PR #936, opened 2015-02-23
GSIP 126, PPIO factories, direction support, and OGR implementation

aaime:
The OGR module has been split in a wfs and a wps part, docs updated accordingly


aaime:
Handled Brad feedback and forced update to keep a clean history.
Merging

PR #939, opened 2015-02-23
Importer rest docs and proper DataStore API usage for dropping schemas

aaime:


aaime:
Escapes removed, merging

jodygarnett:
Thanks @aaime (and @ischneider for review)

PR #941, opened 2015-02-24
GEOS-6905 support WCS 2.0 requests embedded in WPS

jdries:
Added support for embedding WCS 2.0 in WPS, also a basic unit tests which checks that an 1.1 and 2.0.1 xml request can be embedded in a WPS request.

### DIFF
--- a/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/executor/InternalWCSInputProvider.java
+++ b/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/executor/InternalWCSInputProvider.java
@@ -12,6 +12,7 @@ import net.opengis.wps10.MethodType;
 import org.geoserver.ows.KvpRequestReader;
 import org.geoserver.wcs.WebCoverageService100;
 import org.geoserver.wcs.WebCoverageService111;
+import org.geoserver.wcs2_0.WebCoverageService20;
 import org.geoserver.wps.WPSException;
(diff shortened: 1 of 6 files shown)


src/extension/wps/wps-core/src/main/java/org/geoserver/wps/executor/InternalWCSInputProvider.java
```diff
@@ -45,6 +46,9 @@ protected Object getValueInternal(ProgressListener listener) throws Exception {
             KvpRequestReader reader;
             if (version.equals("1.0.0") || version.equals("1.0")) {
                 reader = (KvpRequestReader) context.getBean("wcs100GetCoverageRequestReader");
+                
+            } else if(version.equals("2.0.1") ){
```
aaime: I'd relax this check a bit and also allow for 2.0.0

aaime:
Looking good, just found a minor issue

jdries:
Ok, I added the WCS version. I also had another minor issue with bad handling of unexpected responses from embedded WFS requests, so I simply added that commit within this pull request, hope that's not a problem.

aaime:
Merged, and broke the build due to a missing test dependency, fix here: https://github.com/geoserver/geoserver/commit/a6c84b272562d9353bd95aa8efc8f255ee1debea

Please remember to check that maven builds fine, it does not use the same dependency rules as Eclipse

PR #943, opened 2015-02-24
GEOS-6907 Search wps result item by name or data type

ahuarte47:
This PR request shows as fix the issue https://jira.codehaus.org/browse/GEOS-6907

It needs a junit test, but I only have located the error using an optional sextante process that returns more of one output items (as sxt:kriging process).

### DIFF
--- a/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/executor/InternalWPSInputProvider.java
+++ b/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/executor/InternalWPSInputProvider.java
@@ -53,6 +53,28 @@ public class InternalWPSInputProvider extends AbstractInputProvider {
     @Override
     protected Object getValueInternal(ProgressListener listener) throws Exception {
         Map<String, Object> results = executor.submitChained(executeRequest, listener);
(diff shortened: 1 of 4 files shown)


aaime:
Alvaro, had a look, I see the problem, indeed the code is buggy.
Please do not use so many heuristics, most of the GeoServer code is designed to "fail fast", not to accomodate for user mistakes, I'd suggest to either have a match for the input requested, or fall back on the old behavior for backwards compatibility. 

About writing a test, there is no need to use a real process, just this week I've written a fake process with multiple raw outputs that only runs inside the tests (https://github.com/geoserver/geoserver/blob/master/src/extension/wps/wps-core/src/test/java/org/geoserver/wps/MultiRawProcess.java, full commit here https://github.com/geoserver/geoserver/commit/b827cc3d8c6c8b445e449c175e0abbd01d9815f4 ), I believe you could just use it, or write a similar one, in order to write your own test.

ahuarte47:
Hi Andrea, I provide a test to this fix.

Best Regards
Alvaro

ahuarte47:
Thanks Andrea

PR #926, opened 2015-02-16
[Improvement] Scaling of scale line decoration

onack:
This is nor related to a case, but an improvement that we wanted inhouse. I have tried to make all major changes in separate commits so it should be easy to review. The default double scaleline is not changed(except from the cutting of text), but individual scaleline is a lot prettier so they look more like OpenLayers scaleline.

PR #950, opened 2015-02-25
Improve "scaleline" WMS decoration widget

dalbani:
With this patch, the "scaleline" widget now supports new parameters:
- `strokewidth`
- `borderwidth`
- `padding`
- `suggestedwidth`
- `top-out-unit`
- `top-in-unit`
- `bottom-out-unit`
- `bottom-in-unit`

This makes the use of the widget much more flexible in terms of visual output.

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/decoration/ScaleLineDecoration.java
+++ b/src/wms/src/main/java/org/geoserver/wms/decoration/ScaleLineDecoration.java
@@ -20,13 +20,14 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import org.geoserver.wms.WMSMapContent;
+import org.geotools.util.logging.Logging;
 
 import static org.geoserver.wms.decoration.ScaleLineDecoration.MeasurementSystem.*;
 
 public class ScaleLineDecoration implements MapDecoration {
     /** A logger for this class. */
     private static final Logger LOGGER =
-        org.geotools.util.logging.Logging.getLogger("org.geoserver.wms.responses");
+        Logging.getLogger("org.geoserver.wms.decoration");
 
     private static Map<String, Double>INCHES_PER_UNIT = new HashMap<String, Double> ();
     static {
@@ -75,7 +76,7 @@ public class ScaleLineDecoration implements MapDecoration {
             try {
                 this.fontSize = Float.parseFloat(options.get("fontsize"));
             } catch (Exception e) {
-                this.LOGGER.log(Level.WARNING, "'fontsize' must be a float.", e);
+                LOGGER.log(Level.WARNING, "'fontsize' must be a float.", e);
             }
         }
 
@@ -83,7 +84,7 @@ public class ScaleLineDecoration implements MapDecoration {
             try {
                 this.dpi = Float.parseFloat(options.get("dpi"));
             } catch (Exception e) {
-                this.LOGGER.log(Level.WARNING, "'dpi' must be a float.", e);
+                LOGGER.log(Level.WARNING, "'dpi' must be a float.", e);
             }
         }
 
@@ -91,7 +92,7 @@ public class ScaleLineDecoration implements MapDecoration {
             try {
                 this.strokeWidth = Float.parseFloat(options.get("strokewidth"));
             } catch (Exception e) {
-                this.LOGGER.log(Level.WARNING, "'strokewidth' must be a float.", e);
+                LOGGER.log(Level.WARNING, "'strokewidth' must be a float.", e);
             }
         }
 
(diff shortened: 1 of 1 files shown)


aaime:
Hi, I still ha
See the contributor guide, you need to add a test covering each of the new functionality bits you added: https://github.com/geoserver/geoserver/blob/master/CONTRIBUTING.md

Also, it would be good to document them in the docs: .

aaime:
Bah, sorry, pressed the wrong button

aaime:
Wanted to say that I still haven't looked in the contents of the patch, personally I won't be able to for a few days (other devs feel free to review sooner of course), but from a glance on it tests and docs are missing.

aaime:
Hi, this pull request is hitting similar code areas as #926
Could you work together to solve eventual conflicts?

onack:
I am open to working together on this. I could try to merge our requests if that is ok with you?

snotling:
Sure, please go ahead and do as you wish :)

onack:
I have merged the basics, but I've been thinking. The whole code is a bit misleading when you are adding a 3. type of measurement system. I am not saying adding a 3. is a bad, but we should rewrite the enum system i wrote earlier. More like TOP and BOTTOM or something similar, but if you have just one enabled, its not even top or bottom... I need ideas please.

Other thing is, your implementation of padding doesn't really move the content away from the left side, but I've fixed that part and it looks good now.

I cannot see you have changed anything about tests or documentation for the added features. Do you have them elsewhere?

snotling:
You're right, I didn't provide any change in the tests / doc area.
My patch was more of a quick hack.

onack:
@aaime @snotling 
I have an idea.
How about if we send in an array of units 1 or 2, first one will go to bottom, second one will go to top. We need to define in the code in and out unit for every unit type so it will autoselect accordingly. This will remove a lot of code and makes for easier configuration.

Should default to todays default of [ft,m] (pseudo-code)

Con is that it is not backwards compatible with my imperial/metric earlier pull request.

aaime:
onack, your previous pull request, is it on 2.7.x too? Cause we already released the RC, so we should not break functionality there.

onack:
It was on master before you created 2.7.x i believe.

aaime:
We entered feature freeze for 2.7.x January 18th anyways, so no chance to get these changes there anyways (and 2.7.0 is cut). 
Since your feature is in 2.7.x already, can you think of any way to move this forward without breaking backwards compatibility? Measurement system is there, and documented, so user visible:
http://docs.geoserver.org/latest/en/user/advanced/wmsdecoration.html#decoration-types

aaime:
So I guess this one stalled? Shall I just reference it from a Jira ticket, close it, and wait for someone to resume the work on it?

aaime:
Gah, closed by mistake (pressed wrong button).

onack:
My request does not brake backwards compatibility, but adding more than the existing two(imperial and metric) measurement types will in all cases I can think of brake backwards compatibility because there is no current way of telling where the 3. would go if you enable them all.

jodygarnett:
We have a conflict between this and #950 and #926 how would you both like to proceed?

onack:
I have tried merging, but I cannot do it without breaking backwards compatibility. Hence why I am asking how we would implement multiple(more than two) measurement types. I suggest we do a rewrite that breaks backwards compatibility, but makes it better in the future supporting unlimited types of measurement systems.

aaime:
Normally we try really hard to not do that, because otherwise users try to eat us for breakfast (most people expect to jump 3-4 versions and have everything just working seamlessly without any manual adjustment).
Could we for example leave the old property name there, undocumented but working, and create a new property name for what you're trying to do?

onack:
I can look into that and make a new property and use the old one as a setup for the old default. I dont have time this week to look at it though.

aaime:
I believe we can wait a bit longer :)

onack:
I have a suggestion to solve this here, i have not fixed tests or documentation yet as I want you to have a look at it. https://github.com/onack/geoserver/commit/66fe779aa93a79f8571fa14b4c86cbc20139e3ce

The private classes could be taken out in its own files if desired. With the changes now you can have as many measurement systems as you wish and how many units within a measurement system you want. A lot more flexible than the old solution. I kept the old "measurement-system" config possibilites so it does not break backwards compatibility, but I've also added topunit and bottomunit configuration.

onack:
Fixed some more, left alignment(default and optional), spelling errors, int-casting inaccuracy

aaime:
Hi @onack, I had a very quick look at your branch. I believe the idea is a good one. 
There are a few steps you should take for this to become a pull request and make into the code base:
* Make sure you are following the coding guidelines (spaces, not tabs, for example)
* Squash everything in a single commit
* Add tests for the new options

Eventually we'll need also the documentation, but I'd fix the above and make the pull request so that we can review it better, before using time to also write docs.

aaime:
Oh, also, pretty important, if this does not get merged before August 28th it will become part of the GeoServer 2.9.x feature set, to be released March 2016

jodygarnett:
While I would be happy to help with formatting and squashing into a single commit - I am not familiar with these options enough to quickly write tests for them.

bencaradocdavies:
@onack there has been no activity on this PR for almost a year. Please resubmit as a new pull request to attract attention and trigger a new Travis CI run.

PR #960, opened 2015-03-06
Relaxing Naming Restrictions with STRICT_PATH=false flag [GEOS-6811]

travislbrundage:
This is in response to a naming regression described in the following JIRA ticket: http://jira.codehaus.org/browse/GEOS-6811

The solution introduces a new Java System Property, STRICT_PATH, which if set to false, will relax the restrictions on names. The characters which are allowed with this flag set to false are:

* star (*)

* colon (:)

* comma (,)

* single quote (')

* ampersand (&)
  
* question mark (?)
  
* double quote (")
  
* less than (<)
  
* greater than (>)
  
* bar (|)

This will allow migration from geoserver 2.5.x to future versions for any catalogs which contained these characters.

### DIFF
--- a/doc/en/user/source/datadirectory/migrating.rst
+++ b/doc/en/user/source/datadirectory/migrating.rst
@@ -153,4 +153,33 @@ Backup files
                config.xml.2.2.x
           config.xml.2.2.x        
 
-       
+Migrating between GeoServer 2.5.x and 2.6.x
+-------------------------------------------
+
+The catalog naming conventions became more strict in 2.6, invalidating certain characters within names. This is because certain protocols will not work correctly with certain characters in the name. In 2.6.3 and forward, the naming restrictions are automatically set to relaxed through the STRICT_PATH java system property variable. In order to ensure your names will work with all protocols, set this variable to true. ::
+
+    java -DSTRICT_PATH=true Start
+
+This will invalidate all of the following characters:
+  
+* star (*)
+
+* colon (:)
+
+* comma (,)
+
+* single quote (')
+
+* ampersand (&)
+  
+* question mark (?)
+  
+* double quote (")
+  
+* less than (<)
+  
+* greater than (>)
+  
+* bar (|)
+
+Be warned that some requests or protocols may behave unexpectedly when these characters are allowed. We recommend that you update your catalog and enable strict mode to ensure you follow appropriate naming conventions.
--- a/src/platform/src/main/java/org/geoserver/platform/resource/Paths.java
+++ b/src/platform/src/main/java/org/geoserver/platform/resource/Paths.java
@@ -102,7 +102,41 @@ public class Paths {
         }
         return toPath(names);
     }
-
+    
+    // runtime flag which, if true, throws an error for the WARN characters
+    static final boolean STRICT_PATH = 
+            Boolean.valueOf(System.getProperty("STRICT_PATH", "false"));
+    
+    /**
+     * Pattern used to check for invalid file characters.
+     * <ul>
+     * <li> backslash
+     * </ul>
+     */
+    static final Pattern VALID = Pattern.compile("^[^\\\\]*$");
+    /**
+     * Pattern used to check for ill-advised file characters:
+     * <ul>
+     * <li> star
+     * <li> colon - potential conflict with xml prefix separator and workspace style separator
+     * <li> comma
+     * <li> single quote
+     * <li> ampersand
+     * <li> question mark
+     * <li> double quote
+     * <li> less than
+     * <li> greater than
+     * <li> bar
+     * </ul> 
+     * These characters can cause problems for different protocols.
+     */
+    static final Pattern WARN = Pattern.compile("^[^:*,\'&?\"<>|]*$");
+    /**
+     * Set of invalid resource names (currently used to quickly identify relative paths).
+     */
+    static final Set<String> INVALID = new HashSet<String>(
+            Arrays.asList(new String[] { "..", "." }));
+    
     /**
      * Internal method used to convert a lit of names to a normal Resource path.
      * <p>
@@ -125,6 +159,11 @@ public class Paths {
             }
             if (!VALID.matcher(item).matches()) {
                 throw new IllegalArgumentException("Contains invalid " + item + " path: " + buf);
+            }
+            if (!WARN.matcher(item).matches()) {
+                if (STRICT_PATH == true) {
+                    throw new IllegalArgumentException("Contains invalid " + item + " path: " + buf);
+                }
             }
             buf.append(item);
             if (i < LIMIT - 1) {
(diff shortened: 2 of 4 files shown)


doc/en/user/source/datadirectory/migrating.rst
```diff
@@ -153,4 +153,33 @@ Backup files
                config.xml.2.2.x
           config.xml.2.2.x        
 
-       
\ No newline at end of file
+Migrating between GeoServer 2.5.x and 2.7.x
```
jodygarnett: I think we should backport this change, and change this to GeoServer 2.5.x and GeoServer 2.6.x

src/main/src/main/java/org/geoserver/catalog/impl/CatalogImpl.java
```diff
@@ -1287,7 +1288,7 @@ public StyleInfo getStyleByName(String name) {
             // search by resource name
             String prefix = name.substring(0, colon);
             String resource = name.substring(colon + 1);
-
+
```
jodygarnett: these changes are spurious (white space and a dead import)

src/platform/src/test/java/org/geoserver/platform/resource/PathsTest.java
```diff
@@ -6,6 +6,7 @@
 import java.util.Arrays;
 import java.util.List;
 
+import org.junit.Ignore;
```
jodygarnett: ignore @ignore as it is not used?

src/platform/src/main/java/org/geoserver/platform/resource/Paths.java
```diff
@@ -102,7 +102,41 @@ public static String path(String... path) {
...
+    static final boolean STRICT_PATH = 
+            Boolean.valueOf(System.getProperty("STRICT_PATH", "true"));
+    
+    /**
+     * Pattern used to check for invalid file characters.
+     * <ul>
+     * <li> backslash
+     * </ul>
+     */
+    static final Pattern VALID = Pattern.compile("^[^\\\\]*$");
```
aaime: Backslash is invalid on Windows. Forward slash should be invalid on Linux? http://stackoverflow.com/questions/9847288/is-it-possible-to-use-in-a-filename

src/platform/src/main/java/org/geoserver/platform/resource/Paths.java
```diff
@@ -102,7 +102,41 @@ public static String path(String... path) {
         }
         return toPath(names);
     }
-
+    
+    // runtime flag which, if true, throws an error for the WARN characters
+    static final boolean STRICT_PATH = 
+            Boolean.valueOf(System.getProperty("STRICT_PATH", "true"));
```
aaime: As discussed on the mailing list, this should be false by default since this commit fixes a regression.
But I also understand the desire to have things be more strict moving forward. 
How about making this a property in GeoServerInfo, allowing the users to edit it, and making it "true" in the release data directory, thus enforcing the more strict validation in new installations, while preserving functionality in old ones?

src/platform/src/main/java/org/geoserver/platform/resource/Paths.java
travislbrundage: Good point; updated it to include forward slash as invalid.

src/platform/src/main/java/org/geoserver/platform/resource/Paths.java
travislbrundage: I've set it to false per your suggestion, although I know you and Jody have had a long discussion about how we will interact with the user. Perhaps it is okay to leave it like this for now, and I can re-open discussion on the mailing list for the appropriate way to have the user make this informed choice?

aaime: It is not really back-portable with a "false" setting, which I guess was the main point of the discussion?

travislbrundage:
Need to fix this - apparently the REGEX for validation is used in both the tokenized and raw version of the name, so invalidating "/" is incorrect

travislbrundage:
@aaime Not sure how we can check forward slash to be an invalid character when it will split the file name on that character, treating it as a directory?

Either way, I am not sure that change is necessary here, as this was just meant to update the regression of disallowing certain style names.

jodygarnett:
Yeah, we split the path up based on forward slash - so we are not going to be able to catch forward slash here.

jodygarnett:
From GeoServer meeting:

Want this in, conversation seems to have stopped. This fix allows
GeoServer 2.5 users to update.
Action: merge this in to stable.

doc/en/user/source/datadirectory/migrating.rst
```diff
@@ -153,4 +153,33 @@ Backup files
                config.xml.2.2.x
           config.xml.2.2.x        
 
-       
\ No newline at end of file
+Migrating between GeoServer 2.5.x and 2.6.x
+-------------------------------------------
+
+The catalog naming conventions became more strict in 2.6, invalidating certain characters within names. This is because certain protocols will not work correctly with certain characters in the name. The naming restrictions are automatically set to relaxed through the STRICT_PATH java system property variable. In order to ensure your names will work with all protocols, set this variable to true. ::
```
smithkm: This is a bit hard to follow.  It seems like you mean that the default is relaxed and it can be overridden with the property.  Then you say "This will allow, with warnings".  It's ambiguous whether you mean the relaxed default allows or those characters, or that enabling strict mode enables the warnings.

doc/en/user/source/datadirectory/migrating.rst
```diff
@@ -153,4 +153,33 @@ Backup files
...
+  
+* double quote (")
+  
+* less than (<)
+  
+* greater than (>)
+  
+* bar (|)
+
+Be warned that some requests or protocols may behave unexpectedly when these characters are allowed. We recommend that you update your catalog to follow appropriate naming conventions.
```
smithkm: I'd include "and enable strict mode" to the recommendation, just to be clear.

doc/en/user/source/datadirectory/migrating.rst
smithkm: Also might be worth noting that the property is only available from 2.6.3 onward.

travislbrundage: Yeah, my mistake. I had originally wrote this imagining that it was going to be set to true by default. I need to update that sentence.

doc/en/user/source/datadirectory/migrating.rst
travislbrundage: Another oversight from the fact that I originally assumed it was defaulted to true. Thanks for the suggestion.

smithkm:
The doc issues I pointed out can be dealt with later.  The only issue I see with slash is that it could be included in one of the elements passed to `toPath` which would lead to a path that would not "round trip" when split back apart; however, this issue is present regardless of this change.

So this PR looks good to me.

travislbrundage:
I created a ticket to address the issue noted by Andrea and Kevin: [GEOS-6942](https://jira.codehaus.org/browse/GEOS-6942)

Fixed the minor doc inconsistencies.

PR #961, opened 2015-03-06
Fixing unused/broken test

travislbrundage:
The testMissingName() test in StyleNewPageTest was not actually being executed before. Further, it's call to form.setValue() was incorrect and caused a failure. The test has been updated to run and passes correctly.


jodygarnett:
Appled to master and 2.7.x.

PR #962, opened 2015-03-06
GEOS-6059 Allow user specification of aliases for localhost

smithkm:
Adds mechanism to specify hostnames that are aliases for the local host as well as logging to aid in diagnosing issues.

Provided unit tests for aliases and proxy baseurl but was unable to mock the NetworkInterface discovery as NetworkInterface is final and has no public constructor.

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/xml/WFSURIHandler.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/xml/WFSURIHandler.java
@@ -13,9 +13,14 @@ import java.net.InetAddress;
 import java.net.NetworkInterface;
 import java.net.SocketException;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.Enumeration;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.logging.Level;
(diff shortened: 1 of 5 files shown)


smithkm:
https://github.com/geotools/geotools/pull/767 reduces 6059 from locking up GeoServer to just being an error for a specific request.  This patch provides a means to resolve the issue on a case by case basis by appending aliases to the list of known ones.

src/wfs/src/main/java/org/geoserver/wfs/xml/WFSURIHandler.java
```diff
@@ -51,32 +58,95 @@
...
+                builder.append(" ");
+            }
+            LOGGER.log(Level.INFO, builder.toString());
         }
     }
 
+    private static void initAliases() {
+        assert ADDITIONAL_HOSTNAMES.isEmpty();
+        // User configurable hostnames
+        String additional=GeoServerExtensions.getProperty(WFSURIHandler.class.getName()+".additionalHostnames");
```
aaime: Hi, do you have the coding conventions loaded up? This should read "String additional = GeoServerExtensions..."

aaime:
Looks good besides minor formatting issues

smithkm:
Yeah I was bouncing between machines while working on this, one of which didn't have the formatting config intalled.    Fixed now.

PR #963, opened 2015-03-08
GEOS-6608: Start removing links to confluence from docs.

bradh:
This address part of, not all of, GEOS-6608. More work is required, ideally before codehaus disappears.

Also fix a few typos noted along the way.


aaime:
Thanks

PR #964, opened 2015-03-08
[GEOS-6923] Slow JSON response on EPSG:900913 defined vector layer

aaime:
Pure performance optimization, thus no tests (the existing ones will guard against regressions)


aaime:
Oh, the reason the change speeds up things is that CRS.lookupEPSGCode will internally leverage the CRS internal identifiers, if avaialble.

simboss:
backport please :)

smithkm:
Looks good to me.

NBroomfield:
Just FYI it looks like you have a superfluous nested if statement that has the same condition at the parent if statement.

PR #965, opened 2015-03-09
[GEOS-6842] Add option to enable Dbase filters with ODBC driver

ahuarte47:


src/web/core/src/main/resources/GeoServerApplication.properties
```diff
@@ -478,6 +478,7 @@ ShapefileStoreEditPanel.charset=DBF charset
 ShapefileStoreEditPanel.memoryMapped=Use memory mapped buffers (Disable on Windows)
 ShapefileStoreEditPanel.spatialIndex=Create spatial index if missing/outdated
 ShapefileStoreEditPanel.cacheMemoryMaps=Cache and reuse memory maps (Requires 'Use Memory mapped buffers' to be enabled)
+ShapefileStoreEditPanel.filterWithOdbcDbaseDriver=Use an ODBC provider to fast execution of Dbase filters (Now it only works for "Microsoft ODBC FoxPro Driver (x86)" or "Advantage StreamlineSQL ODBC driver (x86/x64)" drivers running in Windows platforms)
```
aaime: I'd reword it to say:

Currently supports Microsoft ODBC FoxPro Driver (x86)" or "Advantage StreamlineSQL ODBC driver (x86/x64)" drivers running on Windows platform only

ahuarte47: Done, thank you very much Andrea!

ahuarte47:
This PR depends of other PR in GeoTools project ( https://github.com/geotools/geotools/pull/816 )

aaime:
Hi Alvaro, I'm afraid I have bad news, I'll have to revert your original pull request and close this one too.
It pains me, also because I've just spent half a day fixing this pull request (the way it was coded, it was leaking connection pools, much worse than the original connection leak, and basically had no tests, so in the state it was, I would have had to revert it anyways).

The reason for reverting it instead of keep on trying to fix it like I did today is simple: Oracle removed the JDBC-ODBC bridge permanently from Java 8 onwards, and we are about to switch to Java 8 anyways: https://blogs.oracle.com/Lance/entry/removal_of_the_jdbc_odbc

So this code is dead on arrival, Java 7 has reached the end of life in May and there are no more security releases for it, and Java 8 does not have JDBC-ODBC anymore.

I have spent some time checking if there are pure java B+Tree implementations that could allow indexing both equality and range queries, the closes thing to anything useful I've found is jdbm, the original version in particular, as jdbm2, jdbm3, and mapdb seem to have spun out of control, without usable releases in a lot of time. If you need BDF indexing I'd recommend looking into one of them.

ahuarte47:
Hi Andrea, I honestly feel the problems caused by my pull request. Do not worry, I'll look at other ways of indexing when we have to migrate the geoserver for our client using this capability. I appreciate your interest and ask again apologize.

Cheers
Alvaro

PR #966, opened 2015-03-09
Copyediting composite blend docs

bmmpxf:
Split into multiple pages, reordered, and updated text for added clarity.


aaime:
The order of the items seems wrong to me, like, you should see the blending modes before you get the syntax to specify them, or if you get the syntax first, you should get a list of the possible values, and then a link to the page with the visual examples.
Also, in the top page, the source and destination bits are showing on a single line, but they look like they were though out to be a unnumbered list instead.

bmmpxf:
Thanks Andrea. The thinking was to explain how everything works before being shown every possible option. But you're correct that that's not usually how we do it in the GS reference, so I'm happy to move from syntax/example/modes to syntax/modes/example. I don't think throwing the big list of modes out first is the right move. When I read it like that my first thought was "yeah great, but how/where do I do all this?" This way seems like a nice compromise.

And good catch on the unnumbered list.

PR #970, opened 2015-03-11
Fix an infinite loop in GeoServerPersister.renameStyle [GEOS-6983]

mproboeuf:
Variable "i" was never incremented while testing the existence of sld file.
It causes an infinite loop when file1.sld already exists.

Update: https://osgeo-org.atlassian.net/browse/GEOS-6983


simboss:
test please :)

jodygarnett:
While your fix seems fine the code itself seems to assume an "sld" file. Should use the style extension to account for sld and css files during rename ...

jodygarnett:
Revised as #1015 closing this pull request.

PR #974, opened 2015-03-14
Updated citewcs README

tbarsballe:
Updated the readme to match current behavior


tbarsballe:
Due to some strangeness of how the CITE tests work, this fix may not be valid. Please wait to merge until this can be further verified.

PR #980, opened 2015-03-19
Fix for 2.7-RC1 security regression [GEOS-6943]

n-lagomarsini:
For details see: https://jira.codehaus.org/browse/GEOS-6943


src/main/src/test/java/org/geoserver/security/impl/AbstractAuthorizationTest.java
```diff
@@ -208,6 +209,7 @@ protected LayerInfo buildLayer(String name, WorkspaceInfo ws,
 
         LayerInfo layer = createNiceMock(LayerInfo.class);
         expect(layer.getName()).andReturn(name).anyTimes();
+        expect(layer.getPrefixedName()).andReturn(ws.getName() + ":" + name).anyTimes();
```
jodygarnett: Matches up with the GEOS-6943 description.

jodygarnett:
Thanks @n-lagomarsini testing on master before back port.

PR #982, opened 2015-03-19
Improved strict path testability and coverage

bencaradocdavies:
@jodygarnett @travislbrundage this pull request refactors strict path handling to permit testing both with and without strict path, and increases test coverage. (Current tests depend on STRICT_PATH which will always be false, so strict path behaviour is never tested.) Full build passes on master. Not yet tested on 2.7.x or 2.6.x but backport would improve maintainability.

### DIFF
--- a/src/platform/src/main/java/org/geoserver/platform/resource/Paths.java
+++ b/src/platform/src/main/java/org/geoserver/platform/resource/Paths.java
@@ -93,6 +93,17 @@ public class Paths {
      * @return path Path used to identify a Resource
      */
     public static String path(String... path) {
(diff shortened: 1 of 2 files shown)


jodygarnett:
Thanks Ben, that looks good ... although I really feel we are doing this in
the wrong spot :(

We need to catch these names on creation (gui or rest) rather than at the
path level when it is in a sense already too late.



--
Jody Garnett

bencaradocdavies:
Thanks, @jodygarnett. I agree that validation at the UI will give a better user experience.

PR #983, opened 2015-03-19
Add hazelcast cluster module to community extensions build

smithkm:
Build a plugin zip for the Hazelcast clustering community module

### DIFF
--- a/src/community/pom.xml
+++ b/src/community/pom.xml
@@ -55,6 +55,7 @@
           <descriptor>release/ext-gpx.xml</descriptor>
           <descriptor>release/ext-wps-download.xml</descriptor>
           <descriptor>release/ext-jms-cluster.xml</descriptor>
+          <descriptor>release/ext-hz-cluster.xml</descriptor>
           <descriptor>release/ext-activeMQ.xml</descriptor>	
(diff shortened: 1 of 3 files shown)


src/community/pom.xml
```diff
@@ -55,6 +55,7 @@
           <descriptor>release/ext-gpx.xml</descriptor>
           <descriptor>release/ext-wps-download.xml</descriptor>
           <descriptor>release/ext-jms-cluster.xml</descriptor>
+          <descriptor>release/ext-hz-cluster.xml</descriptor>
```
jodygarnett: adding it to the community profile for nightly builds

src/community/pom.xml
```diff
@@ -181,10 +181,12 @@
         <module>kmlppio</module>
         <module>wps-download</module>
         <module>jms-cluster</module>
+        <module>hz-cluster</module>
         <module>solr</module>
 	<module>rest-upload</module>
         <module>geofence</module>
         <module>sldService</module>
+        <module>release</module>
```
jodygarnett: moved release to end? makes it a bit more readable I guess.

jodygarnett:
We can back port this after the release goes out? Don't suppose it matters as the release is not affected...

PR #986, opened 2015-03-25
GEOS-6905 2.7.x

jdries:
Backport of GEOS-6905 to 2.7.x, including fix for missing test dependency


aaime:
Jeroen, we are migrating Jira to another host, if you could you remember to update the "fix for" date once the new Jira instance is up that would be fantastic

PR #987, opened 2015-03-25
GEOS-6947: added support for custom variables from various sources to HT...

mbarto:
...ML GetFeatureInfo ftl templates


jodygarnett:
Thanks for the docs @mbarto - update the headers to 2015 and we are good to go :)

PR #988, opened 2015-03-25
[GEOS-6949] Fixed 16 showing twice

tbarsballe:
Fix for https://jira.codehaus.org/browse/GEOS-6949


aaime:
Oh wow, this one is so trivial it does not make sense to ask for a test (imho).

tbarsballe:
If there are no other concerns, I will merge and backport then.

PR #989, opened 2015-03-26
[GEOS-6958] Use relative paths when importing from data directory

tbarsballe:
See https://jira.codehaus.org/browse/GEOS-6958

While this is technicaly a bugfix, it affects the Catalog, so it might be an idea to keep this fix just on master for a bit before any backporting is considered.

In addition to the automated tests, I have tested this manually, and have not seen any detrimental effects to the catalog.

### DIFF
--- a/src/extension/importer/core/src/main/java/org/geoserver/importer/DataStoreFormat.java
+++ b/src/extension/importer/core/src/main/java/org/geoserver/importer/DataStoreFormat.java
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2015 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
@@ -81,7 +81,7 @@ public class DataStoreFormat extends VectorFormat {
     }
 
     public DataStoreInfo createStore(ImportData data, WorkspaceInfo workspace, Catalog catalog) throws IOException {
-        Map<String,Serializable> params = createConnectionParameters(data);
+        Map<String,Serializable> params = createConnectionParameters(data, catalog);
         if (params == null) {
             return null;
         }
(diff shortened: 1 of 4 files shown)


src/extension/importer/core/src/main/java/org/geoserver/importer/DataStoreFormat.java
```diff
@@ -191,7 +191,7 @@ public int getFeatureCount(ImportData data, ImportTask task) throws IOException
     public DataStore createDataStore(ImportData data) throws IOException {
         DataStoreFactorySpi dataStoreFactory = factory();
 
-        Map<String,Serializable> params = createConnectionParameters(data);
+        Map<String,Serializable> params = createConnectionParameters(data, null);
```
tbarsballe: createDataStore(ImportData data) seems to only ever be used for lightweight tasks such as verifying format, listing layers, and indirect import. Data stores created using this method are not added to the catalog, (in part because this seems to mostly be used where the catalog is not available), so we are safe using an absolute path here.

simboss:
Torben,
I would back port right away to 2.7.x and then wait a bit for 2.6.x.

By experience not many people will be using 2.7.x atm and even less are on master. If we need to test this fix a little we need to go back to 2.7.x at least.

My 0.02€.

tbarsballe:
That sounds reasonable Simone, especially since importer is an optional extension.

aaime:
Hi Torben, a bit confused here.... all your changes are inside the importer module, what do you mean by "it affects the Catalog"? The Catalog interface does not seem to be modified?

tbarsballe:
It doesn't have any direct affect on the catalog, but the changes to importer alter what is actually stored in the catalog slightly by changing the file URL of the StoreInfo object. I am mostly concerned about cases where this change of behaviour could lead to broken URL's being stored in StoreInfo and persisted to the catalog. This doesn't seem to be the case, but still seems like it could be possible.

tbarsballe:
Merged. Also, tested on 2.7.x and backported so that we can get some field testing. Holding off backporting to 2.6.x.

PR #990, opened 2015-03-27
1. Disabling clicking on labels and far-right whitespace to toggle check...

eshon:
Fix provided:

1. Disabling clicking on labels and far-right whitespace to toggle checkboxes
2. Removing closing input tags because input tags do not have closing tags 

For https://jira.codehaus.org/browse/GEOS-6948

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/data/layer/CascadedWFSStoredQueryAbstractPage.html
+++ b/src/web/core/src/main/java/org/geoserver/web/data/layer/CascadedWFSStoredQueryAbstractPage.html
@@ -18,7 +18,7 @@
 
 <wicket:fragment wicket:id="newPage">
   <div>
-    <wicket:message key="chooseNativeName">Layer name </wicket:message> <input type="text" wicket:id="nativeName"></input>
+    <wicket:message key="chooseNativeName">Layer name </wicket:message> <input type="text" wicket:id="nativeName">
   </div>
   <div>
     <wicket:message key="selectStoredQuery">Select Stored Query </wicket:message> <select wicket:id="storedQueriesDropDown"></select>
@@ -33,7 +33,7 @@
 </wicket:fragment>
 
 <wicket:fragment wicket:id="parameterMappingValue">
-  <input type="text" wicket:id="text"></input>
+  <input type="text" wicket:id="text">
 </wicket:fragment>
 
 </wicket:extend>
--- a/src/web/core/src/main/java/org/geoserver/web/services/BaseServiceAdminPage.java
+++ b/src/web/core/src/main/java/org/geoserver/web/services/BaseServiceAdminPage.java
@@ -55,16 +55,16 @@ import org.geoserver.web.wicket.LiveCollectionModel;
  *   &lt;li>
  *       &lt;span>
  *         &lt;label><wicket:message key="maxFeatures.title">Maximum Features</wicket:message></label>
- *         &lt;input wicket:id="maxFeatures" class="field text" type="text"></input>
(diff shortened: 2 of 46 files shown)


jodygarnett:
Hi Eva! Is there a jira issue or something for this one? I am familiar (and frustrated by this issue).

src/web/core/src/main/java/org/geoserver/web/services/BaseServiceAdminPage.java
```diff
@@ -55,16 +55,16 @@
  *   &lt;li>
```
jodygarnett: Update header to say (c) 2014-2015

src/web/core/src/main/java/org/geoserver/web/services/BaseServiceAdminPage.java
```diff
@@ -55,16 +55,16 @@
...
  *       &lt;p class="instruct">
  *       &lt;/p>
  *     &lt;/li>
- *   
+ *
  * </wicket:extend>
  *   </pre>
  * </p>
- * 
+ *
```
jodygarnett: lots of small white space changes

jodygarnett:
So this looks good to me, I would like to manually merge to master (so I can clean up the headers) and then backport to stable. For reference Eva is covered under boundless contribution agreement.

Anyone else want to review this one?

aaime:
Jody, any chance all the whitespace changes can be removed as well?

aaime:
Btw, just to clarify, I can see the whitespace changes are improvements too (removal of extra space at endlines and so on), it's just that in order to backport to the stable series a line by line check would be a good thing to do, and like this the change is massive.

aaime:
Jody, maybe during your manual merge you can do that line by line check? ;-)

src/web/core/src/main/java/org/geoserver/web/data/layer/CascadedWFSStoredQueryAbstractPage.html
```diff
@@ -18,7 +18,7 @@
 
 <wicket:fragment wicket:id="newPage">
   <div>
-    <wicket:message key="chooseNativeName">Layer name </wicket:message> <input type="text" wicket:id="nativeName"></input>
+    <wicket:message key="chooseNativeName">Layer name </wicket:message> <input type="text" wicket:id="nativeName">
```
jodygarnett: Why are we removing the training </input> ?

Should this be:

````
<input type="text" wicket:id="nativeName"/>
````

Comment (aaime) [inline src/web/core/src/main/java/org/geoserver/web/data/layer/CascadedWFSStoredQueryAbstractPage.html]: From what I read <input> is valid in HTML, if you want valid XHTML is should be <input/> instead.
I'm not sure, but believe that some of our CSS was depending on the browser not to enter in quircks mode, and wondering if using <input> alone would do that (sorry, little web frontend experience...)

Comment (ohmu808) [inline src/web/core/src/main/java/org/geoserver/web/data/layer/CascadedWFSStoredQueryAbstractPage.html]: Not clear on what you mean by "training".
Yes, it should be <input /> but was just trying to focus on getting this ticket done and most browsers are ok with just <input>.

Comment (jodygarnett) [inline src/web/core/src/main/java/org/geoserver/web/data/layer/CascadedWFSStoredQueryAbstractPage.html]: Training was a typo, updated my comment above. I am going through and removing whitespace locally.

Only one that has given me pause is:
```
<input type="radio">mode 1</input>
```

Being turned into:
```
<input type="radio">mode 1
```

jodygarnett:
Replaced by #996

PR #992, opened 2015-03-29
[GEOS-6964] Collapse zero-length date ranges to just a date when formatting

bradh:
Also adds unit tests for existing ISO8601Formatter functionality.

### DIFF
--- a/src/main/src/main/java/org/geoserver/util/ISO8601Formatter.java
+++ b/src/main/src/main/java/org/geoserver/util/ISO8601Formatter.java
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2015 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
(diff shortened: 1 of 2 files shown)


aaime:
Hum... I don't have the spec in front of my but wondering, is it going to be a problem if we end up mixing ranges and single values in the same list? Say that there are legit ranges, but mixed among them, some single value ones.

bradh:
I had a look at the OGC best practice for time/elevation data (which basically "clarifies" WMS 1.3 spec in this area, but I couldn't see anything specific). On review of WMS 1.3 (OGC 06-042) Table C.2, I think that it would *not* be valid to have a mix of ranges and single values.

I'll have to think harder about this, almost certainly adding a formatter (or simplifier) at the whole dimension level.

aaime:
I'd look into a fix in the netcdf reader if that is possible, it might have more information to decide how to present the data without having to scan the whole collection of values every time.
As an alternative, it could be a flag in GeoServer that forces it to flatten ranges to single values but... it kind of seems like a workaround for a problem that should be solved elsewhere.

bradh:
I did look at netcdf, but its coupled to a TemporalDomain interface that only returns SortedSet<DateRange>, which makes it difficult...

PR #993, opened 2015-03-29
Fixing security related build failure on windows

aaime:


aaime:
Manually checked it build on linux too, merging

PR #994, opened 2015-03-30
GEOS-5857: Fix StringIndexOutOfBoundsException on some WFS-T requests

awaterme:
See discussion here: https://jira.codehaus.org/browse/GEOS-5857

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/WFSException.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/WFSException.java
@@ -95,18 +95,7 @@ public class WFSException extends ServiceException {
             }
 
             if (locator == null) {
-                //default to name of operation, use the request object class name to determine the
-                // operation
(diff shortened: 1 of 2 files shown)


aaime:
Hi, the change looks good, but you are using tabs for indentation, our coding convention requires usage of spaces

src/wfs/src/main/java/org/geoserver/wfs/WFSException.java
```diff
@@ -116,6 +105,32 @@ else if (className.endsWith("TypeImpl")) {
         return this;
     }
 
+	private String determineDefaultLocator(Object request) {
```
jodygarnett: If you changed the visibility of this method you could write tests cases directly for it.

jodygarnett:
Cleaned up the formatting and merged to master and 2.7.x.

Sorry I cannot close the Jira issue.

PR #995, opened 2015-03-30
Updating community module process

travislbrundage:
Updating the docs for contributing a community module. Some links were outdated and it needed to be updated to use git instead of subversion.


jodygarnett:
Thanks for the update, and good catch on linking to the outdated contribution agreement.

PR #991, opened 2015-03-27
GEOS-6960: Updating GS code after viewsManager cleanup (GEOT-5050)

dromagnoli:
Make sure to not merge it before this one:
https://github.com/geotools/geotools/pull/793

PR #997, opened 2015-03-31
GSIP-128 JAIEXT integration in GeoServer

n-lagomarsini:
Please look at the following Proposal: https://github.com/geoserver/geoserver/wiki/GSIP-128---JAI-EXT-integration-in-GeoServer

Notice that the following PR will require that https://github.com/geotools/geotools/pull/799 and https://github.com/geoserver/geoserver/pull/991 have already been merged.


jodygarnett:
@bmmpxf can you help review this one as it some doc changes.

@n-lagomarsini the screen snap provided is very wide for our docs - not sure if mike has any guidance on this?

src/community/dds/src/main/java/org/geoserver/wms/worldwind/util/BilWCSUtils.java
```diff
@@ -41,34 +41,36 @@
...
+    //
+    // ///////////////////////////////////////////////////////////////////
+    private static final CoverageProcessor processor = CoverageProcessor.getInstance();
+    
+    //private final static SelectSampleDimension bandSelectFactory = new SelectSampleDimension();
+    //private final static Crop cropFactory = new Crop();
+    //private final static Interpolate interpolateFactory = new Interpolate();
+    //private final static Scale scaleFactory = new Scale();
+    //private final static FilteredSubsample filteredSubsampleFactory = new FilteredSubsample();
+    //private final static Resample resampleFactory = new Resample();
```
jodygarnett: is it worth keeping commented out code?

src/community/dds/src/main/java/org/geoserver/wms/worldwind/util/BilWCSUtils.java
```diff
@@ -106,13 +108,13 @@ public static GridCoverage2D reproject(GridCoverage2D coverage,
              * Operations.DEFAULT.resample( coverage, targetCRS, null,
              * Interpolation.getInstance(Interpolation.INTERP_NEAREST))
              */
-            final ParameterValueGroup param = (ParameterValueGroup) resampleParams.clone();
+            final ParameterValueGroup param = (ParameterValueGroup) processor.getOperation("Resample").getParameters().clone();
```
jodygarnett: Side comment: the use of clone in the geotools api for parameter appears dated.

n-lagomarsini:
Hi Jody,

I have reduced the image size as you requested.

Regards,
Nicola

src/main/src/main/java/org/geoserver/config/impl/JAIEXTInfoImpl.java
```diff
@@ -0,0 +1,90 @@
...
+    public static final TreeSet<String> JAIEXT_OPS = new TreeSet<String>();
+    
+    private Set<String> jaiExtOperations = JAIEXT_OPS;
+
+    static{
+        JAIExt.initJAIEXT();
+        List<OperationItem> jaiextOps = JAIExt.getJAIEXTOperations();
+        for(OperationItem item : jaiextOps){
+            String name = item.getName();
+            if (name.equalsIgnoreCase("algebric") || name.equalsIgnoreCase("operationConst")
```
aaime: Hum... one more or and you don't need two branches? Or do we have a bug in the branches code? (It's the same)

src/main/src/main/java/org/geoserver/config/impl/JAIEXTInfoImpl.java
```diff
@@ -0,0 +1,90 @@
...
+
+    public JAIEXTInfoImpl() {
+        if(jaiOperations == null ){
+            jaiOperations = JAI_OPS;
+        }
+        if(jaiExtOperations == null ){
+            jaiExtOperations = JAIEXT_OPS;
+        }
+        List<OperationItem> jaiextOps = JAIExt.getJAIEXTOperations();
+        for(OperationItem item : jaiextOps){
```
aaime: This code duplicates the static initializer, please factor out and share?

src/web/core/src/main/java/org/geoserver/web/admin/JAIEXTPanel.java
```diff
@@ -0,0 +1,71 @@
...
+        add(jaiextSelector);
+    }
+
+    
+    static class ChoiceRenderer implements IChoiceRenderer<String>{
+
+        @Override
+        public Object getDisplayValue(String object) {
+            if(object.equalsIgnoreCase("Stats")){
+                return "Statistics";
```
aaime: These should be internationalized too

src/wms/src/main/java/org/geoserver/wms/map/GeoTIFFMapResponse.java
```diff
@@ -81,8 +87,21 @@ public void formatImageOutputStream(RenderedImage image, OutputStream outStream,
         image = applyPalette(image, mapContent, IMAGE_GEOTIFF8, false);
         
         // crating a grid coverage
-        final GridCoverage2D gc = factory.create("geotiff", image,
+        GridCoverage2D gc = factory.create("geotiff", image,
                 new GeneralEnvelope(mapContent.getRenderingArea()));
+        
+        // NoData stuff
+        if(image instanceof PlanarImage){
```
aaime: Why do we need to add this information? Is the writer not able to do its work without the nodata property? This would be a backwards compatibility problem, the writer must still accept coverages without nodata info

src/wms/src/main/java/org/geoserver/wms/map/PNGMapResponse.java
```diff
@@ -102,7 +98,8 @@ public void formatImageOutputStream(RenderedImage image, OutputStream outStream,
         }
         
         // check to see if we have to see a translucent or bitmask quantizer
-        image = applyPalette(image, mapContent, "image/png8", true);
+        image = applyPalette(image, mapContent, "image/png8", true);
```
aaime: This file changes seem pure whitespace?

src/wms/src/main/java/org/geoserver/wms/map/RenderedImageMapOutputFormat.java
```diff
@@ -736,9 +736,12 @@ private static RenderedImage optimizeSampleModel(RenderedImage source) {
         layout.setTileWidth(w);
         layout.setTileHeight(h);
         RenderingHints hints = new RenderingHints(JAI.KEY_IMAGE_LAYOUT, layout);
+        LookupTable table = LookupTableFactory.create(IDENTITY_TABLE);
         // TODO SIMONE why not format?
-        return LookupDescriptor.create(source, IDENTITY_TABLE, hints);
-
+        ImageWorker worker = new ImageWorker(source);
+        worker.setRenderingHints(hints);
+        worker.looukp(table);
```
aaime: Oops typo, lookup, not looukp

src/wms/src/main/java/org/geoserver/wms/map/RenderedImageMapOutputFormat.java
```diff
@@ -1139,7 +1148,10 @@ protected GridCoverage2D readCoverage(GridCoverage2DReader reader,
                     }
                 } else {
                     if (transparent) {
-                        image = addAlphaChannel(image);
+                        final ImageWorker iw = new ImageWorker(image);
```
aaime: Wouldn't it be better to have back a addAlphaChannel method, and avoid having two image workers with different names in the same context? (worker and iw)

src/wms/src/main/java/org/geoserver/wms/map/GeoTIFFMapResponse.java
n-lagomarsini: This information is added only in case NoData has been defined as a property for the input image, otherwise everything will be left untouched.

PR #999, opened 2015-04-02
[GEOS-6036] Remove unnecessary span elements from UI

smithkm:
Fix for [GEOS-6036](http://jira.codehaus.org/browse/GEOS-6036) by removing span elements that weren't needed but were interfering with Firefox.


jodygarnett:
Change seems straightforward, would you like to merge this to master and ask Jukka to test?

PR #1000, opened 2015-04-02
Test ignore. Travis CI health check before bouncycastle jar signing work...

bencaradocdavies:
...around.


bencaradocdavies:
Ooh, ooh, I got PR #1000! But this is just a test. Under no circumstances merge this PR.

bencaradocdavies:
The test worked: with the most recent travis.yml changes, openjdk7 passed and the oraclejdk builds failed only on the gs-web-sec-ldap unsigned jar problem. See https://github.com/geoserver/geoserver/pull/1001 for the fix (I hope).

PR #1006, opened 2015-04-07
GEOS-6970 Support store creation in Import

mm64z:
If the store does not exist but a store name is provided, create it.
The functionality is virtually identical to not providing a store name,
  except the provided store name is used instead of a default

### DIFF
--- a/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/gs/ImportProcess.java
+++ b/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/gs/ImportProcess.java
@@ -137,9 +137,21 @@ public class ImportProcess implements GSProcess {
         		storeInfo = catalog.getCoverageStoreByName(ws.getName(), store);
         	}
             if (storeInfo == null) {
-                throw new ProcessException("Could not find store " + store + " in workspace "
-                        + workspace);
-                // TODO: support store creation
+                // mirroring "features != null" below
+                if (features != null) {
+                    storeInfo = catalog.getDefaultDataStore(ws);
+                    if (storeInfo == null) {
+                        throw new ProcessException("Could not find a default store in workspace "
+                                + ws.getName());
+                    }
+                }
+                else if (coverage != null) {
+                    // since the store doesn't exist, create it
+                    // mirroring "create a new coverage store" below
+                    storeInfo = cb.buildCoverageStore((store));
+                    add = true;
+                    LOGGER.info("Creating store " + store + " since it did not exist");
+                }
             }
         } else if (features != null) {
             storeInfo = catalog.getDefaultDataStore(ws);
--- a/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/gs/ImportProcess.java
+++ b/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/gs/ImportProcess.java
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014-2015 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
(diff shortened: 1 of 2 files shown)


src/extension/wps/wps-core/src/main/java/org/geoserver/wps/gs/ImportProcess.java
```diff
@@ -137,9 +137,21 @@ else if (coverage != null)
         		storeInfo = catalog.getCoverageStoreByName(ws.getName(), store);
         	}
             if (storeInfo == null) {
-                throw new ProcessException("Could not find store " + store + " in workspace "
-                        + workspace);
-                // TODO: support store creation
+                // mirroring "features != null" below
```
jodygarnett: Usual feedback, if you touch the file update the header with the latest year.

So this file should now say:

````
/* (c) 2014-2015 Open Source Geospatial Foundation - all rights reserved
 * (c) 2001 - 2013 OpenPlans
 * application directory.
 */
````

File: src/extension/wps/wps-core/src/test/java/org/geoserver/wps/gs/ImportProcessTest.java [hunk 2]
```diff
@@ -31,6 +35,10 @@
     @After 
     public void removeNewLayers() {
         removeLayer(SystemTestData.CITE_PREFIX, "Buildings2");
+        removeLayer(SystemTestData.CITE_PREFIX, "Buildings4");
+        removeLayer(SystemTestData.CITE_PREFIX, "Buildings5");
+        removeStore(SystemTestData.CITE_PREFIX, SystemTestData.CITE_PREFIX + "data");
+        removeStore(SystemTestData.CITE_PREFIX, SystemTestData.CITE_PREFIX + "raster");
```
jodygarnett: I cannot remember how specific the CITE test data is (like is it only for CITE testing).

I guess in this case you are just cleaning up after layers you may of added during testing ...

src/extension/wps/wps-core/src/test/java/org/geoserver/wps/gs/ImportProcessTest.java
```diff
@@ -103,4 +111,42 @@ private void checkBuildings(String result,String expected) throws IOException {
...
+        SimpleFeatureCollection rawSource = (SimpleFeatureCollection) ti.getFeatureSource(null,
+                null).getFeatures();
+        ForceCoordinateSystemFeatureResults sampleData = new ForceCoordinateSystemFeatureResults(
+                rawSource, CRS.decode("EPSG:4326"));
+        String storeName = SystemTestData.CITE_PREFIX + "data";
+        DataStoreInfo storeInfo = catalog.getDataStoreByName(storeName);
+        assertNull("Store already exists " + storeInfo, storeInfo);
+        ImportProcess importer = new ImportProcess(getCatalog());
+        String result = importer.execute(sampleData, null, SystemTestData.CITE_PREFIX, storeName,
+                "Buildings5", CRS.decode("EPSG:4326"), null, null);
```
jodygarnett: So this is direct testing of the import process. Do we bother testing these through WPS protocol?

src/extension/wps/wps-core/src/test/java/org/geoserver/wps/gs/ImportProcessTest.java
mm64z: Right, I saw the other tests were deleting the layer they used, so I followed that example and deleted what the tests created.

aaime: Not cite specific, it's just test data that we use in all tests

src/extension/wps/wps-core/src/test/java/org/geoserver/wps/gs/ImportProcessTest.java
aaime: Would be nice, but none of the other tests do that, so I'm ok with the new one ignoring the protocol too.

aaime:
Works for me, merging

PR #1010, opened 2015-04-08
[GEOS-6971] Color-blending example doesn't validate in SLD

smithkm:
Fix for [GEOS-6971](https://osgeo-org.atlassian.net/browse/GEOS-6971) by updating schema file.  Also updated test data to cover this case.


smithkm:
Will backport once merged.

jodygarnett:
So we just have a straight up duplication here with the geotools schema.

Since this is a src/main/resources/ file could we access the content out of the geotools jar dependency instead?

jodygarnett:
Merging to master, please back port to stable.

We can take further discussion to the email list now that you have found the duplication.

PR #1012, opened 2015-04-14
Changes to compile wps-sextante module with sextante libs 1.0 (update)

ahuarte47:
New changes to compile the wps-sextante plugin using last master branch


ahuarte47:
Thanks Andrea!!!

PR #1014, opened 2015-04-16
[GEOS-6979] GWC caching with CQL_FILTER returns no parameter filter exis...

n-lagomarsini:
...ts for FILTER

Look at the following JIRA: https://osgeo-org.atlassian.net/browse/GEOS-6979

### DIFF
--- a/src/gwc/src/main/java/org/geoserver/gwc/GWC.java
+++ b/src/gwc/src/main/java/org/geoserver/gwc/GWC.java
@@ -835,7 +835,8 @@ public class GWC implements DisposableBean, InitializingBean, ApplicationContext
             }
         }
         if (null != request.getFilter() && !request.getFilter().isEmpty()) {
-            if (!filterApplies(filters, request, "FILTER", requestMistmatchTarget)) {
+            boolean sameFilters = checkFilter(request.getFilter(), request.getCQLFilter(), filters);
+            if (!sameFilters && !filterApplies(filters, request, "FILTER", requestMistmatchTarget)) {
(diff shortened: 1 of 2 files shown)


src/gwc/src/main/java/org/geoserver/gwc/GWC.java
```diff
@@ -887,6 +888,41 @@ boolean isCachingPossible(TileLayer layer, GetMapRequest request,
...
+            // In this case the Filter List has been taken from the CQL_FILTER list
+            if (hasCQLFilter && hasFilter) {
+                // First check on the size
+                int size = filter.size();
+                if (size == cqlFilter.size()) {
+                    // Check same elements
+                    boolean equals = true;
+                    // Loop on the elements
+                    for (int i = 0; i < size; i++) {
+                        equals &= filter.get(i).equals(cqlFilter.get(i));
```
jodygarnett: What are these list contents? Do we need to call simplify on the filter before checking?

smithkm:
Looks good.

src/gwc/src/main/java/org/geoserver/gwc/GWC.java
n-lagomarsini: Hi @jodygarnett ,

the contents of the 2 lists are the filters and cql_filters defined in the input WMS requests. I would only tests if the contents of the two lists are equals in order to check if the FILTER list has been copied by the CQL_FILTER list.

Regards,
Nicola Lagomarsini.

PR #1015, opened 2015-04-17
style rename collision produces infinite loop

jodygarnett:
Reported from Matthieu via: #970

The provided fix is fine, and included a test case.

I have cleaned up the formatting and made it format aware.

### DIFF
--- a/src/main/src/main/java/org/geoserver/config/GeoServerDataDirectory.java
+++ b/src/main/src/main/java/org/geoserver/config/GeoServerDataDirectory.java
@@ -6,6 +6,7 @@
 package org.geoserver.config;
 
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -1202,7 +1203,7 @@ public class GeoServerDataDirectory implements ResourceStore {
     protected @Nonnull Style parsedStyleResources(StyleInfo s) throws IOException {
         final Resource styleResource = style(s);
         if ( styleResource.getType() == Type.UNDEFINED ){
-            throw new IOException( "No such resource: " + s.getFilename());
+            throw new FileNotFoundException( "No such resource: " + s.getFilename());
         }
         final DefaultResourceLocator locator = new DefaultResourceLocator();
         locator.setSourceUrl(resourceToUrl(styleResource));
@@ -1355,6 +1356,9 @@ public class GeoServerDataDirectory implements ResourceStore {
                 }
                 
             });
+        }
+        catch(FileNotFoundException e){
+            GeoServerPersister.LOGGER.log(Level.WARNING, "Error loading style:"+ e);
         }
         catch(IOException e) {
             GeoServerPersister.LOGGER.log(Level.WARNING, "Error loading style", e);
--- a/src/main/src/main/java/org/geoserver/config/GeoServerPersister.java
+++ b/src/main/src/main/java/org/geoserver/config/GeoServerPersister.java
@@ -582,8 +582,8 @@ public class GeoServerPersister implements CatalogListener, ConfigurationListene
         Resource style = dd.style(s);
         String sldFileName = newName + ".sld";
         Resource target = style.parent().get(sldFileName);
-        int i = 1;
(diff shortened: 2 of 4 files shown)


src/main/src/main/java/org/geoserver/config/GeoServerPersister.java
```diff
@@ -576,23 +583,53 @@ private void addStyle( StyleInfo s ) throws IOException {
     
     private void renameStyle( StyleInfo s, String newName ) throws IOException {
         LOGGER.fine( "Renameing style " + s.getName() + " to " + newName );
```
tbarsballe: `Renaming style` instead of `Renameing style`

src/main/src/test/java/org/geoserver/config/GeoServerPersisterTest.java
```diff
@@ -665,7 +689,7 @@ public void testModifyStyleWithResourcesInParentDirChangeWorkspace() throws Exce
         StyleInfo s = catalog.getStyleByName( "foostyle" );
         
         s.setWorkspace(catalog.getDefaultWorkspace());
-        catalog.save( s );
+        catalog.save( s ); // note:
```
tbarsballe: `// note:` with no note...

src/platform/src/main/java/org/geoserver/platform/resource/FileSystemResourceStore.java
```diff
@@ -128,7 +128,9 @@ public boolean move(String path, String target) {
         if (!file.exists() && !dest.exists()) {
             return true; // moving an undefined resource
         }
-
+        if( !file.exists() ){
+            throw new IllegalStateException("Unable to find " + path + " to move to " + target);
+        }
```
tbarsballe: This seems wrong/redundant. The check `!file.exists() && !dest.exists()` will often return true before reaching here. We will only get the IllegalStateException if `!file.exists() && dest.exists()`, and the cach block after Files.move already throws this exception anyway.

src/main/src/main/java/org/geoserver/config/GeoServerPersister.java
```diff
@@ -576,23 +583,53 @@ private void addStyle( StyleInfo s ) throws IOException {
...
-            sldFileName = newName + i + ".sld";
-            target = style.parent().get(sldFileName);
+        StyleHandler format = Styles.handler( s.getFormat() );
+        
+        Resource target = uniqueResource( style, newName, format.getFileExtension() );
+        renameRes(style, target.name());
+        s.setFilename(target.name());
+        
+        // rename generated sld if appropriate
+        if( !"sld".equals(format.getFormat())){
```
tbarsballe: Using `format.getFileExtension()` above, but `format.getFormat()` here?
Should use `SLDHandler.FORMAT.equals(format.getFormat())) ` instead.
Alternatively, `"sld".equalsIgnoreCase(format.getFileExtension())`

tbarsballe:
Bunch of relatively minor comments (although they should still be fixed).

Tested, and it seems to work fine with non-SLD styles (eg. YSLD) as well as SLD styles.

tbarsballe:
Updated with changes. FileSystemResourceStore.move() should probably still be looked at (Why does it return true when `!file.exists() && !dest.exists()`; would expect it to throw an exception)

src/main/src/main/java/org/geoserver/config/GeoServerDataDirectory.java
```diff
@@ -1202,7 +1203,7 @@ else if ( l.getResource() instanceof WMSLayerInfo ) {
     protected @Nonnull Style parsedStyleResources(StyleInfo s) throws IOException {
         final Resource styleResource = style(s);
         if ( styleResource.getType() == Type.UNDEFINED ){
-            throw new IOException( "No such resource: " + s.getFilename());
+            throw new FileNotFoundException( "No such resource: " + s.getFilename());
```
jodygarnett: Throwing a more specific exception here so we do not have to produce the full stack trace.

jodygarnett:
I am not sure about the !file.exists() && !dest.exists() - only thing I can think of is that both resources are in the same state. User has renamed a file that they have not created yet etc...

jodygarnett:
Merged to master (squashed to a single commit) and will backport to stable.

PR #1017, opened 2015-04-21
GEOS-6986: Extended namespace validation for WFS-T replace

awaterme:
Same handling for WFS-T Insert and Replace.


awaterme:
BTW: The entire class has no test cases, AFAIK, so I have not created one. Verified manually.

aaime:
There used to be a time when GeoServer had no tests. Now we don't accept changes without one :-)

awaterme:
Makes sense, I have created test cases.

awaterme:
see https://osgeo-org.atlassian.net/browse/GEOS-6986

PR #1021, opened 2015-04-21
Update to Oracle 12 driver (ojdbc7)

tbarsballe:
This change is primarily intended for use with app-schema online tests, but I upgraded everything that used the old ojdbc.jar for consistency.


tbarsballe:
Update: Moved specification of the Oracle version to the \<dependencyManagement/\> section of the root pom

PR #1025, opened 2015-04-23
Improve Doc for WebAdmin Store and Layer pages

dr-jts:
Added subheadings, made formatting more consistent, minor text changes, and added a hint or two about layer naming.

### DIFF
--- a/doc/en/user/source/webadmin/data/layers.rst
+++ b/doc/en/user/source/webadmin/data/layers.rst
@@ -9,7 +9,7 @@ In the Layers section of the web interface, you can view and edit existing layer
 
 .. figure:: ../images/data_layers.png
 
-   Layers View
+   *Layers View*
    
 Layer types
 -----------
@@ -35,13 +35,16 @@ At the upper left-hand corner of the layers view page there are two buttons for 
 
 .. figure:: ../images/data_layers_add_remove.png
    
-   Buttons to Add or Remove a Layer
+   *Buttons to Add or Remove a Layer*
+
+Add a Layer
+^^^^^^^^^^^
 
 Clicking the :guilabel:`Add a new resource` button brings up a :guilabel:`New Layer Chooser` panel. The menu displays all currently enabled stores. From this menu, select the Store where the layer should be added. 
 
 .. figure:: ../images/data_layers_add_chooser.png
 
-   List of all currently enabled stores
+   *List of all currently enabled stores*
 
 Upon selection of a Store, a list is displayed of resources within the store.
 Resources which have already been published as layers are listed first, followed by other resources which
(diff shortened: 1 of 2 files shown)


aaime:
@bmmpxf any time to review this one? :-)

doc/en/user/source/webadmin/data/layers.rst
```diff
@@ -80,61 +86,55 @@ Clicking the layer name opens a layer configuration panel. The :guilabel:`Data`
...
-   Edit Layer: Data tab
+   *Edit Layer: Data tab*
 
 Basic Info
 ^^^^^^^^^^
 
-The beginning sections—Basic Resource Info, Keywords and Metadata link—are analogous to the :ref:`service_metadata` section for WCS, WFS, and WMS. These sections provide "data about the data," specifically textual information that make the layer data easier to work with it. 
+The beginning sections—Basic Resource Info, Keywords and Metadata link—are analogous to the :ref:`service_metadata` section for WCS, WFS, and WMS. 
+These sections provide "data about the data," specifically textual information that make the layer data easier to understand and work with.
+The metadate information will appear in the capabilities documents which refer to the layer.
```
aaime: Typo, metadata, not metadate.

doc/en/user/source/webadmin/data/layers.rst
```diff
@@ -80,61 +86,55 @@ Clicking the layer name opens a layer configuration panel. The :guilabel:`Data`
...
+* **Declared SRS**—Specifies the coordinate system GeoServer publishes to clients 
+* **SRS Handling**—Determines how GeoServer should handle projection when the two SRSes differ 
 
 Bounding Boxes
 ^^^^^^^^^^^^^^
 
-The bounding box determines the extent of a layer. The :guilabel:`Native Bounding Box` is the bounds of the data projected in the Native SRS. You can generate these bounds by clicking the :guilabel:`Compute from data` button. The :guilabel:`Lat/Long Bounding Box` computes the bounds based on the standard lat/lon. These bounds can be generated by clicking the :guilabel:`Compute from native bounds` button.
+The bounding box determines the extent of the data within a layer. 
+
+* **Native Bounding Box**—The bounds of the data specified in the Native SRS. These bounds can be generated by clicking the :guilabel:`Compute from data` button.
```
aaime: Hmm... eh... native is a bit confusing here, in a sense, it's what we claim it's native to the OGC clients, not the actual native CRSs. The difference between the two is evident when one has a native CRS that cannot be matched to a EPSG code, you setup a different one as declared and this will be the the actual CRS in which this bbox should be computed... we should probably rename it to "declared bounding box". Anyways, not something you should change here, more or a clarification of something I've realized clearly just recently. See also https://osgeo-org.atlassian.net/browse/GEOS-7004

doc/en/user/source/webadmin/data/layers.rst
```diff
@@ -194,15 +220,23 @@ The Publishing tab is for configuring HTTP and WMS/WCS settings.
...
 
      The list will be used only for the capabilities document generation, but will not be used to limit the actual target SRS usage in GetFeature requests.
 
-* **WCS Settings**—Provides a list of SRSs the layer can be converted to. :guilabel:`New Request SRS` allows you to add an SRS to that list. 
+WCS Settings
+^^^^^^^^^^^^
+* **Request SRS**—Provides a list of SRSs the layer can be converted to. :guilabel:`New Request SRS` allows you to add an SRS to that list. 
 * **Interpolation Methods**—Sets the raster rendering process, if applicable.
 * **Formats**—Lists which output formats a layers supports.
 * **GeoSearch**—When enabled, allows the Google Geosearch crawler to index from this particular layer. See `What is a Geo Sitemap? <http://www.google.com/support/webmasters/bin/answer.py?hl=en&answer=94554>`_ for more information.
```
aaime: GeoSearch is not part of the WCS settings. Is it showing up at all in a normal UI? The GeoSearch module is an extension

aaime:
Added a few comments, nothing major.

doc/en/user/source/webadmin/data/layers.rst
dr-jts: This is left unchanged from the current doc.  I didn't check too hard for correctness - this patch was mostly focussed on formatting and clarification.  Suggest leaving as is for now, it can be removed later if needed.

doc/en/user/source/webadmin/data/layers.rst
dr-jts: Not sure I understand this well enough to explain it, but a better explanation could be provided in a subsequent patch.

doc/en/user/source/webadmin/data/layers.rst
dr-jts: Ok, add a commit to fix this.

aaime:
Merged, thanks for the work, docs are not exactly where most people put their efforts, nice to see someone that does ;-)

dr-jts:
Thanks...   it's actually easier than coding  :).  GeoServer has pretty good docs really, so it's rewarding to be able to polish them up here and there without a ton of effort.

PR #1026, opened 2015-04-24
GEOS-6994, Cache provider conflict resolution

smithkm:
Resolve conflicts between CacheProviders by picking one and then logging a warning.  The user can specify priority with a system property.

This was tricky to test so i made some changes to unit testing support.

I also added automatic conflict resolution for hz-cluster without the need for user action.

See: https://osgeo-org.atlassian.net/browse/GEOS-6994

### DIFF
--- a/src/community/hz-cluster/src/main/java/org/geoserver/cluster/hazelcast/HzExtensionFilter.java
+++ b/src/community/hz-cluster/src/main/java/org/geoserver/cluster/hazelcast/HzExtensionFilter.java
@@ -1,0 +1,46 @@
+/* (c) 2015 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.cluster.hazelcast;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import org.geoserver.platform.ExtensionFilter;
+import org.geoserver.util.CacheProvider;
+import org.geotools.util.logging.Logging;
+
+/**
+ * Filter to exclude extensions that conflict with hz-cluster.
+ * 
+ * @author Kevin Smith, Boundless
+ *
+ */
+public class HzExtensionFilter implements ExtensionFilter {
+    HzCluster cluster;
+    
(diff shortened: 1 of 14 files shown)


src/community/hz-cluster/src/main/java/org/geoserver/cluster/hazelcast/HzExtensionFilter.java
```diff
@@ -0,0 +1,42 @@
...
+    public HzExtensionFilter(HzCluster cluster) {
+        this.cluster = cluster;
+    }
+    
+    @Override
+    public boolean exclude(String beanId, Object bean) {
+        if(cluster.isEnabled()) {
+            if((bean instanceof CacheProvider) && !(bean instanceof HzCacheProvider)) {
+                // If another extension does this too then we're in trouble as only the default will be used.
+                LOGGER.log(Level.INFO, "hz-cluster module is supressing conflicting CacheProvider {0}", new Object[]{beanId});
```
jodygarnett: This could probably be at CONFIG level as well.

jodygarnett:
Feedback / plan:

- fix headers :)
- the creative code (use of JUnit Rules) creates a fair bit of churn for a bugfix we can backport, but since the churn is all in tests we can ask on geoserver-devel
- if we expect developers to use these new JUnit Rules we will need to add a note to the Developers Guide [Testing](http://docs.geoserver.org/latest/en/developer/programming-guide/testing/index.html) page

smithkm:
The second to last commit "hz-cluster supresses rival CachProviders" by itself is sufficient to fix the issue for hz-cluster.  It doesn't have any changes to unit test support and is entirely confined to the community hz-cluster module.  The unit tests for it do depend on the changes to GeoServerExtensionsHelper but are separated into their own commit.

jodygarnett:
Thanks Kevin, I will merge and add a note to the developers guide.

aaime:
Hi... I was not aware we even had a CacheProvider class. What are these doing in core, without javadocs, and without apparently nothing in core modules using them?

smithkm:
It looks like Gabriel added it as part of GSIP-69

aaime:
Yes, I can see that. That does not answer my question though :)

PR #1028, opened 2015-04-29
[GEOS-6992] Add gdalwarp/gdal_translate/gdaladdo transformation support in importer

aaime:


aaime:
In order to implement this one I had to implement for good the raster import chain, found that there were significant overlaps with the vector one so generalized things a bit, turning the vector Pre/Post/Abstract interfaces/classes into generic base ones that can be shared among raster and vector transformations.
The new tranfsormations are documented, and like others already in there, are available only from the REST api (no GUI support for them at the moment).

tbarsballe:
I've looked over this Andrea - It looks good, although most of it is orthogonal to the parts of importer I have worked on, so I'm not sure that I can provide any more usefull feedback.

src/extension/importer/core/src/main/java/org/geoserver/importer/transform/AbstractCommandLineTransform.java
```diff
@@ -0,0 +1,334 @@
...
+    public void apply(ImportTask task, ImportData data) throws Exception {
+        boolean inline = isInline();
+        File executable = getExecutable();
+        File inputFile = getInputFile(data);
+        Map<String, File> substitutions = new HashMap<>();
+        substitutions.put("input", inputFile);
+        File outputDirectory = null;
+        File outputFile = null;
+        if (!inline) {
+            outputDirectory = inline ? null : getOutputDirectory(data);
```
mbarto: since we are inside if(!inline) the inline ? null : getOutputDirectory(data) can be replaced by getOutputDirectory()

src/extension/importer/rest/src/main/java/org/geoserver/importer/rest/ImportJSONReader.java
```diff
@@ -281,12 +286,37 @@ ImportTransform transform(JSONObject json) throws IOException {
             catch(Exception e) {
                 throw new ValidationException("Error parsing reproject transform", e);
             }
+        } else if ("GdalTranslateTransform".equalsIgnoreCase(type)) {
```
mbarto: Some ability to register / read new transformers in a more dynamic way here is probably advisable in the long term, but is probably out of the context of this work.

mbarto:
In general, it looks good to me.

PR #1029, opened 2015-04-30
[GEOS-7000] Allow importer to add granules to a mosaic as an indirect import

aaime:
This one requires #1028 to be merged first


mbarto:
It looks good to me

aaime:
Refactored according to Mauro's suggestion and merged manually

PR #1032, opened 2015-04-30
[GEOS-7003] Fixed the layer preview search not searching layer titles and keywords and layer group titles and abstracts in 2.7.0.

sikeoka:
[GEOS-7003] Fixed the layer preview search not searching layer titles and keywords and layer group titles and abstracts in 2.7.0.


aaime:
Thank you Steve.

PR #1033, opened 2015-05-03
Build speed up improvements

aaime:
Ask all tests to setUpSecurity to avoid data migrations, reduce the number of data migrations, etc...


src/extension/app-schema/sample-data-access-test/src/test/java/org/geoserver/test/SampleDataAccessWfsTest.java
```diff
@@ -175,8 +174,7 @@ public void prettyPrint(Document doc, OutputStream out) throws Exception {
 
     @Override
     protected void setUpTestData(SystemTestData testData) throws Exception {
-        // TODO Auto-generated method stub
-        
+        testData.setUpSecurity();
```
aaime: Tests that do not need test data better add security config, otherwise the time saved by not configuring data will be lost to the security migrations. I've added this line to all tests I could locate that do not need pre-canned data

src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
```diff
@@ -678,7 +678,7 @@ protected Class impl(Class interfce) {
      * Custom reflection provider which unwraps proxies, and skips empty collections
      * and maps.
      */
-    class CustomReflectionProvider extends Sun14ReflectionProvider {
+    class CustomReflectionProvider extends SunUnsafeReflectionProvider {
```
aaime: We were extending a old and deprecated class, the new one is somewhat faster and it's supported.

src/main/src/main/java/org/geoserver/security/GeoServerSecurityManager.java
```diff
@@ -305,6 +328,8 @@ public ConfigurationPasswordEncryptionHelper getConfigPasswordEncryptionHelper()
     @Override
     public void setApplicationContext(ApplicationContext appContext) throws BeansException {
         this.appContext = appContext;
+        this.xp = buildPersister();
```
aaime: Building and configuring a XStream instance is very expensive, so do it just once (it kept on showing up as the biggest offender even after the version check in the migrations, see below)

src/main/src/main/java/org/geoserver/security/GeoServerSecurityManager.java
```diff
@@ -326,11 +351,25 @@ public void onApplicationEvent(ApplicationEvent event) {
 
             // migrate from old security config
             try {
-                boolean migratedFrom21 = migrateFrom21();
-                removeErroneousAccessDeniedPage();
-                migrateFrom22(migratedFrom21);
-                migrateFrom23();
-                migrateFrom24();                
+                Version securityVersion = getSecurityVersion();
```
aaime: In order to avoid running a series of expensive security dir migrations I've added a version property file, that the code uses to figure out the starting version, from there it will do the migrations required and upgrade the version number. Blindly doing the 2.3 -> 2.4 and 2.4 -> 2.5 migratinos was the number one performance offender

src/main/src/main/java/org/geoserver/security/GeoServerSecurityManager.java
```diff
@@ -2687,7 +2758,14 @@ public boolean accept(File pathname) {
     }
 
     XStreamPersister globalPersister() throws IOException {
-        XStreamPersister xp = persister();
+        if (gxp == null) {
```
aaime: Should not be necessary, but some tests are not doing proper initialization, so we make sure to check.

src/main/src/test/java/org/geoserver/catalog/CatalogIntegrationTest.java
```diff
@@ -42,6 +42,11 @@
 public class CatalogIntegrationTest extends GeoServerSystemTestSupport {
     
     @Override
+    protected boolean isFullInitRequired() {
```
aaime: This test requires all beans to be initialized, as some programmatically attach catalog listeners, and this test is checking them. 99% of the test do not care/need this.

aaime: See also my notes in the GeoServer system test classes

src/main/src/test/java/org/geoserver/security/AbstractSecurityServiceTest.java
```diff
@@ -304,11 +298,11 @@ protected void checkValuesInserted(GeoServerUserGroupService userGroupService) t
         assertEquals(5, userGroupService.getUsers().size());
         assertEquals(5, userGroupService.getUserCount());
         
-        GeoServerUser admin = (GeoServerUser) userGroupService.getUserByUsername(GeoServerUser.ADMIN_USERNAME);
-        GeoServerUser user1 = (GeoServerUser) userGroupService.getUserByUsername("user1");
-        GeoServerUser user2 = (GeoServerUser) userGroupService.getUserByUsername("user2");
-        GeoServerUser disableduser = (GeoServerUser) userGroupService.getUserByUsername("disableduser");
-        GeoServerUser groupAdminUser = (GeoServerUser) userGroupService.getUserByUsername("groupAdminUser");
+        GeoServerUser admin = userGroupService.getUserByUsername(GeoServerUser.ADMIN_USERNAME);
```
aaime: Here and there we have the removal of unecessary casts.

src/main/src/test/java/org/geoserver/security/xml/XMLUserDetailsServiceTest.java
```diff
@@ -44,6 +37,11 @@ protected SystemTestData createTestData() throws Exception {
     }
 
     @Override
+    protected void setUpTestData(SystemTestData testData) throws Exception {
+        // do nothing here, we want the live test data info and run its migration
```
aaime: This is the one case in which for some reason we need migrations to run, or the test will fail.

src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java
```diff
@@ -164,9 +165,33 @@ protected SystemTestData createTestData() throws Exception {
     /**
      * credentials for mock requests
      */
-    protected String username, password; 
+    protected String username, password;
+
+    /**
+     * Cached dispatcher, it has its own app context, so it's expensive to build
+     */
+    protected static DispatcherServlet dispatcher;
```
aaime: Building a dispatcher servlet for every mock request we do is rather expensive, do it once and cache the result (a dispatcher servlet has its own local app context to initialize, even if for us it's empty, creating it is costly)

src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java
```diff
@@ -164,9 +165,33 @@ protected SystemTestData createTestData() throws Exception {
...
+    protected boolean isFullInitRequired() {
+        return false;
+    }
 
     protected final void setUp(SystemTestData testData) throws Exception {
+        // speed up xpath evaluations
+        try {
+            // see
+            // http://stackoverflow.com/questions/6340802/java-xpath-apache-jaxp-implementation-performance
+            Class.forName("com.sun.org.apache.xml.internal.dtm.ref.DTMManagerDefault");
```
aaime: This speeds up xpath evaluations, we are doing a number of them. The Class.forName should be guarding against JDK that do not have the required classes.

src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java
```diff
@@ -164,9 +165,33 @@ protected SystemTestData createTestData() throws Exception {
...
+            System.setProperty("com.sun.org.apache.xml.internal.dtm.DTMManager",
+                    "com.sun.org.apache.xml.internal.dtm.ref.DTMManagerDefault");
+        } catch (Exception e) {
+            // ignore on VM where this optimization does not apply
+        }
+
+        // disable security manager to speed up tests, we are spending a lot of time in privileged
+        // blocks
+        System.setSecurityManager(null);
```
aaime: We are spending a lot of time in doPrivileged, this setup reduces the time we spent as we are not checking over and over the security configs.

aaime: Worth noticing that Oracle is aware of the issue and there is a JEP for JDK 9 to speed up security checks in doPrivileged.

src/main/src/test/java/org/geoserver/test/GeoServerTestApplicationContext.java
```diff
@@ -33,15 +37,20 @@
...
+        this(new String[] { configLocation }, servletContext, true);
     }
 
-    public GeoServerTestApplicationContext(String[] configLocation, ServletContext servletContext)
+    public GeoServerTestApplicationContext(String[] configLocation, ServletContext servletContext,
+            boolean fullInitRequired)
         throws BeansException {
         super(configLocation, false);
+        this.fullInitRequired = true;
+        setValidating(false);
```
aaime: We are spending visible amount of time doing xml validation, disable it.

src/main/src/test/java/org/geoserver/test/GeoServerTestApplicationContext.java
```diff
@@ -96,4 +105,25 @@ protected void initPropertySources() {
         WebApplicationContextUtils.initServletPropertySources(
                 this.getEnvironment().getPropertySources(), this.servletContext);
 	}
+
+    @Override
+    protected void initBeanDefinitionReader(XmlBeanDefinitionReader reader) {
+        super.initBeanDefinitionReader(reader);
+        if (!fullInitRequired) {
```
aaime: If we can, load the beans lazily instead of fully on startup. This gives a significant benefit during the build, and a very significant one when running the tests on the IDE. 
Reason: each test is only depending on a subset of all the beans, we do not need to initialize them all.

src/main/src/main/java/org/geoserver/security/GeoServerSecurityManager.java
jodygarnett: Nice, I like it.

jodygarnett:
Okay, reviewed - you seem to have knocked this one completely out of the part Andrea.

Can I merge this, or do we need to carefully warning geoserver-devel.

aaime:
Two things still left out I believe. One is that one test failed on Travis, but it did not locally... have to check what happened there.
The second is about lazy init in Spring, right now I'm asking specific tests to force full init of the spring context as opposed to lazy init, as an alternative that I wanted to try, ask people to explicitly mark the beans that need immediate loading (because they do some programmatic alteration of the environment on load, such as attaching listeners) as lazy-init="false"

aaime:
Oh, about my comments, they were meant as a guide for the reviewer, as I did many changes :-)

src/wms/src/test/java/org/geoserver/wms/svg/SVGTest.java
```diff
@@ -60,13 +62,7 @@ public void testBasicSvgGeneratorMultipleFts() throws Exception {
...
     @Test
     public void testBatikSvgGenerator() throws Exception {
-        //batik includes DTD reference which forces us to be online, skip test
-        // in offline case
-        try {
-            new URL( "http://www.w3.org").openConnection().connect();
-        } catch (Exception e) {
-            return;
-        }
+        Assume.assumeTrue(isw3OrgReachable());
```
jodygarnett: Interesting have not used Assume before

src/wms/src/test/java/org/geoserver/wms/svg/SVGTest.java
```diff
@@ -82,15 +78,26 @@ public void testBatikSvgGenerator() throws Exception {
...
-            new URL( "http://www.w3.org").openConnection().connect();
+            HttpURLConnection connection = (HttpURLConnection) new URL("http://www.w3.org")
+                    .openConnection();
+            connection.setConnectTimeout(5000);
+            connection.connect();
+            connection.disconnect();
+            return true;
         } catch (Exception e) {
-            return;
+            e.printStackTrace();
```
jodygarnett: Given that this feeds into assume, perhaps we could just print the failure on a single line, the stack trace itself if not interesting - only the result that www.w3.org was not available.

jodygarnett:
Squashing commits and merging to master.

PR #1038, opened 2015-05-05
Improvements for Layer Group page

dr-jts:
Some improvements to wording and information on Web Admin Layer Group page.


jodygarnett:
Result looks fine, squish commits when merging so we can cherrypick back to stable series.

dr-jts:
Yes, squashing would be good, but in this case I was editing them directly in GitHub, so not sure I can squash them myself?

jodygarnett:
squished and merged see ed9abaa9ba8640455ece12e0945eb76aeb477164

PR #1039, opened 2015-05-05
[GEOS-7008] fix cascading delete for duplicate layers in layer group

tbarsballe:


tbarsballe:
Successfully backported to 2.7.x
I was unable to backport this to 2.6.x - the fix seemed fine, but the test framework was different enough to cause the test to fail. It looks like an older mock framework that is more strict about what operations are permitted. A quick attempt at rewriting the test case also failed.

PR #1040, opened 2015-05-05
Minor syntax fixes to CSS tutorial

tbarsballe:
Pull request fixes syntax regressions in tutorial (or codebase) hard to know which.

### DIFF
--- a/doc/en/user/source/extensions/css/workshop/linestring.rst
+++ b/doc/en/user/source/extensions/css/workshop/linestring.rst
@@ -371,7 +371,7 @@ Vendor options can be used to enable some quite spectacular effects, while still
    .. code-block:: css
 
        * {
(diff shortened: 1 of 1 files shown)


jodygarnett:
@aaime do you think the fixes above are a regression in the css java code? Appears to be a difference parsing colors and floats:

* stroke: ededff; -->  stroke: #ededff; 
* stroke-width: .75; --> stroke-width: 0.75;

aaime:
@jodygarnett we can see it both ways... those two seem undocumented (back at the time, the workshop was part of the GeoServer docs) behavior of the old css parser. We can fix them (the parser is not too hard to change, adding a test is equally easy, want to try?) or we can fix the docs.

jodygarnett:
would rather be consistent, and change the docs.

PR #1043, opened 2015-05-07
Adjust for change of Matrix library (GEOS-7009)

jodygarnett:
Adjust expected result test result for GetFeatureInfo test due to difference in transform accuracy.

This request requires that https://github.com/geotools/geotools/pull/829 be applied first

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -23,10 +23,10 @@
  </scm>
 
  <organization>
-  <name>The Open Planning Project</name>
-  <url>http://www.openplans.org</url>
+  <name>Open Source Geospatial Foundation</name>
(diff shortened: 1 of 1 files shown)


aaime:
Please allow for some more time to review. I was too busy trying to get the windows server going and the build sped up to look at this one.

aaime:
See review on the mailing list, the shift is too big, there is something going on that we need to understand better. Please do not merge this one.

jodygarnett:
Understood, I have only observed differences in invert result. I could use a hand here as I am a bit out of my depth on the math front.

jodygarnett:
merged just the change to the pom.xml organization.

PR #1044, opened 2015-05-07
Geofence: REST API

NielsCharlier:
The Geofence REST API.

Some questions for the reviewers though:
- The user/group and role rest controllers don't really belong in the geofence-server module. Should I perhaps make a new community module or do they belong in an existing module?
- The user/group and role rest controllers seem too trivial to me to write unit tests for, each method only calls one other method. Suggestions for what I could write for testing is welcome though

### DIFF
--- a/src/community/geofence-server/src/main/java/org/geoserver/geofence/rest/RolesRestController.java
+++ b/src/community/geofence-server/src/main/java/org/geoserver/geofence/rest/RolesRestController.java
@@ -1,0 +1,186 @@
+/* (c) 2015 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.geofence.rest;
+
+import java.io.IOException;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import org.geoserver.geofence.rest.xml.JaxbRoleList;
+import org.geoserver.security.GeoServerRoleService;
+import org.geoserver.security.GeoServerRoleStore;
+import org.geoserver.security.GeoServerSecurityManager;
+import org.geoserver.security.impl.GeoServerRole;
+import org.springframework.http.HttpStatus;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@Controller
+public class RolesRestController {
+
+    protected GeoServerSecurityManager securityManager;
+
+    public RolesRestController(GeoServerSecurityManager securityManager) {
+        this.securityManager = securityManager;
+    }
+    
+    @ExceptionHandler(IllegalArgumentException.class)
+    public void somethingNotFound(IllegalArgumentException exception, HttpServletRequest request, HttpServletResponse response) throws IOException {
+    	response.sendError(404, exception.getMessage());
+    }
+
+    @RequestMapping(value = "/rest/roles", method = RequestMethod.GET, produces = {"application/xml", "application/json"})
+    public @ResponseBody JaxbRoleList get() throws IOException {
+        return get(securityManager.getActiveRoleService());
+    }
+
+    @RequestMapping(value = "/rest/roles/user/{user}", method = RequestMethod.GET, produces = {"application/xml", "application/json"})
+    protected @ResponseBody JaxbRoleList getUser(@PathVariable("user") String userName)
+            throws IOException {
+        return getUser(securityManager.getActiveRoleService(), userName);
+    }
+
+    @RequestMapping(value = "/rest/roles/role/{role}", method = RequestMethod.POST, produces = {"application/xml", "application/json"})
+    public @ResponseStatus(HttpStatus.CREATED) void insert(@PathVariable("role") String roleName)
+            throws IOException {
+        insert(securityManager.getActiveRoleService(), roleName);
+    }
+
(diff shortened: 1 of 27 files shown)


src/community/geofence-server/src/main/java/org/geoserver/geofence/rest/RolesRestController.java
```diff
@@ -0,0 +1,149 @@
+package org.geoserver.geofence.rest;
```
jodygarnett: headers

src/community/geofence-server/src/main/java/org/geoserver/geofence/rest/RulesRestController.java
```diff
@@ -0,0 +1,189 @@
...
+        @RequestParam(value = "instanceName", required = false) String  instanceName,
+        @RequestParam(value = "instanceAny", required = false) Boolean instanceDefault,
+        @RequestParam(value = "service", required = false) String  serviceName,
+        @RequestParam(value = "serviceAny", required = false) Boolean serviceDefault,
+        @RequestParam(value = "request", required = false) String  requestName,
+        @RequestParam(value = "requestAny", required = false) Boolean requestDefault,
+        @RequestParam(value = "workspace", required = false) String  workspace,
+        @RequestParam(value = "workspaceAny", required = false) Boolean workspaceDefault,
+        @RequestParam(value = "layer", required = false) String layer,
+        @RequestParam(value = "layerAny", required = false) Boolean layerDefault
```
jodygarnett: Looks nice and clean.

src/community/geofence-server/src/main/java/org/geoserver/geofence/rest/UsersRestController.java
```diff
@@ -0,0 +1,151 @@
+package org.geoserver.geofence.rest;
```
jodygarnett: headers :)

Not sure where/if you want to move this, I kind of like having all the REST security work in one place.

jodygarnett:
docs required for user manual, covering functionality of rest api.

src/community/geofence-server/src/main/java/org/geoserver/geofence/rest/UsersRestController.java
NielsCharlier: But what about people who want security REST but not geofence?

NielsCharlier: Should there perhaps be a gs-rest-sec, and then any rest for the default security framework can be put there? And keep the geofence rest in this one

mcrmcr: Hi all

I am about to finish GISP 120 and I put the security rest code in
gs-restconfig, hope this is ok

On Thu, May 14, 2015 at 4:33 PM, NielsCharlier <notifications@github.com>
wrote:

> In
> src/community/geofence-server/src/main/java/org/geoserver/geofence/rest/UsersRestController.java
> <https://github.com/geoserver/geoserver/pull/1044#discussion_r30325318>:
>
> > @@ -0,0 +1,151 @@
> > +package org.geoserver.geofence.rest;
>
> Should there perhaps be a gs-rest-sec, and then any rest for the default
> security framework can be put there? And keep the geofence rest in this one
>
> —
> <https://github.com/geoserver/geoserver/pull/1044/files#r30325318>.
>



-- 
DI Christian Mueller MSc (GIS), MSc (IT-Security)
OSS Open Source Solutions GmbH

NielsCharlier: @mcrmcr , are you using restful? Cause I made everything with spring as you can see.

NielsCharlier: perhaps we should make a new module for spring rest, like restconfig-ng as we move things gradually. @jodygarnett  please advise

mcrmcr: Hi Niels

I  did it like the other REST stuff in geoserver, I have no idea about
using Spring for the REST API.

On Sat, May 16, 2015 at 9:51 AM, NielsCharlier <notifications@github.com>
wrote:

> In
> src/community/geofence-server/src/main/java/org/geoserver/geofence/rest/UsersRestController.java
> <https://github.com/geoserver/geoserver/pull/1044#discussion_r30459367>:
>
> > @@ -0,0 +1,151 @@
> > +package org.geoserver.geofence.rest;
>
> perhaps we should make a new module for spring rest, like restconfig-ng as
> we move things gradually. @jodygarnett <https://github.com/jodygarnett>
> please advise
>
> —
> <https://github.com/geoserver/geoserver/pull/1044/files#r30459367>.
>



-- 
DI Christian Mueller MSc (GIS), MSc (IT-Security)
OSS Open Source Solutions GmbH

NielsCharlier:
added docs + extension configuration

jodygarnett:
Q: The user/group and role rest controllers seem too trivial to me to write unit tests for, each method only calls one other method. Suggestions for what I could write for testing is welcome though

A:

Write a test case as a small workflow that uses the API to configure a couple of users with an appropriate group/role and then delete one of the users. This test case would be handy as bug reports come in as it could be adapted as needed.

jodygarnett:
Thanks for adding the docs Niels I will review, build, merge.

NielsCharlier:
Jody, an additional test as you suggested has been added. Apart from the spring build issue due to changes in geoserver test bases (see mailing list) it is good to go I think

NielsCharlier:
Changes: added JSON support, added security, user of a default servicename for user/group service, better exception reporting with invalid service/username/groupname/rolename, fixed priority bug, added test for it updated docs

etj:
Basic REST calls are ok. 
Anyway some corner cases (creation of duplicated entries, entries not found) should be handled, because usually they return a generic 500 (and an unwanted stacktrace) instead of proper error codes. 
These cases may be handled later, but I guess a skeleton test structure should be implemented (probably imported from restconfig) for testing HTTP return codes.

NielsCharlier:
It is now returning error codes. The errors are all checked in the unit tests but only the exceptions thrown because there are unit tests in place for the methods of the service on the java level but not as a webservice (the marshalling is also not tested, spring is expected to do its work...)

PR #1047, opened 2015-05-07
fixed [GEOS-7012] and a related issue master. 

smesdaghi:
using the rest api to import geojson was failing because the LayerInfo object return by the importer was not properly populated. The importer was working through geoserver but attempting to import from geonode exposed the issues. for example, the following expand=3 would cause geoserver to throw an exception as listed in [GEOS-7012]. 
http://localhost:8080/geoserver/rest/imports/1/tasks/0?expand=3

Second commit computes the bounds properly as well. These changes are safe for 2.6/2.7/master as well since GeoJSONFormat has not changed. Will generate a pull request for each branch.

### DIFF
--- a/src/extension/importer/core/src/main/java/org/geoserver/importer/format/GeoJSONFormat.java
+++ b/src/extension/importer/core/src/main/java/org/geoserver/importer/format/GeoJSONFormat.java
@@ -51,6 +51,7 @@ import com.google.common.collect.Iterables;
 public class GeoJSONFormat extends VectorFormat {
 
     static Logger LOG = Logging.getLogger(GeoJSONFormat.class);
+    private static ReferencedEnvelope EMPTY_BOUNDS = new ReferencedEnvelope();
 
     @Override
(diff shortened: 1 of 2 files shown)


aaime:
As a documented rule (https://github.com/geoserver/geoserver/blob/master/CONTRIBUTING.md) we ask that every change contains a test. Citing:

"Include a test case that shows your patch fixes an issue (or adds new functionality). If you do not include a test case the developer reviewing your work will need to create one."

(aka, you might have to wait months for someone to create it, and if it does not happen, the pull request gets closed without being merged).

smesdaghi:
Andrea, I added a test and also applied it to the 2.7 and 2.6 pull requests (we are currently on 2.6 snapshot). Thanks.

aaime:
Manually merged to squash the commits (we normally ask that at one ticket corresponds one commit) and cherry-picked on 2.7.x and 2.6.x. Closing all related pull requests. Thanks for the patch and test.

PR #1055, opened 2015-05-12
Added timeout format option to WMS GetMap

tbarsballe:
Changes:
* Added a "timeout" format option to the RenderedImageMapOutputFormat. Allows users to specify a timeout in ms. The renderer will use the nonzero minimum of this timout and the WMS global timeout.
* Moved the actual rendering out to a seperate thread. This means that if we reach the timeout, we can return immedialtly, and let GeoServer clean up the renderer on its own time, rather than waiting for the current request to finish.

There is one bit of code that I am a bit leery of adding. The clean up process for the renderer will forcibly terminate the thread (Thread.stop()) if it does not end itself after 1000 ms of being asked to. This helps keep geoserver running quickly after timing out a large request, but leads to its own complications. I feel like there are two options here:
* Keep as-is. Means we are using Thread.stop(), which may result in memory leaks, headless exceptions, etc.
* Allow the renderer to stop itself, even if it takes a while. If geoserver gets a large number of massive requests, could lead to all resources being used waiting for the renderers to clean up after themselves.


jodygarnett:
The previous implementation also used Thread.stop() did it not?

tbarsballe:
With Thread.stop(), I am more concerned that it is deprecated functinality and can cause resource leaks. On the flip side, it will lower the stress on the server if we are timeing out lots of requests.

tbarsballe:
Update: Added PARTIALMAP exception type (Primarily intended for returning a partially rendered image on a timeout, but also works with other WMS Configuration limits), and documentation for both new features.

src/wms/src/main/java/org/geoserver/wms/WMSServiceException.java
```diff
@@ -0,0 +1,133 @@
...
+package org.geoserver.wms;
+
+import org.geoserver.platform.ServiceException;
+
+/**
+ * Extension of ServiceException for returning a WebMap with the exception.
+ * Used with WMS option EXCEPTIONS=PARTIALMAP to return a partial image if there is an 
+ * exception (such as a timeout) thrown when rendering a WMS request.
+ */
+public class WMSServiceException extends ServiceException {
```
aaime: When Gabriel made the giant WMS refactor (2010?) he removed this class, which was present, under the notion that a single ServiceException was enough for everything. I did not agree, but at least the module is uniform now, if you want to re-introduce this class, please let's discuss it on the ml, and let's keep the module and its dependencies (kml) uniform (wmsexception everywhere or serviceexception everywhere)

aaime: Alternatively, let's give this one a more specific name, dunno, WMSTimeoutException.

src/wms/src/main/java/org/geoserver/wms/map/RenderedImageMapOutputFormat.java
```diff
@@ -505,57 +509,86 @@ public RenderedImageMap produceMap(final WMSMapContent mapContent, final boolean
...
         }
-
-        // check if a non ignorable error occurred
-        if (nonIgnorableExceptionListener.exceptionOccurred()) {
-            Exception renderError = nonIgnorableExceptionListener.getException();
-            throw new ServiceException("Rendering process failed", renderError, "internalError");
+        
+        //Run rendering in a separate thread so that we can exit promptly if the timeout is reached,
+        //rather than waiting for the renderer to finish the current task.
+        MapPainter painter = new MapPainter(renderer, paintArea, mapContent, graphic, layout);
```
aaime: The renderer already has a timeout machinery, and you cannot stop the renderer anyways (besides via its stop request, and/or by throwing exceptions in the listeners, please don't introduce another level of thread handling, it's unecessary complication and violence). The thing that you will have troubles stopping is the raster processing anyways, which needs some extra machinery

aaime:
Btw, did you consider the direct rendering path? That one will generate a JAI chain, and will exit pretty much immediately, all the time will actually be spent encoding, as it's there that the jai chain will be actually pulled and start computing tiles.

src/wms/src/main/java/org/geoserver/wms/WMSServiceException.java
tbarsballe: Given that this is only intended for exceptions that also return a WebMap, I think refactoring the name would be better; maybe something like WMSPartialMapException. (Since we could get this thrown if we hit other WMS configuration limits such as max request memory or max errors, not just timout).

src/wms/src/main/java/org/geoserver/wms/map/RenderedImageMapOutputFormat.java
tbarsballe: Good to get some input from someone who knows the internals of the renderer better than I do. I'll move back to the previous method of termiating the rendering process.

From what I have observed, it looks like renderer will not stop rendering in the middle of drawing one feature. One example is the natural earth oceans layer (which is one big feature) - if we start rendering this, the renderer will try to render the whole thing (or at least the first symbolizer) regardless of timeout. Is there anything that can be done about this? A more reasonable case is some more ordinary layer that just happens to have some very large features in it - such a layer will not respect any timeout very well.

aaime: The timeout is there to make sure that we stop the rendering and thus free resources (cpu and memory) locked by the renderer, which are both quite significant in terms of consumption (a reason why we normally tell people to keep a low value for concurrent wms requests in the flow controller).

About painting a large feature, the java2d code needs to go over a PathIterator to do the painting, so we could make it so that the path iterators returned by our custom Shape objects (LiteShape) start throwing exceptions when the stop rendering flag is raised. That happens in the the painting thread, which is a secondary one, so we could also just check if the thread has been interrupted, just not sure how much overhead that will cause (might be very significant, we are talking about millions of extra checks, one per coordinate pair... maybe we could make it so that the check is done only once every 1000 coordinates, or something like that).

But that is just one case. You could be waiting on the database to finish executing your request (see http://blog.grovehillsoftware.com/2009/08/interruptible-jdbc-statements.html for an approach), or you could be waiting for jai to render an image (tile cache poisoning could be a way there, we could build a wrapper around the tile cache that checks if the current thread has been interrupted), or we could be encoding the output, e.g.,  a large png output takes several seconds to perform the encoding, even if the source image is already fully computed (not the jai streaming case you get with direct raster renderer), which might be addressed via output stream poisoning instead. So... as you can see there is a list of things that would need to be done to get the system more or less behave like you want... not a small amount of work, but a "right" solution. The approach you're proposing is instead going to give the illusion of timeout, but cause more problems down the line.

tbarsballe:
Made some updates based on your feedback @aaime 
* Removed the seperate-thread rendering. Back to using the standard timeout task, though I extended its functionality a bit to allow saving the map before cleanup.
* Renamed WMSServiceException to WMSPartialMapException. We are only using this in the case the user requests EXCEPTIONS=PARTIALMAP, so it makes the most sense to me to just have this special case, and leave everything else as ServiceException.

I have not looked at the direct rendering path, no. Is there currently any timeout support (via WMS Configuration limits) for this? When is it actually used?

aaime:
Hi Torben, the direct rendering path is used any time the request contains a single raster layer, with no rendering transformations over it. This path is pure jai and significantly faster and more scalable than going though java2d

src/wms/src/main/java/org/geoserver/wms/map/RenderedImageMapOutputFormat.java
```diff
@@ -505,10 +508,39 @@ public RenderedImageMap produceMap(final WMSMapContent mapContent, final boolean
         
         onBeforeRender(renderer);
 
-        // setup the timeout enforcer (the enforcer is neutral when the timeout is 0)
+        //Timeout on the smallest nonzero value of the WMS timeout and the timeout format option
+        //If both are zero then there is no timeout
         int maxRenderingTime = wms.getMaxRenderingTime() * 1000;
+        int localMaxRenderingTime = 0;
```
aaime: Can we encapsulate all these calculation in their own method? It seems they would just return the max rendering time anyways.

aaime:
Just added another minor comment, beside that it seems ok code wise.
Build wise not so much, it seems your new test breaks on Travis, check it out.

tbarsballe:
Pulled timeout calculations out into their own method
Updated the test case so that it is (pretty much) guaranteed to cause a timout - the earlier test failure was likely caused by the renderer beating the 1 ms timeout used.
Travis Build passes now.

jodygarnett:
I would like to merge this in for milestone release? Is there any remaining feedback to address?

tbarsballe:
I believe everything @aaime commented on in his review has been addressed.
The one thing is supporting timeout for the fast path rendering, but that has never been supported (even with global WMS configuration limits), and should be its own change.
I feel like this in a state to be merged, so I would say go for it.

jodygarnett:
Rebased and applied to master

PR #1056, opened 2015-05-13
Regularize WebAdmin Data documentation

dr-jts:
This patch makes the structure of the WebAdmin Data data type pages all follow the same pattern.  It also makes section wording consistent, and clearer in some cases.  A few section headings have been added, to make structure clearer and allowing easier linking.

### DIFF
--- a/doc/en/user/source/webadmin/data/index.rst
+++ b/doc/en/user/source/webadmin/data/index.rst
@@ -17,8 +17,8 @@ Each subsection describes a data type page which provides add, view, edit, and d
    layergroups
    styles
    
-Using Data type List views
---------------------------
+**Using Data type List views**
+
 
 The main page for each data type is a list view showing the items of that data type configured in the GeoServer instance.
 The page displays links to the data type items, and where applicable their parent data types.
@@ -31,8 +31,7 @@ In the example below, the :guilabel:`Layers` list page displays a table of layer
 
    *Layers list page*
 
-Sorting
-^^^^^^^
+**Sorting**
 
 To sort a data type alphabetically, click the column header. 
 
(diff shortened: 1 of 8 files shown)


doc/en/user/source/webadmin/data/index.rst
```diff
@@ -4,15 +4,36 @@ Data
...
+.. toctree::
+   :maxdepth: 2
+
+   workspaces
+   stores
+   layers
+   layergroups
+   styles
+   
+Using Data type List views
```
bmmpxf: Perhaps take off the headings from here down on this page. A second-level heading here shows up in the parent TOC as the same level as workspaces stores, et al, which is probably not what we want.

aaime:
This one has been stuck for some time. Any progress in sight?

dr-jts:
Ah, right, I forgot there was some work to do here.  I'll action this ASAP.

doc/en/user/source/webadmin/data/index.rst
dr-jts: Seems a shame to eliminate the headings.  Any alternative to totally removing them?

Perhaps the content could move to the Interface Basics page?  Then it can have headings and will show up in an appropriate place in the parent TOC.

dr-jts: @bmmpxf any thoughts on this?

bmmpxf: How about "fake" headings? Just bold the text, and it won't affect the TOC.

dr-jts: Sigh... I suppose so.  I was thinking it would be nice if the sections were linkable.  But that can be done later - don't want to hold up progress on this.

bmmpxf: Me neither. Moving somewhere else is fine too.

dr-jts:
Added a commit to remove headings on index page.  @bmmpxf can you pls review?

dr-jts:
Hold on... @bmmpxf  just saw your note about moving the sections.  That's better - I'll do that now, and  add a commit.

dr-jts:
Ok, this is good to review

bmmpxf:
Looks good to me. Thanks.

PR #1058, opened 2015-05-14
GEOS-7020: fixed VectorRenderingLayerIdentifier bounds / size calcula…

mbarto:
…tion to have a more precise scale factor calculation.

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/featureinfo/VectorRenderingLayerIdentifier.java
+++ b/src/wms/src/main/java/org/geoserver/wms/featureinfo/VectorRenderingLayerIdentifier.java
@@ -171,12 +171,12 @@ public class VectorRenderingLayerIdentifier extends AbstractVectorLayerIdentifie
             if(radius < buffer) {
                 radius = buffer;
             }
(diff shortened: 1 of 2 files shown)


aaime:
+1

bencaradocdavies:
@aaime why +1 when this breaks Travis CI on OpenJDK?
https://osgeo-org.atlassian.net/browse/GEOS-7023

aaime:
@bencaradocdavies why do we only hear complaints from you?
I've reviewed the change and gave a plus one, but I'm not a compiler, is that so hard to understand for you?

mbarto:
fyi: I am working on it, seems to be an issue with antialiasing not turned off and handled differently by the different JVM

bencaradocdavies:
Thanks @mbarto, I saw https://github.com/geoserver/geoserver/pull/1067 and will test it on openjdk7.

No problem @aaime, I do not think everyone has learned to expect working Travis CI as it was broken for so long. I think it is merger's responsibility to check, but this is not yet agreed or clear.

bencaradocdavies:
@mbarto, I think your analysis is correct. OpenJDK uses a different antialiasing implementation and we have previously seen failures caused by this in, if I recall correctly, gt-render.

PR #1059, opened 2015-05-14
GEOS-6874 REST API bug - The specified coverageName is not supported

smithkm:
Using the native name rather than the catalog name so that coverage readers will recognize the name.


aaime:
Hi Kevin, the request seems fine, I don't see a reference to a jira ticket in the commit message though, this makes it harder to track commits.

smithkm:
Commit name changed.

aaime:
Hi Kevin, was the plan to just merge it on master, or do you feel it's safe to backport to 2.7.x and 2.6.x too?

simboss:
Please, do consider a backport.

If needed Andrea, please, help Kevin out.

smithkm:
I expect it to be back portable as it's just switching which getName method is called and getNativeCoverageName has been around for a while.  The client issue that prompted me to fix it involves GS 2.5 so I may have to backport back that far anyway.

PR #1063, opened 2015-05-15
Add documentation for non standard auto CRS namespace in WMS

SimonSchafer:
I have added a couple of notes that refer to the version of GeoTools.  This might be confusing in GeoServer documentation but I felt they were needed.

### DIFF
--- a/doc/en/user/source/services/wms/nonstandardautonamespace.rst
+++ b/doc/en/user/source/services/wms/nonstandardautonamespace.rst
@@ -1,0 +1,50 @@
+.. _non_standard_auto_namespace:
+
+Non Standard AUTO Namespace
+===========================
+
+The WMS standard supports a small number of "automatic" coordinate reference systems that include a user-selected centre of projection.  These are specified using::
+
+    AUTO:auto_crs_id,factor,lon0,lat0
+
+for example::
+
+    CRS=AUTO:42003,1,-100,45
+
+.. note::   in GeoTools 13 AUTO and AUTO2 namespaces are treated identically.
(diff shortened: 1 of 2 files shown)


doc/en/user/source/services/wms/nonstandardautonamespace.rst
```diff
@@ -0,0 +1,50 @@
...
+
+The WMS standard supports a small number of "automatic" coordinate reference systems that include a user-selected centre of projection.  These are specified using::
+
+    AUTO:auto_crs_id,factor,lon0,lat0
+
+for example::
+
+    CRS=AUTO:42003,1,-100,45
+
+.. note::   in GeoTools 13 AUTO and AUTO2 namespaces are treated identically.
```
aaime: This is the GeoServer documentation, people might not be aware of GeoTools, so I'd reword with "GeoServer 2.7.x" or something like that. Also mind, the documentation itself is versioned, this one is particular is the documentation for GeoServer 2.8.x, so there might be no need to specify versions. What we normally do, to help people that stumble into the docs from a web search, is to add something like "Support for this vendor parameter has been added since version x.y.z"

doc/en/user/source/services/wms/nonstandardautonamespace.rst
```diff
@@ -0,0 +1,50 @@
...
+    * - 42001   
+      - Universal Transverse Mercator
+    * - 42002   
+      - Transverse Mercator
+    * - 42003   
+      - Orthographic
+    * - 42004   
+      - Equirectangular
+    * - 42005
+      - Mollweide   (not supported in GeoTools 13)
```
aaime: Same as above

PR #1067, opened 2015-05-16
GEOS-7026: fix failing tests on OpenJDK after GEOS-7020 fix

mbarto:


bencaradocdavies:
@mbarto thanks for the quick turnaround. I can confirm that this change fixes gs-wms tests for openjdk7, openjdk8, oraclejdk7, and oraclejdk8. If Travis CI passes and @aaime is happy to have antialiasing turned off, please merge and backport to 2.7.x and 2.6.x.

mbarto:
@bencaradocdavies @aaime I think it was the original intention to turn off antialias (please @aaime confirm or deny), there is code to set an hint, but unfortunately it is set too late, when the Graphics2D object has already been created using other hints created from the antialias format_options request param and this one defaults to antialiasing set to on

bencaradocdavies:
@mbarto full build with -Prelease passes with locally for me with openjdk7. Travis CI fails for this PR only on oraclejdk7 with a failure in gs-web-sec-core. Looks unrelated. I have seen similar failures before with oraclejdk7, and only in a full build. Nasty, but AFAIKT not your problem.

aaime:
Yes @mbarto, that's fine.

PR #1070, opened 2015-05-16
[GEOS-7027] New generic built-in style working with most layers

aaime:


aaime:
Mind, it depends on https://github.com/geotools/geotools/pull/842 being merged first

aaime:
Applied feedback, fixed headers.

PR #1071, opened 2015-05-17
[GEOS-7029] Add GUI to enable/disable advanced projection handling and map wrapping

aaime:


aaime:
This pull request depends on https://github.com/geotools/geotools/pull/844, although it will probably work even without it being applied (the geotools one ensures we avoid the two bbox queries)

aaime:
Still lacks docs, a test, and a fix in GetFeatureInfo, please do not merge

aaime:
Added docs and tests, merging

PR #1073, opened 2015-05-18
ConfigurableBlobStore must add listeners to delegate

groldan:
ConfigurableBlobStore must add listeners to delegate regardless of its configuration status.

ConfigurableBlobStore is precluding blob store listeners from being added to
the delegate blob store at start up time, since GWCInitializer.initialize
calls ConfigurableBlobStore.setChanged() after the beans have been loaded.

This is making the disk-quota subsystem unusable, since it needs to register
a blobstore listener at startup time. Moreover, adding the listener should
not depend on whether the ConfigurableBlobStore has set up its caching
preferences.

### DIFF
--- a/src/gwc/src/main/java/org/geoserver/gwc/ConfigurableBlobStore.java
+++ b/src/gwc/src/main/java/org/geoserver/gwc/ConfigurableBlobStore.java
@@ -21,6 +21,7 @@ import org.geoserver.platform.GeoServerExtensions;
 import org.geotools.util.logging.Logging;
 import org.geowebcache.storage.BlobStore;
 import org.geowebcache.storage.BlobStoreListener;
+import org.geowebcache.storage.BlobStoreListenerList;
 import org.geowebcache.storage.StorageException;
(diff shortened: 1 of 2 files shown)


aaime:
Hi Gabriel, the delegate can be modified at runtime from the GUI, won't you lose the listeners when that happens? How about collecting the listeners in the ConfigurableBlobStore, and then attach them to the current delegate on configuration when configureBlobStore is called instead?

groldan:
Hi Andrea, sorry for the long delay, been swamped by other stuff. I've updated the PR so that the listeners are not lost but reapplied instead when the config changes. Let me know if it looks better.

groldan:
hey @aaime I'm preparing a PR for GWC pluggable blobstores + S3 support. It'd be nice if this PR can be applied first, or taken care of, given its an independent bug introduced by ConfigurableBlobStore which makes disk quota unusable. Let me know if this last patch looks OK to you. TIA.

aaime:
Ah sorry, I did look at this one yesterday but then forgot to comment!
+1

PR #1075, opened 2015-05-18
GEOS-7020: additional fix to avoid overflows of hitArea outside of pa…

mbarto:
…inted area


aaime:
Ouch, this one did not make into the release??

mbarto:
nope

aaime:
How easy is it to trigger this failure? Do we have a 2.7.1 with GetFeatureInfo not working in common cases?

mbarto:
From what I have seen it can happen quite easily

PR #1076, opened 2015-05-19
[GEOS-7025] Stopped GeoServerDataDirectory.parsedStyle from creating files for nonexistant URLs

tbarsballe:
Resource.file() will create the file if it doesn't exist. Changed to use a method that does not create the file, only generates the correct path.

### DIFF
--- a/src/main/src/main/java/org/geoserver/config/GeoServerDataDirectory.java
+++ b/src/main/src/main/java/org/geoserver/config/GeoServerDataDirectory.java
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2015 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
@@ -1234,7 +1234,9 @@ public class GeoServerDataDirectory implements ResourceStore {
             public URL locateResource(String uri) {
                 URL url = super.locateResource(uri);
                 if(url.getProtocol().equalsIgnoreCase("resource")) {
-                    return fileToUrlPreservingCqlTemplates(urlToResource(url).file());
+                    //GEOS-7025: Just get the path; don't try to create the file
+                    return fileToUrlPreservingCqlTemplates(
+                            Paths.toFile(root(), urlToResource(url).path()));
                 } else {
                     return url;
                 }
--- a/src/main/src/test/java/org/geoserver/config/GeoServerDataDirectoryTest-applicationContext.xml
+++ b/src/main/src/test/java/org/geoserver/config/GeoServerDataDirectoryTest-applicationContext.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- 
+ Copyright (C) 2014 - Open Source Geospatial Foundation. All rights reserved.
+ This code is licensed under the GPL 2.0 license, available at the root
+ application directory.
+ -->
+<!DOCTYPE beans PUBLIC "-//SPRING//DTD BEAN//EN" "http://www.springframework.org/dtd/spring-beans.dtd">
+
+<beans>
+  <bean id="geoserverExtensions" class="org.geoserver.platform.GeoServerExtensions" />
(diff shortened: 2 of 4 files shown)


src/main/src/main/java/org/geoserver/config/GeoServerDataDirectory.java
```diff
@@ -1234,7 +1234,9 @@ else if ( l.getResource() instanceof WMSLayerInfo ) {
             public URL locateResource(String uri) {
                 URL url = super.locateResource(uri);
                 if(url.getProtocol().equalsIgnoreCase("resource")) {
-                    return fileToUrlPreservingCqlTemplates(urlToResource(url).file());
+                    //GEOS-7025: Just get the path; don't try to create the file
+                    return fileToUrlPreservingCqlTemplates(
+                            Paths.toFile(root(), urlToResource(url).path()));
```
aaime: Could work, although I'd have expected it to test the type of resource, and check if it existsed, returning it in one case, returning null in the other. Could be that returnig null is bad semantic wise though, no strong opinion

aaime:
Added a comment, but no strong opinion on it.
Can you add a test?

src/main/src/main/java/org/geoserver/config/GeoServerDataDirectory.java
tbarsballe: That was my origional approach as well, but that caused it to throw a very misleading exception down the line.
This way the error message actually reflects the fact that the file is missing.

aaime: All good then... besides the missing test :-)

tbarsballe:
Test case coming shortly.

tbarsballe:
Test case added. I was kind of suprised that we didn't already have a GeoServerDataDirectoryTest - in the future this could definately be updated with tests for existing methods in GeoServerDataDirectory.

aaime:
Looks good, maybe wait for Travis to verify it builds fine, and then feel free to merge.
About tests missing, you should have seen GeoServer when I met it 8 years ago, there was no test whatsoever, we only had the cite tests to run manually before the releases... we came a long way since then :-)

tbarsballe:
Okay, I am a bit confused about the Travis Build failure we are seeing here - builds fine locally (Oracle JDK 7), the failure is seemingly unrelated to my changes, and the other two builds pass. How stable is the Travis Build right now?

bencaradocdavies:
@tbarsballe I have seen the failure in gs-web-sec-core testFill(org.geoserver.security.web.data.EditDataAccessRulePageTest) when building locally with oraclejdk7 and -T4 and occasionally on Travis CI. It appears to be intermittent. The failure only occurs for me with oraclejdk7 and only for multithreaded builds. Running tests in this module alone always seems to pass for me (single threaded). I have not seen the failure on Boundless Jenkins.

@tbarsballe in my view this failure is unrelated to your changes. I will now merge.

bencaradocdavies:
The merged build passed on Travis CI ... and failed on Boundless Jenkins.  :-/

bencaradocdavies:
Boundless Jenkins passed when kicked. The failure is intermittent.

tbarsballe:
Given that the test failures we were seeing seem intermittent and unrelated, and this change does not appear to be causing any problems over the last few weeks on master, I will backport this bugfix to 2.7.x and 2.6.x, if no one has any objections.

NielsCharlier:
I think this change likely caused https://osgeo-org.atlassian.net/browse/GEOS-7741

When you are using another kind of resource store, like jdbcresourcestore, it is crucial that any resource referred to from the parsed style is cached as a file. The style parser is in geotools and does not know about resources, it needs files.

tbarsballe:
This fix was specifically intended for when a file that doesn't exist is referenced in an SLD (See the JIRA ticket for more details), although it is possible it has side-effects that could be causing the issue.

Also, doesn't this predate most of the ResourceStore work? - I thought that was mostly introduced in 2.9?

tbarsballe:
If changing this so that the file gets created if it doesn't exist, we will need to make sure the file only gets created if it is a valid resource in the resource store, else we will just get the original GEOS-7025 problem again.

NielsCharlier:
The call to file() would have been introduced by the resource work (before that, it would have used files to begin with). okay with your solution

PR #1079, opened 2015-05-20
[GEOS-7038] Added support for prettyPrint to REST GET Style

tbarsballe:
No test case for this one since the parameter only really has any effect when using extended style formats such as CSS (eg convert a CSS style to SLD, or vise-versa). This change was tested upstream with a style extension and did work correctly.


tbarsballe:
Backported to 2.7.x as 41338fc0bff448858d2cfb566f49b46c032c372a

PR #1083, opened 2015-05-21
[GEOS-7041] - Use online INSPIRE schemaLocation

bgsmase:
The INSPIRE view schema is hosted at http://inspire.ec.europa.eu/schemas/inspire_vs/1.0/inspire_vs.xsd
Current local copy imports other online Schemas so doesn't provide any
benefit for disconnected servers.

See https://osgeo-org.atlassian.net/browse/GEOS-7041


src/extension/inspire/src/main/java/org/geoserver/inspire/wfs/WFSExtendedCapabilitiesProvider.java~
```diff
@@ -1,137 +0,0 @@
-/* Copyright (c) 2001 - 2013 OpenPlans - www.openplans.org. All rights reserved.
```
bencaradocdavies: @bgsmase thanks for removing this unwanted file!

src/extension/inspire/src/test/java/org/geoserver/inspire/wms/WMSExtendedCapabilitiesTest.java
```diff
@@ -28,7 +27,7 @@ public void testExtendedCaps() throws Exception {
         getGeoServer().save(wms);
 
         Document dom = getAsDOM("wms?request=getcapabilities");
-        assertEquals(NAMESPACE, dom.getDocumentElement().getAttribute("xmlns:inspire_vs"));
+        assertEquals(VS_NAMESPACE, dom.getDocumentElement().getAttribute("xmlns:inspire_vs"));
```
bencaradocdavies: @bgsmase are there any suitable locations to make asserts against the schemaLocation? This would confirm that your desired behaviour is implemented and help ensure that it is retained in future. Need test coverage!

bgsmase:
I've just added a test for the schemaLocation. 

(N.B. I have added various other tests with the code I intend to commit for https://osgeo-org.atlassian.net/browse/GEOS-7042 and will add a similar schemaLocation test for WFS download service with it too.)

bgsmase:
Don't understand the Travis CI build failure but it doesn't seem to be anything to do with the INSPIRE module

aaime:
Yeah, that's a known intermittent failure that tends to happen on OS running inside virtual machines. @mcrmcr should be having a look at it

bencaradocdavies:
@bgsmase this intermittent gs-web-sec-core failure is reported here and unrelated to your changes:
https://osgeo-org.atlassian.net/browse/GEOS-7037

@aaime it is not just inside VMs: any system under load can exhibit the failure in Maven or Eclipse.

bencaradocdavies:
Thanks, @bgsmase. The test coverage you added looks sufficient.

PR #1084, opened 2015-05-23
[GEOS-7032] You can see the content of the server by sending the right request to GeoServer

aaime:
This one prevents XML entity expansion attacks that can make GeoServer display the contents of arbitrary files in the file system, and should also prevent "XML bombs".

* https://osgeo-org.atlassian.net/browse/GEOS-7032


aaime:
This one prevents XML entity expansion attacks that can make GeoServer display the contents of arbitrary files in the file system, and should also prevent "XML bombs".

This work follows up pr #193 by tapping some overlooked xml parsing paths and covering new ones that weren't there back at the time

aaime:
Ah, please review carefully as this is a fix for a security issue, that we need to backport to the stable and maintenance branches.

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ppio/XMLPPIO.java
```diff
@@ -76,6 +80,33 @@ public void encode( Object value, OutputStream os) throws Exception {
...
+     * @param configuration
+     * @return
+     */
+    protected Parser getParser(Configuration configuration) {
+        Parser parser = new Parser(configuration);
+        // spring injection would have been preferred, but it changes the order in which PPIO
+        // are listed, which in turn chances the behavior of WPS (e.g., default mimeType)
+        parser.setEntityResolver(getResolveProvider().getEntityResolver());
+        return parser;
+    }
```
jodygarnett: So this makes sense, the fundamental change is supplying a EntryResolver to parsers before use.

src/main/src/main/java/org/geoserver/catalog/SLDHandler.java
```diff
@@ -5,11 +5,28 @@
  */
```
jodygarnett: So general feedback that header needs to be updated (yeah I know it is not that important ...)

src/main/src/main/java/org/geoserver/util/NoExternalEntityResolver.java
```diff
@@ -25,6 +25,7 @@
  */
 public class NoExternalEntityResolver implements EntityResolver {
 
+    public static final String ERROR_MESSAGE_BASE = "Entity resolution disallowed for ";
```
jodygarnett: What does taking this out as a static final accomplish?

jodygarnett:
So this pull request does what it sets out to and should be good to go.

As this is a follow up to #193 is there anyway we can keep this problem from being reintroduced in the future?

smithkm:
Fixed the headers and the conflict here: https://github.com/smithkm/geoserver/tree/entity_resolve_fix

src/main/src/main/java/org/geoserver/util/NoExternalEntityResolver.java
aaime: Avoid repeating over and over the same string among differen tests, see all the tests checking for this string

aaime:
Kevin, if it's good to go, feel free to close this one, push yours, and merge

smithkm:
@aaime Yes that was the plan but I keep running into build issues when I try to do a local test. (three way merges always make me wary)  It seems to be working now and hopefully will be done (successfully) by the time I have the GWC 1.8-M0 release done.

PR #1091, opened 2015-05-27
[GEOS-7049] Allow autonomous and asynchonous data fetching in importer REST api

aaime:


doc/en/user/source/extensions/importer/rest_reference.rst
```diff
@@ -180,6 +184,13 @@ The response to the above POST request will be::
 	    ]
 	  }
 	}
+	
+The operation of populating the tasks can require time, especially if done against a large set of
+files, or against a "remote" data (more on this later), in this case the POST request can include ``?async=true``
+at the end of the URL to make the importer run it asynchronously. 
+In this case the import will be created in INIT state and will remain in such state until all
+the data transfer and task creation operations are completed. In case of failure to fetch data
+the import will immediately switch from
```
tbarsballe: This sentance ends prematurely

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -323,12 +360,37 @@ public void init(ImportContext context) throws IOException {
...
+
+                addTasks(context, data, prepData);
+            }
+
+            // switch from init to pending as needed
+            if (context.getState() == ImportContext.State.INIT) {
+                context.setState(ImportContext.State.PENDING);
+            }
+        } catch (Exception e) {
+            LOGGER.log(Level.SEVERE, "Failed to resolve remote data", e);
```
tbarsballe: If we are doing a REST import (for example), is there any way of viewing capturing this error? As it stands, it looks like there is no way (without looking at the GeoServer logs) of telling the difference between an empty remote directory and an error - both result in a completed import with no tasks/layers.
Maybe an error field could be included in the import context, similar to tasks?

tbarsballe:
I've reviewed this, looks good for the most part, added comments on a couple issues.

The only other minor concern I have is that the password is stored in plain text / String form everywhere. This seems like a bit of a security concern, but I don't know what (if anything) should be done about it.

doc/en/user/source/extensions/importer/rest_reference.rst
aaime: Will fix

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
aaime: In another spot I actually put the error message in the Data itself, I guess I can do the same for this one.

aaime:
Eh, security wise, guides suggest to use char[] because you can wipe its contents immediately after you used the password, without having to wait for GC to clean stuff, and I guess I could do that, except that CommonsVFS requires us to build a string representation of the password, so we're almost back to square one... one could argue that string is attached to a shorter lived object, and it's true, but GC might kick in quite seldomly if the server has a lot of memory. Once the data transfer is done the RemoteData will indeed be replaced by another data and become eligible for garbage collection, so the time the string stays in unaltered form is pretty much the same.
In GeoServer we encrypt passwords when storing them in persistent storage though, so the bdb storage should be modified, I guess I can register a custom bindings in Importer.initXStreamPersister that does encryption for the RemoteData password.

aaime:
Ok, security issues fixed as possible (in the bdb module)... and discovered the xstream serializer was not even working, fixed it as well, and made sure passwords are encrypted both with standard serialization and the xstream one

tbarsballe:
Looks good, passes Travis CI. Do you want to squash the two "Work in progress" commits and then merge?

aaime:
Doh, forgot to squash them, will do

PR #1093, opened 2015-05-29
[GEOS-7042] - Assorted INSPIRE extension improvements

bgsmase:
Remove namespace declarations not used in extended capabilities
Remove xsi:type attributes unnecessary because XML Schema fully defines concrete type
Allow the service MetadataURL/MediaType element to be omitted if no value set.
Don't output the extended capabilities element at all if required parameters have not been set.
More tests for different combinations of parameters including test for schemaLocation of download services Schema.


aaime:
Had a look, seems to be well done. Only thing, the copyright headers were not updated, I bet some file does not have 2014 - 2015 yet.  Kevin has shared a useful command to do the updates automatically on unix like systems: 
``sed -i 's/\(([cC])\) 2014 Open Source Geospatial Foundation/\1 2014 - 2015 Open Source Geospatial Foundation/' $(git diff-tree --no-commit-id --name-only -r HEAD)``

bgsmase:
Ran the script to update headers and added headers to some files where there wasn't any.

aaime:
Merged. Wondering, do the docs for the INSPIRE extension need to be updated? http://docs.geoserver.org/latest/en/user/extensions/inspire/using.html
The docs source is also in github: https://github.com/geoserver/geoserver/tree/master/doc/en/user/source/extensions/inspire

bgsmase:
Yes docs will need updating but I was going to do this and submit a pull request in one go after submitting pull requests and seeing results for https://osgeo-org.atlassian.net/browse/GEOS-7043 and https://osgeo-org.atlassian.net/browse/GEOS-6435 as well. I will do 7043 next.

PR #1095, opened 2015-05-29
GEOS-6987 - Support new GWC Parameter Filter options via UI

smithkm:
https://osgeo-org.atlassian.net/browse/GEOS-6987

Adds new UI strings that require localization.  

Uncovered and fixed issues upstream in GWC: https://github.com/GeoWebCache/geowebcache/pull/338 https://github.com/GeoWebCache/geowebcache/pull/339


aaime:
Hi Kevin, I don't see any doc, it may be nice for people that, like me, do not know what these new parameter filters are about.

aaime:
Thanks. Played a bit with the GUI, did you know that if you add two param filters, edit the first, then go and delete the second one, you also lose all the values of the first? (e.g., add two param filters, test1, test2, fill test1 values, delete test2, all the test1 values are gone).

Also, as a matter of consistency, there are other lists of values in GeoServer (like the limited srs list), they are comma separated, not one per line.

smithkm:
Well, the multiline input has been like that for almost 2 years.  Given the parameters might include things like CQL filters that contain commas, lines seem like a safer choice  (Come to think of it, I'm not sure if newlines would survive the trip though XStream)

I hadn't noticed that behaviour with delete before though.

aaime:
I see... too bad for the inconsistency, but I guess the users are already used to the multiline thing then

smithkm:
Ticketed the delete bug separately as it wasn't introduced by this change (GEOS-7068)[https://osgeo-org.atlassian.net/browse/GEOS-7068]

smithkm:
@NielsCharlier fixed the delete issue #1103

smithkm:
@aaime anything else?

aaime:
Nope, go ahead

PR #1096, opened 2015-05-30
Update configuration doc to incluse gs- prefix for web jars

kalxas:
I recently tried to remove web-*.jar files from a GeoServer 2.7.1 installation and I figured out the docs are outdated.


aaime:
Yep, thanks!

PR #1100, opened 2015-06-07
Added missing escape symbol

kolosov-sergey:
This caused transifex to think, that another property has began with key "override"


aaime:
Yep, thank you

aaime:
Backported fix to 2.7.x and 2.6.x too

PR #1101, opened 2015-06-09
[GEOS-7065] Fix missing null check

tbarsballe:


aaime:
Looks ok to me, let's see what Travis thinks of it

tbarsballe:
One test in the CSW module fails on Travis on Oracle JDK7 only - doesn't seem related to the changes. How stable is the Travis build right now? I know it has been unreliable in the past.
Oracle JDK 8 and OpenJDK 7 both pass.

bencaradocdavies:
Looks like an intermittent CSW failure. Merged during committee meeting.

PR #1102, opened 2015-06-11
Fix GEOS-7069

jia020:
https://osgeo-org.atlassian.net/browse/GEOS-7069
Fix the Jenkins geoserver app-shcema online testing failure.
the spatialIndex should be cleared after using.


bencaradocdavies:
Thanks, @jia020 . Looks reasonable to me, and I see that @victortey was happy to merge. A couple of comments:

(1) An alternate solution would be to move the declaration and initialisation of "spatialIndex" inside the scope of the outer for loop (where "file" is declared and initialised) so a new StringBuffer instance is created for each iteration. I think this would have the same effect and, by reducing the scope, reduce the likelihood of state leakage across iterations.

(2) @jia020 your pull request includes an unnecessary merge commit of a remote tracking branch and includes several commits that are unrelated to this change. This pull request has a complicated history with seven commits for a single line change. (It does no harm to the code base as git merges everything into the single line change; only history becomes more complicated.) To improve the clarity of pull requests, please consider making a branch for your pull request and cherry-picking or making a squash commit with just your desired change. This will make history easier to follow.

Thanks again! I will close the Jira issue.

Kind regards,
Ben.

PR #1104, opened 2015-06-14
Fixing GEOS-7071, Security Test Setup not compatible with IBM Java

mcrmcr:
After merging, I will backport to 2.7.x series, hope this is ok


aaime:
Looks good to me.

PR #1107, opened 2015-06-17
[GEOS-7064] Full text search unit tests in JDBCConfig failing

smithkm:
[GEOS-7064](https://osgeo-org.atlassian.net/browse/GEOS-7064)

Fixed a mistake in the tests, and corrected a bug they uncovered in JDBCConfig.

@groldan, `lookUpRelatedObject` seems to all be your code.  Do my changes make sense to you?


groldan:
@smithkm patch looks good to me

smithkm:
Thanks Gabriel.

smithkm:
It appears that although this gets the unit test to pass, full text search is still broken for keywords in JDBCConfig.

PR #1108, opened 2015-06-18
JSON matched features count bug with spatial filter

PaulBiskup:
Patch for  JSON matched features count.bug when applying a spatial filter in a different projection than layer-projection.

See https://osgeo-org.atlassian.net/browse/GEOS-7075 for more information.

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
@@ -430,14 +430,11 @@ public class GeoJSONGetFeatureResponse extends WFSGetFeatureOutputFormat {
             }
             
             int count = 0;
-            count = source.getCount(countQuery);
-            if (count == -1) {
-                // information was not available in the header!
-                org.geotools.data.Query gtQuery = new org.geotools.data.Query(countQuery);
-                FeatureCollection<? extends FeatureType, ? extends Feature> features = source
-                        .getFeatures(gtQuery);
-                count = features.size();
-            }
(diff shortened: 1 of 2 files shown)


src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
```diff
@@ -430,14 +430,11 @@ private int getFeatureCountFromWFS11Request(Operation operation, WFSInfo wfs)
...
             int count = 0;
-            count = source.getCount(countQuery);
-            if (count == -1) {
-                // information was not available in the header!
-                org.geotools.data.Query gtQuery = new org.geotools.data.Query(countQuery);
-                FeatureCollection<? extends FeatureType, ? extends Feature> features = source
-                        .getFeatures(gtQuery);
-                count = features.size();
-            }
+            // information was not available in the header!
```
aaime: Confused about this changes, how is this handling reprojection any better than the old one? I was expecting to see some usage of the reprojecting filter visitor instead. 
Also, while normally the performance of this one should be good, some collection implementations may go directly to actually fetch the actual features just to count them, instead of trying to use the feature source count method, which will try some optimized access (e.g., select count(*) in a dbms). I say some, because the common collection we get nowadays, ContentFeatureCollection, will try to call source.getCount(...) internally before falling back on fetching features.

PaulBiskup:
I was not that deep in the GeoServer/GeoTools-code and did not know about the ReprojectFilterVisitor.

Thanks for the hint.

I have created another pull request using the ReprojectFilterVisitor.

https://github.com/geoserver/geoserver/pull/1117

PR #1110, opened 2015-06-18
Removed geoserver-*-src.zip from assembly phase

bencaradocdavies:
See discussion with @jodygarnett:
http://osgeo-org.1560.x6.nabble.com/Smoke-testing-GeoServer-2-6-4-release-artifacts-td5211514.html


aaime:
Yep, seems like the right way to remove the assembly (at least to me)

bencaradocdavies:
Thanks, @aaime.

PR #1114, opened 2015-06-19
GWC pluggable blobstores configuration

groldan:
Incorporate GWC support for S3 blob stores and implements the configuration UI to set up multiple blob stores and assign blobstores to tile layers.


groldan:
Any comment on this one anyone? If not that's fine, I'll just assume it's ok to merge.

src/gwc/src/main/java/org/geoserver/gwc/GWC.java
```diff
@@ -2058,5 +2066,151 @@ public void testQuotaConfiguration(JDBCConfiguration jdbcConfiguration) throws C
...
+                c.setDefault(false);
+            }
+        }
+        stores.add(config);
 
+        setBlobStores(stores);
+    }
+
+    /**
+     * Convenience method to modify a blobstore; calling {@link #setBlobStores(List)} with the
```
smithkm: "Modify a blobstore" is a bit vague.  Maybe "Update the configuration of a blobstore" would be clearer.

src/gwc/pom.xml
```diff
@@ -85,6 +85,10 @@
       <groupId>org.geowebcache</groupId>
       <artifactId>gwc-diskquota-jdbc</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.geowebcache</groupId>
+      <artifactId>gwc-aws-s3</artifactId>
```
smithkm: Should this maybe be a plugin instead of in the core GS GWC module?  How heavy are the AWS dependencies that get pulled in as a result?

src/web/gwc/src/main/java/org/geoserver/gwc/web/blob/BlobStoreTypeStore.java
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * A singleton store of all supported blobstore types.
+ *
+ * @author Niels Charlier
+ */
+public class BlobStoreTypeStore {
+
+    protected static BlobStoreTypeStore THE_STORE = new BlobStoreTypeStore();
+
+    public static BlobStoreTypeStore getInstance() {
```
smithkm: The static singleton pattern always makes me wary.  It can be a serious pain to build unit tests around it properly.  Might this be done using Spring instead?

NielsCharlier: I don't particularly see the point to use Spring here, to be honest. This class is only used for wicket and wicket and spring don't really go together well, you always have to call the static "bean" method to obtain the instance anyway which actually breaks the ioc principle that spring is for. It seems to me that adding spring stuff for this class is like generating unnecessary code, but I could be wrong. What sort of test would be problematic with this class? It only serves to provide data and that could be tested

src/gwc/pom.xml
groldan: quite a bit:
```
[     ] +- org.geowebcache:gwc-aws-s3:jar:1.8-SNAPSHOT:compile
[544K ] |  \- com.amazonaws:aws-java-sdk-s3:jar:1.9.30:compile
[247K ] |     \- com.amazonaws:aws-java-sdk-kms:jar:1.9.30:compile
[488K ] |        \- com.amazonaws:aws-java-sdk-core:jar:1.9.30:compile
[577K ] |           +- org.apache.httpcomponents:httpclient:jar:4.3.4:compile
[276K ] |           |  \- org.apache.httpcomponents:httpcore:jar:4.3.2:compile
[894K ] |           +- com.fasterxml.jackson.core:jackson-databind:jar:2.3.2:compile
[ 35K ] |           |  +- com.fasterxml.jackson.core:jackson-annotations:jar:2.3.0:compile
[195K ] |           |  \- com.fasterxml.jackson.core:jackson-core:jar:2.3.2:compile
[608K ] |           \- joda-time:joda-time:jar:2.8.1:compile (version selected from constraint [2.2,))

[3864K total]
```
Yet making it a plugin would add some unneeded complexity esp. to the UI.
I'll ask on the mailing list.

NielsCharlier: Just realised... if S3 becomes an extension we would need to change the singleton design anyway... the S3 gui would also need to be separated and the blobstore types (with their own gui pages) would need to be extendible, with indeed something like spring, and we couldn't use the read-only singleton obviously.

src/web/gwc/src/main/java/org/geoserver/gwc/web/blob/BlobStoreTypeStore.java
groldan: I concur the singleton pattern is prone to errors even if in this case it doesn't look like. I removed it in commit 417663f45d496e1c57a0d64c9f0d83ceb066d8da anyways.

jodygarnett:
what is happening on this one?

groldan:
Just updated with the S3 bits as a community module with its own, resolved merge conflicts.
If no other objections arise this should be ready for a merge onto master.

smithkm:
Looks good to me.

PR #1117, opened 2015-06-21
Json featurecount spatialfilter bug reproject using ReprojectingFilterVisitor

PaulBiskup:
Patch for JSON matched features count.bug when applying a spatial filter in a different projection than layer-projection.

This patch is using ReprojectingFilterVisitor if a spatial filter is passed.

See https://osgeo-org.atlassian.net/browse/GEOS-7075 for more information.


src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
```diff
@@ -417,9 +420,19 @@ private int getFeatureCountFromWFS11Request(Operation operation, WFSInfo wfs)
             FeatureSource<? extends FeatureType, ? extends Feature> source = meta.getFeatureSource(
                     null, null);
             Filter filter = query.getFilter();
+            
             if (filter == null) {
                 filter = Filter.INCLUDE;
             }
+            else if (filter instanceof org.geotools.filter.spatial.AbstractPreparedGeometryFilter){
```
aaime: This makes the bug easy to reproduce with a filter like "myatt = 5 && bbox(the_geom, x1,y,1,x2,y2)"
There is no need to check for the type of root filter, the visitor will scan the entire filter and reproject every spatial bit in it.

jodygarnett:
This looks okay to me, you missed updating the header but we can do that when merging.

Any additional feedback @aaime or can I merge this in.

aaime:
We have @sampov2 working on a more general fix, but not sure when it's going to be ready. @sampov2 do you have a deadline for us? The 2.8-beta release is now scheduled for Aug 28th if I remember correctly

sampov2:
Regrettably my client does not have enough budget for the fix @aaime referred to (https://osgeo-org.atlassian.net/browse/GEOS-7150). So that will not happen in time for 2.8.

aaime:
Replaced by https://github.com/geoserver/geoserver/pull/1186 (fixed copyright headers, code convention violations, squashed in a single commit). If the Travis build is green I'll backport to 2.7.x and if possible 2.6.x

PR #1119, opened 2015-06-22
[GEOS-7086] Make sure if namespace is deleted, associated workspace is deleted too

ridethepenguin:
See mailing list and associated [JIRA issue](https://osgeo-org.atlassian.net/browse/GEOS-7086) for discussion.

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/NamespaceWorkspaceConsistencyListener.java
+++ b/src/main/src/main/java/org/geoserver/catalog/NamespaceWorkspaceConsistencyListener.java
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2015 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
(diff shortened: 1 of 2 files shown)


src/main/src/main/java/org/geoserver/catalog/NamespaceWorkspaceConsistencyListener.java
```diff
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2015 Open Source Geospatial Foundation - all rights reserved
```
mbarto: You should add 2015 without removing 2014, 2014 - 2015

src/restconfig/src/test/java/org/geoserver/catalog/rest/NamespaceTest.java
```diff
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2015 Open Source Geospatial Foundation - all rights reserved
```
mbarto: Same as above

ridethepenguin:
Thanks @mbarto, headers fixed.

simboss:
Is there a JIRA issue for this PR? Are we going to backport?

ridethepenguin:
@simboss created issue [GEOS-7086](https://osgeo-org.atlassian.net/browse/GEOS-7086) on JIRA. Backporting should be straightforward, I'll do it after this is merged.

aaime:
Merged, thanks

PR #1120, opened 2015-06-22
[GEOS-7085] Add the proxy base URL to metadata links

sikeoka:
[GEOS-7085] Add the proxy base URL to metadata links, if necessary, in WCS DescribeCoverage and WCS/WFS GetCapabilities responses.


aaime:
Looks good, thank you

PR #1121, opened 2015-06-22
Importer default transformations and direct execute

aaime:
See docs for an example of both in the same request.


src/extension/importer/core/src/test/java/org/geoserver/importer/ImporterDataTest.java
```diff
@@ -986,4 +986,5 @@ public void testIllegalNames() throws Exception {
         assertEquals("a_1_", task.getLayer().getName());
         assertEquals("a_1_", task.getLayer().getResource().getName());
     }
+
```
tbarsballe: Whitespace is the only change in this file

src/extension/importer/rest/src/test/java/org/geoserver/importer/rest/ImporterIntegrationTest.java
```diff
@@ -0,0 +1,262 @@
+package org.geoserver.importer.rest;
```
tbarsballe: Missing header?

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -832,10 +832,13 @@ public void changed(ImportTask task)  {
         changed(task.getContext());
     }
 
-    public Long runAsync(final ImportContext context, final ImportFilter filter) {
+    public Long runAsync(final ImportContext context, final ImportFilter filter, final boolean init) {
```
tbarsballe: In the interests of preserving API consistency, might it be better to overload this method with:
```
public Long runAsync(final ImportContext context, final ImportFilter filter) {
    runAsync(context, filter, false);
}
```

doc/en/user/source/extensions/importer/rest_examples.rst
```diff
@@ -566,4 +566,60 @@ a lot of shapefiles to process, also the import run will be done in asynchronous
...
+Importing and optimizing a large image with a single request
+------------------------------------------------------------
+
+A large image appears every now and then on a mounted disk share, the image needs to be
+optimized and imported into GeoServer as a new layer.
+Since the source is large and we need to copy it on the local disk where the data dir resides,
+a "remote" data is the right tool for the job, an asynchronous execution is also recommended
+to avoid waiting on a possibly large command.
+In this case the request will also contains the "exec=true" parameter to force the importer
+an immediate execution of the command.
```
tbarsballe: Some small tweaks to grammer/wording to improve readability:
```
When a large image is published on a mounted disk share, it may need to be optimized and imported it into GeoServer as a new layer.
Since the source is large, the "remote" data type is used copy it to the local disk where the data directory resides. An asynchronous excecution is also recommended to avoid waiting on a potentially large command.
In this case the request will also contain the "exec=true" parameter to force the importer to excecute the command immediately after initializing the context.
```

doc/en/user/source/extensions/importer/rest_examples.rst
```diff
@@ -566,4 +566,60 @@ a lot of shapefiles to process, also the import run will be done in asynchronous
...
+
+The request will then look as follows::
+
+    curl -u admin:geoserver -XPOST -H "Content-type: application/json" -d @import.json "http://localhost:8080/geoserver/rest/imports?async=true&exec=true"
```
tbarsballe: Based on the code, I think this should be `&execute=true`

src/extension/importer/core/src/test/java/org/geoserver/importer/ImporterDataTest.java
aaime: I know, just fixed a coding convention error

src/extension/importer/rest/src/test/java/org/geoserver/importer/rest/ImporterIntegrationTest.java
aaime: Adding it

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
aaime: Yep, done

doc/en/user/source/extensions/importer/rest_examples.rst
aaime: I've modified the code instead, since the other parameters are already in abbreviated form (async instead of asynchronous)

PR #1122, opened 2015-06-24
Fix build (tests) on Windows with Java 8

bradh:
I think both these fixes are Windows specific. They were done with java 8 though, which might have exposed the issues.


bradh:
GEOS-7090 - https://osgeo-org.atlassian.net/browse/GEOS-7090

aaime:
Hi, I've just tested on Windows with java 7, the test fails as modified, but works fine in its current version.

bradh:
OK, thanks for the advice. I don't see this as a noted change in Java 8, so will need to investigate further.

aaime:
Hi Brad, I have made one more test, with java 8 (u40)... the pull fails for me, and master builds fine instead. Maybe it's environment specific. Venturing a guess... are you building under cygwin, or the git shell?

bradh:
I'm doing it under netbeans.

aaime:
Interesting... have you tried with maven on the command line?

PR #1127, opened 2015-07-01
Restconfig: add GeoPackage format for store file upload

NielsCharlier:
Allows direct geopackage file upload through restconfig. Depends on https://github.com/geotools/geotools/pull/896 (not to build, but to work).


aaime:
+1

PR #1128, opened 2015-07-02
Store field deprecation, CQL filter in layers, and SOLR improvements

aaime:
This one packs the work needed to deprecate the field strategy in the SOLR store in 3 separate commits, in particular, handling deprecacted store config fields, exposing the CQL filter, and updating the SORL configurations so that it can effectively replublish the same "root" layer over and over with different configurations (sql view like).

CQL filtering wise, we followed the advices on the list, and avoided parsing and generating the CQL from a Filter object representation, but storing the CQL instead and parsing it as needed.


aaime:
Anyone has feedback or wants to review?
Just let me know, otherwise I'll merge this one tomorrow

mbarto:
Just one System.out.println to remove, apart from that looks good

PR #1129, opened 2015-07-03
Moved WCS 1.1 schema to GeoTools

tbarsballe:
Migrated WCS 1.1 schema to GeoTools for consistency with WCS 1.0 and WCS 2.0.
Depends on https://github.com/geotools/geotools/pull/904

PR #1130, opened 2015-07-03
[GEOS-7095] Fix for exploitable bypass for XXE fix

tbarsballe:
In order to get the test cases to pass, WFS 1.1 uses the wfs.xsd from geotools (This is because maven/eclipse will use `file:` paths for resources in the current module, instead of the `jar:` paths used when the module is packaged).
Depends on https://github.com/geoserver/geoserver/pull/1129 to pass tests.


aaime:
Looks good. Please squash commits and remember to update the copyright headers last modified year for the files you touched. Kevin shared a nice script that does the updates to the headers as needed, check it out on the geoserver-devel ml.

tbarsballe:
Updated headers (where applicable) and squashed.

PR #1131, opened 2015-07-07
developers guide tutorial review and cleanup (GSIP-129)

jodygarnett:
Clean up of developers guide sections that are outdated or confusing.

See GSIP-129


aaime:
+1
Wondering, is there any header to change in the pom files?

jodygarnett:
yep you got me, should of updated the header...

PR #1132, opened 2015-07-07
replace roadmap planning with timeboxed release model (GSIP-129)

jodygarnett:
Clarifications on community process and release model as per https://github.com/geoserver/geoserver/wiki/GSIP-129


doc/en/developer/source/policies/community-process.rst
```diff
@@ -12,116 +12,58 @@ The GeoServer community process adheres to the following guidelines to ensure
 that it retains open development practices:
 
 * **Transparency** 
-  Project decisions are made in an open and transparent way. When a decision is
-  made it is clear as to why it was made.
+  Project decisions are made in an open and transparent way. Key decisions are presented *GeoServer Improvement Proposals* (:ref:`GSIP <gsip>`), discussed in public. When a decision is made it is clear as to why it was made.
```
aaime: I'd add, anyone can put together a proposal, no need for direct commit access.

doc/en/developer/source/policies/community-process.rst
```diff
@@ -12,116 +12,58 @@ The GeoServer community process adheres to the following guidelines to ensure
...
+The community maintains three active branches:
+
+* master: available for development of new functionality, documentation improvements and bug fixes
+* stable: bug fixes and back-port of new functionality that do not affect the GeoServer API or significantly affect stability.
+* maintenance: bug fixes
+
+GeoServer uses a six month development cycle on the master branch:
+
+  * month 1-5 master is open for development
+  * month 5 a beta release is made on a feature freeze
```
aaime: Actually at month 4 the beta, at month 5 the rc, at month 6 the final, otherwise we don't have a yearly cycle anymore

bencaradocdavies:
@jodygarnett do you think this should be reworded to maintain the yearly cycle?

doc/en/developer/source/policies/community-process.rst
jodygarnett: The final showed up as the initial release in the next section.

Did I do that wrong?

doc/en/developer/source/policies/community-process.rst
jodygarnett: Except that that kind of is as the wiki is not open to everyone right?

jodygarnett:
Not sure what you mean @bencaradocdavies - but if it is not clear it is not clear.

I agree this staggered release is a terrible thing to document clearly. I should just steal the diagram from the [website](http://geoserver.org/roadmap/) page.

doc/en/developer/source/policies/community-process.rst
aaime: Ah, that's correct.. but the wiki is also git managed, they cannot make a pull request, but technically they can still play this way: http://roman-ivanov.blogspot.it/2013/11/how-to-merge-github-wiki-changes-from.html
It's quite convoluted, not sure we can enough to describe that approach.

doc/en/developer/source/policies/community-process.rst
aaime: Ah, here is why it looked strange, you took a developer perspective, so the series begins when you can develop against it, I always took a user perspective, the series begins when the .0 is released. 

What you are describing in your comment however sounds like going back to the old development model, where we have a month of freeze between the beta and RC, didn't we say it was overkill?

Let me try to reword:
* First day of Month 1: RC of series N-1 is cut, branching it off, master is open for development of the series N
* First day of Month 2: final series N is released, development continues on master for N+1
* Month 3, 4, 5: development of N+1
* First day of month 6: Beta of N+1 is released
* First day of month 7 (Month 1 of N+1): RC of N+1 is released
* First day of month 8 (Month 2 of N+1): final of N+1 is released

And then we repeat. Overall, 7 months of work, with a month of overlap between the stabilization of N-1 and development of N.

doc/en/developer/source/policies/community-process.rst
jodygarnett: We are currently set up with "Any GitHub user can create and edit pages to use for documentation, examples, support, or anything you wish."

With this in mind I will add your clarification that proposals can be made by anyone.

PR #1134, opened 2015-07-07
responsible disclosure (GSIP-129)

jodygarnett:


aaime:
+1 
Maybe we should add a link to the PSC page so that people can see who's part of it?

jodygarnett:
I considered that, especially since PSC is not defined anywhere here. So yeah good idea.

PR #1135, opened 2015-07-07
Clarification on our expectations for submitting fixes (GSIP 129)

jodygarnett:


doc/en/developer/source/policies/patches.rst
```diff
@@ -9,6 +9,25 @@ is a great way to give back to the project.
 
 This document is a guide designed to help users through the process of successfully submitting a patch.
 
+.. note:: 
+
+   We would like to see all fixes applied to master first, and then back ported to the stable and maintenance branches. 
+   
+   We require the following for fixes submitted via pull requests or patch:
```
aaime: I would stop encouranging people submitting patches, we have a few of them in Jira, they are all rotting. With a pull request we know if the change builds, we can review it, and we know if the thing can actually merge cleanly.
With a patch there is too much work to apply, and it has to be done as soon as the patch arrives.

For all intents and purposes, patches have been ignored for the past few years, so I'd suggest we are explicit against people providing them, they are quite likely just wasting their time.

jodygarnett: This has been merged, want to take this to email and I can clean up this section to only talk about pull requests.

PR #1137, opened 2015-07-07
[GEOS-7098] Improve getStoreByName null handling

tbarsballe:
This fixes most of GEOS-7098; as long as there is a default store that matches the argument class, it will get returned. Default workspaces for getStoreByName have also been fixed.
Becuase CoverageStore and WMSStore do not support default store behavior, the behavior when trying to get the default stores of these classes is undefined.


aaime:
+1

tbarsballe:
Looks like the earlier build failure was intermittent. I'll go ahead and merge.

PR #1138, opened 2015-07-07
Case insensitive XXE fix (whitelist)

tbarsballe:
Using a whitelist of `systemId.matches("^(?i)(jar:file|http).*\\.xsd$")` to test for valid schema URLs

Updated test cases with `FILE://` and `.XSD` in place of `file://` and `.xsd` to confirm case insensitivity.

Old blacklist approach, with case instensitive fix, here: https://github.com/geoserver/geoserver/compare/master...tbarsballe:xxe-blacklist


bencaradocdavies:
See: https://osgeo-org.atlassian.net/browse/GEOS-7095

PR #1150, opened 2015-07-20
[i18n] Dutch properties files synched from Transifex

fgdrf:
Translation Team work at Transifex lead by @milovanderlinden updated Dutch properties. Thanks for that! This pull is an update for existing core modules and initial work for community modules and extensions.
- Frank


src/web/core/src/main/resources/GeoServerApplication_nl.properties
```diff
@@ -1,32 +1,33 @@
...
+AboutGeoServerPage.wiki          = Wiki
 
 AbstractCoverageStorePage.dataSrcName = Bron
 AbstractCoverageStorePage.description = Omschrijving
 
 AbstractDataAccessPage.cantGetDataStoreFactory = Kan geen plugin vinden voor de opgegeven bron. Controleer of de benodigde plugin is geinstalleerd
-AbstractDataAccessPage.dataSrcName = Bron
-AbstractDataAccessPage.description = Omschrijving
+AbstractDataAccessPage.dataSrcName             = Bron
+AbstractDataAccessPage.description             = Omschrijving
```
jodygarnett: The above looks wrong, as it only provides whitespace changes - and those whitespace changes may be a tab? Is this change intentional?

fgdrf: Actually we haven't edited the files manually, just synched from Transifex. I guess the tool tries to format each section to be better readable (each "=" is down below the one in line before). However, its not a tab, just whitespaces. The fomat is still machine readable and should'nt break any functionality.

PR #1160, opened 2015-07-28
OpenLayers 3 update to fix map preview issue

ahocevar:
This fixes an issue with the map preview when the map units are degrees and certain extents with long floating point coordinates are used. See https://github.com/openlayers/ol3/pull/3933 for details.


aaime:
Is there a GeoServer ticket in Jira for the release changelogs?
Is it something we need to backport to 2.7.x?

smithkm:
I've backported to 2.7 and 2.6 and created a ticket: [GEOS-7135](https://osgeo-org.atlassian.net/browse/GEOS-7135)

PR #1161, opened 2015-07-28
[GEOS-7128] Catch/Log DispatcherCallback Exceptions on fireFinishedCallback

tbarsballe:
Fix for [GEOS-7128](https://osgeo-org.atlassian.net/browse/GEOS-7128), to ensure all callbacks are handled if an exception is thrown in fireFinishedCallback. The other callbacks (init, serviceDispatched, ...) will not catch exceptions, but in this case the exceptions are caught at a higher level, and returned as an <ows:Exception> response. However, for the finished() callback, the (valid) reponse has already been written, so we cannot return an Exception response, hence this fix.
Note that this means that if an exception is thrown from any fire___Callback other than fireFinishedCallback(), subsequent callbacks of that type will not be fired, and the request will terminate with an OWS Exception response. 

I have an alternate fix that catches and logs all exceptions fired by callbacks here: https://github.com/tbarsballe/geoserver/tree/callback-all (As it stands, this fails some OWS tests that expect an error response on a failing callback)
However, I think the current fix (this PR) is probably closer to what we want, since it actually reports the failures to the user where possible. My only concern is if there are callbacks other than finished that do important configuration/cleanup (the finished callbacks will be called even if we are returning an exception response).

I've added test cases for all the callbacks, just to make sure everything behaves as expected even when an error is thrown while handling callbacks.

### DIFF
--- a/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
+++ b/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
@@ -307,7 +307,11 @@ public class Dispatcher extends AbstractController {
 
     void fireFinishedCallback(Request req) {
         for ( DispatcherCallback cb : callbacks ) {
-            cb.finished( req );
+            try {
+                cb.finished( req );
+            } catch (Exception e) {
+                logger.log(Level.WARNING, "Error firing finished callback for "+cb.getClass(), e);
+            }
         }
     }
     
--- a/src/ows/src/test/java/org/geoserver/ows/DispatcherTest.java
+++ b/src/ows/src/test/java/org/geoserver/ows/DispatcherTest.java
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2015 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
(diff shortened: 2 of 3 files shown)


src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
```diff
@@ -307,7 +307,11 @@ void flagAsSOAP(Operation op) {
 
     void fireFinishedCallback(Request req) {
         for ( DispatcherCallback cb : callbacks ) {
-            cb.finished( req );
+            try {
+                cb.finished( req );
+            } catch (Exception e) {
+                logger.log(Level.WARNING, "Error firing finished callback for "+cb.getClass(), e);
+            }
```
jodygarnett: Nice clean fix!

src/ows/src/test/java/org/geoserver/ows/DispatcherTest.java
```diff
@@ -472,6 +473,246 @@ public void setCharacterEncoding(String encoding) {
...
+        dispatcher.callbacks.add(callback1);
+        dispatcher.callbacks.add(callbackFail);
+        dispatcher.callbacks.add(callback2);
+        
+        dispatcher.handleRequest(request, response);
+        assertEquals("Hello world!", response.getOutputStreamContent());
+        assertTrue(firedCallback.get());
+        assertEquals(TestDispatcherCallback.Status.FINISHED, callback1.dispatcherStatus.get());
+        assertEquals(TestDispatcherCallback.Status.FINISHED, callback2.dispatcherStatus.get());
+    }
```
jodygarnett: And an amy of tests to make sure it works. Nice work @tbarsballe !

PR #1162, opened 2015-07-29
[GEOS-7086] Backport to 2.7.x branch

ridethepenguin:
Makes sure when namespace is deleted via REST API, also associated workspace is deleted.


jodygarnett:
If this is just a bug fix we should be able to back port straight away.

ridethepenguin:
Yes, it's just a bug fix. I'll merge also the PR with the backport to the maintenance (2.6.x) branch.

PR #1166, opened 2015-07-30
[GEOS-7132] Added docs and examples for app-schema mappings REST API

ridethepenguin:
Documentation for REST API extension introduced by [this commit](https://github.com/ridethepenguin/geoserver/commit/51583d36a85720250e67f7b647177fee7a9fa91f).

### DIFF
--- a/doc/en/user/source/data/app-schema/app-schema-resolution.rst
+++ b/doc/en/user/source/data/app-schema/app-schema-resolution.rst
@@ -53,6 +53,7 @@ Java applications such as GeoServer can load resources from the Java classpath. 
 
 The ability to load schemas from the classpath is intended to support testing, but may be useful to users whose communities supply JAR files containing their application schemas.
 
+.. _app-schema-cache:
 
 Cache
 -----
--- a/doc/en/user/source/rest/api/datastores.rst
+++ b/doc/en/user/source/rest/api/datastores.rst
@@ -117,7 +117,7 @@ The ``quietOnNotFound`` parameter avoids to log an Exception when the data store
 ``/workspaces/<ws>/datastores/<ds>/[file|url|external][.<extension>]``
 ----------------------------------------------------------------------
 
-These endpoints (``file``, ``url``, and ``external``) allow a file containing spatial data to be added (via a POST or PUT) into an existing data store, or will create a new data store if it doesn't already exist. The three endpoints are used to specify the method that is used to upload the file:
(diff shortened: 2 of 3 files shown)


bencaradocdavies:
@ridethepenguin that looks quite comprehensive. The HTML pages build cleanly.

There is a slight layout problem (in Firefox 38) with a missing background for the top part of the table of contents sidebar on the curl example page, but this is also present before your change, which merely made it bigger. Do you see this?

I am also unsure how imports are resolved for uploaded schemas in the curl example. Can you please confirm that a configuration with schemas uploaded as in this example will work if app-schema-cache is deleted and geoserver is started with no network connection other than localhost?

Otherwise, great work!

Kind regards,
Ben.

ridethepenguin:
@bencaradocdavies 

I tried Firefox 38.0 on a VM running Windows 7 32-bit
Firefox 38.0

and  Firefox 39.0 on my laptop runnig Fedora Linux 21 64-bit
Firefox 39.0

but I can see no problem with the table of contents sidebar...

I tried to clear the app-schema cache, disconnect from the Wi-Fi and start GeoServer, but it would complain about unresolved GML schemas, which are referred to in pretty much all GeoSciML schemas...
However, if I do the same thing after restoring Internet connectivity, I see the app-schema cache filling up with GML 3.1.1 schemas, but none of the GeoSciML schemas included in the ZIP archive ends up in the cache, so it seems to me local referencing is working as expected.

Should I explicitly mention that for the example to work in a completely offline environment one would have to either pre-load the GML schemas (and a few others like `cgiUtilities.xsd`) in the app-schema cache or replace all remote references to them in the GeoSciML schemas?

bencaradocdavies:
@ridethepenguin thanks for investigating the layout problem. I tested again and found that the error is caused by my local browser setting of a minimum font size, which breaks the layout. Removing this setting fixes the page and makes them as nice as your screenshots.  :-)

Your test has confirmed that schema resolution is behaving correctly.

Addition of a sentence mentioning the requirements to work in a completely offline environment (your last paragraph above) would be helpful. Some GeoServer deployments are in firewalled subnets that cannot make outgoing connections.

ridethepenguin:
@bencaradocdavies Added note about reproducing the exmaple in a completely offline environment.

bencaradocdavies:
Thanks @ridethepenguin that looks good to me. I confirmed locally that the docs build cleanly. The Travis CI failure on oraclejdk8 is an unrelated Java compilation error.

PR #1170, opened 2015-08-03
[GEOS-7133] Community modules adding gdal_translate output formats to WCS/WPS

ridethepenguin:
The new modules are very similar to the `gs-ogr-*` modules from the OGR extension, so a set of reusable superclasses / interfaces has been factored out and put in a shared module, `gs-ogr-core`.

### DIFF
--- a/src/community/gdal/gdal-wcs/src/main/java/org/geoserver/wcs/response/GdalCoverageResponseDelegate.java
+++ b/src/community/gdal/gdal-wcs/src/main/java/org/geoserver/wcs/response/GdalCoverageResponseDelegate.java
@@ -1,0 +1,435 @@
+/* (c) 2015 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.wcs.response;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.zip.ZipOutputStream;
+
+import org.geoserver.config.GeoServer;
+import org.geoserver.data.util.IOUtils;
+import org.geoserver.ogr.core.Format;
+import org.geoserver.ogr.core.FormatConverter;
+import org.geoserver.ogr.core.ToolWrapper;
+import org.geoserver.ogr.core.ToolWrapperFactory;
+import org.geoserver.platform.ServiceException;
+import org.geoserver.wcs.WCSInfo;
+import org.geoserver.wcs.responses.CoverageResponseDelegate;
+import org.geotools.coverage.grid.GridCoverage2D;
+import org.geotools.coverage.grid.io.AbstractGridFormat;
+import org.geotools.gce.geotiff.GeoTiffFormat;
+import org.geotools.gce.geotiff.GeoTiffWriteParams;
+import org.geotools.gce.geotiff.GeoTiffWriter;
+import org.geotools.util.Utilities;
+import org.opengis.parameter.GeneralParameterValue;
+import org.opengis.parameter.ParameterValueGroup;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+import org.vfny.geoserver.wcs.WcsException;
+
+/**
+ * Implementation of {@link CoverageResponseDelegate} that leverages the gdal_translate utility to encode coverages in any output format supported by the
+ * GDAL library available on the system where GeoServer is running.
+ * 
+ * <p>
+ * The encoding process involves two steps:
+ * <ol>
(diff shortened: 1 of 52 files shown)


doc/en/user/source/community/gdal.rst
```diff
@@ -0,0 +1,188 @@
...
+
+The gdal_translate utility is usually able to convert more formats than the default setup of this output format allows for, but the exact list depends on how the utility was built from sources. To get a full list of the formats available by your ogr2ogr build just run::
+
+   gdal_translate --long-usage 
+
+and you'll get the full set of options usable by the program, along with the supported formats.
+
+For example, the above produces the following output using gdal 1.11.2 compiled with libgeotiff 1.4.0, libpng 1.6,  libjpeg-turbo 1.3.1, libjasper 1.900.1 and libecwj2 3.3::
+
+   Usage: gdal_translate [--help-general] [--long-usage]
```
jodygarnett: Thanks for docs!

For this you may wish to isolate the goal_translate usage into a seperate file and ".. include:" it ... it will be easier to maintain as we update gdal version numbers.

doc/en/user/source/community/gdal.rst
```diff
@@ -0,0 +1,188 @@
...
+     PDF: Geospatial PDF
+     ZMap: ZMap Plus Grid
+
+The full list of formats that gdal_translate is able to support is available on the `GDAL site <http://www.gdal.org/formats_list.html>`_. Mind that this output format can handle only outputs that are file based and that do support creation. So, for example, you won't be able to use the PostGIS Raster output (since it's database based) or the Arc/Info Binary Grid (creation not supported).
+
+Customisation
+-------------
+
+If gdal_translate is not available in the default path, the GDAL_DATA environment variable is not set, or if the output formats needs tweaking, a ``gdal_translate.xml`` file can be put in the root of the GeoServer data directory to customize the output format.
+
```
jodygarnett: Interesting I was using Env Variable / Context Variable / System property for consistency with the rest of the application.

If we are comfortable with this approach (config file at the root of the data directory) I may change my approach.

I would be slightly more computable if we used GEOSERVER_DATA_DIRECTORY/gdal/gdal_translate.xml as a convention so we do not have modules tripping over each other.

src/community/gdal/gdal-wcs/src/main/java/org/geoserver/wcs/response/GdalCoverageResponseDelegate.java
```diff
@@ -0,0 +1,435 @@
...
+     * The environment variables to set before invoking gdal_translate
+     */
+    Map<String, String> environment = null;
+
+    /**
+     * Map holding the descriptors of the supported GDAL formats (keyed by format name).
+     */
+    static Map<String, Format> formats = new HashMap<String, Format>();
+    /**
+     * Map holding the descriptors of the supported GDAL formats (keyed by mime type).
```
jodygarnett: Is the read/write lock really earning its keep here? Guess it must be? I figured the configuration file would be read once ...

doc/en/user/source/community/gdal.rst
ridethepenguin: Good point, done.

Docs are the result of a "copy/paste + small adjustments" processing applied to the OGR extension's docs, so not much merit on my side :wink:.

doc/en/user/source/community/gdal.rst
ridethepenguin: Fine with me, I changed the config file location to `GEOSERVER_DATA_DIRECTORY/gdal/gdal_translate.xml`.

The OGR extension looks for a `ogr2ogr.xml` file in the root of the data directory, but I guess changing this behavior is not advisable, since it would break existing installations.

src/community/gdal/gdal-wcs/src/main/java/org/geoserver/wcs/response/GdalCoverageResponseDelegate.java
ridethepenguin: I'd be happy to drop the lock, but I think it's really needed, because the configuration may be updated while GeoServer is running by the [watcher](https://github.com/geoserver/geoserver/pull/1170/files#diff-512a173652411c0c49965d33fce69aaeR44) listening for changes to the `gdal_translate.xml` file.

I added a note to the docs to draw the user's attention to the fact that the configuration can be updated without a need to restart GeoServer.

jodygarnett: Thanks, yeah if you have a watcher willing to update this file then the lock is a good idea.

Still there is something odd here - if resource change is the only way to update this thing we should perhaps be using a different lock.

The watcher takes a Resource (which provides its own lock from the LockProvider which is configured under admin control - allowing this to be set up in a clustered environment).

jodygarnett:
This pull request looks good to me, anyone else care to review?

If not please merge when ready.

ridethepenguin:
I committed a [small improvement](https://github.com/geoserver/geoserver/pull/1170/files#diff-512a173652411c0c49965d33fce69aaeR122) to reduce locking overhead as much as possible when configuration is updated.

aaime:
I had reviewed Stefano's work before the pull, it looked ok to me. +1 on merging

PR #1171, opened 2015-08-04
[GEOS-7087] App-schema mapping file upload via REST

ridethepenguin:
Backport from master to 2.7.x branch.


aaime:
Merging, the build is green and more than one month passed since the master merge

PR #1174, opened 2015-08-05
[GEOS-7143] Allow WFS service settings with incomplete INSPIRE metadata

bgsmase:
See: https://osgeo-org.atlassian.net/browse/GEOS-7143


src/extension/inspire/src/main/java/org/geoserver/inspire/web/UniqueResourceIdentifiersEditor.java
```diff
@@ -152,24 +152,6 @@ protected void onError(AjaxRequestTarget target, Form<?> form) {
...
-                  ValidationError error = new ValidationError();
-                  String message = new ParamResourceModel(
-                          "noSpatialDatasetIdentifiers", UniqueResourceIdentifiersEditor.this).getString();
-                  error.setMessage(message);
-                  validatable.error(error);
-                }
-                
-            }
-            
-        });
```
jodygarnett: So we are just removing validation completely? Does it not provide any value ... 

Q: Is there any way to just provide a warning in the GUI if the reference is incomplete?

jodygarnett:
I see from the linked bug report that the plan was to ..."just to include fixed text in both the WMS and WFS web pages that says no inspire element will be produced unless the required parameters have been entered"

This pull request just removes the validation, please add the clarification (to either the web page or the geoserver docs and we are good to go).

src/extension/inspire/src/main/java/org/geoserver/inspire/web/UniqueResourceIdentifiersEditor.java
bgsmase: I guess the nicest thing would be to have one of those yellow warning triangles with a text message saying that no element will be produced when insufficient parameters have been supplied similar to the warnings about not having set a strong password etc. I'm not familiar enough with the wicket stuff to know how to do that immediately although I'm sure I could have a look at some of the other examples to get a hint. I'll have a look today.

aaime: There is no way to show a warning when pressing submit, the button either exits the page (and thus no warning will be shown) or does not.
What one would do is to add a warning on the side of the inspire components, and use ajax listeners to show/hide it every time a inspire component gets edited.

jodygarnett:
I would consider adding the * in the code, rather than in each translation - or looking how mandatory fields are handled on other pages.

We can ask @fgdrf about how to notify translators.

bgsmase:
I've now made code to add a check box to enable/disable the INSPIRE extended capabilities with validation still occuring if it is switched off. This is really abandoning the changes in this branch so I'm closing this pull request and will issue another one instead.

PR #1175, opened 2015-08-05
[GEOS-7139] Add vfs to supported protocols

tbarsballe:
Cherry-picked https://github.com/geoserver/geoserver/pull/1173 against master.

### DIFF
--- a/src/main/src/main/java/org/geoserver/util/NoExternalEntityResolver.java
+++ b/src/main/src/main/java/org/geoserver/util/NoExternalEntityResolver.java
@@ -34,8 +34,11 @@ public class NoExternalEntityResolver implements EntityResolver {
             LOGGER.finest("resolveEntity request: publicId=" + publicId + ", systemId=" + systemId);
         }
         
-        // allow schema parsing for validation (jar or external only)
-        if (systemId != null && systemId.matches("(?i)(jar:file|http)[^?#;]*\\.xsd")) {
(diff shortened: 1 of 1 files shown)


tbarsballe:
OracleJDK8 failed on "OWS request flow controller" (The other two builds passed). Seems unreleated, I think we are good to merge here.

src/main/src/main/java/org/geoserver/util/NoExternalEntityResolver.java
```diff
@@ -35,7 +35,7 @@ public InputSource resolveEntity(String publicId, String systemId) throws SAXExc
         }
         
         // allow schema parsing for validation (jar or external only)
-        if (systemId != null && systemId.matches("(?i)(jar:file|http)[^?#;]*\\.xsd")) {
+        if (systemId != null && systemId.matches("(?i)(jar:file|http|vfs)[^?#;]*\\.xsd")) {
```
jodygarnett: Q: Why not http? 
A: The wild card only checks http (so https will still match!)

The change itself is straight forward, we could probably use a comment to explain what we are letting through (since this code is fairly annoying to read/understand)

````
// http(s) - external schema reference
// jar - internal schema reference (as constructed on tomcat)
// vfs - internal schema reference (as constructed on boss)
```

tbarsballe:
I've updated the comment. Note that jar is for more than just tomcat.

PR #1176, opened 2015-08-06
Resources: closing outputstream bug, recursive delete

NielsCharlier:
According to the Closeable API (http://docs.oracle.com/javase/7/docs/api/java/io/Closeable.html#close%28%29 ) the close method should not throw (or cause, in this case) an exception when called on an already closed resource. The FileSystemResource and ResourceWrapper do do this.

Furthermore, I propose support for recursively deleting directories with their contents. This is already the case for the store.remove() function, but not resource.delete(). Note that the rule on whether to return true/false when the resource doesn't exist is different for remove and delete. Only for remove this appears to be documented.

### DIFF
--- a/src/platform/src/main/java/org/geoserver/platform/resource/FileSystemResourceStore.java
+++ b/src/platform/src/main/java/org/geoserver/platform/resource/FileSystemResourceStore.java
@@ -228,13 +228,16 @@ public class FileSystemResourceStore implements ResourceStore {
                     @Override
                     public void close() throws IOException {
                         delegate.close();
-                        Lock lock = lock();
-                        try {
-                            // no errors, overwrite the original file
-                            Files.move(temp, actualFile);
-                        }
-                        finally {
-                            lock.release();
+                        //if already closed, there should be no exception (see spec Closeable)
+                        if (temp.exists()) {
+                            Lock lock = lock();
+                            try {
+                                // no errors, overwrite the original file
+                                Files.move(temp, actualFile);
+                            }
+                            finally {
+                                lock.release();
+                            }
                         }
                     }
                 
@@ -429,7 +432,7 @@ public class FileSystemResourceStore implements ResourceStore {
 
         @Override
         public boolean delete() {
-            return file.delete();
+            return file.exists() && Files.delete(file);
         }
 
         @Override
--- a/src/platform/src/main/java/org/geoserver/platform/resource/Files.java
+++ b/src/platform/src/main/java/org/geoserver/platform/resource/Files.java
@@ -105,7 +105,10 @@ public final class Files {
                     @Override
                     public void close() throws IOException {
                         delegate.close();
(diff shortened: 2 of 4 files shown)


src/platform/src/main/java/org/geoserver/platform/resource/FileSystemResourceStore.java
```diff
@@ -228,13 +228,16 @@ public OutputStream out() {
...
                     public void close() throws IOException {
                         delegate.close();
-                        Lock lock = lock();
-                        try {
-                            // no errors, overwrite the original file
-                            Files.move(temp, actualFile);
-                        }
-                        finally {
-                            lock.release();
+                        //if already closed, there should be no exception (see spec Closeable)
```
jodygarnett: That is tricky logic, would love to know the file was written out at lest once :)  Thinking about a resource against a directory with no write access. The tmp file would be created, the move would fail, and each successive close would produce the same error ...

You can decide if you care.

src/platform/src/main/java/org/geoserver/platform/resource/FileSystemResourceStore.java
```diff
@@ -429,7 +432,7 @@ public String toString() {
 
         @Override
         public boolean delete() {
-            return file.delete();
+            return file.exists() && Files.delete(file);
         }
```
jodygarnett: Not sure about this one, the functionality was isolated into the Resources class - specifically so implementors (like you) would have less work to do when implementing a Resource.

See: That functionality was already accounted for ...

The functionality is factored out into the Resources class here:

https://github.com/geoserver/geoserver/blob/master/src/platform/src/main/java/org/geoserver/platform/resource/Resources.java#L290

If you are going to change the method contract you will need to update the Resource.delete() javadoc to explain what you are about, and remove the Resources.delete( Resource ) method as it will now be useless.

jodygarnett:
Output stream close looks fine.

The Resource.delete() change requires further discussion/work.

jodygarnett:
Revised and applied to master and 2.7.x

src/platform/src/main/java/org/geoserver/platform/resource/FileSystemResourceStore.java
NielsCharlier: I think that is a bad idea, because deleting recursively can happen a lot more efficiently in a JDBC database (with a lot less queries than would happen if you call the Resources.delete method)

Note that there is also a store.remove() function which _is_ recursive (and returns true if the resource didn't exist yet - which is also a contradiction).

jodygarnett: Your feedback is a bit late, we can take it to geotools-devel. The issue with this one is you combined a good fix (that needed to be back ported) with an API change which did not - so I had to split it up.

PR #1178, opened 2015-08-06
[GEOS-7140] WCS 1.1 now advertises workspace prefixes

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-7140
WCS 1.1 now advertises workspace prefixes
* Updated GetCapabilitiesTest to account for this
* Added test case to check that identifiers returned by DescribeCoverage match those returned by GetCapapabilites, and that two coverages with the same name in different workspaces do not cause conflicts.

### DIFF
--- a/src/wcs1_1/src/main/java/org/geoserver/wcs/response/CoveragesTransformer.java
+++ b/src/wcs1_1/src/main/java/org/geoserver/wcs/response/CoveragesTransformer.java
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2015 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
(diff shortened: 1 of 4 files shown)


src/wcs1_1/src/main/java/org/geoserver/wcs/response/CoveragesTransformer.java
```diff
@@ -124,7 +124,7 @@ void handleCoverage(CoverageInfo ci) throws Exception {
             start("wcs:Coverage");
             element("ows:Title", ci.getTitle());
             element("ows:Abstract", ci.getDescription());
-            element("ows:Identifier", ci.getName());
+            element("ows:Identifier", ci.prefixedName());
```
jodygarnett: yep this looks appropriate, I assume this is how WCS 2 0 was functioning as well.

jodygarnett:
looks good, surprised more test cases did not break.

tbarsballe:
Thanks for the review.
The only test cases that broke were the ones that examined the response for the coverage name using xpath; since WCS still accepts non-prefixed names as input, the minimial test failures actually don't surprise me.

PR #1182, opened 2015-08-11
Style Generation for all formats

tbarsballe:
* Migrated StyleGenerator to gs-main
* Added getTemplate to StyleHandler for generating styles of different formats
* Implemented getTemplate for SLD and CSS
* Added style generation to New/Edit style page

There are still some things that could be added (namely a REST endpoint for generating styles, and the ability to specify the style generation for an import), but the core change has been implemented here.

See [GSIP-131](https://github.com/geoserver/geoserver/wiki/GSIP-131)


src/main/src/main/java/org/geoserver/catalog/StyleHandler.java
```diff
@@ -84,6 +84,20 @@ public String getFileExtension() {
...
+     * Generates a template style using provided substitutionMap from StyleGenerator.
+     *
+     * @param type: the template type, see {@link org.geoserver.catalog.StyleType}.
+     * @param color: java.aw.Color to use during substitution
+     * @param colorName: Human readable color name, for use generating comments
+     * @param layerName: Layer name, for use generating comments
+     * 
+     * @return The text content of the style template after performing substitutions
+     */
+    public String getTemplate(StyleType type, Color color, String colorName, String layerName) {
```
jodygarnett: Consider renaming - now that are passing substitutions in this method produces a style definition (rather than a template).

Good idea changing from a Map<String,Object> to parameters, will force us to extend/document in the future if additional variations are allowed.

src/web/wms/src/main/resources/GeoServerApplication.properties
```diff
@@ -2,6 +2,8 @@
 #pom.xml
 
 AbstractStylePage.confirmOverwrite = Really want to overwrite the current editor contents?
+AbstractStylePage.generate         = Generate ...
+AbstractStylePage.generateNew      = Generate from template style
```
jodygarnett: Would describe this as "Generate default style" (since from the users perspective they cannot supply templates anywhere).

src/web/wms/src/main/java/org/geoserver/wms/web/publish/StyleTypeModel.java
```diff
@@ -0,0 +1,31 @@
+/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
```
jodygarnett: 2015

src/web/wms/src/main/java/org/geoserver/wms/web/publish/StyleTypeChoiceRenderer.java
```diff
@@ -0,0 +1,35 @@
...
+public class StyleTypeChoiceRenderer implements IChoiceRenderer {
+
+    public Object getDisplayValue(Object object) {
+        return getIdValue(object, 0);
+    }
+
+    public String getIdValue(Object object, int index) {
+        switch ((StyleType) object) {
+            case GENERIC:
+                return "Generic";
```
jodygarnett: Consider moving to the Enum.

tbarsballe: It is a unique wicket id. However, I also need a display value (for which I use the same method, see `getDisplayValue` above; given that the type names are unique, this makes more sense than duplicating code). Now that the enum is its own class, moving this to the enum is definately a good idea.

tbarsballe:
Thanks for the review. I have updated with suggested changes (and fixed a few more missing headers).

jodygarnett:
Nifty, now all we need is a few more votes on the proposal.

src/extension/css/src/main/java/org/geoserver/community/css/web/CssHandler.java
```diff
@@ -5,6 +5,7 @@
  */
```
aaime: Header copyrigth was not updated. Kevin shared a nice script to do that on all files in a commit, in one shot, please do so :)

aaime:
The documentation for the style page has not been updated. It we help us  see what it does (don't have time to interactively try it...)

tbarsballe:
Added documentation the webadmin style page (it was very out of date, so I have updated it to be inline with current behavior, including style generation).

I also had time to add a style generator REST Endpoint yesterday (with basic documentation).

Both these changes have been added as their own commit so it is easier to see the individual changes. Once this is ready for merge, I can squash the commits together (allthough each commit is its own self contained change, so not sure this is necessary, or even a good idea?)

src/extension/css/src/main/java/org/geoserver/community/css/web/CssHandler.java
tbarsballe: Fixed.

doc/en/user/source/rest/api/stylegenerator.rst
```diff
@@ -0,0 +1,125 @@
+.. _rest_api_stylegenerator:
+
+Style Generator
+===============
+
+The Style Generator endpoint is used for generating default styles. Styles generated using this endpoint may be added to geoserver using the :ref:`styles <rest_api_styles>` endpoint.
+
+``/stylegenerator[.<format>]``
+------------------------------
```
jodygarnett: Interesting, I may of done this as an endpoint on the styles REST API endpoint.

Something like PUT /workspaces/<ws>/styles/<s>[.<format>] would complete the table?

tbarsballe: For integrating with the styles endpoint, I was mostly concerned about
colliding with the existing endpoints.

I do like your PUT idea however Jody, it integrates much better with the
existing API - in this case I guess the template to be used would be passed
as a query paramete (defaulting to generic)?

Torben

jodygarnett: Should be able to determine from the `<format>` included in the PUT url.

I actually prefer this approach, creating a REST style generator is confusing as it does not actual modify the geoserver configuration (so does not really belong with the other REST API endpoints).

tbarsballe: This (and the existing style api) support both methods - I believe the extension is checked first, then the PUT URL.

aaime:
The style page is sure gotten long... we should probably look into some way to use the horizontal space available too (e.g., two columns), but I guess this is going to be done another day.

About the commits, that's fine to keep them as is.

doc/en/user/source/rest/api/stylegenerator.rst
tbarsballe: @jodygarnett - I looked at the code / documentation and PUT is already implemented - see [/styles/\<s\>\[.\<format\>\]](http://docs.geoserver.org/stable/en/user/rest/api/styles.html#styles-s-format). I am not sure why you thought it wasn't implemented? Wherever we know the style name, PUT is used, otherwise POST is used, in accordance with REST conventions.

jodygarnett: I guess I was just miss understanding of our confusing REST docs :)

tbarsballe: The best idea might be to add `generate` and `type` parameters to the styles API, such that:
* POST `/styles?name=newstyle&generate=true&type=point` creates a new style named "newstyle", generated using the point template
* PUT `/styles/existingstyle?generate=true&type=point` replaces the existing style with a generated style.

I do have a couple concerns about this approach however:
* Unlike other POST/PUT requests against the styles API, these would not expect a form upload. Is this likely to cause issues? (I am not really familiar enough with how our REST API works to determine if this could be a problem)
* This is extending an existing endpoint with functionality that can easily wipe out an the style; seems a bit risky (althoug a blank form upload would probably do the same thing, so maybe not an issue).

@aaime Does this method for exposing the style generator using REST seem better then my current implementation that adds a new GET endpoint? Do you think an empty POST/PUT will cause problems?

jodygarnett: I like it - similar to recalculating a layer bounds.

I would recommend:

* ``/styles/foo.sld?generate=point`` to generate an point SLD style
* ``/styles/bar.css?generate=line`` to generate an line string css style

And then PUT/POST can just be used based on if a new style is being created or not.

tbarsballe: If you take a look at the existing documentation, when creating a new style, we do a POST to `/styles&name=foo` (for example). From what I can determine, this is done to maintain RESTful behaviour, as it may be the case that the style foo already extists. In order to be consistent with the existing behavior, we should stick to this convention, as demonstrated in my example above.

jodygarnett: Sounds good - thanks for taking care with this one.

jodygarnett:
Okay would like to see this one merged.

tbarsballe:
I agree this should be merged in time for 2.8. The REST API changes haven't been done yet as I was rather busy with other stuff last week and am on vacation this week - maybe just merge without the REST commit (https://github.com/tbarsballe/geoserver/commit/9c9a0aca00183496013d400c10caf18d5759fc02), and I can add the API change later (it will basically judt be adding a few options, so not really an API-breaking change)?

jodygarnett:
Good thinking, will work on merging now. Please enjoy your vacation!

PR #1188, opened 2015-08-17
Extra control over WPS execution limits

julian1:
Changes the wps execution limit to apply only to job execution time. Existing behavior that includes both queue and execution time is maintained using the totalTime attributes.

Depends on https://github.com/geotools/geotools/pull/942


aaime:
If I'm interpreting the Travis logs correctly, the change makes the WPS test never terminate.

julian1:
@aaime Yes, the test fails without the Geotools event fix. 

I'm hoping Travis will pull the latest Geotools.  I'll close and reopen the PR to trigger it again.

aaime:
@julian1 the fix in GeoTools seems to be causing the same problem in the normal GeoServer build (without this pull request applied).

aaime:
Hopefully fixed the build. See my comment on the other pull request on what needs to be done before this one can be merged. 
I could not read the code yet, but I have a worry looking at the UI changes, will the change be backwards compatible? As in, what happens if I upgrade from a data directory that do not have the new settings? And will the old settings still have the same meaning?

aaime:
Keep in mind the 28th is feature freeze, to be in 2.7.x this one should be reviewed by the 27th, which basically leaves you just the 26th to do this.

jonescc:
@aaime, @julian1 has moved onto other things and has asked me to address oustanding issues with this PR.

I've fixed merge conflicts introduced by recent activity in master and an issue with start event delegation that was causing synchronous requests to fail.

I've read through the other pull requests and I don't see anything else that hasn't been done?  Copyright headers seem OK.

With reference to the UI changes, yes, @julian1 has changed the existing maximum execution time for synchronous and asynchronous requests to exclude queuing time and has added new total time limits which behave as per the original execution time limits.   When upgrading to this version,  the previous execution time limit is used for the new execution time limit and the total time limit will be set to zero, so yes there will be a behaviour change on upgrade without manual intervention, queuing time will no longer be included when testing against the specified execution time limit and queuing time will not be limited without changing the total time allowed from the defaulted zero value.

Happy to make changes here if you can see a better way of doing this.

aaime:
Hi, had another review, the feedback is similar to the one before: people upgrading should not perform any manual intervention on their configuration, the upgrade must be transparent, what worked before (for them) should keep on working unmodified (going by memory, we never had instructions to manually update the config, not even jumping from 1.7.x to 2.x when the entire set of config files got a relayout).
I admit this is one case where preserving the backwards compatibility is not easy.

Here is a possible idea, make the new total time a Integer in WPSInfoImpl, if the field is null, assume it's equal to the execution time (e.g., in the getter, do a null check and return the execution time instead). This preserves basically preserves the existing behavior, those that were happy with it before will be after, those unhappy now have a config to tune in order to get a different behavior.
The code as is would allow a non acceptable behavior instead, max time = 0 for those upgrading, meaning the jobs could stay in queue forever (which might be ok for asynch ones, but not for synch ones, which depend on the HTTP connection to be still alive in order to return a result).

The two new settings also imply some dependency, the total time should be greater than the execution time (unless one of the two is zero of course). How about adding a validation in the UI, and emit a warning in the listener if the two are still not consistent (people can make them inconsistent by directly modifying the xml files)?

The existing conflicts should be solved, hopefully it's just due to the wicket 7 upgrade, if that's the case it should be easy to solve.

Final note, as a user facing change the documentation should be updated with the new settings, in particular, this page: http://docs.geoserver.org/latest/en/user/services/wps/administration.html#execution-and-resource-management-options

pblain:
Thanks for reviewing and providing the suggestions @aaime - @jonescc will probably be the one to look into this when he returns next week.  Hopefully we will be able to update the PR accordingly.

jonescc:
@aaime, I have now made the suggested changes. Behaviour will be
preserved when upgrading, total time limits will be required to be
greater than the execution time limits if specified, documentation has
been added and conflicts resolved.  I have also added some additional
test cases for upgrading to the new settings and application of the
queuing and execution time limits. Thanks

aaime:
Verified, looks good, merging

aaime:
The commit message should have referred to a ticket in Jira, but I did not want to add yet another review round-trip to this one. Yet, a ticket is needed or the change won't show up in the release notes of 2.10.
Yet I cannot find one... is it there, or should I create it?

aaime:
The merge broke the build: http://ares.boundlessgeo.com/jenkins/job/geoserver-master/2622/consoleFull
Checked the test, it was vulnerable to cpu scheduling issues, modified it so that it should be less so, build passes now

jonescc:
Apologies for the broken test.  I can't find a Jira issue either.  I've created https://osgeo-org.atlassian.net/browse/GEOS-7649 .  I hope that's suitable.  I assume that you will need to update the commit now.

aaime:
Hi, it's suitable. The commit cannot be updated now that it's in the main repo, but it not the end of the world.

PR #1189, opened 2015-08-18
Option to disable synchronous WPS

julian1:


aaime:
Hi Julian,
there are two basic problems with your pull request: 
* You are contributing more than one (non test) file, but you have not signed and sent the OSGeo contribution agreement as requested in our contribution guidelines, see https://github.com/geoserver/geoserver/blob/master/CONTRIBUTING.md
* You are providing two new features, but you have not discussed them in advance in the GeoServer developer list (you should always get feedback on what you want to do _before_ writing code, or you risk to have to redo part of it, or in the worst cases, to throw it away)

Can you please comply with both?

aaime:
Btw, as the WPS maintainer, my immediate feedback on this one is that to line up with the current WPS architecture, you should have applied the synch flag on a per process basis, some processs make sense to run only in asynch mode, others do not.
Also, one should warn very explictly the admins that if they use that flag, they will break WPS compliance, as WPS asynch support is optional and could be disabled, but synch support is never optional, and disabling it will break standard compliant clients.

aaime:
Btw, the 28th we have the feature freeze, if we don't merge this one before it, it will become a GeoServer 2.9.x feature. Which is fine by me, just wanted to check you're aware and ok with it.

aaime:
Hum, no changes in a month. Are you still around?

julian1:
@aaime My apologies for the delay - the wps work has not been prioritized in our planning.  

I emailed a signed contributor agreement this morning. Follow-up discussion is provided in the thread - "[Geoserver-devel] Need advice on managing resources for WPS in Geoserver".

julian1:
closed after mailing list discussion

PR #1190, opened 2015-08-20
Fix WFS ResourceAccessManager tests

julian1:
The http://www.openplans.org/topp schema resource appears to have been recently deleted. 

Unit tests fail (stall) when getAsDOM() is called and the resource can't be retrieved.

I experience the same issues on the release branch (2.7.1).

### DIFF
--- a/src/wfs/src/test/java/org/geoserver/wfs/ResourceAccessManagerWFSTest.java
+++ b/src/wfs/src/test/java/org/geoserver/wfs/ResourceAccessManagerWFSTest.java
@@ -41,7 +41,7 @@ public class ResourceAccessManagerWFSTest extends WFSTestSupport {
     static final String INSERT_RESTRICTED_STREET = "<wfs:Transaction service=\"WFS\" version=\"1.0.0\"\n"
             + "  xmlns:wfs=\"http://www.opengis.net/wfs\" xmlns:cite=\"http://www.opengis.net/cite\"\n"
             + "  xmlns:gml=\"http://www.opengis.net/gml\" xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\"\n"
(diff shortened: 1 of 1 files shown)


aaime:
Hi @julian1, after fixing all the build failiure caused by your sub-progress-listener changes in GeoTools all the build servers are completing the build correctly (as does my local computer)
Wondering if the issue is happening only under certain conditions?

Also, your "fix" is making no sense, you are replacing a syntactically correct schemalocation attribute with  an invalid one. The correct syntax is a sequence of "namespace location namespace location ..."
whilst with your fix we have "namespace location location".

The namespace you have removed is also largely used in the release data directory, so everybody trying GeoServer is using it... it would be a bit strange that no one complained yet.

julian1:
@aaime If it doesn't affect anyone else then it's clearly an issue with my local setup. 

Testing - by deleting the m2 repo and freshly compiling geotools and geoserver on master and release branches both stall at the getAsDom() call for me. If the test mocks a real http service, then maybe there is a port/network issue in my Docker environment. 

Thanks for your comment, and I'll close the PR.

aaime:
Interesting... can you grab a full stack trace using jstack, to see where exactly it's getting blocked?

julian1:
@aaime 

If there's a pre-existing service listening on port 8080, then the behavior of org.geoserver.wfs.ResourceAccessManagerWFSTest depends on that service, and not the mocked http service.

For example, starting netcat and running maven, shows that nc receives the socket traffic, and the test hangs when there's no response.
```
# while true ; do nc -l -p 8080; done
GET /geoserver/wfs/DescribeFeatureType?typename=topp:tasmania_roads HTTP/1.1
User-Agent: Java/1.7.0_80
Host: localhost:8080
Accept: text/html, image/gif, image/jpeg, *; q=.2, */*; q=.2
Connection: keep-alive
```

It would almost certainly be better if the http mocking setup failed cleanly if it cannot exclusively bind the port.

aaime:
The mocked tests does not bind a  port at all, the HTTP for the OGC requests is mocked too... but I'm wondering, it might be the XML processing machinery that tries to get stuff out of localhost:8080.
What requests do you see?

PR #1193, opened 2015-08-21
update link to rgeoserver gem

eliotjordan:
Active development of the rgeoserver gem referenced in the documentation has moved to a new repository and is now maintained by @drh-stanford. Rubygems points to this repo as well. The examples provided on the old gem repo are broken.

### DIFF
--- a/doc/en/user/source/rest/examples/ruby.rst
+++ b/doc/en/user/source/rest/examples/ruby.rst
@@ -3,7 +3,7 @@
 Ruby
 ====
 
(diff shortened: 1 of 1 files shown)


aaime:
Hi, had a quick look, seems ok, just one question: how comes the latest gem seems to be 0.7.10 (https://github.com/sul-dlss/rgeoserver/releases) whilst the github repository contains only up to 0.7.9?

The two releases seem to be only 4 days apart, which is also weird.

eliotjordan:
Hi! Thanks for the quick reply. It looks like the 0.7.10 version is in the relax-rest-client-dep branch that was never merged into master. I created an issue for this: sul-dlss/rgeoserver/issues/7.

PR #1194, opened 2015-08-22
[GEOS-7167] GML2 xlinks import diverge fixed

koordo:
Now both feature.xsd and geometry.xsd imports the same xlinks.xsd and the XSD validates correctly.


aaime:
Well, let's merge and see if we still pass the CITE tests tomorrow :-p

PR #1195, opened 2015-08-24
Accepts check box clicks outside the boxes [GEOS-6948]

jodygarnett:
We have had an issue with label spanning preventing the of checkboxes. This issue has been held open due to two checkboxes in WMSLayerConfig - which provide the same behavior due to correctly marking a label as for an assoicated checkbox.

This pull request removes the assoication until such time as the CSS can be fixed.

For details: https://osgeo-org.atlassian.net/browse/GEOS-6948


smithkm:
I think I see how to fix this without breaking the `for` relationship between label and input.

PR #1196, opened 2015-08-25
Accepts check box clicks outside the boxes [GEOS-6948]

smithkm:
Applied a class to the list rather than to the label allowing for manipulation of the CSS higher up the tree.

Retains the pseudo-hanging indent so long labels won't wrap underneath the checkbox while also making the label an inline element so it doesn't result in clicking on the whitespace toggling the input.


jodygarnett:
Replacement for #1195

jodygarnett:
Thanks @smithkm this does fix GEOS-6948.

For consistency we can hunt down all the other checkboxes at a later date [GEOS-7172](https://osgeo-org.atlassian.net/browse/GEOS-7172)

PR #1197, opened 2015-08-26
resolve conflict between printing and monitor extension

jodygarnett:
From Javier Avalos on mailing list:

> I answered myself : 
> adding the following line:
> /pdf/ **
> to the filter.properties configuration file of the monitoring module and restart geoserver the error disappears .
> Cheers!


aaime:
This really looks like a band-aid, but I guess it's better than nothing... +1 :-p

PR #1201, opened 2015-08-28
[GEOS-7177]: Support system property to enable JAIExt operations

dromagnoli:


tbarsballe:
This change broke tests on systems where native JAI is not supported (eg. OS X). Fix here: https://github.com/geoserver/geoserver/pull/1206

PR #1204, opened 2015-08-31
[GEOS-7151] 2.7.x Backport patch to fix CloseableIterator warnings

sikeoka:
Use the factory method instead of the constructor to create the CloseableIteratorAdapter to ensure that the wrapped CloseableIterator is closed properly.


jodygarnett:
Can this fix be applied to 2.6.x as well?

PR #1205, opened 2015-08-31
[GEOS-7085] 2.7.x Backport patch to add the proxy base URL to WCS/WFS metadata links

sikeoka:
[GEOS-7085] Add the proxy base URL to metadata links, if necessary, in WCS DescribeCoverage and WCS/WFS GetCapabilities responses.


jodygarnett:
Can/should this be back ported to the 2.6.x series?

PR #1206, opened 2015-08-31
[GEOS-7177] Fix JAIPageTest for systems without native JAI

tbarsballe:
https://github.com/geoserver/geoserver/pull/1201 changed the JAI page to only show JAI options if native JAI is enabled. This breaks the JAIPageTest for systems where narive JAI is not enabled or not supported.

This PR fixes the test case to account for this.
Tested on OS X (No native JAI). Since the parts that require JAI are unchanged, I assume they still work.


src/web/core/src/test/java/org/geoserver/web/admin/JAIPageTest.java
```diff
@@ -22,6 +22,7 @@
 import org.geoserver.config.JAIEXTInfo;
```
jodygarnett: Go on update the header you know you want to...

jodygarnett:
So other than header this looks good, want to correct and merge?

PR #1207, opened 2015-08-31
Fixes so that GeoServer can be packaged for OS X

dmkelsey:


build/mac_installer.sh
```diff
@@ -27,98 +27,22 @@ fi
...
-     end tell
-   end tell
-' | osascript
-
-# convert to compressed image, delete temp image
-hdiutil convert "${dmg_tmp}" -format UDZO -imagekey zlib-level=9 -o "${dmg_final}"
-if [ -f "${dmg_tmp}" ]; then
-  rm -f "${dmg_tmp}"
-fi
+pushd tmp/geoserver-$tag-mac/ > /dev/null
```
jodygarnett: It feels like you are killing tones of stuff here?

jodygarnett:
As per the CONTRIBUTING.md file we ask that you have an individual contributors agreement or a corporate contributors agreement to provide changes to OSGeo. In this case you are covered by Boundless.

jodygarnett:
Normally I would ask you to squash this to a single commit (for back porting) - do you want to give it a go?

Since we always use the build scripts from "master" it is not strictly necessary, just neater.

jodygarnett:
Unless @aaime corrects me we do no usually create Jira issues for build issues (or if we do we just mark them closed and do not provided a fixedVersion - preventing them appearing in release notes).

build/mac_installer.sh
dmkelsey: The `./build.sh` script, which lives in the `geoserver-$tag-mac` director duplicates everything I removed..  The difference is that `./build.sh` works correctly.

jodygarnett:
Thanks mutely

PR #1211, opened 2015-09-06
[GEOS-7191] Update POMs that point to old Codehaus Maven repositories.

bradh:
Compile tested the Excel and Charts extensions. Compile tested the spatialite
community module.

The sync community module partly builds but fails some tests in geoserver-sync-server,
which appears unrelated to this change.

### DIFF
--- a/src/community/spatialite/pom.xml
+++ b/src/community/spatialite/pom.xml
@@ -21,13 +21,13 @@
 
 
 <repositories>
-   <repository>
-     <id>codehaus</id>
-     <name>codehaus</name>
-     <url>http://snapshots.repository.codehaus.org/</url>
-     <snapshots>
-      <enabled>true</enabled>
-     </snapshots>
+    <repository>
+        <id>boundless</id>
+        <name>Boundless Maven Repository</name>
+        <url>http://repo.boundlessgeo.com/main</url>
+        <snapshots>
+            <enabled>true</enabled>
+        </snapshots>
     </repository>
     <repository>
       <id>opengeo</id>
--- a/src/extension/charts/pom.xml
+++ b/src/extension/charts/pom.xml
@@ -14,9 +14,9 @@
 
   <repositories>
     <repository>
-      <id>codehaus</id>
-      <name>codehaus</name>
-      <url>http://snapshots.repository.codehaus.org/</url>
+      <id>osgeo</id>
+      <name>Open Source Geospatial Foundation Repository</name>
+      <url>http://download.osgeo.org/webdav/geotools/</url>
       <snapshots>
         <enabled>true</enabled>
       </snapshots>
--- a/src/maven/archetype/wfsOutputFormat/src/main/resources/archetype-resources/pom.xml
+++ b/src/maven/archetype/wfsOutputFormat/src/main/resources/archetype-resources/pom.xml
@@ -7,14 +7,6 @@
  <name>My Output Format</name>
 
   <repositories>
-   <repository>
-     <id>codehaus</id>
-     <name>codehaus</name>
-     <url>http://snapshots.repository.codehaus.org/</url>
-     <snapshots>
-      <enabled>true</enabled>
-     </snapshots>
-    </repository>
     <repository>
       <id>opengeo</id>
       <name>opengeo</name>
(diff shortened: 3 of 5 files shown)


fgdrf:
Good point to remove codehaus repositories. I'm wondering a bit about the different places repositories are defined. Boundless repository is defined in root-pom already. Maybe it makes sense to remove all repository definitions in all submodules consequently to avoid duplications.

Are there repository definitions that is only required by specific module? Or with other words: Would/Could others use dependencies from theses too?

```
 <repositories>
  <repository>
   <id>boundless</id>
   <name>Boundless Maven Repository</name>
   <url>http://repo.boundlessgeo.com/main</url>
   <snapshots>
     <enabled>true</enabled>
   </snapshots>
  </repository>
```

bradh:
It should be possible to build these modules without building from the root POM. That is, we shouldn't need to rely on the root POM.

fgdrf:
@bradh Well, my understanding of maven is : Because of the multi-module structure of the build where parents are defined correctly a specific module inherits configurations from parent/root pom. That means every module can be build independently. If the parent cannot be found relative to the project the dependency-resolutions tries to find parent in local/remote Maven repository. Since GeoServer is a Multi-module maven-project it works quite well without having definitions on every level / module again.
Therefore I would remove all repository definitions from archetype pom and since the module stucture for spatialite, excel, geoserver-sync, and chart are correct, I'd prefere to remove repository definitions from these too.
```
root
  + extension
      + charts
      + excel
  + community
      + geoserver-sync
      + spatialite
```

bradh:
I'd prefer to do that separately, and make it common across all modules. I think that is a more invasive change, and since it is late in 2.8.x release, perhaps a little too much for now. I just want to get off codehaus before they stop redirection and start server invalid JARs with status 200: http://www.codehaus.org/mechanics/maven/

For the archetype POM, I think it should stay in any case. That would support out-of-tree builds (e.g. of a new module).

jodygarnett:
I am happy to remove any stray references to codehaus - as volunteers we don't tend to look out for community models until they are past their doc/qa/ip checks and graduate to an extension.

I think our overall plan is to manage repositories at the top-level, just like we do with a dependency management section. Part of this is so that ares can be better managed (and hopefully limited to repo.boundlessgeo.com during nightly builds).

Can I merge this codehaus change and ask you two carry on your discussion on geoserver-devel?

src/community/spatialite/pom.xml
```diff
@@ -21,13 +21,13 @@
...
-     <snapshots>
-      <enabled>true</enabled>
-     </snapshots>
+    <repository>
+        <id>boundless</id>
+        <name>Boundless Maven Repository</name>
+        <url>http://repo.boundlessgeo.com/main</url>
+        <snapshots>
+            <enabled>true</enabled>
+        </snapshots>
```
jodygarnett: This should already be present in the root pom.xml

src/extension/charts/pom.xml
```diff
@@ -14,9 +14,9 @@
 
   <repositories>
     <repository>
-      <id>codehaus</id>
-      <name>codehaus</name>
-      <url>http://snapshots.repository.codehaus.org/</url>
+      <id>osgeo</id>
+      <name>Open Source Geospatial Foundation Repository</name>
+      <url>http://download.osgeo.org/webdav/geotools/</url>
```
jodygarnett: We use OSGeo repository for stable versions, and boundless repo for snapshots.

src/maven/archetype/wfsOutputFormat/src/main/resources/archetype-resources/pom.xml
```diff
@@ -23,6 +15,14 @@
...
        </snapshots>
     </repository>
+    <repository>
+      <id>osgeo</id>
+      <name>Open Source Geospatial Foundation Repository</name>
+      <url>http://download.osgeo.org/webdav/geotools/</url>
+      <snapshots>
+        <enabled>true</enabled>
+      </snapshots>
+    </repository>
```
jodygarnett: we use osgeo repository for releases, and boundless repo for snapshots.

bradh:
@jodygarnett So the change is OK for you if I get rid of the "snapshot" parts on download.osgeo.org?

Is the removal of repo.boundlessgeo.com from src/community/spatialite/pom.xml required?

smithkm:
I've removed the unnecessary repos that were being added, did a local test build and merged it manually.

PR #1212, opened 2015-09-06
Update links to point to new JIRA instance

bradh:
Split the patch into multiple parts. The only one that actually modifies code is  "Fix bug tracker links". The rest modify data or documentation / comments.

See https://osgeo-org.atlassian.net/browse/GEOS-7192


fgdrf:
Patch looks good. Is there a plan to backport to 2.7.x and 2.6.x as well?

bradh:
I wasn't planning to backport this. 2.7.x might become a candidate after the required "stability" time, but 2.6.x will be EOL by that point.

PR #1217, opened 2015-09-14
GEOS-7189: Fix deserialization of boolean metadata parameter.

bgsmase:
https://osgeo-org.atlassian.net/browse/GEOS-7189


bgsmase:
Possibly ignore this and use https://github.com/geoserver/geoserver/pull/1218 instead if version branched from master is preferred way round to do it.

PR #1219, opened 2015-09-15
GEOS-7116, Applying a CRS in Importer clears other found CRSes

smithkm:
Added unit tests for ImportTaskTable, fixed CRSPanel by forcing it to update the model when the find dialog closes to match the behaviour when manually updating.


aaime:
Merged on master... not sure if to backport to 2.8.x, I had no time to manually test this change in the other places using the CRSPanel, like the layer page (and the test coverage for the UI tends to be rather light). Anyone has time to give it a kick?

simboss:
Ideally Andrea this should go back to 2.7.
Let's talk about it tomorrow...

PR #1232, opened 2015-09-25
[GEOS-7214] Add scaling support to WPS gs:Download process

ridethepenguin:


src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/DownloadUtilities.java
```diff
@@ -50,6 +52,10 @@
     /** The LOGGER. */
     private static final Logger LOGGER = Logging.getLogger(DownloadUtilities.class);
 
+    /** Nearest Neighbour interpolation method */
```
simboss: We don't need this, there is already an instance available in [ImageUtilities](https://github.com/geotools/geotools/blob/master/modules/library/coverage/src/main/java/org/geotools/resources/image/ImageUtilities.java#L183):

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/ScaleToTarget.java
```diff
@@ -0,0 +1,416 @@
...
+     * <p>
+     * If one of the two inputs is omitted, the missing value is inferred from the provided one so that the aspect ratio of the specified envelope is
+     * preserved.
+     * </p>
+     *
+     * @param targetSizeX the size of the target image along the X axis
+     * @param targetSizeY the size of the target image along the Y axis
+     * @throws TransformException
+     */
+    public void setTargetSize(Integer targetSizeX, Integer targetSizeY) throws TransformException {
```
simboss: I am not a fun of using Integer VS int, nice style but inefficient.

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/RasterEstimator.java
```diff
@@ -126,25 +128,25 @@ public boolean execute(final ProgressListener progressListener, CoverageInfo cov
...
+            ReferencedEnvelope refEnvelope = JTS.toEnvelope(roiInNativeCRS_.getEnvelope());
+            scaling = new ScaleToTarget(reader, refEnvelope);
 
+            // TODO investigate on improved precision taking into account tiling on raster geometry
         } else {
             // No ROI, we are trying to read the entire coverage
-            final Rectangle2D originalGridRange = (GridEnvelope2D) reader.getOriginalGridRange();
-            areaRead = originalGridRange.getWidth() * originalGridRange.getHeight();
-
+            scaling = new ScaleToTarget(reader);
```
simboss: You are doing an actual read and scale to estimate the size of the final coverage. If by any chance we are not doing a deferred load this might take forever.

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/ScaleToTarget.java
```diff
@@ -0,0 +1,416 @@
...
+import org.opengis.referencing.operation.MathTransform;
+import org.opengis.referencing.operation.TransformException;
+
+/**
+ * Class encapsulating the logic to scale a coverage to a pre-defined target size.
+ *
+ * @author Stefano Costa, GeoSolutions
+ *
+ */
+class ScaleToTarget {
```
simboss: It could be interesting to share this code between WPS and WCS 2.0 (since the latter is a core module WPS can lean on it) otherwise we would be duplicating things throughout the code.

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/DownloadEstimatorProcess.java
```diff
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2015 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
```
dromagnoli: Minor note: I think you should extend the time period by setting 2014 - 2015

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/DownloadProcess.java
```diff
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2015 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
```
dromagnoli: Same as before

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/DownloadUtilities.java
```diff
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2015 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
```
dromagnoli: Another one :)

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/RasterEstimator.java
```diff
@@ -1,24 +1,22 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2015 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
```
dromagnoli: Date in header

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/DownloadUtilities.java
ridethepenguin: Fixed.

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/ScaleToTarget.java
ridethepenguin: Agreed, but this method will not be called often and using `Integer` allows to deal with missing values more easily.

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/ScaleToTarget.java
ridethepenguin: @simboss  Definitely. However, in order to have this working ASAP, we could leave the refactoring to a second PR. That would also allow us to give it the necessary attention.

ridethepenguin:
I've addressed feedback and fixed the copyright headers. We also agreed to postpone the WCS refactoring.

Travis build is green: I'm going to merge this soon.

PR #1234, opened 2015-09-27
WPS JDBC Status Store

ianturton:
An new community module allowing for the long term storage of process statuses in (clustered) WPS systems.


src/community/pom.xml
```diff
@@ -64,8 +64,9 @@
           <descriptor>release/ext-gwc-distributed.xml</descriptor>  
           <descriptor>release/ext-geofence-server.xml</descriptor>
           <descriptor>release/ext-gwc-s3.xml</descriptor>	
-	  <descriptor>release/ext-gdal-wcs.xml</descriptor>
-	  <descriptor>release/ext-gdal-wps.xml</descriptor>
+	  	  <descriptor>release/ext-gdal-wcs.xml</descriptor>
+	      <descriptor>release/ext-gdal-wps.xml</descriptor>
+	      <descriptor>release/ext-wps-jdbc.xml</descriptor>
```
aaime: What happened to indentation here? I can see the original was already using tabs, which is not nice, but it looks like it got worse or something

src/community/wps-jdbc/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+	<parent>
+		<groupId>org.geoserver</groupId>
+		<artifactId>community</artifactId>
+		<version>2.8-SNAPSHOT</version>
+	</parent>
+	<artifactId>gs-wps-jdbc</artifactId>
+	<name>WPS cluster on jdbc datastores</name>
+	<dependencies>
+		<dependency>
+			<groupId>org.geoserver.extension</groupId>
```
aaime: tabs...

src/community/wps-jdbc/src/main/java/org/geoserver/wps/jdbc/JDBCStatusStore.java
```diff
@@ -0,0 +1,493 @@
...
+                            + PROPERTIES+":String," 
+                            + SIMPLE_PROCESS_NAME+":String,"
+                            + TASK+":String," 
+                            + USER_NAME+":String," 
+                            + ASYNC+":Boolean,"
+                            + EXCEPTION_CLASS+":String,"
+                            + EXCEPTION_MESSAGE+":String," 
+                            + STACK_TRACE+":String");
+                } catch (SchemaException e) {
+                    // TODO Auto-generated catch block
```
aaime: I guess this was meant to be replaced with something more useful. An actual log, or rethrowing the exception maybe? Wondering how this will work if the loader is not able to build a store, like, a misconfigured/lack of configuration setup? Does GeoServer at least start? (it's ok if any attempt at using aynch WPS does not work until the store is configured)

aaime: Ah, I see this one should actually never throw, sorry about that. But my comment is still valid for the catch below.

src/community/wps-jdbc/src/main/java/org/geoserver/wps/jdbc/JDBCStatusStore.java
```diff
@@ -0,0 +1,493 @@
...
+                    e.printStackTrace();
+                }
+              //@formatter:on
+                statuses.createSchema(schema);
+            }
+        } catch (IOException e1) {
+            // TODO Auto-generated catch block
+            e1.printStackTrace();
```
aaime: This one :-)

src/community/wps-jdbc/src/main/java/org/geoserver/wps/jdbc/JDBCStatusStore.java
```diff
@@ -0,0 +1,493 @@
...
+                // if the feature exists delete it
+                Filter filter = ECQL.toFilter(
+                        PROCESS_ID + " = '" + status.getExecutionId() + "'");
+                store.removeFeatures(filter);
+                store.addFeatures(featureCollection);
+                transaction.commit();
+            }
+        } catch (IOException e) {
+            // TODO Auto-generated catch block
+            e.printStackTrace();
```
aaime: Really, non test code should never use these, use the logger instead ;-)

src/community/wps-jdbc/src/main/java/org/geoserver/wps/jdbc/JDBCStatusStore.java
```diff
@@ -0,0 +1,493 @@
...
+        }
+
+    }
+
+    @Override
+    public void save(ExecutionStatus status) {
+        DefaultTransaction transaction = new DefaultTransaction("create");
+        try {
+            SimpleFeatureSource source = statuses.getFeatureSource(STATUS);
+            if (source instanceof SimpleFeatureStore) {
```
aaime: What if by any accident the code does not return a SimpleFeatureStore? We should inform the user somehow

src/community/wps-jdbc/src/main/java/org/geoserver/wps/jdbc/JDBCStatusStore.java
```diff
@@ -0,0 +1,493 @@
...
+            SimpleFeatureSource source = statuses.getFeatureSource(STATUS);
+            SimpleFeatureCollection features = source.getFeatures(filter);
+            ret = features.size();
+            if (ret == 0) {
+                return ret;
+            }
+            SimpleFeatureIterator itr = features.features();
+
+            itr.close();
+            if (source instanceof SimpleFeatureStore) {
```
aaime: Seems like quite inefficient, why not call directly removeFeatures(Filter)?

src/community/wps-jdbc/src/main/java/org/geoserver/wps/jdbc/JDBCStatusStore.java
```diff
@@ -0,0 +1,493 @@
...
+        } catch (InstantiationException | IllegalAccessException | ClassNotFoundException e) {
+            // too bad, I don't care
+        } catch (NoSuchMethodException e) {
+            // too bad, I don't care
+        } catch (SecurityException e) {
+            // too bad, I don't care
+        } catch (IllegalArgumentException e) {
+            // too bad, I don't care
+        } catch (InvocationTargetException e) {
+            // too bad, I don't care
```
aaime: Just one catch would have been sufficient I guess? ;-)
Anyways, I understand you don't care, but this will result in some differences/misbehaviors, maybe at least log it at FINE level?

src/community/wps-jdbc/src/main/java/org/geoserver/wps/jdbc/JDBCStatusStoreLoader.java
```diff
@@ -0,0 +1,93 @@
...
+        // see if we have the JDBCDatastore configuration ready, otherwise create one from the classpath
+        dataDir = dd;
+        try {
+            Properties params = getParameters();
+            
+            store = DataStoreFinder.getDataStore(params);
+
+        } catch (IOException e) {
+            // TODO Auto-generated catch block
+            e.printStackTrace();
```
aaime: See above, do something meaningful here :-)

src/community/wps-jdbc/src/test/java/org/geoserver/wps/jdbc/JDBCStatusStoreTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+    private DataStore datastore;
+
+    JDBCStatusStore statusStore;
+
+    @Override
+    protected ProcessStatusStore buildStore() {
+        Properties props = new Properties();
+        String home = System.getenv("HOME");
+        File f = new File(home, ".geotools/postgis.properties");
```
aaime: This should really be an online type test, e..g, it should be skipped, not fail, if the postgis.properties file is not there

src/extension/wps/web-wps/src/main/java/org/geoserver/wps/web/ProcessStatusProvider.java
```diff
@@ -20,14 +40,22 @@
...
-
         @Override
         public Object getPropertyValue(ExecutionStatus item) {
             // we might want to have a "C" state for the chained processes
             return item.isAsynchronous() ? "A" : "S";
         }
 
+        @Override
+        public boolean isSearchable() {
+            // we really it isn't sortable
```
aaime: The comment does not make much sense written like this. I guess you wanted to say it's not searchable?

src/extension/wps/web-wps/src/main/java/org/geoserver/wps/web/ProcessStatusProvider.java
```diff
@@ -65,4 +96,139 @@ public Object getPropertyValue(ExecutionStatus item) {
...
+                for(String word:keywords) {
+                    for(Property<?> prop:getProperties()) {
+                        if(prop.isSearchable()) {
+                            if(prop.equals(NODE)||
+                               prop.equals(PHASE)||
+                               prop.equals(TASK)||
+                               prop.equals(USER)) {
+                                likes.add(FF.like(FF.property(prop.getName()), "*"+word+"*"));
+                            }
+                            //TODO: support other properties
```
aaime: Yes, at least the process name and maybe the status?

src/extension/wps/web-wps/src/main/java/org/geoserver/wps/web/ProcessStatusProvider.java
```diff
@@ -65,4 +96,139 @@ public Object getPropertyValue(ExecutionStatus item) {
...
+
+            }
+        }
+        LOGGER.info("built query "+query+" to filter statuses");
+        return store.list(query);
+    }
+
+    @Override
+    public Iterator<ExecutionStatus> iterator(int first, int count) {
+        this.first = first;
```
aaime: This is not thread safe... but I'm wondering if we should care, wicket pages should be locked in the http session.

src/extension/wps/web-wps/src/main/java/org/geoserver/wps/web/ProcessStatusProvider.java
```diff
@@ -65,4 +96,139 @@ public Object getPropertyValue(ExecutionStatus item) {
...
+            query.setMaxFeatures(count);
+        }
+        SortParam sort = getSort();
+        if(sort!=null) {
+            SortByImpl[] sortBys = new SortByImpl[1];
+            final Property<?> property = getProperty(sort);
+            if(property.isSearchable()) {//we really need another flag
+
+                //hard to believe there isn't an implementing class!
+                sortBys[0] = new SortByImpl(new PropertyName() {
```
aaime: Confused here, why can't we use new SortByImple(property.getName()) directly?

aaime:
Hum hum... as a final remark, what happens if we are using Oracle, where all the attributes are forced to be uppercase?
Here is an idea: have the feature type be created, look it up using a case insensitive scan, check if the attributes or name have been in any way altered, if so, wrap the feature source and stores with TransformFeatureSource, it will handle the filter renaming on the fly for you.

src/community/pom.xml
ianturton: No idea, I've had a go at cleaning it up and ended up reformatting

src/community/wps-jdbc/src/main/java/org/geoserver/wps/jdbc/JDBCStatusStore.java
ianturton: because it doesn't tell us how many it will remove which we need to return

src/community/wps-jdbc/src/test/java/org/geoserver/wps/jdbc/JDBCStatusStoreTest.java
ianturton: does GS have a mechanism for that?

src/community/wps-jdbc/src/main/java/org/geoserver/wps/jdbc/JDBCStatusStore.java
aaime: Still probably better to do a count and then a delete by filter in a single transaction, the removal could involve hundred of thousands of items in a long lived store

src/community/wps-jdbc/src/test/java/org/geoserver/wps/jdbc/JDBCStatusStoreTest.java
aaime: As far as I remember it does, check LiveDbmsData and its subclasses/usage points

src/extension/wps/web-wps/src/main/java/org/geoserver/wps/web/ProcessStatusProvider.java
ianturton: Phase is status

src/extension/wps/web-wps/src/main/java/org/geoserver/wps/web/ProcessStatusProvider.java
ianturton: In my (limited) understanding it should be ok

src/extension/wps/web-wps/src/main/java/org/geoserver/wps/web/ProcessStatusProvider.java
ianturton: because (sadly) getName returns a String and SortByImpl demands a PropertyName (and I can't find an implementing class for that)

aaime: Ah... well, you are not supposed to instantiate SortByImpl directly, but you can get a hold of the filter factory (statically, once, the lookup is expensive) and then call sort, which takes a string :-)

ianturton: I knew there had to be an easier way of doing that!!!!

ianturton:
Is there an immediate need to support Oracle (I'd assumed that the SQLDIalect would cover everything for me) or can I leave it as an enhancement?

aaime:
Immediate need I'm afraid

src/community/wps-jdbc/src/main/java/org/geoserver/wps/jdbc/JDBCStatusStore.java
ianturton: It does do a count and then remove by filter. But there is a spurious iterator left over which I'll remove

aaime:
Superceded by #1241

PR #1237, opened 2015-09-29
[GEOS-7217] Ensure we provide a Title element for the Style in Capabilities document

bradh:
The WMS 1.1.1 DTD requires Style to match
"(Name,Title,Abstract?,LegendURL*,StyleSheetURL?,StyleURL?)". Adds
small test data file from the release data set.

Includes a small amount of refactoring to reduce code duplication.

More at https://osgeo-org.atlassian.net/browse/GEOS-7217


aaime:
Looks good to me, the failure on travis is caused by the main build server being unable to deploy on repo.boundlessgeo.com due to network issues, thus, it's unrelated

aaime:

PR #1238, opened 2015-09-29
Doc primary key: Correct table name and schema qualified

emacgillavry:
Following up http://sourceforge.net/p/geoserver/mailman/message/34484253/ and my own experience of creating the table in the 'public' schema instead...


aaime:
Looks good, thank you!

aaime:

PR #1240, opened 2015-09-29
[GEOS-7216] WFS GetCapabilities throws NPE if published layer has no geometry

ridethepenguin:
Fix acts on `CatalogBuilder` to make sure even geometryless layers are assigned a (fake) SRS and lat/lon BBOX.

See [related Mailing list discussion](http://osgeo-org.1560.x6.nabble.com/Should-CatalogBuilder-set-a-default-SRS-if-none-was-found-td5226811.html).

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java
+++ b/src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java
@@ -83,6 +83,9 @@ public class CatalogBuilder {
 
     static final Logger LOGGER = Logging.getLogger(CatalogBuilder.class);
 
+    /** Default SRS; will be set on the provided feature type by lookupSRS methods if none was found */
+    public static final String DEFAULT_SRS = "EPSG:404000";
+
     /**
      * the catalog
      */
@@ -658,6 +661,8 @@ public class CatalogBuilder {
             } catch (FactoryException e) {
                 throw (IOException) new IOException().initCause(e);
             }
+        } else {
+            ftinfo.setSRS(DEFAULT_SRS);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java
```diff
@@ -515,6 +518,10 @@ public void setupMetadata(FeatureTypeInfo ftinfo, FeatureSource featureSource)
     public ReferencedEnvelope getLatLonBounds(ReferencedEnvelope nativeBounds,
             CoordinateReferenceSystem declaredCRS) throws IOException {
         if (nativeBounds != null && declaredCRS != null) {
+            // shortcut: return empty envelope if nativeBounds is empty
+            if (nativeBounds.isEmpty()) {
+                return new ReferencedEnvelope(nativeBounds, DefaultGeographicCRS.WGS84);
```
aaime: Hi Stefano, wondering, is we go this way, are the bounds published in WFS valid schema wise? 
Normally for geometryless data I still keep a valid bbox, e.g., maxx > minx, maxy > miny

ridethepenguin: Well, it is XML valid, not sure if it is spec compliant though... it just felt weird to see a null envelope (i.e. minx > maxx, miny > maxy) being transformed to a non-null one, as the previous code was doing.

I can revert this change, if you think it is not correct.

aaime: You should check the WFS spec and see if it has an explicit notion of empty envelope, if not, it must be a non empty one.

ridethepenguin: I couldn't find any notion of empty envelope in the WFS spec, so I restored the original code.

PR #1242, opened 2015-09-30
[GEOS-7221] Document how to debug extensions from eclipse

tbarsballe:
For https://osgeo-org.atlassian.net/browse/GEOS-7221

### DIFF
--- a/doc/en/developer/source/eclipse-guide/index.rst
+++ b/doc/en/developer/source/eclipse-guide/index.rst
@@ -18,6 +18,15 @@ Running and debugging
 Run or debug the class ``org.geoserver.web.Start`` in the ``web-app`` 
 module. The steps to do so are detailed in the :ref:`quickstart`.
 
+Running GeoServer with Extensions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+By default, GeoServer will run without any extensions enabled. In order to run GeoServer with extensions, the ``web-app`` module declares a number of profiles used to enable specific extensions when running ``Start``. To enable an extension, re-generate the root eclipse profile with the appropriate maven profile(s) enabled::
+
+  % mvn eclipse:eclipse -P wps
+
+The full list of supported profiles can be found in ``src/web/app/pom.xml``.
+
 Setting the data directory
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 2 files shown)


doc/en/developer/source/eclipse-guide/index.rst
```diff
@@ -18,6 +18,16 @@ Running and debugging
 Run or debug the class ``org.geoserver.web.Start`` in the ``web-app`` 
 module. The steps to do so are detailed in the :ref:`quickstart`.
 
+Running GeoServer with Extensions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+By default, GeoServer will run without any extensions enabled. In order to run GeoServer with extensions, the ``web-app`` module declares a number of profiles used to enable specific extensions when running ``Start``. To enable an extension, re-generate the ``web-app`` eclipse profile with the appropriate maven profile(s) enabled::
+
+  % cd web/app
+  % mvn eclipse:eclipse -P wps
```
aaime: Hum, this is a sub-par way imho, as it does not link the project with the wps project, but only with the jars.
Better to stay in the root directory, and do the same command, it will create the eclipse project for the wps module too and link the two project by project, so that you can work directly against the sources and make fixes during the debug session

mcrmcr: I think Andrea is right. If I want to debug an extension I do a mvn eclipse:clean eclipse:eclipse for the extension, import the project in my workspace and add the project to the class path (project settings)  of web-app.

aaime: Christian, I did that in the past but it occasionally caused issues because the project imported depended on the jars of other core projects that are also linked as project dependencies in web-app. Depending on the final classpath order you might end up wondering why a change made on the source (without a maven rebuild) does not get reflected in your debug sessions.

mcrmcr: Upps, I was lucky in the past, never had such troubles : -)

tbarsballe: @aaime - I will update based on your suggestions, thanks. I wasn't quite sure of the best way of doing this.
@mcrmcr - That is how I had been doing it until now, then I finally ran into issues, hence this update.

tbarsballe:
Updated instructions based on feedback

PR #1244, opened 2015-10-01
Fix env-var name for data dir

bradh:


aaime:

PR #1246, opened 2015-10-01
[GEOS-7226] GML links in Layer Preview don't work for app-schema layers

ridethepenguin:
A couple of test classes extending `GeoServerWicketTestSupport` have been added to `gs-app-schema-test` module.

### DIFF
--- a/src/extension/app-schema/app-schema-test/pom.xml
+++ b/src/extension/app-schema/app-schema-test/pom.xml
@@ -163,6 +163,14 @@
             <version>2.0.0-1</version>
             <scope>test</scope>
         </dependency>
+        <!-- required for Web UI tests -->
+        <dependency>
+            <groupId>org.geoserver.web</groupId>
+            <artifactId>gs-web-demo</artifactId>
+            <version>${project.version}</version>
+            <classifier>tests</classifier>
+            <scope>test</scope>
+        </dependency>
 	</dependencies>
 
     <profiles>
--- a/src/extension/app-schema/app-schema-test/pom.xml
+++ b/src/extension/app-schema/app-schema-test/pom.xml
@@ -166,6 +166,19 @@
         <!-- required for Web UI tests -->
         <dependency>
             <groupId>org.geoserver.web</groupId>
+            <artifactId>gs-web-core</artifactId>
+            <version>${project.version}</version>
+            <classifier>tests</classifier>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.geoserver.web</groupId>
+            <artifactId>gs-web-demo</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.geoserver.web</groupId>
             <artifactId>gs-web-demo</artifactId>
             <version>${project.version}</version>
             <classifier>tests</classifier>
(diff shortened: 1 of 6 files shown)


src/extension/app-schema/app-schema-test/pom.xml
```diff
@@ -163,6 +163,14 @@
...
             <scope>test</scope>
         </dependency>
+        <!-- required for Web UI tests -->
+        <dependency>
+            <groupId>org.geoserver.web</groupId>
+            <artifactId>gs-web-demo</artifactId>
+            <version>${project.version}</version>
+            <classifier>tests</classifier>
+            <scope>test</scope>
+        </dependency>
```
bencaradocdavies: @ridethepenguin this branch fails to build (compile errors) in Maven as it has a different dependency model from Eclipse, which lumps all phases and classifiers together. I think you need test phase dependencies on gs-web-core tests and gs-web-demo main to get GeoServerWicketTestSupport and MapPreviewPage respectively. I got compilation working in Maven with this:

        
        <dependency>
            <groupId>org.geoserver.web</groupId>
            <artifactId>gs-web-demo</artifactId>
            <scope>test</scope>
        </dependency>
        <dependency>
            <groupId>org.geoserver.web</groupId>
            <artifactId>gs-web-core</artifactId>
            <scope>test</scope>
        </dependency>

Even with this I get a test failure:

Tests in error: 
  initializationError(org.geoserver.test.web.AbstractMapPreviewPageTest): Test class should have exactly one public constructor

Kind regards,
Ben.

ridethepenguin: @bencaradocdavies Ouch, looks like I forgot to check the Maven build.

I pushed another commit to fix the build, but refrained from squashing it and rebasing the branch (I remember you advised against it): I can do it in a new branch and submit a new PR, if you like.

bencaradocdavies:
@ridethepenguin that looks great. Because you refrained from rebasing, I was able to use "git pull" to update my local branch tracking your branch.

I can confirm that, with your most recent changes, the Maven offline build passes. However, the app-schema online tests fail because they use testSourceDirectory to recycle the tests in gs-app-schema-test. The solution is to include your additional dependencies in their poms as well. I have tested this fix and will cherry-pick it on to master as soon as I have merged this pull request.

I also manually tested the GML link created for gsml:MappedFeature in the app-schema tutorial configuration and it works well. I notice that there are no links for the gsml:GeologicUnit type; while OpenLayers and KML will not work as it has no top-level geometry, did you expect a GML link for this type? You mentioned on the mailing list adjusting the behaviour of the OpenLayers/KML/WMS links in another PR; this change could also permit geometryless GML, which I think was your intent.

Kind regards,
Ben

bencaradocdavies:
Convenience link to Jira:
https://osgeo-org.atlassian.net/browse/GEOS-7226

ridethepenguin:
@bencaradocdavies By all means, I expected a link for the gsml:GeologicUnit type (there's also a test covering this case); I just tested the app-schema tutorial configuration (you refer to the one included in the app-schema plugin archive, right?), and it seems that it doesn't come with a pre-configured layer.xml file for the GeologicUnit type. As soon as I added one, I could see the layer listed in the layer preview page and the GML link was working fine.

Can you confirm that this is working for you as well?

bencaradocdavies:
@ridethepenguin thanks for solving this problem. I can confirm that, when I add a layer.xml, the GeologicUnit type appears in the layer preview and its GML link works as expected. I have now added a layer.xml for GeologicUnit and fixed the titles and abstracts for both types (neglected before as they never worked). Thanks!

PR #1247, opened 2015-10-02
[GEOS-7227] Defer bounds calculation during import

tbarsballe:
Moved bounds calculation from `createContext()` to `doImport()` (DataStoreFormat only)
See https://osgeo-org.atlassian.net/browse/GEOS-7227

Some changes to test cases:
* Removed checks for null bounds; no longer applicible
* Added check to confirm a user-defined CRS is preserved after bounds are calculated

### DIFF
--- a/src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
+++ b/src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
@@ -917,6 +917,29 @@ public class Importer implements DisposableBean, ApplicationListener {
             }
 
             addToCatalog(task);
+            
+            //Calculate bounds, if applicable
+            if (task.getLayer().getResource() instanceof FeatureTypeInfo) {
+                FeatureTypeInfo featureType = (FeatureTypeInfo) task.getLayer().getResource();
+                FeatureTypeInfo resource = getCatalog().getResourceByName(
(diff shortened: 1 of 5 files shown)


src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -987,6 +1010,11 @@ void doIndirectImport(ImportTask task) throws IOException {
                         resource.setLatLonBoundingBox(cb.getLatLonBounds(nativeBounds,
                                 resource.getCRS()));
                         getCatalog().save(resource);
+                        
+                        //Do not re-calculate on subsequent imports
+                        if (resource.getMetadata().get("recalculate-bounds").equals(Boolean.TRUE)) {
```
aaime: Hum... wondering, shall we really keep this one around "forever"? Can't we remove it instead?

tbarsballe: Excellent point, I had forgotten that was an option. I will change this.

aaime:
Merged. Backport as you see fit, and remember to close the associated Jira ticket.

tbarsballe:
Thanks for the review and merge. Backported to 2.8.x and 2.7.x.

PR #1248, opened 2015-10-03
[GEOS-7150] Features counted twice for WFS queries with GeoJSON responses

aaime:


src/wfs/src/main/java/org/geoserver/wfs/CountExecutor.java
```diff
@@ -40,7 +40,8 @@ public int getCount() throws IOException {
         if(providedCount != COUNT_UNSET) {
             return providedCount;
         } else {
-            return source.getCount(query);
+            // make sure we get a count, feature source can return -1
+            return source.getFeatures(query).size();
```
jodygarnett: not sure if the comment goes with the code here? Do you:

````
// mean make sure we get a count, source.getCount(query) can return -1
```

src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java
```diff
@@ -296,7 +298,7 @@ public FeatureCollectionResponse run(GetFeatureRequest request)
 
         boolean isNumberMatchedSkipped = false;
         int count = 0; // should probably be long
-        int totalCount = 0;
+        BigInteger totalCount = BigInteger.ZERO;
```
jodygarnett: BigInteger really?

src/wfs/src/main/java/org/geoserver/wfs/CountExecutor.java
aaime: Yes, that is the idea.

src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java
aaime: That's what the EMF interface takes, complain with the EMF code generator, not me :)

src/wfs/pom.xml
```diff
@@ -42,6 +42,10 @@
       <artifactId>gt-shapefile</artifactId>
     </dependency>
     <dependency>
+      <groupId>cglib</groupId>
```
aaime: Verified, it's indeed not going to add a new jar in the GeoServer release:

````
aaime@hydra ~/devel/gs_releases/geoserver-2.7.2/webapps/geoserver/WEB-INF/lib $ ls *cglib*
   cglib-nodep-2.2.jar
````

File: src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java [hunk 4]
```diff
@@ -130,20 +130,10 @@ protected void write(FeatureCollectionResponse featureCollection, OutputStream o
         boolean hasGeom = false;
 
         // get feature count for request
-        BigInteger featureCount = null;
-        // for WFS 1.0.0 and WFS 1.1.0 a request with the query must be executed
-        if(describeFeatureType != null) {
-            if (describeFeatureType.getParameters()[0] instanceof GetFeatureType) {
-                featureCount = BigInteger.valueOf(getFeatureCountFromWFS11Request(describeFeatureType, wfs));
```
sampov2: I think we could get rid of getFeatureCountFromWFS11Request() as well

PR #1249, opened 2015-10-05
[GEOS-7200] Remove snapshot use for OSGEO repo.

bradh:
I did a full build (clean install), a release build (install -Prelease), and a "really clean" build where I removed the whole $HOME/.m2 directory, then did a clean install -Prelease build.

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -64,9 +64,6 @@
     <id>osgeo</id>
     <name>Open Source Geospatial Foundation Repository</name>
     <url>http://download.osgeo.org/webdav/geotools/</url>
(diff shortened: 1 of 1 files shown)


aaime:
Merged... unsure whether to backport, or not. Opinions?

bradh:
I think could be worth waiting to make sure nothing goes awry. I probably should make a test on the branch first too.

bradh:
In retrospect, probably not worth backporting. The only common builders of the branches are likely the CI bots, and for saving a little traffic, it is isn't worth the effort.

PR #1253, opened 2015-10-06
Fixing pom related to GEOT-5148 (NetCDF 4.6.2 upgrade) since UCAR cha…

dromagnoli:
…nged scope of a dependency from compile to runtime


dromagnoli:
Required as consequence of  GEOT-5148 backport to 13.x, made with 
https://github.com/geotools/geotools/pull/917

PR #1254, opened 2015-10-06
update oracle-readme.txt, adding note about ojdbc14jar

rduivenvoorde:


aaime:
Hi Richard, I believe that ojdbc4 still works fine for those that have Oracle 10, and it's actually the recommended driver (I did a full test run against a Oracle XE 10 a few days ago and it worked all fine),
have had issues using ojdbc7 against Oracle 10 for example.

rduivenvoorde:
Hi Andrea, are you sure? Because I understand we are running:
Oracle Database 10g Enterprise Edition Release 10.2.0.5.0
And I had the error I talk about in the readme...

By the way, I understand that ojdbc14.jar is actually the jdbc jar for java 1.4(!) while ojdbc7.jar is for java 7.... strange numbering

But if you are sure, I'll drop the PR, hoping people will just google and find it if they encounter the problem. Jody also told me het put it in the release notes now:

http://docs.geoserver.org/latest/en/user/data/database/oracle.html#installing-the-oracle-extension

He talks about ojdbc4.jar, but downloading an old extension version 2.5.2 it is ojdbc14.jar too.... Not sure where the ojdbc4.jar comes from.

aaime:
Hum... I have to double check, I'm sure I had to go back to a ojdbc14 in a recent work I did, because with ojdbc7 some parts were misbehaving against a local Oracle XE 10 I have handy, but it might have had to do with blob fields.

aaime:
Personal note, still haven't had time to check what's the situation vs Oracle 10 with ojdbc14. Anyone has a running 2.8.x installation with it and can comment?

mprins:
Note that Oracle 10.2 premier support ended July 2010 and extended support ended July 2013

dion-ict:
The latest version of Oracle instantclient 10 (10.2.0.5.0) still contains ojdbc14.jar (dated 2010-04-20) so for Oracle 10 this likely remains the recommended jar to use. From version 11 on, the instantclient distribution contains newer ojdbc jars versioned after the java version: ojdbc5.jar, ojdbc6.jar (for 11.2.0.4.0) and ojdbc6.jar, ojdbc7.jar (for 12.1.0.2.0).

jodygarnett:
@dion-ict can you propose alternate wording? would like to have this doc fix in for the RC on march 3rd.

rduivenvoorde:
I'm not sure if it is authorative, but see: https://osgeo-org.atlassian.net/browse/GEOT-5098
There Torben Barsballe  says: "GeoTools expects ojdbc7.jar as of version 14." AND I was hitting exact this problem (and fixed it with ojdbc7.jar).

mprins:
I just want to point out that in reality there is a multitude of "ojdbc7.jar" (as well as ojdbc6, ojdbc5, ojdbc14 and classes12.jar) so just talking about the Java API level is not a best practice. Changing one for the other and not running into a bug may be because you have unknowingly changed patch-level unknowingly (and just got lucky)

In general one should use the version that matches the java runtime and the database version/patchlevel (the latter may be found in the manifest file inside the jar file)

YMMV with any version since this is oracle :bomb:  we ran into this exotic little gem the other day: https://community.oracle.com/thread/3647914?start=0&tstart=0 

I've also had drivers reporting STRUCT instead of OTHER for geometry columns and only having a patch number difference. And drivers unable to retrieve column metadata in prepared statements (eg ojdbc6-12.1.0.1.jar and ojdbc7-12.1.0.1.jar should be avoided)

@dion-ict 's note about the instantclient seems off-topic since geoserver uses just JDBC and not most of what instantclient provides (OCI, ODBC, OCCI etc, etc) and Database v10 is long gone from the list of supported versions

That being said I don't think this PR adds anything, as the original doc does not mention using 14.

rduivenvoorde:
Ok, clear. Not sure where I had the ojdbc14.jar from. I thought it came with the plugin, but I was wrong.

Closing then.

aaime:
ojdbc14 used to come with the geoserver plugin... until we removed it in GeoServer 2.8

aaime:
@mprins brrr... it's no surprise that everybody loves to hate oracle :-p

PR #1255, opened 2015-10-06
[GEOS-7230] gs:DownloadEstimator should check size is less than Integer.MAX_VALUE

ridethepenguin:
This is not to be seen as a definitive solution to the problem, just a bare minimum sanity check, as OOM exceptions may still occur and cause the actual download to fail, even if the estimator said it was OK to download.


simboss:
backport as needed

ridethepenguin:
Travis build is green, merging. Backports to stable and maintenance will follow.

PR #1258, opened 2015-10-07
Fix unit test failures

smithkm:
Fixes unit test failures that seem to be local to my machine in the WFS and OGR-WPS modules.

It's not clear why the WFS tests only fail on my machine.  The OGR-WPS ones are ignored if ogr2ogr is not present.  Adding it to ares so the tests run may be a good idea.

### DIFF
--- a/src/platform/src/test/java/org/geoserver/util/XmlTestUtil.java
+++ b/src/platform/src/test/java/org/geoserver/util/XmlTestUtil.java
@@ -1,0 +1,200 @@
+/* (c) 2015 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+
+package org.geoserver.util;
+
+import static org.hamcrest.Matchers.any;
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.hasItem;
+
+import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.io.PrintWriter;
+import java.util.AbstractList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import javax.xml.transform.Transformer;
+import javax.xml.transform.TransformerException;
+import javax.xml.transform.TransformerFactory;
+import javax.xml.transform.dom.DOMSource;
+import javax.xml.transform.stream.StreamResult;
+
+import org.custommonkey.xmlunit.SimpleNamespaceContext;
+import org.custommonkey.xmlunit.XMLUnit;
+import org.custommonkey.xmlunit.XpathEngine;
+import org.custommonkey.xmlunit.exceptions.XpathException;
+import org.hamcrest.BaseMatcher;
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.w3c.dom.Document;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+
+/**
+ * An aid for XML related testing.
+ * 
+ * @author Kevin Smith, Boundless
+ *
+ */
+public class XmlTestUtil {
+    private OutputStream showXML = null;
+    private Map<String, String> namespaces = new HashMap<>();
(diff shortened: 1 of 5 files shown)


smithkm:
@aaime, @ridethepenguin Does the fix for the OGR-WPS test look right?  Is it just a case of there being two different valid outputs (with and without the Folder element).

aaime:
Kevin, unsure, 99% of the change is just a different way to make the same tests, the actual change is lost somewhere in there.

aaime:
Btw, I agree xmlunit is basically dead and the matchers approach looks better, but I cannot see what the actual fix is in the half minute I have to look at this.

smithkm:
Right the fixes are each just a line or so and the rest is the changes I made to diagnose the problem.

The change in the OGR-WPS test was here https://github.com/geoserver/geoserver/pull/1258/files#diff-cd91506c58e33f22e9d862e5872b53abL147

I changed `//kml:kml/kml:Document/kml:Schema` to `//kml:kml/kml:Document/kml:Schema | //kml:kml/kml:Document/kml:Folder/kml:Schema`

So it allows a `Folder` element as an intermediate between the `Document` and the `Schema`.   I think that shouldn't be significant, but I don't know exactly what you and @ridethepenguin were testing.

aaime:
Hi Kevin, thanks a lot for the explanation. About the test where you had to disable entity resolution, I believe Torben (or was that Travis... sorry, can't remember) had a different take on the issue, and tried to make sure each and every xsd was resolved from the classpath. Could you double check with them?

smithkm:
I spoke to @tbarsballe  earlier the first time I ran into the issue and we were unable to figure out why this XSD wasn't being found correctly on my machines (Two different ones) but was working elsewhere (Ares, Travis, Jody and Torben's laptops, and presumably every other GeoServer developer's machine) and for other XSDs on my machines.

src/platform/src/test/java/org/geoserver/util/XmlTestUtil.java
```diff
@@ -0,0 +1,200 @@
...
+            public void describeMismatch(Object item, Description description) {
+                XpathEngine engine = XMLUnit.newXpathEngine();
+                engine.setNamespaceContext(namespaceContext);
+                try {
+                    List<Node> nodes = nodeCollection(engine.getMatchingNodes(xPath, (Document) item));
+                    
+                    matcher.describeMismatch(nodes, description);
+                    
+                    if(showXML != null) {
+                        printDom((Document) item, System.out);
```
tbarsballe: Shouldn't this print to showXML, not System.out?

src/wfs/src/test/java/org/geoserver/wfs/v1_1/GetFeatureTest.java
```diff
@@ -17,21 +18,44 @@
...
-	
+    
+    XmlTestUtil xmlUtil;
+    
+    @Before
+    public void setupXmlUtil() {
+        xmlUtil = new XmlTestUtil();
+        xmlUtil.addNamespace("wfs", "http://www.opengis.net/wfs");
+        xmlUtil.addNamespace("sf", "http://cite.opengeospatial.org/gmlsf");
+        xmlUtil.setShowXML(System.out); // Uncomment to display XML on failure
```
tbarsballe: Was this intended to be commented out?

src/platform/src/test/java/org/geoserver/util/XmlTestUtil.java
smithkm: Yes, I'd initially made showXML a boolean then changed it to a stream for more flexibility.  I forgot to update this.

src/wfs/src/test/java/org/geoserver/wfs/v1_1/GetFeatureTest.java
smithkm: Yes, I forgot that, thanks.

tbarsballe:
Looks good overal, XSD changes look fine.
I found a couple of minor issues; added comments.
I tried building this on my machine to see if the changes that fixed your build will cause failures on another machine, and the build ran fine on my machine.

PR #1259, opened 2015-10-07
GEOS-7172, Fix checkbox/radio inputs in UI so they can be toggled by their labels.

smithkm:
Changes to geoserver.css and to most of the checkboxes in the geoserver UI.  Allows the checkboxes to be toggled by their labels while not toggling then when clicking on whitespace near the label and also gives a hanging indent of the label if it is multiline.  Also fixes/restores missing `for` attributes.

I tried to use `wicket:for` but that doesn't seem to work, possibly because of the ancient wicket we're using.  I couldn't find a clear indication of when that feature was added to wicket.


jodygarnett:
Pull request seems fine, since we have gone back and forth on this bug you may wish to discuss back porting to stable/maintenance on geoserver-devel.

PR #1267, opened 2015-10-15
Revisit Layer Bound calculations [GEOS-7248]

dvntucker:
Small feature originally proposed by @tbarsballe to calculate layer bound from the declared/native CRS bounds.


dvntucker:
The specific subtask is actually GEOS-7248

jodygarnett:
Thanks updated pull request title as above.

Thanks for include a test case, can I ask you to update the documentation page (so users can know what to expect when clicking this button - it will especially help since not all SRS have a useful bounds recorded against them).

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
```diff
@@ -37,6 +37,7 @@
 import org.geoserver.web.wicket.SRSToCRSModel;
 import org.geotools.geometry.jts.ReferencedEnvelope;
 import org.geotools.referencing.CRS;
+import org.opengis.geometry.Envelope;
```
jodygarnett: please update header (c) range :)

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
```diff
@@ -135,6 +137,36 @@ public boolean getDefaultFormProcessing() {
 
         };
     }
+    
+    /**
+     * Compute the native bounds from the native CRS. Acts as an alternative to computing the bounds
+     * from the data itself.
+     */
+    AjaxSubmitLink computeBoundsFromSRS(final Form refForm, 
+            final EnvelopePanel nativeBoundsPanel) {
```
jodygarnett: You may want to look at your code formatting settings, think we have a 100 col width here?

src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java
```diff
@@ -1530,4 +1531,39 @@ public void attach(WorkspaceInfo wsInfo) {
         
         return attributes;
     }
+
+    /**
+     * Creates referenced envelope from resource based off the native or declared SRS. This bbox
+     * depends on the projection policy.
+     * 
+     * - force declared, reproject native to declared: use the declared SRS bounding box 
+     * - keep native:
```
aaime: Incomplete javadoc

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
aaime: Yes we do

aaime:
Documentation needs updating too

src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java
jodygarnett: sadly javadoc does html, not markup, so if you want a bulleted list use:
```
<ul>
<li>force declared, reproject native to declared: use the declared SRS bounding box </li>
<li>keep native: ...</li>
</ul>
```

aaime:
Btw, once the documentation is added, remember to have a single commit with a reference to a Jira ticket in the pull (either force pull or, if you think someone else might have worked on it, close and do a new one)

jodygarnett:
Merged (as a single commit) e4237a09dede401ccca76dd49df63efdba6b0f1f 

Thanks @dvntucker and welcome to GeoServer.

PR #1268, opened 2015-10-16
removed arbitrary setting of default data store

mtCarto:
Removed the arbitrary setting of the default DataStore. The store selector will now default to "Create New" in the importer screen. This in preparation for a minor UI change to let users set and view the default Store.

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/impl/CatalogImpl.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/CatalogImpl.java
@@ -158,11 +158,6 @@ public class CatalogImpl implements Catalog {
         StoreInfo added;
         synchronized (facade) {
             added = facade.add(resolve(store));
-            
-            // if there is no default store use this one as the default
-            if(getDefaultDataStore(store.getWorkspace()) == null && store instanceof DataStoreInfo) {
-                setDefaultDataStore(store.getWorkspace(), (DataStoreInfo) store);
-            }
         }
         added(added);
     }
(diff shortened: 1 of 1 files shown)


aaime:
Broke the build, not sure you can actually change this expected behavior, it has been around for a long-ish time and I believe there is code expecting to find a default store.
Also, any chance must have a test case going along with it, this one does not. Please refer to the contribution guide here:

jodygarnett:
The failure is no longer visible, what kind of problem are we looking at? As far as I know only importer uses the "default" store setting?

aaime:
My comment was about the notion of default store, which is available to all plugins, not just importer, not just the official ones, if we are changing it we should go though a proposal.

That said... the current version of the patch is just adding one whitespace in front of the line, or am I missing something?

jodygarnett:
You are not missing something, @mtCarto is learning github and did not know that reverting his change would show up as above.

I was hoping this change was simple (allowing Importer to choose "create new" if a default datastore had not been selected by the user, and then provide a user interface to allow a default datastore to be selected. Think this is a case of a plan not surviving first contact with the codebase.

Do you feel a proposal is needed for the above UI change?

aaime:
Err... for a UI change only of course not, but if you need to change the catalog internal machinery and behavior, then we have a problem of course. Why don't you just default to "create new" and have the user select the target store otherwise? 
The GUI does not allow to set a store as the workspace default anyways.

jodygarnett:
Okay, we will make a change proposal Monday. 

(The current setup where the "first" datastore is recorded as the default datastore for a workspace prevents the idea of the UI choosing "create new" when no store has been chosen - since a store is "always" chosen)

jodygarnett:
Closing this pull request pending proposal / discussion / etc...

PR #1277, opened 2015-10-20
GeoServer updates needed as consequency of GEOT-5247

dromagnoli:
Make sure to not merge this pull request before merging 
https://github.com/geotools/geotools/pull/999


dromagnoli:
The travis check failure is expected since it doesn't take into account the code contained in pull request 
geotools/geotools#999

dromagnoli:
Travis seems getting old classes (from October, the 6th) instead of the latest ones.

PR #1285, opened 2015-10-20
Update sld-introduction.rst

rod-engineer:
The line numbers in the text, lines 109-114, don't match the SLD code block in lines 77-105.
Add 1 to the line number references.

### DIFF
--- a/doc/en/user/source/styling/sld-introduction.rst
+++ b/doc/en/user/source/styling/sld-introduction.rst
@@ -106,12 +106,12 @@ to style them as red circles with a size of 6 pixels.
 
    
 Although the example looks long, only a few lines are really important to understand.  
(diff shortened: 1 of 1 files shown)


aaime:
Indeed number line up perfectly for me. Not sure we can do anything to avoid that though, the line numbering is done automatically by sphinx... but I'm afraid that would also invalidate most available examples which are assuming the current wrapping.

rod-engineer:
I verified on my Chrome browser. Changing the font size sometimes gives the correct behavior (per aaime's comment) and sometimes makes it off one line (per bradh's comment).

I will close close this pull request. The documentation is clear enough, even if the line numbers are off. 

Thank you for your comments.

PR #1287, opened 2015-10-21
Sync SLD encoding with web page encoding

Vampouille:
In style editor, we can edit SLD file but browser will always use
character encoding from html page not xml encoding. So we need to sync
encoding of SLD file to encoding of web page.


aaime:
What misbehavior have you seen, that made you propose this pull request?

Vampouille:
I try to create a new SLD style by copying an existing style. If original SLD was not encoded to UTF-8 then browser will POST an XML file with UTF-8 encoding because HTML pages are in UTF-8.

PR #1295, opened 2015-10-22
[GEOS-7231]: CSW DirectDownload operation and download links

dromagnoli:


bradh:
I realise that the documentation does only talk about coverage layers, but is there any reason why this doesn't work on all layers?

bradh:
I'm having trouble testing this with NetCDF and GRIB files (can't create store), but I did manage to get this going with a TIFF (the wind.tiff test data set from the GeoSolutions multidim training course). That produced the expected entries, but when I tried to download them with curl/wget, it appeared to stall at about 11%. It appears that its all there, and what is happening is that the content-length is being set to the uncompressed size, not the compressed size. I don't yet see why that is occurring though.

bradh:
Does #1302 replace this PR?

dromagnoli:
Worked on reviewed items and created a new one:
https://github.com/geoserver/geoserver/pull/1302.

Removed the content-length setting. Discussing with Andrea, common practice in GeoServer is not setting it in case to avoid the need of fully writing a zip file to disk before streaming it back.

In response to the Coverage-management-only topic, we got mandate to work on the coverage download only. However, it was our intention to open a JIRA to add the vectorial support too (which shouldn't hard work). We will create the JIRA in a few.

In relation to the default config settings in the UI it should use default values.

bradh:
Thanks for the feedback.

dromagnoli:
Thanks to you for the review.

PR #1298, opened 2015-10-24
[GEOS-6876] Update shapefile quickstart to reflect new UI elements.

bradh:
https://osgeo-org.atlassian.net/browse/GEOS-6876


aaime:
Images should never be larger than 740 pixels (can't remember the exact value), otherwise we have issues in PDF generation and I believe in HTML generation too. See also https://github.com/geoserver/geoserver/blob/master/doc/en/themes/geoserver/static/default.css#L142

aaime:
The rest looks good to me

bradh:
OK, I thought it would be resized, based on http://docs.geoserver.org/latest/en/docguide/sphinx.html#images

aaime:
Oh ok, I was not aware of that bit. Does that actually happen when you generated HTML? I guess you are not setup to generate PDF eh?

bradh:
It doesn't appear to rescale on generation (images in the build/html/_images directory appear to be the same as the source), but it does rescale on display - to about 650px.

When I tried generating PDF (which has a lot of other issues), it actually looks better with the higher resolution images. There is definitely scaling, but it appears the images actually get scaled up, so the source images that are small end up looking more blurred, although I can't say any of it looks great.

aaime:
Good, merged then

PR #1302, opened 2015-10-27
[GEOS-7231]: CSW DirectDownload operation and download links

dromagnoli:


bradh:
Does this version deal with the partial download issue (see https://github.com/geoserver/geoserver/pull/1295#issuecomment-150761108)?

bradh:
Retested with updated geotools and all seems good now. Recommend it be merged.

dromagnoli:
Thanks for review and feedbacks.
Merging it.

PR #1303, opened 2015-10-27
[GEOS-7280] Avoid catalog linear scans in GWC integration listeners

aaime:


aaime:
Posted for review, I guess with GWC/JDBCConfig being involved @smithkm would be a good candidate

src/gwc/src/main/java/org/geoserver/gwc/GWC.java
```diff
@@ -1248,18 +1254,6 @@ public GridSetBroker getGridSetBroker() {
...
-    // }
-
-    public List<LayerInfo> getLayerInfos() {
-        return getCatalog().getLayers();
-    }
-
-    public List<LayerGroupInfo> getLayerGroups() {
-        return getCatalog().getLayerGroups();
-    }
-
```
smithkm: Changing the public API here?

smithkm:
Looks OK.  Lack of unit tests makes sense given it's a refactoring for performance.  If backported `getLayerInfos` and `getLayerGroups` probably shouldn't be removed and marking them deprecated in 1.8.x and 1.7.x would be a good idea even if this is not backported.

src/gwc/src/main/java/org/geoserver/gwc/GWC.java
aaime: The idea is that these methods are dangerous, when used the trigger lots of db loads.  I can add them back as deprecated, no problem, I removed only because I wanted to make sure not to leave anything using them.
And then I left them removed because the GWC facade is, to the limits of my knowledge, not used anywhere outside of the GWC module.

smithkm: Yeah, I rather doubt anything is using them either, and I agree they are dangerous so removing them outright in master is fine by me. I would favour merely making them deprecated in a backport though, just in case someone somewhere has decided to use them in some custom extension, as unlikely as it may be,

PR #1305, opened 2015-10-27
[GEOS-7281] move WPS Process to RUNNING state during execution.   

DBlasby:
See ticket https://osgeo-org.atlassian.net/browse/GEOS-7281

### DIFF
--- a/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/executor/WPSExecutionManager.java
+++ b/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/executor/WPSExecutionManager.java
@@ -391,6 +391,13 @@ public class WPSExecutionManager implements ApplicationContextAware,
                 SubProgressListener executionListener = new SubProgressListener(listener,
                         inputPercentage, executionPercentage);
                 notifier.checkDismissed();
+                
+                //The system only moves from Queued to Running when the progress notifier fires (!!!).  This fires when inputs are read.
+                //   However, if there are no inputs, it never moves to running state!
+                //Since we are executing, we can move this from queued to running state. 
+                if (status.getPhase() == ProcessState.QUEUED) {
+                	status.setPhase(ProcessState.RUNNING);
+                }
                 processManager.submit(status.getExecutionId(), status.getProcessName(), inputs,
                         executionListener, status.isAsynchronous());
 
--- a/src/extension/wps/wps-core/src/test/java/org/geoserver/wps/other/NoArgWPS.java
+++ b/src/extension/wps/wps-core/src/test/java/org/geoserver/wps/other/NoArgWPS.java
@@ -1,0 +1,16 @@
+package org.geoserver.wps.other;
+
+
+//import org.geotools.process.factory.DescribeParameter;
+import org.geotools.process.factory.DescribeProcess;
+import org.geotools.process.factory.DescribeResult;
+import org.geoserver.wps.gs.GeoServerProcess;
+
+@DescribeProcess(title="NoArgWPS", description="NoArgWPS - test case for no argument process")
+public class NoArgWPS implements GeoServerProcess {
+
+   @DescribeResult(name="result", description="output result")
+   public String execute( ) {
+        return "Completed!"  ;
+   }
+}
--- a/src/extension/wps/wps-core/src/test/java/org/geoserver/wps/other/NoArgWPSTest.java
+++ b/src/extension/wps/wps-core/src/test/java/org/geoserver/wps/other/NoArgWPSTest.java
@@ -1,0 +1,58 @@
+package org.geoserver.wps.other;
+
+import static org.junit.Assert.*;
+
+import java.io.InputStream;
+import java.io.StringWriter;
+import java.util.Map;
+
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.transform.OutputKeys;
+import javax.xml.transform.Transformer;
+import javax.xml.transform.TransformerConfigurationException;
+import javax.xml.transform.TransformerFactory;
+import javax.xml.transform.dom.DOMSource;
+import javax.xml.transform.stream.StreamResult;
+
+import org.apache.commons.io.IOUtils;
+import org.geoserver.data.test.SystemTestData;
+import org.geoserver.wps.WPSTestSupport;
+import org.junit.Test;
+import org.w3c.dom.Document;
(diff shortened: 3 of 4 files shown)


src/extension/wps/wps-core/src/test/java/org/geoserver/wps/other/NoArgWPS.java
```diff
@@ -0,0 +1,16 @@
+package org.geoserver.wps.other;
```
jodygarnett: Will need project header.

src/extension/wps/wps-core/src/test/java/org/geoserver/wps/other/NoArgWPS.java
```diff
@@ -0,0 +1,16 @@
...
+
+
+//import org.geotools.process.factory.DescribeParameter;
```
jodygarnett: We tend to not leave in commented out code.

src/extension/wps/wps-core/src/test/java/org/geoserver/wps/other/NoArgWPSTest.java
```diff
@@ -0,0 +1,58 @@
+package org.geoserver.wps.other;
```
jodygarnett: file header

src/extension/wps/wps-core/src/test/java/org/geoserver/wps/other/NoArgWPSTest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.w3c.dom.Document;
+
+public class NoArgWPSTest extends WPSTestSupport  {
+
+	
+	
+	 @Override
+	    protected void registerNamespaces(Map<String, String> namespaces) {
+	        namespaces.put("feature", SystemTestData.BUILDINGS.getNamespaceURI());
+	    }
```
jodygarnett: Check the developers guide for code formatting, looks like indention is off (perhaps tabs used rather than spaces?)

src/extension/wps/wps-core/src/test/java/org/geoserver/wps/other/NoArgWPS.java
tbarsballe: This file should include a copyright header, along the lines of:

```
 /* (c) 2015 Open Source Geospatial Foundation - all rights reserved
  * application directory.
  */
```
See the developers guide for more info: http://docs.geoserver.org/stable/en/developer/policies/committing.html#commit-guidelines

src/extension/wps/wps-core/src/test/java/org/geoserver/wps/other/NoArgWPSTest.java
```diff
@@ -0,0 +1,58 @@
...
+	 
+	/**
+	 * This test runs a no-argument WPS Process and checks the result. 
+	 * @throws Exception
+	 */
+	 @Test 
+	public void NoArgumentProcessTest() throws Exception
+	{
```
jodygarnett: We use the other style of brace indents, check developers guide (amounts to java standard formatting with 100 col width)

src/extension/wps/wps-core/src/test/java/org/geoserver/wps/other/NoArgWPSTest.java
tbarsballe: Missing copyright header

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/executor/WPSExecutionManager.java
```diff
@@ -391,6 +391,13 @@ private ExecuteResponseType execute() {
                 SubProgressListener executionListener = new SubProgressListener(listener,
                         inputPercentage, executionPercentage);
                 notifier.checkDismissed();
+                
+                //The system only moves from Queued to Running when the progress notifier fires (!!!).  This fires when inputs are read.
+                //   However, if there are no inputs, it never moves to running state!
+                //Since we are executing, we can move this from queued to running state. 
+                if (status.getPhase() == ProcessState.QUEUED) {
```
jodygarnett: Do you need to check if there are no parameters before invoking this logic?

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/executor/WPSExecutionManager.java
```diff
@@ -391,6 +391,13 @@ private ExecuteResponseType execute() {
                 SubProgressListener executionListener = new SubProgressListener(listener,
                         inputPercentage, executionPercentage);
                 notifier.checkDismissed();
+                
+                //The system only moves from Queued to Running when the progress notifier fires (!!!).  This fires when inputs are read.
```
jodygarnett: 100 col limit on code.

jodygarnett:
Fix looks solid, some code formatting and headers to fix up. Thanks for including test case.

I have updated Jira with a link to a related bug which you may of also just fixed.

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/executor/WPSExecutionManager.java
aaime: Please keep comment professional (in general, no exclamations, no personal opinions and the like) and current to the actual state of things, like, if this one fixes the code not moving the process to running state, then don't mention it does not move there.

aaime: I have some doubts about this one, the thing is, it reports the service as running even if the thread pools in the process manager cannot actually run it, like, it might stay waiting for an actual start for minutes with the wrong state and progress report at zero... Wondering... could it be better to add a wrapper that forcefully notifies the start of execution before delegating the run method?

jodygarnett: I think I understand, I considered recommending a "first" input parameter that is evaluated but not included in the list, so a no argument process would only have the one entry. It seemed a bit of a hack.

aaime: A pure wrapper is cleaner, we don't end up needing extra parameters, it just needs to send the right event on the listener before delegating the execution call, and it's independent on how the process manager is implemented (there are custom ones out there)

jodygarnett:
closing out

PR #1306, opened 2015-10-27
[GEOS-7282] - changed whitespace is test case

DBlasby:


jdeolive:
Thanks Dave! Great to see you back in the GeoServer world!

PR #1307, opened 2015-10-28
fix svn links in developers guide

jodygarnett:
We should probably make our own code template for GeoServer rather then requiring a manual step

### DIFF
--- a/doc/en/developer/source/eclipse-guide/index.rst
+++ b/doc/en/developer/source/eclipse-guide/index.rst
@@ -135,7 +135,7 @@ Eclipse preferences
 Code formatting
 ^^^^^^^^^^^^^^^
 
-#. Download http://svn.osgeo.org/geotools/trunk/build/eclipse/formatter.xml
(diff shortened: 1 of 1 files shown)


fgdrf:
@jodygarnett I like to suggest to backport to 2.8.x because its the current stable branch and docs are published as well : http://docs.geoserver.org/stable/en/developer/eclipse-guide/index.html

jodygarnett:
Go for it, build locally first to ensure cherry pick does not break docs.
On Wed, Oct 28, 2015 at 1:53 AM Frank Gasdorf <notifications@github.com>
wrote:

> @jodygarnett <https://github.com/jodygarnett> I like to suggest to
> backport to 2.8.x because its the current stable branch and docs are
> published as well :
> http://docs.geoserver.org/stable/en/developer/eclipse-guide/index.html
>
> —
> <https://github.com/geoserver/geoserver/pull/1307#issuecomment-151769289>.
>
-- 
--
Jody Garnett

PR #1308, opened 2015-10-28
[GEOS-6432] Include all supported WMS exception formats in GetCapabilities

bradh:
This is a pretty old bug: https://osgeo-org.atlassian.net/browse/GEOS-6432.

The PR extends current test for 1.1.1, and adds a new for 1.3.0. If there is an existing 1.3.0 test, would appreciate someone pointing it out to me.


aaime:
Looks good, merged. I believe it can be freely backported as you see fit.

PR #1309, opened 2015-10-28
[GEOS-7227] Improve null handling for Import bounds calculation

tbarsballe:
Continuation/improvement of https://github.com/geoserver/geoserver/pull/1247

When doing some downstream testing, I ran into a NPE in the vicinity of Importe.java:927. However, I was unable to reproduce this failure in a debugger.
Changes: 
* Updated the bounds status check to be more null-safe.
* Also updated the check to match string values of "true" (which can ocurr if the metadata has been saved to disk).

### DIFF
--- a/src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
+++ b/src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
@@ -918,27 +918,11 @@ public class Importer implements DisposableBean, ApplicationListener {
 
             addToCatalog(task);
             
-            //Calculate bounds, if applicable
             if (task.getLayer().getResource() instanceof FeatureTypeInfo) {
                 FeatureTypeInfo featureType = (FeatureTypeInfo) task.getLayer().getResource();
                 FeatureTypeInfo resource = getCatalog().getResourceByName(
                         featureType.getName(), FeatureTypeInfo.class);
-                if (resource.getNativeBoundingBox().isEmpty()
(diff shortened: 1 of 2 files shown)


src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -1001,8 +1002,9 @@ void doIndirectImport(ImportTask task) throws IOException {
                     // the result of a transform. there may be another way...
                     FeatureTypeInfo resource = getCatalog().getResourceByName(
                             featureType.getQualifiedName(), FeatureTypeInfo.class);
-                    if (resource.getNativeBoundingBox().isEmpty()
-                            || resource.getMetadata().get("recalculate-bounds").equals(Boolean.TRUE)) {
+                    if (resource.getNativeBoundingBox() == null || resource.getNativeBoundingBox().isEmpty()
+                            || Boolean.TRUE.equals(resource.getMetadata().get("recalculate-bounds"))
+                            || "true".equals(resource.getMetadata().get("recalculate-bounds"))) {
```
jodygarnett: Is it worth factoring out this check into a method (so you have a chance to javadoc it?) - and write a test case.

tbarsballe: Given that it is in two places, a method is probably a good idea. The functionality is already integration tested in ImportDataTest (see changes in #1247), but a more thorough test of just the method is probably a good idea.

tbarsballe:
Moved bounds calculation to its own method. Added unit test for calculateBounds method.

PR #1310, opened 2015-10-29
Fixed gdal-wcs usage example

ridethepenguin:
Added `community/gdal/usage_example.rst` to `exclude_patterns` configuration variable to avoid warning on page not being included in any toctree.


ridethepenguin:
Found a better solution, closing.

PR #1313, opened 2015-10-29
[GEOS-7286] GDAL WCS output format fails to encode gray/alpha images in Geospatial PDF

aaime:


aaime:
Applied feedback by @bradh

PR #1314, opened 2015-10-29
GEOS-4972 Allow the link attribute in a GeoRSS feed to be templated like the title and description attributes

kc7bfi:
This allows the link to be templated in GeoRSS feeds

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/georss/AtomUtils.java
+++ b/src/wms/src/main/java/org/geoserver/wms/georss/AtomUtils.java
@@ -86,22 +86,22 @@ public final class AtomUtils {
 
     //TODO: use an html based output format
     public static String getEntryURL(WMS wms, SimpleFeature feature, WMSMapContent context){
-		try {
-			return featureTemplate.link(feature);
-		} catch (IOException ioe){
-			String nsUri = feature.getType().getName().getNamespaceURI();
-			String nsPrefix = wms.getNameSpacePrefix(nsUri);
- 
-			HashMap<String,String> params = new HashMap<String,String>();
-			params.put("format", "application/atom+xml");
-			params.put("layers",  nsPrefix + ":" + feature.getType().getTypeName());
-			params.put("featureid", feature.getID());
+        try {
+            return featureTemplate.link(feature);
+        } catch (IOException ioe){
+            String nsUri = feature.getType().getName().getNamespaceURI();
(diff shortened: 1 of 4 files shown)


src/wms/src/main/java/org/geoserver/wms/georss/AtomUtils.java
```diff
@@ -86,18 +86,22 @@ public static String dateToRFC3339(Date d){
...
-        params.put("format", "application/atom+xml");
-        params.put("layers",  nsPrefix + ":" + feature.getType().getTypeName());
-        params.put("featureid", feature.getID());
-        
-        return ResponseUtils.buildURL(context.getRequest().getBaseUrl(),
-             "wms/reflect",
-             params,
-             URLType.SERVICE);
+		try {
+			return featureTemplate.link(feature);
```
aaime: Not respecting coding conventions. Please check and follow the guidelines for pull requests: https://github.com/geoserver/geoserver/blob/master/CONTRIBUTING.md#commit-guidelines

Also check the rest of the document (for example, have you already sent a contribution agreement?, did the files need to have copyright headers updated?)

aaime:
The change is in line with the rest of the templating engine, thanks for that, there are visible formatting issues though. 
I was about to write "the documentation should be updated too" but it seems we have none on the template customization to start with. Not a requirement, but we'll love if you could write some docs too.

kc7bfi:
If you can tell me where in the documentation I should create a page for GeoRSS templating I could add a page.

aaime:
I believe docs would have to go in this page:
http://docs.geoserver.org/latest/en/user/tutorials/georss/georss.html
that is, source wise, here:
https://github.com/geoserver/geoserver/blob/master/doc/en/user/source/tutorials/georss/georss.rst

aaime:
If you check the tutorials, there are pages talking about templates already that you can link/take inspiration from

kc7bfi:
GEOS-4972 Allow the link attribute in a GeoRSS feed to be templated like the title and description attributes

aaime:
Squashed into a single commit and manually merged here: https://github.com/geoserver/geoserver/commit/4a9e065d6e98cd8a83d7fdb787d492ab25889e7a

aaime:
To clarify, the changes are now merged on master. Since this is a new feature, you will be able to ask for a backport on the developer list in a month from today (assuming the patch does not cause any trouble)

PR #1316, opened 2015-10-31
[GEOS-6226] Make maintainer and onlineResource elements consistent

bradh:
This updates the various configuration files (mainly used in
GetCapabilities) to be consistent in what they specify for
<maintainer> and <onlineResource>.

The original bug report (https://osgeo-org.atlassian.net/browse/GEOS-6226) talked about invalid entries which is not the case any longer. However we still had some variation, which is made consistent in this change.

For extra value, there are a few spelling fixes too.

### DIFF
--- a/data/citewcs-1.0/services.xml
+++ b/data/citewcs-1.0/services.xml
@@ -79,7 +79,7 @@ NONE
 <keyword>WMS</keyword>
 <keyword>GEOSERVER</keyword>
 </keywords>
-<onlineResource>http://geoserver.sourceforge.net/html/index.php</onlineResource>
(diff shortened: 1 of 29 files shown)


aaime:
Hi Brad, this might be debatable, but I don't think the "maintainer" section should point to the commercial support page, that page is not representing all the contributors anyways. 
Maybe just point at geoserver.org instead?

bradh:
@aaime if you are talking about http://geoserver.org/comm then that is the communications page (i.e. mailing lists, responsible disclosure, IRC, skype).

aaime:
Oh right, "comm" is communication, and "commercial" is instead "support". Sigh, confusing page names... all right, merging

PR #1320, opened 2015-11-04
[GEOS-7295] Authkey community module breaks OpenLayers preview

ridethepenguin:
See [JIRA issue](https://osgeo-org.atlassian.net/browse/GEOS-7295) for details.

PR has no unit test because it was not trivial to write one and I wasn't sure it's needed for such a simple fix, but I can add one if required.


simboss:
We should backport to 2.7.x whenever that's possible as it's the maintenance version

ridethepenguin:
Added unit test, turned out not so difficult as I thought. I'll wait for Travis before merging.

ridethepenguin:
Whops, I had forgotten to update the copyright headers... fixed now.

PR #1321, opened 2015-11-05
[GEOS-7296] Fix read the features twice in WFS v1.1/2.0 GetFeature requests

ahuarte47:
It fixes the issue https://osgeo-org.atlassian.net/browse/GEOS-7296

The original FeatureCollection is cached to avoid that features are readed from data source (shapefiles...) each iterator from the original class in called.

Sorry, this fix haven't test. The issue is generated from mixed use of GeoServer and GeoTools.

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/CachedFeatureCollection.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/CachedFeatureCollection.java
@@ -1,0 +1,58 @@
+package org.geoserver.wfs;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+import org.geotools.data.DataUtilities;
+import org.geotools.data.simple.SimpleFeatureCollection;
+import org.geotools.data.simple.SimpleFeatureIterator;
+import org.geotools.feature.collection.AbstractFeatureCollection;
+import org.geotools.geometry.jts.ReferencedEnvelope;
+
+import org.opengis.feature.simple.SimpleFeature;
+
+/**
+ * A feature collection wrapping a base collection to cache the features managed.
+ * It avoids read from data source each an iterator is called.
+ * 
+ * @author Alvaro Huarte
+ */
+class CachedFeatureCollection extends AbstractFeatureCollection {
+    private List<SimpleFeature> cache = new ArrayList<SimpleFeature>();
+    
+    public CachedFeatureCollection(final SimpleFeatureCollection featureCollection) {
+        super(featureCollection.getSchema());
+        
+        SimpleFeatureIterator featureIterator = featureCollection.features();
+        if (featureIterator!=null) {
+            try {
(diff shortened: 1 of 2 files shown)


ahuarte47:
Ummm, I must revise the wrong tests

aaime:
Are you really loading all the data in memory?

ahuarte47:
Hi Andrea, yes. but I can change to cache the collection only the size in lesser than a maximum number of features.

What do you think ?

I have rewrite the fix, but still I break the testPostMultipleQueriesDifferentNamespaces test.

aaime:
That you have to rethinking it from the basics, check the specification, make sure you are not breaking compliance by removing the count, and if is there a way to make the count disappear, don't assume your needs match other people ones, and thus disable feature counting as an option  in WFSInfo that's not enabled by default. Please do your research, come up with a plan, and share it on geoserver-devel before proposing another pull request.

ahuarte47:
Hi Andrea, I don't want to remove the count. Only I want to avoid that a new reading task for each iterator created. This issue is especially problematic for us with big data sources.

I wanted use an idea similar to https://github.com/geoserver/geoserver/blob/master/src/wms/src/main/java/org/geoserver/wms/featureinfo/FeatureCollectionDecorator.java class does.

Thanks, I will remember your advices.

Best regards
Alvaro

aaime:
Ah, I see... ok, in that case either one makes count faster (could be done in a few ways in the shapefile module, including implementing a pure java search index similar to what shapelib does) or indeed a admin controllable memory cache (number of elements in the cache configurable) could be a way.

ahuarte47:
About of implement a pure java search index similar to what shapelib does, I am developing it using  CDX-index files (https://github.com/geotools/geotools/compare/master...ahuarte47:GEOS-6842_CdxDbaseIndex). This code is in test fase but my organization want to propose it as a new feature (It is feasible also write the use of ndx-mdx-idx index files).

Dou you remember this reverted pull?, https://github.com/geotools/geotools/pull/816#issuecomment-133702784 :-) This code replace it with none ODBC-JDBC dependency. As you can remember it avoids the bruteforce read of the shapefile to solve a pure alphaumeric filter. For big shapefiles it is very fast.

About of use a cache from a configurable maximum number of elements, Do you agree with I develop this new setting to reactivate this pull?

Best regards
Alvaro

aaime:
Hi Alvaro, good news about the filtering. Regarding this pull request, personally I have no objection, but could you first present what you want to propose on geoserver-devel so that all core devs care review it?

ahuarte47:
Ok, thanks Andrea, if you agree, before we finish the test fase.

About this pull? Can I continue doing it configurable in the WFS admin panel ?

aaime:
No, you should discuss the new configuration and memory caching on the devel list first (both need discussion, the shapefile index on the geotools-devel list, the memory caching one on the geoserver-devel list)

ahuarte47:
oh! OK, thanks Andrea!

Alvaro

PR #1322, opened 2015-11-05
GEOS-7298, remove list items associated with removed parameter filters

smithkm:
https://osgeo-org.atlassian.net/browse/GEOS-7298

### DIFF
--- a/src/web/gwc/src/main/java/org/geoserver/gwc/web/layer/ParameterFilterEditor.java
+++ b/src/web/gwc/src/main/java/org/geoserver/gwc/web/layer/ParameterFilterEditor.java
@@ -196,6 +196,9 @@ class ParameterFilterEditor extends FormComponentPanel<Set<ParameterFilter>> {
                     @Override
                     protected void onSubmit(AjaxRequestTarget target, Form form) {
                         getList().remove((ParameterFilter) getDefaultModelObject());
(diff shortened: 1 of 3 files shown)


aaime:
The test failed in the JDK8 build:

Tests run: 9, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 3.113 sec <<< FAILURE!
testRemoveFirstParamFilter(org.geoserver.gwc.web.layer.LayerCacheOptionsTabPanelTest)  Time elapsed: 75 sec  <<< FAILURE!

smithkm:
Yes, there's also another problem when deleting from more than 2 filters I'm looking into.

smithkm:
So far the best result I've gotten is that adding filters continues to work normally via ajax with changes to other filters preserved.  I can get removing filters to work without throwing errors, but it doesn't use ajax resulting in a page refresh, returning to the top of the page and any changes made to the other filters are lost.

aaime:
This one has been around for a while, any update?

smithkm:
Ended up assigned to other things and haven't had a chance to figure out how to make Wicket do what I want. This might be best to leave until the Wicket upgrade sprint.  A newer Wicket might provide more/better options.

smithkm:
I think @NielsCharlier fixed this during the Wicket update.

PR #1323, opened 2015-11-06
Feature disallow count

sampov2:
For some datasources, querying the full feature count can be very slow. This is the case with for example complex database views. To improve this, I suggest making it be possible to configure per-feature that counting the number of features can be skipped if possible in the context of the GetFeature request.

See https://osgeo-org.atlassian.net/browse/GEOS-7301


src/web/core/src/main/java/org/geoserver/web/data/resource/FeatureResourceConfigurationPanel.html
```diff
@@ -16,6 +16,16 @@
          </ul>
       </fieldset>
     </div>
+    <div>
+      <fieldset><legend><span><wicket:message key="advanced">curves</wicket:message></span></legend>
```
aaime: Curves?

src/main/src/main/java/org/geoserver/catalog/FeatureTypeInfo.java
```diff
@@ -193,4 +193,13 @@
 	
 	void setCircularArcPresent(boolean arcsPresent);
 
+
+	/**
+	 * Whether querying the full number of features is banned for this feature type.
+	 * 
+	 * @return True if a full feature count should not be attempted, false otherwise
+	 */
+	boolean isTotalCountBanned();
```
aaime: Primitive... are you sure that old configurations load fine if they don't find this element, and likewise, old REST requests? I guess so, it's probably defaulting to false, but it's important that you double check.

src/main/src/main/java/org/geoserver/catalog/FeatureTypeInfo.java
```diff
@@ -193,4 +193,13 @@
...
+
+	void setTotalCountBanned(boolean totalCountBanned);
```
aaime: Banning is sort of a strong word, what about totalCountOmitted?

src/web/core/src/main/resources/GeoServerApplication.properties
```diff
@@ -189,6 +189,8 @@ PointPanel.y  = y
 FeatureResourceConfigurationPanel.curves = Curved geometries control
 FeatureResourceConfigurationPanel.linestrings-are-curves = Linear geometries can contain circular arcs
 FeatureResourceConfigurationPanel.tolerance = Linearization tolerance (useful only if your data contains curved geometries)
+FeatureResourceConfigurationPanel.advanced = Advanced configuration
+FeatureResourceConfigurationPanel.total-count-banned = Disallow querying the full number of features
```
aaime: Disallow -> Omit

aaime:
Hi Sampo, two other issues:
* There are no tests. 
* You should verify that the XML generated in WFS 1.0, 1.1 and 2.0 resposes is still schema valid (like, I have memories that when missing in some specs one has to specify "unknown" or something similar).

src/main/src/main/java/org/geoserver/catalog/FeatureTypeInfo.java
sampov2: The primitive defaults to false which means business as usual. Updating your configuration with a new geoserver then ends up writing this value to the config and thus upgrading the configuration "naturally". I think this is the way to go.

src/main/src/main/java/org/geoserver/catalog/FeatureTypeInfo.java
sampov2: Banning is a strong word and selecting the right word is a bit problematic: using this feature potentially breaks the query interface (knowingly) as the count is not done at all, but on the other hand, because of multi-query GetFeature calls, this "ban" cannot be enforced at all times.

Would totalCountOmittedWhenPossible work for you, or is it too long and unwieldy?

PR #1327, opened 2015-11-07
Upgrading spring version to 3.2.15

jdeolive:
Pretty straight forward, no code changes. Only significant change is that some spring code that the wms module depeds on (some freemarker) stuff now lives in a new spring module, so there is a new dependency on spring-context-support required there.


jdeolive:
Any issue with me merging this one?

aaime:
No issues here, +1

jdeolive:
Thanks Andrea!

PR #1329, opened 2015-11-10
GEOS-7305 - Make KML module code compilable with Java 1.8 source

bgsmase:
Switched source and target Java versions to 1.8 from 1.7 and fixed a compilation error that occurs in KML module when this change is made. I may have gone a bit beyond scope of original bug report by switching versions, maybe this should be a separate task?

### DIFF
--- a/src/kml/src/main/java/org/geoserver/kml/icons/IconService.java
+++ b/src/kml/src/main/java/org/geoserver/kml/icons/IconService.java
@@ -8,6 +8,7 @@ package org.geoserver.kml.icons;
 import java.awt.image.BufferedImage;
 import java.io.IOException;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.logging.Level;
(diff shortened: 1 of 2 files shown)


aaime:
Hum.. yeah, we have to modify travis not to build with java 7, do the same with the linux and windows build servers, and check a full build works with java 8. I'd defer that to a time when we can coordinate such move (it seems everybody is dead busy right now?)

bgsmase:
OK, was a bit too much to switch completely now but would it be useful to add fixes for stuff that will need changing for a full switch as and when found so there is less work to do when it comes to it? (As long as it still builds under 1.7 of course). So I could remove the change of source and target in pom from this pull request but leave the changes in IconService.java... Seem like a good idea?

aaime:
Yep, it will make things easier when we can coordinate the upgrade

aaime:
Manually merged to fix the copyright header year.

PR #1331, opened 2015-11-10
GEOS-6435 - Add INSPIRE extended capabilities to WCS

bgsmase:
…ilities response

### DIFF
--- a/src/extension/inspire/src/main/java/org/geoserver/inspire/wcs/WCSExtendedCapabilitiesProvider.java
+++ b/src/extension/inspire/src/main/java/org/geoserver/inspire/wcs/WCSExtendedCapabilitiesProvider.java
@@ -1,0 +1,126 @@
+/* (c) 2014 - 2015 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2013 OpenPlans
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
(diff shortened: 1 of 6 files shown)


bgsmase:
https://osgeo-org.atlassian.net/browse/GEOS-6435

src/extension/inspire/src/main/java/org/geoserver/inspire/wcs/WCSExtendedCapabilitiesProvider.java
```diff
@@ -0,0 +1,126 @@
+/* (c) 2014 - 2015 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2013 OpenPlans
```
aaime: New file, the copyright is just "(c) 2015 - Open Source Geospatial Foundation - all rights reserved"

aaime: Oh... unless you copy/pasted it from another pre-existing provider... in that case the header would be right as is

aaime: Yeah, verified that is the case

aaime:
Ok, all good, merging. Cheers for a well done and easy to review pull request ;-)

PR #1333, opened 2015-11-12
[GEOS-7311]: Fix extraction of nullValues of a coverage

ahuarte47:
This pull fix the issue https://osgeo-org.atlassian.net/browse/GEOS-7311.

The testSingleBandedCoverage test function (https://github.com/geoserver/geoserver/blob/master/src/main/src/test/java/org/geoserver/catalog/impl/CatalogBuilderTest.java#L192) now works in non-english environments.

Alvaro

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java
+++ b/src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java
@@ -1099,7 +1099,7 @@ public class CatalogBuilder {
             if (categories != null) {
                 for (Category cat : categories) {
 
-                    if ((cat != null) && cat.getName().toString().equalsIgnoreCase("no data")) {
+                    if ((cat != null) && cat.getName().toString(Locale.ENGLISH).equalsIgnoreCase("no data")) {
                         double min = cat.getRange().getMinimum();
                         double max = cat.getRange().getMaximum();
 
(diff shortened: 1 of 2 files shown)


bradh:
Looks OK, but I wonder if its really coming from geotools, or instead perhaps the JAI code. There are some other places in geotools that do similar comparisons.

ahuarte47:
Debugging GeoServer+Geotools, this "localized" string is assigned in this line:

https://github.com/geotools/geotools/blob/master/modules/plugin/geotiff/src/main/java/org/geotools/gce/geotiff/GeoTiffReader.java#L820

Best regards
Alvaro

ahuarte47:
Hi Brad, For security I searched other "no data" texts in all GeoServer, no results.
Thanks!

aaime:
Looks good, but lacks a test. Please review the contribution guidelines before making pull requests and verify all the required and recommended bits are satisfied:

ahuarte47:
Ok, sorry, I will follow your recommendations. In this case I did not write a test case because of a exsting test function named testSingleBandedCoverage failed (Only in no-english environments).

I write other.

Thanks @aaime !

ahuarte47:
Hi @aaime, I added a test

Best regards
Alvaro

ahuarte47:
Thanks @aaime !

PR #1334, opened 2015-11-13
Patch for GEOS-4986

cippino:
Hi,
after long time, here there's the patch relate VirtualTable xml/json order.

Cheers.


cippino:
Dropped.

PR #1335, opened 2015-11-13
Layergroups: attribution & metadatalinks API+rest+getcapabilities

NielsCharlier:
See https://github.com/geoserver/geoserver/wiki/GSIP-135

https://osgeo-org.atlassian.net/browse/GEOS-5888
https://osgeo-org.atlassian.net/browse/GEOS-7313

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/impl/LayerGroupInfoImpl.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/LayerGroupInfoImpl.java
@@ -8,12 +8,14 @@ package org.geoserver.catalog.impl;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.geoserver.catalog.AttributionInfo;
 import org.geoserver.catalog.AuthorityURLInfo;
 import org.geoserver.catalog.CatalogVisitor;
 import org.geoserver.catalog.LayerGroupHelper;
 import org.geoserver.catalog.LayerGroupInfo;
 import org.geoserver.catalog.LayerIdentifierInfo;
 import org.geoserver.catalog.LayerInfo;
+import org.geoserver.catalog.MetadataLinkInfo;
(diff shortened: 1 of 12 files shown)


aaime:
The copyright header year has not been updated in any of the files. Kevin shared a script to fix the headers in all the files changed in the last commit, look for it in the devel list.

src/main/src/main/java/org/geoserver/catalog/impl/LayerGroupInfoImpl.java
```diff
@@ -388,4 +394,23 @@ public String getPrefixedName() {
     public PublishedType getType() {
         return PublishedType.GROUP;
     }
+
+    @Override
```
aaime: Hashcode and equals have not been updated to include the two new fields

src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/CapabilitiesTest.java
```diff
@@ -261,6 +264,41 @@ public void testAttribution() throws Exception {
...
+        info.setType("text/html");
+        info.setMetadataType("FGDC");
+        info.setContent("http://my/metadata/link");
+        lg.getMetadataLinks().add(info);
+        getCatalog().add(lg);
+        
+        try {
+            System.out.println(getAsString("wms?service=WMS&request=getCapabilities&version=1.1.1"));
+            Document doc = getAsDOM("wms?service=WMS&request=getCapabilities&version=1.1.1", true);
+            print(doc);
```
aaime: Handy for debugging, but nasty to have at runtime, please comment out this "print"

aaime:
Found some small issues, besides that, looks good.

NielsCharlier:
Is this good to go after fixes?

aaime:
Yep, looks good to me

PR #1337, opened 2015-11-13
GEOS-7312 - close inputstream after finishing with it

jonescc:
Fixes issue with RawData input streams not being closed when encoding RawData


src/extension/wps/wps-core/src/test/java/org/geoserver/wps/ppio/RawDataPPIOTest.java
```diff
@@ -0,0 +1,64 @@
+package org.geoserver.wps.ppio;
```
aaime: Lacks copyright header

aaime:
Looks good, there is just a minor issue with copyright headers, then we can merge it.

jonescc:
Copyright header added.

aaime:
Backported to 2.8.x and 2.7.x, ticket marked as resolved

PR #1339, opened 2015-11-16
[GEOS-7297] Fix issue where stored queries fail under load.

dpovey:
- Don't rewrite the stored query definition
- Added unit test for StoredQueryProvider


aaime:
You should update the copyright headers "last modification" year, bring it from "2014" to "2014 - 2015".
Also please squash all commits into one to make this easier to backport

PR #1340, opened 2015-11-17
Fix for GEOS-6968

mfilocha:
This is a proposed fix for: https://osgeo-org.atlassian.net/browse/GEOS-6968


aaime:
Checked, not really, if you look at the WFS settings they are same as WMS. WCS came later:

image

I believe the problem is that for WCS we have complex settings that would really need some extra title wrapping them, if not their tab... We can leave it as you suggested too, but then we'd also need to change the WFS ones, and we likely need a better title style.

mfilocha:
You're right. I use mainly raster layers so I haven't noticed that's more general problem. It's up to you then!

PR #1345, opened 2015-11-21
[GEOS-4179] Importer and monitoring REST resources are not thread-safe

aaime:
Found this one while reviewing old tickets. Not affecting much the core REST api, but unfortunately the importer and monitoring REST API are fully non thread-safe (facepalm!).

I've made sure that the BeanResourceFinder rebels against non thread safe usage and then fixed the application contexts.

If someone used BeanResourceFinder in REST plugins this change will pose backwards compatibility issues but... do we really want to allow thread unsafe usage?

### DIFF
--- a/src/extension/importer/rest/src/main/resources/applicationContext.xml
+++ b/src/extension/importer/rest/src/main/resources/applicationContext.xml
@@ -6,49 +6,49 @@
   <!--
    rest bindings
     -->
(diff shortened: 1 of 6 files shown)


aaime:
Btw, there are no visible tests, but in fact all REST test in all modules fail to start up without the related app-contex fixes.

PR #1347, opened 2015-11-25
[GEOS-7294] Allow make a Layer Group not queryable

ahuarte47:
This pull allows make a Layer Group explicitly not queryable. Currently, a layer Group is queryable if it contains at least a queryable layer, and the user can not override it.

See issues:
https://osgeo-org.atlassian.net/browse/GEOS-7294

This feature was funded by: Tracasa (http://www.tracasa.es/en)

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/LayerGroupInfo.java
+++ b/src/main/src/main/java/org/geoserver/catalog/LayerGroupInfo.java
@@ -88,6 +88,24 @@ public interface LayerGroupInfo extends PublishedInfo {
     void setMode( Mode mode );    
 
     /**
+     * Get whether the layer group is forced to be not queryable and hence can not be subject of a GetFeatureInfo request.
+     * <p>
+     * In order to preserve current default behavior (A LayerGroup is queryable when at least a
+     * child layer is queryable), this flag allows explicitly indicate that it is not queryable 
+     * independently how the child layers are configured.
+     * </p>
+     * <p>
+     * Default is {@code false}
+     * </p>
+     */
+    boolean isQueryDisabled();
+    
+    /**
+     * Set the layer group to be not queryable and hence can not be subject of a GetFeatureInfo request.
+     */
+    void setQueryDisabled(boolean queryDisabled);
+    
+    /**
      * Returns a workspace or <code>null</code> if global.
      */
     WorkspaceInfo getWorkspace();    
--- a/src/main/src/main/java/org/geoserver/catalog/impl/LayerGroupInfoImpl.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/LayerGroupInfoImpl.java
@@ -29,6 +29,7 @@ public class LayerGroupInfoImpl implements LayerGroupInfo {
     protected String id;
     protected String name;
     protected Mode mode = Mode.SINGLE;
+    protected Boolean queryDisabled;
     
     /**
      * This property in 2.2.x series is stored under the metadata map with key 'title'.
@@ -114,6 +115,16 @@ public class LayerGroupInfoImpl implements LayerGroupInfo {
         this.mode = mode;
     }
        
+    @Override
+    public boolean isQueryDisabled() {
+        return queryDisabled != null ? queryDisabled.booleanValue() : false;
+    }
+    
+    @Override
+    public void setQueryDisabled(boolean queryDisabled) {
+        this.queryDisabled = queryDisabled ? Boolean.TRUE : null;
+    }
+    
     @Override
     public String getTitle() {
         if(title == null && metadata != null) {
@@ -272,6 +283,7 @@ public class LayerGroupInfoImpl implements LayerGroupInfo {
         result = prime * result + ((identifiers == null) ? 0 : identifiers.hashCode());
         result = prime * result + ((attribution == null) ? 0 : attribution.hashCode());
         result = prime * result + ((metadataLinks == null) ? 0 : metadataLinks.hashCode());
+        result = prime * result + ((queryDisabled == null) ? 0 : queryDisabled.hashCode());
         return result;
     }
 
@@ -370,6 +382,12 @@ public class LayerGroupInfoImpl implements LayerGroupInfo {
             if (other.getMetadataLinks() != null)
                 return false;
         } else if (!metadataLinks.equals(other.getMetadataLinks()))
+            return false;
+        
+        if (queryDisabled == null) {
+            if (other.isQueryDisabled())
+                return false;
+        } else if (!queryDisabled.equals(other.isQueryDisabled()))
             return false;
         
         return true;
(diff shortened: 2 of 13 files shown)


src/wms/src/main/java/org/geoserver/wms/featureinfo/GetFeatureInfoKvpReader.java
```diff
@@ -114,23 +181,6 @@ public Object read(Object req, Map kvp, Map rawKvp) throws Exception {
                     + "It should be a proper subset of those instead");
         }
 
-        // remove no-queryable layers
-        for (MapLayerInfo l : request.getQueryLayers().toArray(new MapLayerInfo[0])) {
```
ahuarte47: Now, this verification is unnecessary, the objects are already filtered in the new GetFeatureInfoKvpRequestReader and GetFeatureInfoKvpParser classes.

src/main/src/main/java/org/geoserver/catalog/LayerGroupInfo.java
```diff
@@ -88,6 +88,19 @@ public Integer getCode() {
     void setMode( Mode mode );    
 
     /**
+     * Get whether the layer group is forced to be not queryable and hence can not be subject of a GetFeatureInfo request.
+     * <p>
+     * Default is FALSE
+     * </p>
+     */
+    Boolean isNotQueryableEnabled();
```
aaime: Make this isQueryable like in LayerInfo. Actually, you could just push up the isQueryable/setQueryable methods up from LayerInfo to PublishedInfo, which is the base interface for LayerInfo and PublishedInfo

src/main/src/main/java/org/geoserver/catalog/impl/LayerGroupInfoImpl.java
```diff
@@ -29,6 +29,7 @@
     protected String id;
     protected String name;
     protected Mode mode = Mode.SINGLE;
+    protected Boolean forceNotQueryable = Boolean.FALSE;
```
aaime: Just "queryable" like in LayerInfo (and you'll have to make opposite logic to handle null values, layer group is queryable by default).

src/web/core/src/main/java/org/geoserver/web/data/layergroup/AbstractLayerGroupPage.java
```diff
@@ -113,6 +116,9 @@ protected void onUpdate(AjaxRequestTarget target) {
         
         form.add(modeChoice);
         
+        queryableCheckBox = new CheckBox("queryable", new Model<Boolean>(!layerGroup.isNotQueryableEnabled()));
```
aaime: You should be able to directly bind the property once the layer group is property is called "queryable"

src/web/core/src/main/java/org/geoserver/web/data/layergroup/AbstractLayerGroupPage.java
```diff
@@ -262,6 +268,10 @@ public void onSubmit() {
                     lg.getStyles().add(entry.getStyle());
                 }
 
+                // update not queryable flag
+                Boolean queryable = queryableCheckBox.getModelObject();
+                if (!queryable) lg.setNotQueryableEnabled(true); else lg.setNotQueryableEnabled(null);
```
aaime: You should be able to directly bind the property once the layer group is property is called "queryable"

src/wms/src/main/java/org/geoserver/wms/kvp/MapLayerInfoKvpParser.java
```diff
@@ -33,6 +33,13 @@ public MapLayerInfoKvpParser(final String key, final WMS wms) {
         rawNamesParser = new FlatKvpParser(key, String.class);
     }
 
+    /**
+     * Returns whether the specified resource is valid in the context of the current request.
+     */
+    protected boolean passRequestedObject(Object theResource) {
```
aaime: By your description it seems a "skipResource" name would be better, but I'll leave this to native English speakers to confirm.

aaime:
Copyright headers have not been updated, please do so.

ahuarte47:
Ummm, I think all copyright headers were updated.

aaime:
They are not, see for example this pull request for examples of proper header updates: https://github.com/geoserver/geoserver/pull/1348/files

By the way, both you and @NielsCharlier are working on layer group updates with active pull requests, it's quite likely that the first one getting in will generate merge conflicts in the other.
Not sure if you want to discuss and see who goes first, otherwise I guess the only impartial rule would to be merge first the one that is ready for merge (full pass of code review) first.

src/wms/src/main/java/org/geoserver/wms/kvp/MapLayerInfoKvpParser.java
ahuarte47: Changed

src/main/src/main/java/org/geoserver/catalog/LayerGroupInfo.java
ahuarte47: I would not alter current behavior. Now, a LayerGroup is queryable if some child layer is queryable (https://github.com/geoserver/geoserver/blob/master/src/wms/src/main/java/org/geoserver/wms/WMS.java#L972). Do you prefer that I change this ?

ahuarte47:
I apologize, but I do not see what copyright headers you mean. About discuss and see who goes first, this pull or @NielsCharlier work, I adjust me to your comments.

aaime:
Doh, you are right, I was watching the existing headers in the wrong branch, the copyright headers are indeed ok.
@NielsCharlier, I'd merge this one first since it's ready and small, any objection?

aaime:
Actually no, wait, it's not ready, my other inline comments have not been addressed yet.
Uff... I need a vacation...

ahuarte47:
Do not worry, :-) I need a vacation too. About ....

_"Make this isQueryable like in LayerInfo. Actually, you could just push up the isQueryable/setQueryable methods up from LayerInfo to PublishedInfo, which is the base interface for LayerInfo and PublishedInfo"_

I tried to answer here:
https://github.com/geoserver/geoserver/pull/1347#discussion_r46120159

src/main/src/main/java/org/geoserver/catalog/LayerGroupInfo.java
aaime: Ah right, you have a point... let's call it isQueryDisabled() then,  pointing out in the javadoc that the default behavior is to respect the layers own queriable flags (isNotQueryableEnabled is not readable english).

aaime:
And oh, search the docs for the page that describes layer group editing and update it with the description of this new flag.

ahuarte47:
Changed, I hope you are comfortable with last changes. Thanks @aaime !

ahuarte47:
Thank you very much @aaime !

PR #1348, opened 2015-11-25
Refactor layer/layergroup UI - joined base with tabs for layergroup -…

NielsCharlier:
… layergroup supports attribution and metadatalinks


src/extension/geosearch/src/main/java/org/geoserver/geosearch/web/GeoSearchLayerConfigPanel.java
```diff
@@ -1,41 +0,0 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
```
aaime: This file seems to be still referred by the application context in this module?

````
<bean id="geosearchLayerGroupConfig" class="org.geoserver.web.publish.LayerGroupConfigurationPanelInfo">
    <property name="id" value="geosearchLayerGroupConfig"/>
    <property name="titleKey" value="publish.layergroup.config.geosearch"/>
    <property name="componentClass" value="org.geoserver.geosearch.web.GeoSearchLayerGroupConfigPanel"/>
  </bean>
````

File: src/web/core/src/main/java/org/geoserver/web/data/layergroup/RootLayerEntryPanel.java [hunk 2]
```diff
@@ -39,7 +39,11 @@ public RootLayerEntryPanel(String id, final Form form, WorkspaceInfo workspace)
         final TextField<LayerInfo> rootLayerField = new TextField<LayerInfo>("rootLayer") {
             @Override
             public IConverter getConverter(Class<?> type) { 
-                return form.getConverter(type);
+                if (LayerInfo.class.isAssignableFrom(type)) {
```
aaime: What's going on here?

aaime:
The diff is heavily obfuscated by generics and serialization changes but... oh well, they are going to come in handy for the Wicket upgrade I believe.

src/web/core/src/main/java/org/geoserver/web/publish/CommonPublishedConfigurationPanelInfo.java
```diff
@@ -0,0 +1,14 @@
+package org.geoserver.web.publish;
```
aaime: Bleck, no copyright header, no javadoc

src/web/core/src/main/java/org/geoserver/web/publish/CommonPublishedEditTabPanelInfo.java
```diff
@@ -0,0 +1,14 @@
+package org.geoserver.web.publish;
```
aaime: Same as above

aaime:
Sigh, too many changes for me to review in a airport, not clear minded enough, will try again later.

src/web/core/src/main/java/org/geoserver/web/data/layergroup/RootLayerEntryPanel.java
NielsCharlier: The form in question doesn't exist any longer (because the LayerGroupEditPage was refactored), and the converter creation code was moved to a more appropriate place. However, I guess the "if" construction has become obsolete

NielsCharlier:
@aaime In my re-factoring I rely on generics, and it was much easier for me to see there were no generic warnings left on my stuff if all of the @suppresswarnings stuff were gone and generics were consistently applied properly. While I was doing that I thought I might as well fix the serialisation stuff for the classes I refactored as well since Jody told me before to do that properly in the UI classes from now on

src/web/core/src/main/java/org/geoserver/web/data/layergroup/LayerGroupEditPage.java
```diff
@@ -34,38 +72,222 @@ public LayerGroupEditPage(PageParameters parameters) {
...
+            add(modeChoice);
+            
+            add(new TextField<String>("title"));
+            add(new TextArea<String>("abstract"));
+            
+            DropDownChoice<WorkspaceInfo> wsChoice = 
+                    new DropDownChoice<WorkspaceInfo>("workspace", new WorkspacesModel(), 
+                            new WorkspaceChoiceRenderer());
+            wsChoice.setNullValid(true);
+            if (!isAuthenticatedAsAdmin()) {
```
aaime: Hmmm... this change is at least non obvious. Before the ws editor was made uneditable by anyone besides a full admin, now it seems that it's left editable, but I believe, it will contain only a single possible choice? (or worse, contains two choices, leaving the current workspace or setting it to null, but won't actually allow to set it to null).
The previous behavior seemed to be at least more intuitive.

Also, since this is now both the new and edit page (a questionable change, not much because it cannot work, but because it makes these pages different to everything else in the UI), we also have potentially a different behavior in the "new" mode, where before the choice was initialized with the first workspace available to the ws specific admin, but now, nothing like that seems to be happening.

aaime: Squitinting at the diffs a bit longer (this one has been made way harder than necessary to review) it seems that this code block comes from the old AbstractLayerGroupPage, and the extra behavior in the edit/new pages has been simply lost.

src/web/core/src/main/java/org/geoserver/web/data/layergroup/LayerGroupEditPage.java
```diff
@@ -34,38 +72,222 @@ public LayerGroupEditPage(PageParameters parameters) {
...
+            
+            rootLayerPanel = new RootLayerEntryPanel("rootLayer", getPublishedInfo().getWorkspace(), myModel);
+            rootLayerPanelContainer.add(rootLayerPanel);
+
+            updateRootLayerPanel(getPublishedInfo().getMode());
+            
+            TextField<String> name = new TextField<String>("name");
+            name.setRequired(true);
+            //JD: don't need this, this is validated at the catalog level
+            //name.add(new GroupNameValidator());
```
aaime: While it's true that it's validated at the catalog level,  the GroupNameValidator provides internationalized error messages, the catalog exceptions do not... @jdeolive, do you remember if there was any other reason to comment this out?

src/web/core/src/main/java/org/geoserver/web/data/resource/ResourceConfigurationPage.java
```diff
@@ -50,49 +33,32 @@
...
+import org.geoserver.web.data.layer.LayerPage;
 
 /**
  * Page allowing to configure a layer and its resource.
  * <p>
  * The page is completely pluggable, the UI will be made up by scanning the Spring context for
- * implementations of {@link ResourceConfigurationPanel} and {@link LayerConfigurationPanel}.
+ * implementations of {@link ResourceConfigurationPanel} and {@link PublishedConfigurationPanel}.
  * <p>
  * WARNING: one crucial aspect of this page is its ability to not loose edits when one switches from
```
aaime: Can I assume this type of manual testing was carried out?

src/web/core/src/main/java/org/geoserver/web/publish/LayerEditTabPanelInfo.java
```diff
@@ -0,0 +1,14 @@
+package org.geoserver.web.publish;
```
aaime: Copyright header missing

src/web/core/src/main/resources/GeoServerApplication_ca.properties
```diff
@@ -21,23 +21,23 @@ AbstractDataAccessPage.cantGetDataStoreFactory = No s'ha trobat la factoria del
...
-AbstractLayerGroupPage.th.code                 = Codi\u0009
-AbstractLayerGroupPage.th.defaultStyle         = Estil per defecte
-AbstractLayerGroupPage.th.description          = Descripci\u00f3
-AbstractLayerGroupPage.th.layer                = Capa
-AbstractLayerGroupPage.th.position             = Posici\u00f3
-AbstractLayerGroupPage.th.remove               = Esborra
-AbstractLayerGroupPage.th.style                = Estil
-AbstractLayerGroupPage.title                   = Grup de capes
-AbstractLayerGroupPage.up                      = Mou la capa a sobre
+LayerGroupEditPage.bounds                  = L\u00edmits
```
aaime: Hum, here we have a ton of i18n keys changing name. @fgdrf will this be handled gracefully by transifex or we are going to make a mess for translators?

aaime:
Did some more review and some interactive testing. All in all the result looks ok, once the regressions have been fixed and we got some plan to handle the i18n changes (or just some assurance that we are not going to throw away a lot of work from translators).

Niels, if you fix the outstanding stuff very quick I'm going to merge it so that I have a baseline for the wicket7 upgrade branch. If I don't hear anything I'll start the wicket upgrade branch, and you can adapt and merge this one onto it later.

src/web/core/src/main/java/org/geoserver/web/data/resource/ResourceConfigurationPage.java
NielsCharlier: It has been, extensively.

src/web/core/src/main/java/org/geoserver/web/data/layergroup/LayerGroupEditPage.java
NielsCharlier: @aaime You are mistaken, there is no lost behaviour. All behaviour towards workspaces is left the same. See https://github.com/geoserver/geoserver/pull/1348/files#diff-eec1d0f43ca47e20dc28a41b97cd41beR213 and https://github.com/geoserver/geoserver/pull/1348/files#diff-eec1d0f43ca47e20dc28a41b97cd41beR79.

The purpose of this patch is to make the layergroup page more similar to the layer page, and of course share the necessary code between them. Contrary to what you are saying, the layer (resource) page was already using a single class both for new and well as existing layers. As I was porting the layergroup page to become part of the same structure I followed the same system. So I was actually adapting to existing code.
I could make a derived class for a new layergroup but I see little purpose since it only regards a tiny piece of custom code that can easily be solved with an if condition (considered the boolean isNew is there regardless).

src/web/core/src/main/java/org/geoserver/web/data/layergroup/LayerGroupEditPage.java
NielsCharlier: Only moved this, happy to change this is required.

src/web/core/src/main/resources/GeoServerApplication_ca.properties
NielsCharlier: Hmmm okay good point. This could be one reason to leave the class names intact. Let me know.

NielsCharlier:
@aaime I added the missing headers. I am at this moment still of the opinion there is no regression issue.
Please let me know if there is anything else I need to change.

src/web/core/src/main/java/org/geoserver/web/data/layergroup/LayerGroupEditPage.java
aaime: Sorry Niels, did not remember about the layer page setup, it has been setup many years ago. Also consider that I'm reviewing your pull request in spare time.
Next time plan for a review and try to minimize the changes, that will help getting a review faster, and with less misunderstandings.

src/web/core/src/main/resources/GeoServerApplication_ca.properties
aaime: I'm not sure if @fgdrf will follow up quickly on this one, if he does not, please restore the class names so that I can merge this one and move on with the Wicket upgrade sprint preparation, I only have a few hours sprinkled up to Jan 6th to do that, past it, I won't be able to work on it any more until the sprint actually begins.

fgdrf: Everything looks good! 

The resource file src/web/core/src/main/resources/GeoServerApplication.properties is synchronized from github (master) into Transifex project https://www.transifex.com/GeoServer/geoserver (once merged back this pull into master). Every Translation team gets a message that the resource has been changed. 
I would suggest to add translated Cataln properties at Transifex (makes it easier to update from Transifex) and remove changes for this pull request for src/web/core/src/main/resources/GeoServerApplication_ca.properties

However, its just a suggestion.

At the moment of writing @oscarfonts managing Catalan Translation Team, @NielsCharlier feel free to join and help improving translations.

aaime:
Hi Niels, I believe it's good for merging... except I cannot press the merge button, there are merge conflicts (I have tried to tell you in another pull request that seemed like it would have caused conflicts)

aaime:
Merged

NielsCharlier:
Thank you for your review, @aaime

ahuarte47:
Hi @NielsCharlier, the spanish interface shows bad texts. The spanish properties file seems broken.

https://github.com/geoserver/geoserver/blob/master/src/web/core/src/main/resources/GeoServerApplication_es.properties

image

NielsCharlier:
Hi Alvaro,

Something went wrong with the encoding during the merge. I have fixed it.

Thanks for letting me know.

Regards
Niels

PR #1350, opened 2015-11-25
Consolidate dependencies on proprietary JDBC drivers in profiles separate from building their data store extensions

smithkm:
Was causing build issues when the `oracle` or `sqlserver` profiles were enabled to build the data store extensions along with the `jdbcconfig` or `importer` extensions.  The latter modules should have been using the `sqlserverDriver` and `oracleDriver` profiles instead.


tbarsballe:
Looks good

PR #1351, opened 2015-11-26
Update psc list

jodygarnett:
As per nomination of @smithkm :

Alessio Fabiani
Andrea Aime +1
Ben Caradoc-Davies +1 (second)
Christian Mueller +1
Jody Garnett +1 (motion)
Jukka Rahkonen +1
Phil Scadden
Simone Giannecchini +1


smithkm:
Hmm, seems to be missing a unit test, but I suppose we can let that slide.

jodygarnett:
Updating the developers guide is the best unit test I could find for a new PSC members - you hitting the merge button *was* the unit test!

PR #1352, opened 2015-11-26
WMS GetCapabilities Mix/MaxDenominators for layer groups [GEOS-7333]

NielsCharlier:
https://osgeo-org.atlassian.net/browse/GEOS-7333

### DIFF
--- a/src/wms/src/test/java/org/geoserver/wms/capabilities/GetCapabilitiesScaleDenominatorTest.java
+++ b/src/wms/src/test/java/org/geoserver/wms/capabilities/GetCapabilitiesScaleDenominatorTest.java
@@ -4,15 +4,13 @@
  * application directory.
  */package org.geoserver.wms.capabilities;
 
-import static junit.framework.Assert.assertEquals;
-import static junit.framework.Assert.assertTrue;
+import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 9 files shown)


src/wms/src/test/java/org/geoserver/wms/capabilities/GetCapabilitiesScaleDenominatorTest.java
```diff
@@ -0,0 +1,237 @@
+/* (c) 2014 - 2015 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2013 OpenPlans
```
jodygarnett: Is this a new file? Or are you reusing existing code (and thus need the open plans reference)

jodygarnett:
This seems clear, thanks for test case.

src/wms/src/test/java/org/geoserver/wms/capabilities/GetCapabilitiesScaleDenominatorTest.java
NielsCharlier: largely copied from GetCapabilitiesScaleHintTest

NielsCharlier:
So this can be merged?

jodygarnett:
Yep, you answered my question at least :)

PR #1355, opened 2015-12-02
[GEOT-5315] Cache features under a low limit for later successive readings of data

ahuarte47:
This pull is a new proposal to fix the issue:
https://osgeo-org.atlassian.net/browse/GEOT-5315

The changes are done in geoserver code instead geotools code (https://github.com/geoserver/geoserver/pull/1355).

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/FeatureSizeFeatureCollection.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/FeatureSizeFeatureCollection.java
@@ -1,0 +1,157 @@
+/* (c) 2017 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.wfs;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.geotools.data.FeatureSource;
+import org.geotools.data.Query;
+import org.geotools.data.simple.SimpleFeatureCollection;
+import org.geotools.data.simple.SimpleFeatureIterator;
+import org.geotools.feature.FeatureIterator;
+import org.geotools.feature.collection.DecoratingSimpleFeatureCollection;
+
+import org.opengis.feature.Feature;
+import org.opengis.feature.simple.SimpleFeature;
+import org.opengis.feature.type.FeatureType;
+
+
+/**
+ * A feature collection which caches a configurable (and small) amount of features.
+ * 
+ * @author Alvaro Huarte
+ */
+class FeatureSizeFeatureCollection extends DecoratingSimpleFeatureCollection {
+    
+    /**
+     * The original feature source.
+     */
+    protected FeatureSource<? extends FeatureType, ? extends Feature> featureSource;
+    
+    /**
+     * The feature cache to manage.
+     */
+    protected List<SimpleFeature> featureCache;
+    
+    /**
+     * The original query.
+     */
+    protected Query query;
+    
+    /**
+     * Defines the maximum number of cacheable features to avoid successive readings 
+     * of the data source.
+     * <p>
+     * With a minimum overload of memory, it takes advantage of a previous reading 
+     * of features when the feature source does not support direct count of the 
+     * collection managed (e.g. shapefile stores).
+     * </p>
+     * <p>
+     * It is very useful when the feature store needs to execute costly queries 
+     * and the filter returns empty or low-count feature results. 
+     * In some contexts, WFS-GetFeature requests need to precalculate the size 
+     * of the results, and (e.g. shapefile stores) the query is executed at 
+     * least twice causing two readings of the data source.
+     * </p>
+     */
+    static int FEATURE_CACHE_LIMIT = 
+        Integer.valueOf(System.getProperty("org.geoserver.wfs.getfeature.cachelimit", "16"));
+    
+    /**
+     * Allows to programmatically set the maximum number of cacheable features.
+     */
+    public static void setFeatureCacheLimit(int featureCacheLimit) {
+        FEATURE_CACHE_LIMIT = featureCacheLimit;
+    }
+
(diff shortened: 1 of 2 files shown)


jodygarnett:
Files changed ∞ <--- what is this?

jodygarnett:
This approach sounds better (using a wrapper) - we do a similar thing for spatial index ...

Do we need a flag QueryCapabilities flag so we can selectively apply this to only things like shapefile datastore that are *slow*.

bencaradocdavies:
@ahuarte47 did you rebase your branch? This pull request seems to be broken.

ahuarte47:
@jodygarnett @bencaradocdavies, yesterday I did a rebase. I see two files changed.

ahuarte47:
Hi @jodygarnett, about:

_Do we need a flag QueryCapabilities flag so we can selectively apply this to only things like shapefile datastore that are slow._

This change takes effect only when the getCount method of the FeatureSource returns "-1" (e.g. shapefile stores). This value can act itself as that flag, do you agree?

The code also take advantage when getCount returns "0", in this case we can create safely an empty cache to avoid successive readings (regardless of type of datastore).

aaime:
@jodygarnett any more feedback on this one? There is an observation for you by @ahuarte47 in his last comment

ahuarte47:
Hi, any more feedback on this one? Did I miss anything?
Thanks!

jodygarnett:
The getCount method returning -1 just means that the answer is not quickly available from checking a file header (or index root node). It is not a flag, it is part of the api.

ahuarte47:
Hi @jodygarnett, [this](https://github.com/geoserver/geoserver/blob/master/src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java#L531) is the reason of this pull. It does that the ShapeFileDataStore (or others) reads the data twice in some cases. Initially I wrote this pull in [GeoServer side](https://github.com/geoserver/geoserver/pull/1321) but it looked better to move the code to GeoTools side.

ianturton:
what happens if the size of the underlying datastore changes between calls? Say a database gets added to in real time while we are mapping. Will this prevent us seeing new data that may have appeared?

ahuarte47:
No, this doesn't take care of store changes. The new FeatureCollection is only used in WFS GetFeature requests, Is it habitual here store be changed?

aaime:
So if I read this correctly, the cache is not static, and does not survive past the current request. It only avoids re-reading the data in the following case:

- size() is called more than one time
- features() is called after size (which is what the GML encoding does)
- the number of features read is small enough

The use case seems really really narrow, how much speedup do you get?

src/wfs/src/main/java/org/geoserver/wfs/FeatureSizeFeatureCollection.java
```diff
@@ -0,0 +1,158 @@
...
+                
+                // we have to iterate, save to cache to avoid later successive readings of data.
+                List<SimpleFeature> tempFeatureCache = new ArrayList<SimpleFeature>();
+                
+                // bean counting...
+                it = featureSource.getFeatures(query).features();
+                count = 0;
+                while (it.hasNext()) {
+                    SimpleFeature feature = (SimpleFeature) it.next();
+                    if (tempFeatureCache.size() < FEATURE_CACHE_LIMIT) tempFeatureCache.add(feature);
```
aaime: Maybe I'm mis-reading it, but working this way, if there are more than limit features, don't we get exactly FEATURE_CACHE_LIMIT In the tempFeatureCache? Which would make the code below cache it?

src/wfs/src/main/java/org/geoserver/wfs/FeatureSizeFeatureCollection.java
```diff
@@ -0,0 +1,158 @@
...
+import org.geotools.feature.FeatureIterator;
+import org.geotools.feature.collection.DecoratingSimpleFeatureCollection;
+
+import org.opengis.feature.Feature;
+import org.opengis.feature.simple.SimpleFeature;
+import org.opengis.feature.type.FeatureType;
+
+
+/**
+ * A feature collection which cache features when the feature source
```
aaime: "which caches a configurable (and small) amount of features"

ahuarte47: Thanks @aaime, I changed it

src/wfs/src/main/java/org/geoserver/wfs/FeatureSizeFeatureCollection.java
ahuarte47: tempFeatureCache will be only used when the [count is lower than FEATURE_CACHE_LIMIT](https://github.com/geoserver/geoserver/pull/1355/files#diff-3855896b9c66cd8763ca54e143f382caR140), then we can only add features to this array upto this limit. Otherwise this collection wrapper has not effect because [featureCache is null](https://github.com/geoserver/geoserver/pull/1355/files#diff-3855896b9c66cd8763ca54e143f382caR104).

ahuarte47:
> So if I read this correctly, the cache is not static, and does not survive past the current request. It only avoids re-reading the data in the following case:
> 
> size() is called more than one time
> features() is called after size (which is what the GML encoding does)
> the number of features read is small enough

It takes effect for most of WFS 2.0 GetFeature requests, because of this [code](https://github.com/geoserver/geoserver/blob/master/src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java#L549).

> The use case seems really really narrow, how much speedup do you get?

Data providers (e.g. shapefiles), which do not support "select count..." queries (features.size() does this), read all data twice. For pure alphanumeric filters, this driver reads all data because there is not alphanumeric indexing by default. Then, we get a 200% of speedup for requests without results, or with counts lower than the limit defined. It is very effective for shapefiles with many geometries.

aaime:
Replaced by #2664

PR #1356, opened 2015-12-03
[GEOS-7270] Edges of image with NoData have white artifacts when reprojected

dvntucker:
Updated the "direct" raster path to use a null background, which is just inline with how the full path does it (in StreamingRenderer.RenderRasterRequest and RenderCoverageReaderRequest). This fixes the artefacts seen when reprojecting.


src/wms/src/test/java/org/geoserver/wms/map/RenderedImageMapOutputFormatTest.java
```diff
@@ -465,4 +504,6 @@ public DummyRasterMapProducer(WMS wms) {
         }
     }
 
+
+
```
tbarsballe: Unnecessary whitespace

tbarsballe:
Overall, looks good. Could you sqash your two commits together though? It will make for a cleaner commit history.

PR #1357, opened 2015-12-04
[GEOS-7337] UTFGrid output format

aaime:


bradh:
I built it locally, and gave it a try. Obviously a useful addition, and I'd like to see it merged.

Some considerations:
 - Probably needs some documentation changes (at least list in `doc/en/user/source/services/wms/outputformats.rst`, possibly worth saying which version of UTFGrid we're attempting conformance to  - I assume 1.3).
 - Layer preview would look a little nicer with a translation entry (so you see "UTFGrid" in the "All Formats" combobox, rather than "application/json;type=utfgrid"). 
 - It would be even nicer if the Layer Preview suppressed display of UTFGrid on layers that can't make vector output (as for GML, but I don't have a clean solution). Perhaps more of a future scope change.

bradh:
Github will probably break this as a patch, but its simple enough to just copy...

```
--- a/src/web/demo/src/main/resources/GeoServerApplication.properties
+++ b/src/web/demo/src/main/resources/GeoServerApplication.properties
@@ -80,6 +80,7 @@ format.wms.openlayers=OpenLayers
 format.wms.application/openlayers=OpenLayers
 format.wms.application/x-sqlite3=MBTiles
 format.wms.application/x-gpkg=GeoPackage
+format.wms.application/json;type\u003Dutfgrid=UTFGrid
 
 # Well known WFS formats
 format.wfs.GML2=GML2
```

groldan:
@aaime this is great.
What do you think of moving this piece of code to the vectortiles module and try to combine approaches where possible? I'm not suggesting the other way around to keep main from depending on no.ecc.vectortile

aaime:
Hi @groldan, I cannot go there, the vectortiles modules is unsupported (haven't heard of any plans to make it so?), this one is going to be instead, as you can see, it's going to be part of core.
For a 2.8.x backport I'm going to make it into a community module to allow people stuck on 2.8.x to try it anyways, but it's going to be a short-lived module, 2.9.x has it in core.

The placement of the code has been discussed on the geoserver-devel list a month ago and has been confirmed during the last PSC skype meeting: 
http://osgeo-org.1560.x6.nabble.com/New-feature-UTFGrid-output-format-td5234423.html

aaime:
About suppressing the display of UTFGrid from layers that cannot support it, indeed more work is needed, we would need to extend GetMapOutputFormat with a "canHandle(PublishedInfo pi)" method that would tell us if the output can be generated... not hard to do indeed, but I'm afraid I'd get push back by mixing a UI concern with the WMS core interface. I'll pass for another day.

aaime:
Added a reference in the docs, and merged @bradh output format translation.

PR #1358, opened 2015-12-07
improvments

DBlasby:
I have made some changes to the module;

a) turned on clipping for all formats (applicationContext.xml)
     * OpenLayers handles this properly now
b) add new option (applicationContext.xml) for oversampling 
     * Results looked "blocky", setting this to "2" makes it more detailed
     * please verify that I didn't miss anything!!
c) Changed from DouglasPeuckerSimplifier to TopologyPreservingSimplifier
      * DP was not properly simplifying polygons with "thin" parts


tbarsballe:
Backported to 2.8.x. Could not backport to 2.7.x due to API changes.
Is there a GEOS ticket / bug report for this change, so it can be updated with affected / fixed versions?

aaime:
Mind, the ticket should have been created before the commits, and referred from the commit message

PR #1359, opened 2015-12-07
Backport GEOS-7116

smithkm:


smithkm:
Original PR #1219

smithkm:
I forgot that Travis doesn't do builds on PRs to stable and maintenance.  I had intended to avoid doing it locally.

PR #1362, opened 2015-12-09
Property to control  inheritance of Layer Groups in virtual services

smithkm:
Also included a bean factory utility I had intended for a related fix that turned out to be impractical.  The factory class should be useful elsewhere though so I included it.

### DIFF
--- a/src/main/src/main/java/org/geoserver/ows/LocalWorkspaceCatalogFilter.java
+++ b/src/main/src/main/java/org/geoserver/ows/LocalWorkspaceCatalogFilter.java
@@ -16,6 +16,7 @@ import org.geoserver.catalog.ResourceInfo;
 import org.geoserver.catalog.StyleInfo;
 import org.geoserver.catalog.WorkspaceInfo;
 import org.geoserver.catalog.Wrapper;
+import org.geoserver.platform.GeoServerExtensions;
 import org.geoserver.security.AbstractCatalogFilter;
 import org.geotools.filter.expression.InternalVolatileFunction;
 import org.opengis.filter.Filter;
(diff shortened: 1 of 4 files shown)


aaime:
Just a quick note, when I read "inheritance" I was thinking of properties, or something like that, the idea that it meant making global groups visible in virtual services. Maybe it's just me and a better term is indeed not coming to mind, but wanted to report my "surprise", could have the same effect to others.

src/main/src/main/java/org/geoserver/ows/LocalWorkspaceCatalogFilter.java
```diff
@@ -79,15 +80,35 @@ public boolean hideStyle(StyleInfo style) {
         return hideWorkspace(style.getWorkspace());
     }
 
+    /**
+     * Should local workspaces include layer groups from the global workspace
+     * @return
+     */
+    public static boolean workspaceLayerGroupInherit() {
+        String value = GeoServerExtensions.getProperty("GEOSERVER_GLOBAL_LAYER_GROUP_INHERIT");
```
aaime: The property cannot really be changed at runtime can it? So why not put it in a static field and avoid the repeated lookups.

aaime: Could have a package property setter for testing purposes

aaime:
Would seem like a useful property to have in GeoServerInfo + UI, but I understand that takes more time, could be added later.

aaime:
Hi Kevin, while not 100% sure I believe that while working on this pull request https://github.com/geoserver/geoserver/pull/1539 I actually made the global groups disappear from workspace specific services as a side effect.

jodygarnett:
@smithkm any update here?

bencaradocdavies:
Closing as inactive for eight months, now has merge conflicts with master, and even when it was submitted, failed to compile on Travis CI. Please revise and resubmit.

PR #1361, opened 2015-12-08
JDBC ResourceStore

NielsCharlier:
https://github.com/geoserver/geoserver/wiki/GSIP-136

PR #1363, opened 2015-12-11
Don't fail geoserver when failing to parse global configuration files.

NielsCharlier:


bencaradocdavies:
Thanks, @NielsCharlier, this looks good.

Block changes are also an opportunity for local reformatting (as always at your discretion).

bencaradocdavies:
@NielsCharlier this fix does not work because the IOException is chained in a RuntimeException. I have pushed a straightforward fix (catch Exception) in  d98e27c8958d1e6c2ef5f76d64694b75dba75e93.

bencaradocdavies:
cite-wms-1.3 on ares is now fixed:
http://ares.boundlessgeo.com/jenkins/view/testing-cite/job/cite-wms-1.3/783/

NielsCharlier:
It is not a good strategy to wrap things in RuntimeException and then 
afterwards do a catch (Exception). That is basically using Runtime 
exceptions in order to avoid using "throws" api contracts altogether, 
which exists for a good reason. RuntimeExceptions should be limited to 
fatal errors that point to a bug, not for external circumstances that 
needs to be reported such as configuration errors and problems 
reading/writing to a medium.

In this case I find it particularly strange because there is a "throws 
IOException" declared in the method anyway.

Regards
Niels

bencaradocdavies:
@NielsCharlier your position on checked exceptions is indeed the official view from Sun/Oracle. However, many (including myself) think that checked exceptions are a flaw in the design of Java. This is the great checked exception controversy. The main evidence for this is the common need for exception nesting. Of special concern is the need to nest concrete checked exceptions when passing them through abstract interfaces that cannot know beforehand how they are going to be implemented so have no throws clause or implement their own parallel exception interface. It seems to me that the better a framework separates concerns, the more likely it is to need nesting to overcome exception checking, which in my view is evidence against checked exceptions. XStreams is just one example. Dynamic languages for the JVM often nest all checked exceptions. Clojure nests all Java checked exceptions in ClojureException, a RuntimeException. Anyway, this is way off topic but very interesting!

Kind regards,
Ben.

bencaradocdavies:
And the Resource Store implementation nests almost every checked exception in an IllegalStateException! Yes, I was thinking of this discussion while reviewing #1361.

NielsCharlier:
@bencaradocdavies I hadn't thought of it that way yet, you make some valid points.
However, I see benefits to exceptions being part of the API too, most importantly that you know in advance what exactly can go wrong when you call a certain method, and you can't accidentally forget to take account of that. For example, with I/O access several things can go wrong that have to do with external circumstances, and the programmer is forced to consider these possibilities. Those kind of exceptions should never fail the program (except perhaps in a unit test), but are a way to communicate back to the caller, kind of a more advanced way of returning a success boolean. I was surprised about the design choice for Resources to use IllegalStateException instead of IOException. This in my opinions will invite many programmers to forget to consider them and they are likely to perpetuate problems higher up.

And another thing, even if we'd go for the non-enforced C++ style exception model through using subclasses of RuntimeException as a design choice, at least we should still be throwing and catching specific exceptions. In the case of resources we can do catch(IllegalStateException). I really disapprove of  catch(Exception) or catch (RuntimeException), because I've seen it happen many times that a programmer-caused bug causes something like a IllegalArgumentException, which is then somewhere caught in such a catch-all and re-represented as something that it is not at all! (Like a failure to read a file or something)

bencaradocdavies:
@NielsCharlier I agree that we should throw and catch specific exceptions. In my view this is the best way to communicate exception information to the caller. Unfortunately, Java does not permit throwing a checked exception across an API that lacks a matching throws clause. I think Java sometimes forces implementers to handle checked exceptions prematurely when the exception could be permitted to propagate to a caller better positioned to handle the exception. The response of implementers is varied when they are forced to handle a checked exception: they might nest and throw, swallow and log, or throw a new exception without nesting. In my opinion, checked exceptions were well-intended but in practice discourage the throwing and catching of specific exceptions, and so cause a loss of exception quality. One solution is to put "throws Exception" on every interface method: such a practice would be an admission that checked exceptions are in the way.

PR #1365, opened 2015-12-15
split multi-valued CQL_FILTER across layers in layer group

lisalarson:
See https://osgeo-org.atlassian.net/browse/GEOS-7344

Previous behavior parsed only one clause out of a multi-valued CQL_FILTER and applied it to each layer in the layer group, so it was not actually possible to use different filters on different layers. This change associates each clause with each layer of the group, in order.

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/map/GetMapKvpRequestReader.java
+++ b/src/wms/src/main/java/org/geoserver/wms/map/GetMapKvpRequestReader.java
@@ -400,7 +400,7 @@ public class GetMapKvpRequestReader extends KvpRequestReader implements HttpServ
                         // expand the filter on the layer group to all its sublayers
                         if (filters != null) {
                             for (int j = 0; j < layers.size(); j++) {
-                                newFilters.add(getFilter(filters, i));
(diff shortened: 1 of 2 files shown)


aaime:
Thanks, this makes it so much easier to review... and unfortunately it changes the intended behavior, causing regressions to those that have been using layer groups with filters up to today.
The comment before the code you changed is clear, it states:

`` // expand the filter on the layer group to all its sublayers ``

That is, the filter is applied to the group as a whole, and thus it is expanded on each layer in the group. 
Your change instead takes all the filters and treats them as if the group was already expanded, which is a change in behavior, but will also break existing applications that were passing just one filter for the group (getFilter will throw an exception on those).

As such, I don't believe we can accept this change as is. As a suggestion, you can subscribe on geoserver-devel and argue about getting a system variable or configuration option of sorts that would allow the behavior your desire to be enabled, leaving the current one the default one for backwards compatibility (the one you want pretty much assume the client knows how many layers are inside a layer group at a given time, which is not something one can guess, in general, unless has admin rights and can get into the REST config API to count the children).

PR #1366, opened 2015-12-15
[GEOS-7297] Fix issue where Stored Queries fail under load.

dpovey:
This is a resubmitted fix for GEOS-7297.  The original pull request is now out of date and I made several changes as a result of feedback.


aaime:
Re-kicked the build, looks like it was a transient download failure from the repos.

aaime:
Had to revert the backport to 2.8.x and 2.7.x as it breaks the build, the test is using some API that's available only on trunk. Please prepare a separate pull request for 2.8.x if you need the fix to be backported there.

PR #1369, opened 2015-12-17
test feature by grid code at row / col location

jodygarnett:
Adjust UTFGridIntegrationTest to test features using grid code at a location (rather than assuming grid code).

This is a follow up to GEOT-5322 which changed the number of featuers returned due to clipping.


aaime:
-1, the change in grid code is alerting you of a change in rendering that might be caused by a newly introduced bug, and should make you check what changed and ensure it's really not a bug.

jodygarnett:
That is fine.

PR #1370, opened 2015-12-17
GEOS-7349 GeoJSON geometry and geometry_name inconsistent on multi-geometry features

awaterme:
Pull request for proposed solution in [GEOS-7349](https://osgeo-org.atlassian.net/browse/GEOS-7349)

Created fix and new unit test to prove correctness.

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
@@ -137,7 +137,7 @@ public class GeoJSONGetFeatureResponse extends WFSGetFeatureOutputFormat {
             // execute should of set all the header information
             // including the lockID
             //
-            // execute should also fail if all of the locks could not be aquired
+            // execute should also fail if all of the locks could not be acquired
             List<FeatureCollection> resultsList = featureCollection.getFeature();
             CoordinateReferenceSystem crs = null;
             for (int i = 0; i < resultsList.size(); i++) {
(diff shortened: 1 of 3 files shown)


src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
```diff
@@ -165,10 +165,10 @@ else if ( id_option.length() != 0){
                             jsonWriter.key("id").value(value);
                         }
                         
-                        GeometryDescriptor defaultGeomType = fType.getGeometryDescriptor();
-                        if(defaultGeomType != null) {
+                        GeometryDescriptor mainGeomType = fType.getGeometryDescriptor();
```
aaime: Improper rename, we never had a concept of a "main" geometry in GeoTools feature types, but one of a "default" geometry, as made evident by the call to "feature.getDefaultGeometry()" a few lines down. 
Please revert.

src/wfs/src/test/java/org/geoserver/wfs/json/GeoJSONTest.java
```diff
@@ -480,6 +488,67 @@ public void testGetFeatureCRS() throws Exception {
         assertThat(aCRS, encodesCRS(expectedCrs));
     }
     
+    @Test
+    public void testGeometryAndGeometryNameConsistency() throws Exception {
+        JSONObject collection = (JSONObject) getAsJSON(
+                "wfs?request=GetFeature&version=1.0.0&typename="
+                        + getLayerId(MULTI_GEOMETRIES_WITH_NULL) + "&outputformat=" + JSONType.json);
+         print(collection);
```
aaime: Don't leave un-commented print outs

aaime:
I can see show this is one way to solve the inconsistency... it's however not one that I would have chosen. The current approach results in a structure where the geometry type and meaning can change from feature to feature (e.g., think a case where one has polygon and label point, and for some reason, a few features only have a label point).

According to the spec: 'A feature object must have a member with the name "geometry". The value of the geometry member is a geometry object as defined above or a JSON null value.'

So a null geometry seems to be a legit value, I'd rather keep structure constant and just make the other geometries always be "normal" attributes. But maybe it's just me, it's a topic that might warrant discussion on geoserver-devel (core devs should also see the comment here, but they are generally a bit shy to comment on pull requests...)

PR #1372, opened 2015-12-17
[GEOS-7351] Add prototype scopes for REST beans

nmco:


aaime:
Reviewed offline before the pull request was made, good to go

PR #1374, opened 2015-12-17
[GEOS-7352] Error rendering coverage with transformation when out of bounding box

sikeoka:
Fix for:
[GEOS-7352] Error rendering coverage with transformation when out of bounding box

This can be backported to 2.8.x and 2.7.x.


aaime:
Yep, makes sense. I'll merge once I have time to also do the backports

PR #1375, opened 2015-12-17
[GEOS-7353] Make GetFeatureInfo HTML template loading thread-safe.

sikeoka:
Fix for:
[GEOS-7353] GetFeatureInfo HTML template loading is not thread-safe

Can be backported to 2.8.x and 2.7.x.

### DIFF
--- a/src/main/src/main/java/org/geoserver/template/GeoServerTemplateLoader.java
+++ b/src/main/src/main/java/org/geoserver/template/GeoServerTemplateLoader.java
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2015 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
@@ -83,7 +83,7 @@ public class GeoServerTemplateLoader implements TemplateLoader {
      * Feature type directory to load template against. Its presence is mutually
      * exclusive with coverageName
      */
-    ResourceInfo resource;
+    protected ResourceInfo resource;
     
     /**
      * Feature type directory to load template against. Its presence is mutually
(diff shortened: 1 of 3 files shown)


src/main/src/main/java/org/geoserver/template/GeoServerTemplateLoader.java
```diff
@@ -83,7 +83,7 @@
      * Feature type directory to load template against. Its presence is mutually
      * exclusive with coverageName
      */
-    ResourceInfo resource;
+    protected ResourceInfo resource;
```
aaime: Is this change actually used anywhere?

src/wms/src/main/java/org/geoserver/wms/featureinfo/HTMLFeatureInfoOutputFormat.java
```diff
@@ -68,7 +68,7 @@ public TemplateModel wrap(Object object) throws TemplateModelException {
         });
     }
 
-    GeoServerTemplateLoader templateLoader;
+    volatile GeoServerTemplateLoader templateLoader;
```
aaime: Normally I see volatile being used along with a double checked locking pattern, but here it does not seem the case. Wasn't moving the loader init in the synchronized block enough?

src/main/src/main/java/org/geoserver/template/GeoServerTemplateLoader.java
aaime: Ah, seems to be used in the tests

src/wms/src/main/java/org/geoserver/wms/featureinfo/HTMLFeatureInfoOutputFormat.java
sikeoka: Yes, you are correct.  The volatile modifier is not necessary now that it is in the synchronized block.  I will remove it from the pull request.

sikeoka:
I updated the pull request to remove the unnecessary volatile modifier.

aaime:
Manually merged to squash the two commits into one, and then backported to 2.8.x and 2.7.x

PR #1381, opened 2015-12-18
Duplicated isGmlPrefixing and setGmlPrefixing?

cgore:
These appear to be unintentionally duplicated identically, removing the second copy.


aaime:
Yep, looks like a leftover indeed. Thank you!

PR #1382, opened 2015-12-18
Add scalable font size example

riverstream:
Add scalable font size example for TextSymbolizer


aaime:
Yep, nice example. Mind one thing, one could get a similar effect with less effort by setting the size in real world unit of measure, except it will become too small at lower scales (high scale denominator) and giant at high scale (low scale denominator)

riverstream:
Thanks.  Correct, I did try using the real world unit of measure and ran into the exact issue you mentioned.  Hopes this helps others.

PR #1384, opened 2015-12-20
[GEOS-7358] Fix: use ResourceAccessManager pluggable authorization in UI.

etj:
Authorization architecture allows ResourceAccessManager to be pluggable,
but the WorkspaceAdminComponentAuthorizer class forces the use of the default access manager.


aaime:
Going to merge manually after adding a test that check the pluggability is actually respected (one that that would fail, if you did not fix this issue).

PR #1386, opened 2015-12-31
[GEOS-7281] move input-less processes to RUNNING state during execution

aaime:
Look ma, an insurmountable problem fixed!

### DIFF
--- a/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/executor/ProcessStartupFilter.java
+++ b/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/executor/ProcessStartupFilter.java
@@ -1,0 +1,81 @@
+/* (c) 2015 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.wps.executor;
+
+import java.util.Map;
+import java.util.logging.Logger;
+
+import org.geoserver.platform.ExtensionPriority;
+import org.geoserver.wps.process.DelegatingProcessFactory;
+import org.geoserver.wps.process.ProcessFilter;
+import org.geotools.process.Process;
+import org.geotools.process.ProcessException;
+import org.geotools.process.ProcessFactory;
+import org.geotools.util.DelegateProgressListener;
+import org.geotools.util.logging.Logging;
+import org.opengis.feature.type.Name;
+import org.opengis.util.ProgressListener;
+
+/**
+ * A process filter making sure the {@link ProgressListener#started()} method is called upon execution
+ * no matter if the process has inputs or not
+ * 
+ * @author Andrea Aime - GeoSolutions
+ *
+ */
+public class ProcessStartupFilter implements ProcessFilter, ExtensionPriority {
+
+    public class ProcessStartupWrapper implements Process {
+        
+        Process delegate;
+        
+        public ProcessStartupWrapper(Process delegate) {
+            super();
+            this.delegate = delegate;
+        }
+
+        @Override
+        public Map<String, Object> execute(Map<String, Object> input, ProgressListener monitor)
+                throws ProcessException {
+            if(monitor != null) {
+                monitor.started();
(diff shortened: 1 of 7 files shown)


bradh:
Can't say I fully understand the problem (beyond what is in the JIRA ticket: https://osgeo-org.atlassian.net/browse/GEOS-7281), but it looks like a pretty elegant solution.

aaime:
Thanks @bradh, I should have addressed your feedback

bradh:
:+1: 
Recommend merge.

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/executor/ProcessStartupFilter.java
```diff
@@ -0,0 +1,81 @@
...
+
+        @Override
+        public Map<String, Object> execute(Map<String, Object> input, ProgressListener monitor)
+                throws ProcessException {
+            if(monitor != null) {
+                monitor.started();
+                monitor = new DelegateProgressListener(monitor) {
+                    @Override
+                    public void started() {
+                        // do not pass over, we already called "started"
```
jodygarnett: Okay so started is called *for sure*, even if the delegate does not other to use the monitor. This seems a complicated way to introduce a change to how process progress is recorded.

Is there any reason we should not use this approach *all* the time? Like have the engine call *started* and pass the monitor into the execute function already started.

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/executor/ProcessStartupFilter.java
```diff
@@ -0,0 +1,81 @@
...
+    class ProcessStartupFactory extends DelegatingProcessFactory {
+
+        public ProcessStartupFactory(ProcessFactory delegate) {
+            super(delegate);
+        }
+        
+        @Override
+        public Process create(Name name) {
+            Process process = delegate.create(name);
+            return new ProcessStartupWrapper(process);
```
jodygarnett: So the ProcessStartupWrapper is used in all cases, not only for no argument processes.

Is this problem due to no argument processes, or just a problem with process implementations that do not honor the API contract (by calling started() on the provided monitor).

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/executor/ProcessStartupFilter.java
aaime: @jodygarnett I don't understand what you're suggesting, could you be more specific and show where you would place the "started" call?
For example, at first I placed it in the DefaultProcessMananger, which was simpler, but that would not have worked for custom process managers.

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/executor/ProcessStartupFilter.java
aaime: By any process implementation that does not care to check its inputs (at the low level, the API taking a Map).. Could have limited it to no-args processes, but why bother?

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/executor/ProcessStartupFilter.java
jodygarnett: Chatting here with Dave and am feeling stupid for not noticing it earlier.

The Process execute method contract asks that the monitor be used (my assumption is started, progress, progress, ... done notifications).

The reflection based process implementation is willing to pass in the monitor for developers to meet this API contract (if the last argument is a monitor). For a no-arguments process (or a process that does not accept a monitor argument) I suspect I have a GeoTools bug where "the monitor" is not being correct notified on execute start and execute completion. 

I *like* your delegate approach (of intercepting the started method) and would like to steal it for a GeoTools fix.

aaime: Steal as you see fit :-)

PR #1387, opened 2016-01-01
Wicket 7 conversion

bradh:
POM conversion per advice at https://cwiki.apache.org/confluence/display/WICKET/Migration+to+Wicket+1.5#MigrationtoWicket1.5-Mavenusersarerecommendedtodependonorg.apache.wicket:wicket:pomfromnowon

Convert over usage of AbstractValidator to IValidator. Those are untested, but do at least fix compilation problems.

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/data/store/WMSStoreNewPage.java
+++ b/src/web/core/src/main/java/org/geoserver/web/data/store/WMSStoreNewPage.java
@@ -8,14 +8,15 @@ package org.geoserver.web.data.store;
 import java.io.IOException;
 import java.net.URL;
 import java.util.Collections;
-import java.util.List;
 import java.util.logging.Level;
 
 import javax.management.RuntimeErrorException;
 
 import org.apache.wicket.ajax.AjaxRequestTarget;
 import org.apache.wicket.validation.IValidatable;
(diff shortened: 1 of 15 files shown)


src/web/core/src/main/java/org/geoserver/web/data/store/WMSStoreNewPage.java
```diff
@@ -8,14 +8,15 @@
 import java.io.IOException;
```
aaime: Let's not forget to update copyright headers. With 2016 now :-)

aaime:
Thanks, looks good. About the headers hum... unsure, I guess we are going to modify all those files once more again anyways, we might just fix the headers during the sprint.

src/web/core/src/main/java/org/geoserver/web/data/store/WMSStoreNewPage.java
bradh: I'm sure I did those changes last year...

aaime: Yeah, this file still states only 2014. Anyways, as said, no big deal, it's quite likely that we are going to have to modify these files further during the sprint.

aaime:
Merged :)

PR #1390, opened 2016-01-05
[GEOS-7349] value of default geometry always as GeoJSON geometry...

awaterme:
[GEOS-7349] value of default geometry always as GeoJSON geometry, never within properies

New pull request for [GEOS-7349](https://osgeo-org.atlassian.net/browse/GEOS-7349), after discussion on GS mailing list.

PR removed loop for finding alternative geometry if default geometry is null.

Also fixes writing "null" within properties for default geometry.

Corrected minor typos.

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
@@ -1,5 +1,5 @@
 /* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
- * (c) 2001 - 2016 OpenPlans
+ * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
  */
@@ -216,7 +216,7 @@ public class GeoJSONGetFeatureResponse extends WFSGetFeatureOutputFormat {
                                 } else if(value == null){
                                     jsonWriter.key(ad.getLocalName());
                                     jsonWriter.value(null);
-                                }else {
+                                } else {
                                     jsonWriter.key(ad.getLocalName());
                                     jsonWriter.writeGeom((Geometry) value);
                                 }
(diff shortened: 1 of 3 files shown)


src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
```diff
@@ -1,5 +1,5 @@
-/* (c) 2014 - 2015 Open Source Geospatial Foundation - all rights reserved
- * (c) 2001 - 2013 OpenPlans
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2016 OpenPlans
```
aaime: Nope, this one has to stay up to 2013, it was the previous copyright owner

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
```diff
@@ -212,30 +202,27 @@ else if ( id_option.length() != 0){
...
+                                // If it's the default that you already
+                                // printed above, so you don't need it here.
+                                if (ad.equals(defaultGeomType)) {
+                                    // Do nothing, we wrote it above
+                                    // jsonWriter.value("geometry_name");
+                                } else if(value == null){
                                     jsonWriter.key(ad.getLocalName());
-                                    jsonWriter.value(value);
+                                    jsonWriter.value(null);
+                                }else {
```
aaime: Formatting

src/wfs/src/test/java/org/geoserver/wfs/json/GeoJSONTest.java
```diff
@@ -1,5 +1,5 @@
-/* (c) 2014 - 2015 Open Source Geospatial Foundation - all rights reserved
- * (c) 2001 - 2013 OpenPlans
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2016 OpenPlans
```
aaime: Same as the other file, should stay 2013

aaime:
Requires minor fixes to copyright headers and formatting, besides that, looks good to me.

awaterme:
Fixed copyrights headers and formatting

aaime:
Manually merged to squash the two commits into one, and backported to 2.8.x and 2.7.x

PR #1391, opened 2016-01-06
[GEOS-7366] Fixed a WPS file handle leak.

sikeoka:
LazyInputMap fires a ProcessEvent before it begins loading the inputs and another ProcessEvent after each input is loaded.  The current CoverageResourceListener will only check the inputs on the first event, where all of the inputs will be null.  This can result in leaking file handles.

This pull request updates CoverageResourceListener to check for non-null values.


aaime:
Looks good. I'll merge once I also have time to backport the fix.

aaime:
Merged and backported to 2.8.x and 2.7.x

PR #1392, opened 2016-01-07
[GEOS-2613] GeoJSON support for setting decimal places.

dpovey:
This change adds support for limiting the number of decimal places used
for doubles when outputting GeoJSON responses to WFS queries.  Unit tests
are also added for this functionality.

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
@@ -6,6 +6,7 @@
 package org.geoserver.wfs.json;
 
 import java.io.Writer;
+import java.math.BigDecimal;
 import java.util.Calendar;
 import java.util.logging.Logger;
 
@@ -13,6 +14,7 @@ import net.sf.json.JSONException;
 import net.sf.json.util.JSONBuilder;
 
 import org.geotools.geometry.jts.coordinatesequence.CoordinateSequences;
+import org.geotools.gml.producer.CoordinateWriter;
 import org.geotools.referencing.CRS;
 import org.geotools.util.Converters;
 
@@ -39,10 +41,13 @@ import com.vividsolutions.jts.geom.impl.CoordinateArraySequence;
  *
  */
 public class GeoJSONBuilder extends JSONBuilder {
+
     private final Logger LOGGER = org.geotools.util.logging.Logging
     .getLogger(this.getClass());
     
     private CRS.AxisOrder axisOrder = CRS.AxisOrder.EAST_NORTH;
+
+    private int numDecimals = 6;
 
     public GeoJSONBuilder(Writer w) {
         super(w);
@@ -51,8 +56,8 @@ public class GeoJSONBuilder extends JSONBuilder {
     /**
      * Writes any geometry object.  This class figures out which geometry representation to write
      * and calls subclasses to actually write the object.
-     * @param geometry The geoemtry be encoded
-     * @return The JSONBuilder with the new geoemtry
+     * @param geometry The geometry to be encoded
+     * @return The JSONBuilder with the new geometry
      * @throws JSONException If anything goes wrong
      */
     public JSONBuilder writeGeom(Geometry geometry) throws JSONException {
@@ -159,14 +164,14 @@ public class GeoJSONBuilder extends JSONBuilder {
     private JSONBuilder writeCoordinate(double x, double y, double z) {
         this.array();
         if(axisOrder==CRS.AxisOrder.NORTH_EAST){
-            this.value(y);
-            this.value(x);
+            roundedValue(y);
+            roundedValue(x);
         } else {
-            this.value(x);
-            this.value(y);
+            roundedValue(x);
+            roundedValue(y);
         }
         if(!Double.isNaN(z)) {
-            this.value(z);
+            roundedValue(z);
         }
 
         return this.endArray();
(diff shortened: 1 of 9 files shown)


src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
```diff
@@ -298,11 +302,18 @@ public GeoJSONBuilder value(Object value) {
         if (value instanceof java.util.Date || value instanceof Calendar) {
             value = Converters.convert(value, String.class);
         }
+        if (value instanceof Double) {
```
aaime: "else if" maybe?

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
```diff
@@ -91,8 +91,10 @@ public String getMimeType(Object value, Operation operation) throws ServiceExcep
 
     @Override
     protected void write(FeatureCollectionResponse featureCollection, OutputStream output,
-            Operation describeFeatureType) throws IOException {
-
+            Operation describeFeatureType) throws IOException { 
+    	
+    	int numDecimals = getNumDecimals(featureCollection.getFeature(), gs, gs.getCatalog());
```
aaime: The indenting seems off here?

aaime:
Generally speaking it looks good, minor issues here and there. 
Question: so by default, if you don't change anything in a release data directory, you'd get 8 decimals like in WFS 1.0?

dpovey:
Yes that's the case. I tried to keep it consistent with the approach taken for GML. I'll make those changes you suggest and resubmit.

Thanks!
Dean.

dpovey:
Updated with suggested changes.  I found a few more cases where tabs were used instead of spaces and fixed those (at least the ones that were in the code I changed).

dpovey:
Updated with suggested changes. I found a few more cases where tabs were used instead of spaces and fixed those (at least the ones that were in the code I changed).

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
```diff
@@ -297,11 +301,17 @@ public static int getGeometryType(Geometry geometry) {
...
+        } else if (value instanceof Double) {
+            value = roundToNumberOfDecimalPlaces((Double)value);
         }
         super.value(value);
         return this;
     }
-    
+
+    private double roundToNumberOfDecimalPlaces(Double value) {
+        return new BigDecimal((Double)value).setScale(numDecimals, BigDecimal.ROUND_HALF_UP).doubleValue();
```
aaime: Looking at this one again, this is a massive generator of garbage, by allocating a large object as BigDecimal to format each and every ordinate. While I haven't tried, I believe this is going to affect performance significantly.
The GML2 output format got a nice speedup when we optimized its decimal places handling, see here: 
https://github.com/geotools/geotools/blob/master/modules/library/main/src/main/java/org/geotools/gml/producer/CoordinateWriter.java#L295

aaime: For best performance, we could also dodge the jsonlib creation of a Double wrapper, override value(double d) and make it call directly append(formattedStringValue) instead.

As a separate observation, your change has the side effect of also formatting attributes that happen to be doubles, which we might need to preserve with full accuracy. Probably best to change writeCoordinate to call onto an optimized writeOrdinate method that formats the double the efficient way, and then appends a string

dpovey: Okay, I did consider that, but wanted to keep the change minimal.

I note that the code you reference does not correctly round negative values
(or at least it is not consistent with any standard rounding scheme) it
probably should be:

x = (x < 0) ? -Math.ceil(-x * scale - 0.5) / scale : Math.floor(x * scale +
0.5) / scale;

Which gives you the "Round Half Up" mode.

Using BigDecimal was designed to handle the precision loss. You can find a
small percentage of examples which fail when you round to 8 decimals, e.g:

6251.334203214999 rounded to 8 decimal places gives 6521.33420322 instead
of 6251.33420321

These will be rare.  There are some cases where even BigDecimal gives the
wrong answer here so it is probably something you can live with.

I will make a change to do this in primitive doubles as above if you are
happy with those caveats.

dpovey: append(formattedStringValue) is private in json-lib but the bare writer is
protected, so I would have to call that directly.  That strikes me as a
little fragile, but I can do it if you insist.

Also json-lib handles removing trailing zeros so we would have to duplicate
that code to be consistent.

My suggestion would be to just change the writeCoordinate to do the
rounding but still call super.value() on the resulting double.  The Double
wrapper is going to be much less significant than the StringBuffer
allocation and the other work that is going on there.

dpovey: Ignore that, I realise now that -Math.ceil(-x * scale - 0.5) ==
Math.floor(x * scale + 0.5)!

dpovey: I settled on this implementation after a bunch of testing. It handles the
precision failures as well as can be expected.  Will need to submit a new
pull request with the complete checkin including tests:

dpovey: /* (c) 2016 Open Source Geospatial Foundation - all rights reserved
 * application directory.
 */
package org.geoserver.wfs.json;

/**
 * Utility class for rounding double values.
 *
 * @author Dean Povey
 *
 */
public class RoundingUtil {
    // How to scale the double, indexed by the number of digits
    private static double[] SCALE = {
            1d,
            10d,
            100d,
            1000d,
            10000d,
            100000d,
            1000000d,
            10000000d,
            100000000d
        };

    /**
     * Round a value to the specified number of decimal places using the
"Round Half Up" strategy.
     *
     *
     * <p>
     * Special cases:
     * <ul>
     * <li>NaN is returned as NaN.<li>
     * <li>+/-Infinity are returned as +/-Infinity.
     * </ul>
     *
     * @param value The value to round
     * @param numDecimals The number of decimal places to round to.
     *
     * @return The value rounded to the specified number of decimals
     */
    public static double round(double value, int numDecimals) {
        // Technically this code will handle -numDecimals by rounding
digits to the left of the decimal point, but that is
        // probably a use case that is not really needed in practice.

        double scale = (numDecimals < 8) ? SCALE[numDecimals] :
Math.pow(10, numDecimals);

        // Prevent us exceeding the maximum precision.  We make sure the
minimum spacing between values is sufficient
        // for the given scale otherwise just return the value.
        if (Math.ulp(value) * scale > 1d) return value;

        return Math.floor(value * scale + 0.5) / scale;
    }

}

aaime:
Hi, looks like you have mixed in some changes from another pull request, see changes to RenderedImageMapOutputFormat and TinyRasterBoundingBoxTest (both of which already landed on master).

dpovey:
Yeah, I am going to have to start again on a new branch. Sorry :-(.

dpovey:
New pull request: #1394

PR #1393, opened 2016-01-09
Do not clip geometries at the tile boundary

ahocevar:
In the future, we should apply the clipping with a buffer around the tile. For lines and polygons, a few pixels are enough. Points also need to be included in neighbouring tiles, taking into account the symbol size and potential label sizes.


DBlasby:
Please do not merge this - I have a more indepth patch ready to push.  I'll push it monday morning.

bencaradocdavies:
Superseded by: https://github.com/geoserver/geoserver/pull/1396

PR #1394, opened 2016-01-11
[GEOS-2613] GeoJSON support for setting decimal places.

dpovey:
This change adds support for limiting the number of decimal places used
for doubles when outputting GeoJSON responses to WFS queries.  Unit tests
are also added for this functionality.

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
@@ -8,9 +8,6 @@ package org.geoserver.wfs.json;
 import java.io.Writer;
 import java.util.Calendar;
 import java.util.logging.Logger;
-
-import net.sf.json.JSONException;
-import net.sf.json.util.JSONBuilder;
 
 import org.geotools.geometry.jts.coordinatesequence.CoordinateSequences;
 import org.geotools.referencing.CRS;
@@ -29,6 +26,9 @@ import com.vividsolutions.jts.geom.Point;
 import com.vividsolutions.jts.geom.Polygon;
 import com.vividsolutions.jts.geom.impl.CoordinateArraySequence;
 
+import net.sf.json.JSONException;
+import net.sf.json.util.JSONBuilder;
+
 
 /**
  * This class extends the JSONBuilder to be able to write out geometric types.  It is coded
@@ -39,10 +39,13 @@ import com.vividsolutions.jts.geom.impl.CoordinateArraySequence;
  *
  */
 public class GeoJSONBuilder extends JSONBuilder {
+
     private final Logger LOGGER = org.geotools.util.logging.Logging
     .getLogger(this.getClass());
     
     private CRS.AxisOrder axisOrder = CRS.AxisOrder.EAST_NORTH;
+
+    private int numDecimals = 6;
 
     public GeoJSONBuilder(Writer w) {
         super(w);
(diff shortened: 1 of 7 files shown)


aaime:
Thanks. One thing I don't remember, did you already sign and send a OSGeo contribution agreement, according to our contribution guide? 
This pull request is large enough to require one: 

aaime:
There are test failures in gt-wfs on travis, could you have a look?

jodygarnett:
wonder why it broke, did you leave the number of decimal places setting changed at the end of your test?

dpovey:
Done.

dpovey:
No my tests were borked locally due to another checkin so I missed that I
hadn't converted the bbox correctly.  I have committed a fix for that and
assuming that works I'll squash that commit together with the previous one.

dpovey:
CI tests passed. I squashed these two commits together so once the next build passes it should be good to go.

jodygarnett:
Thanks, happy to merge this in.

stefan0722:
Is there any documentation how to configure the numDecimals? If this is not configurable this is a breaking change because 6 digits means 4 decimal places or 30m in EPSG:4326 which is too inaccurante for many usecases. The name is also misleading, the RoudingUtils algorithm takes the whole double value into accout and not only the decimal places. This means that you get 4 or 5 decimal places which depends on the digits before the comma.

aaime:
The code seems to talk about decimals, not digits. 6 decimals is like 0,1m precision (assuming roughly 100km per degree at the equator), which I agree might not be enough for high precision local mapping (8 decimals would be better).
I just checked a GeoJSON request against a 2.9.x, it has indeed 6 decimals (and 7 or 8 digits):

http://demo.geo-solutions.it/geoserver/tiger/ows?service=WFS&version=1.0.0&request=GetFeature&typeName=tiger:tiger_roads&maxFeatures=50&outputFormat=application%2Fjson

Wondering where your comment is coming from? Are you experiencing any specific problem and traced it back to this pull request?

aaime:
Further quick checks seem to suggest the 6 is not actually used, the decimal count is picked from the geoserver configuration (wfs config, with an override per layer). And indeed if I pick another layer I get 8 decimals, which is the default for wfs:
http://demo.geo-solutions.it/geoserver/geosolutions/ows?service=WFS&version=1.0.0&request=GetFeature&typeName=geosolutions:regioni&maxFeatures=1&outputFormat=application%2Fjson

stefan0722:
Maybe my assumption was too fast but with the same data directory and upgrading vom 2.8.3 to 2.9 I get two complete different responses:

2.8.3: 
 {
      "type": "Feature",
      "id": "vw_object_address_geoms.fid--685c38b5_15608542126_-7ffd",
      "geometry": {
        "type": "Point",
        "coordinates": [
          13.030025366385066,
          47.795177712424
        ]
      },
      "geometry_name": "geometry",
      "properties": {
        "id": 5008,
        "type_id": 3       
      }
    }

2.9:
{
      "type": "Feature",
      "id": "vw_object_address_geoms.fid--10e2c24b_156081c8390_-7afa",
      "geometry": {
        "type": "Point",
        "coordinates": [
          13.03,
          47.7952
        ]
      },
      "geometry_name": "geometry",
      "properties": {
        "id": 5008,
        "type_id": 3
      }

The last one is very inaccurate. Maybe there is a migration of properties in the data_dir necessary?

stefan0722:
Now I found the problem. Geoserver 2.8.3 wrote in the global.xml <numDecimals>4</numDecimals> by default but it had no effect. After upgrading to 2.9 this option suddendly had an effect. I changed it to 8 and now it works like expected (with 8 decimal places). Maybe this global numDecimals parameter could be added to the documentation somehow. Thanks for your time and help!

dpovey:
This change aligns the numDecimals parameter in the JSON output to the usage in GML so they are the same.  This setting is documented in the global settings:

[http://docs.geoserver.org/2.9.0/user/configuration/globalsettings.html](url)

_**Number of Decimals**
Refers to the number of decimal places returned in a GetFeature response. Also useful in optimizing bandwidth. Default is **8**._

It does seem that default should be changed to align with the doc if that's the case.

aaime:
Yes, looks like the default value if the data dir is empty or does not have a global.xml file is 4: https://github.com/geoserver/geoserver/blob/master/src/main/src/main/java/org/geoserver/config/impl/SettingsInfoImpl.java#L28

PR #1395, opened 2016-01-11
[GEOS-7370] Fix use of LocalWorkspaceIncludesPrefix setting

ahuarte47:
The LayerGroup object ignores its LocalWorkspaceIncludesPrefix setting when the object is fetched from "catalog.getLayerGroups()" method. It works ok when the LayerGroup object is fetched from "catalog.getLayerGroupByName('x')" method.

This PR fixes the issue:
https://osgeo-org.atlassian.net/browse/GEOS-7370

This feature was funded by: Tracasa (http://www.tracasa.es/en)

### DIFF
--- a/src/main/src/test/java/org/geoserver/catalog/impl/LocalWorkspaceLayersTest.java
+++ b/src/main/src/test/java/org/geoserver/catalog/impl/LocalWorkspaceLayersTest.java
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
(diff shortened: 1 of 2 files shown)


src/main/src/test/java/org/geoserver/catalog/impl/LocalWorkspaceLayersTest.java
```diff
@@ -59,6 +61,16 @@ public void testGroupLayerInWorkspace() {
         assertNotNull(catalog.getLayerGroupByName("localGroup"));
         assertEquals(1, catalog.getLayerGroups().size());
         assertEquals("localGroup", catalog.getLayerGroupByName("localGroup").prefixedName());
+        
+        GeoServer gs = getGeoServer();
+        SettingsInfo settings = gs.getFactory().createSettings();
+        settings.setLocalWorkspaceIncludesPrefix(true);
```
jodygarnett: Is this setting already available? Not sure I saw it mentioned in the docs...

ahuarte47: Yes, there is a CheckBox in the workspace web-admin panel. Other similar wrap methods in the LocalWorkspaceCatalog class already use this flag. Without this fix you can get a GetCapabilites XML document with layers without prefix even with flag activated.

ahuarte47: Hi @jodygarnett, ....

image

aaime:
Looks good, merging

ahuarte47:
Thanks @aaime

PR #1397, opened 2016-01-12
[GEOS-7372] Allow default value for time and elevation dimensions to be a range

aaime:
Places that can now return a value or a range of values have been switched to return "Object", which matches the practice of the rest of the WMS dimension handling when both point and range values are admitted. One could create a Value<T> that can contain either a value or a range, but it's a widespread change that's out of scope for this pull request (it would require touching lots of code that is not involved in default value handling at all).


aaime:
No feedback, assuming it's ok to merge

PR #1398, opened 2016-01-12
Avoid log about non existing memory leak

ahuarte47:
It avoids a possible memory leak when the GetMap method throws an error.


aaime:
Please rename the commit message, it's not a memory leak, not even a possible one, it's just an ill-advised log. "Avoid log about non existing memory leak" would be better.

ahuarte47:
Done, thanks @aaime

aaime:
Merged on master, backported on 2.8.x and 2.7.x

PR #1400, opened 2016-01-15
GEOS-7377: added Java rendering engine to status page

mtCarto:
https://osgeo-org.atlassian.net/browse/GEOS-7377

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/admin/StatusPage.java
+++ b/src/web/core/src/main/java/org/geoserver/web/admin/StatusPage.java
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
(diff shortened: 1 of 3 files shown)


src/web/core/src/main/resources/GeoServerApplication.properties
```diff
@@ -609,6 +609,7 @@ StatusPage.title                            = Server Status
 StatusPage.availableFonts					= Available Fonts
 StatusPage.availableFontsList			= Full list of available fonts
 StatusPage.fontCount			= GeoServer can access {0} different fonts. 
+StatusPage.javaRenderer						= Java Rendering Engine
```
tbarsballe: Use spaces not tabs please.

src/web/core/src/main/java/org/geoserver/web/admin/StatusPage.java
```diff
@@ -75,6 +75,8 @@
     private static final String KEY_COVERAGEACCESS_KEEP_ALIVE_TIME = "coverage_thread_keepalivetime";
     
     private static final String KEY_UPDATE_SEQUENCE = "update_sequence";
+
```
tbarsballe: Please update the copyright date in the heading

tbarsballe:
Code looks good, but there are a couple of minor style issues (see comments inline)

tbarsballe:
Looks good, I will merge after the CI build completes.

src/web/core/src/main/java/org/geoserver/web/admin/StatusPage.java
```diff
@@ -198,6 +201,7 @@ private void updateModel() {
         values.put(KEY_COVERAGEACCESS_KEEP_ALIVE_TIME, Integer.toString(coverageAccess.getKeepAliveTime()));
 
         values.put(KEY_UPDATE_SEQUENCE, Long.toString(geoServerInfo.getUpdateSequence()));
+        values.put(KEY_JAVA_RENDERER, sun.java2d.pipe.RenderingEngine.getInstance().getClass().getName());
```
aaime: There is a chance this will break on a non Oracle VM (IBM, maybe Azul), maybe move it into a method, guard with try/catch, and report N/A if it cannot be found?

PR #1401, opened 2016-01-15
Consistent use of h3 headings [GEOS-7378]

jodygarnett:
Use h3 for Basic, WMS and WFS Settings to make them consistent with WCS Settings

https://osgeo-org.atlassian.net/browse/GEOS-7378


tbarsballe:
You missed "HTTP Settings" at the top of the publishing page. I have added this to the commit and manually merged.

Everything else looks good. Nice catch on the key collision on the publishing page.

PR #1402, opened 2016-01-15
[GEOS-7376] Deadlock in GWC when reloading config while under heavy tile request load

aaime:


src/gwc/src/test/java/org/geoserver/gwc/layer/CatalogConfigurationTest.java
```diff
@@ -407,4 +424,43 @@ private void testModifyLayer(GeoServerTileLayer orig) {
...
         verify(this.tileLayerCatalog, never()).save(info);
     }
+    
+    @Test
+    public void testConfigurationDeadlock() throws Exception {
+        // to make it reproducible with some reliability on my machine
+        // 100000 loops need to be attempted. With the fix it works, but runs for 
+        // a minute and a half, so not suitable for actual builds.
+        // For in-build tests I've thus settled down for 1000 loops instead 
+        final int LOOPS = 1000;
```
jodygarnett: These kind of tests always scare me in units tests, I thought we had an "intensive tests" flag - but perhaps that is only on GeoTools.

jodygarnett:
Looks good to me @aaime - I don't mind the new class (as an alternative to lock.readLock().tryLock(timeoutMs, TimeUnit.MILLISECONDS) plus boilerplate try/catch code).

Optional suggestion: The actual service exception message produced would be uninformative for a user, consider configuring the class with a timeout duration and a name to use in message reporting - so that it can produce messages such as:

```
"GWC Configuration" locked, unable to access.
```

src/gwc/src/test/java/org/geoserver/gwc/layer/CatalogConfigurationTest.java
aaime: Do not worry, with 1000 loops it takes 0.7 seconds on a old machine, not really intensive.
But you are right, there is not such a thing as a "slow test" classification in GeoServer

aaime:
Error message improved. Right now it's not a big deal, the TimeoutReadWriteLock is package private anyways to keep maintenance  burden in check, but just in case we eventually need it somewhere else, I've added a way to configure the "name".

PR #1405, opened 2016-01-16
Convert Component visitor usage over

bradh:
This PR relates to the https://github.com/geoserver/geoserver/wiki/Wicket-migration-code-sprint#todos for `Component visit has changed, see this bit of the Wicket 1.5 migration guide for reference`.

There are a few whitespace cleanups (courtesy a bad IDE setting) in the early commits, hopefully not too hard to review.

Also added a little cleanup of the last of the AbstractVisitor usage.

### DIFF
--- a/src/community/dyndimension/src/main/java/org/geoserver/web/data/resource/DynamicDimensionsTabPanel.java
+++ b/src/community/dyndimension/src/main/java/org/geoserver/web/data/resource/DynamicDimensionsTabPanel.java
@@ -29,7 +29,6 @@ import org.apache.wicket.markup.html.panel.Fragment;
 import org.apache.wicket.markup.repeater.ReuseIfModelsEqualStrategy;
 import org.apache.wicket.model.IModel;
 import org.apache.wicket.model.PropertyModel;
-import org.apache.wicket.util.visit.IVisit;
 import org.apache.wicket.validation.IValidatable;
 import org.apache.wicket.validation.IValidator;
 import org.geoserver.catalog.DimensionInfo;
@@ -273,7 +272,7 @@ public class DynamicDimensionsTabPanel extends PublishedEditTabPanel<LayerInfo> 
 
         @Override
         public void convertInput() {
-            visitChildren(TextArea.class, (Component component, final IVisit<Void> visit) -> {
+            visitChildren(TextArea.class, (component, visit) -> {
                 if (component instanceof TextArea) {
                     TextArea textArea = (TextArea) component;
                     textArea.updateModel();
--- a/src/community/dyndimension/src/main/java/org/geoserver/web/data/resource/DynamicDimensionsTabPanel.java
+++ b/src/community/dyndimension/src/main/java/org/geoserver/web/data/resource/DynamicDimensionsTabPanel.java
@@ -1,4 +1,4 @@
-/* (c) 2014 - 2015 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
  * (c) 2014 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
@@ -29,6 +29,7 @@ import org.apache.wicket.markup.html.panel.Fragment;
 import org.apache.wicket.markup.repeater.ReuseIfModelsEqualStrategy;
 import org.apache.wicket.model.IModel;
 import org.apache.wicket.model.PropertyModel;
+import org.apache.wicket.util.visit.IVisit;
 import org.apache.wicket.validation.IValidatable;
 import org.apache.wicket.validation.IValidator;
 import org.geoserver.catalog.DimensionInfo;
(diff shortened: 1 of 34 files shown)


src/community/dyndimension/src/main/java/org/geoserver/web/data/resource/DynamicDimensionsTabPanel.java
```diff
@@ -272,12 +273,10 @@ public void validate(IValidatable<DefaultValueConfigurations> validatable) {
 
         @Override
         public void convertInput() {
-            visitChildren(TextArea.class, new org.apache.wicket.Component.IVisitor<TextArea<?>>() {
-
-                @Override
-                public Object component(TextArea<?> component) {
-                    component.updateModel();
-                    return null;
+            visitChildren(TextArea.class, (Component component, final IVisit<Void> visit) -> {
```
aaime: Found out the lambda conversion could be less verbose and just use (component, visit), the compiler seems to be able to infer the argument types on its own.
Or did you do it on purpose to help the reader?

src/community/wms-eo/src/main/java/org/geoserver/wms/eo/web/EoLayerGroupAbstractPage.java
```diff
@@ -501,14 +501,14 @@ private boolean checkDimensions(EoLayerGroupEntry entry) {
...
 
         @Override
-        protected void onValidate(IValidatable validatable) {
-            String name = (String) validatable.getValue();
+        public void validate(IValidatable<String> iv) {
+            String name = (String) iv.getValue();
             LayerGroupInfo other = getCatalog().getLayerGroupByName(name);
             if(other != null && (layerGroupId == null || !other.getId().equals(layerGroupId))) {
-                error(validatable, "duplicateGroupNameError", Collections.singletonMap("name", name));
+                iv.error(new ValidationError("duplicateGroupNameError"));
```
aaime: Error message parameter "name" seems to have been lost in the conversion

src/extension/importer/web/src/main/java/org/geoserver/importer/web/NewWorkspacePanel.java
```diff
@@ -29,21 +29,22 @@ public NewWorkspacePanel(String id) {
...
         this.add(wst);
     }
 
+    static class WorkspaceDoesNotExistValidator implements IValidator<String> {
+
+        @Override
+        public void validate(IValidatable<String> iv) {
+            String value = iv.getValue();
+            if (GeoServerApplication.get().getCatalog().getWorkspaceByName(value) != null) {
+                iv.error(new ValidationError("NewWorkspacePanel.duplicateWorkspace"));
```
aaime: Error message "workspace" parameter seems to have been lost in translation

src/community/dyndimension/src/main/java/org/geoserver/web/data/resource/DynamicDimensionsTabPanel.java
bradh: Didn't realise that it could be so simple...

src/community/wms-eo/src/main/java/org/geoserver/wms/eo/web/EoLayerGroupAbstractPage.java
aaime: Btw, the ValidationError class seems to offer a fluent interface, adding .setVariable("name", name) after the constructor  call should do the trick

bradh: In this particular case the name seemed obvious. I think the previous set of conversions I did also lost a few parameters, so if we want them all, I need to do some rework.

aaime: As far as I know, if the param is not there, but found in the error message in the i18n files, we'll get an exception about the lack of the param, instead of the actual error message.

src/web/security/core/pom.xml
```diff
@@ -59,6 +59,11 @@
 			<artifactId>mockrunner</artifactId>
 			<scope>test</scope>
 		</dependency>
+            <dependency>
+                <groupId>org.apache.wicket</groupId>
+                <artifactId>wicket</artifactId>
+                <type>pom</type>
+            </dependency>
```
aaime: So the pom dependency is added everywhere because otherwise it does not get inherited? This seems quite inconvenient (and not obvious), wouldl't it be better to go back depending on actual jars instead? 
We'd have to upgrade the jar list when we upgrade Wicket, but at least, that would happen in a single place (as opposed to have to remember adding this dependency from each new module with a UI)

aaime:
Review done, battery power gone, I'm out :-)

src/web/security/core/pom.xml
bradh: I think it is just that we don't (yet) have it in a shared parent.

src/community/dyndimension/src/main/java/org/geoserver/web/data/resource/DynamicDimensionsTabPanel.java
bradh: Converted over all the `(Component component, final IVisit<Void> visit)` use to `(component, visit)` as a new commit.

src/community/wms-eo/src/main/java/org/geoserver/wms/eo/web/EoLayerGroupAbstractPage.java
bradh: Fixed in ba2005f8332935170a8b0d72d9d2c7f74f4a5121

src/extension/importer/web/src/main/java/org/geoserver/importer/web/NewWorkspacePanel.java
bradh: Fixed in ba2005f8332935170a8b0d72d9d2c7f74f4a5121

aaime:
And in it goes

PR #1406, opened 2016-01-17
[Wicket 7] Fix errors in conversion at c1aabb94e843263851e376bdff2d28ba45bba7e6

bradh:
I'd failed to realise those were resource keys.

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/data/resource/DimensionEditor.java
+++ b/src/web/core/src/main/java/org/geoserver/web/data/resource/DimensionEditor.java
@@ -464,15 +464,10 @@ public class DimensionEditor extends FormComponentPanel<DimensionInfo> {
                     if (strategyModel.getObject() == Strategy.NEAREST) {
                         if (!DimensionDefaultValueSetting.TIME_CURRENT.equalsIgnoreCase(value
                                 .getValue())) {
-                            IValidationError err = new ValidationError("invalidNearestTimeReferenceValue " + Collections
-                                    .singletonMap("value", (Object) value.getValue()));
-                            value.error(err);
+                            value.error(new ValidationError("invalidNearestTimeReferenceValue"));
                         }
                     } else {
-                        IValidationError err = new ValidationError("invalidTimeReferenceValue" + Collections.singletonMap("value",
-                                (Object) value.getValue()));
-
-                        value.error(err);
+                        value.error(new ValidationError("invalidTimeReferenceValue"));
                     }
                 }
             }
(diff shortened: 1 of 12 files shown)


src/web/core/src/main/java/org/geoserver/web/data/resource/DimensionEditor.java
```diff
@@ -464,15 +464,10 @@ else if (dimension.equals("time")) {
                     if (strategyModel.getObject() == Strategy.NEAREST) {
                         if (!DimensionDefaultValueSetting.TIME_CURRENT.equalsIgnoreCase(value
                                 .getValue())) {
-                            IValidationError err = new ValidationError("invalidNearestTimeReferenceValue " + Collections
-                                    .singletonMap("value", (Object) value.getValue()));
-                            value.error(err);
+                            value.error(new ValidationError("invalidNearestTimeReferenceValue"));
```
aaime: "value" message parameter lost in the conversion

src/web/core/src/main/java/org/geoserver/web/data/resource/DimensionEditor.java
```diff
@@ -464,15 +464,10 @@ else if (dimension.equals("time")) {
...
                         }
                     } else {
-                        IValidationError err = new ValidationError("invalidTimeReferenceValue" + Collections.singletonMap("value",
-                                (Object) value.getValue()));
-
-                        value.error(err);
+                        value.error(new ValidationError("invalidTimeReferenceValue"));
```
aaime: Likewise

src/web/core/src/main/java/org/geoserver/web/data/resource/DimensionEditor.java
bradh: In this case, the parameter was not used (hence removed).

DimensionEditor.referenceValue.invalidNearestTimeReferenceValue = Reference time value must be an ISO 8601 DateTime or word 'CURRENT'

src/web/core/src/main/java/org/geoserver/web/data/resource/DimensionEditor.java
bradh: In this case, the parameter was not used (hence removed).

```DimensionEditor.referenceValue.invalidTimeReferenceValue = Reference time value must be an ISO 8601 DateTime```

File: src/web/core/src/main/java/org/geoserver/web/data/resource/DimensionEditor.java [hunk 3]
```diff
@@ -480,9 +475,7 @@ else if (dimension.equals("elevation")) {
                 try {
                     Double.parseDouble(value.getValue());
                 } catch (NumberFormatException nfe){
-                    IValidationError err = new ValidationError("invalidElevationReferenceValue " + Collections.singletonMap("value",
-                            (Object) value.getValue()));
-                    value.error(err);
+                    value.error(new ValidationError("invalidElevationReferenceValue"));
```
bradh: The parameter was also not used here:

```DimensionEditor.referenceValue.invalidElevationReferenceValue = Reference elevation value must be numeric```

PR #1420, opened 2016-01-18
Tests in core now compile.

jnh5y:


src/web/core/src/test/java/org/geoserver/web/data/layer/LayerPageTest.java
```diff
@@ -81,7 +81,7 @@ public void testBasicActions() {
         String checkBoxPath = "table:listContainer:items:6:selectItemContainer:selectItem";
         CheckBox selector = (CheckBox) tester.getComponentFromLastRenderedPage(checkBoxPath);
         // dirty trick, how to set a form component value without a form
-        tester.getServletRequest().setParameter(selector.getInputName(), "true");
```
smithkm: This appears to be the conflict

PR #1432, opened 2016-01-20
Use the session error when returning to other pages

dvntucker:
Without using the session.error() the errors do not appear.

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/admin/LogPage.java
+++ b/src/web/core/src/main/java/org/geoserver/web/admin/LogPage.java
@@ -118,15 +118,10 @@ public class LogPage extends GeoServerSecuredPage {
                     public String getContentType() {
                         return "text/plain";
                     }
-                };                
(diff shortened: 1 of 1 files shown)


src/web/core/src/main/java/org/geoserver/web/data/store/DataAccessEditPage.java
```diff
@@ -60,9 +60,8 @@ public DataAccessEditPage(PageParameters parameters) {
         try {
```
jodygarnett: update header to 2016

src/web/wms/src/main/java/org/geoserver/wms/web/data/StyleEditPage.java
```diff
@@ -41,7 +41,7 @@ public StyleEditPage(PageParameters parameters) {
             getCatalog().getStyleByName(name);
```
jodygarnett: update header to 2016

dvntucker:
Updated the headers.

PR #1435, opened 2016-01-21
Update the DoubleInfinityConverter to fall back to its super class

dvntucker:
Fall back to regular DoubleConverter after special infinity cases are covered. Previous code caused issues because BigDecimal can't handle Double.INFINITY


tbarsballe:
Looks good, but the header needs updating.

dvntucker:
Updated the headers.

PR #1439, opened 2016-01-21
Major restructuring of the User Manual

bmmpxf:
This is a large restructuring of the GeoServer User Manual, designed to make it easier to add content to GeoServer going forward.

Among its highlights:

* Front page replaced with section headers and short descriptions. Most other index pages have been compacted. **No more giant walls of TOC**.

* The elaborate **web admin section has been removed**. Pages that were contained there have been migrated to each relevant reference section (for example: there's now a section on the security UI pages in the "Security" section). Grouping by functionality will lead to less duplication of content and make it easier for users to find information. The one main page on the web admin that remains links to all relevant sections, so users who prefer the old way of finding content will continue be able to do so.

* **New "Server configuration" section** comprising relevant content pulled from other areas.

* **"Google Earth" section is now part of WMS** and is no longer listed on the front page.

Screenshots and other content were not altered.


bmmpxf:
Thanks!

PR #1444, opened 2016-01-22
Updates to the Layer page

dvntucker:
Updates to the layer page as discussed in the sprint:

Add in Title as the second column
Remove workspace column
Removed ? on Enabled
Create Name column with workspace and layer name


src/web/core/src/main/java/org/geoserver/web/data/layer/LayerProvider.java
```diff
@@ -47,18 +46,20 @@
  */
 @SuppressWarnings("serial")
 public class LayerProvider extends GeoServerDataProvider<LayerInfo> {
-    static final Property<LayerInfo> TYPE = new BeanProperty<LayerInfo>("type",
+    static final Property<LayerInfo> TYPE = new BeanProperty<>("type",
             "type");
 
-    static final Property<LayerInfo> WORKSPACE = new BeanProperty<LayerInfo>(
+    static final Property<LayerInfo> WORKSPACE = new BeanProperty<>(
```
tbarsballe: After your change, this property is no longer used. Remove it?

tbarsballe:
The header LayerPage.java still need updating (You got halfway there ;-) )

dvntucker:
Oops, little too quick on the pushes.

src/web/core/src/main/java/org/geoserver/web/data/layer/LayerPage.java
```diff
@@ -76,10 +77,12 @@ protected Component getComponentForProperty(String id, IModel<LayerInfo> itemMod
                     return f;
                 } else if(property == SRS) {
                     return new Label(id, SRS.getModel(itemModel));
+                } else if (property == TITLE) {
```
tbarsballe: The `if (property == WORKSPACE)` segment above is probably invalid now if you are no longer displaying workspace?

dvntucker:
Removed the unused BeanProperty and its uses in the LayerPage

PR #1446, opened 2016-01-23
[GEOS-7385] Honour default charset when parsing WFS KVP Filter

bencaradocdavies:
See: https://osgeo-org.atlassian.net/browse/GEOS-7385

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/kvp/FilterKvpParser.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/kvp/FilterKvpParser.java
@@ -10,6 +10,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
 import java.io.StringReader;
+import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
@@ -81,7 +82,9 @@ public abstract class FilterKvpParser extends KvpParser {
                 InputStream input = new ByteArrayInputStream(string.getBytes());
 
                 try {
-                    Filter filter = (Filter) parser.parse(input);
+                    InputSource inputSource = new InputSource(input);
+                    inputSource.setEncoding(Charset.defaultCharset().name());
+                    Filter filter = (Filter) parser.parse(inputSource);
 
                     if (filter == null) {
                         throw new NullPointerException();
--- a/src/wfs/src/main/java/org/geoserver/wfs/kvp/FilterKvpParser.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/kvp/FilterKvpParser.java
@@ -1,4 +1,4 @@
-/* (c) 2014 - 2015 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
--- a/src/wfs/src/test/java/org/geoserver/wfs/kvp/Filter_2_0_0_KvpParserTest.java
+++ b/src/wfs/src/test/java/org/geoserver/wfs/kvp/Filter_2_0_0_KvpParserTest.java
@@ -1,0 +1,88 @@
+/* (c) 2016 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+
+package org.geoserver.wfs.kvp;
+
+import java.net.URLDecoder;
+import java.nio.charset.Charset;
+import java.util.List;
+
+import org.junit.Assert;
+import org.junit.Assume;
+import org.junit.Test;
+import org.opengis.filter.Filter;
+import org.opengis.filter.PropertyIsLike;
+import org.opengis.filter.expression.PropertyName;
+
+/**
+ * Tests for {@link Filter_2_0_0_KvpParser}, the parser for Filter 2.0 in KVP requests.
+ * 
+ * @author Ben Caradoc-Davies (Transient Software Limited)
+ */
+public class Filter_2_0_0_KvpParserTest {
+
+    /**
+     * Test that Filter 2.0 {@code fes:PropertyIsLike} can be parsed from percent-encoded form into a {@link PropertyIsLike} object.
+     * 
+     * <p>
+     * 
+     * This test is skipped if {@code expectedLiteral} cannot be encoded in the JVM default {@link Charset}.
+     * 
+     * @param expectedLiteral expected decoded filter literal
+     * @param encodedLiteral percent-encoded filter literal
+     */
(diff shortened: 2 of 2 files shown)


bencaradocdavies:
Test coverage is provided in https://github.com/bencaradocdavies/geoserver/commit/b5d258340a03309403837b0f4389ace16a16944a but the fix is only included in https://github.com/geoserver/geoserver/commit/56c2a2f7da1a7be96a92fba6ba47179e58ee8678 so that the effectiveness of the test coverage can be observed.

In src/wfs the test can be performed with various charsets with:

mvn -o -Dtest=Filter_2_0_0_KvpParserTest -Djvm.opts="-Dfile.encoding=ISO-8859-15" clean compile test

All tests will pass with "UTF-8", "ISO-8859-15", and "windows-1252". The non-ASCII literal test will be skipped for "US-ASCII" and "windows-1251" as these charsets cannot encode "ü*" and we do not want to break the build on these charsets.

Reverting the fix in https://github.com/geoserver/geoserver/commit/56c2a2f7da1a7be96a92fba6ba47179e58ee8678 and re-running the test above for various charsets results in pass for "UTF-8", the non-ASCII literal test skipped for "US-ASCII" and "windows-1251", and failure for "ISO-8859-15" and "windows-1252". The latter two failures represent the original problem report, so this demonstrates that the fix is effective and covered by the tests.

Platforms using a non-UTF-8 charset will not be able to support characters outside of their charset in a WFS KVP Filter.

bencaradocdavies:
And I should also mention that full GeoServer -PallExtensions and app-schema-online-test builds pass with and without -Djvm.opts="-Dfile.encoding=ISO-8859-15".

src/wfs/src/main/java/org/geoserver/wfs/kvp/FilterKvpParser.java
```diff
@@ -81,7 +82,9 @@ public Object parse(String value) throws Exception {
                 InputStream input = new ByteArrayInputStream(string.getBytes());
 
                 try {
-                    Filter filter = (Filter) parser.parse(input);
+                    InputSource inputSource = new InputSource(input);
+                    inputSource.setEncoding(Charset.defaultCharset().name());
```
aaime: Hi Ben, checking the GeoServerInfo.getSettings() you will find a "getCharset()" method that returns "the default charset used for responses":
http://docs.geoserver.org/stable/en/user/webadmin/server/globalsettings.html#character-set

Makes me wonder if the best choice for the input should be the system one, or the one that's configured in GeoServer, for consistency sake between input and output.

src/wfs/src/main/java/org/geoserver/wfs/kvp/FilterKvpParser.java
```diff
@@ -10,6 +10,7 @@
 import java.io.InputStream;
```
aaime: 2016 header

src/wfs/src/main/java/org/geoserver/wfs/kvp/FilterKvpParser.java
ahuarte47: Hi, I have similar problems when the GetFeature request wants to autodetect the charset of the input text. The UTF-8 charset is used by default when this autodetection fails.

See:
https://github.com/geoserver/geoserver/blob/master/src/ows/src/main/java/org/geoserver/ows/util/XmlCharsetDetector.java#L330

I must to append the first line with the encoding.

```
<?xml version="1.0" encoding="ISO-8859-1" ?>

<wfs:GetFeature service="WFS" version="1.1.0"
  xmlns:topp="http://www.openplans.org/topp"
  xmlns:wfs="http://www.opengis.net/wfs"
  xmlns:ogc="http://www.opengis.net/ogc"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://www.opengis.net/wfs http://schemas.opengis.net/wfs/1.1.0/wfs.xsd">
  
  <wfs:Query typeName="IDENA:CATAST_Pol_Municipio">
    
    <ogc:Filter xmlns:ogc="http://www.opengis.net/ogc">
          <PropertyIsLike escape="\" singleChar="_" wildCard="*" matchCase="false">
            <PropertyName>MUNINOAC</PropertyName>
            <Literal>*BARAÑA*</Literal>
          </PropertyIsLike>
    </ogc:Filter>    
  </wfs:Query>
</wfs:GetFeature>
```

It can also be a good place to consider this as well?

aaime: Hi Alvaro, I had a similar concern that I did not voice yet... which is that regardless of what the default OS charset is, the requests are coming from clients that are not running the local OS. I guess that at least for local administrations or services meant to be consumed by a single country, it could make sense to try and predict the charset, but yes, in general, it should be the client itself to add a charset indication in the filter to be parsed.

ahuarte47: Thanks @aaime, similar to your last advice about "GeoServerInfo.getSettings().getCharset()", could it be better replace the hard-coded "UTF-8" encoding in that "getEncodingName" function ?

Similar to:
https://github.com/ahuarte47/geoserver/commit/d658eab273fc61f001a3041d1ed4015378962d54#diff-17cf56a09a7f63467323d1504f02d3daR13

bencaradocdavies:
@aaime the incoming KVP filter has already been decoded from percent-encoded UTF-8 into a String at this point so there is no need or sense in using the global settings character set. This fix merely configures the InputSource so that internal round-tripping through the JVM default encoding succeeds.

However, @aaime, this is a most excellent question, because it made me realise just how silly the implementation is, and that my fix could be even better. The real problem occurs a few lines above where ```string.getBytes()``` does the damage by completely unnecessarily re-encoding the filter into the platform default charset. Any characters that cannot be encoded into the platform default charset will be damaged.

A much better solution is to avoid this pointless re-encoding by using the parser Reader interface, just as is done in ```parseXMLFilterWithOldParser(new StringReader(string))``` a few lines below (doh!). With this change, I can remove the filter test Assume line and KVP filter parsing with a non-ASCII literal "ü*" works even with -Dfile.encoding "US-ASCII" or "windows-1251", despite these being unable to encode "ü*"!.

I will close this PR and make a new one. Thanks again for your great question!

bencaradocdavies:
@ahuarte47 your addition of an XML declaration with an encoding is exactly the right thing to do. In the absence of a decalred encoding, fallback to UTF-8 is the recommendation of Appendix F (Non-Normative) of the XML specification:
https://www.w3.org/TR/REC-xml/#sec-guessing

In my view, this is much better than any server-side configuration as a client should not need to know the server encoding or configuration.

Note that my pull request relates to KVP filters which must be percent-encoded as UTF-8 octets and have no XML declaration.

ahuarte47:
Hi @bencaradocdavies, thanks for your clarifications

bencaradocdavies:
Superseded by: https://github.com/geoserver/geoserver/pull/1449

PR #1451, opened 2016-01-26
Updated the assertDateEquals to compare difference

dvntucker:
@aaime I think you originally added the test, can you take a look at this replacement? I had some issues with it randomly on my machine (I think because the method used autocasts to float from long) and I think the replacement should accomplish the same thing. 

Update the assertDateEquals method to compare that the dates are within the tolerance of one another, rather than using the method from Assert.assertEquals. The assertEquals that it was using
was autocasting to float, which was then causing infrequent issues (at least for me). The updated method should accomplish the intended goal and the tests still pass

### DIFF
--- a/src/wms/src/test/java/org/geoserver/wms/WMSDimensionsTestSupport.java
+++ b/src/wms/src/test/java/org/geoserver/wms/WMSDimensionsTestSupport.java
@@ -5,8 +5,6 @@
  */
 package org.geoserver.wms;
 
-import static org.junit.Assert.assertEquals;
-
 import java.math.BigDecimal;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


src/wms/src/test/java/org/geoserver/wms/WMSDimensionsTestSupport.java
```diff
@@ -192,7 +190,8 @@ protected void setupResourceDimensionDefaultValue(QName name, String dimensionNa
      * @param tolerance
      */
     protected static void assertDateEquals(java.util.Date d1, java.util.Date d2, long tolerance) {
-        assertEquals(d1.getTime(), d2.getTime(), tolerance);
+        long difference = Math.abs(d1.getTime() - d2.getTime());
+        assert(difference <= tolerance);
```
aaime: Should have been assertTrue

aaime:
Quite the hasty merge @jodygarnett, the code is not equivalent anymore unless assertions are enabled (-ea/-enableassertions passed to the jvm). They are enabled by default in our maven build, but might not be in IDEs

aaime:
@dvntucker good catch, but the pull should not have been merged as is, it requires a small change, can you make it?

dvntucker:
Created a new PR with the assertTrue fix: #1454

PR #1452, opened 2016-01-26
wmsMapBoxMapOutputFormat.clipToMapBounds should be false

stvno:
clipToMapBounds doesn't allow for a buffer around the tiles, whereas the VectorTileEncoder.addFeature function has a nice default buffer to clip features just beyond the tile size to prevent tile edges being rendered in the client.
Setting the clipToMapBounds to false sends over the unclipped geometries to the VectorTileEncoder which will clip them accordingly. If kept to true the geometries are already clipped to the tile-edge and the tile-edge will show up in the client

### DIFF
--- a/src/community/vectortiles/src/main/resources/applicationContext.xml
+++ b/src/community/vectortiles/src/main/resources/applicationContext.xml
@@ -56,7 +56,7 @@
   <bean id="wmsMapBoxMapOutputFormat" class="org.geoserver.wms.vector.VectorTileMapOutputFormat">
     <constructor-arg ref="wms"/>
     <constructor-arg ref="wmsMapBoxBuilderFactory"/>
-    <property name="clipToMapBounds" value="true" />
+    <property name="clipToMapBounds" value="false" />
     <property name="transformToScreenCoordinates" value="true" />
      <property name="overSamplingFactor" value="2.0">
       <description>Sub-pixel accuracy - higher value means less generalization (higher resolution results)</description>


DBlasby:
https://github.com/geoserver/geoserver/pull/1396

This PR (merged now) allows all the tile formats (including MVT) to have a buffer around them - is changing this to no-clipping necessary? It might be duplicating the work.

I'm not too familiar with the VectorTileEncoder...

stvno:
Hm, that is weird, it looks like #1396 should have fixed it, but it doesn't appear to do so on my end. I'll probe further and let you know

stvno:
The nightly builds of the vectortiles plugin are not up to date I have the feeling. I build my own jar files from the source and the problem disappeared. So this PR is not that useful anymore. 
The only thing I noticed is that the geometries get clipped twice now: Once by Pipelinebuilder with a buffer of 12 and once by the VectorTIleEncoder with a buffer of 8. Which increases the risk of clipping errors and decreases the performance.

DBlasby:
I used 12 since that's what the streaming renderer used (was going to add a parameter for size).

Is its possible to turn off the MVT clipping?  Do you know what kind of clipping its doing (i.e. is it using JTS or its own clipper).

stvno:
It uses JTS [link](https://github.com/ElectronicChartCentre/java-vector-tile/blob/master/src/main/java/no/ecc/vectortile/VectorTileEncoder.java) 
It always clips the geometry if I read (this)[https://github.com/ElectronicChartCentre/java-vector-tile/blob/master/src/main/java/no/ecc/vectortile/VectorTileEncoder.java#L141] correctly. (I'm no Java coder so I might be mistaken).

DBlasby:
Yes - it looks like it will always clip (and the clipping code is a bit strange).  Perhaps your PR will work out fine (although likely better to have the clipping in the external VectorTileEncoder library be optional).

stvno:
It is a crucial part of the MVT spec that the geometries are clipped, so I can understand why the library does it. BTW: it is possible to set the VectorTIleEncoder (VTE) buffer the same as the PipelineBuilder (PlB); if you set the [polygonClipBuffer](https://github.com/geoserver/geoserver/blob/master/src/community/vectortiles/src/main/java/org/geoserver/wms/mapbox/MapBoxTileBuilder.java#L34) to the same value as PipelineBuilder's 12 the clipping shouldn't change (though still calculated twice). 

@NielsCharlier is involved in both VTE and GeoServer so he might be able to suggest a proper way forward to streamline the integration between the two.

DBlasby:
I think the best solution would be to either 
(a) make clipping an option in the VTE library (and then clip in geoserver)
(b) modify the VTE clipping to be a more robust

The VTE clipping code has two problems (a) the weird WKT stuff and (b) it assumes valid polygons as inputs and ignores topologyexceptions during clipping.

I'm using a topology preserving generalization function to do the geometry generalization - its pretty slow but returns valid polygons.  Its likely something that will change in the future (to a faster method that will return invalid polygons).

Here's the code in the geoserver vector tiles - its using Geotool's GeometryClipper.  It has two algorithms - JTS's method and a very simple clipping algorithm (can produce invalid, but drawable, polygons). The pipeline should be delivering valid polygons (currently using the topology preserving generalization) - if there's a problem clipping, it falls back to the simple clipper. 

 private static final class Clip extends Pipeline {

        private final Envelope clippingEnvelope;

        Clip(Envelope clippingEnvelope) {
            this.clippingEnvelope = clippingEnvelope;
        }

        @Override
        protected Geometry _run(Geometry geom) throws Exception {
            GeometryClipper clipper = new GeometryClipper(clippingEnvelope);
            try {
                return clipper.clip(geom, true);        
            } catch (Exception e) {
                return clipper.clip(geom, false);    // use non-robust clipper
            }
            
        }
    }

aaime:
@DBlasby maybe you just need to upgrade your dependecy on the library, the latest release has this commit in:

https://github.com/ElectronicChartCentre/java-vector-tile/commit/3c0b401dc20e8d3cb240640dc2ab5dc8525ec3b8

aaime:
This one seems to have stalled. Any further action planned?

aaime:
Should we close this pull request?

aaime:
I am going to close this one in a few days, unless you object and would like to work on it a bit more?

DBlasby:
Close this for now - the upstream library needs to be modified.

PR #1457, opened 2016-01-28
Add params-extractor community module

nmco:


src/community/params-extractor/src/main/java/org/geoserver/params/extractor/Filter.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.logging.Logger;
+
+public final class Filter implements GeoServerFilter, ExtensionPriority {
+
+    private static final Logger LOGGER = Logging.getLogger(Filter.class);
+
+    private static List<Rule> rules;
+
+    public Filter(ResourceStore dataDirectory) {
+        Resource resource = dataDirectory.get("rules.xml");
```
mbarto: Is the rules.xml stored in main data_dir folder? Can we move it to a more meaningful subfolder? (params-extractor) or give it a more meaningful name than rules.xml (params-extractor.xml)?

mbarto:
Looks good to me, if we can just make the persisted file name / path more meaningful

nmco:
Thank you for the review, you're right about the persistent path I will use the path 'params-extractor/extraction-rules.xml' if this sound good to you.

mbarto:
yes, that path is good for me

mbarto:
Adding some documentation in rst would be a nice addition

PR #1458, opened 2016-01-28
Improve performance in CSVOutputFormat

anthonyccri:
* Use Guava Escapers to escape embedded quotes in attributes rather
  than String.replace as the latter recompiles the regex for every
  single attribute
* Compute the attribute formatters once and reuse rather than checking
  instanceof for every attribute

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/response/CSVOutputFormat.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/response/CSVOutputFormat.java
@@ -1,21 +1,12 @@
-/* (c) 2014 - 2015 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
  */
 package org.geoserver.wfs.response;
 
-import java.io.BufferedWriter;
-import java.io.IOException;
-import java.io.OutputStream;
-import java.io.OutputStreamWriter;
-import java.text.NumberFormat;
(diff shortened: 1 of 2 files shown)


src/wfs/src/test/java/org/geoserver/wfs/response/CSVOutputFormatTest.java
```diff
@@ -5,17 +5,13 @@
...
 
-import static org.junit.Assert.*;
-
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.io.StringReader;
-import java.util.ArrayList;
-import java.util.List;
-
+import au.com.bytecode.opencsv.CSVReader;
```
aaime: Copyright header should say "2014 - 2016"

src/wfs/src/main/java/org/geoserver/wfs/response/CSVOutputFormat.java
```diff
@@ -5,17 +5,8 @@
  */
```
aaime: Copyright header should say "2014 - 2016" now (from 2014 - 2015)

aaime:
Since this is a performance improvement, can you share how much the performance did improve in your tests? :-)

anthonyccri:
@aaime I fixed the copyright issue.  As far as quantifying the performance gain, I ran a test of serializing 500K features, each with 50 attributes, using the old code and the new code and observed a performance gain of 2.5X.  On my laptop, this dropped processing time from 25s to 10s.

aaime:
Please next time open a ticket and refer it from the commit message. I've created the ticket for you, so that this commit makes into the release notes: https://osgeo-org.atlassian.net/browse/GEOS-7406

PR #1459, opened 2016-01-29
[GEOS-7387] Cached layer removal fails to remove cached tiles

bencaradocdavies:
See: https://osgeo-org.atlassian.net/browse/GEOS-7387

### DIFF
--- a/src/gwc/src/test/java/org/geoserver/gwc/layer/CatalogConfigurationTest.java
+++ b/src/gwc/src/test/java/org/geoserver/gwc/layer/CatalogConfigurationTest.java
@@ -364,7 +364,7 @@ public class CatalogConfigurationTest {
         doThrow(new IllegalArgumentException("failedSave")).when(tileLayerCatalog).save(
                 eq(forceState2));
 
-        verify(mockMediator, never()).layerRemoved(anyString());
+        verify(mockMediator, times(1)).layerRemoved(layerInfo2.getName());
         verify(mockMediator, never()).layerRenamed(anyString(), anyString());
 
         GeoServerTileLayerInfo addedState1 = TileLayerInfoUtil.loadOrCreate(layerWithNoTileLayer,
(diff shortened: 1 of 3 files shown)


src/gwc/src/test/java/org/geoserver/gwc/layer/CatalogConfigurationTest.java
```diff
@@ -364,7 +364,7 @@ private void testModifyLayer(GeoServerTileLayer orig) {
         doThrow(new IllegalArgumentException("failedSave")).when(tileLayerCatalog).save(
                 eq(forceState2));
 
-        verify(mockMediator, never()).layerRemoved(anyString());
```
aaime: Hmm... given the bug description wouldn't it be better to actually check the tiles are gone on disk? However that would require a non mock test. Just an idea, it's ok like this too I assume (just don't know GWC notifications well enough to say "aha, yeah, that does it")

smithkm: If the event gets fired without the tiles being deleted, that's a bug in a component other than the one being tested so I think this is good enough for testing the Configuration.

bencaradocdavies:
@aaime, I agree: we need a non-mock test somewhere because the mocked test is too thin to cover integration with the blob stores. The mocked tests never caught this failure in the first place. I have pushed https://github.com/bencaradocdavies/geoserver/commit/7b67258525eca0622f23750311240e91a6495b49 on to this pull request to add a non-mock test that cached tiles are really removed from disk when the layer is removed from the catalog.

aaime:
Thanks @bencaradocdavies, that makes me feel a lot better about this not regressing, as opposed to just having a way to say, after the regressione happened, "but this is not my fault, my code is doing the right thing" :-p

bencaradocdavies:
Thanks @aaime. There is still a small risk that the additional locking could impact performance at the moment a delete is triggered, but I am not concerned. Both FileBlobStore and S3BlobStore perform bulk deletes on another thread to prevent blocking. At least we can now be confident that cached tiles are removed.

PR #1460, opened 2016-01-29
GEOS-7392 SecuredLayerGroup bug

NielsCharlier:
https://osgeo-org.atlassian.net/browse/GEOS-7392

ModificationProxyCloner change was in the end not necessary to fix the bug, but the change still makes sense so I am leaving it in.

### DIFF
--- a/src/extension/csw/core/src/test/java/org/geoserver/csw/store/internal/LayerGroupTest.java
+++ b/src/extension/csw/core/src/test/java/org/geoserver/csw/store/internal/LayerGroupTest.java
@@ -5,8 +5,8 @@ import static org.custommonkey.xmlunit.XMLAssert.assertXpathExists;
 
 import java.util.Arrays;
 
+import org.geoserver.catalog.LayerGroupInfo;
 import org.geoserver.catalog.PublishedInfo;
-import org.geoserver.catalog.impl.LayerGroupInfoImpl;
 import org.geoserver.data.test.SystemTestData;
 import org.geotools.csw.CSWConfiguration;
 import org.junit.Test;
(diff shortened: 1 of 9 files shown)


NielsCharlier:
the second change to the ModificationProxyCloner is necessary.

aaime:
The Travis failures appear to be legit and caused by this change (at the very least, they do involve layer groups):

````
Failed tests:   testRecordSortedAndPaged(org.geoserver.csw.store.internal.LayerGroupTest): Expecting to find matches for Xpath //csw:SummaryRecord[dc:title='Buildings and Bridges']
  testRecords(org.geoserver.csw.store.internal.LayerGroupTest): Expecting to find matches for Xpath //csw:Record[dc:title='Buildings and Bridges']

Tests in error: 
  testRecordById(org.geoserver.csw.store.internal.LayerGroupTest): org.geoserver.security.decorators.SecuredLayerGroupInfo cannot be cast to org.geoserver.catalog.impl.LayerGroupInfoImpl
````

**Comment by NielsCharlier** [comment]
fixed

File: src/extension/csw/core/src/test/java/org/geoserver/csw/store/internal/LayerGroupTest.java [hunk 1]
```diff
@@ -5,8 +5,8 @@
```
aaime: Lack copyright header

src/extension/csw/core/src/test/java/org/geoserver/csw/store/internal/LayerGroupTest.java
```diff
@@ -61,7 +62,7 @@ public void testRecords() throws Exception {
     
     @Test
     public void testRecordById() throws Exception {
-        String request = "csw?service=CSW&version=2.0.2&request=GetRecordById&typeNames=csw:Record&id=" + ID_FORESTSANDSTREAMS;
+        String request = "csw?service=CSW&version=2.0.2&request=GetRecordById&typeNames=csw:Record&id=" + id_forestsandstreams;
         Document d = getAsDOM(request);
         //print(d);
         checkValidationErrors(d);
```
aaime: testRecordSortedAndPaged has an uncommented print statement

src/main/src/main/java/org/geoserver/catalog/impl/ModificationProxyCloner.java
```diff
@@ -22,6 +22,7 @@
 import java.util.logging.Logger;
```
aaime: Copyright header not updated

src/main/src/main/java/org/geoserver/security/decorators/DecoratingCatalogFactory.java
```diff
@@ -0,0 +1,125 @@
+package org.geoserver.security.decorators;
```
aaime: No copyright header

src/main/src/main/java/org/geoserver/security/decorators/DecoratingCatalogFactory.java
```diff
@@ -0,0 +1,125 @@
...
+import org.geoserver.catalog.MapInfo;
+import org.geoserver.catalog.MetadataLinkInfo;
+import org.geoserver.catalog.NamespaceInfo;
+import org.geoserver.catalog.StyleInfo;
+import org.geoserver.catalog.WMSLayerInfo;
+import org.geoserver.catalog.WMSStoreInfo;
+import org.geoserver.catalog.WorkspaceInfo;
+import org.geoserver.catalog.impl.AbstractDecorator;
+
+public class DecoratingCatalogFactory extends AbstractDecorator<CatalogFactory> implements CatalogFactory {
```
aaime: No javadoc for a public class?

src/main/src/main/java/org/geoserver/catalog/impl/ModificationProxyCloner.java
```diff
@@ -194,7 +195,13 @@ public int compare(Class c1, Class c2) {
             // nothing to copy
             return null;
         }
-        Collection<T> copy = source.getClass().newInstance();
+        Collection<T> copy;
+        try {
+            copy = source.getClass().newInstance();
+        } catch (InstantiationException e) {
+            //we'll just pick something
+            copy = new ArrayList<T>();
```
aaime: This something is I suppose what you needed, but what if the original collection was a Set or a Map?
What was the class that could not be instantiated?

aaime:
I commented on a few, but several classes are missing copyright year updates, please fix them all.

src/main/src/main/java/org/geoserver/catalog/impl/ModificationProxyCloner.java
NielsCharlier: Thanks for the review. The problem is [this](https://github.com/NielsCharlier/geoserver/commit/e919df627299543e2177fc8f814f8a3dfce92f5b#diff-56150f92aa7ba936e390c7a4aa0438ffR50) inner class I made ( a list). I searched for a way to reliably clone any collection easily, but that doesn't seem to exist except if you use some third part library. But I guess the least I could do is add some extra ifs for sets etc... although that shouldn't occur (for now)

aaime: You could switch the class to static and pass around the fields it needs, then give it also a empty constructor, that would fix issues without assuming a non cloneable collection is necessarily a list.

NielsCharlier:
Fixed

PR #1461, opened 2016-02-01
Set slf4j dependencies version in root pom.

groldan:
Homogenize slf4j dependency version, there were
several mixed versions across modules.

Upgrade to latest stable 1.7.14.

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -844,8 +844,18 @@
    </dependency>
    <dependency>
      <groupId>org.slf4j</groupId>
+     <artifactId>slf4j-api</artifactId>
+     <version>1.7.14</version>
+   </dependency>
+   <dependency>
+     <groupId>org.slf4j</groupId>
      <artifactId>slf4j-log4j12</artifactId>
-     <version>1.4.2</version>
+     <version>1.7.14</version>
+   </dependency>
+   <dependency>
+     <groupId>org.slf4j</groupId>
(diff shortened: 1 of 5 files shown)


src/pom.xml
```diff
@@ -844,8 +844,18 @@
    </dependency>
    <dependency>
      <groupId>org.slf4j</groupId>
+     <artifactId>slf4j-api</artifactId>
+     <version>1.7.14</version>
```
jodygarnett: Thanks for adding to dependency management section.

(I agree with only two entries it is not worth breaking out a ``${slfj.version}`` property.

jodygarnett:
Thanks for the build improvement.

jodygarnett:
@groldan - noticed this was to the 2.7.x branch, can I ask you to cherry pick this change to 2.8.x and master.

groldan:
@jodygarnett sure, thanks, my next question was going to be if I should create separate PR's for 2.8.x and master.

jodygarnett:
Just do it.
On Mon, Feb 1, 2016 at 5:56 PM Gabriel Roldan <notifications@github.com>
wrote:

> @jodygarnett <https://github.com/jodygarnett> sure, thanks, my next
> question was going to be if I should create separate PR's for 2.8.x and
> master.
>
> —
> <https://github.com/geoserver/geoserver/pull/1461#issuecomment-178066317>.
>
-- 
--
Jody Garnett

PR #1462, opened 2016-02-01
[GEOS-7395] Add missing JARs to geofence plugin (2.8.x backport)

nmco:
This also corrects some spaces issues.


simboss:
Apparently the failure looks unrelated.  Going ahead with the merge, weill fix right away if something happens...

PR #1465, opened 2016-02-02
GEOS-7360 GWC-diskquota-config exception

NielsCharlier:


dromagnoli:
Hi @NielsCharlier, 
I have locally applied your patch and it seems resolved. So I would say "+1" on merging it once travis is Green
Thanks for the fix.

PR #1478, opened 2016-02-06
Remove JDK8 dependency and make rule tests predictable (backport 2.8.x)

nmco:
Make rule tests predictable


aaime:
Merging, build failure is unrelated, it's likely the Boundless repository with broken metadata causing this (will follow up on the devel list)

PR #1479, opened 2016-02-06
Make Rule tests predictable

nmco:


aaime:
Merge, build failure is unrelated (it's an old random failure that we never figured out)

PR #1484, opened 2016-02-08
[GEOS-7414] Fix LayerGroups styles handling in Geofence

nmco:
Currently if GeoFence is active a request like this:

`curl -X GET 'http://localhost:8080/geoserver/wms?SERVICE=WMS&VERSION=1.1.1&REQUEST=GetMap&FORMAT=image%2Fpng&TRANSPARENT=true&STYLES=,line&LAYERS=tiger-ny,tiger_roads&SRS=EPSG%3A4326&WIDTH=531&HEIGHT=768&BBOX=-74.05900955200195%2C40.640830993652344%2C-73.87670516967773%2C40.9045028686523445' -o result.png`

will provoke an index out of bounds exception.

This patch allow GeoFence to properly handle requests that contains LayerGroups, normal Layers and dynamic styles.

This patch also adds a test case for this.

Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-7414

### DIFF
--- a/src/community/geofence/src/main/java/org/geoserver/geofence/GeofenceAccessManager.java
+++ b/src/community/geofence/src/main/java/org/geoserver/geofence/GeofenceAccessManager.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright (C) 2007 - 2015 GeoSolutions S.A.S.
+ *  Copyright (C) 2007 - 2016 GeoSolutions S.A.S.
  *  http://www.geo-solutions.it
  *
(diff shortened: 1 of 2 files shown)


simboss:
Had a cursory look, it looks fine to me.

Minor thing, there is no JIRA for the issue highlighted here, I would create it and reference it in all PR.

nmco:
Issue created and pull request info updated.

mbarto:
Looks good to me

PR #1486, opened 2016-02-09
GEOS-7415: fixed printing module demo page, broken during wicket upgrade

mbarto:


src/extension/printing/src/test/java/org/geoserver/printing/PrintDemoPageTest.java
```diff
@@ -20,15 +20,17 @@ public void setup() {
     public void testCSS() {
```
aaime: Header not updated

PR #1487, opened 2016-02-10
GEOS-7407 / GSIP 137 ResourceStore Rest API

NielsCharlier:
https://osgeo-org.atlassian.net/browse/GEOS-7407
https://github.com/geoserver/geoserver/wiki/GSIP-137


aaime:
@NielsCharlier can you please review the contribution guidelines so that we don't have to remind you every time about them during reviews? It would be great if the review could focus on the real contents instead of the other details.
with particular focus to:
https://github.com/geoserver/geoserver/blob/master/CONTRIBUTING.md#commit-guidelines

bencaradocdavies:
@NielsCharlier please update copyright headers on ```AbstractResource.java``` and ``RESTUtils.java```.

src/restconfig/src/main/java/org/geoserver/rest/resources/resourceDirectory.ftl
```diff
@@ -0,0 +1,26 @@
...
+<body>
+
+<h1>Resource Directory '${properties.path}'</h1>
+
+<ul>
+  <li>Name: '${properties.name}'</li>
+  <li>Parent: <a href="${properties.parent_href}">${properties.parent_path}</a></li>
+  <li>Last modified: ${properties.lastModified}</li>
+  <li>Children:
+	<ul>
```
bencaradocdavies: @NielsCharlier please use spaces not tabs for indentation.

src/restconfig/src/main/java/org/geoserver/rest/AtomLink.java
```diff
@@ -0,0 +1,66 @@
...
+ * 
+ * @author Niels Charlier
+ *
+ */
+public class AtomLink {
+
+    @SuppressWarnings("unused")
+    private final String NAMESPACE = "http://www.w3.org/2005/Atom";
+    
+    private String href, rel, type;
```
bencaradocdavies: @NielsCharlier the code conventions prefer one declaration per line (section 6.1).

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceResource.java
```diff
@@ -0,0 +1,409 @@
...
+
+/**
+ *
+ *
+ * @author Niels Charlier
+ *
+ */
+public class ResourceResource extends AbstractResource {
+    
+    public enum Operation {DEFAULT, METADATA, MOVE, COPY};
```
bencaradocdavies: @NielsCharlier the introduction of enums was after the Sun code conventions were written, but both existing practice for blocks in declarations and the published GeoTools Eclipse formatter.xml suggest this formatting for this enum:
```
    public enum Operation {
        DEFAULT, METADATA, MOVE, COPY
    };
```

bencaradocdavies:
@NielsCharlier this pull request causes the build to fail in ```ResourceTest``` at line 294:
```
Results :

Failed tests:   testErrorResponseCodes(org.geoserver.rest.resources.ResourceTest): expected:<405> but was:<404>

Tests run: 371, Failures: 1, Errors: 0, Skipped: 3
```
```
------------------------------------------------------------------------------
Test set: org.geoserver.rest.resources.ResourceTest
Tests run: 16, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.439 sec <<< FAILURE!
testErrorResponseCodes(org.geoserver.rest.resources.ResourceTest)  Time elapsed: 9 sec  <<< FAILURE!
java.lang.AssertionError: expected:<405> but was:<404>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:743)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:555)
	at org.junit.Assert.assertEquals(Assert.java:542)
	at org.geoserver.rest.resources.ResourceTest.testErrorResponseCodes(ResourceTest.java:294)
```
The failing assertion is here:
```
        //copy dir
        response = putAsServletResponse("/rest/resource/mynewdir?operation=copy", "/rest/resource/mydir", "text/plain");
        Assert.assertEquals(405, response.getStatusCode());
```
Travis CI did not run for this PR. I do not know why.

src/restconfig/src/test/java/org/geoserver/rest/resources/ResourceTest.java
```diff
@@ -0,0 +1,310 @@
...
+                + "  \"type\": \"resource\""
+                + "}}";
+        JSONAssert.assertEquals(expected, (JSONObject) json);
+    }
+    
+    @Test
+    public void testResourceMetadataHTML() throws Exception {
+        XMLUnit.setXpathNamespaceContext(NS_HTML);
+        Document doc = getAsDOM("/rest/resource/mydir/myres?operation=metadata&format=html");
+        print(doc);
```
bencaradocdavies: @NielsCharlier please do not print to stdout during tests. I see that you commented out the print lines in other tests, but you missed this one.  :-)

doc/en/user/source/rest/api/resources.rst
```diff
@@ -0,0 +1,119 @@
...
+
+Headers
+~~~~~~~
+
+.. list-table::
+   :header-rows: 1
+
+   * - Header
+     - Description
+   * - Last-modified
```
bencaradocdavies: @NielsCharlier while HTTP header field names are case insensitive, conforming to conventional capitalisation would improve readability and recognition of the standard headers. Perhaps ```Last-Modified```, ```Content-Type```, ```Resource-Type```, and ```Resource-Parent```?

src/rest/src/main/java/org/geoserver/rest/AbstractResource.java
```diff
@@ -92,7 +96,7 @@ protected DataFormat getFormatGet() {
         if ( ext == null ) {
             ext = (String) getRequest().getAttributes().get( "type" );
         }
-        if ( ext == null ) {
+        if ( ext == null && includeFileExtension) {
```
bencaradocdavies: @NielsCharlier extra space after opening bracket. Yes, you did not add it, but this is a great time to fix it.

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceFinder.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.geoserver.platform.resource.ResourceStore;
+import org.geoserver.rest.util.RESTUtils;
+import org.restlet.Finder;
+import org.restlet.data.Request;
+import org.restlet.data.Response;
+import org.restlet.resource.Resource;
+
+/**
+ *
```
bencaradocdavies: Please insert a comment here. For example, is this a Spring bean for handling REST requests for resources?

src/restconfig/src/main/java/org/geoserver/rest/AtomLink.java
```diff
@@ -0,0 +1,66 @@
+/* (c) 2016 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.rest;
+
+import com.thoughtworks.xstream.XStream;
+/**
```
bencaradocdavies: Should have a blank line here to improve readability.

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceResource.java
```diff
@@ -0,0 +1,409 @@
...
+import org.restlet.resource.Representation;
+
+import com.thoughtworks.xstream.annotations.XStreamAlias;
+
+import freemarker.ext.beans.CollectionModel;
+import freemarker.template.Configuration;
+import freemarker.template.SimpleHash;
+
+/**
+ *
```
bencaradocdavies: @NielsCharlier please add a comment here to explain the class name, which, while logical and consistent, is also perverse. You may also need a reference to Catch-22:  https://en.wikipedia.org/wiki/Major_Major_Major

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceResource.java
```diff
@@ -0,0 +1,409 @@
...
+        //HTML
+        if (operation == Operation.METADATA || resource.getType() == Type.DIRECTORY) {
+            formats.add(new ReflectiveHTMLFormat(request,response,this) {   
+                @Override
+                protected Configuration createConfiguration(Object data, Class clazz) {
+                    Configuration cfg = super.createConfiguration(data, clazz);
+                    cfg.setClassForTemplateLoading(ResourceResource.class, "");
+                    cfg.setObjectWrapper(new ObjectToMapWrapper<ResourceMetadata>(ResourceMetadata.class) {
+                        @Override
+                        protected void wrapInternal(Map properties, SimpleHash model, ResourceMetadata object) {
```
bencaradocdavies: @NielsCharlier please add ```@SuppressWarnings("unchecked")``` to ```wrapInternal```. Eclipse no like.

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceResource.java
```diff
@@ -0,0 +1,409 @@
...
+                            super.wrapInternal(properties, model, object);
+                            properties.put("path", object.getParent() == null ? "/" :
+                                Paths.get(object.getParent().getPath(), object.getName()) ); 
+                            properties.put("parent_path", object.getParent() == null ? "" : object.getParent().getPath());
```
bencaradocdavies: @NielsCharlier many lines in ```ResourceResource.java``` exceed the 100 character line length specified in our coding conventions. Did you format this class? There is also trailing whitespace that will likewise be fixed by formatting..

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceResource.java
```diff
@@ -0,0 +1,409 @@
...
+    
+    @Override
+    public void handlePut() {
+        if (resource.getType() == Type.DIRECTORY) {
+            throw new RestletException("Attempting to write data to a directory.", Status.CLIENT_ERROR_METHOD_NOT_ALLOWED);
+        } else {
+            try {
+                boolean isNew = resource.getType() == Type.UNDEFINED;
+                
+                if (operation == Operation.COPY || operation == Operation.MOVE) {
```
bencaradocdavies: @NielsCharlier, there is loads of trailing whitespace in this file. Adding a new file is a great opportunity to use the Eclipse formatter.

src/restconfig/src/main/java/org/geoserver/rest/AtomLink.java
```diff
@@ -0,0 +1,66 @@
...
+    private final String NAMESPACE = "http://www.w3.org/2005/Atom";
+    
+    private String href, rel, type;
+    
+    public AtomLink(String href, String rel, String type) {
+        this.href = href;
+        this.rel = rel;
+        this.type = type;
+    }
+
```
bencaradocdavies: Trailing whitespace/ New file? Use the Eclipse formatter!

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceResource.java
```diff
@@ -0,0 +1,409 @@
...
+            this.lastModified = lastModified;
+            this.type = type;
+        }
+
+        /**
+         * Create from resource.
+         * The class must be static for serialization, but output is request dependent so passing on self.
+         */
+        protected ResourceMetadata(Resource resource, ResourceResource self, boolean isDir) {
+            if (!"".equals(resource.path())) {
```
bencaradocdavies: @NielsCharlier this is a strange idiom. Does your intent differ from ```!resource.path().isEmpty()```? Can ```resource.path()``` return null?

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceResource.java
```diff
@@ -0,0 +1,409 @@
...
+         */
+        public ResourceDirectory(Resource resource, ResourceResource self) {
+            super(resource, self, true);
+            for (Resource child : resource.list()) {
+                children.add(new ResourceChild(child.name(), 
+                        new AtomLink(self.href(child.path(), child.getType() == Type.DIRECTORY), 
+                                "alternate", self.getMediaType(child).getName())));
+            }
+        }
+
```
bencaradocdavies: Lots of training whitespace.

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceResource.java
```diff
@@ -0,0 +1,409 @@
...
+                } else {
+                    rep = new InputRepresentation(resource.in(), getMediaType(resource));
+                }
+                rep.setModificationDate(new Date(resource.lastmodified()));
+                getVariants().add(rep);
+
+                if (!"".equals(resource.path())) {
+                    getResponseHeaders().add("Resource-parent", "/" + resource.parent().path());
+                }
+                getResponseHeaders().add("Resource-type", resource.getType().toString().toLowerCase());
```
bencaradocdavies: @NielsCharlier please see my remark in your docs about conventional capitalisation. I think these should be ```Resource-Parent``` and ```Resource-Type```.

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceResource.java
```diff
@@ -0,0 +1,409 @@
...
+    public void handlePut() {
+        if (resource.getType() == Type.DIRECTORY) {
+            throw new RestletException("Attempting to write data to a directory.", Status.CLIENT_ERROR_METHOD_NOT_ALLOWED);
+        } else {
+            try {
+                boolean isNew = resource.getType() == Type.UNDEFINED;
+                
+                if (operation == Operation.COPY || operation == Operation.MOVE) {                    
+                    try (InputStream is = getRequest().getEntity().getStream()) {
+                        Resource source = store.get(IOUtils.toString(is, "UTF-8"));
```
bencaradocdavies: @NielsCharlier I do not understand this. Why UTF-8? What happens if I PUT a resource with ```<?xml version="1.0" encoding="ISO-8859-15"?>``` in which paths contain non-ASCII characters?

src/restconfig/src/main/java/org/geoserver/rest/resources/resourceMetadata.ftl
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
```
bencaradocdavies: I am assuming that the infrastructure will set the charset when the response is returned; does it add a meta header?

Is correct to not add a charset here for the freemarker template itself? Or do we insist that resources are UTF-8? I think Maven is configured to do so. We might need <#ftl encoding="utf-8"> in this file.
http://freemarker.incubator.apache.org/docs/pgui_misc_charset.html

I am unsure whether this file should contain <meta charset="utf-8"> after ```<head>```.

src/restconfig/src/test/java/org/geoserver/rest/resources/ResourceTest.java
```diff
@@ -0,0 +1,310 @@
...
+        XMLAssert.assertXpathEvaluatesTo("Type: resource", "/x:html/x:body/x:ul/x:li[3]", doc);
+        XMLAssert.assertXpathEvaluatesTo("Last modified: " + new Date(myRes.lastmodified()).toString(), 
+                "/x:html/x:body/x:ul/x:li[4]", doc);
+    }
+    
+    @Test
+    public void testResourceHeaders() throws Exception {
+        MockHttpServletResponse response = getAsServletResponse("/rest/resource/mydir2/fake.png");
+        Assert.assertEquals(FORMAT_HEADER.format(myRes.lastmodified()), response.getHeader("Last-modified"));
+        Assert.assertEquals("/mydir2", response.getHeader("Resource-parent"));
```
bencaradocdavies: Conventional header capitalisation, please.

src/restconfig/src/test/java/org/geoserver/rest/resources/ResourceTest.java
```diff
@@ -0,0 +1,310 @@
...
+        XMLAssert.assertXpathEvaluatesTo("http://localhost:8080/geoserver/rest/resources/mydir/myres", 
+                "/x:html/x:body/x:ul/x:li[4]/x:ul/x:li/x:a/@href", doc);
+    }
+    
+    @Test
+    public void testDirectoryHeaders() throws Exception {
+        MockHttpServletResponse response = getAsServletResponse("/rest/resource/mydir?format=xml");
+        Assert.assertEquals(FORMAT_HEADER.format(myRes.lastmodified()), response.getHeader("Last-modified"));
+        Assert.assertEquals("/", response.getHeader("Resource-parent"));
+        Assert.assertEquals("directory", response.getHeader("Resource-type"));
```
bencaradocdavies: Conventional capitalisation, please.

src/restconfig/src/test/java/org/geoserver/rest/resources/ResourceTest.java
```diff
@@ -0,0 +1,310 @@
...
+        try (OutputStreamWriter os = new OutputStreamWriter(myRes.out())) {
+            os.append("This is my test.");
+        }
+        
+        try (OutputStreamWriter os = new OutputStreamWriter(getDataDirectory().get("/mydir2/fake.png").out())) {
+            os.append("This is not a real png file.");
+        }
+
+        IOUtils.copyStream(getClass().getResourceAsStream("testimage.png"),
+                getDataDirectory().get("/mydir2/imagewithoutextension").out(), true, true);
```
bencaradocdavies: @NielsCharlier please include non-ASCII characters in paths and content. This will give you unit test coverage of charset handling. You can also run Maven tests with -Dfile.encoding as this is used to set the JVM default charset and many parsers will choose it if a charset is not specified.

src/restconfig/src/test/java/org/geoserver/rest/resources/ResourceTest.java
```diff
@@ -0,0 +1,310 @@
...
+    @Test
+    public void testResource() throws Exception {
+        String str = getAsString("/rest/resource/mydir/myres");
+        Assert.assertEquals("This is my test.\n", str);
+    }
+    
+    @Test
+    public void testResourceMetadataXML() throws Exception {
+        XMLUnit.setXpathNamespaceContext(NS_XML);
+        Document doc = getAsDOM("/rest/resource/mydir/myres?operation=metadata&format=xml");
```
bencaradocdavies: @NielsCharlier change this to ```operation=mEtAdATa``` to test case insensitivity.

src/restconfig/src/test/java/org/geoserver/rest/resources/ResourceTest.java
```diff
@@ -0,0 +1,310 @@
...
+        XMLUnit.setXpathNamespaceContext(NS_XML);
+        Document doc = getAsDOM("/rest/resource/mydir2?format=xml");
+        //print(doc);
+        XMLAssert.assertXpathEvaluatesTo("image/png", "/ResourceDirectory/children/child[name='imagewithoutextension']/atom:link/@type", doc);
+        XMLAssert.assertXpathEvaluatesTo("image/png", "/ResourceDirectory/children/child[name='fake.png']/atom:link/@type", doc);
+    }
+        
+    @Test
+    public void testUpload() throws Exception {
+        put("/rest/resource/mydir/mynewres", "This is my new test.");
```
bencaradocdavies: @NielsCharlier great opportunity to include non-ASCII characters to see if they survive the round trip.

src/restconfig/src/test/java/org/geoserver/rest/resources/ResourceTest.java
```diff
@@ -0,0 +1,310 @@
...
+        try (InputStream is = newRes.in()) {
+            Assert.assertEquals("This is my new test.", IOUtils.toString(is));
+        }
+        
+        newRes.delete();
+    }
+    
+    @Test
+    public void testCopy() throws Exception {
+        put("/rest/resource/mydir/mynewres?operation=copy", "/mydir/myres");
```
bencaradocdavies: ```operation=cOpY```

src/restconfig/src/test/java/org/geoserver/rest/resources/ResourceTest.java
```diff
@@ -0,0 +1,310 @@
...
+                "/x:html/x:body/x:ul/x:li[3]", doc);
+        XMLAssert.assertXpathEvaluatesTo("http://localhost:8080/geoserver/rest/resources/mydir/myres", 
+                "/x:html/x:body/x:ul/x:li[4]/x:ul/x:li/x:a/@href", doc);
+    }
+    
+    @Test
+    public void testDirectoryHeaders() throws Exception {
+        MockHttpServletResponse response = getAsServletResponse("/rest/resource/mydir?format=xml");
+        Assert.assertEquals(FORMAT_HEADER.format(myRes.lastmodified()), response.getHeader("Last-modified"));
+        Assert.assertEquals("/", response.getHeader("Resource-parent"));
```
bencaradocdavies: If your test fixture had more depth, you could test charset handling on the ```Resource-Parent``` header.

src/restconfig/src/test/java/org/geoserver/rest/resources/ResourceTest.java
```diff
@@ -0,0 +1,310 @@
...
+import org.w3c.dom.Document;
+
+import com.mockrunner.mock.web.MockHttpServletResponse;
+
+/**
+ * 
+ * @author Niels Charlier
+ *
+ */
+public class ResourceTest extends GeoServerSystemTestSupport {
```
bencaradocdavies: @NielsCharlier I do not see a test that output format can be detected from request file extension.

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceResource.java
NielsCharlier: there is only one call to valueOf in the code and there would never be need for any other.
If these were lowercase I would still need to call toLowerCase() on the provided argument because the point is that it is not case sensitive

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceFinder.java
NielsCharlier: the restconfig module exists only of nothing but restlet "Finders" and "Resources" with (unfortunately) no class descriptions anywhere

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceResource.java
bencaradocdavies: Sure, @NielsCharlier I later edited this comment to suggest a getOperation method as valueOf cannot be overridden. And I recognise that you will also need trim(). These are suggestions and not mandatory requirements and I will follow your judgement.

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceFinder.java
bencaradocdavies: Got to start somewhere.

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceResource.java
NielsCharlier: I understand it sounds comical. There are over 20 different classes and interfaces known to my eclipse called "Resource". Even in Geoserver we have two meanings for the word Resource.
 I have already made a Spring "resource" that holds a geoserver "Resource" and I decided to call it "SpringResourceAdaptor".

However, in this case, I felt I had little choice but to call it ResourceResource. First, it is not an adaptor. It is a class that holds nearly all of the logic of the resource rest endpoint.

All of the classes in the restconfig module are called SomethingResource and SomethingFinder. Breaking that name pattern would obscure the fact that my two classes do follow the same class pattern.

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceResource.java
NielsCharlier: I hate how eclipse formatting breaks lines up, very hard to read and ugly

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceResource.java
bencaradocdavies: @NielsCharlier as I indicated above, I understand why this class has that name, and recognise that you are being consistent. The name is nice and simple. I ask only for a few words in the class javadoc. Perhaps even using of ```{@link``` to make your job easier? Just something to indicate what it is to future maintainers. I am guessing you might add something like:
```
{@link org.restlet.resource.Resource} to provide operations on {@link org.geoserver.platform.resource.Resource} instances.
```

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceResource.java
NielsCharlier: I always get rid of all warnings when I submit new code, however I have "ignore" set on unavoidable raw types (because of superclasses using raw types). Add the @SuppressWarnings then introduces a warning (Unnecessary @SuppressWarnings("unchecked") 
Personally I always use generics and try to only use the suppresswarnings("unchecked") when I am forced to do a class cast that I know if safe but eclipse doesn't.

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceResource.java
NielsCharlier: no and no

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceResource.java
NielsCharlier: It's a mistake

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceResource.java
bencaradocdavies: @NielsCharlier the best thing about Eclipse formatting is that you can turn it off and on wherever you choose. Our published formatter enables the default off/on tags (in the Code Style / Formatter preference hit Edit and look for the Off/On Tags tab):
```
//@formatter:off
System
  .out
    .println(
"Eclipse will not format this"
    );
//@formatter:on
```
The purpose of the formatter and line length limits is to enhance readability. Eclipse often wraps lines in ways that reduce readability; a well-placed trailing ```//``` or formatter off/on tags is an acceptable solution where you feel this is justified.

src/restconfig/src/main/java/org/geoserver/rest/resources/resourceMetadata.ftl
NielsCharlier: The restlet TemplateRepresentation class sets the output character encoding the same as the input character encoding. So setting <#ftl encoding="utf-8"> should ensure both input as well as output encoding are utf-8 and I can do <meta charset="UTF-8"> as well

(The existing templates are not caring about this...)

src/restconfig/src/test/java/org/geoserver/rest/resources/ResourceTest.java
NielsCharlier: https://github.com/geoserver/geoserver/pull/1487/files#diff-53d3ea4d2b8d4e1363c0415f421cdda9R208

NielsCharlier: The method used for determining the type in a directory listing is the same as for downloading the file itself...

NielsCharlier: Besides, this test also detects from file extension (but not from content, the above does both
): 
https://github.com/geoserver/geoserver/pull/1487/files#diff-53d3ea4d2b8d4e1363c0415f421cdda9R133

NielsCharlier: Besides, this test also detects from file extension (but not from content, the above does both): 
https://github.com/geoserver/geoserver/pull/1487/files#diff-53d3ea4d2b8d4e1363c0415f421cdda9R133

NielsCharlier:
I made (more or less) all of the changes you requested

bencaradocdavies:
@NielsCharlier your new special character tests in paths and content work beautifully on my default locale. Unfortunately if I build GeoServer with ```-Djvm.opts="-Dfile.encoding=US-ASCII"``` it fails in gs-restconfig```ResourceTest``` with:
```
Failed tests:   testCopy(org.geoserver.rest.resources.ResourceTest): expected:<This is my test. [é ö]> but was:<This is my test. [? ?]>
  testMove(org.geoserver.rest.resources.ResourceTest): expected:<This is my test. [é ö]> but was:<This is my test. [? ?]>
  testResource(org.geoserver.rest.resources.ResourceTest): expected:<This is my test. [é ö]> but was:<This is my test. [? ?]>
  testUpload(org.geoserver.rest.resources.ResourceTest): expected:<...his is my new test. [€ è]> but was:<...his is my new test. [? ?]>

Tests in error: 
  testDirectoryHTML(org.geoserver.rest.resources.ResourceTest): Premature end of file.
  testResourceMetadataHTML(org.geoserver.rest.resources.ResourceTest): Premature end of file.
```

For example, to run just this test, in ```src/restconfig```:
```
mvn -o -Djvm.opts="-Dfile.encoding=US-ASCII" -Dtest=ResourceTest test
```

It even fails with ```ISO-8859-15``` which should support all the characters you used.

I am beginning to think that my request for you to include these special characters was a bad idea.

bencaradocdavies:
@NielsCharlier before this PR, all of GeoServer can be built and all tests run with even ```US-ASCII```. I think it would be a major developer regression to lose this.

Questions:

(1) What is the expected behaviour when restconfig is used to store a resource with content that cannot be represented by the default charset?

(2) What is the expected behaviour when restconfig is used to address a path that includes characters that cannot be represented in the charset used for native filesystem metadata?

bencaradocdavies:
@NIelsCharlier unfortunately all the GeoServer test fixtures are tied to a filesystem, and thus the developer native charset for unit tests. What happens when a developer running Windows Cyrillic (```windows-1251```) tries to run this test?

With careful use of ```Assume``` and testing the default charset it might be possible to keep you test portable and preserve your lovely unicode coverage.

NielsCharlier:
Ben, (1) undefined and (2) undefined.

One thing that should work though - is using ISO-8859-15 instead of UTF-8

bencaradocdavies:
@NielsCharlier agreed, ```ISO-8859-15``` should work and I do not understand why it does not. It contains € which ```ISO-8859-1``` does not.

I have seen some European users on the lists running GeoServer on ```ISO-8859-15```; they reported filter encoding problems that were fixed recently. If the unit test does not work for this charset, there is something wrong.

bencaradocdavies:
Testing with ```ISO-8859-15``` breaks only the HTML parsing; something not escaped?

```
Tests in error: 
  testDirectoryHTML(org.geoserver.rest.resources.ResourceTest): Premature end of file.
  testResourceMetadataHTML(org.geoserver.rest.resources.ResourceTest): Premature end of file.
```

NielsCharlier:
Actually, it's a real bug. Might have been introduced by the UTF-8 stuff.

NielsCharlier:
Hmmmm. removing <#ftl encoding="UTF-8"> seems indeed to "fix" it. Hmmmm

bencaradocdavies:
Aha! Your test coverage is worthwhile.

Perhaps you could guard test methods with (or the whole class) with:
```
Assume.assumeTrue(Charset.defaultCharset().newEncoder().canEncode("éö€è"))
```

NielsCharlier:
What a load of ****! So it says in the link you told me about charset issues in freemarker:

"You can give the charset directly by overriding the encoding setting ....

Also, instead of relying on this charset guessing mechanism, you can specify the charset of the template in the template file itself, with the ftl directive, like <#ftl encoding="utf-8">."

So clearly the FTL directive is offered as an *alternative* to specifying it programmatically. However, if  the encoding in the directive of the template file is not EQUAL to the encoding setting (which defaults to system default), an exception is thrown.

In other words, the directive is competely useless. You still have to set it programmatically as well.

bencaradocdavies:
Good grief!

NielsCharlier:
Ben, I find a solution that is even better than Assume, just perform all the same tests without special characters.

The special characters in the filename still work with ASCII. I assume because this does not involve  I/O, and Java uses unicode for its strings no matter what.

bencaradocdavies:
Niels, I saw your new fix and that looks like a good compromise. 

What if  "/poëzie/café" cannot be encoded? Will the file creation in the test throw?

NielsCharlier:
Good point, if the file system doesn't support special characters in the file name it could cause problems. But all modern OS support ë en é in file names.

What do you reckon, shall we go on and see what happens with the geosolutions windows build?

bencaradocdavies:
Aha! I think NTFS uses UCS-2 same as Java so filesystem names may be less likely to throw than file content. So winbuild should be OK.

I was wondering, are there platforms which forbid filenames with those characters?

I am tempted to merge and see.  :-)

Test build running now.

bencaradocdavies:
The dream: http://www.utf8everywhere.org/

bencaradocdavies:
One day every GeoServer instance will use a jdbc resource store for configuration. Standalone instances will use H2. Filesystems path manipulations? How quaint! A thing of the past. And legacy constraints will be left behind.

bencaradocdavies:
@NielsCharlier I am seeing a compilation failure in gs-rest:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:2.3.2:testCompile (default-testCompile) on project gs-rest: Compilation failure: Compilation failure:
[ERROR] /home/ben/geoserver/src with spaces/geoserver/src/rest/src/test/java/org/geoserver/rest/format/ReflectiveHTMLFormatTest.java:[26,63] error: incompatible types: Class<Foo> cannot be converted to String
[ERROR] /home/ben/geoserver/src with spaces/geoserver/src/rest/src/test/java/org/geoserver/rest/format/ReflectiveHTMLFormatTest.java:[43,73] error: incompatible types: Class<ExceptionBean> cannot be converted to String
```

bencaradocdavies:
@NielsCharlier I have confirmed this compile failure in Eclipse. A couple of constructors in the test need to be updated to reflect your change in ```ReflectiveHTMLFormat```. `In gs-restconfig, ``CatalogFreemarkerHTMLFormat``` also needs to be fixed (but Maven gave up before getting that far)..

bencaradocdavies:
Ooh, I see a forced update on your branch. I will try again.

bencaradocdavies:
Looks good. ResourceTest now passes with all the encodings I tested before and more.

NielsCharlier:
@bencaradocdavies wrote: "I was wondering, are there platforms which forbid filenames with those characters?"

Yes, it's called DOS.

PR #1488, opened 2016-02-11
Added specific EnumPPIO for handling enum parameter types.

jdeolive:
I made this change as part of a pull request pending, but I thought I would break it out separately.

### DIFF
--- a/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ppio/ProcessParameterIO.java
+++ b/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ppio/ProcessParameterIO.java
@@ -1,4 +1,4 @@
-/* (c) 2014 - 2015 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
@@ -125,7 +125,7 @@ public abstract class ProcessParameterIO {
     public static ProcessParameterIO find(Parameter<?> p, ApplicationContext context, String mime) {
         // enum special treatment
         if (p.type.isEnum()) {
-            return new LiteralPPIO(p.type);
+            return new EnumPPIO(p.type);
         }
 
         // TODO: come up with some way to flag one as "default"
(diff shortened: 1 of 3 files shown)


aaime:
Travis says no

jdeolive:
Updated the pull request to handle case where the enum might be private or for whatever other reason unaccessible, and fall back to what was the previous behaviour (ie. LiteralPPIO).

jdeolive:
Travis is happy again so unless there is any other feedback on this one I'll assume it's ok to merge.

src/extension/wps/wps-core/src/test/java/org/geoserver/wps/ppio/EnumPPIOTest.java
```diff
@@ -0,0 +1,38 @@
+package org.geoserver.wps.ppio;
```
aaime: Lacks copyright header

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ppio/ProcessParameterIO.java
```diff
@@ -125,7 +125,7 @@
     public static ProcessParameterIO find(Parameter<?> p, ApplicationContext context, String mime) {
         // enum special treatment
         if (p.type.isEnum()) {
-            return new LiteralPPIO(p.type);
+            return new EnumPPIO(p.type);
```
aaime: Copyright header not updated

aaime: Also, findAll is still using LiteralPPIO

aaime:
Minor copyright header issues, and one use case for enums in ProcessParameterIO not tapped. Once that is solved I won't have anything more to add.

jdeolive:
Cool, thanks @aaime. I'll add the header to the test file and update findAll along with a test. And also to clarify whenever modifying an existing file the header needs to be updated to the current year? Is that correct?

aaime:
@jdeolive correct, as per our commit guidelines: https://github.com/geoserver/geoserver/blob/master/CONTRIBUTING.md#commit-guidelines
Doing it by hand is ok for smaller commits, for larger ones if you check geoserver-devel archives some scripts have been circulated to automate the update.

jdeolive:
Cool, thanks. I'll refresh myself on the commit guidelines.

PR #1489, opened 2016-02-11
Added support for importing files along with styling.

jdeolive:
When importing from a file source it checks if the style came with a
style file, and when that is the case tries to use that for the style
of the new layer rather than generating one from scratch.

### DIFF
--- a/src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
+++ b/src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
@@ -1,4 +1,4 @@
-/* (c) 2014 - 2015 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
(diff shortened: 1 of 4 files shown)


src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -1570,6 +1589,48 @@ public XStreamPersister initXStreamPersister(XStreamPersister xp) {
...
+                        StyleInfo info = catalog.getFactory().createStyle();
+
+                        String styleName = styleGen.generateUniqueStyleName(task.getLayer().getResource());
+                        info.setName(styleName);
+                        
+                        info.setFilename(styleName + "." +ext);
+                        info.setFormat(styleHandler.getFormat());
+                        info.setWorkspace(task.getStore().getWorkspace());
+
+                        catalog.getResourcePool().writeStyle(info, styleFile.in());
```
aaime: I believe the input stream will not get closed here, in a cursory look at the code in writeStyle and called methods I did not see the stream being closed.

aaime:
Copyright headers I have not been updated.
Also found an un-closed stream. Besides that, it looks good, I won't have further comments.

jdeolive:
Thanks for the review Andrea. Adding closing of the input stream (oops) and updated copyright headers. I'll go ahead and merge once travis gives the ok.

PR #1493, opened 2016-02-16
Make SecuredGridCoverage2DReader using a cached Crop operation instance

afabiani:
Related to JIRA GEOS-7427


aaime:
One spurious change in importer, no copyright header update. I guess this one is more of a performance optimization and thus cannot be actually "tested" right?

afabiani:
Yep, right.

Spurious change reverted!

aaime:
Copyright header not updated though :)

aaime:
Manually merged and backported to 2.8.x

PR #1495, opened 2016-02-16
Update resource store to use alias to initialize the resourceStore

dvntucker:
Changed the JDBC resourceStore bean to be initialized using alias, which causes it to take precedence over the default resourceStore definition.


jodygarnett:
Update - going to try and fix the proxy, closing this pull request (it can be reopened as a backup plan)

PR #1496, opened 2016-02-17
Updating the ResourcePool to use non-deprecated method

dvntucker:
Changed the ResourcePool to use a non-deprecated method to get the style body. Deprecated method used the file interface, which caused issues when using JDBC resource store.

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
+++ b/src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
@@ -1,4 +1,4 @@
-/* (c) 2014 - 2015 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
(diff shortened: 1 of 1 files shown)


src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -1792,11 +1792,11 @@ public void clear(StyleInfo info) {
      * @return A reader for the style.
      */
     public BufferedReader readStyle( StyleInfo style ) throws IOException {
-        File styleFile = dataDir().findStyleSldFile(style);
-        if( styleFile == null ) {
-            throw new IOException( "No such file: " + style.getFilename() );
+        Resource styleResource = dataDir().style(style);
```
jodygarnett: Can you provide me some details here?  Any of these methods scare me as they often have odd/subtle consequences...

A bit surprised it was missed previously.

jodygarnett:
Change looks clear (and correct) - can you confirm it is covered by one of the many test cases and/or manual testing.

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
dvntucker: I don't think it should scare you too much, just updating to use a non-deprecated method for style body retrieval that uses Resources under the hood. The same method is already being used in a number of other places such as writeStyle in the same class and it's used numerous times in StyleResource.java in the catalog. The method should already be pretty well tested.

dvntucker:
The existing test coverage for ResourcePool is not great, however the underlying method (style) has good test coverage via the GeoserverDataDirectoryTest. I have also tested it manually through my resource store QA.

I can look at writing tests specific to ResourcePool though if need be.

aaime:
Looks fine to me, it's basically an in-lining of the previously called method without creating a intermediate file in the process.

PR #1503, opened 2016-02-18
Expose feature and raster classification processes via wps.

jdeolive:
Adds a PPIO for the custom result returned by the recently added classification processes in GeoTools.  Along with a couple of unit tests. Also includes a modification to the states.properties file to create an actual distribution of data for testing.


src/extension/wps/wps-core/src/main/java/org/geoserver/wps/gs/ClassificationStatsPPIO.java
```diff
@@ -0,0 +1,65 @@
+/* (c) 2016 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2013 OpenPlans
```
jodygarnett: new file 2001?

jodygarnett:
You wanting this in for tomorrows code freeze @jdeolive ?

jodygarnett:
merged to master

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/gs/ClassificationStatsPPIO.java
jdeolive: I actually wrote the original 2-3 years back so kept the original copyright.

PR #1504, opened 2016-02-18
[GEOS-7396]: ClassCastException fix

dromagnoli:


dromagnoli:
Still failing due to travis using oldest jars.

PR #1505, opened 2016-02-18
GEOS-7407 / GSIP 137 ResourceStore Rest API: fix percent escaping

NielsCharlier:


src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceResource.java
```diff
@@ -229,10 +230,27 @@ private String href(String path, boolean isDir) {
         return href;
     }
 
-    public String formatHtmlLink(String link) {
+    private static String formatHtmlLink(String link) {
         return link.replaceAll("&", "&amp;");
     }
 
+    private static String formatPathURL(String input) {
```
bencaradocdavies: @NielsCharlier while this is nice and fast, you have pretty much reimplemented URLEncoder.encode. For maintainability and consistency, please implement this by splitting the path on "/" and then encoding with URLEncoder.encode to UTF-8. This implementation misses several characters like round and square brackets. Why not let  URLEncoder.encode do the work.

Also, as discussed on the mailing list, ResponseUtils.buildURL is sorely in need of this code, so if you use ResponseUtils.buildURL everywhere and fix its implementation to percent encode path elements, you will fix two problems at once.

src/restconfig/src/test/java/org/geoserver/rest/resources/ResourceTest.java
```diff
@@ -223,7 +230,7 @@ public void testDirectoryHTML() throws Exception {
     public void testDirectoryHeaders() throws Exception {
         MockHttpServletResponse response = getAsServletResponse("/rest/resource/mydir?format=xml");
         Assert.assertEquals(FORMAT_HEADER.format(myRes.lastmodified()), response.getHeader("Last-Modified"));
-        Assert.assertEquals("/", response.getHeader("Resource-Parent"));
+        Assert.assertEquals("http://localhost:8080/geoserver/rest/resource/?format=xml", response.getHeader("Resource-Parent"));
```
bencaradocdavies: As discussed on the mailing list, probably best to remove the query.

src/restconfig/src/test/java/org/geoserver/rest/resources/ResourceTest.java
```diff
@@ -149,17 +149,24 @@ public void testResourceMetadataHTML() throws Exception {
...
 
     @Test
     public void testSpecialCharacterNames() throws Exception {
-        MockHttpServletResponse response = getAsServletResponse("/rest/resource/poëzie/café");
+        XMLUnit.setXpathNamespaceContext(NS_XML);
+        Document doc = getAsDOM("/rest/resource/po%c3%abzie?format=xml");
+        XMLAssert.assertXpathEvaluatesTo("http://localhost:8080/geoserver/rest/resource/po%c3%abzie/caf%c3%a9", 
+                "/ResourceDirectory/children/child/atom:link/@href", doc);
+
+        MockHttpServletResponse response = getAsServletResponse("/rest/resource/po%c3%abzie/caf%c3%a9");
```
bencaradocdavies: You could put "?format=xml" on this request to ensure that it is removed in Resource-Parent.

src/restconfig/src/main/java/org/geoserver/rest/resources/ResourceResource.java
NielsCharlier: Ben, URLEncoder.encode also changes spaces into + instead of %20.
Is that something we want?

bencaradocdavies: @NielsCharlier you are quite right. ```URLEncode.encode``` is intended for encoding form data in which spaces should be encoded as "+". In URL paths, spaces should be encoded as "%20". I am not sure which is the best practice code to use in GeoServer.

NielsCharlier: I'd suggest we keep our own method.. It is specific enough to our case to justify it imo. I have improved it though to make sure every character is escaped.

bencaradocdavies:
Thanks @NielsCharlier, good work. You changes will improve the handling of percent-encoding well beyond the resource store.

PR #1506, opened 2016-02-18
[GEOS-7433] Add support for aggregation process group by results

nmco:
This pull request extends the current AggregatorProcess XML XML output format to support the new group by results. A new tabular JSON output format will be added. Backwards compatibility is granted.

Issue: https://osgeo-org.atlassian.net/projects/GEOS/issues/GEOS-7433?filter=myopenissues


nmco:
I found an error on this.

nmco:
Solve.

PR #1514, opened 2016-02-23
[GEOS-7425] Params extractor echo parameter (backport 2.8.x)

nmco:
This pull request adds the possibility to configure in the params-extractor module the echo of existing URL parameters to a get capabilities result. For example, by default the following request:

`/geoserver/ows?service=wms&version=1.3.0&request=GetCapabilities&cql_filter=CFCC=%27D68%27`

will return a get capabilities document were the URLs will be of the type:

`/geoserver/ows?SERVICE=WMS&`

if this module is configured to echo an existing cql_filter parameter the result would be:

`/geoserver/ows?SERVICE=WMS&CQL_FILTER=CFCC%3D%27D68%27&`

This pull request adds proper test cases and also updates the documentation.

Issue: https://osgeo-org.atlassian.net/projects/GEOS/issues/GEOS-7425?filter=allopenissues


jodygarnett:
Okay looks like you are on a mission here, community module so procedure is light. Please advise how you would like to proceed.

simboss:
I will merge as is, if there is anything we'll go back and fix.

PR #1519, opened 2016-02-26
[GEOS-7440] Document the WPS aggregate process

nmco:
Improves the documentation of the WPS aggregate process adding real examples that can be executed againts data shipped by GeoServer release. It also documents the output formats and how they should be interpreted.

### DIFF
--- a/doc/en/user/source/services/wps/processes/gs.rst
+++ b/doc/en/user/source/services/wps/processes/gs.rst
@@ -10,67 +10,232 @@ As with the "geo" processes, the names and definitions of these processes are su
 Aggregation process
 -------------------
 
-The aggregation process is used to perform common aggregation functions (sum, average, count) on vector data.
+The aggregation process is used to perform common aggregation functions (sum, average, count) on vector data. The available outputs formats for this process are `text/xml` and `application/json`. 
 
-WPS aggregate request WITHOUT group by attributes:
+The process parameters are described in the table bellow:
 
-XML:
+.. list-table::
+   :widths: 10 80 5 5
 
-.. code-block:: xml
+   * - **Parameter**
+     - **Description**
+     - **Mandatory**
+     - **Multiple**
+   * - ``features``
+     - Input feature collection.
+     - yes
+     - no
+   * - ``aggregationAttribute``
+     - Attribute on which to perform aggregation.
+     - yes
+     - no
+   * - ``function``
+     - An aggregate function to compute. Functions include Count, Average, Max, Median, Min, StdDev, and Sum.
+     - yes
+     - yes
+   * - ``singlePass``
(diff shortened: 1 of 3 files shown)


jodygarnett:
This looks great, not sure about the heading levels (ie use of "========") but I can fix that during merge.

This inspires me with a great big scary idea - could we factor out the xml requests:

a) Use "include" to include them as code examples?

````
.. literalinclude:: example/WPS_aggregate_1.0.xml
   :language: xml
   :emphasize-lines: 15-18
````

This would let us highlight lines and make the examples easier to test and maintain.

b) use them from the demo request page?

It would be great to have wps examples available, and if we could reuse them in the docs it would be a win/win.

* https://github.com/geoserver/geoserver/tree/master/data/release/demo

I am totally not sure how to do that - it would probably be like relative path madness if the geotools docs are anything to go by:

````
.. literalinclude:: /../../data/release/demos/WPS_aggregate_1.0.xml
   :language: xml
   :emphasize-lines: 15-18
````

File: doc/en/user/source/services/wps/processes/gs.rst [hunk 1]
```diff
@@ -10,67 +10,345 @@ As with the "geo" processes, the names and definitions of these processes are su
...
+    <Max>2.9760021E7</Max>
+    <Average>5038397.020408163</Average>
+    <Sum>2.46881454E8</Sum>
+    <Count>49</Count>
+  </AggregationResults>
+
+
+===============
+ Example 2
+===============
```
jodygarnett: Think this is the "wrong" header based on sphinx conventions - trying to keep things consistent across the different files.

Since the top of the file has:
```
 GeoServer processes
 ===================
```

This would show up in the sphinx table of contents as a new page!

Consider using::

      Aggregate GroupBy Example
      ```````````````````````````

nmco:
Great ideas ! I have updated the pull request with all your suggestions, if I missed something let me know.

nmco:
The build fail seems to be unrelated with this pull request (the build fail is related with gs-libjpeg-turbo module):
```
ERROR] Failed to execute goal on project gs-libjpeg-turbo: Could not resolve dependencies for project org.geoserver.extension:gs-libjpeg-turbo:jar:2.9-SNAPSHOT: Failed to collect dependencies at it.geosolutions.imageio-ext:imageio-ext-turbojpeg:jar:1.1.13 -> org.libjpegturbo:turbojpeg-wrapper:jar:1.2.1.1: Failed to read artifact descriptor for org.libjpegturbo:turbojpeg-wrapper:jar:1.2.1.1: Could not find artifact it.geosolutions.imageio-ext:imageio-ext-library:pom:1.2-SNAPSHOT in boundless (https://repo.boundlessgeo.com/main/) -> [Help 1]
```

PR #1520, opened 2016-02-27
remove javadoc 8 errors

jodygarnett:
Pull request created during 2.9-beta release, addressing issue creating javadocs using Java 8.

The following errors are still open:
````
[ERROR] ows/src/main/java/org/geoserver/ows/util/ResponseUtils.java:263: error: bad use of '>'
[ERROR] ows/src/main/java/org/geoserver/ows/util/ResponseUtils.java:264: error: bad use of '>'
[ERROR] ows/src/main/java/org/geoserver/ows/util/ResponseUtils.java:265: error: bad use of '>'
[ERROR] ows/src/main/java/org/geoserver/ows/util/ResponseUtils.java:266: error: bad use of '>'
[ERROR] ows/src/main/java/org/geoserver/ows/util/ResponseUtils.java:289: error: bad use of '>'
[ERROR] ows/src/main/java/org/geoserver/ows/util/ResponseUtils.java:290: error: bad use of '>'
[ERROR] ows/src/main/java/org/geoserver/ows/util/ResponseUtils.java:291: error: bad use of '>'
[ERROR] ows/src/main/java/org/geoserver/ows/util/ResponseUtils.java:424: error: bad use of '>'
````

### CODE REVIEW

**Comment by jodygarnett** [comment]
Although stack exchanges describes changing the javadoc configuration I have been unable to get it to work:
```
<additionalparam>-Xdoclint:none</additionalparam>
```

aaime:
@jodygarnett not sure this is the way to go, maintenace wise, since our IDEs keep on generating empty param javadocs.

The reason why your configuration did not work is that calling javadoc:javadoc or javadoc:aggregate needs configuration in the build section, while we have it in the reporting section (used for site generation, which we don't do). If you copy/move it to the build section and disable lint via parameter, then javadoc generation works.

Here is another pull request doing the above, https://github.com/geoserver/geoserver/pull/1521  ... but this one does a lot of cleanup, so maybe we should merge them both?

jodygarnett:
Thanks @aaime you are a champ, I will finish up these last errors (since I have gone so far) and then merge them both

PR #1521, opened 2016-02-27
Make javadoc aggregation work with java 8

aaime:
The upgrade to the 2.10 version turned out not to be necessary, but does not hurt either...

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -1475,6 +1475,67 @@
      </configuration>
    </plugin>
 
+<!-- ======================================================= -->
+   <!--     Javadoc generation.                                 -->
+   <!-- ======================================================= -->
+   <plugin>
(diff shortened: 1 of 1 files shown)


src/pom.xml
```diff
@@ -1475,6 +1475,67 @@
...
+      </tag>
+      <tag>
+       <name>tutorial</name>
+       <placement>Xt</placement>
+       <head>Tutorial:</head>
+      </tag>
+     </tags>
+
+     <links>
+      <link>http://docs.oracle.com/javase/7/docs/api/</link>
```
jodygarnett: my other pull request updates some of these URLS to point at java 8 (and removes dead URLs for JAI)

jodygarnett:
merged to master

PR #1523, opened 2016-03-02
(2.8.x) Support for List and Map types in GeoJson

jahhulbert-ccri:
Signed-off-by: Andrew Hulbert <andrew.hulbert@ccri.com>

Previously lists/maps were toString()'d instead of using JSON Arrays and Objects as allowed in the geojson spec:

```
"someMapAttr": "{key=value, foo=bar}",
"someListAttr": "[Edward, Bill, Harry]",
```

New encoding is like this for properties who bindings are java.util.List or java.util.Map:
```
"someMapAttr": {
   "foo": "bar",
   "key": "value"
},
"someListAttr": [
    "Edward",
    "Bill",
    "Harry"
],
```

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
@@ -7,6 +7,8 @@ package org.geoserver.wfs.json;
 
 import java.io.Writer;
 import java.util.Calendar;
+import java.util.List;
+import java.util.Map;
 import java.util.logging.Logger;
 
 import net.sf.json.JSONException;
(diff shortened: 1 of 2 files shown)


jodygarnett:
Looks like you are already hitting this on the devel list - can you make this pull request against master? We are in a code-freeze right now (so we can merge fixes but are holding off on new features).

I think CCRi may have filled in corporate contribution OSGeo agreement, see the [CONTIRBUTING.md](https://github.com/geoserver/geoserver/blob/master/CONTRIBUTING.md) instructions for links.

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
```diff
@@ -286,19 +288,65 @@ public static int getGeometryType(Geometry geometry) {
...
      * @see net.sf.json.util.JSONBuilder#value(java.lang.Object)
      */
     @Override
     public GeoJSONBuilder value(Object value) {
-        if (value instanceof java.util.Date || value instanceof Calendar) {
-            value = Converters.convert(value, String.class);
+        if (value == null) {
+            super.value(value);
+        } else if (value instanceof Geometry) {
+            this.writeGeom((Geometry) value);
```
jodygarnett: Was writeGeom not being used before? May want to check if it uses the correct "curve aware" geotools wkt writer (rather than just calling geometry toString() ).

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
```diff
@@ -286,19 +288,65 @@ public static int getGeometryType(Geometry geometry) {
...
+        } else if (value instanceof List) {
+            this.writeList((List)value);
+        } else if (value instanceof Map) {
+            this.writeMap((Map)value);
+        } else {
+            if (value instanceof java.util.Date || value instanceof Calendar) {
+                value = Converters.convert(value, String.class);
```
jodygarnett: I am a bit concerned that this logic change only special cases Date and Calendar - there are other things like Color that may wish to make use of Converters for a safe text representation...

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
aaime: GeoJSON is not using WKT, and afaik, does not have a notion of curves

jahhulbert-ccri: Andrea is correct, geojson doesn't use wkt but rather native json arrays...and no curves yet though perhaps we could propose them. The spec is here http://geojson.org/geojson-spec.html 

The reason for using the write geometry here is for supporting List<Geometry> or Map which a user could do.

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
jahhulbert-ccri: Sounds good...I'll make that change... 

@jodygarnett I like your idea of exposing other converters for safe text representations that don't rely on the net.sf.json value convert. Perhaps a ticket? Wouldn't be too bad and would make use of the existing converter SPI.

jahhulbert-ccri:
Here's the master PR

https://github.com/geoserver/geoserver/pull/1524 

I'll update both with any changes. Do you normally merge squash or do a single commit only branch from the developer?

jodygarnett:
I usually use rebase and all the commits on master (to better reflect discussion history) and squash for the backport,

jodygarnett:
I am closing this one, if you want to back port https://github.com/geoserver/geoserver/commit/8e85f6a201be1ae165043679a047719855c975f6 to 2.8.x please ask.

PR #1530, opened 2016-03-14
[GEOS-7459] Securing an app-schema data store may result in ClassCastException being thrown

ridethepenguin:
Takes the simplest path and changes the signature of `SecuredDataStoreInfo.getDataStore` to return a generic `DataAccess<FeatureType, Feature>` instance.

Formally, this is a public API change, but shouldn't present a real problem, as the `getDataStore` method in `DataStoreInfo` (the interface decorated by `SecuredDataStoreInfo`) already returns `DataAccess<FeatureType, Feature>`.

### DIFF
--- a/src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/SecuredFeatureChainingTest.java
+++ b/src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/SecuredFeatureChainingTest.java
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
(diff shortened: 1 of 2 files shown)


ridethepenguin:
Full build passes locally, Travis failure seems to be unrelated...

aaime:
Yep, it's an intermittent failure that I've never managed to reproduce locally. Kicked the build again.

aaime:
Hum... the current failure might be legit though? It's a security related test in the WFS module

ridethepenguin:
Seemed spurious to me so I kicked the build again... and this time it failed building the turbojpeg-wrapper module because of missing dependencies... but you were already expecting this, right? :wink:

bencaradocdavies:
@ridethepenguin @aaime I also saw the gs-wfs failure *before* I merged these changes. It looks like an intermittent problem in the gs-wfs module.

I am pleased to confirm that a full local build passes with these changes, and so does a local app-schema online test integration test against postgis, and a full ```-DskipTests -PcommunityRelease``` compile.

@ridethepenguin I agree that the impact of the API change is limited; master is the right place for this fix, and I argue that it is required by GSIP 31.  :-)

Unit test coverage, fixed copyright headers and imports, squeaky clean generics. Thanks @ridethepenguin for this tidy fix.

PR #1531, opened 2016-03-15
Fix typos

dalbani:


aaime:
Hmm... I not 100% sure, but formally I believe you should be following the contribution guide even for such a trivial change: https://github.com/geoserver/geoserver/blob/master/CONTRIBUTING.md
(e.g. contribution agreement, copyright header updates if needed)

dalbani:
@aaime: OK, I wasn't sure about it either, given how small the fixes are.
But I intend to propose larger fixes in the time to come, so I will look at the contribution guide and follow the procedure.

alex3305:
@aaime @dalbani is a member of the Geodan organisation which has a signed contributor agreement submitted to you. However he is a new employee, so maybe he needs to be added to the said agreement.

bencaradocdavies:
@alex3305 in most countries, and I think in the Netherlands too, Geodan will own the contributions of their employees, so @dalbani should be covered by the Geodan corporate contributor agreement. Geodan should update the schedule in the agreement to include @dalbani, but in practice, we usually accept an assertion by an employee that their contributions are on behalf of their employer and are covered by an existing contributor agreement signed by their employer. Strict conformance requires legal advice and knowledge of the internal contracts and processes of Geodan, but this is very cumbersome (I had to do this at a former employer!).

aaime:
Hi @alex3305, looked, the Geodan agreement is indeed there. Merging

PR #1532, opened 2016-03-15
[GEOS-7459] Securing an app-schema data store may result in ClassCastException being thrown

ridethepenguin:
This is the backport of #1530 to the 2.8.x branch.


ridethepenguin:
Travis build failed due to missing dependencies in the tubojpeg-wrapper module...

PR #1534, opened 2016-03-16
Remote wps loadbalancer

afabiani:
Improve the WPS Remote community module with a load balancer getting info from the average resource consumption on the remote processing nodes.

WARNING: must be used with https://pypi.python.org/pypi/wps-remote 2.9.1 or above


aaime:
Merging, all changes are in the community module

PR #1538, opened 2016-03-18
Excel extension module updates

srstclair:
Three very small updates for the Excel module:

* Add a web app profile for Excel
* Add Wicket `wfs.format` labels for Excel output formats
* Update Apache POI to 3.14

I don't believe any of these changes require new test cases.

### DIFF
--- a/src/extension/excel/src/main/resources/GeoServerApplication.properties
+++ b/src/extension/excel/src/main/resources/GeoServerApplication.properties
@@ -1,0 +1,2 @@
+format.wfs.excel=Excel (.xls)
+format.wfs.excel2007=Excel 2007 (.xslx)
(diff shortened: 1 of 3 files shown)


srstclair:
Also, I'm the Excel module maintainer but I believe the proper protocol is to submit a PR as normal? Or should I merge this myself?

aaime:
Hi, as the module maintainer you should be able to just commit, but we appreciate you asking, it's the right thing to do. Generally speaking, if the changes are limited to the module, it's up to you do decide if it's best to socialize what you're doing and ask for feedback, or just go straight ahead. Since we are apparently not yet in feature freeze (last minute schedule change that caused some confusion) I'm going to merge this one directly.
The Travis build failure is unrelated to your change, it's a repository issue that I hope to find the time to fix this weekend

srstclair:
@aaime Thank you. Should I go ahead and backport this to 2.8.x and 2.7.x?

PR #1539, opened 2016-03-21
[GEOS-7463] GeoServer allows for layer specific catalog/services, but not for layer group ones

aaime:


src/main/src/main/java/org/geoserver/ows/LocalWorkspaceCallback.java
```diff
@@ -57,14 +60,20 @@ public Request init(Request request) {
...
-                        }
-                        else {
-                            //TODO: perhaps throw an exception?
+                            LocalPublished.set(l);
+                        } else {
+                            lg = catalog.getLayerGroupByName(ws, last);
+                            if(lg != null) {
+                                LocalPublished.set(lg);
+                            } else {
+                                // TODO: perhaps throw an exception?
```
smithkm: Probably worth a log warning at least.

smithkm:
Looks good to me.

aaime:
So do we want to merge it, or we consider it unfit for the feature freeze? This is in a gray area between a fix and a new feature.

simboss:
I am going to merge this now on 2.10 and then discuss about eventual backport...

PR #1541, opened 2016-03-22
[GEOS-3964] Fix REST Put for primitive types

tbarsballe:
Fix for https://osgeo-org.atlassian.net/browse/GEOS-3964

Added test cases for all affected classes.
Fix MetadataMap for WMSStore Put (previously, WMSStore Put appeared to be broken for **all** PUT requests)
Inject catalog object before reading Xstream request, and fix 'calculateOptionalFields' acordingly.


tbarsballe:
Note: This fix is not especially urgent, and can wait until after the code freeze.

tbarsballe:
It looks like Travis is failing on `org.geoserver.catalog.rest.LayerTest` (which _was_ altered by my changes). However, I cannot reproduce this failure locally (even if I run with the exact same maven commands as Travis).
The error is not especially helpfull (Test failed with an HTTP 500 instead of a HTTP 200), so if anyone is able to reproduce this failure in a dev environment that would be very helpful.

jodygarnett:
Does not fail for me locally, I would consider this to be a rather important fix given what it does to use of the REST API?

src/main/src/main/java/org/geoserver/catalog/impl/StoreInfoImpl.java
```diff
@@ -123,6 +123,9 @@ public void setConnectionParameters(Map<String, Serializable> connectionParamete
     }
 
     public MetadataMap getMetadata() {
+        if (metadata == null) {
+            metadata = new MetadataMap();
```
jodygarnett: If you are going to do this do you need to @sychronize?

src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
```diff
@@ -1506,10 +1511,49 @@ protected void postDoMarshal(Object source,
             }
         }
     }
+    protected class AbstractCatalogInfoConverter extends AbstractReflectionConverter {
```
jodygarnett: Go on give me a class javadoc, since this is protected it is part of the api and I would like to know what you are doing.

src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
```diff
@@ -1506,10 +1511,49 @@ protected void postDoMarshal(Object source,
...
+                }
+            }
+            return emptyObject;
+        }
+    }
     /**
      * Converter for {@link DataStoreInfo}, {@link CoverageStoreInfo}, and {@link WMSStoreInfo}
      */
-    class StoreInfoConverter extends AbstractReflectionConverter {
+    class StoreInfoConverter extends AbstractCatalogInfoConverter {
```
jodygarnett: See your subclass over here has a class javadoc, good subclass have a cookie.

src/ows/src/main/java/org/geoserver/ows/util/OwsUtils.java
```diff
@@ -420,8 +420,10 @@ else if ( Set.class.isAssignableFrom( type ) ) {
     static void updateCollectionProperty(Object object, Collection newValue, Method getter)
             throws Exception {
         Collection oldValue = (Collection) getter.invoke(object, null);
-        oldValue.clear();
-        oldValue.addAll(newValue);
+        if (oldValue != null) {
+            oldValue.clear();
+            oldValue.addAll(newValue);
```
jodygarnett: This would avoid setting the newValue? Not sure you are meeting the java bean getter/setter api contract for oldValue?

tbarsballe: This is called by OwsUtils.copy, which does a reflective copy based on the setters and getters of an interface. This particular code is only called in the case we have a collection with a getter but no setter. Consequently, if oldValue is null, we have no reasonable way of accessing the underlying collection (since we are dealing only with the interface).

Affected values include ResourceInfo.getMetadataLinks and ResourceInfo.getKeywords(), among others.

Fortunately, this only occurs if when we are copying into a reflectively created object (since normally these collections are initialized during a proper instantiation. Since the XStream implementation deals wiht the actual object classes instead of the interface classes, if any of these are set in the PUT request they should be correctly parsed into the final object. Otherwise, the values remain null and do not get copied into the final catalog object.

I can update the javadocs / comments to make this clearer.

Also note that previously, we would have just gotten an NPE if oldValue was null.

src/restconfig/src/main/java/org/geoserver/catalog/rest/CoverageResource.java
```diff
@@ -150,8 +151,9 @@ protected void handleObjectPut(Object object) throws Exception {
         
         CoverageStoreInfo cs = catalog.getCoverageStoreByName(workspace, coveragestore);
         CoverageInfo original = catalog.getCoverageByCoverageStore( cs,  coverage );
-        new CatalogBuilder(catalog).updateCoverage(original,c);
         calculateOptionalFields(c, original);
+        
+        new CatalogBuilder(catalog).updateCoverage(original,c);
```
jodygarnett: Are you really just moving calculateOptionalFields before updating the coverage?

If you add this "calculateOptionalFields" to AbstractCatalogResource API (ie as part of the resource lifecycle) can CatalogBuilder make use of the method to make this occur in a consistent order?

src/restconfig/src/main/java/org/geoserver/catalog/rest/DataStoreResource.java
```diff
@@ -179,6 +180,16 @@ protected void configurePersister(XStreamPersister persister, DataFormat format)
         persister.setCallback( 
             new XStreamPersister.Callback() {
                 @Override
+                protected CatalogInfo getCatalogObject() {
```
jodygarnett: You are having to override this method in a lot of classes (DataStoreResource, FeatureTypeResource, CoverageStoreResource, CoverageResource) - what was wrong with the previous implementation?

src/restconfig/src/test/java/org/geoserver/catalog/rest/CoverageStoreTest.java
```diff
@@ -373,8 +392,8 @@ public void testDeletePurgeAllAfterConfigure() throws Exception {
 
     @Test
     public void testPut2() throws Exception {
-        Document dom = getAsDOM( "/rest/workspaces/wcs/coveragestores/BlueMarble.xml");
-        assertXpathEvaluatesTo("GeoTIFF", "/coverageStore/type", dom );
+        //Document dom = getAsDOM( "/rest/workspaces/wcs/coveragestores/BlueMarble.xml");
+        //assertXpathEvaluatesTo("GeoTIFF", "/coverageStore/type", dom );
```
jodygarnett: commenting out test cases makes me sad, can you use @Disable?

jodygarnett:
Thanks for pull request, I expect you can answer my questions (fill in comments etc...) and merge. Despite my feedback/questions this appears to be an important fix to get in.

One question? Do you expect to backport this fix to the stable branch?

src/restconfig/src/main/java/org/geoserver/catalog/rest/CoverageResource.java
tbarsballe: calculateOptionalFields itself also got heavily overhauled (see above). The method is part of the restconfig resource implementation, and appears to have origionally been written primarily with REST transactions in mind. See https://github.com/geoserver/geoserver/commit/558e4bad12ecdb596ead3a695cfabc8f7ee162a1 for more info.

src/restconfig/src/main/java/org/geoserver/catalog/rest/DataStoreResource.java
tbarsballe: I added this method to the abstract superclass in XStreamPersister.Callback in the first place, so there was no previous implementation.
If you look in the Callback class, it is basically a bunch of empty methods that get overriden and are called during resource handling. This one is intended to get the proper catalog object for a given resource.

 If you look at the other overrriden methods below, you will see they folow the same coding pattern.

src/restconfig/src/test/java/org/geoserver/catalog/rest/CoverageStoreTest.java
tbarsballe: That was actually just something I forgot to uncomment after debugging, thanks for catching it.

aaime:
Wanted to have a look locally before rushing in a request that fails on the build server, but the output of the build server is not a test failure... it's a compile failure in there?
From the logs:

````
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:2.3.2:testCompile (default-testCompile) on project gs-restconfig: Compilation failure: Compilation failure:
[ERROR] /home/travis/build/geoserver/geoserver/src/restconfig/src/test/java/org/geoserver/catalog/rest/CoverageStoreTest.java:[289,35] error: cannot find symbol
[ERROR] variable response of type MockHttpServletResponse
[ERROR] /home/travis/build/geoserver/geoserver/src/restconfig/src/test/java/org/geoserver/catalog/rest/DataStoreTest.java:[308,35] error: cannot find symbol
[ERROR] variable response of type MockHttpServletResponse
[ERROR] /home/travis/build/geoserver/geoserver/src/restconfig/src/test/java/org/geoserver/catalog/rest/WMSLayerTest.java:[314,35] error: cannot find symbol
[ERROR] variable response of type MockHttpServletResponse
[ERROR] /home/travis/build/geoserver/geoserver/src/restconfig/src/test/java/org/geoserver/catalog/rest/LayerTest.java:[129,35] error: cannot find symbol
[ERROR] variable response of type MockHttpServletResponse
[ERROR] /home/travis/build/geoserver/geoserver/src/restconfig/src/test/java/org/geoserver/catalog/rest/CoverageTest.java:[407,35] error: cannot find symbol
[ERROR] variable response of type MockHttpServletResponse
[ERROR] /home/travis/build/geoserver/geoserver/src/restconfig/src/test/java/org/geoserver/catalog/rest/WMSStoreTest.java:[227,35] error: cannot find symbol
[ERROR] variable response of type MockHttpServletResponse
[ERROR] /home/travis/build/geoserver/geoserver/src/restconfig/src/test/java/org/geoserver/catalog/rest/WMSStoreTest.java:[253,35] error: cannot find symbol
[ERROR] variable response of type MockHttpServletResponse
[ERROR] /home/travis/build/geoserver/geoserver/src/restconfig/src/test/java/org/geoserver/catalog/rest/LayerGroupTest.java:[390,35] error: cannot find symbol
[ERROR] variable response of type MockHttpServletResponse
[ERROR] /home/travis/build/geoserver/geoserver/src/restconfig/src/test/java/org/geoserver/catalog/rest/FeatureTypeTest.java:[367,35] error: cannot find symbol
````

Indeed I have the same compile failures locally. Wondering if this one was thrown off base by the spring 4 upgrade merges?

**Comment by tbarsballe** [comment]
Good, I am not the only one! I was seeing these same failures last week, but @jodygarnett was not able to reproduce. I have since rebased against the current master and all compile failures have gone away. I will be commiting that + fixes based on Jody's feedback shortly.

File: src/main/src/main/java/org/geoserver/catalog/impl/StoreInfoImpl.java (hunk 1 above)
Comment (tbarsballe) [inline src/main/src/main/java/org/geoserver/catalog/impl/StoreInfoImpl.java]: You are right, this is not treadsafe. I will add `synchronized`.

**Comment by tbarsballe** [comment]
Updated, rebased and pushed. Hopefull Travis will be happy with it this time around.
@jodygarnett While this fix is definately important, I am not sure that it is especially urgent. Workarounds exist, and the bug itself has been around since GeoServer 2.0. As for backporting to stable, the changes are significant enough, and have enough of a potential for subtle behavioral changes that I am a bit leery about backporting. Plus, with the spring changes, I think some minor rewriting is necessary to backport. So, if we do backport to stable, it is probably best to wait a bit.

**Comment by aaime** [comment]
Hi Travis, now it compiles. I've found out that if I run the tests from Eclipse, they pass with OpenJDK but fail with Oracle JDK. Maven seems to be doing the same and reports the following:

````
Failed tests:   testPutWorkspaceStyle(org.geoserver.catalog.rest.LayerTest): expected:<200> but was:<500>
  testUpdateStyleJSON(org.geoserver.catalog.rest.LayerTest): expected:<200> but was:<500>
  testPut(org.geoserver.catalog.rest.LayerTest): expected:<200> but was:<500>
  testPutNonDestructive(org.geoserver.catalog.rest.LayerTest): expected:<200> but was:<500>
  testPutWorkspaceAlternateStyle(org.geoserver.catalog.rest.LayerTest): expected:<200> but was:<500>
````

The errors I get all look like:

````
31 mar 19:38:33 ERROR [geoserver.rest] - 
com.thoughtworks.xstream.converters.ConversionException: java.lang.reflect.InvocationTargetException : java.lang.reflect.InvocationTargetException
---- Debugging information ----
message             : java.lang.reflect.InvocationTargetException
cause-exception     : java.lang.RuntimeException
cause-message       : java.lang.reflect.InvocationTargetException
class               : org.geoserver.catalog.impl.LayerInfoImpl
required-type       : org.geoserver.catalog.impl.LayerInfoImpl
converter-type      : org.geoserver.config.util.XStreamPersister$LayerInfoConverter
line number         : 1
version             : not available
-------------------------------
	at com.thoughtworks.xstream.core.TreeUnmarshaller.convert(TreeUnmarshaller.java:79)
	at com.thoughtworks.xstream.core.TreeUnmarshaller.convertAnother(TreeUnmarshaller.java:66)
	at com.thoughtworks.xstream.core.TreeUnmarshaller.convertAnother(TreeUnmarshaller.java:50)
	at com.thoughtworks.xstream.core.TreeUnmarshaller.start(TreeUnmarshaller.java:134)
	at com.thoughtworks.xstream.core.AbstractTreeMarshallingStrategy.unmarshal(AbstractTreeMarshallingStrategy.java:32)
	at com.thoughtworks.xstream.XStream.unmarshal(XStream.java:1185)
	at com.thoughtworks.xstream.XStream.unmarshal(XStream.java:1169)
	at com.thoughtworks.xstream.XStream.fromXML(XStream.java:1049)
	at org.geoserver.config.util.XStreamPersister.load(XStreamPersister.java:599)
	at org.geoserver.catalog.rest.AbstractCatalogResource$1.read(AbstractCatalogResource.java:51)
	at org.geoserver.rest.format.StreamDataFormat.toObject(StreamDataFormat.java:34)
	at org.geoserver.rest.ReflectiveResource.handlePut(ReflectiveResource.java:192)
	at org.restlet.Finder.handle(Finder.java:298)
	at org.geoserver.rest.BeanDelegatingRestlet.handle(BeanDelegatingRestlet.java:38)
	at org.restlet.Filter.doHandle(Filter.java:105)
	at org.restlet.Filter.handle(Filter.java:134)
	at org.restlet.Router.handle(Router.java:444)
	at org.geoserver.rest.RESTDispatcher$1.handle(RESTDispatcher.java:205)
	at com.noelios.restlet.ext.servlet.ServletConverter.service(ServletConverter.java:129)
	at org.geoserver.rest.RESTDispatcher.handleRequestInternal(RESTDispatcher.java:87)
	at org.springframework.web.servlet.mvc.AbstractController.handleRequest(AbstractController.java:147)
	at org.springframework.web.servlet.mvc.SimpleControllerHandlerAdapter.handle(SimpleControllerHandlerAdapter.java:50)
	at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:959)
	at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:893)
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:968)
	at org.springframework.web.servlet.FrameworkServlet.doPut(FrameworkServlet.java:881)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:691)
	at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:844)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:770)
	at org.geoserver.test.GeoServerSystemTestSupport$1.service(GeoServerSystemTestSupport.java:1452)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:770)
	at org.springframework.mock.web.MockFilterChain$ServletFilterProxy.doFilter(MockFilterChain.java:167)
	at org.springframework.mock.web.MockFilterChain.doFilter(MockFilterChain.java:134)
	at org.geoserver.test.GeoServerSystemTestSupport.dispatch(GeoServerSystemTestSupport.java:1479)
	at org.geoserver.test.GeoServerSystemTestSupport.dispatch(GeoServerSystemTestSupport.java:1410)
	at org.geoserver.test.GeoServerSystemTestSupport.dispatch(GeoServerSystemTestSupport.java:1398)
	at org.geoserver.test.GeoServerSystemTestSupport.putAsServletResponse(GeoServerSystemTestSupport.java:1063)
	at org.geoserver.test.GeoServerSystemTestSupport.putAsServletResponse(GeoServerSystemTestSupport.java:1051)
	at org.geoserver.catalog.rest.LayerTest.testPutWorkspaceStyle(LayerTest.java:210)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:271)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:70)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:50)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:238)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:63)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:236)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:53)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:229)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:309)
	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestReference.run(JUnit4TestReference.java:86)
	at org.eclipse.jdt.internal.junit.runner.TestExecution.run(TestExecution.java:38)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:459)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:675)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:382)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.main(RemoteTestRunner.java:192)
Caused by: java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
	at org.geoserver.ows.util.OwsUtils.copy(OwsUtils.java:350)
	at org.geoserver.config.util.XStreamPersister$AbstractCatalogInfoConverter.unsafeCopy(XStreamPersister.java:1526)
	at org.geoserver.config.util.XStreamPersister$AbstractCatalogInfoConverter.instantiateNewInstance(XStreamPersister.java:1554)
	at com.thoughtworks.xstream.converters.reflection.AbstractReflectionConverter.unmarshal(AbstractReflectionConverter.java:256)
	at com.thoughtworks.xstream.core.TreeUnmarshaller.convert(TreeUnmarshaller.java:72)
	... 66 more
Caused by: java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.geoserver.ows.util.OwsUtils.copy(OwsUtils.java:348)
	... 70 more
Caused by: java.lang.NullPointerException: Layer enabled flag must not be set without an underlying resource
	at org.geoserver.catalog.impl.LayerInfoImpl.setEnabled(LayerInfoImpl.java:221)
	... 75 more
````

Hope this helps

**Comment by tbarsballe** [comment]
```
@Override    
    public void setName(String name) {
        // TODO: remove this log and reinstate field assignment when resource/publish split is complete
        LOGGER.log(Level.FINE, "Warning, some code is setting the LayerInfo name, but that will be ignored");
        this.name = name;
        
        if (resource == null) {
            throw new NullPointerException("Layer name must not be set without an underlying resource");
        }
        resource.setName(name);
    }
```

My best guess is OwsUtils.copy is trying to copy the name before the resource, although if this was the case I am surprised this has not come up earlier. I am using Oracle JDK, and the tests pass in Eclipse for me (looks like Travis passes as well).

I've pushed something that should fix this (https://github.com/geoserver/geoserver/pull/1541/commits/8a5c9fabe949f1bba9edaa04329537b593680243), but since I can't reproduce the issue am unable to test. Could you test if this fix works on your setup @aaime ?

aaime:
Hi, the build works with that change. Had a look at the diff, some parts makes my skin crawl in terms of committing it as part of the 2.9.x release, it seems overly risky.
For example, creating on the fly the metadata map should not be necessary, it's known that XStream does not populate those fields and we have custom methods in the catalog to populate them instead (see the resolve methods and their call to resolveCollection at https://github.com/geoserver/geoserver/blob/master/src/main/src/main/java/org/geoserver/catalog/impl/CatalogImpl.java#L1561
On one side it would seem the new code would make those un-necessary as the converter code now instantiated the object via normal constructors (does that mean we don't need the resolve methods anymore?), and on the other side we have the metadata map strangely created on the fly I assume to avoid NPEs (but why just for StoreInfoImp, most catalog object impl have a metadata map?), as well as some null checks which might make one think the collections could be null (live collections should just never be).

My suggestion: do not merge for 2.9.x, see if we can have @jdeolive look at it before merging, and make sure that no live collection can ever be null.

jdeolive:
Took a quick look and don't have much more to add than Andrea already has. Agreed that this patch touches a few places with a lot of subtle logic... I too would be hesitant to commit this during a stabilization period and would probably wait until the next unstable branch becomes available.

tbarsballe:
Glad that fix works.
I fully agree that there a bunch of risky changes in this one, and am happy to wait until after the 2.9.x release to commit this.

The metadata map bit was added specifically because of NPEs thrown by reflectively created instances of WMSStoreInfoImpl (which assumes MetadataMap is never null - see setUseConnectionPooling). Perhaps adding some null checking there instead would be better.

PR #1542, opened 2016-03-27
[GEOS-7474] Speed up usage of git-commit-it

aaime:
On my computer this single change cuts the release build time from 20:43 down to 19:13, or a minute and a half speedup.

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -1469,7 +1469,7 @@
    <plugin>
      <groupId>pl.project13.maven</groupId>
      <artifactId>git-commit-id-plugin</artifactId>
-     <version>2.1.13</version>
+     <version>2.1.15</version>
      <executions>
        <execution>
          <goals>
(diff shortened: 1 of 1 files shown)


aaime:
Also note this ticket opened at the git-commit-id github: https://github.com/ktoso/maven-git-commit-id-plugin/issues/242

aaime:
Unfortunately with runOnlyOnce the build.commit.id variable is not exported in the subsequent runs, meaning the jars won't contain the revision number in the META-INF/MANIFEST file. Changed it to make it optional (someone might want the extra speed and does not care about the revision ids) and opened a ticket against git-commit-id here:

https://github.com/ktoso/maven-git-commit-id-plugin/issues/243

aaime:
Disregard my last comment, I've found a way to export the commit id in all projects in the build, it just needed an extra setting.

bencaradocdavies:
@aaime if it helps, feel free to bump git-commit-id-plugin to 2.2.1 as Maven on ares Jenkins has been upgraded to 3.3.9 and we can now support plugins later than 2.1.13.

aaime:
@bencaradocdavies for now it seems we're good, and I'm not sure everybody is using maven 3.3.x on their dev machines. We can reconsider if/when they fix https://github.com/ktoso/maven-git-commit-id-plugin/issues/242

ianturton:
shaves 30 seconds of my build from 7:09 to 6:34 - seems fine on ubuntu 14.04, Oracle Java 1.8 and mvn 3.3.9

PR #1546, opened 2016-03-31
[GEOS-7480] - NetCDF Importer Fixes

afabiani:
Related to Improvement JIRA GEOS-7480


aaime:
Community module only changes, does not break the build, made by the module owner... good to go :-)

PR #1548, opened 2016-04-01
Use standard UTF-8 encoding for 'é' character

Vampouille:


Vampouille:
This break compilation of geoserver:

```
java.nio.file.InvalidPathException: Malformed input or input contains unmappable chacraters: /home/jacroute/git/georchestra/geoserver/geoserver-submodule/data/citewfs-1.1/workspaces/sf/sf/Entite??Ge??ne??rique
	at sun.nio.fs.UnixPath.encode(UnixPath.java:147)
	at sun.nio.fs.UnixPath.<init>(UnixPath.java:71)
```

bencaradocdavies:
@Vampouille these filename é characters are represented by a perfectly valid UTF-8 byte sequence (0x65 0xCC 0x81, 'e' plus U+0301 'COMBINING ACUTE ACCENT') <http://unicode.org/cldr/utility/character.jsp?a=0301>.

What is your platform character encoding?

Vampouille:
I want to rename this folder because it breaks maven git-commit-id-plugin. This plugin fails with above error. I'm using a ubuntu/wily with UTF-8 as default encoding.
On my system, when I press 'é' key, it produce 0xc3 0xa9 which is a very common encoding for 'é'. If you take a look at the documentation in geoserver : https://github.com/geoserver/geoserver/blob/master/doc/fr/user/source/webadmin/basics.rst, you will only find 'é' encoded as 0xc3 0xa9 and no 'é' encoded as 0x65 0xCC 0x81 at all. No file in geoserver repository uses 'é'. The `data/citewfs-1.1-h2/dataset-sf0.sql` file uses 'é' encoded as 0xc3 0xa9 to represent "Entité générique" folder : https://github.com/geoserver/geoserver/commit/44bacafcf1352bfe0b1836262df686fd58036585#diff-e74b110766ce02c2c439fcc3dbf924afR61
Using 'é' encoded as 0x65 0xCC 0x81 breaks the layout of page:
https://github.com/geoserver/geoserver/commit/44bacafcf1352bfe0b1836262df686fd58036585#diff-516a7bf5ff4e7c508e953085f9f6407b (accents are on the 'g' letter instead of 'e')

So even if it's a valid encoding for é, I would prefer to use a more popular encoding for 'é' in UTF-8 : 0xc3 0xa9.
screenshot from 2016-04-04 12-20-46

aaime:
Weird, geoserver is using git-commit-id without issues?

bencaradocdavies:
@Vampouille this file name exists to catch problems handling unicode. It would be strange then to force it to use an ISO-8859-15 subset. What happens if you change the filename to 通用实体 (Google translate of generic entity into simplified chinese)? Does git-commit-id-plugin still fail? What happens if you build in /家/用户 (/home/user via Google translate)? I would like to know if there is something in your build that round-trips the filename through a non-unicode charset. Can we also please see the full stack trace?

@aaime one difference is that georchestra includes this file under the top-level build directory, but because GeoServer is built from ```src```, this file may be outside the tree examined by git-commit-id-plugin.

bencaradocdavies:
If I downgrade to GeoServer git-commit-id-plugin 2.1.10 as in georchestra and copy the citewfs-1.1 directory under src, I still see no failure.

bencaradocdavies:
@Vampouille have you had a chance to look at my questions above?

jodygarnett:
Hello @Vampouille we were hoping for an update on this one for todays GeoServer meeting - any news?

jodygarnett:
It sounds like this test is correctly checking expected functionality, it is just that your system is being a bad operating environment right?

Can we provide a -D flag to disable this test if needed?

bencaradocdavies:
Closing as no response from submitter. Happy to reopen if comments above addressed.

PR #1540, opened 2016-03-22
[GSIP-144] LDAP UserGroupService + refactoring to share code with RoleService

NielsCharlier:
Please see https://github.com/geoserver/geoserver/wiki/GSIP-144

PR #1550, opened 2016-04-04
Spring Security: fix LDAP AuthenticationProvider

NielsCharlier:


bencaradocdavies:
@NielsCharlier, copyright header? And should this be against master? I think this branch is officially dead.

NielsCharlier:
I did not know this.

bencaradocdavies:
The merge was discussed in this email thread:
http://osgeo-org.1560.x6.nabble.com/time-to-merge-spring-4-update-td5257874.html

I do not know if it was explicitly stated, but my assumption is that a feature branch is dead after it is merged. AFAIK all future fixes are being applied master, which is good as they are covered by Jenkins.

PR #1540 was also mentioned in the email thread (but you also discussed it on that PR so I know you are aware).

PR #1552, opened 2016-04-04
Determine JAVA_HOME on Debian/Ubuntu

terokorp:


src/release/bin/startup.sh
```diff
@@ -5,6 +5,11 @@
 # $Id$
 # -----------------------------------------------------------------------------
 
+# Determine JAVA_HOME on Debian/Ubuntu
+if [ -z "$JAVA_HOME" ]; then
+  JAVA_HOME=$(readlink -fe /usr/bin/java | sed "s:bin/java::")
```
bencaradocdavies: This is elegant because, if the link is missing or broken, ```JAVA_HOME``` will stay empty. I like this fix because I am always annoyed by having to manually set ```JAVA_HOME``` when starting the bin.zip.

My only remark is that, in ```readlink -fe```, the "f" is redundant.

bencaradocdavies:
Another problem is that ```readlink``` on MacOS does supports neither ```-f``` nor ```-e```. Also, ```sed``` on MacOS has some eccentricities.

Another solution (non-MacOS) is ```$(dirname $(dirname $(readlink -f $(which java))))```, with similar MacOS version here:
See http://stackoverflow.com/questions/1117398/java-home-directory

bencaradocdavies:
TL;DR: great idea, not quite ready as would break MacOS.

bencaradocdavies:
A better question is, why do we even want to determine ```JAVA_HOME```? It is only used to locate ```java```. Once ```java``` is running, it can knows its own ```System.getProperty("java.home")``` and AFAIK requires no environment variables. We could just modify ```startup.sh``` to use ```_RUNJAVA=java``` if ```$(which java)``` is nonzero and only consider ```JAVA_HOME``` otherwise.

bencaradocdavies:
I have an alternative solution in https://github.com/geoserver/geoserver/pull/1553 which avoids setting ```JAVA_HOME``` and falls back to the path. It also fixes ```shutdown.sh```.

terokorp:
Yea, #1553 seems to be better. I will close this one.

bencaradocdavies:
@terokorp, #1553 was just merged by @jodygarnett. Please try a subsequent nightly build (or 2.9-beta2 coming very soon) to see if it works for you. Thank you for your pull request, even though it was not merged. This was a necessary fix and we needed your prompting.  :-)

PR #1553, opened 2016-04-05
Detect Java runtime from JAVA_HOME or PATH

bencaradocdavies:
Alternative to https://github.com/geoserver/geoserver/pull/1552 that should be more general and not break MacOS.

### DIFF
--- a/src/release/bin/shutdown.sh
+++ b/src/release/bin/shutdown.sh
@@ -5,19 +5,27 @@
 # $Id$
 # -----------------------------------------------------------------------------
 
-# Make sure prerequisite environment variables are set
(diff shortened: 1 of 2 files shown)


bencaradocdavies:
One downside of this change is that the executable name will be ```java``` in the output of ```ps```, not the canonical path name. On Linux the full path name of the executable image can be found with ```ls -l /proc/$PID/exe```.

bencaradocdavies:
Thanks @jodygarnett. Good to get this one into 2.9-beta2. I am always forgetting to set ```JAVA_HOME``` when testing.

PR #1556, opened 2016-04-06
GEOS-7514 Spring Security: fix LDAP AuthenticationProvider (attempts to use non…

NielsCharlier:
…-ldap login to fetch groups from ldap server)


aaime:
So this one is part of the spring4 related upgrades? Do we have to merge it before beta2?

NielsCharlier:
Yes and preferrably

NielsCharlier:
I assume this can be merged?

PR #1557, opened 2016-04-06
[GEOS-7487] Upgrades JQuery version and replaces the inline-info plugin

nmco:
I also removed unused uses of JQuery and the inline-info plugin.

Issue: https://osgeo-org.atlassian.net/projects/GEOS/issues/GEOS-7487?filter=allopenissues
Mailing list discussion: http://osgeo-org.1560.x6.nabble.com/Geofence-drag-and-drop-and-JQuery-version-td5259780.html


src/web/core/src/main/java/org/geoserver/web/GeoServerBasePage.java
```diff
@@ -223,6 +224,10 @@ protected String load() {
 
     @Override
     public void renderHead(IHeaderResponse response) {
+
+        // includes jquery, rehired by the placeholder plugin
```
aaime: required maybe?

nmco: oops you are correct ... I don't really know why and how I wrote "rehired" instead of "required"

PR #1560, opened 2016-04-08
[GSIP-143] Module status

mtCarto:
Initial pull request with ModuleStatus interface and implementation. Will be adding docs soon for the REST endpoints. 

Any feedback is much appreciated.

### DIFF
--- a/doc/en/user/source/rest/api/manifests.rst
+++ b/doc/en/user/source/rest/api/manifests.rst
@@ -7,6 +7,7 @@ GeoServer provides a REST service to expose a listing of all loaded JARs and res
 
 * ``about/manifest``—Retrieves details on all loaded JARs
 * ``about/version``—Retrieves details for the high-level components: GeoSever, GeoTools, and GeoWebCache
+* ``about/status``-Retrieves details for the status of all loaded and configured modules
 
 
 ``/about/manifest[.<format>]``
@@ -126,7 +127,46 @@ This endpoint shows only the details for the high-level components: GeoServer, G
      - HTML
      - :ref:`manifest <rest_api_manifests_manifest>`, :ref:`key <rest_api_manifests_key>`, :ref:`value <rest_api_manifests_value>`
    * - POST
-     - 
+     -
+     - 405
+     -
+     -
+     -
+   * - PUT
+     -
+     - 405
+     -
+     -
+     -
+   * - DELETE
+     -
+     - 405
+     -
+     -
+     -
+
+``/about/status[.<format>]``
+-----------------------------
+
+This endpoint shows the status details of all installed and configured modules.Status details always include human readable name, and module name. Optional details include version, availability, status message, and links to documentation.
+
+.. list-table::
+   :header-rows: 1
+
+   * - Method
+     - Action
+     - Status Code
+     - Formats
+     - Default Format
+     - Parameters
+   * - GET
+     - List module statuses
+     - 200
+     - HTML, XML, JSON
+     - HTML
+     -
+   * - POST
+     -
      - 405
      - 
      - 
--- a/src/extension/wps/web-wps/src/main/resources/applicationContext.xml
+++ b/src/extension/wps/web-wps/src/main/resources/applicationContext.xml
@@ -1,10 +1,15 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Copyright (c) 2001 - 2013 OpenPlans - www.openplans.org. All rights reserved. 
+<!-- Copyright (c)
+  2014 - 2016 Open Source Geospatial Foundation. All rights reserved.
+  2001 - 2013 OpenPlans - www.openplans.org.
   This code is licensed under the GPL 2.0 license, available at the root application 
   directory. -->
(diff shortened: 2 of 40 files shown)


bradh:
Is this https://github.com/geoserver/geoserver/wiki/GSIP-143  / https://osgeo-org.atlassian.net/browse/GEOS-6068 ?

mtCarto:
Yes, this is addresses GSIP-143 and GEOS-6068

src/platform/src/main/java/org/geoserver/platform/ModuleStatusImpl.java
```diff
@@ -0,0 +1,124 @@
...
+        this.module = status.getModule();
+        this.component = status.getComponent();
+        this.name = status.getName();
+        this.version = status.getVersion();
+        this.documentation = status.getDocumentation();
+        this.message = status.getMessage();
+    }
+
+    public ModuleStatusImpl(String description) {
+        String split[] = description.split(",");
```
smithkm: Why are you passing in a comma separated string rather than two separate strings?

src/main/src/test/java/org/geoserver/ModuleStatusTest.java
```diff
@@ -0,0 +1,21 @@
+package org.geoserver;
```
smithkm: Unit tests should really have copyright headers too.  W aren't always as picky about them as about implementation files but really they should be everywhere.

src/platform/src/main/java/org/geoserver/platform/ModuleStatusImpl.java
```diff
@@ -0,0 +1,124 @@
...
+     */
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    /**
+     * @return the version
+     */
+    public String getVersion() {
+        return version;
```
smithkm: Since this may not have been initialized it could be `null`.  The interface doesn't give any warning about that via docs or `@Nullable` annotation.  I'd prefer to see a clear requirement in the interface that these methods not return `null` with this implementation being updated to match, use the new Java 8 `Optional` object, or clearly mark them as returning `null` so that people using the value are warned that they need to do null checks.

doc/en/user/source/rest/api/manifests.rst
```diff
@@ -34,15 +35,15 @@ The available types are ``core``, ``extension``, or ``community``. To filter mod
      - HTML
      - :ref:`manifest <rest_api_manifests_manifest>`, :ref:`key <rest_api_manifests_key>`, :ref:`value <rest_api_manifests_value>`
    * - POST
-     - 
+     -
```
smithkm: You have a bunch of what appear to be non-semantic whitespace changes here.

src/platform/src/main/java/org/geoserver/platform/RenderingEngineStatus.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Override
+    public String getName() {
+        return "Rendering Engine";
+    }
+
+    @Override
+    public Optional<String> getVersion() {
+        if (provider.contains("OracleJDK")) {
+            return version.ofNullable(System.getProperty("java.version"));
```
smithkm: Why do you have `Optional` instance `version` if all you ever use it for is to call these static methods?

src/restconfig/src/main/java/org/geoserver/rest/AboutStatus.java
```diff
@@ -0,0 +1,125 @@
...
+                        SimpleHash hash = new SimpleHash();
+                        hash.put( "values", new CollectionModel( list, new BeansWrapper(){
+                            public TemplateModel wrap(Object object) throws TemplateModelException {
+                                if( object instanceof ModuleStatus){
+                                    ModuleStatus status = (ModuleStatus) object;
+                                    SimpleHash hash = new SimpleHash();
+                                    hash.put("module", status.getModule() );
+                                    hash.put("name", status.getName() );
+                                    if (status.getComponent() != null) {
+                                        hash.put("component", status.getComponent().get());
```
smithkm: Since you switched to Optional, this needs to be changed.

smithkm:
OK, looks good.  Just need to squash the commits down to make it tidy.

src/extension/wps/web-wps/src/main/resources/applicationContext.xml
```diff
@@ -1,10 +1,15 @@
...
+  2014 - 2016 Open Source Geospatial Foundation. All rights reserved.
+  2001 - 2013 OpenPlans - www.openplans.org.
   This code is licensed under the GPL 2.0 license, available at the root application 
   directory. -->
 <!DOCTYPE beans PUBLIC "-//SPRING//DTD BEAN//EN" "http://www.springframework.org/dtd/spring-beans.dtd">
 <beans>
-
+  <bean class="org.geoserver.platform.ModuleStatusImpl">
+    <constructor-arg index="0" value="gs-web-wps"/>
+    <constructor-arg index="0" value="GeoServer Web UI Web Processing Service"/>
```
smithkm: Wrong Index, it's causing the tests to fail in Travis.

PR #1565, opened 2016-04-12
GEOS-7493 Path.names not removing root slash

NielsCharlier:
See:

Paths.names (method used by multiple ResourceStores) not removing root slash
https://osgeo-org.atlassian.net/browse/GEOS-7493

### DIFF
--- a/src/platform/src/main/java/org/geoserver/platform/resource/Files.java
+++ b/src/platform/src/main/java/org/geoserver/platform/resource/Files.java
@@ -1,4 +1,4 @@
-/* (c) 2014-2015 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014-2016 Open Source Geospatial Foundation - all rights reserved
  * (c) 2014 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
@@ -44,7 +44,7 @@ public final class Files {
         final File file;
 
         private ResourceAdaptor(File file) {
-            this.file = file;
+            this.file = file.getAbsoluteFile();
         }
 
         @Override
--- a/src/platform/src/test/java/org/geoserver/platform/resource/FileWrapperResourceTheoryTest.java
+++ b/src/platform/src/test/java/org/geoserver/platform/resource/FileWrapperResourceTheoryTest.java
@@ -1,3 +1,7 @@
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
 package org.geoserver.platform.resource;
 
 import static org.junit.Assert.fail;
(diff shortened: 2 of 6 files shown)


NielsCharlier:
failure unrelated

aaime:
Check commit guidelines

bencaradocdavies:
@aaime is your objection the lack of description in the initial comment? The title contains a Jira issue id which I have linked. Looks like a bug fix to me.

src/platform/src/test/java/org/geoserver/platform/resource/FileWrapperResourceTheoryTest.java
```diff
@@ -26,7 +30,13 @@
     protected Resource getResource(String path) throws Exception {
         File file = Paths.toFile(null, path);
         if (!file.isAbsolute()) {
-            file = Paths.toFile(folder.getRoot(), path);
+            if (folder.getRoot().getPath().startsWith("/") &&
+                    path.startsWith(folder.getRoot().getPath().substring(1))) {
+                //linux hack
```
bencaradocdavies: @NielsCharlier why is a hack required on Linus?

NielsCharlier: Well, the issue is that the whole FileWrapperResource is a hack. Resources are supposed to have a path relative to the store root (the data directory), not distinguishing between paths that start with "/" and paths that start without it. The FileWrapperResource is a resource without a store and it returns a regular file system path.
Because of this hack, this method distinguishes between relative and absolute paths (although according to the specs, there is no such thing with normal resources). However, in linux it cannot detect absolute paths if they do not start with a slash.

aaime:
@bencaradocdavies I did notice the lack of copyright header updates, which Niels understood and fixed. I linked to the full document as I did not check the other commit guidelines, so there might have been other violations.

bencaradocdavies:
@aaime thanks for the clarification. Sounds good, but all too quick for me!

src/platform/src/test/java/org/geoserver/platform/resource/FileWrapperResourceTheoryTest.java
bencaradocdavies: @NielsCharlier that makes sense. For the benefit of future maintainers, please change the comment to this (or similar):
```
// linux cannot detect absolute resource paths that do not start with a slash
```

NielsCharlier:
Is it okay to merge now @bencaradocdavies ?

bencaradocdavies:
Thanks @NielsCharlier that looks good to me.

src/platform/src/main/java/org/geoserver/platform/resource/Files.java
```diff
@@ -44,7 +44,7 @@
         final File file;
 
         private ResourceAdaptor(File file) {
-            this.file = file;
+            this.file = file.getAbsoluteFile();
```
jodygarnett: This looks to be a scary change, what is the reason for this one?

NielsCharlier: When you to take the path of a resource, it isn't possible to derive from the path whether it is absolute or relative in Linux (Resource paths according to the specs you wrote do not support relative paths and make no difference between a starting slash and no starting slash). I don't see how this would be dangerous or affect anything. ResourceWrappers are used to provide simple files to objects and methods that expect a Resource. They should be threaten like a normal resource, and are mostly used to directly read and write to it. This change simply guarantees that the path returned is unique and makes some kind of consistent sense to somebody who is following the resource API.

PR #1566, opened 2016-04-12
Changes to compile wps-sextante module

ahuarte47:
Changes to compile the wps-sextante plugin using last master branch


ahuarte47:
Thanks @aaime !

PR #1567, opened 2016-04-12
Add upload shapefile to postgis example to importer docs

tbarsballe:
While the Importer REST API examples contained plenty of the more complicated upload into postgis processes, there was no example for a more basic workflow. This commit adds a the basic example of uploading a file to an existing postgis store.

Thanks to @DBlasby for the example content.

### DIFF
--- a/doc/en/user/source/extensions/importer/rest_examples.rst
+++ b/doc/en/user/source/extensions/importer/rest_examples.rst
@@ -295,6 +295,55 @@ And eventually succeed::
 	    ]
 	  }
 	}
+
+Uploading a Shapefile to PostGIS
+--------------------------------
+
+This example shows the process for uploading a Shapefile (in a zip file) to an existing PostGIS datastore (cite:postgis).
+
+Create the import definition::
+
+  {
+    "import": {
+      "targetStore": {
+        "dataStore": {
+          "name": "postgis"
+        }
+      },
+      "targetWorkspace": {
+        "workspace": {
+          "name": "cite"
+        }
+      }
+    }
+  }
+
+POST this definition to /geoserver/rest/imports::
+
+    curl -u admin:geoserver -XPOST -H "Content-type: application/json" -d @import.json "http://localhost:8080/geoserver/rest/imports"
+
+The response will contain the import ID.
+
+We now have an empty import with no tasks. To add a task, POST the shapefile to the list of tasks::
+
+  curl -u admin:geoserver -F name=myshapefile.zip -F filedata=@myshapefile.zip "http://localhost:8080/geoserver/rest/imports/14/tasks"
+
+Since we sent a shapefile, importer assumes the target will be a shapefile store. To import to PostGIS, we will need to reset it.
+Create the following JSON file::
+
+  {
+    "dataStore": {
+      "name":"postgis"
+    }
+  }
+
+PUT this file to /geoserver/rest/imports/14/tasks/0/target::
+
+  curl -u admin:geoserver -XPUT -H "Content-type: application/json" -d @target.json "http://localhost:8080/geoserver/rest/imports/14/tasks/0/target"
+
+Finally, we execute the import by sending a POST to /geoserver/rest/imports/14::
+
+  curl -u admin:geoserver -XPOST "http://localhost:8080/geoserver/rest/imports/14"
 	
 Uploading a CSV file to PostGIS while transforming it
 -----------------------------------------------------


doc/en/user/source/extensions/importer/rest_examples.rst
```diff
@@ -295,6 +295,55 @@ And eventually succeed::
...
+      },
+      "targetWorkspace": {
+        "workspace": {
+          "name": "cite"
+        }
+      }
+    }
+  }
+
+POST this definition to /geoserver/rest/imports  this JSON::
```
DBlasby: post this json to ...  (remove "this JSON")

doc/en/user/source/extensions/importer/rest_examples.rst
```diff
@@ -295,6 +295,55 @@ And eventually succeed::
...
+
+    curl -u admin:geoserver -XPOST -H "Content-type: application/json" -d @import.json "http://localhost:8080/geoserver/rest/imports"
+
+The response will contain the import ID.
+
+We now have an empty import with no tasks. To add a task, POST the shapefile to the list of tasks::
+
+  curl -u admin:geoserver -F name=myshapefile.zip -F filedata=@myshapefile.zip "http://localhost:8080/geoserver/rest/imports/14/tasks"
```
DBlasby: Might want to mention that "14" is the import ID from the previous example

doc/en/user/source/extensions/importer/rest_examples.rst
```diff
@@ -295,6 +295,55 @@ And eventually succeed::
...
+    }
+  }
+
+PUT this file to /geoserver/rest/imports/14/tasks/0/target::
+
+  curl -u admin:geoserver -XPUT -H "Content-type: application/json" -d @target.json "http://localhost:8080/geoserver/rest/imports/14/tasks/0/target"
+
+Finally, we execute the import by sending a POST to /geoserver/rest/imports/14::
+
+  curl -u admin:geoserver -XPUT "http://localhost:8080/geoserver/rest/imports/14"
```
DBlasby: I think this should be -XPOST

DBlasby:
Awesome, thanks for putting this together, torben.  I think i sent you some minor mistakes - I put comments in the changed file!

CHEERS!

doc/en/user/source/extensions/importer/rest_examples.rst
tbarsballe: I was considering it, but all the previous examples also use an arbitrary id without mentioning it. I think it is clear enough as-is, and is well documented in the API definition.

DBlasby: Ok, leave as-is.  :+1:

tbarsballe:
Fixed issues, will merge after build completes.
This is just a doc change, so should be fine to get into the 2.9-beta.

doc/en/user/source/extensions/importer/rest_examples.rst
```diff
@@ -295,6 +295,55 @@ And eventually succeed::
...
+The response will contain the import ID.
+
+We now have an empty import with no tasks. To add a task, POST the shapefile to the list of tasks::
+
+  curl -u admin:geoserver -F name=myshapefile.zip -F filedata=@myshapefile.zip "http://localhost:8080/geoserver/rest/imports/14/tasks"
+
+Since we sent a shapefile, importer assumes the target will be a shapefile store. To import to PostGIS, we will need to reset it.
+Create the following JSON file::
+
+  {
```
aaime: Makes me wonder, why don't set the target store in the initial import like the "csv" example does, there should be no need to fix it afterwards in that case

tbarsballe: If you look above, we actually do do that when creating the import definition. It appears that sending the shapefile resets this (my guess is that since the shapefile is a recognized store type, importer notices this and resets the store type accordingly).

aaime: Weird... I'd call this a bug, but then again, I was not the one that designed the importer API :-p

tbarsballe: Yeah, I am not sure if the behavior is intentional or not.

DBlasby: I thought this was very strange too.  It could be a bug (it seems like a bug).  However, I asked everyone here and no one knew what the expected behaviour was supposed to be.

tbarsballe: I did a bit of further testing, and it looks like this is fixed in 2.9.x. I still prefer the docs as-is, since many people are likely to just read the most recent docs.

DBlasby: I wrote the above against 2.9 (I had to use 2.9 because of the wicket changes).  What was fixed?  Am I doing something wrong?

tbarsballe: I was testing against 2.9.x master in a debugger, so it could be an environment-specific thing I guess. There have not been any changes to importer in the last 3 weeks, so we shoud be running the same code. Given this, I am even more happy to merge the docs as-is.

PR #1573, opened 2016-04-18
[GEOS-7499] Fix unwrapping of SecuredWMSLayerInfo in SecuredLayerInfo and SecureCatalogImpl

nmco:
Issue: https://osgeo-org.atlassian.net/browse/GEOS-7499

The cause of this issue was SecureCatalogImpl and SecuredLayerInfo not properly handling the SecuredWMSLayerInfo wrapper.
The SecureCatalogImpl was not unwrapping resources info of type SecuredWMSLayerInfo, this was provoking the "Not a proxy error" exception.
The SecuredWMSLayerInfo was not unwrapping resources info of type SecuredWMSLayerInfo in the setResource method, so it was overriding the existing layer WMSLayerInfo resource with the secure wrapper. So the first request will go well and the next ones will fail (tricky to found this error).

This pull request add proper handling for the SecuredWMSLayerInfo wrapper and also add tests cases for this.


aaime:
From a cursory look it seems ok, let's hear back from Travis

nmco:
Cool thank you for the review :)

PR #1577, opened 2016-04-19
[GEOS-7505] fix RenderingEngine test

mtCarto:


aaime:
Better, we might still get fun reports from people using more exotic JVMs like IBM or Zulu, or even JDK9 that should have Marlin by default without any sysvar set.
Or did you try with those too?
I'd probably check the JDK and not perform assertions if the JDK is not a known one, and write a big warning in the logs instead

mtCarto:
Thanks for the feedback, updated the test accordingly

simboss:
Thx for this fix. I am merging...

PR #1579, opened 2016-04-20
[GEOS-7482] 404 when logging in from some pages

aaime:


aaime:
Probably not the prettiest fix, but appears to work fine for me.

tbarsballe:
This looks good, do you want to merge?
We should be able to get this into the 2.9 beta, given all the backaging issues currently blocking the release.

aaime:
Yes, merging. I thought this ticket was actually one of the blockers for beta2 :-)

PR #1580, opened 2016-04-20
[GEOS-7510] Upgrade from OL 3.7.0 to OL 3.15.1

tbarsballe:
Using the ol3 dist from https://github.com/openlayers/ol3/releases/tag/v3.15.1


tbarsballe:
I did some basic testing of the OL3 preview page, and it seems to work fine. Confermed that GeoServer is using the new version.

PR #1583, opened 2016-04-27
GEOS-7521: fix external module so it doesn't clip

DBlasby:
See details in ticket.

a) moved to 1.0.9
b) created subclass that doesn't do any clipping
c) use the sub-class

### DIFF
--- a/src/community/vectortiles/pom.xml
+++ b/src/community/vectortiles/pom.xml
@@ -40,7 +40,7 @@
     <dependency>
       <groupId>no.ecc.vectortile</groupId>
       <artifactId>java-vector-tile</artifactId>
-      <version>1.0.8</version>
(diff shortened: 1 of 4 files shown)


DBlasby:
Added Test Case

tbarsballe:
This looks good, merging.
Thanks for the fix.

tbarsballe:
This has been on master for a month, and is a bugfix. Backporting to 2.9.x and 2.8.x.

aaime:
This refers to GEOS-7521, but the ticket at https://osgeo-org.atlassian.net/browse/GEOS-7522 refers back to this one... confused?

aaime:
It seems the suggested fix for GEOS-7522 was applied in this ticket, I've closed 7522.

PR #1585, opened 2016-04-29
[GEOS-7523] Make GWC settings UI only show the accepted eviction policies (backport 2.8.x)

nmco:
The issue associated with this pull request: https://osgeo-org.atlassian.net/browse/GEOS-7523

This pull request fixes the issue, updates the documentation and add a test case.

### DIFF
--- a/doc/en/user/source/webadmin/tilecache/defaults.rst
+++ b/doc/en/user/source/webadmin/tilecache/defaults.rst
@@ -148,7 +148,11 @@ Parameter for configuring in memory cache size in MB.
 
 Cache Eviction Policy
 ```````````````````````````````````````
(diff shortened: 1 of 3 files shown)


nmco:
The build failure is in the main module and the failed test is:
`testOverrideEPSGOperation(org.vfny.geoserver.crs.OvverideTransformationsTest): Database failure while creating a 'CoordinateReferenceSystem' object for code "EPSG:4269".`

It seems to be unrelated with this pull requests changes.

simboss:
It is. Quite annoying though...

aaime:
Agreed. Keep on happening on the build servers, never managed to reproduce it locally. Just run it like 20 times in a row with openjdk 8, and then on oracle jdk 8, no dice, always works.
I've made a blind attempt at fixing it here: https://github.com/geoserver/geoserver/pull/1586

PR #1586, opened 2016-04-30
Blind attempt to fix random failure in crs override test. Hoping that the build target dir is better behaved than the global tmp one

aaime:


bencaradocdavies:
@aaime do you have a link to the failures? Or a Jira? We looked at this in the committee meeting and were not sure which failures this is meant to fix. It does look quite ugly.

src/main/src/test/java/org/vfny/geoserver/crs/OvverideTransformationsTest.java
```diff
@@ -27,14 +27,29 @@
...
+            System.clearProperty(JAVA_IO_TMPDIR);
+        } else {
+            System.setProperty(JAVA_IO_TMPDIR, OLD_TMP_VALUE);
+        }
+    }
 
     @Override
     protected void onSetUp(SystemTestData testData) throws Exception {
+        OLD_TMP_VALUE = System.getProperty(JAVA_IO_TMPDIR);
+        System.setProperty(JAVA_IO_TMPDIR, new File("./target").getCanonicalPath());
```
bencaradocdavies: I do not like tests that change system properties. Why is this needed or expected to fix anything? Is this a Windows fix?

aaime:
Ugly is in the eye of the beholder, the system property gets cleared in the test class teardown. By applying the same logic, you should probably wipe also the tens of tests that are setting and resetting system properties down in GeoTools (a quick find and grep return 53 of them[1])

As for the failure, searching for OvverideTransformationsTest in my mailbox I quickly gathered 101 failed builds reports, are you sure you never seen it? It happens on both windows and linux, occasionally on Travis too.
Here is a link to one such failure from the official build server: http://ares.boundlessgeo.com/jenkins/job/geoserver-2.8.x/221/console

The test has been randomly failing for years, but increased its appearance recently (no idea why, maybe it's just more visible because we have more build servers). The rationale behind this fix is that the global temp directory might be for some reason more fragile on some build servers, e.g., scheduled cleanups, remote mount, whatever.
Never managed to reproduce it locally, as the title says, this is a blind attempt at a fix after the n-th complaint about it.

There is no Jira that I'm aware of, I can create one and change the commit message, no problem. Or I can just close this one if more people think it's too ugly, you might have a cleaner solution for it. Finally, you can just close it and move on, I was just trying to help but there is no one actually sponsoring this fix after all, no attachments here.

[1]: try running ``find . -name "*Test.java" | xargs grep "System.setProperty" | grep -v "//"``

bencaradocdavies:
Thanks for the clarification @aaime. Ugly is orthogonal to useful, and I am well aware that changes to system properties are made in other tests; your fix does the right thing and reverts the change in tearDown. I am just complaining about an architecture that makes you do this.

While I could guess that this was to fix the problem you noted above, I was not sure and wanted clarification as to whether this was for a specific platform, not because I doubted the fix, but for future reference. No need for a Jira, but add a link if there is one. A few lines of exception message would help draw search engines.

PR #1587, opened 2016-05-04
VectorTiles: Applies the buffer used for clipping when querying for features.

bladegun:
Existing code use a bbox filter that is the same as the tile border. Surrounding features that are located near the tile border are not included. When rendering vector tiles using a thick line style (especially road), some thick lines will get cropped near the tile border.

mvt-before-fix-marked
Those lines are NOT in the neighbor tile. So the neighbor tile does not render it at all.

I have updated the code to expand the bbox filter by the clip buffer.

mvt-after-fix-marked
Those lines are now included in the neighbor tile.

### DIFF
--- a/src/community/vectortiles/src/main/java/org/geoserver/wms/vector/PipelineBuilder.java
+++ b/src/community/vectortiles/src/main/java/org/geoserver/wms/vector/PipelineBuilder.java
@@ -243,13 +243,17 @@ class PipelineBuilder {
                 clippingEnvelope = paintArea;
             } else {
                 ReferencedEnvelope renderingArea = context.renderingArea;
-                renderingArea.expandBy( clipBBOXSizeIncreasePixels * context.pixelSizeInTargetCRS);
+                renderingArea.expandBy(getClipDistance());
                 clippingEnvelope = renderingArea;
             }
 
             addLast(new Clip(clippingEnvelope));
         }
         return this;
+    }
+
+    public double getClipDistance() {
+        return clipBBOXSizeIncreasePixels * context.pixelSizeInTargetCRS;
     }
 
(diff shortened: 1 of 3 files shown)


aaime:
An observation, I'm not familiar with the vector tiles code but the streaming renderer down in geotools has logic to detect how much of a buffer apply to avoid this kind of border issues, it basically scans the style to see how big the symbolizers are, this can be also overridden by a GetMap request parameter (buffer). In most cases it just works, if it's not already in use you might want to have a look at it.

bladegun:
The vector tiles plugin doesn't use the streaming renderer. It doesn't generate image. It generates GeoJSON, TopoJSON or PBF (google protobuf) files that store the coordinates of all the geometries within a tile. Rendering is done entirely on the client side. My screenshots are the result of rendering by MapBox GL JS.

aaime:
I know what vector tiles are and how they are used.
The engine producing them is actually using a style as a reference to decide what vector features put in them (based on some code part of, guess what, the streaming renderer package, see VectorTileMapOutputFormat, generally speaking there are 18 references to that package in the vector tile module), you might as well use it to get a default buffer value too. The class deciding how much to expand the buffer based on the style is called MetaBufferEstimator.

bladegun:
The engine does make use of a style, but it only reads the scale denominator and filter from the style to construct a Query. It does not read the symbolizer in the style. Stroke width is not being taken into consideration. See VectorMapRenderUtils.

aaime:
I know, I'm suggesting it could to get a default buffer value. Not all clients can do vector tiles, a website will probably want to provide the same maps as PNG tiles as well, so it's not unlikely to find a corresponding SLD style on the server side.

PR #1589, opened 2016-05-05
[GEOS-7526] Disable unique wicket Resource IDs for clustering 

NielsCharlier:
Hazelcast/JDBConfig clustering issues caused by HTTP 302 response and unique wicket resource IDs

See https://osgeo-org.atlassian.net/browse/GEOS-7526

Discussed at SC meeting


simboss:
Blindly reviewing, I see changes to modules in core without even a test.

This is fine with the HZ code (although I wound not really recommend it...) but It is not for the changes to the core modules.

simboss:
Ops, wrong button :)

simboss:
That said, missing copyright updates as well.

simboss:
Suggestion for the next PR: can we take care of  the obvious stuff  like tests and headers before making a PR so we don't have to waste time on that?

Thx a lot :)

NielsCharlier:
Well this time I checked all copyright headers and they are already up to date! 

I realised I needed a test and I added it before I brought attention to this PR on the list :)

simboss:
Thx for the test, let's see what others think.

aaime:
@NielsCharlier  we are all immediately notified of a pull request the moment you open it

src/main/src/main/java/org/geoserver/config/impl/GeoServerInfoImpl.java
```diff
@@ -47,6 +47,8 @@
     protected Boolean xmlExternalEntitiesEnabled = Boolean.FALSE;
     
     protected String lockProviderName;
+
+    protected WebUIMode webUIMode = WebUIMode.DEFAULT;
```
aaime: Copyright header not updated. There are shell scripts to automatically update the headers that have been shared on the list multiple times, why don't you use them instead of checking manually file by file, and risking of losing an update?
Just commit, run the following (it's called fix_headers on my system), commit again and squash the commit. Only needs to be done once before making a pull request, minimum overhead:

````
sed -i 's/\(([cC])\) 2014 Open Source Geospatial Foundation/\1 2014 - 2016 Open Source Geospatial Foundation/' $(git diff-tree --no-commit-id --name-only -r HEAD)
sed -i 's/\(([cC])\) 2014-2015 Open Source Geospatial Foundation/\1 2014 - 2016 Open Source Geospatial Foundation/' $(git diff-tree --no-commit-id --name-only -r HEAD)
sed -i 's/\(([cC])\) 2014 - 2015 Open Source Geospatial Foundation/\1 2014 - 2016 Open Source Geospatial Foundation/' $(git diff-tree --no-commit-id --name-only -r HEAD)
````

aaime: Btw, works fine only if the pull is made of a single commit, otherwise you have to run it after each commit.

NielsCharlier:
Okay, sorry I still missed that one, next time I'll run the scripts either way. Fixed

NielsCharlier:
Added documentation for new setting.

jodygarnett:
Cherry picked these two commits to 2.9.x, marking GEOS-7539 as resolved.

PR #1597, opened 2016-05-10
[GEOS-7539]Geoserver  2.9beta2 and 2.9rc1 does not start when using the extension imagemap: Update applicationContext.xml

cristianeandrioli:
Geoserver does not start when using the extension imagemap (geoserver-2.9-RC1-imagemap-plugin.zip). Works fine in 2.9beta1, does not work in 2.9beta2 and 2.9rc1. Tested in Linux and Windows.

An error occurred in the file applicationContext.xml :

Attribute "singleton" must be declared for element type "bean"

See https://osgeo-org.atlassian.net/browse/GEOS-7539

### DIFF
--- a/src/extension/imagemap/src/main/java/applicationContext.xml
+++ b/src/extension/imagemap/src/main/java/applicationContext.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 -<!-- 
- - Copyright (C) 2014 - Open Source Geospatial Foundation. All rights reserved.
+ - Copyright (C) 2014 - 2016 - Open Source Geospatial Foundation. All rights reserved.
  - This code is licensed under the GPL 2.0 license, available at the root
  - application directory.
  - -->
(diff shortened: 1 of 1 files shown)


src/extension/imagemap/src/main/java/applicationContext.xml
```diff
@@ -1,16 +1,17 @@
-﻿<?xml version="1.0" encoding="UTF-8"?>
-<!-- 
- Copyright (C) 2014 - Open Source Geospatial Foundation. All rights reserved.
- This code is licensed under the GPL 2.0 license, available at the root
- application directory.
- -->
-<!DOCTYPE beans PUBLIC "-//SPRING//DTD BEAN//EN" "http://www.springframework.org/dtd/spring-beans.dtd">
+<?xml version="1.0" encoding="UTF-8"?>
```
aaime: Please do not remove the copyright headers from this file, you should add them back and update the years to say "2014 - 2016"

aaime:
Thanks a lot for your contribution, the changes to the file look good but we need to retain the copyright headers, please check my comment in the diff.

src/extension/imagemap/src/main/java/applicationContext.xml
cristianeandrioli: Aime.
Now it´s ok.

Cris

jodygarnett:
Thanks, squashed into a single commit and merged to master and 2.9.x.

PR #1601, opened 2016-05-19
Undo faulty serializable resources [GEOS-7513]

NielsCharlier:
Undo faulty serializable resources, Resources.serializable, fix DemoRequestPage and importer module.

For more information see https://osgeo-org.atlassian.net/browse/GEOS-7513

### DIFF
--- a/src/community/jdbcstore/src/main/java/org/geoserver/jdbcstore/JDBCResourceStore.java
+++ b/src/community/jdbcstore/src/main/java/org/geoserver/jdbcstore/JDBCResourceStore.java
@@ -140,9 +140,7 @@ public class JDBCResourceStore implements ResourceStore {
      * 
      */    
     protected class JDBCResource implements Resource {
-        
-        private static final long serialVersionUID = 1829574259576354635L;
-        
+
         private final JDBCDirectoryStructure.Entry entry;
         
         public JDBCResource(JDBCDirectoryStructure.Entry entry) {
--- a/src/extension/importer/core/src/main/java/org/geoserver/importer/FileData.java
+++ b/src/extension/importer/core/src/main/java/org/geoserver/importer/FileData.java
@@ -7,6 +7,7 @@ package org.geoserver.importer;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.ObjectInputStream;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -28,7 +29,7 @@ public class FileData extends ImportData {
     protected Resource file;
 
     public FileData(Resource file) {
-        this.file = file;
+        this.file = Resources.serializable(file);
     }
 
     public FileData(FileData file) {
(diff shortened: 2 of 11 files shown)


src/community/jdbcstore/src/main/java/org/geoserver/jdbcstore/JDBCResourceStore.java
```diff
@@ -140,9 +140,7 @@ public String toString() {
      * 
      */    
     protected class JDBCResource implements Resource {
-        
-        private static final long serialVersionUID = 1829574259576354635L;
-        
+
```
jodygarnett: This space is intentionally left blank

src/extension/importer/core/src/main/java/org/geoserver/importer/FileData.java
```diff
@@ -28,7 +29,7 @@
     protected Resource file;
 
     public FileData(Resource file) {
-        this.file = file;
+        this.file = Resources.serializable(file);
```
jodygarnett: This is not quite what I expected, I expected something like:
````
protected String path;

protected Resource resource(){
     return resources.get( path );
}
````

But I would have to check how FileData is being used, perhaps this is what @cuttlefish was referencing in terms of accidentally loading a geotiff into jdbcstore.

File: src/platform/src/main/java/org/geoserver/platform/resource/Resource.java [hunk 3]
```diff
@@ -20,7 +20,7 @@
  * Resource creation is handled in a lazy fashion, simply use {@link #file()} or {@link #out()} and the resource will be created as required. In a
  * similar fashion setting up a child resource will create any required parent directories.
  */
-public interface Resource extends Serializable {
+public interface Resource {
```
jodygarnett: +1 yay!

src/extension/importer/core/src/main/java/org/geoserver/importer/FileData.java
NielsCharlier: As I suggested, I made a serializable resource wrapper for the sake of convenience. This way we hardly need to change any code in the other modules... seemed like a good idea to me, no?

As far as @cuttlefish is concerned, I believe I can help him through a configuration change.

jodygarnett: Right, got it.

tbarsballe: It looks like this got missed when the ResourceStore work was initially done, but FileData is a representation of a file that is being loaded into Importer (in other words, not yet added to geoserver), and should actually be using File, not Resource.

jodygarnett:
@tbarsballe I am going to review and merge as is, we can handle the FileData revert in a separate fix.

jodygarnett:
Updated headers and squashed into a single commit on master / 2.9.x.

PR #1600, opened 2016-05-19
GEOS 7538: Monitoring plugin doesn't log remote user

bossie:
Hi,

I've done the changes for this issue along with two unit tests. Please review.

Enjoy,
Jan

PR #1602, opened 2016-05-19
Revert "GEOS 7538: Monitoring plugin doesn't log remote user"

simboss:
Reverts geoserver/geoserver#1600


simboss:
Reopened as we have to port first on master and 2.9

PR #1605, opened 2016-05-19
Fix the download endpoint for asynchronous tasks.

jdgarrett:


aaime:
Hi, any fix is accepted first on the master branch, and then eventually backported down to stable and maintenance. Could you open a pull request for the master branch instead?

jdgarrett:
Sorry I should have clarified in the description, this is a fix for the GeoGig plugin, which does not exist on the master branch.

aaime:
Does not exist for the master branch? Ouch, that's highly irregular... but I guess that for a community plugin it's to some extent acceptable.
Any plans to rectify and have a plugin on master and 2.9.x too?
Merging in any case....

PR #1610, opened 2016-05-20
[GEOS-7549] XSS: Reflected Cross Site Script Vulnerability

aaime:


aaime:
Build went well, merging so that we have a nightly that the reporter can use to test

PR #1612, opened 2016-05-20
list max memory on status page (next to used memory)

DBlasby:
Made a minor change to StatusPage. 

It now reports max memory.

For example;
Memory Usage	51 MB **/ 910 MB**


tbarsballe:
This is a helpful change.
However, it would be good if the docs that show the GeoServer status page were updated to reflect this change.

Also, does this need to get into the 2.9.0 release, currently in progress?

DBlasby:
yes - will like to push this to 2.9.  Also, I am working on a few other minor changes to the REST about/status to add some extra info for some of the modules.

tbarsballe:
Ok. I am good to merge this if we get a doc update for all these changes eventually.

jodygarnett:
Here is the doc page:

- http://docs.geoserver.org/latest/en/user/configuration/status.html#config-serverstatus

Proposed replacement text:

````
The amount of memory current used by GeoServer. In the above example, 55.32 MB of memory is being used out of a total of 980 MB. Clicking on the “Free Memory” button, cleans up memory marked for deletion by running the garbage collector.
```

Update the screen snap, and make the numbers match.

aaime:
Please be very careful with this last minute additions, sure, it's one liner, but it's like making a jump without a net... if any change screws the release, it does for good, there is no RC2. 
That said, this one does not seem problematic.

DBlasby:
added doc and new screen shot

aaime:
I believe such large images caused doc build issues in the past... but could have been only for PDF.
Anyways, that' s the reason why most images in the docs are < 700px wide

tbarsballe:
Updated image to 723 px wide, manually merged, and backported to 2.9.x.

jodygarnett:
Oh, okay then ... I did not expect the backport to 2.9.x

PR #1613, opened 2016-05-20
[GEOS-7544] Use File for Importer (not Resource)

tbarsballe:
For https://osgeo-org.atlassian.net/browse/GEOS-7544

Could I get a review @jodygarnett or @NielsCharlier ?

### DIFF
--- a/src/extension/importer/bdb/src/main/java/org/geoserver/importer/bdb/BDBImportStore.java
+++ b/src/extension/importer/bdb/src/main/java/org/geoserver/importer/bdb/BDBImportStore.java
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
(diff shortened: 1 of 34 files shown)


src/extension/importer/bdb/src/main/java/org/geoserver/importer/bdb/BDBImportStore.java
```diff
@@ -21,7 +21,6 @@
 import org.geoserver.importer.ImportStore;
 import org.geoserver.importer.ImportTask;
 import org.geoserver.importer.Importer;
-import org.geoserver.platform.resource.Resource;
```
jodygarnett: You will need to update the header on this file.

src/extension/importer/core/src/main/java/org/geoserver/importer/DataStoreFormat.java
```diff
@@ -224,7 +223,7 @@ public DataStore createDataStore(ImportData data) throws IOException {
 
             if (f != null) {
                 Map<String,Serializable> map = new HashMap<String, Serializable>();
-                map.put("url", relativeDataFileURL(Resources.find(f).toURI().toURL().toString(), catalog));
+                map.put("url", relativeDataFileURL(f.toURI().toURL().toString(), catalog));
```
jodygarnett: The DataUtilities method is slightly more stable then toURI().toURL() on platforms like windows.

tbarsballe:
Updated to fix compile failures caused by revert, added headers to BDBImportStore

jodygarnett:
Thanks @tbarsballe - running checks now and will merge.

PR #1619, opened 2016-05-25
Remove unnecessary repositories from main pom file

cuttlefish:
A 3rd party repo went down and was causing the build to take > 10 hours due to very long timeouts.

Given the way that Maven handles repo failures, I think that we should do our best to minimize the number of repos included. All of the necessary jars seem to be pulling correctly when only the Boundless and OSGeo repositories are used.

I was having a problem pulling a jar when building the community extensions, but the project (geoscript-py) is currently semi abandoned and was relying on unsupported snapshots.

@aaime has already made part of the change, so it won't automatically merge.

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="ISO-8859-1"?>
-<!-- 
- Copyright (C) 2015 - Open Source Geospatial Foundation. All rights reserved.
+<!--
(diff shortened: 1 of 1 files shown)


cuttlefish:
Oops, I accidentally made this PR off of a fork's master branch. I can redo it if necessary.

cuttlefish:
I'm going to create a new PR.
I can't figure out how to undo this... please delete this PR if possible.

PR #1620, opened 2016-05-25
Remove unnecessary repositories from main pom file

cuttlefish:
A 3rd party repo went down and was causing the build to take > 10 hours due to very long timeouts.

Given the way that Maven handles repo failures, I think that we should do our best to minimize the number of repos included. All of the necessary jars seem to be pulling correctly when only the Boundless and OSGeo repositories are used.

I was having a problem pulling a jar when building the community extensions, but the project (geoscript-py) is currently semi abandoned and was relying on unsupported snapshots.

[I submitted this previously, but had accidentally done it off of a branch under active development]

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -72,6 +72,18 @@
     <url>http://download.osgeo.org/webdav/geotools/</url>
     <!-- release repository used by geotools (and third-party dependencies) -->
   </repository>
+   
+   <repository>
+      <id>geosolutions</id>
+      <name>geosolutions repository</name>
+      <url>http://maven.geo-solutions.it/</url>
+      <snapshots>
+        <enabled>true</enabled>
+      </snapshots>
+     <releases>
+       <enabled>true</enabled>
+     </releases>
+    </repository>
  </repositories>
 
  <dependencyManagement>
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="ISO-8859-1"?>
-<!-- 
- Copyright (C) 2015 - Open Source Geospatial Foundation. All rights reserved.
+<!--
+ Copyright (C) 2016 - Open Source Geospatial Foundation. All rights reserved.
  This code is licensed under the GPL 2.0 license, available at the root
  application directory.
  -->
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -72,27 +72,6 @@
     <url>http://download.osgeo.org/webdav/geotools/</url>
     <!-- release repository used by geotools (and third-party dependencies) -->
   </repository>
-   
-  <repository>
-    <id>maven-restlet</id>
-    <name>Restlet Maven Repository</name>
-    <url>http://maven.restlet.org</url>
-    <snapshots>
-      <enabled>false</enabled>
-    </snapshots>
-  </repository>
-
-   <repository>
-      <id>geosolutions</id>
-      <name>geosolutions repository</name>
-      <url>http://maven.geo-solutions.it/</url>
-      <snapshots>
-        <enabled>true</enabled>
-      </snapshots>
-     <releases>
-       <enabled>true</enabled>
-     </releases>
-    </repository>
  </repositories>
 
  <dependencyManagement>


cuttlefish:
It'd be nice to get some independent verification that this works from a totally empty Maven cache.

src/pom.xml
```diff
@@ -72,27 +72,6 @@
...
-    <url>http://maven.restlet.org</url>
-    <snapshots>
-      <enabled>false</enabled>
-    </snapshots>
-  </repository>
-
-   <repository>
-      <id>geosolutions</id>
-      <name>geosolutions repository</name>
-      <url>http://maven.geo-solutions.it/</url>
```
jodygarnett: Although repo.boundlessgeo.org mirrors this repository, we should check with @simboss about removing this reference.

src/pom.xml
```diff
@@ -72,27 +72,6 @@
     <url>http://download.osgeo.org/webdav/geotools/</url>
     <!-- release repository used by geotools (and third-party dependencies) -->
   </repository>
-   
-  <repository>
-    <id>maven-restlet</id>
-    <name>Restlet Maven Repository</name>
-    <url>http://maven.restlet.org</url>
```
jodygarnett: This repo is working for me, but it is cached (https://repo.boundlessgeo.com/artifactory/webapp/#/artifacts/browse/tree/General/restlet-cache) and republished as part of repo.boundlessgeo.com/main

jodygarnett:
@cuttlefish I wish there was a way to use a profile for this, as far as I know profiles can only add repositories.

bencaradocdavies:
@cuttlefish building a fresh checkout with the release and community release profiles with a clean local maven repo (set through a settings file) fails with missing dependencies.
```
$ mvn -version
Apache Maven 3.3.9 (bb52d8502b132ec0a5a3f4c09453c07478323dc5; 2015-11-11T05:41:47+13:00)
Maven home: /home/ben/java/maven
Java version: 1.8.0_91, vendor: Oracle Corporation
Java home: /usr/lib/jvm/java-8-openjdk-amd64/jre
Default locale: en_GB, platform encoding: UTF-8
OS name: "linux", version: "4.5.0-2-amd64", arch: "amd64", family: "unix"
```
Maven command line. Eclipse included to ensure source downloads also tested:
```
mvn -llr -s ~/geoserver/maven/settings.xml -Prelease -PcommunityRelease -T4 -fae -DskipTests -DdownloadSources clean install eclipse:clean eclipse:eclipse
```
Failure:
```
[ERROR] Failed to execute goal on project gs-libjpeg-turbo: Could not resolve dependencies for project org.geoserver.extension:gs-libjpeg-turbo:jar:2.10-SNAPSHOT: Failed to collect dependencies at it.geosolutions.imageio-ext:imageio-ext-turbojpeg:jar:1.1.13 -> org.libjpegturbo:turbojpeg-wrapper:jar:1.2.1.1: Failed to read artifact descriptor for org.libjpegturbo:turbojpeg-wrapper:jar:1.2.1.1: Failure to find it.geosolutions.imageio-ext:imageio-ext-library:pom:1.1-SNAPSHOT in https://repo.boundlessgeo.com/main/ was cached in the local repository, resolution will not be reattempted until the update interval of boundless has elapsed or updates are forced -> [Help 1]
[ERROR] Failed to execute goal on project gs-script-py: Could not resolve dependencies for project org.geoserver.script:gs-script-py:jar:2.10-SNAPSHOT: Failure to find org.geoscript:geoscript-py:jar:1.4-SNAPSHOT in https://repo.boundlessgeo.com/main/ was cached in the local repository, resolution will not be reattempted until the update interval of boundless has elapsed or updates are forced -> [Help 1]
```

bencaradocdavies:
@cuttlefish I will build again with ```-U``` to see if pain persists.

bencaradocdavies:
@cuttlefish no good, same to failures with ```-U``` (snapshot updates).

aaime:
Hey @bencaradocdavies thanks for testing. Yeah, I don't think we can do without the geosolutions repository, I personally already wasted too much time trying to make the boundless repo work in the last six months before giving up and adding back the geosolutions one (I was supposedly given admin access, but when it was needed, I could not actually admin squat, when complained no change was made, it's just did not work out...).

For the short term, we cannot remove the geosolutions one it because Emanuele just added yesterday the hibernate spatial jars we were fetching from the failing repo onto the geosolutions one, but more in general, we have a number of people working on upstream projects (the various jai/imageio exts, geofence) that are deploying on the geosolutions repo, going to add a new user every time to the boundless repo is a chore, in fact I did not have time to fix the hibernatespatial stuff myself, but Emanuele jumped on it exactly because he could upload stuff without any red tape.

In the last few months the boundless repo has had repeated reliability issues that broke several builds during artifact downloads and deploys, I haven't seen an equivalent pattern for the geosolutions one, but if it happens, we're eager to fix it, just let us know.

cuttlefish:
@bencaradocdavies: I mentioned that gs-script-py was failing for me. I'm not sure why, since it's in at least one of the repos. But it's built against the wrong version of GeoTools anyway, so it's probably for the best that it's not being loaded.
It's odd that I don't get the other error. It shouldn't be making a difference since the GeoSolutions repo was added yesterday, right?

aaime: I agree about the reliability of the Boundless repo, I've had many problems with it. But I got started on this because whatever was added to the GeoSolutions repo wasn't working for me. We had to add several jars to the Boundless repo. Perhaps some of them didn't get copied? I'm not sure that two haphazard solutions is better than one.

I've been working on Vagrant box for building GeoServer, so I've been blowing away the cache regularly. I wouldn't have come across these issues otherwise. All I've learned is that I really dislike Maven, especially since adding repos has a huge performance impact.

aaime:
@cuttlefish I haven't seen any huge performance issue with 1 or 100 extra repos unless I'm building from a empty repo (which is not the normal setup, it should only happen once). Doing a parallel build helps a lot to handle with the latency, along with the takari local concurrent repo manager to actually make that work.

If there were issues with the hibernatespatial jars you should have just said so, with this pull request you're cutting away a repository that we depend on and use in anger, it may solve a problem for you, but creates a lot for us on the other side of the pond...

The GeoSolutions repo was added back months ago after wasting the entire winter trying to get the single unified repo to work without any result. Don't want to go back there.

aaime:
Btw, check the Travis build:

````
[ERROR] Failed to execute goal on project gs-libjpeg-turbo: Could not resolve dependencies for project org.geoserver.extension:gs-libjpeg-turbo:jar:2.10-SNAPSHOT: Failed to collect dependencies at it.geosolutions.imageio-ext:imageio-ext-turbojpeg:jar:1.1.13 -> org.libjpegturbo:turbojpeg-wrapper:jar:1.2.1.1: Failed to read artifact descriptor for org.libjpegturbo:turbojpeg-wrapper:jar:1.2.1.1: Could not transfer artifact it.geosolutions.imageio-ext:imageio-ext-library:pom:1.2-SNAPSHOT from/to codehaus-snapshots (https://nexus.codehaus.org/snapshots/): nexus.codehaus.org: Unknown host nexus.codehaus.org -> [Help 1]
````

This is the error we could never fix with the old single repo. It's a mistake to have a release artifact depending on a snapshot for sure, but we did not have the time to do a full release just to fix it.
That said, the main problem is one of control and access, in the current setup I'm playing single point of failure for my company (if I'm not there, sick, or just too busy to help with the repo).

cuttlefish:
@aaime, git blame said that the GeoSolutions repo was added yesterday. I didn't realize it had been there for months.

What do people think about dropping the other repos?

Doesn't maven scan through every repository? Maybe that's only for snapshots, but since people are linking against snapshots everywhere it seems very slow on my setup. Maybe there are configurations to avoid that?

aaime:
And oh, pulling snapshot jars from repos has been historically source of lots of troubles, the boundless repo (even before artifactory) was returning months old snapshots on occasion (but not always... no, would have been too boring otherwise).

bencaradocdavies:
@cuttlefish I always build snapshot dependencies locally and invoke Maven with ```-llr -nsu``` to ensure that (1) there is no memory of the origin of artifacts and repos can be moved, and (2) snapshots are not updated. For (1), see:
https://issues.apache.org/jira/browse/MNG-5181
https://issues.apache.org/jira/browse/MNG-5883

Maven has caused a great me a great deal of grief over the years. With local snapshot builds, ```-llr``` and ```-nsu```, Maven 3.3.9 is fast and reliable. Andrea's recent fixes for our git commit id plugin configuration have made it the fastest ever.

Maven is the worst dependency management tool, except all the others.

cuttlefish:
aaime: Doh. Apparently I just don't know what month it is... I think I need more sleep.

cuttlefish:
@bencaradocdavies Thanks, I'll give that a try. I hate trying to work with a modified GeoServer only to realize half way through a build that it's pulling down snapshots from the repo instead of using the ones it _just_ built :(

PR #1626, opened 2016-05-30
New screenshots and format table for layer preview docs

dvntucker:
Updated screenshots for the layer preview page.


doc/en/user/source/data/webadmin/layerpreview.rst
```diff
@@ -12,17 +12,19 @@ This page provides layer views in various output formats. A layer must be enable
...
-   * - .. image:: img/preview_layergroup.png
-     - Layer group 
+   * - .. image:: img/raster_icon.png
+     - Raster (grid)
+   * - .. image:: img/polygon_icon.png
+     - Polygon
+   * - .. image:: img/line_string_icon.png
+     - Line
+   * - .. image:: img/point_icon.png
+     - Point
```
jodygarnett: Think we still need layergroup here

doc/en/user/source/data/webadmin/layerpreview.rst
```diff
@@ -12,17 +12,19 @@ This page provides layer views in various output formats. A layer must be enable
 Each layer row consists of a type, name, title, and available formats for viewing.
```
jodygarnett: Consider filtering the preview screen snap with "sf" to make a smaller image :)

doc/en/user/source/data/webadmin/layerpreview.rst
aaime: If you want this to be complete (not saying that it has to be) then there's also an icon for WMS cascaded layers, and a "generic geometry" one. See https://github.com/geoserver/geoserver/blob/master/src/web/core/src/main/java/org/geoserver/web/CatalogIconFactory.java#L114 and the associated methods

dvntucker:
Just added the remaining icon types to the layer preview doc page based on the code aaime linked.

PR #1628, opened 2016-06-01
[GEOS-7565] Set model type for cluster web page components

nmco:
Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-7565


aaime:
Yep, looks good to me. Waiting for Travis is not going to add anything, as Travis does not build community modules anyways.Thus, merging right away. Backport as you see fit.

PR #1629, opened 2016-06-01
[GEOS-7565] Set model type for cluster web page components (backport to 2.9.x)

nmco:
Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-7565


aaime:
Ah ha, here's the backport. Merging

PR #1634, opened 2016-06-09
Add GWC MBTiles integration community module

nmco:
This pull request is dependent on GWC pull request: https://github.com/GeoWebCache/geowebcache/pull/388


jodygarnett:
This blocked pending GeoWebCache change, @groldan can you provide an update?

aaime:
Looking at the blobstore UI, the directories for storage and metadata should be close toghether (btw, still unsure what the metadata dirs contain to start with, isn't the metadata already contained in the mbtiles dbs?)

doc/en/user/source/community/gwc-sqlite/index.rst
```diff
@@ -0,0 +1,158 @@
...
+This plugin provides integration with GWC SQLite based blob stores. At the moment only one blob store of this type is available, the MBTiles blob store.
+
+MBTiles BlobStore
+-----------------
+
+This blob store allow us to store tiles using the `MBTiles <https://github.com/mapbox/mbtiles-spec/blob/master/1.1/spec.md>`_ specification (version 1.1) which defines a schema for storing tiles in an `SQLite <https://www.sqlite.org/>`_ database with some restrictions regarding tiles formats and projections.
+
+MBTiles specification only supports JPEG and PNG formats and projection EPSG:3857 is assumed. The implemented blob store will read and write MBTiles files compliant with the specification but will also be able to write and read MBTiles files that use others formats and projections.
+
+Using the MBTiles blob store will bring several benefits at the cost of some performance loss. We will be able to store our data with a significantly smaller number of files, which will help us avoid file system headaches and help us managing our data. In some cases the stored data will be more compacted reducing the size of the data on disk.
```
aaime: Same observations as with the GWC documentation, try to avoid using "we"

doc/en/user/source/community/gwc-sqlite/index.rst
```diff
@@ -0,0 +1,158 @@
...
+
+When compared to the file blob store this store as two limitations:
+
+* This store will not integrate with disk quota, this is a consequence of using database files.
```
aaime: Does not integrate... how? Like, is disk quota disabling itself, ignoring the storage? Or it's just that if enabled the stats kept in disk quota do not make much sense and do not reflect the actual disk usage?
If it's the latter, I'd clarify and state that it's best not to enable disk quota because the size of the dbs cannot be really controlled.

doc/en/user/source/community/gwc-sqlite/index.rst
```diff
@@ -0,0 +1,158 @@
...
+File Path Templates
+~~~~~~~~~~~~~~~~~~~
+
+An MBTiles file will correspond to an SQLitle database file. In order to limit the amount of contention on each single database file we will allow users to decide the granularity of the databases files. When GeoWebCache needs to map a tile to a database file it will only look at the databases files paths, it will not take in account the MBTiles metadata (this is why we are able to handle others formats and projections).
+
+To configure the databases files granularity the user needs to provide a file path template. The default file path template for the MBTiles blob store is this one:
+
+.. code-block:: none
+
+  {grid}\{layer}\{format}\{params}\{z}\tiles_{x}_{y}.sqlite
```
aaime: Same as GWC, use a sample template following the normal storage conventions

aaime:
Added some minor comments, mostly mirroring the same on the GWC side. For the rest seems pretty straightforward, and it's after all a community module.

doc/en/user/source/community/gwc-sqlite/index.rst
nmco: Documentation updated.

doc/en/user/source/community/gwc-sqlite/index.rst
nmco: Documentation updated.

nmco:
"Looking at the blobstore UI, the directories for storage and metadata should be close toghether"

Why ? The storage directory will be common to all SQLite based stores (geopackage, mbtiles, etc ...) and the MBTiles metadata directory is specific to the MBTiles blob store.

"(btw, still unsure what the metadata dirs contain to start with, isn't the metadata already contained in the mbtiles dbs?)."

The user may want to provide extra metadata like attribution or description for example.

"Also the documentation states "sqlite storage" but the dropdown states "mbtiles" instead."

Where ?

aaime:
Oh ok, so the plan is to have other variants of sqlite based storages in the future? That would explain why there are both terms.
I though the current storage was not a strict mbtiles anyways, is it fully compatible instead? (like, besides eventual extra tables, the main one is 100% compatible with mbtiles?)

aaime:
(clicked on the wrong buttong, reopened)

nmco:
Correct. MBTiles specification allow us to add all the tables we need. We stop being MBTiles compliant when we start using projections different from EPSG:436 or images formats different of PNG and JPEG.

PR #1635, opened 2016-06-09
GSIP-143 GEOS-7567 Module status for netcdf and netcdf-out

DBlasby:
Module status for netcdf and netcdf-out.  Looks for native dependencies and tries to report on them.


tbarsballe:
Looks good.

PR #1637, opened 2016-06-12
Fix typo in German translation

chrismayer:
Title says it all...


fgdrf:
Fixed it on Transifex as well https://www.transifex.com/GeoServer/geoserver/translate/#de/web_core

fgdrf:
backported to 2.9.x as well (2fb7113)

PR #1641, opened 2016-06-15
Update memory settings/defaults for Java 8

jodygarnett:


jodygarnett:
Please backport to 2.8.x

doc/en/user/source/production/container.rst
```diff
@@ -15,13 +15,41 @@ Set the following performance settings in the Java virtual machine (JVM) for you
...
 
-For more information about JVM configuration, see the article `Performance tuning garbage collection in Java <http://www.petefreitag.com/articles/gctuning/>`_.
+For more information about JVM configuration, see the article `Performance tuning garbage collection in Java <http://www.petefreitag.com/articles/gctuning/>`_ and `The 4 Java Garbage Collectors <http://blog.takipi.com/garbage-collectors-serial-vs-parallel-vs-cms-vs-the-g1-and-whats-new-in-java-8/>`_.
+
+.. note:: 
+   
+   If you're serving just vector data, you'll be streaming, so having more memory won't increase performance.  If you're serving coverages, however, image processing will use a disk cache and benifit from more memory. It is also a good idea to configure the JAI tile cache size (see the Server Config page in the :ref:`web_admin` section) so that it uses 75% of the heap (0.75).
```
aaime: A disk cache? Not that I know of?  There is the tile cache of course, which when using deferred loading can cache input tiles, but it's not doing any cross-request caching, it just avoids to re-read the same tile from disk within the confines of a single request.

We almost never set JAI to 75% in production btw, it's often set to lower values (indeed with @simboss we discussed a few times about lowering the current default).

aaime:
@jodygarnett this one could have used some review before merging... I did a very quick one right now, just looking at the obvious issues, there are more things that could be discussed (e.g., in none of the benchmarks I've done so far has G1 provided improved performance, but maybe you did some suggesting otherwise?)

Not sure what you mean by "please backport to 2.8.x", also considering that this is targeting Java 8 and 2.8.x supports Java 7 instead?

doc/en/user/source/production/container.rst
jodygarnett: I agre about the disk cache, I copied that text out ... let me check what it was before.

Check it out (http://docs.geoserver.org/latest/en/user/production/container.html#optimize-your-jvm) it says disk cache currently.

jodygarnett:
Thanks, I had intended 2.9. I have not had a convincing test of G1GC being faster, I am happy to take away the recommendation and mark the option as:

Consider experimenting with this garbage collector when using heap sizes of 6GB or more.

doc/en/user/source/production/container.rst
aaime: Eh, I tried to track down that statement in the geoserver-history, but the trail ends with a "move" commit from Mike back in 2009 (so the indication is actually older than that). It's likely that it was added in good faith, but also believe it's not correct. I'll confirm with Simone and Daniele and we'll get back here.

simboss: There is no disk cache in JAI.

Michael implemented one in JAI-Tools but it is not used in GS or GT afaik.

jodygarnett:
Thanks for review, revised text and merged to master and 2.9.x branches.

PR #1643, opened 2016-06-15
[GEOS-7578] Do not package outdated gdal-bindings jar

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-7578

When installing the GDAL extension using current versions of the native GDAL libraries, the user should use the gdal-*.jar bindings build with SWIG for the version of the GDAL binaries being used.

The imageio-ext-gdal-bindings*.jar is an older version of these GDAL bindings, and can cause classloader conflicts with current versions.


jodygarnett:
The pull request does exactly what you describe - but I would like some confirmation of the approach before merging this pull request.

tbarsballe:
I agree, some confirmation of approach would be nice.

jodygarnett:
we are coming up to the feature freeze, without confirmation (after a month) we should probably ask on the devel list and then merge.

PR #1644, opened 2016-06-15
Compress images to save disk space

johanvdw:
Both images take up a 3 Mb --> now 50 Kb


aaime:
Ouch, those files should not have got in in the first place. Given the large version is going to stay in the history though, wondering if the change is actually improving things. I cannot find evidence if git is able to compress binary files in the history or not (git + binary gets results mostly talking about large binaries and how to either manage them or get rid of them by rewriting history). Do you know if binaries gets at least compressed in the git history management?

johanvdw:
Having smaller files is definitely useful for anyone downloading the documentation (eg on OSGeo live, where I detected the files).

Git will use gzip on older blobs when packing (eg when runinning git gc), so in the history, the files will take up less space.
Adding the new images will slightly increase the size of a bare repository, but will reduce the space occupied by a checkout (by a larger margin).

PR #1645, opened 2016-06-16
[GEOS-7591] Allow WMTS service to be invoked per workspace

nmco:
Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-7591

### DIFF
--- a/src/gwc/src/main/java/org/geoserver/gwc/dispatch/GwcServiceDispatcherCallback.java
+++ b/src/gwc/src/main/java/org/geoserver/gwc/dispatch/GwcServiceDispatcherCallback.java
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
@@ -7,11 +7,21 @@ package org.geoserver.gwc.dispatch;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
+import org.geoserver.catalog.Catalog;
+import org.geoserver.catalog.WorkspaceInfo;
+import org.geoserver.gwc.layer.CatalogConfiguration;
 import org.geoserver.ows.AbstractDispatcherCallback;
 import org.geoserver.ows.Dispatcher;
 import org.geoserver.ows.DispatcherCallback;
+import org.geoserver.ows.LocalWorkspace;
 import org.geoserver.ows.Request;
+import org.geoserver.platform.ServiceException;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletRequestWrapper;
 
 /**
  * Adapts plain incoming requests to be resolved to the GWC proxy service.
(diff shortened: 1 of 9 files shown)


src/gwc/src/main/java/org/geoserver/gwc/dispatch/GwcServiceDispatcherCallback.java
```diff
@@ -29,21 +36,95 @@
...
+
+    public GwcServiceDispatcherCallback(Catalog catalog) {
+        this.catalog = catalog;
+    }
+
     @Override
     public Request init(Request request) {
         String context = request.getContext();
-        if (context == null || !context.startsWith("gwc/service")) {
+        if (context == null || !context.toLowerCase().contains("gwc/service")) {
```
jodygarnett: Is this really needed or even a good idea?

src/gwc/src/main/java/org/geoserver/gwc/dispatch/GwcServiceDispatcherCallback.java
```diff
@@ -29,21 +36,95 @@
...
+        public String getContextPath() {
+            // to GWC the workspace is part of the request context
+            return super.getContextPath() + "/" + localWorkspaceName;
+        }
+
+        @Override
+        public String getParameter(String name) {
+            if (layerName != null && name.equalsIgnoreCase("layer")) {
+                return layerName;
+            }
```
jodygarnett: Do you need to consider style references here as well?

src/gwc/src/main/java/org/geoserver/gwc/dispatch/GwcServiceDispatcherCallback.java
nmco: The virtual service workspace name will be part of the context. Let's consider this request:
`http://localhost:8080/geoserver/topp/gwc/service/wmts?REQUEST=GetCapabilities`
the context will be `topp/gwc/service`.

Regarding the good idea, what do you mean ? Using the contains to see if we are in the presence of a GWC service is to wild ?

src/gwc/src/main/java/org/geoserver/gwc/dispatch/GwcServiceDispatcherCallback.java
nmco: Good question ! Let's wait for feedback on this http://osgeo-org.1560.x6.nabble.com/Styles-and-virtual-services-td5273216.html

src/gwc/src/main/java/org/geoserver/gwc/dispatch/GwcServiceDispatcherCallback.java
aaime: I believe @jodygarnett is confused by the use of toLowerCase. OGC services are for the most part case sensitive (with the exception of key values) and the old code was not making case insensitive checks. Also, I'd probably be a bit more strict and check the path contains a known workspace and not whatever path with maybe multiple components in it.

aaime:
Added a few comments, looks otherwise good, mimics GeoServer own virtual service support, and has a good set of tests.

src/gwc/src/main/java/org/geoserver/gwc/dispatch/GwcServiceDispatcherCallback.java
nmco: Fixed and add a proper test use case for this.

src/gwc/src/main/java/org/geoserver/gwc/dispatch/GwcServiceDispatcherCallback.java
nmco: Since the behavior is the same for other virtual services I will let this like this.

nmco:
The pull request was extended to included the following behavior: if we are in the presence of a GWC GetCapabilities request in the context of virtual service we use the no-prefixed layer name. Updated files:

- src/gwc/src/main/java/org/geoserver/gwc/dispatch/GwcServiceDispatcherCallback.java
- src/gwc/src/test/java/org/geoserver/gwc/dispatch/GwcServiceDispatcherCallbackTest.java
- src/gwc/src/main/java/org/geoserver/gwc/layer/GeoServerTileLayer.java
- src/gwc/src/test/java/org/geoserver/gwc/layer/GeoServerTileLayerTest.java

This changes turn the advertised layer name trick obsolete, thanks @aaime for the idea !

aaime:
Thanks for simplifying things, merging

PR #1647, opened 2016-06-20
[GEOS-7595] Fix JDBCConfig import

tbarsballe:
See issue here: https://osgeo-org.atlassian.net/browse/GEOS-7595

Looking at the existing test case, it only test the empty config / no import case, hence why it didn't catch this issue. I would have liked to add a proper test case, but was not quite sure how to implement it properly, given that: (a) jdbcconfig relies upon an existing configuration, and a database and (b) I am not especially familiar with jdbcconfig.

While this is a community module and therefore strictly speaking does not require a test case, I would be much happier if it had one. I would appreciiate it if someone more familiar with the jdbcconfig module would either be able to add a proper test case, or give me a few pointers.

Also: @NielsCharlier - this change reverts some of the changes you made for https://github.com/geoserver/geoserver/commit/17dfdf29e71be9bcfd4eef47596681732aaad6f3. Could you comment on whether my changes seem resonable. Full explanation of changes is in the attached ticket.

### DIFF
--- a/src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/JDBCGeoServerLoader.java
+++ b/src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/JDBCGeoServerLoader.java
@@ -87,8 +87,6 @@ public class JDBCGeoServerLoader extends DefaultGeoServerLoader {
         if (config.isImport()) {
             readCatalog(catalog, xp);
             decImportStep();
-            config.setImport(false);
-            config.save();
         }
     }
 


jodygarnett:
If I understand the workflow, JDBCConfig supports an initial import (ie a migration from disk based config to jdbcconfig). This story is much better told by the pending backup and restore GSIP.

Speculation:  I assume the code you are commenting out explicitly marks the configuration on disk as "import" false (and saves that change out to disk). I assume such a marker flag would preventing JDBCConfig from reloading/replacing the JDBCConfig contents on startup?

tbarsballe:
Not quite. If you look at that file in its entirety, you will see `private int importSteps = 2;`
The import is processed in two stages - first the GeoServer catalog (stores, layers, etc.) is loaded, then the GeoServer configuration (service config, contact info, etc.) is loaded. After each of these stages, importSteps is decremented. When importSteps reaches zero, the on-disk configuration (jdbcconfig.properties) has the import flag set to false.

The code I removed incorectly sets this flag to false after the first stage, meaning that the GeoServer configuration is never imported.

NielsCharlier:
Okay, fair enough

tbarsballe:
Merged, and backported to 2.9.x

PR #1648, opened 2016-06-21
[GEOS-7597] Add JNA to netcdf-out release configuration

tbarsballe:
Fairly basic configuration change based on testing the netcdf-out extension. It turns out it was missing the jar required for native linking (required for NetCDF 4 support).


bencaradocdavies:
@tbarsballe I made extensive use of NetCDF-4 output with the 2.9.x nightly from 17 June. I think you are right: this only works for me because I also have the grib plugin, which includes the JNA jar!

dromagnoli:
Thanks for the fix @tbarsballe. Could you also backport it, please?

tbarsballe:
Backported to 2.9.x and 2.8.x

PR #1649, opened 2016-06-23
Status grib

DBlasby:
GEOS-7567 GSIP-143  Module Status for GRIB Extension 

Grib uses the netcdf library

### DIFF
--- a/src/extension/grib/src/main/java/org/geoserver/grib/GribStatus.java
+++ b/src/extension/grib/src/main/java/org/geoserver/grib/GribStatus.java
@@ -1,0 +1,59 @@
+package org.geoserver.grib;
+
+import java.util.Optional;
+
+import org.geoserver.platform.ModuleStatus;
+import org.geotools.factory.GeoTools;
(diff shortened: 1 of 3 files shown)


src/extension/grib/src/main/java/org/geoserver/grib/GribStatus.java
```diff
@@ -0,0 +1,59 @@
+package org.geoserver.grib;
```
tbarsballe: Missing copyright header

src/extension/grib/src/main/java/org/geoserver/grib/GribStatus.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public boolean isAvailable() {
+        return NetCDFUtilities.isGribAvailable();
+    }
+
+    @Override
+    public boolean isEnabled() {
+        return true;
```
tbarsballe: Is there any way GRIB could not be enabled?

tbarsballe:
This looks good now. I have manually squashed and merged it into master.

PR #1650, opened 2016-06-23
Match geotools commons-pool version

tbarsballe:
GeoTools uses commons-pool 1.5.4 via gt-referencing
GeoServer explicitly requires commons-pool 1.5.3 in the top level pom. However, the only dependancies that use this are gt-referencing (via gs-wcs) and commons-dbcp (via gs-jdbcconfig), both of which actually depend upon 1.5.4.
The GeoServer dependancy appears necessary in order for the release module to actually pick up the jar during packaging.
This PR fixes the patch version so that we are actually consistent with the dependancies that are pulling this jar.

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -726,7 +726,7 @@
    <dependency>
     <groupId>commons-pool</groupId>
     <artifactId>commons-pool</artifactId>
-    <version>1.5.3</version>
+    <version>1.5.4</version>
    </dependency>
    <dependency>
     <groupId>commons-collections</groupId>


src/pom.xml
```diff
@@ -726,7 +726,7 @@
    <dependency>
     <groupId>commons-pool</groupId>
     <artifactId>commons-pool</artifactId>
-    <version>1.5.3</version>
+    <version>1.5.4</version>
```
jodygarnett: Is this dependency even used or can it be transitive?

jodygarnett:
How far does this get backported?

tbarsballe:
It hasn't changed in geoserver or geotools since before we imported the repo into git (pre 2011), so we can backport as far back as you like. 2.8.x and 2.9.x at minimum.

PR #1654, opened 2016-06-28
GEOS-7608 GEOS-7567 GSIP-143 bug fix and improved module status page

DBlasby:
Fixed bug in ModuleStatusImpl (wasn't actually recording isEnabled/isAvailable) - GEOS-7608
Added test case

Added isEnabled/isAvailable to HTML and JSON output.
Added test case (there wasn't any existing REST tests I could find)

### DIFF
--- a/src/platform/src/main/java/org/geoserver/platform/ModuleStatusImpl.java
+++ b/src/platform/src/main/java/org/geoserver/platform/ModuleStatusImpl.java
@@ -5,49 +5,65 @@
 package org.geoserver.platform;
 
 import java.util.Optional;
+
 /**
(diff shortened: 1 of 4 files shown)


tbarsballe:
A little bit excessive on the whitespace changes (generally, major formatting changes should not be mixed in with functional changes), but this is a small enough changeset that I don't have too much of an issue with it. Looks good overall.

tbarsballe:
Travis CI passed, I'll go ahead and merge. Does this affect earlier releases / does this need backporting? The JIRA ticket is rather lacking in information.

PR #1655, opened 2016-06-28
GEOS-7567 GSIP-143 netcdf status page - add native lib version (if available)

DBlasby:
Ask the native lib for its version (if available), put in Status Page message.

### DIFF
--- a/src/extension/netcdf/src/main/java/org/geoserver/netcdf/NetCDFStatus.java
+++ b/src/extension/netcdf/src/main/java/org/geoserver/netcdf/NetCDFStatus.java
@@ -11,6 +11,8 @@ import org.geotools.factory.GeoTools;
 import org.geotools.imageio.netcdf.utilities.NetCDFUtilities;
 import org.geotools.util.Version;
 
+import com.sun.jna.Native;
+
 import ucar.nc2.jni.netcdf.Nc4prototypes;
 
 public class NetCDFStatus implements ModuleStatus {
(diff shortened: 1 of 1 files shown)


tbarsballe:
This does not compile. Also, should it really be pulling in jna as a dependency?
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:2.3.2:compile (default-compile) on project gs-netcdf: Compilation failure: Compilation failure:
[ERROR] /home/travis/build/geoserver/geoserver/src/extension/netcdf/src/main/java/org/geoserver/netcdf/NetCDFStatus.java:[14,18] error: package com.sun.jna does not exist
[ERROR] /home/travis/build/geoserver/geoserver/src/extension/netcdf/src/main/java/org/geoserver/netcdf/NetCDFStatus.java:[56,48] error: cannot find symbol
[ERROR] class NetCDFStatus
[ERROR] /home/travis/build/geoserver/geoserver/src/extension/netcdf/src/main/java/org/geoserver/netcdf/NetCDFStatus.java:[57,22] error: cannot access Library
```

DBlasby:
Torben - it needs jna to communicate with the native lib.  It's just asking the native lib for it's version.

I can move this over to netcdf-out, which definitely requires the jna jar.

tbarsballe:
Nowhere does netcdf or netcdf ecplicitly require the `com.sun.jna` package. The only jna dependency comes from the `net.java.dev.jna` package (not the same as `com.sun.jna`), provided transitively through the edu.ucar netcdf4 library.

Basically, you are using the wrong jna library.

DBlasby:
Humm - I copied the code right out of the edu.ucar library (the load() function logs the library version number).  I will check to see which jna its depending on.

tbarsballe:
Looks like edu.ucar includes it as a runtime dependency, not a compile time dependency. This makes sense as it is an optional jar (netcdf only requires jna for NetCDF v4 output, otherwise defaults to v3).

Consequently, we should not have jna as a compile-time dependeny. Perhaps you can do something with placeholder jars similar to how the oracle extension handles the oracle driver dependency?

PR #1656, opened 2016-06-29
GEOS-7567 GSIP-143 netcdf-out status page - add native lib version (if available)

DBlasby:
Ask the native lib for its version (if available), put in Status Page message.

This implementation uses reflection because JNA is an optional jar.

### DIFF
--- a/src/extension/netcdf-out/src/main/java/org/geoserver/netcdfout/NetCDFOutStatus.java
+++ b/src/extension/netcdf-out/src/main/java/org/geoserver/netcdfout/NetCDFOutStatus.java
@@ -5,6 +5,7 @@
 
 package org.geoserver.netcdfout;
 
+import java.lang.reflect.Method;
 import java.util.Optional;
 
 import org.geoserver.platform.ModuleStatus;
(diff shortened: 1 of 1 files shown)


src/extension/netcdf-out/src/main/java/org/geoserver/netcdfout/NetCDFOutStatus.java
```diff
@@ -50,10 +51,33 @@ public boolean isEnabled() {
         return true;
     }
 
+    public String Nc4Version() {
+        try {
+                //This reflection code is to deal with JNA being an optional jar.
+                //Its the same as;
+                //     Nc4prototypes nc4 = (Nc4prototypes) Native.loadLibrary("netcdf", Nc4prototypes.class);
+                //     return nc4.nc_inq_libvers();
+            Class com_sun_jan_Native = Class.forName("com.sun.jna.Native");
```
tbarsballe: This indentation looks a little wonky - are there tabs hiding in here?

tbarsballe: Variable names should not contain underscores - see the GeoServer style guide.

DBlasby: Not sure what happened - git wasn't picking up my changes.  
Should be correct now.

tbarsballe:
This looks like a viable approach for getting around the runtime-only dependency. While I don't like adding more reflection, this seems like the lesser evil. 

Assuming the build passes on Travis, I am fine to merge this.

Thanks.

PR #1657, opened 2016-06-29
GSIP-143 minor typos noticed in module names

DBlasby:
fixing module names


tbarsballe:
It concerns me that the application context typos weren't caught by existing test cases - are there any integration tests for these?

The changes look good overall.

tbarsballe:
Okay, I see that the stuff in the .xml files is tust a string argument containing the name. In that case, I have no concerns with this one.

PR #1658, opened 2016-06-29
GSIP-143 make html status page more readable - newline to html br

DBlasby:


tbarsballe:
It would be better if you finalized your module status layout before commiting, to avoid all these small PRs for module status (typos, etc).
However, small commits are easier to review and test than giant monolithic ones, so I am not complaining too much.
It will be a pain to backport everything though.

PR #1660, opened 2016-06-30
GSIP-143 archtype test for status page HTML via rest

DBlasby:


DBlasby:
This was missed from an earlier pull request.

PR #1661, opened 2016-06-30
PR: (GEOS-7611) Recursive copy method on Resources fails with infinite recursion

afabiani:


simboss:
Looking good, test is present. Merging.

PR #1663, opened 2016-06-30
PR: (GEOS-7614) Allowing GeoServer to store property placeholders as templates into the catalog

afabiani:
--- WARNING: requires GeoWebCache PR https://github.com/GeoWebCache/geowebcache/pull/399

### DIFF
--- a/src/gwc/src/main/java/org/geoserver/gwc/GWC.java
+++ b/src/gwc/src/main/java/org/geoserver/gwc/GWC.java
@@ -56,9 +56,9 @@ import org.geoserver.gwc.layer.GeoServerTileLayer;
 import org.geoserver.gwc.layer.GeoServerTileLayerInfo;
 import org.geoserver.gwc.layer.GeoServerTileLayerInfoImpl;
 import org.geoserver.ows.Dispatcher;
-import org.geoserver.ows.LocalWorkspace;
 import org.geoserver.ows.Request;
 import org.geoserver.ows.Response;
+import org.geoserver.platform.GeoServerEnvironment;
 import org.geoserver.platform.GeoServerExtensions;
 import org.geoserver.platform.Operation;
 import org.geoserver.security.AccessLimits;
@@ -79,6 +79,7 @@ import org.geotools.ows.ServiceException;
 import org.geotools.referencing.CRS;
 import org.geotools.referencing.CRS.AxisOrder;
 import org.geotools.util.logging.Logging;
+import org.geowebcache.GeoWebCacheEnvironment;
 import org.geowebcache.GeoWebCacheException;
 import org.geowebcache.GeoWebCacheExtensions;
 import org.geowebcache.config.BlobStoreConfig;
@@ -216,6 +217,9 @@ public class GWC implements DisposableBean, InitializingBean, ApplicationContext
     
     private FilterFactory2 ff = CommonFactoryFinder.getFilterFactory2();
     
+    private GeoWebCacheEnvironment gwcEnvironment;
+    
+    final GeoServerEnvironment gsEnvironment = GeoServerExtensions.bean(GeoServerEnvironment.class);
     
     public GWC(final GWCConfigPersister gwcConfigPersister, final StorageBroker sb,
             final TileLayerDispatcher tld, final GridSetBroker gridSetBroker,
@@ -279,6 +283,7 @@ public class GWC implements DisposableBean, InitializingBean, ApplicationContext
                         + " found by GeoServerExtensions");
             }
         }
+        GWC.INSTANCE.syncEnv();
         return GWC.INSTANCE;
     }
 
@@ -2202,6 +2207,28 @@ public class GWC implements DisposableBean, InitializingBean, ApplicationContext
     @Override
     public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
         this.applicationContext = applicationContext;
+        
+        this.gwcEnvironment = GeoServerExtensions.bean(GeoWebCacheEnvironment.class);
+        
+        syncEnv();
+    }
+
+    /**
+     * @throws IllegalArgumentException
+     */
+    public void syncEnv() throws IllegalArgumentException {
+        if (gsEnvironment != null && gsEnvironment.isStale() && gwcEnvironment != null) {
+            if (GeoServerEnvironment.ALLOW_ENV_PARAMETRIZATION && gsEnvironment.getProps() != null) {
+                Properties gwcProps = gwcEnvironment.getProps();
+
+                if (gwcProps == null) {
+                    gwcProps = new Properties();
+                }
+                gwcProps.putAll(gsEnvironment.getProps());
+
+                gwcEnvironment.setProps(gwcProps);
+            }
+        }
     }
     
     /**
@@ -2414,5 +2441,12 @@ public class GWC implements DisposableBean, InitializingBean, ApplicationContext
         checkNotNull(compositeBlobStore);
         return compositeBlobStore;
     }
-    
+
+    /**
+     * @return the gwcEnvironment
+     */
+    public GeoWebCacheEnvironment getGwcEnvironment() {
+        return gwcEnvironment;
+    }
+
 }
--- a/src/main/src/test/java/org/geoserver/security/GeoServerSecurityManagerTest.java
+++ b/src/main/src/test/java/org/geoserver/security/GeoServerSecurityManagerTest.java
@@ -16,6 +16,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Properties;
 
+import org.geoserver.platform.GeoServerEnvironment;
 import org.geoserver.platform.resource.Files;
 import org.geoserver.security.config.SecurityManagerConfig;
 import org.geoserver.security.impl.GeoServerRole;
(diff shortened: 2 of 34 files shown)


src/main/src/main/java/org/geoserver/config/ServiceInfo.java
```diff
@@ -238,4 +238,10 @@
     
     Map<Object,Object> getClientProperties();
 
+    /**
+     *
```
aaime: Lacks Javadoc in an interface. The "allowEnvParametrization" seems more like a "replaceParameters" to me

src/main/src/main/java/org/geoserver/config/impl/ServiceInfoImpl.java
```diff
@@ -247,35 +257,26 @@ public void setVerbose(boolean verbose) {
     public int hashCode() {
         final int prime = 31;
         int result = 1;
-        result = prime * result + ((abstrct == null) ? 0 : abstrct.hashCode());
-        result = prime
-                * result
-                + ((accessConstraints == null) ? 0 : accessConstraints
-                        .hashCode());
+        result = prime * result + ((getAbstract() == null) ? 0 : getAbstract().hashCode());
```
aaime: Confused, why is the commit changing all field access to getters? The method has visibility of the fields and I see no late/lazy inizialization of them.  This kind of change is likely going to be reverted next time someone uses a IDE code generator to update hashCode/equals

src/extension/csw/core/src/main/java/org/geoserver/csw/GetCapabilities.java
```diff
@@ -278,6 +278,8 @@ public GetCapabilities(CSWInfo csw, CatalogStore store, ApplicationContext conte
     }
 
     public CapabilitiesType run(GetCapabilitiesType request) throws CSWException {
+        CSWInfo expandedCsw = (CSWInfo) csw.clone(true);
```
aaime: Thinking out loud here, instead of spreading changes like this all over the place, making all the code aware of the parametrization, isn't it better to have GeoServer.getService return a cloned object for all situation where we know in advance we are not editing configuration? Like, a special thread local could be set to inform GeoServerImpl that the current request is a REST or a GUI one, and maybe we could have a public method to set the mode (expanded or parametric) for any other custom code that needs to work in parametric mode.

To make a parallel with security, we could have made all the code security aware, but it would have been a massive maintenance chore, instead we apply security transparently. I'm hoping parametrization could be done the same way.

I've checked the proposal, it was generally talking about xstreamloaders (which would have been a very centralized point, only a few classes) but would not have allowed to edit the params via the UI (so I understand why a different path was attempted)

src/extension/csw/core/src/test/java/org/geoserver/csw/CSWSimpleTestSupport.java
```diff
@@ -17,6 +17,7 @@
     protected void setUpTestData(SystemTestData testData) throws Exception {
         testData.setUpSecurity();
         System.setProperty("DefaultCatalogStore", "org.geoserver.csw.store.simple.GeoServerSimpleCatalogStore");
+        System.setProperty("ALLOW_ENV_PARAMETRIZATION", "true");
```
aaime: Probably best to set it in the base test class, and make sure it gets cleaned at the end of the test?

src/gwc/src/main/java/org/geoserver/gwc/GWC.java
```diff
@@ -319,6 +323,9 @@ public Catalog getCatalog() {
     }
 
     public GWCConfig getConfig() {
+        if (gsEnvironment != null && gsEnvironment.isStale()) {
```
aaime: This bloc of code is repeated many times, should be centralized in a single method.

src/main/src/main/java/org/geoserver/catalog/impl/CatalogImpl.java
```diff
@@ -680,10 +680,12 @@ public ValidationResult validate( LayerInfo layer, boolean isNew) {
         
         // clean up eventual dangling references to missing alternate styles
         Set<StyleInfo> styles = layer.getStyles();
-        for (Iterator it = styles.iterator(); it.hasNext();) {
-            StyleInfo styleInfo = (StyleInfo) it.next();
-            if(styleInfo == null) {
-                it.remove();
+        if (styles != null) {
```
aaime: The styles collection is a "live" one, can never be null (there is no setter, so no way to set it to a non null value), if it happens, then a significant bug has been introduced or was already present, and should be fixed instead.

src/main/src/main/java/org/geoserver/catalog/impl/ResourceInfoImpl.java
```diff
@@ -376,35 +376,21 @@ public void setAdvertised(boolean advertised) {
     public int hashCode() {
         final int prime = 31;
         int result = 1;
-        result = prime * result
-                + ((_abstract == null) ? 0 : _abstract.hashCode());
+        result = prime * result + ((_abstract == null) ? 0 : _abstract.hashCode());
```
aaime: Looks like pure reformats, if so, can you avoid modifying this file at all?

src/main/src/main/java/org/geoserver/config/impl/GeoServerEnvironmentAwareFacadeWrapper.java
```diff
@@ -0,0 +1,237 @@
...
+import org.geoserver.config.GeoServerInfo;
+import org.geoserver.config.LoggingInfo;
+import org.geoserver.config.ServiceInfo;
+import org.geoserver.config.SettingsInfo;
+import org.geoserver.ows.Dispatcher;
+import org.geoserver.ows.Request;
+import org.geoserver.platform.GeoServerEnvironment;
+
+/**
+ * @author Alessio Fabiani, GeoSolutions
```
aaime: Public class, no javadoc

src/main/src/main/java/org/geoserver/config/impl/GeoServerEnvironmentAwareFacadeWrapper.java
```diff
@@ -0,0 +1,237 @@
...
+        }
+        
+        return wrappedServices;
+    }
+
+    @Override
+    public Collection<? extends ServiceInfo> getServices(WorkspaceInfo workspace) {
+        Collection wrappedServices = new ArrayList();
+        Request request = Dispatcher.REQUEST.get();
+        if (request != null && GeoServerEnvironment.ALLOW_ENV_PARAMETRIZATION) {
```
aaime: Duplicates code in the above method. Maybe get the services from the wrapped entity, and then call a method that will filter them and expand their variables? (mind, jdbc config returns a immutable list, make sure you keep on filtering by creating a new list)

src/main/src/main/java/org/geoserver/security/decorators/DecoratingWMSLayerInfo.java
```diff
@@ -230,6 +230,4 @@ public boolean isAdvertised() {
     public void setAdvertised(boolean advertised) {
         delegate.setAdvertised(advertised);
     }
-
```
aaime: Pure formatting change, please avoid (here and in the other files too, it also makes the pull request look bigger than it is)

src/main/src/test/java/org/geoserver/security/GeoServerSecurityManagerTest.java
```diff
@@ -188,4 +189,25 @@ public void testWebLoginChainSessionCreation() throws Exception {
...
     }
+    
+    @Test public void testGeoServerEnvParametrization() throws Exception {
+        GeoServerSecurityManager secMgr = getSecurityManager();
+        SecurityManagerConfig config = secMgr.loadSecurityConfig();
+        String oldRoleServiceName = config.getRoleServiceName();
+        
+        try {
+            if (GeoServerEnvironment.ALLOW_ENV_PARAMETRIZATION) {
+                System.setProperty("TEST_SYS_PROPERTY", oldRoleServiceName);
```
aaime: I'd suggest to clear the property in finally, might create accidental coupling with other tests

aaime: This is valid in general also for other tests that set sysvars

src/restconfig/src/main/java/org/geoserver/catalog/rest/CoverageStoreFileResource.java
```diff
@@ -194,17 +194,16 @@ public void handlePut() {
         }
         
         //add or update the datastore info
-        if ( add ) {
```
aaime: Pure reformat

src/restconfig/src/main/java/org/geoserver/service/rest/ServiceSettingsResource.java
```diff
@@ -76,18 +66,19 @@ protected Object handleObjectGet() throws Exception {
                     Status.CLIENT_ERROR_NOT_FOUND);
         }
 
-        return service;
+        return service.clone(true);
```
aaime: Given this is configuration, wouldn't it make more sense to show the parameters in the output, accept them in input, and then fix validation like it's done in the UI?

src/wcs1_0/src/main/java/org/geoserver/wcs/response/Wcs10CapsTransformer.java
```diff
@@ -53,6 +50,9 @@
 import org.xml.sax.SAXException;
 import org.xml.sax.helpers.AttributesImpl;
 
+import net.opengis.wcs10.CapabilitiesSectionType;
```
aaime: Pure reformat

src/wcs1_0/src/test/java/org/geoserver/wcs/GetCapabilitiesTest.java
```diff
@@ -254,6 +256,7 @@ public void testWorkspaceQualified() throws Exception {
         
         expected = getCatalog().getCoverageStoresByWorkspace(MockData.CDF_PREFIX).size();
         dom = getAsDOM("cdf/wcs?request=GetCapabilities&service=WCS&version=1.0.0");
+        print(dom);
```
aaime: Comment out

PR #1664, opened 2016-07-01
PR: (GEOS-7616) Allowing plugins to be discovered through appContext and reload/store conifguration via Interface

afabiani:


src/extension/control-flow/src/main/java/org/geoserver/flow/config/DefaultControlFlowConfigurator.java
```diff
@@ -195,4 +202,53 @@ public long getTimeout() {
...
+        }
+    }
+
+    @Override
+    public void loadConfiguration(GeoServerResourceLoader resourceLoader) throws IOException {
+        synchronized (this) {
+            Resource controlflow = resourceLoader.get(PROPERTYFILENAME);
+            if (Resources.exists(controlflow)) {
+                configFile = new PropertyFileWatcher(controlflow);
+                configFile.setKnownLastModified(System.currentTimeMillis());
```
aaime: Done this way, isStale() is not going to return true after a configuration restore, or not? I may be wrong, but given it's already look at the files, it's likely that leaving this method empty will already do the trick... although looking at the FileWatcher code, it might fail if the file was there before, and gets removed later. But I don't see a reason to re-create the watcher otherwise.

aaime: Ah, I see you're doing it because a test is using a fixedwatcher. The consideration about isStale is still valid though.

src/extension/monitor/core/src/main/java/org/geoserver/monitor/MonitorConfig.java
```diff
@@ -228,5 +230,67 @@ Properties props() {
         return props;
     }
 
+    @Override
+    public List<Resource> getFileLocations() throws IOException {
+        List<Resource> configurationFiles = new ArrayList<>();
+        GeoServerResourceLoader loader = GeoServerExtensions.bean(GeoServerResourceLoader.class);
```
aaime: The loader is already available via construtor injection, why look it up dynamically at runtime?

src/extension/monitor/core/src/test/java/org/geoserver/monitor/ows/MonitorCallbackTest.java
```diff
@@ -560,4 +568,29 @@ public void testWMSGetFeatureInfoDifferentCrs() throws Exception {
         BBoxAsserts.assertEqualsBbox(new ReferencedEnvelope(48.4,48.4,-123.3,-123.3,logCrs),data.getBbox(), 0.1);
     }
 
+    @Test
+    public void testConfiguration() throws IOException {
+        System.out.println(monitor.getConfig().getFileLocations());
```
aaime: Leftover printout (comment it out)

src/main/src/main/java/org/geoserver/config/impl/GeoServerImpl.java
```diff
@@ -448,8 +453,22 @@ public void reload() throws Exception {
...
             synchronized (org.geoserver.config.GeoServer.CONFIGURATION_LOCK) {
-                getCatalog().getResourcePool().dispose();
-                loader.reload();
+                getCatalog().getResourcePool().dispose();                
+
+                if(catalog != null) {
+                    dispose();
+                    
+                    //reload catalog, make sure we reload the underlying catalog, not any wrappers
+                    Catalog oldCatalog = getCatalog();
```
aaime: This is actually "the" catalog, the one that we are getting as a parameter is the one that we will copy from, and that will be thrown away right? If so, better name them newCatalog (the parameter) and catalog (the actual gs catalog)

src/main/src/main/java/org/geoserver/config/GeoServerPluginConfigurator.java
```diff
@@ -0,0 +1,32 @@
+/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
```
aaime: The header should only say "2016 Open Source Geospatial Foundation - all rights reserved", without Openplans below

src/main/src/main/java/org/geoserver/config/GeoServerPluginConfigurator.java
```diff
@@ -0,0 +1,32 @@
...
+ * let GeoServer to be aware of their extra-configuration files (usually properties
+ * files customized and specific for each plugin - see control-flow, monitoring and so on...).
+ * 
+ * 
+ * @author Alessio Fabiani, GeoSolutions
+ *
+ */
+public interface GeoServerPluginConfigurator {
+    
+    public List<Resource> getFileLocations() throws IOException;
```
aaime: Public interface methods without javadocs

src/main/src/main/java/org/geoserver/config/GeoServerPropertyConfigurer.java
```diff
@@ -114,4 +114,12 @@ protected void loadProperties(Properties props) throws IOException {
             }
         }
     }
+    
+    public void reload() throws IOException {
```
aaime: Is this new method actually used anywhere?

aaime: Answering myself, yes, by the new backup/restore module.
New public method, please add javadoc

src/main/src/main/java/org/geoserver/security/GeoServerSecurityManager.java
```diff
@@ -364,66 +364,71 @@ public ApplicationContext getApplicationContext() {
...
-                }
-                if (securityVersion.compareTo(CURR_VERSION) < 0) {
-                    writeCurrentVersion();
-                }
-            } catch (Exception e1) {
-                throw new RuntimeException(e1);
+    /**
+     * 
+     */
+    public void reload() {
```
aaime: New public method, please add javadoc (this one is used by backup/restore too)

src/extension/control-flow/src/main/java/org/geoserver/flow/config/DefaultControlFlowConfigurator.java
afabiani: I'm not sure I fully understood your observation, however the idea here is that through this method you can reload the configuration from a different ResourceLoader.

aaime: My remaining point is that by creating a new watcher and setting the last modified to the current time, the next call to "isStale()" will return false and control-flow won't notice the new (restored) property file

PR #1666, opened 2016-07-04
PR: (GEOS-7619) Allow the Wicket UI to show a Server Busy page when updating the configuration instead of locking the server

afabiani:
The idea behind this proposal is the following one:

Currently GeoServer tries to acquire an exclusive lock when accessing pages which can read/write the configuration through the WicketConfigurationCallback, thus causing the UI to hang until the configuration has been released by the catalog.
The problem is that in some cases, if the configuration change requires too much, the UI is not accessible at all.

With this PR I would like to address to goals:

1. Avoid completely freeze the UI when the configuration is locked

2. Allow a developer (who knows what he is doing) to create pages which can be accessed also when the configuration is locked by some process (thing for instance to an extension which needs to safely cancel/stop a long write operation)

The proposed solution is the following one:

- Implement a "tryLock" method on the global GeoServerLock, which returns true/false if the configuration can be taken or not.

- Test the lock before rendering the page: if the lock can be taken, the page will be rendered as usual, otherwise the server will redirect the user to a "ServerBusy" page.

- A new interface "GeoServerUnlockablePage" allows us to recognize the pages which can be always rendered, skipping the locking checks.


src/web/core/src/main/java/org/geoserver/web/ServerBusyPage.java
```diff
@@ -0,0 +1,21 @@
+/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
```
aaime: New page, wrong header, it should be the same as GeoServerUnlockablePage

src/web/core/src/main/java/org/geoserver/web/WicketConfigurationLockCallback.java
```diff
@@ -45,30 +47,59 @@ public void onEndRequest() {
...
+                THREAD_LOCK.set(type);
+                THREAD_REQUESTED_PAGE.set(requestTarget);
+            }
         }
+    }
+    
+    @Override
+    public void onRequestTargetSet(RequestCycle cycle,
+            Class<? extends IRequestablePage> requestTarget) {
+        WicketCallback.super.onRequestTargetSet(cycle, requestTarget);
```
aaime: I'd call directly the other method, or just merge them into one, after all, the other method is not going to get called anymore from GeoServerApplication right?

src/main/src/main/java/org/geoserver/GeoServerConfigurationLock.java
```diff
@@ -25,6 +28,10 @@
  * 
  */
 public class GeoServerConfigurationLock {
+    
+    /** DEFAULT_TRY_LOCK_TIMEOUT_MS */
+    private static final int DEFAULT_TRY_LOCK_TIMEOUT_MS = 5000;
```
aaime: Don't hard code, use GeoServerExtensions.getProperty to allow admins to set this value in multiple ways (even better would be to have a real config option in the global settings, but I understand it's more work and the default should be sane)

aaime: I see it's done down in the code, but it should be doable once here instead

src/web/core/src/main/java/org/geoserver/web/WicketConfigurationLockCallback.java
```diff
@@ -45,30 +47,59 @@ public void onEndRequest() {
...
 
     @Override
     public void onRequestTargetSet(Class<? extends IRequestablePage> requestTarget) {
-        // we can have many of these calls per http call, avoid locking multiple times,
-        // onEndRequest will be called just once
-        LockType type = THREAD_LOCK.get();
-        if (type != null || requestTarget == null) {
-            return;
+        if (!GeoServerUnlockablePage.class.isAssignableFrom(requestTarget)) {
+            // we can have many of these calls per http call, avoid locking multiple times,
```
aaime: We should have just one, where did you see multiple ones?

PR #1667, opened 2016-07-04
[GEOS-7621] Coverage band details editing, issues with locale, list separators and decimal numbers representation

aaime:
@dromagnoli this one is for you to review (as the original author of the editable version of the band properties)


dromagnoli:
Looks good to me. Thanks for the fixes.

PR #1670, opened 2016-07-08
[GEOS-7627] Support "recurse" deletion of style referenced by layers via REST

dalbani:
https://osgeo-org.atlassian.net/browse/GEOS-7627
(incorrect JIRA ID fixed thanks to @afabiani)


jodygarnett:
Thanks for the patch - looks to be complete, with documentation and test-case.

PR #1686, opened 2016-07-19
Mbtiles support 2.9.x

nmco:
Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-7645

Depends on: https://github.com/GeoWebCache/geowebcache/pull/402


nmco:
Realized that GeoServer GWC is not dependent of GWC Spring configuration. To avoid any unexpected side effects the MBTiles blobstore configuration inclusion could be commented. If someone wants to use the MBTiles blobstore he just has to uncomment that line.  Pull request updated.

PR #1675, opened 2016-07-09
[GEOS-7629] GSIP-147 Align with Geotools copyright header management practices

bradh:
This implements the documentation changes as proposed under GSIP-147.
See https://github.com/geoserver/geoserver/wiki/GSIP%20147 for proposal.

It also fixes a few broken links in the affected files, which came up during testing.

PR #1687, opened 2016-07-19
GEOS-7646: Applying default conversion to feature type for PostGIS.

jdeolive:
The default conversion applies logic to rename attributes with spaces or other
cases where the name is not a valid XML element name. This matches the behaviour
of the conversion that is done on the feature by feature level.


aaime:
Looks good to me

jdeolive:
Thanks @aaime . I once again forgot to update copyright headers... is the new policy following the GeoTools decision up and running yet?

aaime:
Yep

jdeolive:
Awesome, thanks! Any issues with a backport on this one? Or should I let it simmer for a while?

aaime:
No simmering for fixes, just for features

bencaradocdavies:
@jdeolive the new copyright header policy is accepted and the documentation updated in https://github.com/geoserver/geoserver/pull/1675 . Good thing you forgot or we would have to make you change them back.  ;-)

PR #1688, opened 2016-07-19
PR: (GEOS-7619) Allow the Wicket UI to show a Server Busy page when updating the configuration instead of locking the server

afabiani:
The idea behind this proposal is the following one:

Currently GeoServer tries to acquire an exclusive lock when accessing pages which can read/write the configuration through the WicketConfigurationCallback, thus causing the UI to hang until the configuration has been released by the catalog.
The problem is that in some cases, if the configuration change requires too much, the UI is not accessible at all.

With this PR I would like to address to goals:

1. Avoid completely freeze the UI when the configuration is locked

2. Allow a developer (who knows what he is doing) to create pages which can be accessed also when the configuration is locked by some process (thing for instance to an extension which needs to safely cancel/stop a long write operation)

The proposed solution is the following one:

- Implement a "tryLock" method on the global GeoServerLock, which returns true/false if the configuration can be taken or not.

- Test the lock before rendering the page: if the lock can be taken, the page will be rendered as usual, otherwise the server will redirect the user to a "ServerBusy" page.

- A new interface "GeoServerUnlockablePage" allows us to recognize the pages which can be always rendered, skipping the locking checks.

### DIFF
--- a/src/main/src/main/java/org/geoserver/GeoServerConfigurationLock.java
+++ b/src/main/src/main/java/org/geoserver/GeoServerConfigurationLock.java
@@ -5,11 +5,13 @@
  */
 package org.geoserver;
 
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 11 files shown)


aaime:
Looks good to me. The old pull request has been out for 2+ weeks with no other comments... if nobody wants to have a look I'll merge this one tomorrow

PR #1692, opened 2016-07-22
[GEOS-7467] Make JMS cluster plugin only serialize the necessary data

nmco:
Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-7467

JMD plugin doesn't serialize and send the whole store anymore. A test case was added for this.


aaime:
Makes sense to me, +1

PR #1695, opened 2016-07-22
[GEOS-5090] Show styles to be deleted in GUI notification

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-5090

### DIFF
--- a/src/web/core/src/test/java/org/geoserver/web/data/ConfirmRemovalPanelTest.java
+++ b/src/web/core/src/test/java/org/geoserver/web/data/ConfirmRemovalPanelTest.java
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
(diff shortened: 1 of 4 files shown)


jodygarnett:
Quesiton - can this be backported as a bug fix? Or is it a "new" feature?

src/web/core/src/test/java/org/geoserver/web/data/ConfirmRemovalPanelTest.java
```diff
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
```
jodygarnett: No need to update the header anymore :)

jodygarnett:
Looks good!

src/web/core/src/test/java/org/geoserver/web/data/ConfirmRemovalPanelTest.java
tbarsballe: Is that now official? Excellent. Has the Contributing guide been updated?

tbarsballe:
This can be backported as far back as we allow workspaced styles, so yes. I will do so after merging.

PR #1697, opened 2016-07-22
[GEOS-7650] Can't delete Default Cached Gridsets

emerkle826:


src/web/gwc/src/main/java/org/geoserver/gwc/web/DefaultGridsetsEditor.java
```diff
@@ -98,7 +99,14 @@ protected void onClick(AjaxRequestTarget target) {
...
                 Property<GridSet> property) {
-            // TODO Auto-generated method stub
+            // Property objects are package access, so we can't statically reference them here
+            // see org.geoserver.gwc.web.gridset.ACTION_LINK
+            final String propertyName = property.getName();
+            // the Remove link property name is the empty string. If that is the property name,
+            // return the actionLink here.
+            if (Strings.isEmpty(propertyName)) {
+                return actionLink(id, itemModel.getObject().getName());
+            }
```
jodygarnett: This is hard to reasonably write a test case for - I am interested in backporting this to 2.9.x as it is a regression brought on by the wicket upgrade.

PR #1700, opened 2016-07-22
[GEOS-7342] Filter layers and layer groups by workspace when adding to layer group

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-7342

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/data/layergroup/LayerGroupEntryPanel.java
+++ b/src/web/core/src/main/java/org/geoserver/web/data/layergroup/LayerGroupEntryPanel.java
@@ -114,7 +114,7 @@ public class LayerGroupEntryPanel extends Panel {
                 popupWindow.setInitialHeight( 375 );
                 popupWindow.setInitialWidth( 525 );
                 popupWindow.setTitle(new ParamResourceModel("chooseLayer", this));
-                popupWindow.setContent( new LayerListPanel(popupWindow.getContentId()) {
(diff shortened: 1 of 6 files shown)


dvntucker:
Looks OK to me.

tbarsballe:
The WCS NetCDF output Module seems to be causing the Travis build to fail sporadicly. I'm going to try re-running the build.

bencaradocdavies:
The Travis CI logs suggest that the first call to WCSTestSupport had some sort of schema failure; I suspect a network problem connecting to http://schema.opengis.net/ as we have from time to time seen things like this before. Earlier today I exchanged email with the chair of an OGC committee who was complaining about it, I so I guess the problem is at the OGC end.

aaime:
Did not try it interactively, but wondering... what happens if someone changes the workspace while editing? Or switches from global to ws specific? E.g., is the switch being picked up by the editors, and what happens if one saves with layers not in the current workspace?

tbarsballe:
It is picking up the workspace from the Wicket model, so it should update
the filter accordingly. The filter does work as expected when creating a
new Layer Group in a workspace, so the update does seem to be working.
However, I dod not test this explicitly - I will take a look.

Torben

tbarsballe:
I looked into it, and this is indeed a problem - it does not pick up on the
currently selected workspace if you change the selection.

Additionally, a fix is going to be nontrivial, since the state of the
workspace drop-down is only available client side prior to submitting the
form. One option would be to change the links that open the layer group
panel to submit links so that the model gets updated, although this would
likely need a bunch of other loging to avoid failing the various built-in
validation checks.

I'm going to close the PR and update the ticket with this info.

PR #1701, opened 2016-07-23
[GEOS-7272] OL3 Preview wraps tiled layers

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-7272

Either of this or https://github.com/geoserver/geoserver/pull/1702 is a viable fix
This is a hackier fix, but carries less risk of introducing problems with new projections, since it is not changing how we handle them.


aaime:
This approach also works but if the global attribute controls the wrapping, I'd say it should be verified in the code that a projection handler can be built from the current code, and it's a wrapping one (thus, adding an extra variable in the FTL context)

tbarsballe:
Updated to check the associated ProjectionHandler type, and only use `global: true` when we have a WrappingProjectionHandler.

PR #1702, opened 2016-07-23
[GEOS-7272] OL3 Preview wraps tiled layers, using OL3 internal srs

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-7272

Either of this or https://github.com/geoserver/geoserver/pull/1701 is a viable fix
This is the more correct approach, but carries a higher risk of causing problems with odd projections not supported by default in OL3.

I have tested with EPSG:4326 and EPSG:3031 (Antarctic Polar) with no issues.


aaime:
We do ship with a bunch of extra projections by default, and people can define their own... this should probably be verified, and if the projection is not known, fall back on the old approach

PR #1710, opened 2016-07-26
[GEOS-7644]: Make secure decorators supporting StructuredGridCoverage…

dromagnoli:
…2DReaders

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/decorators/DefaultSecureDataFactory.java
+++ b/src/main/src/main/java/org/geoserver/security/decorators/DefaultSecureDataFactory.java
@@ -5,13 +5,12 @@
  */
 package org.geoserver.security.decorators;
 
-import java.util.Iterator;
-
 import org.geoserver.platform.ExtensionPriority;
 import org.geoserver.security.AccessLevel;
 import org.geoserver.security.Response;
 import org.geoserver.security.WrapperPolicy;
 import org.geotools.coverage.grid.io.GridCoverage2DReader;
+import org.geotools.coverage.grid.io.StructuredGridCoverage2DReader;
 import org.geotools.coverage.grid.io.AbstractGridFormat;
 import org.geotools.data.DataAccess;
 import org.geotools.data.DataStore;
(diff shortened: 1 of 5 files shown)


src/main/src/main/java/org/geoserver/security/decorators/DefaultSecureDataFactory.java
```diff
@@ -114,14 +115,16 @@ public Object secure(Object object, WrapperPolicy policy) {
         } else if (FeatureIterator.class.isAssignableFrom(clazz)) {
             return new SecuredFeatureIterator((FeatureIterator) object); 
         }
-        
+
         // try coverage readers and formats
         if(GridCoverage2DReader.class.isAssignableFrom(clazz)) {
             return new SecuredGridCoverage2DReader((GridCoverage2DReader) object, policy);
+        } else if(StructuredGridCoverage2DReader.class.isAssignableFrom(clazz)) {
```
aaime: Given that:

{code}
public interface StructuredGridCoverage2DReader extends GridCoverage2DReader {
{code}

Should't you test the structured one before testing its super-interface?

src/main/src/main/java/org/geoserver/security/decorators/SecuredStructuredGridCoverage2DReader.java
```diff
@@ -0,0 +1,98 @@
...
+ */
+public class SecuredStructuredGridCoverage2DReader extends DecoratingStructuredGridCoverage2DReader {
+
+    /** Parameters used to control the {@link Crop} operation. */
+    private static final ParameterValueGroup cropParams;
+
+    /**
+     * Cached crop factory
+     */
+    private final static Crop coverageCropFactory = new Crop();
```
aaime: Is this crop factory still needed?

src/main/src/main/java/org/geoserver/security/decorators/DefaultSecureDataFactory.java
dromagnoli: Right, this remembers me why I put the structured in first place 2 weeks ago when doing the local patch. Fixing it.

aaime:
Looks good now, merging

PR #1711, opened 2016-07-26
Update dependencies to avoid upstream bugs.

smithkm:


aaime:
Hi @smithkm  I don't understand why the same dependency is added twice, with and without the classifier?

smithkm:
@aaime Do you mean the ones with scope provided?  The classifier technically makes the resulting artifact "something else" so transitive dependencies will still pull in the normal artifact.  Adding a dependency with scope provided says that's not needed.

aaime:
I'm asking because for json-lib we already have the classifier:

{code}
 <dependency> 
     <groupId>net.sf.json-lib</groupId> 
     <artifactId>json-lib</artifactId> 
     <version>2.2.3</version> 
     <classifier>jdk15</classifier>
   </dependency>
{code}

but I don't see the double dependency (classified and not classified) being used anywhere, we simply depend on it. Also see https://github.com/geoserver/geoserver/blob/master/src/wfs/pom.xml#L64 . 
Thus me wondering if also adding the non classified one is required.

smithkm:
The only way I'm touching the json-lib dependency is to remove the exclusion of an old version of it in  GWC since GWC no longer depends on json-lib at all.  I'm using classifiers to provide special patched builds of Bean Utils, and Restlet.

smithkm:
Resolved Andrea's question in voice conversation.  Classifiers cause the the artifact to be considered a different "thing" (test artifacts for instance) so a provided scoped version of the original dependency is included to indicate that the the classified one works in its place in order to handle any transitive dependencies.

src/ows/src/main/java/org/geoserver/ows/util/OwsUtils.java
```diff
@@ -206,19 +206,20 @@ public static Method getter(Class clazz, String property, Class type) {
...
+     * @param <T>
      *
      * @param object The java bean.
      * @param property The property to retreive.
      * @param type Teh type of the property to retreive.
      *
      * @return The property, or null if it could not be found..
      */
-    public static Object property(Object object, String property, Class type) {
+    public static <T> T property(Object object, String property, Class<T> type) {
```
smithkm: Should be safe.  Technically I suppose someone somewhere might be storing a class in a raw generic variable and then passing it in, but it seems extremely unlikely.  I can skip this change and cast the expressions using it instead when backporting if anyone thinks this would be a problem.

smithkm:
And that last bean utils removal seems to have fixed up the unit tests in Travis.  Anything else before merging?

aaime:
Well, let's merge the beast and see if CITE tests are ok with it

PR #1717, opened 2016-07-27
[GEOS-7661] Support NaN in DecimalConverter

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-7661

@aaime - this bug was introduced by #1667, could you take a quick look at my fix here?

Also note that after your change, NaN is getting output as `\ufffd` (literally �, the unicode replacement character) in the UI, in accordance with http://docs.oracle.com/javase/8/docs/api/java/text/DecimalFormat.html (See the section on **Special Values**).

This is technically correct programatically, but may be a bit confusing from a user perspective.

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/wicket/DecimalConverter.java
+++ b/src/web/core/src/main/java/org/geoserver/web/wicket/DecimalConverter.java
@@ -5,6 +5,7 @@
 package org.geoserver.web.wicket;
 
 import java.text.DecimalFormat;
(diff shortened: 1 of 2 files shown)


mtCarto:
Looks good to me, using the DecimalFormatSymbols is probably more robust than relying on the unicode characters. Should avoid any localization issues now.

aaime:
+1

tbarsballe:
Merged, backported to 2.9.x and 2.8.x.

PR #1725, opened 2016-07-29
[GEOS-7639] Set the root directory to search for wktlib files

mtCarto:
Ensures symbol.properties type files can be found from a wktlib reference in SLD, test case is pending. @tbarsballe could you review and make sure this is sane.


src/main/src/main/java/org/geoserver/config/GeoServerDataDirectory.java
```diff
@@ -1078,6 +1079,19 @@ else if ( l.getResource() instanceof WMSLayerInfo ) {
...
     }
     
+    /**
+     * Set the the root directory to look for WKTLib files as the styles directory
+     */
+    public void setWKTLibRoot() {
+        WKTMarkFactory wp = new WKTMarkFactory();
+        URL styleUrl = null;
+        try {
+            styleUrl = new URL("file:" + root().toString() + "/" + STYLE_DIR);
```
tbarsballe: This should use:
```
resourceLoader.get(STYLE_DIR).file().toURL();
```

You may want to add some checks (eg. file exists, etc) before calling `toURL()`.

PR #1728, opened 2016-07-29
[GEOS-7632] Fix WMS endpoint possibly returning cached tiles to unauthorized users. Backport to 2.9.x

smithkm:
Backport #1727


smithkm:
Wrong base branch

PR #1729, opened 2016-07-29
[GEOS-7632] Fix WMS endpoint possibly returning cached tiles to unauthorized users. Backport to 2.9.x

smithkm:
Backport #1727

### DIFF
--- a/src/gwc/src/main/java/org/geoserver/gwc/GWC.java
+++ b/src/gwc/src/main/java/org/geoserver/gwc/GWC.java
@@ -16,6 +16,7 @@ import static org.geowebcache.seed.GWCTask.TYPE.TRUNCATE;
 
 import java.io.IOException;
 import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


src/gwc/src/main/java/org/geoserver/gwc/GWC.java
```diff
@@ -672,6 +675,29 @@ public final ConveyorTile dispatch(final GetMapRequest request,
...
+            String srs = request.getRawKvp().get("SRS");
+            ReferencedEnvelope bbox = null;
+            try {
+                bbox = (ReferencedEnvelope) new BBoxKvpParser().parse(bboxstr);
+            } catch (Exception e) {
+                throw new RuntimeException("Invalid bbox for layer '" + layerName + "': " + bboxstr);
+            }
+            if (srs != null) {
+                try {
+                    bbox = new ReferencedEnvelope(bbox, CRS.decode(srs));
```
aaime: Hum... will this actually work for wms 1.3 where some crs are flipped, but keep on using the EPSG:XYZW syntax? The WMS request reader checks the version number and changes the srs syntax accordingly in these cases.

smithkm: GWC doesn't do the magic version dependant swapping for 1.3.0, which would seem to be a bug in its own right under integrated mode which advertises 1.3.0 but then delegates to GWC, but it means that this incorrect behaviour balances out the incorrect behaviour elsewhere and masks the areas we want masked.

I REALLY hate these axis order bugs.

PR #1737, opened 2016-08-03
Updated the appbundle jar URL to point to new repo

dvntucker:


dvntucker:
Closing because this is still not the correct URL. D'oh. My bad.

PR #1742, opened 2016-08-05
temporary upgrade to servlet 2.4 to allow me to run

jodygarnett:
It may be a temporary glitch in the oracle matrix, had to change to servlet 2.4 for debugging today.


jodygarnett:
@tbarsballe you tried this locally, is it still needed?

tbarsballe:
This seemed to only affect our office in victoria, and everything started working again on Monday (Aug 8) so, no this is not needed.

PR #1744, opened 2016-08-05
PR: Fixes (GEOS-7672) Layer publishing regression: need to select the…

afabiani:
… store and Published flag is false

https://osgeo-org.atlassian.net/browse/GEOS-7672

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
+++ b/src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
@@ -42,6 +42,7 @@ import javax.measure.unit.SI;
 import javax.measure.unit.Unit;
 
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.SerializationUtils;
 import org.eclipse.xsd.XSDElementDeclaration;
 import org.eclipse.xsd.XSDParticle;
 import org.eclipse.xsd.XSDSchema;
@@ -56,6 +57,7 @@ import org.geoserver.config.GeoServerDataDirectory;
 import org.geoserver.data.util.CoverageStoreUtils;
 import org.geoserver.data.util.CoverageUtils;
 import org.geoserver.feature.retype.RetypingFeatureSource;
+import org.geoserver.ows.util.OwsUtils;
 import org.geoserver.platform.GeoServerEnvironment;
 import org.geoserver.platform.GeoServerExtensions;
 import org.geoserver.platform.GeoServerResourceLoader;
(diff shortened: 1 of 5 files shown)


src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -2343,35 +2345,38 @@ private void flushState(ContentDataStore contentDataStore, String nativeName)
...
     public DataStoreInfo clone(final DataStoreInfo source, boolean allowEnvParametrization) {
-        DataStoreInfo target = catalog.getFactory().createDataStore();
-        target.setEnabled(source.isEnabled());
-        target.setName(source.getName());
-        target.setDescription(source.getDescription());
-        target.setWorkspace(source.getWorkspace());
-        target.setType(source.getType());
+        DataStoreInfo target;
+        try {
+            target = (DataStoreInfo) SerializationUtils.clone(source);
```
jodygarnett: I guess this is to get out of breaking the code each time a new field is added?

I am used to using copy constructors to hide individual fields being set ...

src/web/core/src/main/java/org/geoserver/web/data/store/CoverageStoreNewPage.java
```diff
@@ -82,7 +82,8 @@ protected void onSuccessfulSave(final CoverageStoreInfo info, final Catalog cata
         CoverageStoreInfo expandedStore = getCatalog().getFactory().createCoverageStore();
         try {
             clone(savedStore, expandedStore);
-            layerChooserPage = new NewLayerPage(expandedStore.getId());
+            // The ID is assigned by the catalog and therefore cannot be cloned
+            layerChooserPage = new NewLayerPage(savedStore.getId());
```
jodygarnett: This appears to be the only functional change

jodygarnett:
@afabiani this is a fix right, are we backporting to 2.9.x, 2.8.x

jodygarnett:
Merge when ready!

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
afabiani: Yep

src/web/core/src/main/java/org/geoserver/web/data/store/CoverageStoreNewPage.java
afabiani: Yep, also only the savedStore has the ID setted at this time.

PR #1746, opened 2016-08-05
[GEOS-7673] Whitelisting dyndimension classes for XStream serialisation

torse:


jodygarnett:
did you backport this @smithkm ?

PR #1749, opened 2016-08-06
[GEOS-7635] Fixed workspaced/global legend graphic path

goudine:
https://osgeo-org.atlassian.net/browse/GEOS-7635


jodygarnett:
One of these commits is by "admin" - I am not sure how you did that :)

src/main/src/main/java/org/geoserver/ows/StylePublisher.java
```diff
@@ -0,0 +1,123 @@
+/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2013 OpenPlans
```
jodygarnett: As a new file it is probably (c) 2016, although I think we have recently updated our header policy.

* http://docs.geoserver.org/latest/en/developer/policies/committing.html

src/main/src/main/java/org/geoserver/ows/StylePublisher.java
```diff
@@ -0,0 +1,123 @@
...
+
+        if ((reqPath.length() > 1) && reqPath.startsWith("/")) {
+            reqPath = reqPath.substring(1);
+        }
+
+        // sigh, in order to serve the file we have to open it 2 times
+        // 1) to determine its mime type
+        // 2) to determine its encoding and really serve it
+        // we can't coalish 1) because we don't have a way to give jmimemagic the bytes at the
+        // beginning of the file without disabling extension quick matching
```
jodygarnett: I do not see evidence of opening the file in the following code? Is this comment up to date?

src/main/src/main/java/org/geoserver/ows/StylePublisher.java
```diff
@@ -0,0 +1,123 @@
...
+                    throw new IllegalArgumentException("Workspace not found");
+                }
+                resource = data.getStyles(workspace, stylePath);
+                if (resource.getType() == Type.UNDEFINED) {
+                    // workspace style not found, try the global styles directory as a fallback
+                    resource = data.getStyles(stylePath);
+                }
+            } else {
+                // something went wrong :(
+                throw new IllegalStateException("Unexpected number of tokens in reqest path.");
```
jodygarnett: This error message is unlikely to help a normal user (who does not think in terms of tokens unless they are playing in an arcade)

src/main/src/test/java/org/geoserver/ows/StylePublisherTest.java
```diff
@@ -0,0 +1,145 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
```
jodygarnett: Please match project standards for the creation of new files :)

src/main/src/test/java/org/geoserver/ows/StylePublisherTest.java
```diff
@@ -0,0 +1,145 @@
...
+                "workspaces/cite/styles/override.png");
+    }
+
+    private MockHttpServletResponse request(String[] path, String modifiedSince) throws Exception {
+        MockHttpServletRequest request = new MockHttpServletRequest();
+        request.setContextPath("/geoserver");
+        request.setMethod("GET");
+        StringBuilder b = new StringBuilder("/geoserver");
+        for (int i = 0; i < path.length; i++) {
+            b.append('/').append(path[i]);
```
jodygarnett: If the path is only being used to construct a URL then you may be able to test with just a string?

src/main/src/test/java/org/geoserver/ows/StylePublisherTest.java
```diff
@@ -0,0 +1,145 @@
...
+        if (modifiedSince != null) {
+            request.addHeader("If-Modified-Since", modifiedSince);
+        }
+        MockHttpServletResponse response = new MockHttpServletResponse();
+        publisher.handleRequest(request, response);
+        return response;
+    }
+
+    @Test
+    public void testEncoding() throws Exception {
```
jodygarnett: this test appears to be testing the contents (via response.getContentAsByteArra(), rather then the "encoding" - consider renaming?

src/main/src/test/java/org/geoserver/ows/StylePublisherTest.java
```diff
@@ -0,0 +1,145 @@
...
+                response.getContentAsByteArray());
+
+        path = new String[] { "styles", "cite", "override.png" };
+        response = request(path, null);
+        assertEquals(Arrays.toString(path), 200, response.getStatus());
+
+        assertArrayEquals(
+                IOUtils.toByteArray(this.getClass().getClassLoader()
+                        .getResourceAsStream("org/geoserver/ows/grass_fill.png")),
+                response.getContentAsByteArray());
```
jodygarnett: really glad to see this work

src/main/src/test/java/org/geoserver/ows/StylePublisherTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+        // add style - global
+        resourceLoader.copyFromClassPath("org/geoserver/ows/smileyface.png",
+                "styles/smileyface.png");
+        paths.add(new String[] { "styles", "smileyface.png" });
+
+        // add style - workspaced
+        resourceLoader.copyFromClassPath("org/geoserver/ows/house.svg",
+                "workspaces/cite/styles/house.svg");
+        paths.add(new String[] { "styles", "cite", "house.svg" });
```
jodygarnett: while I can see house.svg being added to the test fixture, I am not sure it was used in any of the tests?

jodygarnett:
I am not sure why the above build has failed?

data/release/global.xml
```diff
@@ -37,7 +37,7 @@
     <queueType>UNBOUNDED</queueType>
     <imageIOCacheThreshold>10240</imageIOCacheThreshold>
   </coverageAccess>
-  <updateSequence>152</updateSequence>
+  <updateSequence>154</updateSequence>
```
jodygarnett: do not think we need to change the update sequence here/

data/release/styles/default_point.sld
```diff
@@ -1,38 +1,37 @@
...
-              <Size>6</Size>
-            </Graphic>
-          </PointSymbolizer>
-        </Rule>
-      </FeatureTypeStyle>
-    </UserStyle>
-  </NamedLayer>
-</StyledLayerDescriptor>
-
+<?xml version="1.0" encoding="UTF-8"?>
```
jodygarnett: Do not think we need to change the default_poing.sld

data/release/styles/point.xml
```diff
@@ -1,5 +1,15 @@
...
+  <format>sld</format>
+  <languageVersion>
+    <version>1.0.0</version>
+  </languageVersion>
   <filename>default_point.sld</filename>
+  <legend>
+    <width>32</width>
+    <height>32</height>
+    <format>image/png; charset=UTF-8</format>
+    <onlineResource>grass_fill.png</onlineResource>
```
jodygarnett: Okay I see what happened, you were testing locally and accidentally saved the change

src/main/src/main/java/org/geoserver/ows/StylePublisher.java
```diff
@@ -104,11 +86,9 @@ protected URL getUrl(HttpServletRequest request) throws IOException {
                     resource = data.getStyles(stylePath);
                 }
             } else {
-                // something went wrong :(
                 throw new IllegalStateException("Unexpected number of tokens in reqest path.");
```
jodygarnett: Please consider a more helpful error message:
````
throw new IllegalStateException("Restricted to global styles and workspace styles folder");
````

File: src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java [hunk 13]
```diff
@@ -1323,7 +1324,8 @@ protected void handleLegendURL(LayerInfo layer, LegendInfo legend,
                 attrs.addAttribute("", "xmlns:xlink", "xmlns:xlink", "", XLINK_NS);
                 attrs.addAttribute(XLINK_NS, "type", "xlink:type", "", "simple");
                 
-                String legendUrl = buildURL(request.getBaseUrl(), legend.getOnlineResource(), null, URLType.RESOURCE);
+                String legendUrl = buildURL(request.getBaseUrl(), 
+                		appendPath("styles", legend.getOnlineResource()), null, URLType.RESOURCE);
```
jodygarnett: Thanks, was looking for this

src/main/src/test/java/org/geoserver/ows/StylePublisherTest.java
goudine: @jodygarnett Yes while it is not yet implemented in a test case, it will be useful in future stages when we ensure that only the correct file types (ie. *.png, *.svg) are accepted (rather than *.xml for example).

PR #1751, opened 2016-08-09
Backport Dependency Updates and GEOS-7632

smithkm:


bencaradocdavies:
@smithkm Travis CI fails because ```org.springframework:spring-test``` is missing. I think it is available for Spring 3.

smithkm:
OK, fixed, I ran into another problem where the unit test failed because the response code was incorrect.  I don't really have time to fix that (There's probably a fix that didn't get backported somewhere) but I did change the unit test to ensure that even if the response code is wrong, the content is empty.  So even if there's another bug, 7632 is fixed.

bencaradocdavies:
Thanks, @smithkm. I see you also ported the challenge test to use mockrunner servlet response not spring-test.

smithkm:
Yes this was part of the spring update.  Spring required a new version of servlet api, which mockrunner didn't like so we had to switch to the spring test classes that did the same thing.

PR #1755, opened 2016-08-09
[GEOS-7467] Make JMS cluster plugin only serialize the necessary data (backport 2.9.x)

nmco:
Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-7467

JMS plugin doesn't serialize and send the whole store anymore. A test case was added for this.


jodygarnett:
If this is ready please merge

PR #1757, opened 2016-08-09
Remove unnecessary directory creation (legacy code from before resour…

NielsCharlier:
…ces)

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/GeoServerSecurityManager.java
+++ b/src/main/src/main/java/org/geoserver/security/GeoServerSecurityManager.java
@@ -2212,14 +2212,6 @@ public class GeoServerSecurityManager implements ApplicationContextAware,
         }
         
         LOGGER.info("Start security migration");
-        
-        //create required directories
-        get("security/role").dir();
-        get("security/usergroup").dir();
-        get("security/auth").dir();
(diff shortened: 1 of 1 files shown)


src/main/src/main/java/org/geoserver/security/GeoServerSecurityManager.java
```diff
@@ -2212,14 +2212,6 @@ boolean migrateFrom21() throws Exception{
...
         
         LOGGER.info("Start security migration");
-        
-        //create required directories
-        get("security/role").dir();
-        get("security/usergroup").dir();
-        get("security/auth").dir();
-        get("security/pwpolicy").dir();
-        get("security/filter").dir();
-        get("security/masterpw").dir();
```
jodygarnett: I understand not needing to force directory creation, not sure how to write, or if a test case is needed

NielsCharlier: This is the code that I should have been removed last year but I missed. The method does never use the file directories created there. There isn't really a way to test, because there is no expected behaviour from this. The fact that the tests for this method still pass should be enough...

jodygarnett:
Thanks Niels, do you want to backport this change?

PR #1758, opened 2016-08-09
[GEOS-7675] ImageMosaic documentation refactor

simboss:
Initial work for refactoring the ImageMosaic documentation. More will come soon.

PR #1759, opened 2016-08-09
Revert "[GEOS-7675] ImageMosaic documentation refactor"

jodygarnett:
Reverts geoserver/geoserver#1758


jodygarnett:
reading through it looks like this is still a work in progress, some sections marked "not true".

@bmmpxf please advise?

PR #1762, opened 2016-08-10
[GEOS-7676] Makes the importer share the request context when starting a job asynchronously

nmco:
Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-7676

This pull request also provides a test case for this.

### DIFF
--- a/src/extension/importer/rest/src/test/java/org/geoserver/importer/rest/RequestContextListener.java
+++ b/src/extension/importer/rest/src/test/java/org/geoserver/importer/rest/RequestContextListener.java
@@ -1,0 +1,45 @@
+/* (c) 2016 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.importer.rest;
+
+import org.geoserver.catalog.ResourceInfo;
+import org.geoserver.security.AbstractResourceAccessManager;
+import org.geoserver.security.DataAccessLimits;
+import org.springframework.security.core.Authentication;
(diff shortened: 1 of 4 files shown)


src/extension/importer/rest/src/test/java/org/geoserver/importer/rest/RequestContextListener.java
```diff
@@ -0,0 +1,45 @@
...
+import org.geoserver.security.AbstractResourceAccessManager;
+import org.geoserver.security.DataAccessLimits;
+import org.springframework.security.core.Authentication;
+import org.springframework.web.context.request.RequestContextHolder;
+import org.springframework.web.context.request.ServletRequestAttributes;
+
+import javax.servlet.http.HttpServletRequest;
+
+/**
+ * The propose of this class is to test that requests spring context are correctly shared
```
aaime: purpose

aaime:
Tiny typo fix needed, then good to go, merge at will. Mind you used java 8 syntax so if you need to backport the fix to 2.8.x, it will have to be amended

src/extension/importer/rest/src/test/java/org/geoserver/importer/rest/RequestContextListener.java
nmco: Well it was the only way I found to test that the spring context was share between several threads.

nmco: haha ok I got it sorry fixing the typo ...

nmco:
Type fixed, thanks for the review.

PR #1764, opened 2016-08-11
[GEOS-7443] fix for layergreoup json 

cippino:
Fix for issue [GEOS-7443](https://osgeo-org.atlassian.net/browse/GEOS-7443).


aaime:
Looking at the code it seems good, but there is no read part, so how are we going to handle POST requests from json clients that pick the json, alter it, and post it back (common approach). Is that going to work?

aaime:
Btw, the formatting is all off, it's using tabs... can you fix it? (only the parts that you have changed/added, not the whole file).

cippino:
Sorry for that I'll fix as soon as possible.

cippino:
The build of geoserver is broker 

Tests in error: 
  testWfsCompatitibility(org.geoserver.wfsng.WfsCompatibilityTest): Unknown type sf_archsites

I thought it was a fail of mine so I made a rebase, but the problem is related wfs, as you can read.

About the new attributes on style, You get right: there's no parsing action related, because I didn't take mind to that. If I think it is necessary, I'll implement it. The usercases will be: 

- default: true, explicit style: explicit style will be ignored 
- default: true, no style: set default (as actually happen)   
- default: false, explicit style: set explicit style (as actually happen) 
- default: false, no style: set default (as the second usecase)

Do you like?

aaime:
Superceded by #1852

PR #1766, opened 2016-08-12
[GEOS-7674] Fixup for GSIP-149 Improved GeoServer Style Page

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-7674
* Add apply button for saving style while remaining on page
  * For style new page, this adds the style and redirects to the edit page
  * Associations and Preview tabs have been removed from the style new page, as they are incapable of functioning before the style is added to the catalog. Once apply is clicked, the user is redirected to the edit page for the new style, which contains these tabs
  * Some extra fixup to avoid certain NPEs (althogh these should not be reachable due to the aboive point)
* Remove the CSS Styling page (Keep the Handler / Validator + templates for style generation).


tbarsballe:
@jodygarnett or other, could I get a sanity check here? I have done some manual testing (which led to a few rewrites of these fixes), but it is quite possible I missed something.

bencaradocdavies:
No review for me, just a PR title change because (1) GSIP-149 looks like a Jira tag but is not, and (2) I like works so I do not have to search for Jira tag / GSIP title when skimming a Git log.  :-)

src/web/wms/src/main/java/org/geoserver/wms/web/data/AbstractStylePage.html
```diff
@@ -14,9 +14,8 @@
     </form>
     <div class="button-group selfclear">
        <a href="#" wicket:id="validate"><wicket:message key="validate">Validate</wicket:message></a>
-       <a href="#" id="mainFormSubmit" onclick="document.forms['styleForm'].submit();">
-           <wicket:message key="submit">Submit</wicket:message>
-       </a>
+       <a href="#" wicket:id="apply"><wicket:message key="apply">Apply</wicket:message></a>
+       <a href="#" wicket:id="submit"><wicket:message key="submit">Submit</wicket:message></a>
        <a href="#" wicket:id="cancel"><wicket:message key="cancel">Cancel</wicket:message></a>
```
jodygarnett: A great usability improvement, I think the word "apply" works.

jodygarnett:
The commit bcd21c21aa4ece3790dddac3439b492bd92b5c68 looks clean, and checks off one of the tasks.

For the apply button, and tab NPE fixes .... I expect you will have a steady stream of fixes as you and @smithkm are working. Do you want to build them up in this single pull request or as a series of small commits?

tbarsballe:
Building them up here seems fine, I'm going to try and get as much fixed up before the M0 release today.
Everything remaining after that will be its own JIRA issue and will be tracked individually.

jodygarnett:
Would prefer to keep GEOS-7674 open to track commitments, such as docs.

tbarsballe:
I have no problem keeping it open, just want to make sure individual bugs are getting tracked so that we don't get intermingling discussions of different issues on one JIRA issue.

PR #1770, opened 2016-08-14
Get back some vertical space in the style editor

aaime:
Removes the style name inside the page "contents" and puts it in the title instead

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/css/geoserver.css
+++ b/src/web/core/src/main/java/org/geoserver/web/css/geoserver.css
@@ -1070,6 +1070,11 @@ height: 2.1em;
 margin: 1.5em 0 2.5em;
 }
 
+.tab-row-compact {
+margin: 1.5em 0 1em;
+}
+
+
 .tab-row ul {
 list-style-type: none;
 margin: 0;
--- a/src/web/wms/src/main/resources/GeoServerApplication.properties
+++ b/src/web/wms/src/main/resources/GeoServerApplication.properties
@@ -46,7 +46,7 @@ KMLLayerConfigPanel.kmlFormatSettings           = KML Format Settings
 
 StyleEditPage.description = Edit the current style. The editor can provide syntax highlighting and automatic formatting. Click on the "validate" button to verify the style is a valid SLD document.
 StyleEditPage.name        = Name
-StyleEditPage.title       = Style Editor
+StyleEditPage.title       = Style Editor - {0}
 StyleEditPage.notFound    = Could not find style "{0}"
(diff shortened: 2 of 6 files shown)


src/web/wms/src/main/resources/GeoServerApplication.properties
```diff
@@ -46,7 +46,7 @@ KMLLayerConfigPanel.kmlFormatSettings           = KML Format Settings
 
 StyleEditPage.description = Edit the current style. The editor can provide syntax highlighting and automatic formatting. Click on the "validate" button to verify the style is a valid SLD document.
 StyleEditPage.name        = Name
-StyleEditPage.title       = Style Editor
+StyleEditPage.title       = Style Editor - {0}
```
tbarsballe: This seems to be causing an issue with the page/tab title - the wildcard replacement does not seem to be working there:
screen shot 2016-08-15 at 9 03 10 am

(Works fine in the UI though)

src/web/core/src/main/java/org/geoserver/web/css/geoserver.css
```diff
@@ -1070,6 +1070,11 @@ height: 2.1em;
 margin: 1.5em 0 2.5em;
 }
 
+.tab-row-compact {
+margin: 1.5em 0 1em;
```
tbarsballe: Given the margin of the `page-header`, a style of `margin: -1em 0 1em;` might work better here (Or `margin: 0 0 1em;` at least) . Result:

screen shot 2016-08-15 at 9 18 25 am

tbarsballe:
I quite like this change - I wonder if we could/should do something similar for the Layer/Layer Group edit pages, as they do the same thing the style page before this change.

src/web/wms/src/main/resources/GeoServerApplication.properties
aaime: Ah, good catch, I did not notice. It seems there are two different template methods to override, getTitle and getPageTitle. I'll override both.

src/web/core/src/main/java/org/geoserver/web/css/geoserver.css
aaime: I'll give it a try

aaime:
I believe it sort of makes sense in general, except that the sections titles have a negative margin, so applying it to all pages might result some overlapping elements, like for example, in the "dimensions" tab of the layer editor.

tbarsballe:
Ok, so something we could consider, but will likely need to be tested relatively thoroughly, and may involve rewriting some of the geoserver CSS. Something we can keep in mind for the future at least.

aaime:
Page title fixed.

tbarsballe:
Looks good to me.

PR #1775, opened 2016-08-23
Compress (very) large png files in documentation

johanvdw:
The 3 files were taking up 4Mb, reduced to 89 Kb (98% reduction).

This is very similar to https://github.com/geoserver/geoserver/pull/1644


bradh:
The `doc/en/user/source/community/jms-cluster/activemq/images/MasterRestarted.png` change has introduced an artifact in the top part of Database1. They do generally look better except for that.

johanvdw:
Ok, thanks for having a look. I will see if I can fix that. It will be a new PR (not a commit on top of this one) (otherwise the unused image ends up in the bare repo anyway).

PR #1776, opened 2016-08-24
[GEOS-7698] Raster legend border colour now set by configuration

robward-scisys:
Enables the raster legend border colour to be taken into account when creating WMS legend.

Issue [https://osgeo-org.atlassian.net/browse/GEOS-7698](https://osgeo-org.atlassian.net/browse/GEOS-7698)

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/legendgraphic/RasterLayerLegendHelper.java
+++ b/src/wms/src/main/java/org/geoserver/wms/legendgraphic/RasterLayerLegendHelper.java
@@ -192,6 +192,9 @@ public class RasterLayerLegendHelper {
             cmapLegendBuilder.setBand(channelSelection != null ? channelSelection.getGrayChannel()
                     : null);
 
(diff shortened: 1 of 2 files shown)


tbarsballe:
Thanks for the fix. Could you please add a test case?

Also, have you signed a [contributor license agreement](http://docs.geoserver.org/latest/en/developer/policies/committing.html)? If not, could you please do so?

robward-scisys:
Added test case.

Corporate contributor agreement has been completed and sent to OSGeo.

aaime:
It actually does not work in a real GetLegendGraphic call because the legendoptions are provided as a CaseInsensitiveMap, which uppercases everything, and then the code checks for "borderColor" (which is a break in OGC service expectations, keys always have to be treated as case insensitive). 
I've changed the code to use another CaseInsensitiveMap internally, updated the tests so that they would break without my fix. Will merge manually.

aaime:
Merged on master and backported to 2.9.x

PR #1779, opened 2016-08-26
GeoServerResourceLoader: use resourcestore rather than basedirectory …

NielsCharlier:
…for url to file conversion

Motivation: We are running geoserver without a data directory at all, with an in-memory resource store.
Note that this change has no effect when using the default resource store (DataDirectoryResourceStore)

Testing: This method is already being tested extensively in two places: GeoServerDataDirectoryTest and PropertyDataStoreRelativeUrlTest.


NielsCharlier:
Sorry, ignore this, it will cause the whole dir to be cached. We need a different solution.

PR #1780, opened 2016-08-26
GeoServerResourceLoader: allow setting data directory without changin…

NielsCharlier:
…g resource store

Motivation: we are running a geoserver without a web context (so the data directory is never set in the setServletContext method as it is supposed to) and with its own resource store. We need to programmatically set the data directory without changing the resource store a filesystembased one. (Without a data directory set, the url() method fails).

Testing: this method is actually nowhere used in geoserver, there isn't really anything to test.

### DIFF
--- a/src/platform/src/test/java/org/geoserver/platform/GeoServerResourceLoaderTest.java
+++ b/src/platform/src/test/java/org/geoserver/platform/GeoServerResourceLoaderTest.java
@@ -5,15 +5,23 @@
 
 package org.geoserver.platform;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
 import java.io.File;
+import java.io.IOException;
 
 import javax.servlet.ServletContext;
 
 import org.easymock.EasyMock;
+import org.geoserver.platform.resource.FileSystemResourceStore;
+import org.geoserver.platform.resource.ResourceStore;
 import org.junit.Assert;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
+import org.junit.rules.TemporaryFolder;
 
 /**
  * Tests for {@link GeoServerResourceLoader}.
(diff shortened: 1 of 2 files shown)


aaime:
The method seems testable, even if not used, please add one test. It's for your own good at the end, since the method is not used today we might end up changing its behavior without considering that someone actually needs it to be coded that way. A failing test is always a good wake up call ;-)

NielsCharlier:
@aaime ok, done

src/platform/src/test/java/org/geoserver/platform/GeoServerResourceLoaderTest.java
```diff
@@ -124,5 +132,28 @@ public void testLookupRequireMissingFileServlet() {
...
         GeoServerResourceLoader.lookupGeoServerDataDirectory(context);
     }
+    
+    @Test
+    public void testSetBaseDirectory() throws IOException {
+        GeoServerResourceLoader loader = new GeoServerResourceLoader();
+        assertNull(loader.getBaseDirectory());
+        assertEquals(ResourceStore.EMPTY, loader.getResourceStore());
+        
+        TemporaryFolder tempFolder = new TemporaryFolder();
```
aaime: TemporaryFolder needs to be set as a field with a @Rule to be cleaned up at test end, or not?
http://junit.org/junit4/javadoc/4.12/org/junit/rules/TemporaryFolder.html

NielsCharlier: really?  I have never done that before.

aaime: That's at least my understanding, and how the documentation shows to use a temp folder

NielsCharlier: sorted

NielsCharlier:
@aaime ok to merge now?

PR #1783, opened 2016-08-27
Donating YSLD reference from Boundless Suite

bmmpxf:


bmmpxf:
@jodygarnett This is the first draft. We may wish to adjust this further.

jodygarnett:
Would like a review from @jdeolive :)

jodygarnett:
Thanks @bmmpxf - I noticed a couple pictures we will need to update (wrong geoserver logo used). Other than that these are excellent docs thanks you.

jdeolive:
These docs are awesome, great work Mike. Big +1 from me.

PR #1788, opened 2016-08-31
updated styling workshop

jodygarnett:
Updated workshop materials from foss4g 2016 from @smithkm and @tbarsballe


jodygarnett:
Please do not merge yet, would like a review from @bmmpxf first, and I need to double check page references.

tbarsballe:
I belive there are also a few bugs we found in the workshop that should be fixed here

doc/en/user/source/styling/workshop/ysld/linestring.rst
```diff
@@ -0,0 +1,671 @@
...
+* :ref:`LineString <styling/sld-reference/linesymbolizer.html>` (User Manual | SLD Reference )
+
+Line
+----
+
+A line symbolizer is represented by a :kbd:`line` key.  You can make a completely default symbolizer by giving it an empty map
+
+.. code-block:: yaml
+   
+        line: {}
```
tbarsballe: Remove `{}`, turn this into valid YSLD. This was confusing most workshop attendees.

doc/en/user/source/styling/workshop/ysld/polygon.rst
```diff
@@ -0,0 +1,1024 @@
...
+
+      symbolizers:
+      - polygon:
+          fill-graphic:
+            symbols:
+            - mark:
+                shape: square
+                fill-color: 'gray'
+                stroke-color: 'black'
+                stroke-width: 1
```
tbarsballe: Somewhere around here we need to also add
```
            stroke-color: 'black'
            stroke-width: 1
```
at a lower indent level. @mtCarto might remember better

doc/en/user/source/styling/workshop/ysld/ysld.rst
```diff
@@ -0,0 +1,339 @@
...
+
+Finished early? For now please help your neighbour so we can proceed with the workshop.
+
+If you are really stuck please consider the following challenge rather than skipping ahead.
+
+.. admonition:: Explore
+
+   #. Return to the :guilabel:`Data` tab and use the :guilabel:`Compute` link to determine the minimum and maximum for the **scalerank** attribute.
+   
+      .. only:: instructor
```
jodygarnett: Can we figure out a way to make these white on white text?

jodygarnett:
Thanks @tbarsballe, this is on a branch you have write access to :)

doc/en/user/source/styling/workshop/ysld/polygon.rst
jodygarnett: Adding it in for now, but this seems to be a regression occurring in a couple places with YSLD. 

@smithkm I expected SLD defaults to be respected (such as a black stroke) but that does not seem to be the case. Is that intensional?

smithkm: This is likely due to defaults being handled as part of the parsing rather than in the renderer.  I expected the renderer to fill in proper defaults and so parsed the YSLD exactly as it was, leaving things blank if they were blank.    It seems like the SLD parser was filling those defaults in itself instead.  

The short term easy solution would be to change the YSLD parser to fill in the same defaults.  I kind of prefer the idea of doing defaults at a more fundamental level and changing the SLD parser to work like the YSLD one.  Actually looking at how much work is involved or how much stuff might be broken might change that opinion.

PR #1790, opened 2016-09-01
vector-to-raster render transform avoids direct raster path

rfecher:
I'll send an email to the dev list, but in general the assumption is that a vector-to-raster rendering transform is still going to make good use of the bbox in the query that is added to the query object in the StreamingRenderer code path, but not in the directRasterRender() code path.  For example the geotools HeatmapProcess by default will not make use of bbox in the query because it is using  Query.ALL supplied by the direct raster rendering path.

I can file a contributor agreement assuming this seems reasonable.


aaime:
Needs a test, you can check and extend RenderedImageMapOutputFormatTest

rfecher:
yep makes sense, added a test and squashed it back down to a single commit

aaime:
Yes, looks good now, merging

aaime:
Btw, could you please create a ticket in jira and mark it as resolved, otherwise it won't show up in the release notes

rfecher:
sure, I created https://osgeo-org.atlassian.net/browse/GEOS-7735, but it doesn't seem to let me close or resolve the issue, change the assignee, or various other actions unless I'm missing something.  It makes complete sense to not open that open to everyone.

PR #1791, opened 2016-09-01
minor updates to reflect the newer namespace of the processes

rfecher:
it doesn't seem like this documentation reflects a change to the namespaces for the geotools processes that are referenced.  I thought it may be nice to update this minor detail.


aaime:
Yep, thanks

PR #1795, opened 2016-09-02
[GEOS-7712] Importer throws NullPointerException

emerkle826:
In some cases, importnig a raster causes a NullPointerException
  when trying to fetch the Namespace from the Catalog, using
  the Workspace name of the StoreInfo of the supplied ResourceInfo,
  because the StoreInfo is not set on the ResourceInfo.
  The ResourceInfo does seem to have a Namespace, so use that if
  we can't get it from the Store.

### DIFF
--- a/src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
+++ b/src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
@@ -1412,7 +1412,14 @@ public class Importer implements DisposableBean, ApplicationListener {
 
         //TODO: put an upper limit on how many times to try
         StoreInfo store = resource.getStore();
-        NamespaceInfo ns = catalog.getNamespaceByPrefix(store.getWorkspace().getName());
+        NamespaceInfo ns;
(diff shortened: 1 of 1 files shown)


aaime:
We do not normally accept pull requests without a test

emerkle826:
@aaime I'll try to write one, stand by.

emerkle826:
@aaime I'm looking at https://github.com/geoserver/geoserver/pull/1762 which addresses GEOS-7676. The issue in that ticket might have something to do with the issue in GEOS-7712. I'm going to try to pull that change in and test against the issues for GEOS-7676. I may not need this PR at all if 7676 fixes it for me.

aaime:
@emerkle826 not sure  @nmco pull request has relations with this one, #1762 is about passing down in the thread pool the Spring request context, I don't think it has effects on the ImportContext creation.

emerkle826:
No luck. The fix in #1762 does not resolve the issue with GEOS-7712. Still working on a good test for this.

emerkle826:
This still does not fix this issue. Closing until the correct fix can be identified....

PR #1799, opened 2016-09-05
[GEOS-7717] Malformed WMS GetFeatureInfo XML response if raster sampl…

bencaradocdavies:
…e dimension description not valid NCName

https://osgeo-org.atlassian.net/browse/GEOS-7717


aaime:
Seems like a reasonable change. At the same time I have an hesitation, because it's imposing a GML restriction to all output formats. Like, I supposed people would prefer to see the name they have configured in the UI in text,html and geojson outputs. Or not? 
An alternative approach could be to wrap the collection and force ncnames in the GML output formats instead.

bencaradocdavies:
@aaime in my view this is deeper than just an output format because a ```Feature``` is being constructed. As far as I know, all properties in the OGC General Feature Model are referred to by names that are pretty close to NCNames (perhaps not quite this strict until you get to the GML encoding). ```AttributeDescriptorImpl``` takes a constructor argument that is a ```Name``` (```NameImpl```) which is equivalent to a ```QName```. The current implementation of ```RasterLayerIdentifier``` is using ```SimpleFeatureTypeBuilder```. It looks to me like XML all the way down. Use of non-NCName identifiers in this context is quite surprising to me. I expected that any ```Feature``` can be encoded as GML without having to rewrite names. I thought that it was the responsibility of users of feature builders to use only NCNames.

ISO 19013 requirement 16 forbids the presence of whitespace in attribute names:
https://github.com/ISO-TC211/UML-Best-Practices/wiki/Naming-of-packages,-classes-and-attributes

```RasterLayerIdentifier``` permits whitespace in attribute names and in my view deviates from ISO 19103 and the intent of GeoAPI.

aaime:
Remember you are talking coverages here, not features, the translation to feature is an implementation accident. Check what are the limitations on band names in WCS, that's what the user is actually configuring.

aaime:
Checked WCS 2.0, it seems they are enforcing NCName indeed (weird, but then again, WCS 2.0 made odd choices, like not allowing : in the coverage names like previous implementation):

````
    <complexType name="DataRecordType">
        <complexContent>
            <extension base="swe:AbstractDataComponentType">
                <sequence>
                    <element maxOccurs="unbounded" minOccurs="1" name="field">
                        <annotation>
                            <documentation>Definition of the field provided as a nested data component. The field can be scalar or can itself be an aggregate such as a record, choice or array</documentation>
                        </annotation>
                        <complexType>
                            <complexContent>
                                <extension base="swe:AbstractDataComponentPropertyType">
                                    <attribute name="name" type="NCName" use="required"/>
                                </extension>
                            </complexContent>
                        </complexType>
                    </element>
                </sequence>
            </extension>
        </complexContent>
    </complexType>
````

So ok, no more objections here.

bencaradocdavies:
Thanks @aaime, that is even more complicated and interesting than I had imagined. I think your concerns are valid for WCS 1.0. Will forcing NCNames on plain text and HTML output annoy users? FTL should be able to turn ```_``` back into spaces.

If you are happy with this one then can I merge it?

aaime:
Eh, it's a change in behavior that will please on GML users, most of the people I know use GeoJSON or HTML as the preferred output format. Maybe discuss the two options on the devel list?

bencaradocdavies:
@aaime, I have started a thread on the geoserver-devel mailing list:
http://osgeo-org.1560.x6.nabble.com/Convert-raster-GetFeatureInfo-band-names-into-NCNames-td5284128.html

bencaradocdavies:
@aaime, should this be backported to 2.9.x?

bencaradocdavies:
@aaime, any objection to backporting this change to 2.9.x for inclusion in 2.9.2?

bencaradocdavies:
Backported to 2.9.x.

PR #1800, opened 2016-09-05
[GEOS-7719] Communtiy Module: OAuth2 filter and Google OAuth2 filter

afabiani:
JIRA

https://osgeo-org.atlassian.net/browse/GEOS-7719


simboss:
Open a Jira and reference it in the PR

src/community/pom.xml
```diff
@@ -462,5 +464,17 @@
         <module>geogig</module>
       </modules>
     </profile>
+    <profile>
+      <id>oauth2</id>
```
aaime: Why having a second profile? web/app/pom.xml only has oauth2-google in any case.

afabiani: Fixed!

Best Regards,
Alessio Fabiani.

==
GeoServer Professional Services from the experts!
Visit http://goo.gl/it488V for more information.
==

Ing. Alessio Fabiani
@alfa7691
Founder/Technical Lead

GeoSolutions S.A.S.
Via di Montramito 3/A
55054  Massarosa (LU)
Italy
phone: +39 0584 962313
fax:     +39 0584 1660272
mob:   +39 331 6233686

http://www.geo-solutions.it
http://twitter.com/geosolutions_it

-------------------------------------------------------

*AVVERTENZE AI SENSI DEL D.Lgs. 196/2003*

Le informazioni contenute in questo messaggio di posta elettronica e/o
nel/i file/s allegato/i sono da considerarsi strettamente riservate. Il
loro utilizzo è consentito esclusivamente al destinatario del messaggio,
per le finalità indicate nel messaggio stesso. Qualora riceviate questo
messaggio senza esserne il destinatario, Vi preghiamo cortesemente di
darcene notizia via e-mail e di procedere alla distruzione del messaggio
stesso, cancellandolo dal Vostro sistema. Conservare il messaggio stesso,
divulgarlo anche in parte, distribuirlo ad altri soggetti, copiarlo, od
utilizzarlo per finalità diverse, costituisce comportamento contrario ai
principi dettati dal D.Lgs. 196/2003.



The information in this message and/or attachments, is intended solely for
the attention and use of the named addressee(s) and may be confidential or
proprietary in nature or covered by the provisions of privacy act
(Legislative Decree June, 30 2003, no.196 - Italy's New Data Protection
Code).Any use not in accord with its purpose, any disclosure, reproduction,
copying, distribution, or either dissemination, either whole or partial, is
strictly forbidden except previous formal approval of the named
addressee(s). If you are not the intended recipient, please contact
immediately the sender by telephone, fax or e-mail and delete the
information in this message that has been received in error. The sender
does not give any warranty or accept liability as the content, accuracy or
completeness of sent messages and accepts no responsibility  for changes
made after they were sent or for other risks which arise as a result of
e-mail transmission, viruses, etc.

---------------------------------------------------------------------

doc/en/user/source/community/oauth2/index.rst
```diff
@@ -0,0 +1,151 @@
+.. _security_tutorials_oauth2:
+
+Authentication with OAuth2
+==========================
+
+This tutorial introduces GeoServer OAuth2 support and walks through the process of
+setting up authentication aganist an OAuth2 provider. It is recommended that the 
+:ref:`security_auth_chain` section be read before proceeding.
+
+OAuth2 Protocol and GeoServer OAuth2 core module
```
aaime: Most of this section is too technical, this is user documentation (I don't understand it myself, someone that has been exposed to the details of a auth protocol implementation probably would). It's still good material, but should not be the first thing a user reads. Maybe move it at the very end of the document, just telling users that they need to install the core module plus the specific binding one, and if they don't find the one they want, they can go and reference the new section at the end.

aaime:
Most (but not all) of the screenshots are heavily pixxelated and hard to read. It's like antialiasing was turned off, and then heavy jpeg compression was applied to the result. Not the end of the world for a community module anyways, but it makes the docs harder to go through.

doc/en/user/source/community/oauth2/index.rst
```diff
@@ -0,0 +1,151 @@
...
+#. Obtain OAuth 2.0 credentials from the Google API Console.
+
+   Visit the `Google API Console <https://console.developers.google.com/>`_ to obtain OAuth 2.0 credentials such as a client ID and client secret 
+   that are known to both Google and your application. The set of values varies based on what type of application you are building. 
+   For example, a JavaScript application does not require a secret, but a web server application does.
+   
+    .. figure:: images/google1.jpg
+       :align: center
+
+#. Create a new ``OAuth client ID``
```
aaime: Looks like there is a missing step here, when I get to the credentials page I'm offered to create a new project:

selezione_062

and then:

selezione_064

If the project already exists then I am offered to select it. Only after that I reach the screen at point 2.

doc/en/user/source/community/oauth2/index.rst
```diff
@@ -0,0 +1,151 @@
...
+
+   Click on the ``Client credentials`` context menu as shown in the figure below.
+   
+    .. figure:: images/google2.jpg
+       :align: center
+
+#. Configure a new ``Web application``
```
aaime: Before I can move on, I also have to go though this steps:

image

doc/en/user/source/community/oauth2/index.rst
```diff
@@ -0,0 +1,151 @@
...
+
+   Add the ``Name`` and the ``Authorized redirect URIs`` like shown here below.
+   
+    .. note:: Always add two entries for each URI. One without the ending ``/`` and another one with it.
+   
+    .. figure:: images/google3.jpg
```
aaime: localhost does not seem to be accepted, I cannot save the form unless I remove it. If a true internet address is to be given, it's best to warn users about it and maybe point to some dyndns solution for people interested to try from home.

selezione_067

doc/en/user/source/community/oauth2/index.rst
```diff
@@ -0,0 +1,151 @@
...
+       :align: center
+
+#. Take note of the ``Client ID`` and the ``Client Sectet``.
+
+   At the end of the procedure Google will show-up a small dialog box with the ``Client ID`` and the ``Client Sectet``.
+   Those info can be always accessed and updated from the `Google API Console <https://console.developers.google.com/>`_
+   
+    .. figure:: images/google4.jpg
```
aaime: I assume you already have invalidated those right? ;-) (since they can be read from the screenshot)

doc/en/user/source/community/oauth2/index.rst
```diff
@@ -0,0 +1,151 @@
...
+
+    .. figure:: images/filter2.jpg
+       :align: center
+
+#. Click the ``OAuth2`` link.
+
+    .. figure:: images/filter3.jpg
+       :align: center
+
+#. Fill in the fields of the settings form as follows:
```
aaime: No, please don't make people copy those links from a screenshot (actually, could the oauth plugin just provide defaults so that people do not have to type/copy them at all?)... is there any page documenting them, just in case they might be changing in the future. Anyways, what I had to type was:
* https://accounts.google.com/o/oauth2/token
* https://accounts.google.com/o/oauth2/auth
* http://localhost:8080/geoserver (not sure if this is going to work, see google not accepting it as a redirect url before)
* https://www.googleapis.com/oauth2/v1/tokeninfo
* https://accounts.google.com/logout

doc/en/user/source/community/oauth2/index.rst
```diff
@@ -0,0 +1,151 @@
...
+	
+	   #. ``Scopes`` should be::
+	       
+		   https://www.googleapis.com/auth/userinfo.email,https://www.googleapis.com/auth/userinfo.profile
+		
+	   #. ``Client ID`` and ``Client Secret`` are the ones Google provided
+	   
+	   #. Choose a ``Role Service`` able to recognize user emails as IDs. By default a connected user will have ``ROLE_USER`` role
+	
+#. Update the filter chains by adding the new OAuth2 filter.
```
aaime: Which filter chains? All of them?

doc/en/user/source/community/oauth2/index.rst
```diff
@@ -0,0 +1,151 @@
...
+	   #. Choose a ``Role Service`` able to recognize user emails as IDs. By default a connected user will have ``ROLE_USER`` role
+	
+#. Update the filter chains by adding the new OAuth2 filter. 
+
+   .. figure:: images/filter5.jpg
+      :align: center
+
+#. Select the OAuth2 Filter for each filter chain you want to protect with OAuth2.
+
+   .. figure:: images/filter6.jpg
```
aaime: Issue in screenshot, it shows in unreadable chars something looking like "oauth2", but I have "oauth2-google" instead. By adding it, it goes to the end of the list, I assume the right position is after rememberme, but before anonymous (not sure if before or after form, probably does not matter)

doc/en/user/source/community/oauth2/index.rst
```diff
@@ -0,0 +1,151 @@
...
+      :align: center
+
+   Be sure to select and order correctly the OAuth2 Filter.
+
+#. Save.
+
+Test a Google OAuth2 login
+--------------------------
+
+#. Navigate to the GeoServer home page and log out of the admin account.
```
aaime: Something went wrong here, and I cannot login anymore, google makes me login and then brings me to the account settings page instead of the geoserver page. My guess is that it's due to the redirect url. Some instructions to un-brick GeoServer would be useful.
In my case I went to modify GeoServer config at $GEOSERVER_DATA_DIR/release/security/filter/oauth2-google and set the redirect URI to one of the authorized ones, restarted and then I get automatically logged in.
However, GeoServer is still "bricked", as I can only login with my google account now, no way to get to the home page as anonymous, and no way to get back as admin using the local authentication. I believe this needs to be solved, or instructions on how to make sure admin control is not lost.

aaime: To get back admin control of GeoServer, one can go to $GEOSERVER_DATA_DIR//home/aaime/devel/gs_localdata/release/security and edit the config.xml, removing the oauth filter from the web chain.

doc/en/user/source/community/oauth2/index.rst
aaime: Indeed setting a non authorized redirect URL will break the login.

src/community/security/oauth2/src/main/java/org/geoserver/security/oauth2/GeoServerOAuthAuthenticationFilter.java
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    @Override
+    protected void doAuthenticate(HttpServletRequest request, HttpServletResponse response) {
+
+        String principal = null;
+        try {
+            principal = getPreAuthenticatedPrincipal(request, response);
+        } catch (IOException e1) {
+            // TODO Auto-generated catch block
```
aaime: Do something here, log the exception, rethrow it, but printing it to the std out/err is not a valid option.

src/community/security/oauth2/src/main/java/org/geoserver/security/oauth2/OAuth2SecurityConfiguration.java
```diff
@@ -0,0 +1,20 @@
...
+import org.springframework.security.oauth2.client.resource.OAuth2ProtectedResourceDetails;
+
+/**
+ * Interface for {@link GeoServerOAuth2SecurityConfiguration}
+ * 
+ * @author Alessio Fabiani, GeoSolutions S.A.S.
+ */
+public interface OAuth2SecurityConfiguration {
+
+    public OAuth2ProtectedResourceDetails geoServerOAuth2Resource();
```
aaime: Public methods without javadoc

doc/en/user/source/community/oauth2/index.rst
mprins: After authorization the auth service redirects back to this address/URL, since every IP host has it's own localhost it would never work

afabiani: Of course in a production system you need to configure an access key for every public host and specify the public ip.

simboss: Alessio, make sure this is properly documented

afabiani: about this specific issue... in this case Google does not accept the URI because you typed "locahost" instead of "localhost".

Google does checks on the address. It cannot be a public IP or a local domain different from localhost (like http://mydomain.local for instance).

PR #1809, opened 2016-09-07
[GEOS-7960] Ensure edits are not lost when switching tabs

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-7690

Uses a submit link to save the model status when switching tabs.

As with the [PublishedConfigurationPage](https://github.com/geoserver/geoserver/blob/master/src/web/core/src/main/java/org/geoserver/web/publish/PublishedConfigurationPage.java#L44) upon which this fix is modeled, there is no easy way of testing this behavior with the wicket tester.

### DIFF
--- a/src/web/wms/src/main/java/org/geoserver/wms/web/data/AbstractStylePage.java
+++ b/src/web/wms/src/main/java/org/geoserver/wms/web/data/AbstractStylePage.java
@@ -24,6 +24,7 @@ import org.apache.wicket.extensions.ajax.markup.html.tabs.AjaxTabbedPanel;
 import org.apache.wicket.extensions.markup.html.tabs.AbstractTab;
 import org.apache.wicket.extensions.markup.html.tabs.ITab;
 import org.apache.wicket.extensions.markup.html.tabs.PanelCachingTab;
+import org.apache.wicket.markup.html.WebMarkupContainer;
 import org.apache.wicket.markup.html.basic.Label;
 import org.apache.wicket.markup.html.form.Form;
 import org.apache.wicket.markup.html.link.Link;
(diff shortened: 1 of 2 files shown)


jodygarnett:
This important fix is held up on merge conflicts.

tbarsballe:
Merge conflicts resolved, waiting on travis CI.

src/web/wms/src/main/java/org/geoserver/wms/web/data/AbstractStylePage.java
```diff
@@ -45,14 +46,18 @@
...
-import org.geoserver.web.data.style.StyleDetachableModel;
 import org.geoserver.web.wicket.CodeMirrorEditor;
 import org.geoserver.web.wicket.GeoServerAjaxFormLink;
 import org.geoserver.web.wicket.ParamResourceModel;
 import org.xml.sax.SAXParseException;
 
 /**
  * Base page for creating/editing styles
+ * <p>
+ * WARNING: one crucial aspect of this page is its ability to not loose edits when one switches from
```
jodygarnett: Not sure that this needs to be in a javadoc :)

jodygarnett:
Tested locally, merged to master in anticipation of beta release.

src/web/wms/src/main/java/org/geoserver/wms/web/data/AbstractStylePage.java
tbarsballe: Thats fair, regular comment probably would have been fine. PublishedConfigurationPage does the same thing, so I guess this is more consistent?

PR #1810, opened 2016-09-08
[GEOS-7731] CoverageViewReader supporting AbstractGridFormat.BANDS parameter

mantesat:
[GEOS-7731](https://osgeo-org.atlassian.net/browse/GEOS-7731)

This change allows CoverageViewReader to support the BANDS reading parameter for using selected bands from a CoverageView. Changes  have been done also to RenderedImageMapOutputFormat to support rendering of CoverageView rasters using a symbolizer with selection channels. In this case, selection channels determine the BANDS parameter used by the CoverageViewReader to read bands. This closely follows changes in GeoTools described at [GEOT-5513](https://osgeo-org.atlassian.net/browse/GEOT-5513)

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/CoverageViewReader.java
+++ b/src/main/src/main/java/org/geoserver/catalog/CoverageViewReader.java
@@ -15,6 +15,7 @@ import java.awt.image.SampleModel;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -23,6 +24,7 @@ import java.util.Set;
 import javax.media.jai.ImageLayout;
 import javax.media.jai.RasterFactory;
 
+import org.apache.commons.lang.ArrayUtils;
 import org.geoserver.catalog.CoverageDimensionCustomizerReader.GridCoverageWrapper;
 import org.geoserver.catalog.CoverageView.CoverageBand;
 import org.geoserver.catalog.CoverageView.InputCoverageBand;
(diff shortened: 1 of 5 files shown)


src/main/src/main/java/org/geoserver/catalog/CoverageViewReader.java
```diff
@@ -284,35 +292,115 @@ public GridCoverage2D read(GeneralParameterValue[] parameters) throws IllegalArg
...
-                checker = new CoveragesConsistencyChecker(reader);
-            } else {
-                checker.checkConsistency(reader);
+        
+        int coverageBandsSize = bands.size();
+        
+        //Check params, populate band indices to read if BANDS param has been defined
+        ArrayList<Integer> selectedBandIndices = new ArrayList<Integer>();
+    	for (int m=0;m<coverageBandsSize;m++)
+    		selectedBandIndices.add(m);
```
aaime: There are tabs in the code, please follow the coding conventions: http://docs.geoserver.org/latest/en/developer/eclipse-guide/#code-formatting

src/main/src/main/java/org/geoserver/catalog/CoverageViewReader.java
```diff
@@ -378,14 +466,67 @@ public static GridCoverage2DReader wrap(GridCoverage2DReader reader, CoverageVie
...
+                HashMap<String, String> info = new HashMap<String, String>();
+
+                info.put("name", getName());
+                info.put("description", getDescription());
+                info.put("vendor", getVendor());
+                info.put("docURL", getDocURL());
+                info.put("version", getVersion());
+                
+                // reading parameters
+                return new ParameterGroup(
```
aaime: This read actually delegate reads to a wrapped format, so it should expose all wrapped read params, plus the band one, not just the band one.

src/wms/src/main/java/org/geoserver/wms/map/RenderedImageMapOutputFormat.java
```diff
@@ -919,6 +922,25 @@ private RenderedImage directRasterRender(WMSMapContent mapContent, int layerInde
             tileSizeY = mapContent.getMapHeight();
         }
         
+        // 
+        // Band selection
+        //
+        
+        int[] bandIndicesTmp = null;
```
aaime: This code is repeated in geotools as well. Could be factored out to a static method in the new style visitor for example.

mantesat:
Comments by aaime have been addresed on commit ee72f28

aaime:
Manually merged to add a build fix

PR #1812, opened 2016-09-08
[GEOS-7651] Use IOUtils rename method for FileSystemResource

tbarsballe:
Depends upon https://github.com/geoserver/geoserver/pull/1811; Travis build will fail until that PR is merged to master.

### DIFF
--- a/src/platform/src/main/java/org/geoserver/platform/resource/FileSystemResourceStore.java
+++ b/src/platform/src/main/java/org/geoserver/platform/resource/FileSystemResourceStore.java
@@ -5,6 +5,8 @@
  */
 package org.geoserver.platform.resource;
 
+import static org.geoserver.util.IOUtils.rename;
+
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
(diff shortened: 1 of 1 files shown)


aaime:
Uh, we have a compile failure here

aaime:
Aaah ok, depends on the other one

tbarsballe:
Now that the other one is merged, I have restarted the build.

src/restconfig/src/test/java/org/geoserver/rest/resources/ResourceTest.java
```diff
@@ -295,7 +295,6 @@ public void testMove() throws Exception {
     
     @Test
     public void testMoveDirectory() throws Exception {
-        put("/rest/resource/mydir/mynewdir?operation=move", "/mydir");
```
tbarsballe: As far as I can tell, this was intended to fail before, but works fine with the new Apache Commons FileUtils move/rename methods, causing the subsequent line to fail.

Hoping to get more details from @NielsCharlier

tbarsballe:
Got some feedback from @NielsCharlier on the mailing lists, now updated to explicitly disallow renaming a resource to a descendant of itself.

src/platform/src/main/java/org/geoserver/platform/resource/FileSystemResourceStore.java
```diff
@@ -476,15 +478,24 @@ public boolean delete() {
...
-                return Resources.renameByCopy(this, dest);
+            if (dest.parent().path().contains(path())) {
+                LOGGER.log(Level.FINE, "Cannot rename a resource to a descendant of itself");
+                return false;
             }
+            try {
+                if(dest instanceof FileSystemResource) {
+                    rename(file, ((FileSystemResource)dest).file);
+                } else if(dest instanceof Files.ResourceAdaptor) {
+                    rename(file, ((Files.ResourceAdaptor)dest).file);
```
jodygarnett: Interesting we have to special case the adaptor, I am not sure how often these are used/useful.

jodygarnett:
This looks good, all tests pass.

PR #1813, opened 2016-09-08
[GEOS-7651] Use IOUtils rename method for FileSystemResource (2.9.x)

tbarsballe:
Slightly hacky method that does not depend upon https://github.com/geoserver/geoserver/pull/1811. This should be sufficient to fix GEOS-7651 for 2.9.x and earlier in a way that does not require API changes.

### DIFF
--- a/src/platform/src/main/java/org/geoserver/platform/resource/FileSystemResourceStore.java
+++ b/src/platform/src/main/java/org/geoserver/platform/resource/FileSystemResourceStore.java
@@ -19,6 +19,7 @@ import java.util.UUID;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import org.apache.commons.io.FileUtils;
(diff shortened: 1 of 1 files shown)


src/restconfig/src/test/java/org/geoserver/rest/resources/ResourceTest.java
```diff
@@ -291,7 +291,6 @@ public void testMove() throws Exception {
     
     @Test
     public void testMoveDirectory() throws Exception {
-        put("/rest/resource/mydir/mynewdir?operation=move", "/mydir");
```
tbarsballe: As far as I can tell, this was intended to fail before, but works fine with the new Apache Commons FileUtils move/rename methods, causing the subsequent line to fail.

Hoping to get more details from @NielsCharlier

PR #1814, opened 2016-09-09
[GEOS-7734] Support coverage GetFeatureInfo wrapping and shifted nati…

bencaradocdavies:
…ve longitudes

https://osgeo-org.atlassian.net/browse/GEOS-7734


DBlasby:
Tested this and it's working (suite 4.9)

bencaradocdavies:
Thanks @DBlasby.

@aaime, you know about continuous map wrapping; do you have any objection to this PR?

aaime:
It bothers me that it's not a proper match for actual continuous map wrapping for projected CRS (e.g., google mercator), but only for geographic ones. However, to use the projection handlers the code would have to be expressed in terms of an envelope instead of a point. The handling of the 0-360 case is also not handled in the projection handler code, so it's not just a matter of using existing facilities instead of writing new code. 
So -0, but go ahead and merge.

bencaradocdavies:
Thanks, @aaime. I would also like to have supported wrapping for projections that have a periodic axis (including Mercator and Web Mercator), but as you note, it is not straightforward to reuse the existing code at this point. If I can figure out a robust way to detect such projections and the periodicity of the axis, I will submit a patch to support them.

I would also like to backport this change to 2.9.x for inclusion in 2.9.2. Test coverage is decent.

bencaradocdavies:
Backported to 2.9.x.

PR #1816, opened 2016-09-13
PR: Fix for JIRA GEOS-7738 Importer should preserve auth on async requests

afabiani:


src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -886,6 +889,7 @@ protected ImportContext call(ProgressMonitor monitor) throws Exception {
                     // set the parent request spring context, some interceptors like the security ones
                     // for example may need to have access to the original request attributes
                     RequestContextHolder.setRequestAttributes(parentRequestAttributes);
+                    SecurityContextHolder.getContext().setAuthentication(auth);
```
aaime: I don't see it being reset anywhere, are there risks to have cross-request pollution in that case? Maybe not, if this is the only way to get in and the new auth is always set, but I wanted to make sure. If you check the RequestContextHolder path, its being reset in the finally block just to make sure

afabiani: Fixed!

aaime:
Merged, backport as you see fit

PR #1817, opened 2016-09-13
PR - Fixes GEOS-7739 Importer tries to write features on closed connection

afabiani:


simboss:
test?

afabiani:
Testing this stuff automatically is not trivial. Tests using DB as target datastore on the Importer are already present, but this issue comes out only on certain cases.

In particular it can be reproduced by using as target a JNDI based store with a lot of free connections. After the transaction the pool physically closes the connections causing an exception on write, even if the schema has been correctly created on the DB.

aaime:
Yep, I cannot think of a reliable way to test it either. Merging, backport as you see fit.

PR #1819, opened 2016-09-13
[GEOS-7740] Make ResourceNotification serializable

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-7740


jodygarnett:
These are transitory objects used for notification - why do they need to be serializable?

tbarsballe:
See details about the problematic behavior in the linked ticket.

jodygarnett:
Niels has answered my question on the mailing list.

tbarsballe:
Trivial bugfix; backported to 2.9.x

PR #1822, opened 2016-09-16
[GEOS-7745] CSW based on internal layers ignores security restrictions

aaime:


jodygarnett:
I assume this can now be closed?

PR #1823, opened 2016-09-16
[GEOS-7745] CSW based on internal layers ignores security restrictions (facade wrapping version)

aaime:


src/main/src/main/java/org/geoserver/security/SecureCatalogFacade.java
```diff
@@ -0,0 +1,584 @@
...
+import org.geoserver.catalog.StyleInfo;
+import org.geoserver.catalog.WorkspaceInfo;
+import org.geoserver.catalog.event.CatalogListener;
+import org.geoserver.catalog.util.CloseableIterator;
+import org.geoserver.platform.GeoServerResourceLoader;
+import org.opengis.feature.type.Name;
+import org.opengis.filter.Filter;
+import org.opengis.filter.sort.SortBy;
+
+public class SecureCatalogFacade implements CatalogFacade {
```
jodygarnett: public class javadoc

src/main/src/main/java/org/geoserver/security/SecureCatalogFacade.java
```diff
@@ -0,0 +1,584 @@
+package org.geoserver.security;
```
jodygarnett: new file header

src/extension/csw/core/src/test/java/org/geoserver/csw/store/internal/ResourceAccessManagerCSWTest.java
```diff
@@ -0,0 +1,147 @@
...
+public class ResourceAccessManagerCSWTest extends CSWTestSupport {
+
+    static final Logger LOGGER = Logging.getLogger(ResourceAccessManagerCSWTest.class);
+
+    XpathEngine xpath;
+
+    // @Override
+    // protected String getLogConfiguration() {
+    // return "/GEOTOOLS_DEVELOPER_LOGGING.properties";
+    // }
```
jodygarnett: commented out code, but also test case

PR #1824, opened 2016-09-16
Fix URI building/parsing of PostgreSQL URIs for GeoGig

emerkle826:
- Address GeoGig issue: https://github.com/locationtech/geogig/pull/151
 - Uses Apache HttpClient URIBuilder to correclty encode special characters
   in the query parameter values for URIs


emerkle826:
@groldan I'm closing this for now as it might be too risky to use the Apache HttpComponent HttpClient library. There might be conflicts with the existing Apache Commons HttpClient library already being used by GeoServer.

PR #1825, opened 2016-09-17
[GEOS-7748] JdbcResourceStore: fix ignore directories

NielsCharlier:
Following up the discussion in https://github.com/geoserver/geoserver/pull/1821


tbarsballe:
While the fix itself looks good, I am concerned that this could exacerbate the problem identified in [GEOS-7747](https://osgeo-org.atlassian.net/browse/GEOS-7747). I would like to understand that issue a bit better before possibly making it worse.

NielsCharlier:
Since I identified GEOS-7747 as unrelated to jdbcstore, I think we should be fine

tbarsballe:
I agree, since GEOS-7747 is a core bug, I think this is fine to merge.

PR #1828, opened 2016-09-20
[GEOS-7749] Fixed ArrayIndexOutOfBounds bug on sample dimensions calculation.

mantesat:
This is a fix for [GEOS-7749](https://osgeo-org.atlassian.net/browse/GEOS-7749) bug, regarding an ArrayIndexOutOfBounds exception that occurs on CoverageViewReader at the procedure of the resulting coverage dimension wrapping when any of the source coverages of the view have less bands than the resulting coverage.

A small test is also included to prove the fix.


dromagnoli:
Thanks for the fix @mantesat .
Going to merge it.

PR #1831, opened 2016-09-20
WMTS multi dimensional fixes

nmco:
This pull request provide two commits that fix this issues (one commit for each issue):
- https://osgeo-org.atlassian.net/browse/GEOS-7722
- https://osgeo-org.atlassian.net/browse/GEOS-7721

All the changes were made in the WMTS multi dimensional community module.

The existing dimensions retrieval machinery assumes that we always want the unique values (no duplicates) and hence returns the domain values in a TreeSet. The GetHistogram operation needs to access the full domain values (including the duplicates ones).

For every DescribeOperation a bounding box of the used features will be used as spatial domain and properly encoded in the response. When is not possible to compute a bounding  box for the selected features (no geometry, non structured reader, etc ...) a default referenced envelope is returned.

Testes have been updated to reflect the new changes.


simboss:
I am merging as it looks good and this is a commmunity module.

PR #1832, opened 2016-09-20
GEOS-7741: while parsing style, not caching dependent resources

NielsCharlier:


tbarsballe:
This has a compile failure:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:2.3.2:testCompile (default-testCompile) on project gs-jdbcstore: Compilation failure: Compilation failure:
[ERROR] /home/travis/build/geoserver/geoserver/src/community/jdbcstore/src/test/java/org/geoserver/jdbcstore/AbstractJDBCResourceStoreTest.java:[21,27] error: cannot find symbol
[ERROR] package org.geoserver.config
[ERROR] /home/travis/build/geoserver/geoserver/src/community/jdbcstore/src/test/java/org/geoserver/jdbcstore/AbstractJDBCResourceStoreTest.java:[391,69] error: cannot find symbol
```

tbarsballe:
The actual fix looks good however

NielsCharlier:
A maven dependency issue that didn't occur in eclipse. Fixed now.

tbarsballe:
Okay, I just tested this out locally, and it mostly works.

I have found one rather seriouos bug in the fix. If you change the image in the resource store, the graphic will not get updated (even if you reload the configuration or restart geoserver).

As far as I can tell, resource.file() will not replace the image file if it already exists. I was able to get the new image to render by manually deleting the file from geoserver, AND reloading the Resource Cache and Configuratioin. This is obviously not a viable workaround each time you want to update a graphic.

NielsCharlier:
Wow, that doesn't sound right. resource.file() **will** replace a file if it already exists. That is what it is supposed to do, and I can't find any reason in the code why it wouldn't. (see SimpleResourceCache). Could you determine the reason that the file doesn't get updated?

tbarsballe:
Found the problem: https://github.com/geoserver/geoserver/blob/master/src/community/jdbcstore/src/main/java/org/geoserver/jdbcstore/cache/SimpleResourceCache.java#L73

SimpleResourceCache only updates the file if the file is NEWER than the resource. It should update the file if the file is OLDER than the resource.
The fix for this is realy easy, but I am getting hung up in writing a proper uit test - the JSBCStore test are quite sparse, and there are not a lot of good examples.

tbarsballe:
Here is my fix, with a test case: https://github.com/tbarsballe/geoserver/tree/jdbcstore
Could you please review the fix & pull this commit over to your branch to update this PR?

NielsCharlier:
Okay, reviewed, approved, and added.
Quite insane that this bug was not detected before.
Thanks for finding this.

tbarsballe:
Ok, this looks good, I will go ahead and merge.
Despite this being a bugfix, I am a bit hesitant to backport until after the 2.9.2 release, given that this includes core changes.

PR #1833, opened 2016-09-20
OS X App bundle improvements

tbarsballe:
* Add code-signing support for GeoServer.app bundle on OS X
* Migrate OS X DMG generation to dmgbuild utility; no GUI login required


tbarsballe:
Merged, and cherry-picked to 2.9.x

PR #1839, opened 2016-09-21
Follow up on GEOT-5514 (2.9.x branch)

aaime:
No known usable attack vectors as the WMS client code parsing caps does not report details about the parse fail, and cascaded GetFeatureInfo was alredy covered, but just to be safe...


aaime:
Depends on https://github.com/geotools/geotools/pull/1317

PR #1842, opened 2016-09-22
[GEOS-7755] [GEOS-7269] [GEOS-7268][GEOS-7758] Makes JMS plugin listen on all services and settings events and be aware of virtual services

nmco:
Makes JMS plugin listen on all service events (add, modify and remove) and also makes the plugin aware of virtual services. This pull request also makes the JMS plugin aware of settings events.

Tests for the reported issues were added.

The following issues reports different problems for the same root cause:
- https://osgeo-org.atlassian.net/browse/GEOS-7755
- https://osgeo-org.atlassian.net/browse/GEOS-7269
- https://osgeo-org.atlassian.net/browse/GEOS-7268
- https://osgeo-org.atlassian.net/browse/GEOS-7758

### DIFF
--- a/src/community/jms-cluster/jms-geoserver/src/main/java/org/geoserver/cluster/impl/events/configuration/JMSSettingsModifyEvent.java
+++ b/src/community/jms-cluster/jms-geoserver/src/main/java/org/geoserver/cluster/impl/events/configuration/JMSSettingsModifyEvent.java
@@ -1,0 +1,39 @@
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2013 OpenPlans
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.cluster.impl.events.configuration;
+
+import org.geoserver.cluster.impl.events.JMSModifyEvent;
+import org.geoserver.config.SettingsInfo;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * This class defines an event for modified settings. This settings represents metadata
+ * information for the GeoServer instance or for a specific workspace. By default an
+ * workspace doesn't have any settings and use GeoServer global settings
+ * <p>
+ * <p>A settings modified event can represent three situations, the settings were
+ * added, removed or the settings were modified.
+ */
+public class JMSSettingsModifyEvent extends JMSModifyEvent<SettingsInfo> {
+
+    private static final long serialVersionUID = 1L;
+
+    public JMSSettingsModifyEvent(SettingsInfo source, List<String> propertyNames,
+                                  List<Object> oldValues, List<Object> newValues) {
+        this(source, propertyNames, oldValues, newValues, JMSEventType.MODIFIED);
+    }
+
+    public JMSSettingsModifyEvent(SettingsInfo source, JMSEventType eventType) {
+        this(source, Collections.emptyList(), Collections.emptyList(), Collections.emptyList(), eventType);
+    }
+
+    public JMSSettingsModifyEvent(SettingsInfo source, List<String> propertyNames,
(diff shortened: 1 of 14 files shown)


src/community/jms-cluster/jms-geoserver/src/main/java/org/geoserver/cluster/impl/events/configuration/JMSSettingsModifyEvent.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    public JMSSettingsModifyEvent(SettingsInfo source, JMSEventType eventType) {
+        this(source, Collections.emptyList(), Collections.emptyList(), Collections.emptyList(), eventType);
+    }
+
+    public JMSSettingsModifyEvent(SettingsInfo source, List<String> propertyNames,
+                                  List<Object> oldValues, List<Object> newValues, JMSEventType eventType) {
+        super(source, propertyNames, oldValues, newValues);
+        this.eventType = eventType;
```
aaime: Does it make sense to have this for a class that states it's "modified"?

src/community/jms-cluster/jms-geoserver/src/main/java/org/geoserver/cluster/impl/events/configuration/JMSSettingsModifyEvent.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    public JMSEventType getEventType() {
+        return eventType;
```
aaime: Does it actually need to be extended?

src/community/jms-cluster/jms-geoserver/src/main/java/org/geoserver/cluster/impl/handlers/configuration/JMSServiceHandler.java
```diff
@@ -46,14 +51,29 @@ public boolean synchronize(JMSServiceModifyEvent ev) throws Exception {
...
+                case MODIFIED:
+                    // localize service
+                    final ServiceInfo localObject = localizeService(geoServer, ev);
+                    // save the localized object
+                    geoServer.save(localObject);
+                    break;
+                case ADDED:
+                    // checking that this service is not already present, we don't synchronize this check
+                    // if two threads add the same service well one of them will fail and throw an exception
+                    if (geoServer.getService(ev.getSource().getId(), ev.getSource().getClass()) == null) {
```
aaime: How can this happen?

src/community/jms-cluster/jms-geoserver/src/main/java/org/geoserver/cluster/impl/handlers/configuration/JMSSettingsHandler.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+        // save the updated settings
+        geoServer.save(settingsInfo);
+    }
+
+    private void handleAddedSettings(JMSSettingsModifyEvent event) {
+        // we only need to save the new settings, if the workspace associated
+        // with this settings doesn't exists or this settings already exists
+        // GeoServer will complain about it with a proper exception
+        geoServer.add(event.getSource());
```
aaime: Does this work even if the source object already has an id? Normally GeoServer won't save an object with an "id" as "new" but maybe for settings it's not caring. Just wondering

src/community/jms-cluster/jms-geoserver/src/main/java/org/geoserver/cluster/impl/events/configuration/JMSSettingsModifyEvent.java
nmco: Correct this method should go to the JMSModifyEventType.

aaime:
Ok, good to go for me then

PR #1848, opened 2016-09-23
[GEOS-7762] Validates the provided matrix tile set and updates bounding box CRS

nmco:
This pull request makes sure that the bounding box coordinate reference system is correctly updated and adds a test case for this. 

Associated issue:
https://osgeo-org.atlassian.net/browse/GEOS-7762


simboss:
Failure seems unrelated, merging...

PR #1849, opened 2016-09-26
[GEOS-7766] Adds parsed filter to parsed WMS POST request

koordo:
Adds the already parsed filter to the request parse result along side the layers and styles.
Added test to check the addition of filters.


aaime:
Thank you. Do you want to backport?
From the ticket description it might even sound like a regression, unless you think you had a customized geoserver version?

koordo:
Thanks for accepting!
No, we'll wait on the 2.10 release. Backporting would just delay upgrading to new version.

PR #1850, opened 2016-09-26
[GEOS-7767] Coverage views do not reproduce dynamically added alpha band (e.g., footprint management)

aaime:


dromagnoli:
Looks good to me. 
+1 on merge.

PR #1852, opened 2016-09-27
[GEOS-7771] Fix JSON encoding for a list of nulls

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-7771

This particular solution causes a null value to be returned as "null" when encoded as JSON.

For example `{styles: {style: ["null", "null", "null", "null", "null"]}}`

### DIFF
--- a/src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
+++ b/src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
@@ -24,6 +24,7 @@ import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import com.thoughtworks.xstream.io.json.JettisonStaxWriter;
 import org.apache.commons.collections.MultiHashMap;
 import org.geoserver.catalog.AttributeTypeInfo;
 import org.geoserver.catalog.AttributionInfo;
@@ -1155,8 +1156,9 @@ public class XStreamPersister {
                     String typeName = cam.serializedClass( theClass );
                     writer.addAttribute("type", typeName);
                 }
-                
                 context.convertAnother( item, new ReferenceConverter( clazz ) );
+            } else if (writer instanceof JettisonStaxWriter) {
+                writer.setValue("null");
             }
             writer.endNode();
         }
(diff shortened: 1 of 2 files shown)


tbarsballe:
Test failure seems to be:
```
testGetAsXML(org.geoserver.catalog.rest.LayerGroupTest)  Time elapsed: 7 sec  <<< ERROR!
java.lang.IllegalArgumentException: Layer group named 'citeLayerGroup' already exists
	at org.geoserver.catalog.impl.CatalogImpl.validate(CatalogImpl.java:823)
	at org.geoserver.catalog.impl.CatalogImpl.add(CatalogImpl.java:794)
	at org.geoserver.security.SecureCatalogImpl.add(SecureCatalogImpl.java:1109)
	at org.geoserver.catalog.impl.AbstractFilteredCatalog.add(AbstractFilteredCatalog.java:541)
	at org.geoserver.catalog.impl.AbstractCatalogDecorator.add(AbstractCatalogDecorator.java:438)
	at org.geoserver.catalog.impl.LocalWorkspaceCatalog.add(LocalWorkspaceCatalog.java:214)
	at org.geoserver.catalog.rest.LayerGroupTest.revertChanges(LayerGroupTest.java:73)
```

This is in the area that was changes, but _I cannot reproduce this test failure locally_ (and it seems like the sort of thing that should have failed before). It looks like a multithreading issue.

tbarsballe:
Got "Your test run exceeded 120 minutes" (but no other errors I could see) this time around. Restarting again...

tbarsballe:
Travis CI failed in the same place as before - `testGetAsXML(org.geoserver.catalog.rest.LayerGroupTest)`.

aaime:
Wait, isn't this one trying to address the same problem as #1764 ? Can you and @cippino synchronize on the issue and its solution?

cippino:
Yes, we found the same bug. If it works, it should be a better solution.  

If your test doesn't not works, you can take inspiration from mine

[#1764](https://github.com/geoserver/geoserver/pull/1764/commits/712969189a03d0937b7ad0bb0fb07659085cd690)

tbarsballe:
Looks like we did find the same bug. I've linked the two JIRA issues as duplicates.

I also think I figgured out the test failure - it looks like I had a bit of uncommited code locally :(
I've pushed the change, tests should pass now.

aaime:
Build is green now indeed, I'll merge this one and close #1764 
Should it be backported? Looks like a bug fix to me.

tbarsballe:
Yes, this is a bug fix. Now that this has been merged, I have backported to 2.9.x / 2.10.x.

PR #1858, opened 2016-09-30
PR - Fix for JIRA [GEOS-7776] GEOS-7753 fix that makes it impossible to create any WMS Cascaded store

afabiani:


tbarsballe:
Thanks for the fix. Works on my end. I am going to go ahead an backport this to 2.9.x, as it affects that as well.

PR #1863, opened 2016-10-01
[GEOS-7744] Makes layers data by default only editable by admin users (backport 2.8.x)

nmco:
This pull request changes the default configuration so that by default  layers data can only be edited by admin users. Documentation was also updated.

Mailing list discussion: http://osgeo-org.1560.x6.nabble.com/Who-changed-my-data-or-is-it-sane-to-have-WFS-T-open-to-all-by-default-td5286288.html

Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-7744

Backport of pull request: https://github.com/geoserver/geoserver/pull/1835


aaime:
The failure may be unrelated, and 2.8.x has not been built in a while, but I cannot assure it's a fluke. Nuno, does this build for you locally?

nmco:
The failure is provoked by my commit (the error only doesn't happen always but is my fault). I'm taking care of it.

PR #1864, opened 2016-10-02
[GEOS-7780] Add WCS scale factor XML demo

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOS-7780

This pull request targets master.

### DIFF
--- a/data/release/demo/WCS_getCoverage_2.0.1_scale_factor.xml
+++ b/data/release/demo/WCS_getCoverage_2.0.1_scale_factor.xml
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Demonstrates use of WCS scale factor -->
+<wcs:GetCoverage service="WCS" version="2.0.1"
+    xmlns:wcs="http://www.opengis.net/wcs/2.0" xmlns:scal="http://www.opengis.net/WCS_service-extension_scaling/1.0"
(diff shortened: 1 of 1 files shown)


data/release/demo/WCS_getCoverage_2.0.1_scale_factor.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Demonstrates use of WCS scale factor -->
+<wcs:GetCoverage service="WCS" version="2.0.1"
+    xmlns:wcs="http://www.opengis.net/wcs/2.0" xmlns:scal="http://www.opengis.net/WCS_service-extension_scaling/1.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.opengis.net/wcs/2.0 http://schemas.opengis.net/wcs/2.0/wcsAll.xsd">
+    <!--
```
aaime: Hmm... will we remember to remove the comment once the other issue is fixed? Does it actually need to be here?

bencaradocdavies:
This comment is needed so that, when users cut and paste, they can uncomment the full {{schemaLocation}} to enable schema validation.

bencaradocdavies:
Backported to 2.9.x.

PR #1866, opened 2016-10-03
Re-set entityResolverPrivider on configuration reload. Pass entityResolver to WMSStore

tbarsballe:
Add GeoServerReinitializer marker interface for GeoServerInitializers that should be run on both inital load and reload.
Updated test case, including improved cause verification based on GEOT-5533.

### DIFF
--- a/src/main/src/main/java/org/geoserver/config/GeoServerLoader.java
+++ b/src/main/src/main/java/org/geoserver/config/GeoServerLoader.java
@@ -152,6 +152,19 @@ public abstract class GeoServerLoader {
         }
     }
     
+    protected void reloadInitializers(GeoServer geoServer) throws Exception {
+        //reload applicable initializer extensions
+        List<GeoServerReinitializer> initializers = GeoServerExtensions.extensions( GeoServerReinitializer.class );
+        for ( GeoServerReinitializer initer : initializers ) {
+            try {
+                initer.reinitialize( geoServer );
+            }
+            catch( Throwable t ) {
+                LOGGER.log(Level.SEVERE, "Failed to run initializer " + initer, t);
+            }
+        }
+    }
+    
     /**
      * Does some post processing on the catalog to ensure that the "well-known" styles
      * are always around.
@@ -206,6 +219,8 @@ public abstract class GeoServerLoader {
         
         loadCatalog( catalog, xp );
         loadGeoServer( geoserver, xp);
+        
+        reloadInitializers(geoserver);
     }
 
     protected void readCatalog(Catalog catalog, XStreamPersister xp) throws Exception {
--- a/src/main/src/main/java/org/geoserver/config/GeoServerReinitializer.java
+++ b/src/main/src/main/java/org/geoserver/config/GeoServerReinitializer.java
@@ -1,0 +1,25 @@
+/* (c) 2016 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.config;
+
+
+/**
+ * Extension point interface for initializing based on configuration.
+ * Intended to be run both upon configuration initialization, and upon reload
+ *
+ */
+public interface GeoServerReinitializer extends GeoServerInitializer { 
+
+    /**
+     * Performs initialization  of GeoServer configuration, as well as any actions that should be
+     * performed only when reloading the configuration.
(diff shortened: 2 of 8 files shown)


src/main/src/main/java/org/geoserver/config/GeoServerReinitializer.java
```diff
@@ -0,0 +1,13 @@
...
+ */
+package org.geoserver.config;
+
+
+/**
+ * Extension point interface for initializing based on configuration.
+ * Intended to be run both upon configuration initialization, and upon reload
+ *
+ */
+public interface GeoServerReinitializer extends GeoServerInitializer { }
```
jodygarnett: Could I suggest adding a {{reinit}} method here, you can use a default implementation that just calls {{init}}. That way you have the flexibility to perform a different action on reload.

src/web/core/src/test/java/org/geoserver/web/data/store/WMSStoreNewPageTest.java
```diff
@@ -99,4 +102,47 @@ public void testSaveNewStore() {
 
         catalog.validate(expandedStore, false).throwIfInvalid();
     }
+    
+    @Test
+    public void testSaveNewStoreEntityExpansion() throws Exception {
+
+        WMSStoreNewPage page = startPage();
+        // print(page, true, true);
```
jodygarnett: kill me

tbarsballe: Good catch. Serves me right for blindly copy-pasting (see line 74 above)

src/main/src/main/java/org/geoserver/config/GeoServerReinitializer.java
tbarsballe: That seems sensible

tbarsballe:
Alright, I've pushed up changes based on your feedback @jodygarnett

jodygarnett:
Reviewing now, how do you feel about testing this on master before consider a backport.

src/main/src/main/java/org/geoserver/config/GeoServerLoader.java
```diff
@@ -206,6 +219,8 @@ public void reload() throws Exception {
         
         loadCatalog( catalog, xp );
         loadGeoServer( geoserver, xp);
+        
+        reloadInitializers(geoserver);
```
jodygarnett: Can I confirm the order of these methods?

src/main/src/main/java/org/geoserver/config/GeoServerReinitializer.java
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Extension point interface for initializing based on configuration.
+ * Intended to be run both upon configuration initialization, and upon reload
+ *
+ */
+public interface GeoServerReinitializer extends GeoServerInitializer { 
+
+    default void reinitialize( GeoServer geoServer ) throws Exception {
+        initialize(geoServer);
+    };
```
jodygarnett: Thanks, that is what I had in mind, can we get a javadoc (since this is a public method).

tbarsballe: Javadoc added (also added one for GeoServerInitializer.initialize)

src/main/src/main/java/org/geoserver/config/GeoServerLoader.java
tbarsballe: Order is correct, same as is done when initially loading the GeoServer configurion. I origionally tried with reloadInitializers first, but it turns out loadCatalog is responsible for creating a new ResourcePool() object.

jodygarnett: That makes me a little uncomfortable. Is it worth clearing the ResourcePool after the initializers run? Or making the ResourcePoolInitalizer clear the ResourcePool datastore cache when it is done.

tbarsballe: That is a good question. I don't have a good answer - what do you think?

jodygarnett: discuss on email, this pull requests accomplishes what you set out to do.

tbarsballe:
Backported to 2.9.x

PR #1867, opened 2016-10-04
[GEOS-7782] Add INSPIRE grid set when INSPIRE extensions is loaded

nmco:
This pull request added a new method on GWC singleton to allow GeoServer extensions to register non editable grid sets.

Associated issue:
https://osgeo-org.atlassian.net/browse/GEOS-7782

### DIFF
--- a/src/extension/inspire/src/main/java/org/geoserver/inspire/wmts/InspireGridSetLoader.java
+++ b/src/extension/inspire/src/main/java/org/geoserver/inspire/wmts/InspireGridSetLoader.java
@@ -1,0 +1,55 @@
+/* (c) 2016 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.inspire.wmts;
+
+import org.geoserver.gwc.GWC;
+import org.geoserver.platform.ContextLoadedEvent;
+import org.geowebcache.grid.BoundingBox;
+import org.geowebcache.grid.GridSet;
+import org.geowebcache.grid.GridSetFactory;
+import org.geowebcache.grid.SRS;
+import org.springframework.context.ApplicationListener;
+
+/**
+ * Loads the inspire grid set and mark it as non editable by the user.
+ */
+public class InspireGridSetLoader implements ApplicationListener<ContextLoadedEvent> {
+
+    public static final String INSPIRE_GRID_SET_NAME = "InspireCRS84Quad";
+
+    @Override
+    public synchronized void onApplicationEvent(ContextLoadedEvent event) {
+        GWC gwc = GWC.get();
+        // this grid set should not be editable by the user
+        gwc.addEmbeddedGridSet(INSPIRE_GRID_SET_NAME);
+        GridSet gridSet = gwc.getGridSetBroker().get(INSPIRE_GRID_SET_NAME);
+        if (gridSet != null) {
+            // this grid set already exists
+            return;
+        }
+        // the grid set resolutions
+        double[] resolutions = new double[]{
+                0.703125, 0.3515625, 0.17578125, 0.087890625, 0.0439453125, 0.02197265625, 0.010986328125, 0.0054931640625,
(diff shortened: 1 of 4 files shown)


src/gwc/src/main/java/org/geoserver/gwc/GWC.java
```diff
@@ -223,6 +223,9 @@
     private GeoWebCacheEnvironment gwcEnvironment;
     
     final GeoServerEnvironment gsEnvironment = GeoServerExtensions.bean(GeoServerEnvironment.class);
+
+    // list of GeoServer contributed grid sets that should not be editable by the user
+    private final List<String> geoserverEmbeddedGridSets = new ArrayList<>();
```
aaime: This should be a Set, or not?

src/extension/inspire/src/test/java/org/geoserver/inspire/wmts/InspireGridSetLoaderTest.java
```diff
@@ -0,0 +1,35 @@
...
+ * Testing that INSPIRE grid set is correctly loaded and configured.
+ */
+public class InspireGridSetLoaderTest extends GeoServerSystemTestSupport {
+
+    @Test
+    public void testGridSetLoading() throws Exception {
+        // let's wait a max of 30 seconds for the grid set registration
+        GWC gwc = GWC.get();
+        int waited = 0;
+        while(waited <= 30000 && GWC.get().getGridSetBroker().get(INSPIRE_GRID_SET_NAME) == null) {
```
aaime: Confused, why do we have to wait, are the context loaded notifications happening in a background thread? By the time the method is run the full initialization should have happened already or not?

src/extension/inspire/src/main/java/org/geoserver/inspire/wmts/InspireGridSetLoader.java
```diff
@@ -0,0 +1,55 @@
...
+import org.geowebcache.grid.BoundingBox;
+import org.geowebcache.grid.GridSet;
+import org.geowebcache.grid.GridSetFactory;
+import org.geowebcache.grid.SRS;
+import org.springframework.context.ApplicationListener;
+
+/**
+ * Loads the inspire grid set and mark it as non editable by the user.
+ */
+public class InspireGridSetLoader implements ApplicationListener<ContextLoadedEvent> {
```
aaime: What if we are reloading the data directory, and maybe the data directory is coming from a node that does not have the inspire extensions... how does this play out? Will we lose the gridset definition? Not a big deal, as it means the data dir being reloaded does not contain any cached layer with that gridset to start with I guess.

src/gwc/src/main/java/org/geoserver/gwc/GWC.java
nmco: Yes you are correct a set will be better here, updating it ...

src/extension/inspire/src/test/java/org/geoserver/inspire/wmts/InspireGridSetLoaderTest.java
nmco: Agrh ... correct  and I forgot to remove this form the tests ... (I tested it putting a thread sleep in a listener which blocked spring initiation which means that listeners are not executed on a background thread ... if it was not the case this solution would not have make sense)

src/extension/inspire/src/main/java/org/geoserver/inspire/wmts/InspireGridSetLoader.java
nmco: If the grid set is not present it will be added and consequently stored in ``geowebcache.xml``. If a data dircetory coming from a GeoServer that has the INSPIRE extension enabled is used in a a GeoServer that doesn't have the INSPIRE extension enabled nothing will be done to the grid set but the user will be able to edit it or remove it from GeoServer UI.

simboss:
Apparently concerns have been addressed, I will wait a little then i will merge.

PR #1871, opened 2016-10-05
[GEOS-7468] Change JMS default virtual topic default name to geoserver

nmco:
Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-7468


simboss:
I would update the doc with also some information on how to use different topics for different instances in GeoServer (mention a use case for example).

nmco:
Check this PR https://github.com/geoserver/geoserver/pull/2293.

PR #1877, opened 2016-10-06
Rename NoExternalEntityResolver to PreventLocalEntityResolver

jodygarnett:
Depends on https://github.com/geotools/geotools/pull/1340


DBlasby:
Looks good - just renaming the entity resolver + minor test case changes (compare vs indexOf).

PR #1883, opened 2016-10-14
license moved into root

teeschke:
License information belongs into the project's root. As an extra Github is able to detect the license model of the project automagically.


teeschke:
Dear Devs,
can I ask you to put the "hacktoberfest" label onto the PR?
Thanks and best regards!

aaime:
Hum hum... I have a doubt about this one, the root contains also the documentation, but the docs are not under GPL

teeschke:
What type of license does the docs have? Right now doc are in "default" license mode as well as the build and data folders. It looks like it's not allowed to reproduce, distribute, or derivative them.
https://help.github.com/articles/open-source-licensing/#what-happens-if-i-dont-choose-a-license

Maybe we can create an extra license file inside the docs folder.

aaime:
So the documentation is licenced as https://creativecommons.org/licenses/by/3.0/
The data directories, I don't believe they have one.... we might want to go for http://opendatacommons.org/licenses/pddl/summary/ as they are meant to be re-used freely. I'll start a discussion on the devel list about this.

PR #1884, opened 2016-10-14
formating changes - ran Geoserver code formatter + removed commented out code

DBlasby:
for GSIP-138


aaime:
Hi David, for this kind of changes against a community module you should be able to just go with a direct commit.

PR #1885, opened 2016-10-17
[GEOS-7792] WFS ignores disabling of stores (master)

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOS-7792

This PR implements both the simple and ```getSecurityFilter``` changes proposed by @aaime. Note that the ```getSecurityFilter``` change does not fix the bug by itself. I would also like some feedback on the ```shouldApplyFilter``` logic; see the [Jira comment](https://osgeo-org.atlassian.net/browse/GEOS-7792?focusedCommentId=60278&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-60278) for details. This PR leaves the ```shouldApplyFilter``` logic as-is. I am happy to change it based on feedback.

This PR also fixes WFS 2.0 ```GetFeatureJoinTest``` and ```GetFeaturePagingTest```, which were running with disabled stores.

### DIFF
--- a/src/wfs/src/test/java/org/geoserver/wfs/GetFeatureTest.java
+++ b/src/wfs/src/test/java/org/geoserver/wfs/GetFeatureTest.java
@@ -21,8 +21,10 @@ import org.custommonkey.xmlunit.XMLAssert;
 import org.custommonkey.xmlunit.XMLUnit;
 import org.custommonkey.xmlunit.XpathEngine;
 import org.geoserver.catalog.Catalog;
+import org.geoserver.catalog.DataStoreInfo;
 import org.geoserver.catalog.FeatureTypeInfo;
 import org.geoserver.catalog.ResourceInfo;
+import org.geoserver.catalog.StoreInfo;
 import org.geoserver.config.GeoServer;
 import org.geoserver.data.test.MockData;
 import org.geoserver.data.test.SystemTestData;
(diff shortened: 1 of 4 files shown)


aaime:
getSecurityFilter is the primary filter, it's more of an optimization for the jdbc store, the in memory one has actually no need for it. The other changes are the secondary filter instead. That said, I would have expected it to fix the issue, but oh well :-)

src/wfs/src/test/java/org/geoserver/wfs/GetFeatureTest.java
```diff
@@ -375,7 +377,25 @@ public void testRequestDisabledResource() throws Exception {
...
-    
+
+    /**
+     * Test that a request for a resource from a disabled store fails.
+     */
+    @Test
+    public void testRequestDisabledStore() throws Exception {
+        Catalog catalog = getCatalog();
+        StoreInfo store = catalog.getStoreByName("cdf", DataStoreInfo.class);
+        store.setEnabled(false);
```
aaime: So the store does not get re-enabled and the other tests pass anyways? Stroke of luck and it's one of the last tests to be run?

bencaradocdavies: @aaime no luck at all: there is already a ```@Before resetFifteen()``` above in this file to ensure that every test gets a working ```cdf:Fifteen```. The new test method is not the first to modify the store. Seeing the ```@Before``` method, I knew that fixture clean up was not required.  :-)

bencaradocdavies: @aaime I have force-pushed an updated version that re-enables the store in a ```finally``` block to restore the fixture state. Our tests should not rely on luck!

bencaradocdavies:
@aaime the store should now be re-enabled to restore the fixture state. I would also like clarification on the ```shouldApplyFilter``` logic.

One you are happy with this one, I would like to backport to 16.x and 15.x.

aaime:
The shouldApply logic makes sure the filter hides disabled resources for OGC request, but allows REST and UI admin requests to see them

bencaradocdavies:
Yes @aaime, but why is is applied only to layers and not also to resources? You wrote:
```
if (shouldApplyFilter() && LayerInfo.class.isAssignableFrom(clazz) || ResourceInfo.class.isAssignableFrom(clazz))
```
but did you mean
```
if (shouldApplyFilter() && (LayerInfo.class.isAssignableFrom(clazz) || ResourceInfo.class.isAssignableFrom(clazz)))
```
(note extra brackets)? The former seemed strange to me. I have preserved this logic.

aaime:
I believe that was a typo that needs fixing

bencaradocdavies:
Thanks @aaime. I have changed the logic to apply the ```shouldApplyFilter```  to both layers and resources. Once Travis CI passes, I will merge and backport to 2.10.x and 2.9.x.

PR #1624, opened 2016-05-25
GEOS-7557: vectortiles - remove degenerate geometries produced by clipping

DBlasby:
https://osgeo-org.atlassian.net/browse/GEOS-7557
https://github.com/geoserver/geoserver/wiki/GSIP-138

After clipping, the resulting geometries are filtered so that lower-dimensional results are removed.

I also re-formatted the code so it follows GS coding standards.

PR #1887, opened 2016-10-17
GEOS-7557 - filter clipped geometries so degenerate geometries are removed

DBlasby:
This is the same as https://github.com/geoserver/geoserver/pull/1624 except this does NOT contain the formating changes as well (moved to separate PR).

### DIFF
--- a/src/community/vectortiles/src/main/java/org/geoserver/wms/vector/PipelineBuilder.java
+++ b/src/community/vectortiles/src/main/java/org/geoserver/wms/vector/PipelineBuilder.java
@@ -8,6 +8,8 @@ import static org.geotools.renderer.lite.VectorMapRenderUtils.buildTransform;
 
 import java.awt.Rectangle;
 import java.awt.geom.AffineTransform;
+import java.util.ArrayList;
+import java.util.List;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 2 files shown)


src/community/vectortiles/src/main/java/org/geoserver/wms/vector/PipelineBuilder.java
```diff
@@ -313,4 +321,133 @@ protected Geometry _run(Geometry geom) throws Exception {
...
+     * also points and line (degenerative).
+     * 
+     * This will remove the degenerative geometries from the result. ie. input is polygon(s), only polygons are returned input is line(s), only lines
+     * are returned input is point(s), only points returned
+     * 
+     * For mixed input (GeometryCollection), we do the above for each component in the GeometryCollection. i.e. for GeometryCollection( POLYGON(...),
+     * LINESTRING(...) ) it would ensure that the POLYGON(...) only adds Polygons and that the LINESTRING() only adds Lines
+     * 
+     */
+    public static final class ClipRemoveDegenerateGeometries extends Clip {
```
tbarsballe: Maybe add a javadoc here since this is a public class?

src/community/vectortiles/src/main/java/org/geoserver/wms/vector/PipelineBuilder.java
```diff
@@ -293,7 +301,7 @@ protected Geometry _run(Geometry geom) throws Exception {
         }
     }
 
-    private static final class Clip extends Pipeline {
+    public static class Clip extends Pipeline {
```
tbarsballe: Does this need to be public? Would protected work fine?

tbarsballe:
Looks good, although it would be nice if it was rebased into a single commit for ease of tracking/cherry-picking.

PR #1890, opened 2016-10-18
GSIP-138 documentation for VT module

DBlasby:


doc/en/user/source/community/vectortiles/tutorial.rst
```diff
@@ -0,0 +1,237 @@
...
+.. list-table::
+   :header-rows: 1
+   :class: non-responsive
+
+   * - Format
+     - MIME
+     - Description
+   * - `GeoJSON <http://geojson.org/>`_
+     - ``application/json;type=geojson``
+     - Human readable, format (not tiles) widely supported
```
aaime: "format (not tiles)" --> what does this mean? If we ask for a tile don't we get back only what's in the tile, even if the format is GeoJSON?

DBlasby: I'll make it more understandable.  Thanks for reading + feedback.

PR #1892, opened 2016-10-18
[GEOS-7797] Add null check to Style Editor Publishing Tab

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-7797

In addition to a test case for the error in question, also added a generic test case for the Publishing tab.


aaime:
Nice. Will you backport?

tbarsballe:
Yes. I am planning on gettting #1893 and #1894 merged in, then backporting all three.

PR #1893, opened 2016-10-18
[GEOS-7692] Add error checking to Style Editor Layer Attributes tab

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-7692

If the layer is invalid, an error message will be displayed at the top of the page, similar to the one that gets displayed on the Layer Edit page under similar circumstances.


tbarsballe:
Currently, I have no test case for this fix. I have so far been unable to find a way of testing the layer chooser popup using the wicket test framework. I am still looking into a viable test case, but am not especially confident of finding a solution.
I have done some manual testing to verify the fix actually works.

tbarsballe:
Added test case. Will likely need some refactoring depending on which of #1892, #1893, #1894 get merged first.

aaime:
Yep, looks good. Has merge conflicts.

mprins (approved):

PR #1894, opened 2016-10-18
[GEOS-7693] Do not allow WMS layers for Style Page Layer Preview / Attributes

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-7693

Excludes all WMS layers from the layer chooser list.
WMS Layers cannot be styled, so there is no reason to show them in the Style Page Layer / Attribute preview.

### DIFF
--- a/src/web/wms/src/main/java/org/geoserver/wms/web/data/LayerChooser.java
+++ b/src/web/wms/src/main/java/org/geoserver/wms/web/data/LayerChooser.java
@@ -5,6 +5,7 @@
  */
 package org.geoserver.wms.web.data;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


tbarsballe:
Currently, I have no test case for this fix. I have so far been unable to find a way of testing the layer chooser popup using the wicket test framework. I am still looking into a viable test case, but am not especially confident of finding a solution.
I have done some manual testing to verify the fix actually works.

tbarsballe:
Added test case. Will likely need some refactoring depending on which of #1892, #1893, #1894 get merged first.

aaime:
Looking good. Has merge conflicts. The layer list code will cause scalability issues with large catalogs on jdbcconfig (not introduced in this pull request, was already there, just saying..)

mprins (approved):

PR #1895, opened 2016-10-19
[GEOS-7788] original BandNames from coverageViews are lost, being replaced by color model names

aaime:


dromagnoli:
Looks good to me.
Merging it.

mprins (approved):

PR #1896, opened 2016-10-19
PR for [GEOS-7803] GeoFence Embedded does not check for plain Rule Services

afabiani:


src/community/geofence-server/src/main/java/org/geoserver/geofence/internal/InternalUserResolver.java
```diff
@@ -42,6 +42,13 @@ public boolean existsUser(String username) {
                     return true;
                 }
             }
+            
+            for (String roleServiceName : securityManager.listRoleServices()) {
+                if (securityManager.loadRoleService(roleServiceName).getRolesForUser(username) != null && 
+                        !securityManager.loadRoleService(roleServiceName).getRolesForUser(username).isEmpty()) {
+                    return true;
+                }
```
nmco: I would split the user roles retrieve and condition check, I think it will make it more clear and we avoid retrieving two times the same information, something like this:

```
SortedSet<GeoServerRole> userRoles = securityManager.loadRoleService(roleServiceName).getRolesForUser(username);
if (userRoles != null && !userRoles.isEmpty()) {
    return true;
}
```

nmco:
Just a minor comment, there is a test for this use case and community module tests are passing. This pull requests is against 2.9.x this needs to be ported to master.

afabiani:
Redo PR based on master

mprins (approved):

mprins (approved):

PR #1903, opened 2016-10-24
PR - Fix for (GEOS-7805) CSW Service does not work anymore due to EntityResolver bean circular reference

afabiani:


aaime:
Can you elaborate on how/why the pull request works? The readers do depend explicitly on the EntityResolverProvider, which in turn depends on GeoServer... so it seems the change only makes a transitive dependency explicit. What am I missing?

afabiani:
At least in Tomcat container, since geoServer is an alias, if the loader tries to instantiate a bean depending on that, even transitively, before gs-main jar is loaded, we get a"BeanInCreation" exception. At least this is what I am experiencing on Tomcat7 containers.

afabiani:
At least in Tomcat container, since geoServer is an alias, if the loader
tries to instantiate a bean depending on that, even transitively, before
gs-main jar is loaded, we get a"BeanInCreation" exception. At least this is
what I am experiencing on Tomcat7 containers.

afabiani:
Closing this as replaced by

https://github.com/geoserver/geoserver/pull/1921

PR #1905, opened 2016-10-24
PR - Fix for (GEOS-7810) GeoFence Embedded Server should not overwrite configuration custom parameters

afabiani:


etj:
The property file is automatically regenerated, and has this line in the header:
```
### This file may be automatically regenerated by GeoServer, so any changes beside the property values may be lost
```
so it's **not** supposed to hold custom parameters.

PR #1909, opened 2016-10-25
PR for [GEOS-7803] GeoFence Embedded does not check for plain Rule Services

afabiani:
- (GEOS-7803) improve code


nmco:
As a test case and looks good to me, merging it.

PR #1916, opened 2016-10-25
Comment web.xml DTD reference to start when disconnected [GEOS-7813]

jodygarnett:
For more information: https://osgeo-org.atlassian.net/browse/GEOS-7813

### DIFF
--- a/src/web/app/src/main/webapp/WEB-INF/web.xml
+++ b/src/web/app/src/main/webapp/WEB-INF/web.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE web-app PUBLIC "-//Sun Microsystems, Inc.//DTD Web Application 2.3//EN" "http://java.sun.com/dtd/web-app_2_3.dtd">
+<!-- !DOCTYPE web-app PUBLIC "-//Sun Microsystems, Inc.//DTD Web Application 2.3//EN" "http://java.sun.com/dtd/web-app_2_3.dtd"-->
 <web-app>
     <display-name>GeoServer</display-name>
   


jodygarnett:
This is a small change suitable for backport.

aaime:
Wondering if we can put the dtd in the package instead? We do the same for all other OGC schemas for the same reason (to run offline)

jodygarnett:
Kind of up to the configuration of the parser done by Jetty, Tomcat, etc...  (best case is they would contain their own copy of these files as part of "implementing" the Java EE spec).

aaime:
But if they were doing that already, they'd be ignoring the local DTD declaration. What bothers me about removing it is that it makes editing the file harder in tools that are aware of DTDs.

jodygarnett:
Can we include the local DTD, and reference it at the top of the file? (Downside some application servers may freak out since it is not matching one of the standards they support).

jodygarnett:
Tried this a couple ways, including the file locally did not work.

Digging into jetty code they do have a copy of dtd23, and even set up a redirect:

````

redirectEntity("-//Sun Microsystems, Inc.//DTD Web Application 2.3//EN",dtd23);
````

jodygarnett:
I could not determine why jetty is not using its dtd23, this bug fix is still a valid approach.

An alternative would be to upgrade to a newer version of servlet spec - @dvntucker do you have an opinion on this?

Sammers21:
why this pull request in already open?

jodygarnett:
@Sammers21 because it has not been merged (accepted) or closed (rejected) yet.  We are kind of stuck. Do you know anything about how jetty handles redirects?

Sammers21:
@jodygarnett nope. I really have no idea. Have you tried to solve this problem by watching through jetty sources?

aaime:
@jodygarnett it's been many months, time to close this one?

jodygarnett:
I would still like this fix merged, as far as I know I cannot start geoserver when disconnected.

aaime:
All right, you know what the practice is, pull requests that stay there too long get closed and linked from a ticket.

PR #1917, opened 2016-10-25
[GEOS-7691] Update OL preview on save without reloading page

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-7691


jodygarnett:
Looking forward to this, assume it can be backported directly to 2.10.x

tbarsballe:
Yes, this should be safe to backport to 2.10.x. It would be nice to include this in the 2.10.0 release (although cutting it a bit tight)

jodygarnett:
Thansk, works great on master, changing tabs resets the preview but that is life :)

I will backport for the release

PR #1918, opened 2016-10-25
Prevent image being serialized, form uses wicketImage [GEOS-7689]

jodygarnett:
For details see: https://osgeo-org.atlassian.net/browse/GEOS-7689


src/web/wms/src/main/java/org/geoserver/wms/web/data/StyleAdminPanel.java
```diff
@@ -84,6 +84,7 @@
     
     protected WebMarkupContainer legendContainer;
     protected ExternalGraphicPanel legendPanel;
+
```
tbarsballe: Rogue whitespace makes me sad :(

jodygarnett: I am a bad person.

jodygarnett:
Thanks @tbarsballe, backporting to stable. Wow that is all the style page stuff done!

PR #1920, opened 2016-10-26
PR - Fix for (GEOS-7806) JDBCRoleService may cause NPE for roles with NULL property name

afabiani:
Weird case which may happen in some situations where the table has a different definition (see rolesddl.xml sample in the DB2 test use case).


mbarto:
Looks good to me

bencaradocdavies:
This PR causes build failures when the online postgis test fixture {{~/.geoserver/postgis.properties}} is configured:

[GEOS-7847] Invalid SQL "auto_increment" causes PostGisRoleServiceTest and PostGisUserDetailsServiceTest to fail
https://osgeo-org.atlassian.net/browse/GEOS-7847

afabiani:
Fixes available at this PR:

https://github.com/geoserver/geoserver/pull/1959

PR #1922, opened 2016-10-26
Replace Jettison libraries with JSONP (JSR-353)

emerkle826:
Update the GeoGig plugin to use the new Response Representation framework in GeoGig that removes the coupling to XMLStreamWriter and Jettison bridging libraries.


emerkle826:
This should build now that the GeoGig dev branch has been deployed to the maven repository

PR #1928, opened 2016-10-28
fix miss initialized flag in WFSPPIO to fix GEOS-5663

ianturton:
See [GEOS-5663](https://osgeo-org.atlassian.net/browse/GEOS-5663) for more details.


aaime:
+1

PR #1931, opened 2016-10-28
[GEOS-7822] Support unset primitive fields on REST PUT against ServiceInfo

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-7822

This fix uses the same approach as in https://github.com/geoserver/geoserver/pull/1541, but for ServiceInfo instead of CatalogInfo this time around

### DIFF
--- a/src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
+++ b/src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
@@ -1542,24 +1542,8 @@ public class XStreamPersister {
             Object catalogObject = callback.getCatalogObject();
             
             if (catalogObject != null) {
-                Class i = null;
-                if (emptyObject instanceof CoverageInfo) {
-                    i = CoverageInfo.class;
-                } else if (emptyObject instanceof CoverageStoreInfo) {
-                    i = CoverageStoreInfo.class;
-                } else if (emptyObject instanceof DataStoreInfo) {
-                    i = DataStoreInfo.class;
-                } else if (emptyObject instanceof FeatureTypeInfo) {
-                    i = FeatureTypeInfo.class;
-                } else if (emptyObject instanceof LayerGroupInfo) {
-                    i = LayerGroupInfo.class;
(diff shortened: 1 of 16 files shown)


tbarsballe:
Part of this fix adds the getObjectClass() to method, in order to obtain what type of ServiceInfo is being converted. This is necessary because gs-wms is downstream of gs-main, meaning gs-main is unaware of the WMSInfo class.

For consistency, I have also modified AbstractCatalogInfoConverter with the same implementation (as a seperate commit).

src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
```diff
@@ -2291,4 +2278,39 @@ public Object doUnmarshal(Object result,
...
+        public Object doUnmarshal(Object result,
+                HierarchicalStreamReader reader, UnmarshallingContext context) {
+            ServiceInfoImpl obj = (ServiceInfoImpl) super.doUnmarshal(result, reader, context);
+            
+            return obj;
+        }
+        @Override
+        protected Object instantiateNewInstance(HierarchicalStreamReader reader,
+                UnmarshallingContext context) {
+            Object emptyObject = super.instantiateNewInstance(reader, context);
```
jodygarnett: This is where the bug is actually fixed, can we get a javadoc or something to show how important this logic is

src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
```diff
@@ -2291,4 +2278,39 @@ public Object doUnmarshal(Object result,
             return obj;
         }
     }
+    
+    /**
+     * Converter for ServiceInfo class
+     */
+    public class ServiceInfoConverter extends AbstractReflectionConverter {
```
jodygarnett: Can we have some more details for this new public class?

tbarsballe:
Javadocs updated

tbarsballe:
Backported to 2.10.x.

PR #1937, opened 2016-11-01
[GEOS-7821] Fixup gs-restconfig services ftl templates and add minimal  HTML tests

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-7821

Added `!` to most properties (show "" when a value is null instead of throwing an error).
Added proper validation and parsing for list properties.
Minor HTML syntax fixup.
Minimal HTML GET test cases.


jodygarnett:
Looks good @tbarsballe, assume this is suitable for backport.

tbarsballe:
Yes, this should be definitely be backported as far back as is feasible. It looks like this never worked properly.

I have backported to 2.9.x and 2.10.x.

PR #1940, opened 2016-11-03
GEOS-7834: new WMS GetLegendGraphic options to improve labels to icon…

mbarto:
…s margins

### DIFF
--- a/doc/en/user/source/services/wms/get_legend_graphic/index.rst
+++ b/doc/en/user/source/services/wms/get_legend_graphic/index.rst
@@ -84,6 +84,7 @@ Here is a description of the various parameters that can be used in ``LEGEND_OPT
     - **bgColor (hex)** background color for the generated legend, values are expressed in ``0xRRGGBB`` format
     - **dpi (integer)** sets the DPI for the current request, in the same way as it is supported by GetMap. Setting a DPI larger than 91 (the default) makes all fonts, symbols and line widths grow without changing the current scale, making it possible to get a high resolution version of the legend suitable for inclusion in printouts 
     - **forceLabels** "on" means labels will always be drawn, even if only one rule is available. "off" means labels will never be drawn, even if multiple rules are available. Off by default.
+    - **labelMargin** margin (in pixels) to use between icons and labels.
     - **layout** sets icons layout to be **vertical** (default) or **horizontal**.
     - **columnheight** enables **multicolumn** layout when layout is **vertical**. Each column height is limited by the columnheight value (in pixels).
     - **rowwidth** enables **multirow** layout when layout is **horizontal**. Each row width is limited by the rowwidth value (in pixels).
@@ -141,6 +142,7 @@ The producer for the raster legend will make use of this elements in order to bu
     - the height of each row is set to the maximum height of the single elements
     - the width of each row is set to the sum of the width of the various elements plus the various paddings
     - **dx,dy** the spaces between elements and rows are set to the 15% of the requested width and height. Notice that **dy** is ignored for the colormaps of type **ramp** since they must create a continuous color strip.
+    - **absoluteMargins** true/false, if true dx is considered as a fixed number of pixels, instead of a percentage of the width.
     - **mx,my** the margins from the border of the legends are set to the 1.5% of the total size of the legend
 
 Just to jump right to the conclusions (which is a bad practice I know, but no one is perfect ), here below I am adding an image of a sample legend with all the various options at work. The request that generated it is the following::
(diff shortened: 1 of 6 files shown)


src/wms/src/main/java/org/geoserver/wms/legendgraphic/ColorMapLegendCreator.java
```diff
@@ -160,6 +160,8 @@
         private double columnMarginPercentage = LegendUtils.columnPaddingFactor;
 
         private double hMarginPercentage = LegendUtils.marginFactor;
+        
+        private boolean absoluteMargins = false;
```
nmco: If we want by default consider absolute values this should be true no ?

doc/en/user/source/services/wms/get_legend_graphic/index.rst
```diff
@@ -141,6 +142,7 @@ The producer for the raster legend will make use of this elements in order to bu
     - the height of each row is set to the maximum height of the single elements
     - the width of each row is set to the sum of the width of the various elements plus the various paddings
     - **dx,dy** the spaces between elements and rows are set to the 15% of the requested width and height. Notice that **dy** is ignored for the colormaps of type **ramp** since they must create a continuous color strip.
+    - **absoluteMargins** true/false, if true dx is considered as a fixed number of pixels, instead of a percentage of the width.
```
nmco: Maybe it should be clear that this option only affects rasters legends and that it may be combined with the dx option to get the desirable margin.

src/wms/src/main/java/org/geoserver/wms/legendgraphic/BufferedImageLegendGraphicBuilder.java
```diff
@@ -671,7 +675,7 @@ private RenderedImage getLayerLegend(LegendRequest legend, int w, int h, boolean
      */
     private BufferedImage mergeGroups(List<RenderedImage> imageStack, Rule[] rules, GetLegendGraphicRequest req,
             boolean forceLabelsOn, boolean forceLabelsOff) {
-        LegendMerger.MergeOptions options = LegendMerger.MergeOptions.createFromRequest(imageStack, 0, 0, 0, req, forceLabelsOn, forceLabelsOff);
+        LegendMerger.MergeOptions options = LegendMerger.MergeOptions.createFromRequest(imageStack, 0, 0, 0, 0, req, forceLabelsOn, forceLabelsOff);
```
nmco: If I understood correctly the idea is that the label margin default value is 3, right ? But in this case the label margin value will have no effect so we set it to zero, right ?

src/wms/src/main/java/org/geoserver/wms/legendgraphic/ColorMapLegendCreator.java
```diff
@@ -924,7 +934,7 @@ private BufferedImage mergeRows(Queue<BufferedImage> legendsQueue) {
         } else {
             List<RenderedImage> imgs = new ArrayList<RenderedImage>(legendsQueue);
             
-            LegendMerger.MergeOptions options = new LegendMerger.MergeOptions(imgs, (int) dx, (int) dy, (int) margin, backgroundColor, transparent, true, layout, rowWidth, 
+            LegendMerger.MergeOptions options = new LegendMerger.MergeOptions(imgs, (int) dx, (int) dy, (int) margin, 0, backgroundColor, transparent, true, layout, rowWidth,
```
nmco: The same as above, if I understood correctly the idea is that the label margin default value is 3, right ? But in this case the label margin value will have no effect so we set it to zero, right ?

src/wms/src/main/java/org/geoserver/wms/legendgraphic/BufferedImageLegendGraphicBuilder.java
mbarto: Yes, that's correct

doc/en/user/source/services/wms/get_legend_graphic/index.rst
mbarto: Yes, that's correct

src/wms/src/main/java/org/geoserver/wms/legendgraphic/ColorMapLegendCreator.java
mbarto: Yes, that's correct

src/wms/src/main/java/org/geoserver/wms/legendgraphic/ColorMapLegendCreator.java
mbarto: Right, going to fix

nmco:
Looks good to me. The changes have been discussed in the mailing list (http://osgeo-org.1560.x6.nabble.com/WMS-GetLegendGraphic-td5293025.html). The introduced changes are covered by test cases and properly documented. Merging.

PR #1944, opened 2016-11-04
PR - Fix for (GEOS-7810) GeoFence Embedded Server should not overwrite configuration custom parameters

afabiani:


etj:
The initial concern was about setting the user and role service name in a property file, that's why 
GEOS-7810 is titled "GeoFence Embedded Server should not overwrite configuration custom parameters".

This PR anyway solves the issue in a better way, by getting the effective service name from the Geoserver internals.

NielsCharlier:
@afabiani @etj  Something very strange happened here. The default role service name is used as default user/group service name. Why??? I don't understand the logic.
Also, the documentation was never updated that this configuration was removed.

afabiani:
Mmm, the PR tries to avoid this behavior. It falls back to DEFAULT ROLE SERVICE only when no active one can be found, instead of always using it.

afabiani:
also... this property could not be configured

protected String defaultServiceName;

it was hardcoded on spring XML.

NielsCharlier:
but this is the user/group REST, not the roles REST. the roles REST was already using the default role service.
A user/group service and a role service are two entirely different things. 

You are requesting the role service name and then using it as a user/group service name.

OK, there was a bug there... my fault. but the solution introduces a new bug...

afabiani:
Can you open a new issue for that?

Thanks for reporting,
-A.

NielsCharlier:
https://osgeo-org.atlassian.net/browse/GEOS-8205

BUt, I have suggested moving the user/group/role REST to the core modules (since they are geofence independent, and now the core rest uses spring mvc as well):
https://osgeo-org.atlassian.net/browse/GEOS-8204

I have made a patch for that which fixes this one as well
https://github.com/geoserver/geoserver/pull/2439

afabiani:
@NielsCharlier seems good to me. @etj what do you think?

PR #1949, opened 2016-11-07
[GEOS-7840] Consistently use java-vector-tile:1.0.9

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-7840

This one will need to be backported to 2.10.x and 2.9.x

### DIFF
--- a/src/community/script/groovy/pom.xml
+++ b/src/community/script/groovy/pom.xml
@@ -15,6 +15,12 @@
  <artifactId>gs-script-groovy</artifactId>
  <packaging>jar</packaging>
  <name>Groovy Scripting Extension</name>
+ <repositories>
+  <repository>
+   <id>ECC</id>
+   <url>https://github.com/ElectronicChartCentre/ecc-mvn-repo/raw/master/releases</url>
+  </repository>
+ </repositories>
  <dependencies>
   <dependency>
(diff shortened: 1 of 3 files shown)


src/community/script/groovy/pom.xml
```diff
@@ -27,10 +33,10 @@
    <version>1.7-SNAPSHOT</version>
   </dependency>
   <dependency>
-    <!-- this dependency taken from geoscript-groovy pom but upgraded to 1.0.8 -->
+    <!-- this dependency taken from geoscript-groovy pom but upgraded to 1.0.9 -->
     <groupId>no.ecc.vectortile</groupId>
     <artifactId>java-vector-tile</artifactId>
-    <version>1.0.8</version>
+    <version>1.0.9</version>
```
jodygarnett: should this be in a dependency management section?

tbarsballe: Based on https://github.com/geoserver/geoserver/blob/master/src/extension/vectortiles/pom.xml#L17 , apparently not?

tbarsballe:
Moved version information to dependency management.

tbarsballe:
Backported to 2.10.x and 2.9.x

PR #1950, opened 2016-11-08
Fix for GEOS-7841

silenius:
Geoserver should work with a read-only GEOSERVER_DATA_DIR
More info here: https://osgeo-org.atlassian.net/browse/GEOS-7841


jodygarnett:
Short feedback - we ask that fixes include a test case as per our contribution guidelines.

Longer feedback - we have not designed GeoServer with this use-case in mind. If you would like to pursue this course of action consider a change proposal.  What could be effected: cross process file locking, communication of catalog change, file formats that require disk access, untangling log files, etc...

PR #1952, opened 2016-11-08
[Backport 2.9.x] - Fix for (GEOS-7806) JDBCRoleService may cause NPE for roles without properties

afabiani:


bencaradocdavies:
This PR causes build failures when the online postgis test fixture ```~/.geoserver/postgis.properties``` is configured:

[GEOS-7847] Invalid SQL "auto_increment" causes PostGisRoleServiceTest and PostGisUserDetailsServiceTest to fail
https://osgeo-org.atlassian.net/browse/GEOS-7847

PR #1956, opened 2016-11-09
[GEOS-7826] Add check for no layers when instantiating Style Editor page

tbarsballe:
Fix for https://osgeo-org.atlassian.net/browse/GEOS-7826

Added null check to the page constructor.
Added error messages when trying to navigate to other tabs when no layers are available.


jodygarnett:
Nice corner case, happens every time we startup without a data directory.

aaime:
Hmm... the Travis build keeps on failing early in the build, but only on Travis (and only for this pull request... that's a bit crazy).

tbarsballe:
That is odd, this should not be able to affect gs-main.
Re-kicked the build and looks like it is working again. I will go ahead and merge (+ backport to 2.10)

tbarsballe:
Cherry-picked to 2.10.x

PR #1957, opened 2016-11-09
GEOS-7844: move Importer to use high level api

DBlasby:
This PR does two things;
a) splits the (quite long) loadIntoDataStore() method into two parts - setup and transferring features
b) provide two implementation for moving features one based on the low-level api (read-a-feature write-a-feature) and one based on the high-level api (FeatureCollection/featureStore.addFeatures).

For the non-datastore formats (GeoJSON, GML, KML), this PR will not affect functionality.  However, for Datastore-backed data sources, this uses the FeatureCollection/featureStore.addFeatures() method.

The original feature transformation (cf copyFeaturesLowLevelAPI) has been re-implemented inside a FeatureCollection/FeatureIterator.

The FeatureIterator provides two pieces of functionality;
a) cancelability (cf CancelableFeatureIterator)
b) feature xformation (cf TransformingFeatureIterator)

The actual functionality remains the same - this just does the same thing inside the FeatureIterator instead of in a read-a-feature write-a-feature while loop.

Notes;
a) I re-formated the modified code
b) there was already existing adequate test coverage

### DIFF
--- a/src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
+++ b/src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
@@ -65,11 +65,14 @@ import org.geotools.coverage.grid.io.StructuredGridCoverage2DReader;
 import org.geotools.data.DataStore;
 import org.geotools.data.DefaultTransaction;
 import org.geotools.data.FeatureReader;
+import org.geotools.data.FeatureSource;
 import org.geotools.data.FeatureStore;
 import org.geotools.data.FeatureWriter;
 import org.geotools.data.Transaction;
 import org.geotools.data.directory.DirectoryDataStore;
 import org.geotools.data.shapefile.ShapefileDataStore;
+import org.geotools.feature.FeatureCollection;
+import org.geotools.feature.collection.DecoratingFeatureCollection;
 import org.geotools.feature.simple.SimpleFeatureTypeBuilder;
 import org.geotools.geometry.GeneralEnvelope;
 import org.geotools.geometry.jts.ReferencedEnvelope;
@@ -1027,8 +1030,9 @@ public class Importer implements DisposableBean, ApplicationListener {
                     calculateBounds(resource);
                 }
             }
-            catch(Exception e) {
-                LOGGER.log(Level.SEVERE, "Error occured during import", e);
+            catch(Throwable th) {
+                LOGGER.log(Level.SEVERE, "Error occured during import", th);
+                Exception e = (th instanceof Exception) ? (Exception) th : new Exception(th);
                 task.setError(e);
                 task.setState(ImportTask.State.ERROR);
                 return;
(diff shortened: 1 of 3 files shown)


src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -1172,192 +1175,107 @@ boolean doPostTransform(ImportTask task, ImportData data, TransformChain tx) {
...
-        FeatureWriter writer = null;
+
         // using this exception to throw at the end
         Exception error = null;
         Transaction transaction = new DefaultTransaction();
         try {
-            reader = format.read(data, task);
 
-            SimpleFeatureType featureType = (SimpleFeatureType) reader.getFeatureType();
+            SimpleFeatureType featureType = (SimpleFeatureType) task.getMetadata()
```
jodygarnett: Is that really the api? That looks pretty informal ...

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -1172,192 +1175,107 @@ boolean doPostTransform(ImportTask task, ImportData data, TransformChain tx) {
...
+
                 if (!uniquifiedFeatureTypeName.equals(featureTypeName)) {
-                    //update the metadata
+                    // update the metadata
                     task.getLayer().getResource().setName(uniquifiedFeatureTypeName);
                     task.getLayer().getResource().setNativeName(uniquifiedFeatureTypeName);
-                    
-                    //retype
+
+                    // retype
```
tbarsballe: While I appreciate the formatting fixes, such changes should really be a separate commit. Mixing format and code changes makes a PR rather more difficult to review.

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -1369,16 +1287,146 @@ else if (isPostGISDataStore(dataStore)) {
...
         // finally, throw any error
         if (error != null) {
             throw error;
         }
     }
 
+    private Exception copyFeaturesHighLevelAPI(ImportData data, ImportTask task,
+            DataStoreFormat format, DataStore dataStore, Transaction transaction,
+            String featureTypeName, String uniquifiedFeatureTypeName,
+            FeatureDataConverter featureDataConverter, VectorTransformChain tx) throws IOException {
```
jodygarnett: Would prefer "copyFromFeatureSource" since we do not have a formal high / low / internal api.

Is it worth untangling some of these featureTypeName, dataStore, transaction parameters by passing in a single configured featureStore to this method?

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -1369,16 +1287,146 @@ else if (isPostGISDataStore(dataStore)) {
...
+        Exception error = null;
+        ProgressMonitor monitor = task.progress();
```
jodygarnett: Should this be a sub progress monitor? Just asking I am not sure ...

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -1369,16 +1287,146 @@ else if (isPostGISDataStore(dataStore)) {
...
+        try {
+            long startTime = System.currentTimeMillis();
+            task.clearMessages();
+
+            task.setTotalToProcess(format.getFeatureCount(task.getData(), task));
+            LOGGER.info("begining import - highlevel api");
```
jodygarnett: Set logging level to fine for this one

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -1369,16 +1287,146 @@ else if (isPostGISDataStore(dataStore)) {
...
+
+            FeatureSource fs = format.getFeatureSource(data, task);
+            FeatureCollection fc = fs.getFeatures();
+
+            FeatureStore featureStore = (FeatureStore) dataStore
+                    .getFeatureSource(uniquifiedFeatureTypeName);
+            featureStore.setTransaction(transaction);
+
+            fc = new TransformingCancelableInteratingFC(fc, monitor, featureDataConverter,
+                    featureStore.getSchema(), tx, task, dataStore);
```
jodygarnett: Would like to follow class name conventions here; there should be some base classes to start from.
- fc this is your original featureCollection
- monitor - this could only be used once, not suitable for a reusable FeatureCollection implementation
- tx, datastore - this is already known to the fc why so many parameters?

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -1369,16 +1287,146 @@ else if (isPostGISDataStore(dataStore)) {
...
+
+        if (error != null || monitor.isCanceled()) {
+            // all sub exceptions in this catch block should be logged, not thrown
+            // as the triggering exception will be thrown
+
+            // failure, rollback transaction
+            try {
+                transaction.rollback();
+            } catch (Exception e1) {
+                LOGGER.log(Level.WARNING, "Error rolling back transaction", e1);
```
jodygarnett: provide some context here as this will be the only message end users will see in their logs.
```
LOGGER.log(Level.WARNING,"Unable to load data into "+ uniquifiedFeatureTypeName+", rolling back data insert:"+e1, e1);
```

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -1369,16 +1287,146 @@ else if (isPostGISDataStore(dataStore)) {
...
+            }
+
+            // attempt to drop the type that was created as well
+            try {
+                dropSchema(dataStore, featureTypeName);
+            } catch (Exception e1) {
+                LOGGER.log(Level.WARNING, "Error dropping schema in rollback", e1);
```
jodygarnett: interesting, createSchema does not allow the table to be crated on a transaction.

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -1369,16 +1287,146 @@ else if (isPostGISDataStore(dataStore)) {
...
+            }
+        }
+
+        return error;
+    }
+
+    Exception copyFeaturesLowLevelAPI(FeatureReader reader, ImportTask task, VectorFormat format,
+            DataStore dataStore, Transaction transaction, String featureTypeName,
+            String uniquifiedFeatureTypeName, FeatureDataConverter featureDataConverter,
+            VectorTransformChain tx) throws IOException {
```
jodygarnett: Same feedback as above:
- prefer "copyFromFeatureReader"
- that is a lot of parameters

aside: There is a DataUtilities method that can wrap a FeatureReader as a FeatureCollection, but it may store the results in memory. uDig has a single use feature collection wrapper which would be suitable but is a pretty brutal solution.

src/extension/importer/core/src/main/java/org/geoserver/importer/TransformingCancelableInteratingFC.java
```diff
@@ -0,0 +1,150 @@
...
+import org.opengis.feature.type.FeatureType;
+
+import com.vividsolutions.jts.geom.Geometry;
+
+/**
+ * FeatureCollection that does two things required by the importer; a) provide cancel functionality b) Do some FeatureType Transforming
+ *
+ * This class is simply wraps the FeatureIterator with two iterators wrappers that provide the above functionality.
+ */
+public class TransformingCancelableInteratingFC extends DecoratingFeatureCollection {
```
jodygarnett: Nice, you found DecoratingFeatureCollection. 

- You may wish to make this package visible, rather than commit to additional public api
- Can we rename this to "ImportTransformFeatureCollection" or something to follow the naming convention.

src/extension/importer/core/src/main/java/org/geoserver/importer/TransformingCancelableInteratingFC.java
```diff
@@ -0,0 +1,150 @@
...
+
+    ProgressMonitor monitor;
+
+    FeatureDataConverter featureDataConverter;
+
+    FeatureType resultingFT;
+
+    VectorTransformChain tx;
+
+    ImportTask task;
```
jodygarnett: This task can provide a monitor as needed, no need to pass it in?

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
tbarsballe: No, I don't believe it should

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
DBlasby: This is the original implementation

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
DBlasby: I updated the name.

I don't think untangling is going to help at all - there's just going to be another method added that pretty much does what this one does.

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
DBlasby: Not sure what you're saying here...

src/extension/importer/core/src/main/java/org/geoserver/importer/TransformingCancelableInteratingFC.java
DBlasby: I changed visibility and name.

src/extension/importer/core/src/main/java/org/geoserver/importer/TransformingCancelableInteratingFC.java
DBlasby: fixed

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
DBlasby: updated

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
DBlasby: I updated this, as well as the other  one.

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
DBlasby: If you look at the VectorTransformChain, it requires a lot of classes.
tx - is the VectorTransformChain (I'll rename - that was it's original name)
datastore - this is actually the destination datastore (I'll rename - that was it's original name)

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -1369,16 +1287,146 @@ else if (isPostGISDataStore(dataStore)) {
...
+            FeatureStore featureStore = (FeatureStore) dataStoreDestination
+                    .getFeatureSource(uniquifiedFeatureTypeName);
+            featureStore.setTransaction(transaction);
+
+            fc = new ImportTransformFeatureCollection(fc, featureDataConverter,
+                    featureStore.getSchema(), tx, task, dataStoreDestination);
+
+            featureStore.addFeatures(fc);
+
+        } catch (Exception e) {
```
jodygarnett: Think we want to catch RuntimeExceptions here as well - or use a try/finally block to ensure the transaction is either committed or rolled back.

jodygarnett:
Excellent Dave, I had one thing I was unsure of with try/finally, but this is looking great.

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
DBlasby: Yes - further up the chain.  CF line 1333-1344.  It rolls back the transaction as well as dropping the created schema.

tbarsballe:
I would still like to make sure we squash the commits before merging, but yes, this is looking good.

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
jodygarnett: Sorry Dave I ment Throwable, we need to catch Throwable here or we have a chance of missing the clean up code on 1333-1344

DBlasby: changed

aaime:
So, before we had a single code path to maintain using readers, and now there are two, one using readers, one using feature collections? Looks more code, more options to understand, where is the improvement?

PR #1958, opened 2016-11-10
GEOS-7844: move importer to use high-level DS api

DBlasby:
This PR does two things;
a) splits the (quite long) loadIntoDataStore() method into two parts - setup and transferring features
b) provide two implementation for moving features one based on the low-level api (read-a-feature write-a-feature) and one based on the high-level api (FeatureCollection/featureStore.addFeatures).

For the non-datastore formats (GeoJSON, GML, KML), this PR will not affect functionality. However, for Datastore-backed data sources, this uses the FeatureCollection/featureStore.addFeatures() method.

The original feature transformation (cf copyFeaturesLowLevelAPI) has been re-implemented inside a FeatureCollection/FeatureIterator.

The FeatureIterator provides two pieces of functionality;
a) cancelability (cf CancelableFeatureIterator)
b) feature xformation (cf TransformingFeatureIterator)

The actual functionality remains the same - this just does the same thing inside the FeatureIterator instead of in a read-a-feature write-a-feature while loop.

Notes;
a) I re-formated the modified code
b) there was already existing adequate test coverage

CF https://github.com/geoserver/geoserver/pull/1957
There are no changes in this PR relative to that one, other than squishing.

### DIFF
--- a/src/extension/importer/core/src/main/java/org/geoserver/importer/DataStoreFormat.java
+++ b/src/extension/importer/core/src/main/java/org/geoserver/importer/DataStoreFormat.java
@@ -28,6 +28,7 @@ import org.geotools.data.DataStore;
 import org.geotools.data.DataStoreFactorySpi;
 import org.geotools.data.DataUtilities;
 import org.geotools.data.FeatureReader;
+import org.geotools.data.FeatureSource;
 import org.geotools.data.FileDataStoreFactorySpi;
 import org.geotools.data.Query;
 import org.geotools.data.Transaction;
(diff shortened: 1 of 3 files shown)


aaime:
So, before we had a single code path to maintain using readers, and now there are two, one using readers, one using feature collections? Looks more code, more options to understand, where is the improvement?

DBlasby:
Andrea - the main reason for doing this is speed improvements.  In particular, the geogig store is 40 times faster for a smallish import and even faster for a larger import.

aaime:
Ah ok, target store specific performance improvement then. All right, hopefully it will make Boundless more interested in sharing maintenance of this module ;-)

+1

DBlasby:
Certainly - this module is used heavily by geonode and gsconfig. Cheers!

jodygarnett:
Can this be backported?

PR #1966, opened 2016-11-15
[GEOS-7852] Ensure passed catalog objects exist when calling setDefault

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-7852


tbarsballe:
Moved fix to CatalogImpl level

aaime:
+1 Looks good to me

PR #1967, opened 2016-11-15
Remove duplicate entry in "GeoServerApplication.properties"

dalbani:


tbarsballe:
Looks Good, thanks for the fix.

PR #1968, opened 2016-11-15
[GEOS-7837]: Add sample for WPS Download output as a reference + some…

dromagnoli:
… fixes on bandIndices params example


aaime:
Looks good to me

PR #1969, opened 2016-11-15
[GEOS-7853] Provide field for "max map buffer" setting on WMS admin page

dalbani:
JIRA: https://osgeo-org.atlassian.net/browse/GEOS-7853


tbarsballe:
Looks good, thanks for the fix.

PR #1973, opened 2016-11-16
[GEOS-7464] Using this tile cache cleans up the tile cache at the end of a request with very large images a lot faster

aaime:


dromagnoli:
Looks good to me.

PR #1974, opened 2016-11-16
[GEOS-7857] Fix CascadeDeleteVisitor for nested layer groups

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-7857

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/CascadeDeleteVisitor.java
+++ b/src/main/src/main/java/org/geoserver/catalog/CascadeDeleteVisitor.java
@@ -134,7 +134,7 @@ public class CascadeDeleteVisitor implements CatalogVisitor {
                 
                 // either update or remove the group
                 if(group.getLayers().size() == 0) {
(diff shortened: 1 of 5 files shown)


src/main/src/main/java/org/geoserver/catalog/CascadeDeleteVisitor.java
```diff
@@ -134,7 +134,7 @@ public void visit(LayerInfo layer) {
                 
                 // either update or remove the group
                 if(group.getLayers().size() == 0) {
-                    catalog.remove(group);
+                    visit(catalog.getLayerGroup(group.getId()));
```
jodygarnett: Had a bit of a hunt to find that the remove( group ) method is still being called inside the visit.

jodygarnett:
Can this fix be backport to 2.10 / 2.9 ?

tbarsballe:
Yes, I have backported to 2.9.x and 2.10.x

PR #1976, opened 2016-11-17
[GEOS-7860] Add test use case for the relaxed GML encoder and update documentation

nmco:
Associated issue:
https://osgeo-org.atlassian.net/browse/GEOS-7860

Depends on GT pull request:
https://github.com/geotools/geotools/pull/1386


nmco:
The builds fails because as expected in the new added test:
```
Tests run: 27, Failures: 1, Errors: 0, Skipped: 2, Time elapsed: 7.23 sec <<< FAILURE!
testNonValidNestedGML(org.geoserver.test.FeatureChainingWfsTest)  Time elapsed: 64 sec  <<< FAILURE!
java.lang.AssertionError: expected:<5> but was:<0>
```

Once th PR is merged on GeoTools side I will start the build again.

bencaradocdavies:
@nmco I still see these failures even after a local test merge and build of both GeoTools and GeoServer, in both Maven and Eclipse. In Eclipse I can see the test enter your new method and still fail:
```
Failed tests:   testNonValidNestedGML(org.geoserver.test.FeatureChainingWfsTest): expected:<5> but was:<0>
  testGml311Links(org.geoserver.test.web.Gml311LinksTest): expected:<5> but was:<6>
```

doc/en/user/source/data/app-schema/mapping-file.rst
```diff
@@ -278,6 +278,7 @@ E.g.::
 
 Note that the GML encoding rules require that complex types are never the direct property of another complex type; they are always contained in a property type to ensure that their type is encoded in a surrounding element. Encoded GML is always type/property/type/property. This is also known as the GML "striping" rule. The consequence of this for app-schema mapping files is that ``targetAttributeNode`` must be applied to the property and the type must be set to the XSD property type, not to the type of the contained attribute (``gsml:CGI_TermValuePropertyType`` not ``gsml:CGI_TermValueType``). Because the XPath refers to a property type not the encoded content, ``targetAttributeNode`` appears in a mapping with ``targetAttribute`` and no other elements when using with complex types.
 
+The GML complex features encoder was relaxed to be able to encode nested complex features that are mapped to a complex type that doesn't respect the rule above. The Java configuration property ``encoder.relaxed`` can be set to ``false`` to disable this feature.
```
bencaradocdavies: Rather than using the past tense here, perhaps just stick to the present state of the XML encoder. The first sentence then becomes: "The XML encoder will encode nested complex features that are mapped to a complex type that does not respect the GML striping rule."

bencaradocdavies:
@nmco other than the test failures, good clean code with thorough tests. Nearly ready to be merged.

doc/en/user/source/data/app-schema/mapping-file.rst
nmco: Thanks, fixed.

nmco:
Thanks for testing this @bencaradocdavies, the test failure was valid, I made a mistake when committing the isNonStripedNestedElement method, I fixed it in the GeoTools PR (sorry).

bencaradocdavies:
@nmco I see a single app-schema online test failure against postgis:
```
Failed tests:   testNonValidNestedGML(org.geoserver.test.FeatureChainingWfsTest): expected:<5> but was:<4>
```

This test passes in non-database mode (using the plain properties file).

Here is how to run from JUnit for debugging:
http://docs.geoserver.org/latest/en/developer/programming-guide/app-schema/index.html#running-tests-from-junit
```
I used:
```
-Dapp-schema.joining=true -DtestDatabase=postgis -Xmx1024m
```

nmco:
Interesting, this is the data of the "SIMPLECONTENT" type:

|ID|STRING|LINK_ONE|LINK_TWO|URN_ID|PKEY|
|---|----------|--------------|---------------|----------|-------|
| sc.1 | string_one|cc.1 |cc.2|1GRAV|sc.1|
| sc.2 | string_two|cc.1 |cc.2|1TILL|sc.2|
| sc.3 | string_three||cc.2|6ALLU|sc.3|

Note that "LINK_ONE" value for the third element is NULL.

In my test I was doing a "self join" by column "LINK_ONE", which returns this results when using the properties file:

| PKEY | ID | STRING | LINK_ONE | LINK_TWO | URN_ID | 
|---------|----|-------------|---------------|----------------|------------| 
| sc.1 | sc.1 | string_one | cc.1 | cc.2 | 1GRAV | sc.1 | 
| sc.2 | sc.2 | string_two | cc.1 | cc.2 | 1TILL | sc.1 | 
| sc.1 | sc.1 | string_one | cc.1 | cc.2 | 1GRAV | sc.2 | 
| sc.2 | sc.2 | string_two | cc.1 | cc.2 | 1TILL | sc.2 | 
| sc.3 | sc.3 | string_three |  | cc.2 | 6ALLU |  | 

When using a database a "INNER JOIN" is used which means that NULL values are ignored:

| PKEY | ID | STRING | LINK_ONE | LINK_TWO | URN_ID | 
|---------|----|-------------|---------------|----------------|------------| 
| sc.1 | sc.1 | string_one | cc.1 | cc.2 | 1GRAV | sc.1 | 
| sc.2 | sc.2 | string_two | cc.1 | cc.2 | 1TILL | sc.1 | 
| sc.1 | sc.1 | string_one | cc.1 | cc.2 | 1GRAV | sc.2 | 
| sc.2 | sc.2 | string_two | cc.1 | cc.2 | 1TILL | sc.2 |

This is why the test fails, I cannot really say which one of the results is valid, anyway in my case I just update my test case and do the nested feature mapping by column  LINK_TWO.

bencaradocdavies:
@nmco that is an interesting finding. It might be that for the property file, NULL == NULL, but SQL JOIN has different semantics. @NielsCharlier and @riniangreani do you have any comment on this edge case? The discrepancy is a tiny buglet. I suspect that the database version is less surprising to users.

I will run a final integration test.

bencaradocdavies:
@nmco merge blocked by an unrelated failure in gs-main.

bencaradocdavies:
Thanks @nmco, all integration tests passed.

NielsCharlier:
@bencaradocdavies  I agree that the INNER JOIN results makes most sense. I would even go as far as saying that if we assume NULL = NULL for property files, that is in fact a bug. Three-valued logic should be respected for property files also.

aaime:
Three valued logic you say? As far as I remember work was done in the JDBC filter encoders to make sure the two value logic of OGC filter is correctly mapped onto databases.

NielsCharlier:
I see. That is a choice made for OGC filters, I guess they must have good reasons for doing that. But in this case it does make more sense to reason the traditional way, where NULL represents the absence of a link (and should therefore not be linked)

PR #1977, opened 2016-11-17
[GEOS-7760] Fix padding on workspace settings

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-7760


tbarsballe:
Backported to 2.9.x and 2.10.x

PR #1979, opened 2016-11-17
[GEOS-7858] Verify layer group is not used before removing from catalog

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-7858

Actual bugfix is in CatalogImpl/CatalogImplTest 

There are a bunch of additional test fixes to make other downstream test cases behave properly


jodygarnett:
Thanks, looking at backport now.

PR #1981, opened 2016-11-18
[GEOS-7584] Improve security of GetFeatureInfo as HTML

dalbani:
https://osgeo-org.atlassian.net/browse/GEOS-7584

### DIFF
--- a/src/community/jms-cluster/jms-geoserver/src/main/java/org/geoserver/cluster/impl/rest/ClusterResource.java
+++ b/src/community/jms-cluster/jms-geoserver/src/main/java/org/geoserver/cluster/impl/rest/ClusterResource.java
@@ -123,7 +123,7 @@ public class ClusterResource extends ReflectiveResource {
 
 		@Override
 		protected Configuration createConfiguration(Object data, Class clazz) {
-			Configuration cfg = new Configuration();
+			Configuration cfg = new Configuration(Configuration.VERSION_2_3_25);
 			cfg.setClassForTemplateLoading(ClusterResource.class, "templates");
 
 			cfg.setObjectWrapper(new ObjectToMapWrapper<Properties>(
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -1012,7 +1012,7 @@
    <dependency>
     <groupId>org.freemarker</groupId>
     <artifactId>freemarker</artifactId>
-    <version>2.3.18</version>
+    <version>2.3.25-incubating</version>
    </dependency>
    <dependency> 
      <groupId>net.sf.json-lib</groupId> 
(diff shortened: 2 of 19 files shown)


dalbani:
One of the unit tests now failing is at https://github.com/geoserver/geoserver/blob/master/src/wms/src/test/java/org/geoserver/wms/featureinfo/HTMLFeatureInfoOutputFormatTest.java#L249.
Expected `"(empty string)"`, found `"VALUE1,VALUE2,testLayer"`.
But I don't really understand why, because [test_content.ftl](https://github.com/geoserver/geoserver/blob/master/src/wms/src/test/resources/org/geoserver/wms/featureinfo/test_content.ftl) contains:
```
${request.ENV.TEST1},${request.ENV.TEST2},${request.LAYER}
```
Those values being defined [starting from line 96](https://github.com/geoserver/geoserver/blob/master/src/wms/src/test/java/org/geoserver/wms/featureinfo/HTMLFeatureInfoOutputFormatTest.java#L96).

src/pom.xml
```diff
@@ -1012,7 +1012,7 @@
    <dependency>
     <groupId>org.freemarker</groupId>
     <artifactId>freemarker</artifactId>
-    <version>2.3.18</version>
+    <version>2.3.25-incubating</version>
```
aaime: Incubating?

dalbani: Yeah, looked stranged to me at first, but nothing to be worried about: http://freemarker.org/freemarkerdownload.html#downloadApacheDisclaimer

src/community/jms-cluster/jms-geoserver/src/main/java/org/geoserver/cluster/impl/rest/ClusterResource.java
```diff
@@ -123,7 +123,7 @@ public JMSConfigHTMLFormat(Request request, Response response,
 
 		@Override
 		protected Configuration createConfiguration(Object data, Class clazz) {
-			Configuration cfg = new Configuration();
+			Configuration cfg = new Configuration(Configuration.VERSION_2_3_25);
```
aaime: I am checking the Freemarker documentation, setting this flag seems to enable a lot of non backwards compatible behavioral changes... this needs to be carefully evaluated and documented for people upgrading GeoServer, or better, if possible, not done at all unless strictly necessary.
Is it possible to get the html security improvements without throwing in also all these changes?

aaime:
@dalbani should I assume this pull request will not be updated? It's been stuck six months after my initial review.

dalbani:
With regard to your remark about (preferably) not changing the default `Configuration` version flag of Freemarker, I think there are only options here:
1. keep the existing templates untouched and change the version flag to >= 2.3.24 to enable auto-escaping;
2. change all the (HTML) templates shipped by GeoServer to explicitly escape variables and keep the Freemarker version (flag) as is.

The main difference between these 2 options concerns the user-made templates:
1. provides a secure-by-default template engine configuration, with the possibility of breakage due to encoding of values;
2. leaves the responsibility on the template creator / maintainer, who will need to learn (via documentation?) what he has to change in his templates.

From a security point of view, I prefer solution 1. but I understand your concern about non backwards compatible changes.

bencaradocdavies:
@dalbani this pull request now has conflicts. Please revise and resubmit.

aaime:
Closing due to lack of progress

PR #1982, opened 2016-11-18
[GEOS-7862] Add documentation for the WMTS multidimensional community module

nmco:
This pull requests adds documentation for the WMTS multidimensional community module.

Associated issue:
https://osgeo-org.atlassian.net/browse/GEOS-7862


nmco:
This failed is unrelated to this PR, it only adds documentation. The fail happens in the web processing module (is this normal or this module doesn't like me, I have always a fail related to this module ?).

PR #1984, opened 2016-11-18
[GEOS-7848] Proxy vulnerability

aaime:


src/wfs/src/main/java/org/vfny/geoserver/wfs/servlets/TestWfsPost.java
```diff
@@ -303,4 +316,24 @@ protected void processRequest(HttpServletRequest request, HttpServletResponse re
...
+                throw new IllegalArgumentException("Invalid url requested, the demo requests should be hitting: " + proxyBase);
+            }
+        }
+        
+        // use the requested url then, and remove the TestWfsPort
+        String requestedUrl = request.getRequestURL().toString();
+        // this should not happen, but let's not make it an open proxy if it does
+        if(!requestedUrl.endsWith(TEST_WFS_POST_PATH)) {
+            throw new IllegalStateException("Unepected, the TestWfsPost was accessed by a path not ending with TestWfsPost: " + requestedUrl);
+        }
```
jodygarnett: While I agree that this servlet has not moved in 10+ years - I am not sure I am getting why this check is required/useful? I was more focused on the urlString (since that is what we are opening up a connection to).

src/wfs/src/main/java/org/vfny/geoserver/wfs/servlets/TestWfsPost.java
```diff
@@ -303,4 +316,24 @@ protected void processRequest(HttpServletRequest request, HttpServletResponse re
...
+        String base = requestedUrl.substring(0, requestedUrl.lastIndexOf(TEST_WFS_POST_PATH));
+        if(!url.startsWith(base)) {
+            throw new IllegalArgumentException("Invalid url requested, the demo requests should be hitting: " + base);
+        }
```
jodygarnett: I like it, tricky but avoids having to know about port.

Could this be messed up when PROXY_BASE_URL above?

aaime: Uh you are probably right, best to put this part in the else of ``if(proxyBase != null)``
I'll make this quick change

PR #1985, opened 2016-11-18
[GEOS-7863] Responsive doc width + basic mobile support

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-7863

Supports a 640-740 px range for the doc body, plus a new narrow mode for widths below 875 px, where the sidebar (normally 200px) and other border content is hidden.

Also adds rudimentary mobile support based on this narrow mode


jodygarnett:
Thanks @tbarsballe this is good fun. Do we have a recommend image/figure width for the documentation guide?

tbarsballe:
We settled on 640px recommended width.

Wider images are OK, but don't go above 720px, and try to make sure they are still legible if rescaled to 640px

PR #1986, opened 2016-11-18
Adding docs for WCS request builder

bmmpxf:


jodygarnett:
Thanks @bmmpxf merging

PR #1989, opened 2016-11-19
Fix build after fast lookup changes, use harder to lookup CRSs

aaime:
Based on https://github.com/geotools/geotools/pull/1394, not sure if it will work without

### DIFF
--- a/src/main/src/main/java/org/geoserver/GeoserverInitStartupListener.java
+++ b/src/main/src/main/java/org/geoserver/GeoserverInitStartupListener.java
@@ -75,7 +75,7 @@ public class GeoserverInitStartupListener implements ServletContextListener {
 
     private final static String COMPARISON_TOLERANCE_PROPERTY = "COMPARISON_TOLERANCE";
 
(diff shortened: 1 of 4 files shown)


src/main/src/main/java/org/geoserver/GeoserverInitStartupListener.java
```diff
@@ -75,7 +75,7 @@
 
     private final static String COMPARISON_TOLERANCE_PROPERTY = "COMPARISON_TOLERANCE";
 
-    private final static double DEFAULT_COMPARISON_TOLERANCE = 1e-9;
+    private final static double DEFAULT_COMPARISON_TOLERANCE = 1e-8;
```
jodygarnett: Is there a reason to lower the tolerance?

aaime: The obvious one you already stated, there are a number of mismatches that are due to the too strict tolerance, lowering it allowed to increase the match ratio without causing (in my testing) false positives

aaime:
Merging to fix build after @jodygarnett merged the corresponding geotools pull request

PR #1996, opened 2016-12-01
[GEOS-7882] Fix status NPE when using custom GDAL

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-7882

### DIFF
--- a/src/extension/gdal/src/main/java/org/geoserver/gdal/GDALStatus.java
+++ b/src/extension/gdal/src/main/java/org/geoserver/gdal/GDALStatus.java
@@ -54,7 +54,7 @@ public class GDALStatus implements ModuleStatus {
 
     @Override
     public Optional<String> getMessage() {
-        String message = "JNI GDAL Wrapper Version: " + getGDALWrapperJarVersion();
+        String message = "JNI GDAL Wrapper Version: " + getGDALWrapperJarVersion().orElse("null");
         if (!isAvailable()) {
             message += "\njava.library.path: " + System.getProperty("java.library.path", "");
         } else {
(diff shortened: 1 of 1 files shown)


src/extension/gdal/src/main/java/org/geoserver/gdal/GDALStatus.java
```diff
@@ -68,9 +68,13 @@ public boolean isEnabled() {
         return Optional.ofNullable("");
     }
 
-    public String getGDALWrapperJarVersion() {
+    public Optional<String> getGDALWrapperJarVersion() {
         if (isAvailable()) {
-            return GeoTools.getVersion(gdal.class).toString();
+            Version v = GeoTools.getVersion(gdal.class);
+            if (v == null) {
+                return Optional.empty();
```
jodygarnett: So when does this occur? Probably when running from a development environment? I kind of thought you would use the GDAL bindings if available and report back the version of GDAL used.

DBlasby: I'm not sure when this would return null - gdal.class should be available as its part of the module that this class is in.  This might be an issue with the #getVersion() method or your deployment (i.e. if it's running directly on classes instead of a .jar).  

Looking at the implementation;
https://github.com/geotools/geotools/blob/master/modules/library/metadata/src/main/java/org/geotools/factory/GeoTools.java#L431

it will return null if the version is unavailable, so this should be fine.  

Are you running directly on .class files or does your jar have not have a "abc-version.jar" filename?

tbarsballe: This occurred in "normal" (not dev - running from built jars) environment with a custom build of GDAL.
Sadly, the GeoTools.getVersion call is not especially smart, and just inspects the jar filename for a version number (otherwise returning null).

tbarsballe: @DBlasby - The jar is just called "gdal.jar". I'm fairly sure the NPE was coming from the toString() call.

tbarsballe:
Backported to 2.10.x

PR #1999, opened 2016-12-04
Do not check stores when capabilites generation is set in tolerant mode

aaime:
This is a tentative, and incomplete, pull request disabling data store check on startup when the capabilities document generation is configured to skip over invalid layers. There is a test failure in geogig that seems to be related to code not being up to date (I also had a app-schema one, but that seems to be only local, I probably need to update code there).


aaime:
If there are no comments I'll clean it up (mostly adding docs) and commit

PR #2006, opened 2016-12-07
[GEOS-7890] Fix false negative in GeoServerInfoImpl.equals()

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-7890

### DIFF
--- a/src/main/src/main/java/org/geoserver/config/impl/GeoServerInfoImpl.java
+++ b/src/main/src/main/java/org/geoserver/config/impl/GeoServerInfoImpl.java
@@ -7,6 +7,7 @@ package org.geoserver.config.impl;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 
 import org.geoserver.catalog.MetadataMap;
 import org.geoserver.config.ResourceErrorHandling;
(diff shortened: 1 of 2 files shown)


src/main/src/main/java/org/geoserver/config/impl/GeoServerInfoImpl.java
```diff
@@ -343,7 +344,7 @@ public boolean equals(Object obj) {
         if (updateSequence != other.getUpdateSequence())
             return false;
       
-        if (globalServices != other.isGlobalServices())
+        if (!Objects.equals(globalServices, other.isGlobalServices()))
```
tbarsballe: Boolean

src/main/src/test/java/org/geoserver/config/GeoServerImplTest.java
```diff
@@ -44,6 +44,21 @@ public void testGlobal() throws Exception {
         geoServer.setGlobal( global );
        
         assertEquals( global, geoServer.getGlobal() );
+
+        //GEOS-7890
+        GeoServerInfo global1 = geoServer.getFactory().createGlobal();
+        GeoServerInfo global2 = geoServer.getFactory().createGlobal();
+        global1.setGlobalServices(new Boolean(true));
+        global1.setXmlExternalEntitiesEnabled(new Boolean(false));
```
jodygarnett: Oh dear it is a Boolean boolean, and you needed to do this XML stuff to get two Boolean.TRUE values in the system?

jodygarnett:
Horrified at the test case - good work!

src/main/src/test/java/org/geoserver/config/GeoServerImplTest.java
tbarsballe: XmlExternalEntities is just another Boolean value that I added to the test case in case of regressions

PR #2008, opened 2016-12-09
GEOS-7884 - Slow startup fixes

aaime:


ianturton:
looks good to me

aaime:
Hi @ianturton I actually have the eclipse code formatter setting from geotools loaded and it's not adding the extra braces. Checked, I cannot find a setting to do that. I wanted to use the formatter since the whole equals method in that class, along with others in the same package, has non fully braced elses. 
Not sure why, probably eclipse autogenerates it like that...

aaime:
Merging to keep this moving, but let's revisit the formatting issue, either here or on the mailing list.

PR #2010, opened 2016-12-12
Make WFS and WFS-T work properly with latest QGIS

aaime:


src/main/src/main/java/org/geoserver/config/CapabilitiesCacheHeadersCallback.java
```diff
@@ -0,0 +1,192 @@
...
+import org.geoserver.ows.Response;
+import org.geoserver.platform.GeoServerExtensions;
+import org.geoserver.platform.Operation;
+import org.geoserver.platform.ServiceException;
+import org.geotools.util.logging.Logging;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpStatus;
+
+/**
+ * Addds proper caching headers to capabilites response clients paying attention to HTTP headers do
```
bradh: Typo: Addds -> Adds

src/wfs/src/test/java/org/geoserver/wfs/TransactionTest.java
```diff
@@ -159,6 +159,29 @@ public void testInsert() throws Exception {
         assertEquals(2, dom.getElementsByTagName("gml:featureMember")
                 .getLength());
     }
+	
+	@Test
+    public void testInsertWithGetFeaturInThePath() throws Exception {
```
bradh: GetFeaturIn -> GetFeatureIn as a micro-improvement.

PR #2014, opened 2016-12-12
Fix test for ex schemaLocation includes [GEOS-7860]

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOS-7860

See discussion on geoserver-devel mailing list:
http://osgeo-org.1560.x6.nabble.com/app-schema-test-keeps-on-failing-locally-td5299512.html


aaime:
Yep, this fixed the build on my computer

bencaradocdavies:
Thanks for the confirmation @aaime. GeoSolutions OpenJDK is also fixed and winbuild has moved on to a different failure.

PR #2023, opened 2016-12-15
Adding Java Service Wrapper license to LICENSE.txt

bmmpxf:
The license adds a duplicate copy of GPLv2 (and an unnecessary copy of GPL v3), but I didn't alter the license at all.


aaime:
+1

tbarsballe:
This looks good.

To avoid duplication in the future, does it make sense to copy `LICENSE.txt` to `src/release/LICENSE.txt` during a build?

PR #2024, opened 2016-12-16
Fix the NPE message when requesting the WPS GetCapabilities

goudine:
Rather than 'java.lang.NullPointerException', the error message now contains a little bit more explanation.

### DIFF
--- a/src/extension/wps/wps-core/src/test/java/org/geoserver/wps/GetCapabilitiesTest.java
+++ b/src/extension/wps/wps-core/src/test/java/org/geoserver/wps/GetCapabilitiesTest.java
@@ -8,17 +8,10 @@ package org.geoserver.wps;
 import static junit.framework.Assert.assertEquals;
 import static junit.framework.Assert.assertTrue;
 import static org.custommonkey.xmlunit.XMLAssert.assertXpathEvaluatesTo;
-import static org.junit.Assert.*;
-
-import java.io.InputStream;
-
 import org.custommonkey.xmlunit.XMLUnit;
 import org.custommonkey.xmlunit.XpathEngine;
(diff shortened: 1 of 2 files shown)


src/extension/wps/wps-core/src/test/java/org/geoserver/wps/GetCapabilitiesTest.java
```diff
@@ -8,11 +8,17 @@
...
 import static org.custommonkey.xmlunit.XMLAssert.assertXpathEvaluatesTo;
+import static org.junit.Assert.*;
+
+import java.io.InputStream;
 
 import org.custommonkey.xmlunit.XMLUnit;
 import org.custommonkey.xmlunit.XpathEngine;
 import org.geoserver.config.GeoServer;
+import org.geoserver.config.util.XStreamPersister;
+import org.geoserver.config.util.XStreamPersisterFactory;
```
jodygarnett: is this still required

src/extension/wps/wps-core/src/test/java/org/geoserver/wps/GetCapabilitiesTest.java
```diff
@@ -40,6 +46,22 @@ public void testGetDisabledService() throws Exception {
             gs.save(info);
         }
     }
+    
+    @Test
+    public void testWPSScriptEnable() throws Exception {
```
jodygarnett: you may wish to rename this test; as you have not really managed to configure an invalid process group

jodygarnett:
Thanks @goudine - waiting for travis and then we can merge.

jodygarnett:
I think this is suitable for backporting; leave the ticket open as we still have the wps page to fix.

PR #2026, opened 2016-12-16
[GEOS-7905] Handle store display after REST upload.

bradh:
See https://osgeo-org.atlassian.net/browse/GEOS-7905 for background.

This probably addresses (part of) https://osgeo-org.atlassian.net/browse/GEOS-4616 too

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/data/store/panel/FileModel.java
+++ b/src/web/core/src/main/java/org/geoserver/web/data/store/panel/FileModel.java
@@ -7,6 +7,7 @@ package org.geoserver.web.data.store.panel;
 
 import java.io.File;
 import java.io.IOException;
+import java.net.URL;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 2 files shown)


aaime:
Looks good, lacks a test

bradh:
Added test that showed a problem with the original implementation, updated it.

src/web/core/src/main/java/org/geoserver/web/data/store/panel/FileModel.java
```diff
@@ -50,8 +51,14 @@ private boolean isSubfile(File root, File selection) {
         return isSubfile(root, selection.getParentFile());
     }
 
+    @Override
     public String getObject() {
-        return delegate.getObject();
+        Object obj = delegate.getObject();
```
aaime: I was looking at how the code could have possibly compiled before, looks like delegate is a IModel<String>, but in some cases it actually contains a URL? Which is however wrong since StoreInfo.getConnectionParameters returns a Map<String, Serializable>...

bradh: Indeed the "full" fix would be to change the REST code to insert a string instead of a URL. However I wasn't confident enough to find all the places where it potentially relied on the existing behaviour.

richb8:
I can confirm this fix is working.

tbarsballe:
Merged. Should this be backported?

bradh:
Backport is possible, but I don't need it.

PR #2030, opened 2016-12-16
[GEOS-7342] Filter layers and layer groups by workspace when adding to layer group

aaime:
Hi @tbarsballe I've added a small change that apparently succeeds in bringing back the chosen workspace to the server side. Could you double check, and if you like the change, merge?


aaime:
Nice... should we backport for the 2.10.1 release?

tbarsballe:
Yes; I am currently backporting.

aaime:
Great

PR #2032, opened 2016-12-16
Check if JAI-ext has been initialised in GeoServerInitStartupListener.

NielsCharlier:


aaime:
Ah, I was not aware that Nicola cloned that class in jai-ext... hum... isn't it better to empty the GeoServer class, deprecate it, and make it extend the jai-ext one instead? @dromagnoli what do you think?

aaime:
Uff.. the class in jai-ext is final, so the trick won't work without changes in jai-ext too...

PR #2035, opened 2016-12-16
[GEOS-7909] Upgrade SRS List demo to use OL3

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-7909

Converted OL2 preview to OL3
Also made some improvements to the bounds calculation used by the preview.

Tested with EPSG:4326, EPSG:3857, EPSG:2000 (Anguilla 1957 / British West Indies Grid), and EPSG:3031 (Antarctic Polar Stereographic). All render fine and start at a sensible zoom.

### DIFF
--- a/src/web/demo/src/main/java/org/geoserver/web/demo/SRSDescriptionPage.java
+++ b/src/web/demo/src/main/java/org/geoserver/web/demo/SRSDescriptionPage.java
@@ -5,9 +5,7 @@
  */
 package org.geoserver.web.demo;
 
-import java.util.Collection;
-import java.util.Locale;
-
+import com.vividsolutions.jts.geom.Envelope;
 import org.apache.wicket.Component;
 import org.apache.wicket.behavior.Behavior;
 import org.apache.wicket.markup.head.CssUrlReferenceHeaderItem;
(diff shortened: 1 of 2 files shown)


src/web/demo/src/main/java/org/geoserver/web/demo/SRSDescriptionPage.java
```diff
@@ -139,35 +156,25 @@ public void renderHead(Component component, IHeaderResponse response) {
...
-                    x1 = Math.min(dst1[0], dst2[0]);
-                    y1 = Math.min(dst1[1], dst2[1]);
-                    x2 = Math.max(dst1[0], dst2[0]);
-                    y2 = Math.max(dst1[1], dst2[1]);
+                    Envelope destEnvelope = JTS.transform(envelope, null, tr, 10);
+
+                    x1 = destEnvelope.getMinX();
+                    y1 = destEnvelope.getMinY();
+                    x2 = destEnvelope.getMaxX();
+                    y2 = destEnvelope.getMaxY();
```
jodygarnett: @aaime do you wish to double check the above?

PR #2036, opened 2016-12-17
[GEOS-7631] Set store name on REST upload.

bradh:
https://osgeo-org.atlassian.net/browse/GEOS-7631 provides the background. I took a
different approach to that suggest by the reporter to ensure keywords and title are
set up correctly too.


tbarsballe:
Looks good to me. Thanks for the fix.

Can this be backported?

bradh:
Its a minor change, so ambivalent about backporting.

PR #2039, opened 2016-12-20
[GEOS-7912] Layer group based security

aaime:
This is GSIP-152 implementation: https://github.com/geoserver/geoserver/wiki/GSIP-152


src/main/src/main/java/org/geoserver/security/impl/DataAccessRule.java
```diff
@@ -135,7 +163,7 @@ public String getValue() {
      * and if anything else is equal, read comes before write
      */
     public int compareTo(DataAccessRule other) {
-        int compareWs = compareCatalogItems(workspace, other.workspace);
+        int compareWs = compareCatalogItems(root, other.root);
```
bradh: Could rename `compareWS` to `compareRoot`

src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
```diff
@@ -156,14 +191,105 @@ public boolean canAccess(Authentication user, ResourceInfo resource, AccessMode
             // it's a layer whose resource we don't know about
             return true;
         }
-
+
```
bradh: Looks like unneeded whitespace.

src/main/src/test/java/org/geoserver/security/impl/AbstractAuthorizationTest.java
```diff
@@ -170,12 +195,30 @@ public void setUp() throws Exception {
...
         cascadedLayer = buildLayer("cascaded", toppWs, WMSLayerInfo.class);
         cascaded = (WMSLayerInfo) cascadedLayer.getResource();
     }
 
+    @After
+    public void cleanupRequestThreadLocal() throws Exception {
+        Dispatcher.REQUEST.remove();
+    }
+    
+    protected void setupRequesThreadLocal(String service) {
```
bradh: Rename `setupRequesThreadLocal` to `setupRequestThreadLocal`. Could this do both WFS and WMS service and have a @Before?

src/main/src/test/java/org/geoserver/security/impl/bothGroupABMilitaryOnly.properties
```diff
@@ -0,0 +1,3 @@
+# make tree A and B disappear to everybody but military
+namedTreeA.r=MILITARY
+containerTreeB.r=MILITARY
```
bradh: Missing newline at end of file.

src/main/src/test/java/org/geoserver/security/impl/containerTreeGroupBMilitaryOnly.properties
```diff
@@ -0,0 +1,2 @@
+# make container tree B disappear to everybody but military
+containerTreeB.r=MILITARY
```
bradh: Missing newline at end of file.

src/main/src/test/java/org/geoserver/security/impl/namedTreeAMilitaryOnly.properties
```diff
@@ -0,0 +1,2 @@
+# make named tree A (container) available only to military
+namedTreeA.r=MILITARY
```
bradh: Missing newline at end of file.

src/main/src/test/java/org/geoserver/security/impl/singleGroupCMilitaryOnly.properties
```diff
@@ -0,0 +1,2 @@
+# make group tree C (single) available only to military
+singleGroupC.r=MILITARY
```
bradh: Missing newline at end of file.

src/main/src/test/java/org/geoserver/security/impl/wsContainerGroupDMilitaryOnly.properties
```diff
@@ -0,0 +1,2 @@
+# make ws specific container tree D disappear to everybody but military
+nurc.wsContainerD.r=MILITARY
```
bradh: Missing newline at end of file.

src/main/src/test/java/org/geoserver/security/impl/wsContainerGroupDMilitaryOnly.properties
```diff
@@ -0,0 +1,2 @@
+# make ws specific container tree D disappear to everybody but military
```
bradh: "disappear" might be better as "invisible"

src/main/src/test/java/org/geoserver/security/impl/containerTreeGroupBMilitaryOnly.properties
```diff
@@ -0,0 +1,2 @@
+# make container tree B disappear to everybody but military
```
bradh: disappear might be better as "invisible"

src/main/src/test/java/org/geoserver/security/impl/bothGroupABMilitaryOnly.properties
```diff
@@ -0,0 +1,3 @@
+# make tree A and B disappear to everybody but military
```
bradh: disappear might be better as "invisible"

src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
```diff
@@ -156,14 +191,105 @@ public boolean canAccess(Authentication user, ResourceInfo resource, AccessMode
...
+        }
+        
+        // layer groups are used only in WMS
+        final String service = request.getService();
+        boolean isMapService = "WMS".equalsIgnoreCase(service) || "gwc".equalsIgnoreCase(service);
+        if(!isMapService) {
+            return false;
+        }
+        
+        return true;
```
nmco: The last six lines can be squashed into: return "WMS".equalsIgnoreCase(service) || "gwc".equalsIgnoreCase(service);

doc/en/user/source/security/layer.rst
```diff
@@ -14,16 +14,19 @@ Providing access to layers is linked to :ref:`roles <security_rolesystem_roles>`
 Rules
 -----
 
-The syntax for a layer security rule is as follows (``[]`` denotes optional
-parameters)::
+The syntax for a layer security rule can follow three different patterns (``[]`` denotes optional parameters)::
 
+  globalLayerGroup.permission=role[,role2,...]
   workspace.layer.permission=role[,role2,...]
+  workspace.workspaceLayerGroup.permission=role[,role2,...]
```
nmco: May be worth make it clear that a globalLayerGroup is just a a layerGroup without a workspace.

doc/en/user/source/security/layer.rst
```diff
@@ -14,16 +14,19 @@ Providing access to layers is linked to :ref:`roles <security_rolesystem_roles>`
...
-* ``layer``—Name of a resource (featuretype/coverage/etc...). The wildcard ``*`` is used to indicate all layers.
-* ``permission``—Type of access permission/mode. 
+* ``globalLayerGroup``— Name of a global layer group.
+* ``workspace``— Name of the workspace. The wildcard ``*`` is used to indicate all workspaces.
+* ``layer``— Name of a resource (featuretype/coverage/etc...). The wildcard ``*`` is used to indicate all layers.
+* ``workspaceLayerGroup``— Name of a workspace specific layer group
+* ``permission``— Type of access permission/mode. 
    
    * ``r``—Read access
    * ``w``—Write access
```
aaime: Nope, the code really does not make this consideration today, one has to remember to also state the read rule

src/main/src/test/java/org/geoserver/security/impl/AbstractAuthorizationTest.java
aaime: Nope, a request can only be in a single service at a given time, it's either WMS or WFS, cannot be both.

aaime:
Followed up with @bradh  and @nmco  review, plus fixed a bug Nuno reported to me directly that required quite some rework of the auth logic and some extra examples and clarifications in the documentation. Those are kept into two separate commits to allow easier review, I'll squash everything into one commit before merging.

src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
```diff
@@ -423,5 +657,95 @@ private boolean canAccess(Authentication user, SecureTreeNode node) {
...
+        public void handleModifyEvent(CatalogModifyEvent event) throws CatalogException {
+            final CatalogInfo source = event.getSource();
+            if(source instanceof LayerGroupInfo) {
+                LayerGroupInfo lg = (LayerGroupInfo) event.getSource();
+                // was the layer group renamed or moved?
+                int nameIdx = event.getPropertyNames().indexOf("name");
+                int wsIdx = event.getPropertyNames().indexOf("workspace");
+                if(nameIdx != -1 || wsIdx != -1) {
+                    // group has been modified, clear the old node (the new one will be updated in ppost modify)
+                    clearGroupInfo(lg);
```
nmco: I made the following test, created a global layer group and defined a rule for him. Then I updated the layer group adding him to an workspace. The created rule will still exists and reference the global layer group, not sure what should be done here.

doc/en/user/source/security/layer.rst
bradh: That would be worth adding a comment for. Suggest something like "Note that Write access (``permission`` of ``w``) does not imply read access."

src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
aaime: Wait, the rule is still there and that's normal, and "your fault", but the security subsystem should stop tracking the  catalog ids contained in that rule because it's now dangling. Is that not happening? I?ll check.

nmco: Agree, is up to the user to remove the rule and as far I can see indeed the security subsystem stop tracking the catalog ids contained in that rule. Thanks fro the clarification

nmco:
I made a review of this:
- [x] code
- [x] tests
- [x] documentation

This build locally with Java8, with JDK8 and OpenJDK8.

aaime:
Excellent, thanks for the detailed review. Merging

PR #2042, opened 2016-12-28
[GEOS-7918] Invalid scope attribute in gs-restconfig applicationContext.xml (master)

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOS-7918

### DIFF
--- a/src/restconfig/src/main/java/applicationContext.xml
+++ b/src/restconfig/src/main/java/applicationContext.xml
@@ -4,7 +4,7 @@
  This code is licensed under the GPL 2.0 license, available at the root
  application directory.
  -->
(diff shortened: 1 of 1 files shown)


bencaradocdavies:
An alternative to changing the DTD is to use ```singleton="false"```:
http://docs.spring.io/spring/docs/2.0.x/reference/beans.html#beans-factory-scopes

aaime:
+1

bencaradocdavies:
Thanks @aaime. Answering my own question, we have several other instances of later ```spring-beans``` DTD in the code base and no ```singleton="false"```, so for consistency I will merge this version.

PR #2044, opened 2017-01-02
GEOS-7921: Updates dispatcher to not use sendError when non error code http status.

jdeolive:
Rather it uses setStatus and writes content like a normal response.
Using sendError causes the servlet container to include it’s own
content indicating the request is an error, which throws off the
desired response.

### DIFF
--- a/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
+++ b/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
@@ -1634,14 +1634,25 @@ public class Dispatcher extends AbstractController {
             } else {
                 logger.log(Level.FINER, "", t);
             }
-            
+
+            boolean isError = ece.getErrorCode() >= 400;
+            HttpServletResponse rsp = request.getHttpResponse();
             try {
-                if(ece.getMessage() != null) {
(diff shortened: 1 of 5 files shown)


aaime:
Ticket? ;-)

src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
```diff
@@ -1634,14 +1634,24 @@ void exception(Throwable t, Service service, Request request) {
...
-                    request.getHttpResponse().sendError(ece.getErrorCode(),ece.getMessage());
-                } else {
-                    request.getHttpResponse().sendError(ece.getErrorCode());
+                if (isError) {
+                    if (ece.getMessage() != null) {
+                        rsp.sendError(ece.getErrorCode(), ece.getMessage());
+                    } else {
+                        rsp.sendError(ece.getErrorCode());
+                    }
+                }
```
aaime: Formatting :-p

aaime:
Looks good, found two minor issues

jdeolive:
Thanks for the fast review @aaime! Two issues should be fixed now.

aaime:
+1

PR #2048, opened 2017-01-09
[GEOS-7913] Opaque container layer group mode

aaime:


nmco:
Layers present in a opaque container and no present in any other group will still be available for WMS GetMap and GetFeature requests, i.e. there is no way to hide this layers for those operations. This is the case for logged and non logged users. Note, that the GetCapabilities document behavior is correct, i.e, the contained layers are not shown.

nmco:
Doubt, security rules applied to opaque container groups will not be considered right ? Basically they will always be considered public.

Lets say I have two groups:
- group1 -- opaque container -- layer1,layer2
- group2 -- named tree -- layer1,layer3

and I defined this security rules:
- *.*.r=*
- group1.r=ADMIN

Layer1 will be accessible by everyone, since the security rule is ignored.
Layer2 will not be accessible by anyone.
Layer3 will be accessible everyone.

This is the current behavior.

nmco:
Apart from the comments above this seems to work as expected. I was able to successfully build locally with JDK8 and OpenJDK8.

src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
```diff
@@ -192,85 +207,91 @@ public boolean canAccess(Authentication user, ResourceInfo resource, AccessMode
...
-        if(nodes.stream().anyMatch(n -> n.canAccess(user, mode) && n.getDepth() > catalogNodeDepth)) {
-            return true;
+        // there are groups, so there might be more specific rules overriding the catalog one, search for them
+        List<LayerGroupSummary> groupOverrides = containers.stream().filter(sg -> {
+            LayerGroupInfo gi = rawCatalog.getLayerGroup(sg.getId());
+            if(gi == null) {
+                return false;
+            }
+            SecureTreeNode node = getNodeForGroup(gi);
+            return (node != null && node.getDepth() > catalogNodeDepth) || (sg.getMode() == Mode.OPAQUE_CONTAINER);
```
nmco: This lines confuses me a bit, I don't understand why we have the ``sg.getMode() == Mode.OPAQUE_CONTAINER`` and the node depth check at the same time.

src/main/src/main/java/org/geoserver/security/impl/LayerGroupContainmentCache.java
```diff
@@ -0,0 +1,426 @@
...
+     */
+    static final Predicate<PublishedInfo> IS_LAYER = p -> p.getId() != null
+            && p instanceof LayerInfo && ((LayerInfo) p).getResource() != null;
+
+    /**
+     * Verifies a certain {@link PublishedInfo} is actually a {@link LayerGroupInfo}
+     */
+    static final Predicate<PublishedInfo> IS_GROUP = p -> p.getId() != null
+            && p instanceof LayerGroupInfo;
+
```
nmco: Nice one, the filter functions (IS_LAYER and IS_GROUP) make the code more readable :)

src/main/src/main/java/org/geoserver/security/impl/LayerGroupContainmentCache.java
```diff
@@ -0,0 +1,426 @@
...
+            addGroupInfo(lg);
+        }
+
+        // now populate the containment structure
+        for (LayerGroupInfo lg : groups) {
+            registerContainedGroups(lg);
+        }
+
+    }
+
```
nmco: So basically before processing any contained groups we need to be sure that they are already cached right ?

src/main/src/main/java/org/geoserver/security/impl/LayerGroupContainmentCache.java
```diff
@@ -0,0 +1,426 @@
...
+            }
+        });
+        // this group does not contain anything anymore, remove from containment
+        for (LayerGroupSummary d : groupCache.values()) {
+            d.containerGroups.remove(lg.getId());
+        }
+    }
+
+    /**
+     * Returns all groups containing directly or indirectly containing the resource
```
nmco: Returns all groups **containing** directly or indirectly **containing** the resource

src/main/src/main/java/org/geoserver/security/impl/LayerGroupContainmentCache.java
```diff
@@ -0,0 +1,426 @@
...
+        for (LayerGroupSummary lg : groups) {
+            collectContainers(lg, result);
+        }
+        return result;
+    }
+
+    /**
+     * Returns all groups containing directly or indirectly the specified group, and relevant for security (e.g., anything but
+     * {@link LayerGroupInfo.Mode#SINGLE} ones
+     *
```
nmco: So opaque container groups are relevant for security because they can actually hide a layer right ?
By the way the method above should probably also say that the single groups will not be considered since they are filtered in the ``collectContainers`` method :)

src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/GetFeatureInfoTest.java
```diff
@@ -960,4 +966,31 @@ public void testNotQueryableGroupLayer() throws Exception {
...
+           Document dom = getAsDOM(url);
+           //print(dom);
+           
+           // should not be found
+           XMLAssert.assertXpathEvaluatesTo("1", "count(/ServiceExceptionReport)", dom);
+           XMLAssert.assertXpathEvaluatesTo("layers", "//ServiceException/@locator", dom);
+           XMLAssert.assertXpathEvaluatesTo("LayerNotDefined", "//ServiceException/@code", dom);
+       }
+   }
 }
```
nmco: Manual testing shows that layers contained in an opaque container group are actually queryable, so wondering how this test can succeeded :astonished:.

src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/GetMapIntegrationTest.java
```diff
@@ -1169,4 +1173,31 @@ public void testFeatureIdMultipleLayers() throws Exception {
...
+                    + "&SERVICE=WMS&VERSION=1.1.1&REQUEST=GetMap&SRS=EPSG%3A4326&WIDTH=256&HEIGHT=256&BBOX=-0.0043,-0.0025,0.0043,0.0025";
+            Document dom = getAsDOM(url);
+            //print(dom);
+            
+            // should not be found
+            XMLAssert.assertXpathEvaluatesTo("1", "count(/ServiceExceptionReport)", dom);
+            XMLAssert.assertXpathEvaluatesTo("layers", "//ServiceException/@locator", dom);
+            XMLAssert.assertXpathEvaluatesTo("LayerNotDefined", "//ServiceException/@code", dom);
+        }
+    }
```
nmco: Same here.

aaime:
I've tried interactively, indeed I can still access the nested layers... until a restart/reload hits, at that point it behaves properly. Probably an issues with the containment cache catalog listener not doing its job on group mode switch.

aaime:
I just defined a rule like this:

````
*.*.r=*
tiger-ny.r=GROUP_ADMIN
mode=HIDE
````
and the tiger-ny group is not available anymore to anonymous users. So I cannot reproduce the second issue, the group and its layers are not available anymore

**Comment by nmco** [comment]
I was not clear on this one sorry, the group is not visible anymore but layer1 will be accessible to anyone I was just wondering if this is the expected behavior.

**Comment by aaime** [comment]
Given that the second group is accessible, yes, it's expected that layer1 is visible.

File: doc/en/user/source/data/webadmin/layergroups.rst [hunk 8]
```diff
@@ -9,9 +9,10 @@ A layer group also provides a consistent, fixed ordering of the layers it contai
 Layer Group modes
 -----------------
 
-Layer group behaviour can be configured by setting its :guilabel:`mode`. There are 4 available values:
+Layer group behaviour can be configured by setting its :guilabel:`mode`. There are 5 available values:
 
-* **single**: the layer group is exposed as a single layer with a name.
+* **single**: the layer group is exposed as a single layer with a name, acting as an alias for a list of layers. The layers are still showing up as top level entries in the WMS capabilities document (unless explicitly referred by a tree group).
```
aaime: If a layer is also accessible via a named, EO, or container group it's going to be accessible anyways. Same as with the previous group security logic, a layer is accessible if there is at least a non single group that makes it so.

doc/en/user/source/data/webadmin/layergroups.rst
```diff
@@ -9,9 +9,10 @@ A layer group also provides a consistent, fixed ordering of the layers it contai
...
+* **opaque container**: the layer group is exposed as a single layer with a name, acting as an alias for a list of layers. However, the layers and sub-groups contained in it won't show up in the capabilities document (unless explicitly referred by a tree group) and won't be available by themselves in WMS calls, but only  as part of the group.
```
aaime: Roger

src/main/src/main/java/org/geoserver/security/CatalogFilterAccessManager.java
```diff
@@ -94,6 +102,16 @@ public LayerGroupAccessLimits getAccessLimits(Authentication user, LayerGroupInf
...
     
+    @Override
+    public LayerGroupAccessLimits getAccessLimits(Authentication user, LayerGroupInfo layerGroup, List<LayerGroupInfo> containers) {
+        if (hideLayerGroup(layerGroup)) {
+            return new LayerGroupAccessLimits(CatalogMode.HIDE);
+        }
+        else {
+            return super.getAccessLimits(user, layerGroup, containers);
+        }
+    }
```
aaime: Roger

nmco:
Yes correct, I was fooled by the rules specificity.

aaime:
First batch of updates. @nmco can you try again with these changes and see if that residual issue is still there?

nmco:
Thanks, tested with the last updates everything seems to be working fine :tada:.

aaime:
Awesome!

PR #2052, opened 2017-01-12
[GEOS-7855] Fix for new style editor thinking old CSS styles are SLD

mattkrusz:
Adds logic to the AbstractStylePage and StyleEditPage to check for and recover CSS styles generated by the old (pre-pluggable styles) CSS extension. If a CSS style is found, it is only recovered it if the derived SLD has not subsequently been edited.


src/web/wms/src/main/java/org/geoserver/wms/web/data/AbstractStylePage.java
```diff
@@ -430,7 +432,59 @@ public void setRawStyle(Reader in) throws IOException {
...
+        } catch (Exception e) {
+            return;
+        }
+
+        // Use this tolerance to prevent erasing an SLD that was manually edited after being
+        // generated from a CSS file. (Generated SLDs will always be newer than the CSS).
+        long favorSLDIfNewerByMS = 600000;
+
+        // The problem only exists for styles that were not explicitly set as "css".
+        if (!"css".equalsIgnoreCase(si.getFormat())) {
```
tbarsballe: Can we modify this check so that only styles whose format is sld get updated?

If we have a YSLD style we probably don't want to convert it to CSS

src/web/wms/src/main/java/org/geoserver/wms/web/data/AbstractStylePage.java
```diff
@@ -430,7 +432,59 @@ public void setRawStyle(Reader in) throws IOException {
...
+            String filename = si.getFilename();
+            String filenameCss = filename.substring(0, filename.lastIndexOf('.')) + ".css";
+            GeoServerDataDirectory g = new GeoServerDataDirectory(getCatalog().getResourceLoader());
```
tbarsballe: Bit iffy on `g` as the var name here. Too many other things it could refer to.

src/web/wms/src/main/java/org/geoserver/wms/web/data/AbstractStylePage.java
```diff
@@ -430,7 +432,59 @@ public void setRawStyle(Reader in) throws IOException {
...
+            Resource cssResource = g.get(si, filenameCss);
+
+            if (!cssResource.getType().equals(Resource.Type.UNDEFINED)) {
+                // If there is an existing CSS file with the style's name, check if it should be recovered.
+                Resource sldResource = g.get(si, filename);
+                long sldNewerByMs = sldResource.lastmodified() - cssResource.lastmodified();
+
+                if (sldNewerByMs > favorSLDIfNewerByMS) {
+                    LOGGER.log(Level.WARNING,
+                            "A CSS version of this style has been recovered (" + filenameCss
```
tbarsballe: Rather than "this style", refer to the style by the StyleInfo name, since there will be no way of determining the current style when looking back at the logs.

e.g.
"A CSS version of the style OldCssStyle has been recovered ..."

If the style is in a workspace, the name should be prefaced by the workspace (<ws>:<name>) to ensure a unique name.

tbarsballe:
Overall looks very good.

A few fairly minor comments which need addressing, but once those are dealt with, it should be ready to merge.

src/web/wms/src/main/java/org/geoserver/wms/web/data/AbstractStylePage.java
mattkrusz: Yes, I think you're right. The old CSS extension created styles with no <format>, so I was thinking that the logic had to be as I wrote it. But on second thought, I believe the StyleInfo format gets defaulted to sld even if it is missing in the actual xml file. I'll make this change.

src/web/wms/src/main/java/org/geoserver/wms/web/data/AbstractStylePage.java
mattkrusz: Sure, easy to fix.

src/web/wms/src/main/java/org/geoserver/wms/web/data/AbstractStylePage.java
mattkrusz: Yes, agreed, that'd be much clearer.

src/web/wms/src/main/java/org/geoserver/wms/web/data/AbstractStylePage.java
tbarsballe: Yep, plus stuff like YSLD wasn't added until much more recently, so we don't want to be accidentally clobbering it.

mattkrusz:
Updated PR.

tbarsballe:
Ok, changes look good. Could you squash the two commits together into one and push back up to this PR before I merge? It makes it easier to track and backport changes for a single issue when they are all in one commit. (Multiple commits are of course easier when actually _reviewing_ a PR)

mattkrusz:
Sure -- squashed it down this morning.

tbarsballe:
Backported to 2.10.x

PR #2053, opened 2017-01-12
[GEOS-7926] Fix WMTS blind re-enabling

nmco:
Removes the WMTS service blind enabling in the initialization and fix the test that was not invoking the initialization.

Associated issue:
https://osgeo-org.atlassian.net/browse/GEOS-7926


nmco:
If nobody have nothing against this I will merge.

aaime:
Looks good. At first I though a service created during a data dir upgrade would have had the service disabled, but looking at the code it seems the ServiceImpl actually starts in enabled mode. May be worth having a test to ensure that, just in case.

aaime:
@nmco wasn't it you that someone who put in the initialization when you've made WMTS a top level GeoServer service?

PR #2058, opened 2017-01-18
GEOS-7933: Adding contentType to HttpErrorCodeException, 2.10.x backport

jdeolive:
(cherry picked from commit a1bebe6c5e58bb9c0ec8f28e0fa99b70a4474a8a)

### DIFF
--- a/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
+++ b/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
@@ -1637,6 +1637,10 @@ public class Dispatcher extends AbstractController {
 
             boolean isError = ece.getErrorCode() >= 400;
             HttpServletResponse rsp = request.getHttpResponse();
+
+            if (ece.getContentType() != null) {
+                rsp.setContentType(ece.getContentType());
+            }
             try {
                 if (isError) {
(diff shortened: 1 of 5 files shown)


jdeolive:
This backport doesn't pass travis but I think the failure is unrelated (in the NetCDF output module). Tried rebooting the build a few times... no luck. I think I'll merge this and just watch the main jenkins builds to see if they fail or not. If this does indeed introduce an issue I'll be sure to revert.

bencaradocdavies:
@jdeolive the failure was legit but no need to revert. The cause was [a fix for an unrelated GeoTools change](https://github.com/geoserver/geoserver/commit/2e38f7bbd3d43650e11b7000f4d00a549a4f387c) that was not backported to 2.10.x and 2.9.x when the [corresponding GeoTools change](https://github.com/geotools/geotools/commit/f9d146d944b17e1eb41795aaafbd6acbac7eaebe) was backported to 16.x and 15.x. I have cherry-picked the fix on to 2.10.x. and 2.9.x and expect these builds to be back to normal in under two hours.

jdeolive:
Thanks @bencaradocdavies! Had just noticed that the boundless jenkins had failed so was going to look into it. Thanks for the sleuthing and the fix!

PR #2062, opened 2017-01-23
[GEOS-7939]/[GSIP-154] - Add YSLD cookbook to GeoServer docs

tbarsballe:
To bring YSLD up to parity with SLD and CSS we would need a YSLD cookbook.

Boundless has donated the YSLD cookbook from https://github.com/boundlessgeo/suite/tree/master/docs/usermanual/source/cartography/ysld/cookbook (which was originally adapted from the GeoServer SLD cookbook.

When migrating this to geoserver, I have retained all history for transparency.


tbarsballe:
Updated to use images from sld cookbook, for efficiency and consistency.

tbarsballe:
Cherry-picked to 2.10.x

PR #2065, opened 2017-01-25
GEOS-7937 - Remove use of deprecated Files method in ResourcePool

dvntucker:
This one is a little tricky, I don't have a test right now because the actual conditions under which this is a problem are extremely specific:

- Using a non-default ResourceStore. When doing this no base directory gets set in GeoServerResourceLoader, which causes a few other things to fail when they start trying to munge things into a File based off the base directory.

- Using a custom GridFormat AND a custom DataStorePanelInfo implementation for it. We're testing a GridFormats that use things like S3 and HDFS internally right now, which have custom URL protocols (eg. s3://). We have a custom UI that skips all file-y validations for these URLs. However the ResourcePool still attempts to convert this to a file in order to pass to the GridReader.

Thankfully this is all fixed by using Resources.fromURL instead of the now-deprecated loader.fromURL. 

Unfortunately I don't have a test for this, but thankfully the test coverage for getGridCoverageReader is already quite good, and the change is moving from a deprecated method to its non-deprecated alternative, so maybe we can merge it anyway? If there are concerns I can look at writing a test for the specific bug being addressed.

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
+++ b/src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
@@ -20,6 +20,8 @@ import java.io.OutputStreamWriter;
 import java.io.Serializable;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -1484,17 +1486,13 @@ public class ResourcePool {
                     // Getting coverage reader using the format and the real path.
                     //
                     // /////////////////////////////////////////////////////////
-                    final String url = expandedStore.getURL();
-                    GeoServerResourceLoader loader = catalog.getResourceLoader();
-                    final File obj = loader.url(url);
-
-                    // In case no File is returned, provide the original String url
-                    final Object input = obj != null ? obj : url;  
+                    final String urlString = expandedStore.getURL();
+                    Object readObject = getObjectToRead(urlString);
 
                     // readers might change the provided hints, pass down a defensive copy
-                    reader = gridFormat.getReader(input, new Hints(hints));
+                    reader = gridFormat.getReader(readObject, new Hints(hints));
                     if(reader == null) {
-                        throw new IOException("Failed to create reader from " + url + " and hints " + hints);
+                        throw new IOException("Failed to create reader from " + urlString + " and hints " + hints);
                     }
                     if(key != null) {
                         if(hints != null) {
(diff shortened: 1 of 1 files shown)


aaime:
How hard is it to write a test? 
Mind, it seems the change broke a test in the importer module:

````
Tests run: 6, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 1.756 sec <<< FAILURE!
testPopulateEmptyMosaic(org.geoserver.importer.mosaic.ImporterMosaicTest)  Time elapsed: 65 sec  <<< ERROR!
java.lang.IllegalArgumentException: Contains invalid . path: 
	at org.geoserver.platform.resource.Paths.toPath(Paths.java:168)
	at org.geoserver.platform.resource.Paths.path(Paths.java:114)
	at org.geoserver.platform.resource.Paths.path(Paths.java:96)
	at org.geoserver.platform.resource.FileSystemResourceStore$FileSystemResource.get(FileSystemResourceStore.java:422)
	at org.geoserver.platform.resource.Resources.fromURL(Resources.java:580)
	at org.geoserver.platform.GeoServerResourceLoader.fromURL(GeoServerResourceLoader.java:187)
	at org.geoserver.platform.resource.Resources.fromURL(Resources.java:512)
	at org.geoserver.catalog.ResourcePool.getGridCoverageReader(ResourcePool.java:1488)
````

dvntucker:
Just saw the failure, not sure why I missed it when I ran tests.

dvntucker:
Actually, thinking about it this may be one of those instances where we *do* want to rely on the direct file system access rather than the Resources APIs, since most likely grid coverages are actual files and not typically stored along with GeoServer resources...

I may need to go back to the drawing board a bit on this one.

jodygarnett:
Curious about lack of "base directory" - even with JDBCStore we need a "base directory" for those rare api calls where a file has to be passed in; so the blob is saved out to the "base directory" (assumed to be a tmp folder) just so we can make the API call. 

I hope such calls are now rare - I think templates were some of the new remaining examples.

dvntucker:
Ok, I changed this one around a little. Instead of using the Resources code it checks the URL to see if it's a file or doesn't have a scheme and converts it to a file, otherwise it uses the String on its own. It's basically an explicit version of what the code was doing by happenstance anyway.

@aaime WRT how difficult it is to test the actual bug we ran into, it only happens under the following conditions:

- Not using the default file based store, instead using one such as JDBCStore.
- Using a GridFormat that relies on actual URLs rather than files (such as S3 or HDFS)
- Using a custom grid edit panel that skips the file check validations.

I'm certainly open to advice on testing it. Maybe by mocking out the ResoureStore implementation some how? But then I'm not sure how to inject that implementation in during a single test.

The good news is that it's a fairly minor refactor to a method that already has very good test coverage.

@jodygarnett It's possible that baseDirectory should be set for other resource store implementations... but, to what? I'd like to look at code that still relies on a directory because any code that explicitly requires file system access at this point shouldn't really be going through the ResourceStore, right?

aaime:
Looks good to me

tbarsballe:
Merged + Backported to 2.10.x

PR #2070, opened 2017-01-26
[GEOS-7944] Fix for null LegendInfo objects in StyleEditPage

mattkrusz:
When the AbstractStylePage is loaded with an existing style, it will now check if the StyleInfo object has a null LegendInfo. If so, it will insert a blank LegendInfo into the StyleInfo to act as the model for the ExternalGraphicPanel. This was already being done for new styles, but was missing for existing styles.

Added additional tests for discarding a legend on the StyleEditPage.

Also fixed another issue that was uncovered by this fix: Changed the “discard legend” functionality to fully reset user input — otherwise, supposedly discarded values could still fail to validate and prevent a save.


jodygarnett:
Thanks for the fix, and the very complete tests cases

jodygarnett:
Is this fix suitable for backport?

tbarsballe:
Looks to be backportable to 2.10.x. Merged and backported

PR #2071, opened 2017-01-26
[GEOS-7752] - Disabling the style name when editing a default style

dvntucker:
https://osgeo-org.atlassian.net/browse/GEOS-7752

Disabled renaming of default styles, since this causes problems elsewhere

### DIFF
--- a/src/web/wms/src/main/java/org/geoserver/wms/web/data/StyleAdminPanel.java
+++ b/src/web/wms/src/main/java/org/geoserver/wms/web/data/StyleAdminPanel.java
@@ -43,6 +43,9 @@ import org.apache.wicket.model.CompoundPropertyModel;
 import org.apache.wicket.model.IModel;
 import org.apache.wicket.model.Model;
 import org.apache.wicket.request.resource.AbstractResource;
+import org.apache.wicket.validation.IValidatable;
+import org.apache.wicket.validation.IValidator;
+import org.apache.wicket.validation.ValidationError;
 import org.geoserver.catalog.StyleGenerator;
 import org.geoserver.catalog.StyleInfo;
 import org.geoserver.catalog.StyleType;
@@ -107,6 +110,9 @@ public class StyleAdminPanel extends StyleEditTabPanel {
                 nameTextField.setEnabled(false);
                 uploadLink.setEnabled(false);
             }
+            if (StylePage.isDefaultStyle(getStylePage().getStyleInfo())) {
+                nameTextField.setEnabled(false);
+            }
             // format only settable upon creation
             formatChoice.setEnabled(false);
             formatReadOnlyMessage.setVisible(true);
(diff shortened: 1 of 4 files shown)


dvntucker:
Strange, seems to be a test failure but the test seemingly runs for me. Will investigate.

bencaradocdavies:
@dvntucker I can confirm the failure locally:
```
Tests run: 14, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 4.104 sec <<< FAILURE!
testCopyStyleFrenchLocale(org.geoserver.wms.web.data.StyleEditPageTest)  Time elapsed: 133 sec  <<< FAILURE!
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at org.geoserver.wms.web.data.StyleEditPageTest.testCopyStyleFrenchLocale(StyleEditPageTest.java:377)
```

bencaradocdavies:
In Eclipse, this method passes by itself but fails when the entire ``StyleEditPageTest``` suite is run.

bencaradocdavies:
If I ```@Ignore testRenameDefaultStyle``` then ```testCopyStyleFrenchLocale``` passes, so the failure depends on method run order to see the side-effect.

src/web/wms/src/main/java/org/geoserver/wms/web/data/StyleAdminPanel.java
```diff
@@ -107,6 +110,9 @@ public StyleAdminPanel(String id, AbstractStylePage parent) {
                 nameTextField.setEnabled(false);
                 uploadLink.setEnabled(false);
             }
+            if (StylePage.isDefaultStyle(getStylePage().getStyleInfo())) {
```
aaime: The ticket correctly suggested to disable workspace assignment as well no?

src/web/wms/src/main/java/org/geoserver/wms/web/data/StylePage.java
```diff
@@ -79,12 +79,7 @@ protected Component headerPanel() {
         header.add(removal = new SelectionRemovalLink("removeSelected", table, dialog) {
             @Override
             protected StringResourceModel canRemove(CatalogInfo object) {
-                StyleInfo s = (StyleInfo) object;
-                if ( StyleInfo.DEFAULT_POINT.equals( s.getName() ) || 
-                    StyleInfo.DEFAULT_LINE.equals( s.getName() ) || 
-                    StyleInfo.DEFAULT_POLYGON.equals( s.getName() ) || 
-                    StyleInfo.DEFAULT_RASTER.equals(s.getName()) ||
-                    StyleInfo.DEFAULT_GENERIC.equals(s.getName())) {
+                if (isDefaultStyle(object)) {
```
aaime: Wondering about the REST API comment made in the ticket, and the possibility of moving this check down into the catalog.

dvntucker: Moving it to the catalog probably makes sense, would that fix it for the REST API as well? I didn't want to worry too much about the rest API, since I wanted to at least get this improvement out of the way and there are already plenty of differences between what the REST API does and the UI does.

src/web/wms/src/main/java/org/geoserver/wms/web/data/StyleAdminPanel.java
dvntucker: You're right, I'll take a look at that.

src/web/wms/src/main/java/org/geoserver/wms/web/data/StylePage.java
aaime: No big deal, as said, the pull request is a valid improvement as is, just make clear in Jira what actually has been modified

tbarsballe:
Thanks for the fix, so we don't lose track of the other issues can we get a seperate JIRA ticket for not allowing default styles to be workspaced?

Also, can this be squashed down to a single commit with the ticket number in the commit message?

tbarsballe:
Oh, and this will also need to be rebased against master

PR #2072, opened 2017-01-26
[GEOS-7771] [GEOS-7873] Upgrade to jettison 1.2

tbarsballe:
While working on https://osgeo-org.atlassian.net/browse/GEOS-7873, I revisited my fix for GEOS-7771 and discovered that GeoServer is using jettison 1.0.1, rather than 1.2. Upgrading jettison fixed GEOS-7771 and GEOS-7873 without requiring any workarounds.

This did result in a slight change in JSON syntax for representing arrays, and as such should probably not be backported. (See ResourceTest.java for an example of the syntax change - effectively, objects containing single-element arrays are now single-element arrays containing objects. This appears to occur whenever a List is encoded to JSON.

### DIFF
--- a/src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
+++ b/src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
@@ -1163,8 +1163,6 @@ public class XStreamPersister {
                     writer.addAttribute("type", typeName);
                 }
                 context.convertAnother( item, new ReferenceConverter( clazz ) );
-            } else if (writer instanceof JettisonStaxWriter) {
-                writer.setValue("null");
             }
             writer.endNode();
         }
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -655,7 +655,7 @@
    <dependency>
     <groupId>org.codehaus.jettison</groupId>
     <artifactId>jettison</artifactId>
-    <version>1.0.1</version>
+    <version>1.2</version>
    </dependency>
    <dependency>
     <groupId>lucene</groupId>
--- a/src/restconfig/src/test/java/org/geoserver/catalog/rest/CoverageStoreTest.java
+++ b/src/restconfig/src/test/java/org/geoserver/catalog/rest/CoverageStoreTest.java
@@ -62,7 +62,7 @@ public class CoverageStoreTest extends CatalogRESTTestSupport {
         JSON json = getAsJSON( "/rest/workspaces/wcs/coveragestores.json");
         assertTrue( json instanceof JSONObject );
         
-        Object coveragestores = ((JSONObject)json).getJSONObject("coverageStores").get("coverageStore");
+        Object coveragestores = ((JSONObject)json).getJSONArray("coverageStores").getJSONObject(0).get("coverageStore");
         assertNotNull( coveragestores );
         
         if( coveragestores instanceof JSONArray ) {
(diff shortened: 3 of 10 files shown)


tbarsballe:
I agree the transposition is somewhat scary (and all the other test changes  in the commit are manifestations of that change). I expect it is mostly an artifact of converting from XStream to JSON, and I think the new structure is a bit better than the old one.

aaime:
Won't this require changing the documentation examples as well?
If I read correctly, does the client code have to handle a "object or array" case nowadays? If so, it would seems like using a fixed choice would be an improvement, although it's indeed going to break someone, worth mentioning it in the release notes.

tbarsballe:
As far as I can find, we don't actually have any JSON documentation examples (which is a separate problem in itself, but makes this easy).
I'm not quite sure I understand your question about handling an "object or array" case?

* The Java libraries have always treated JSONObject and JSONArray differently
* Other client code would generally have to deal with objects separately from arrays, yes.

I agree, this change should definitely be noted in the release notes.

aaime:
Maybe I misunderstood "objects containing single-element arrays are now single-element arrays containing object". It looked like a client would have got an object if the array had one entry, and an array if it contained more than one entry, and now it's always an json array instead?

tbarsballe:
No, not quite:

Before the change, the client would get an object containing an array: {[...]}
After the change, the client gets and array containing an object: [{...}]

In both cases, the array always has a size of one.
See ResourceTest for an example of this in practice: https://github.com/geoserver/geoserver/pull/2072/files#diff-53d3ea4d2b8d4e1363c0415f421cdda9

aaime:
Ouch, yeah, not backwards compatible, needs to be advertised well.
Many times people have to upgrade to get a security fix but cannot upgrade clients too because of that... is there any way to configure the new jettison to behave like the old one, like, not by default, but on demand, setting a system variable or something like that?

tbarsballe:
I highly doubt there is a way to make this work. 
When initially looking to upgrade, I found some discussions that noted that the jettison 1.2 JSON output will be slightly different from the jettison 1.0.1 output (no mention of workarounds)
Additionally, the jettison documentation is poor and virtually nonexistent (i've mostly just found javadocs).

I expect any manual way of making this work from within GeoServer would wind up as a pile of hacky fixes, to the point where we would be better off just forking jettison (which could be a viable solution, but is likely nontrivial).
If we would rather stick with jettison 1.0.1 and keep its current behavior, I do have an alternative fix to GEOS-7873 [here](https://github.com/geoserver/geoserver/compare/master...tbarsballe:GEOS-7873-reflect#diff-68b4cc50e6268484493c72df83cb5301R1175), but it is really ugly.

The jettison project seems to be poorly supported overall, and at this point looks to be mostly dead. I've found their [github](https://github.com/codehaus/jettison), and it hasn't seen any commits in 2 years

aaime:
Oh well, if there is no way, there is no way.

jodygarnett:
Can we add a documentation example for the above json story, and make a note for those upgrading from GeoServer 2.10.x - would that allow us to continue?

I am not that stuck up on fixing it properly; but I have seen the evil that is @tbarsballe fix - and reflection to relax class field permissions is an extra level of evil I would not expect to work on different JVM implementations (such as IBM).

With that in mind we have a choice between invalid JSON (eek) or breaking backwards compatible (see fixed test cases).

Can we do this for the 2.11 release; and update gsconfig and geoserver-manager at the same time? I do not know how sophisticated those two libraries are (do they use xml or json). 

Going to take this discussion to the email list...

tbarsballe:
Discussion moved to email list. Proposal created: https://github.com/geoserver/geoserver/wiki/GSIP-156

While putting this proposal together, I discovered the encoding changes are slightly more complicated than just transposing arrays. Instead, top-level list objects now get wrapped in  single-element arrays, and single element child arrays now get unwrapped as objects. For examples, see the proposal.

tbarsballe:
Closing in favor of https://github.com/geoserver/geoserver/pull/2084

This can be re-visited if/when we upgrade to Jettison 1.2

PR #2073, opened 2017-01-27
GEOS-7941:WfsCompatibilityTest fails if a local GeoServer is running without the sf:archsites layer

vickdw:
Added try/catch clause to the test to catpure the IOException being thrown when the Schema isn't present.

### DIFF
--- a/src/main/src/test/java/org/geoserver/wfsng/WfsCompatibilityTest.java
+++ b/src/main/src/test/java/org/geoserver/wfsng/WfsCompatibilityTest.java
@@ -67,12 +67,19 @@ public class WfsCompatibilityTest extends GeoServerSystemTestSupport {
         }
                 
         assertTrue(store instanceof WFSDataStore);
+        
+        try{
+            FeatureType type = ftInfo.getFeatureType();
+            
+            assertEquals("sf_archsites", type.getName().getLocalPart());        
+            
+            assertEquals("sf_archsites", ftInfo.getFeatureSource(null, null).getName().getLocalPart());
+        }catch(IOException e){
+            String expectedMessage = "Schema 'sf_archsites' does not exist.";
+            assertEquals("Exception message must be correct", expectedMessage, e.getMessage());
+        }
                 
-        FeatureType type = ftInfo.getFeatureType();
         
-        assertEquals("sf_archsites", type.getName().getLocalPart());        
-        
-        assertEquals("sf_archsites", ftInfo.getFeatureSource(null, null).getName().getLocalPart());
 
     }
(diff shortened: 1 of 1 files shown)


src/main/src/test/java/org/geoserver/wfsng/WfsCompatibilityTest.java
```diff
@@ -67,12 +67,19 @@ public void testWfsCompatitibility() throws IOException {
         }
                 
         assertTrue(store instanceof WFSDataStore);
-                
-        FeatureType type = ftInfo.getFeatureType();
         
-        assertEquals("sf_archsites", type.getName().getLocalPart());        
+        try{
```
aaime: Still not following formatting rules, it should be "try {" and "} catch(...) {", no need to add extra white lines in the block instead

vickdw: so, like this?
`try{
            FeatureType type = ftInfo.getFeatureType();
            assertEquals("sf_archsites", type.getName().getLocalPart());
            assertEquals("sf_archsites", ftInfo.getFeatureSource(null, null).getName().getLocalPart());
        } catch(IOException e){
            String expectedMessage = "Schema 'sf_archsites' does not exist.";
            assertEquals("Exception message must be correct", expectedMessage, e.getMessage());
        }`

src/main/src/test/java/org/geoserver/wfsng/WfsCompatibilityTest.java
```diff
@@ -67,12 +67,14 @@ public void testWfsCompatitibility() throws IOException {
...
                 
         assertTrue(store instanceof WFSDataStore);
-                
-        FeatureType type = ftInfo.getFeatureType();
-        
-        assertEquals("sf_archsites", type.getName().getLocalPart());        
         
-        assertEquals("sf_archsites", ftInfo.getFeatureSource(null, null).getName().getLocalPart());
-
+        try{
```
aaime: Still wrong, there is a space between try and the curly brace, same at the end of the catch exception. As a general rule, always have a space before and/or after curly braces

aaime:
Yes, looking good. I've manually merged it to squash the 3 commits into one, and make the commit message follow the usual "jira title" pattern (jira code plus title).

aaime:
Squashred commit here 2673e472ec8c2a82c93dd59bd4bb64aacefc2b1b

aaime:
Backported on 2.10.x too, on 2.9.x the test apparently does not exist

PR #2075, opened 2017-01-27
[GEOS-7930] Prevent overwrite of updated styles when also changing workspace

mattkrusz:
https://osgeo-org.atlassian.net/browse/GEOS-7930

Change the StyleEditPage submit function to first save the StyleInfo, then save the updated style body. When done in the opposite order, updates to the style body would be overwritten if the style was also moved.

Also added a test that isolates the bug by updating and moving a style at the same time.

### DIFF
--- a/src/web/wms/src/main/java/org/geoserver/wms/web/data/StyleEditPage.java
+++ b/src/web/wms/src/main/java/org/geoserver/wms/web/data/StyleEditPage.java
@@ -108,18 +108,21 @@ public class StyleEditPage extends AbstractStylePage {
                 style.setLegend(null);
             }
 
-            // write out the SLD
+            // save the updated StyleInfo
+            getCatalog().save(style);
+
+            // save the updated style contents to the new file location
             try {
-                getCatalog().getResourcePool().writeStyle(style, 
+                getCatalog().getResourcePool().writeStyle(style,
                         new ByteArrayInputStream(rawStyle.getBytes()));
             } catch (IOException e) {
                 throw new WicketRuntimeException(e);
             }
-            getCatalog().save(style);
+
             styleForm.info("Style saved");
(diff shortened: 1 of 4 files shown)


src/web/wms/src/test/java/org/geoserver/wms/web/data/StyleEditPageChangeWorkspaceTest.java
```diff
@@ -0,0 +1,94 @@
...
+    protected void onSetUp(SystemTestData testData) throws Exception {
+        super.onSetUp(testData);
+        testData.addStyle(TEST_STYLE_NAME, TEST_STYLE_FILENAME, this.getClass(), getCatalog());
+    }
+
+    /**
+     * Test that a user can update the .sld file contents and move the style into a workspace in a single edit.
+     */
+    @Test
+    public void testMoveWorkspaceAndEdit() throws Exception {
```
tbarsballe: Given that this is just a single test, it might be better of in StyleEditPageTest.java rather than a separate file. Its probably fine as-is though

tbarsballe:
Looks good.

src/web/wms/src/test/java/org/geoserver/wms/web/data/StyleEditPageChangeWorkspaceTest.java
aaime: Yeah, given the time it takes to setup the integration tests of a test it's often a good idea to add a test in a pre-existing test class unless it's really out of place. Not a major issue mind, but keep in mind for the future.

mattkrusz:
Moved the new test case into the existing integration test & squashed the commit.

tbarsballe:
Merged. Should this be backported?

PR #2076, opened 2017-01-28
DAEMON to follow JAVA_HOME from config var file

thinknirmal:
Made changes to Debian init.d file so the `$DAEMON` uses custom `$JAVA_HOME` specified in config variable file `/etc/default/$NAME`.


jodygarnett:
This a doc change, thanks for the contribution.

PR #2077, opened 2017-01-28
[GEOS-7953] GSIP 155 - Core improvements for catalogs with many layers, follow up on the write side

aaime:
The other changes helped some in terms of build time on Travis (it's a bit shorter than usual, couple of minutes), so I prepared a patch mirroring the Resource.getContents() adding a Resource.setContents(byte[]). Let's see if this helps too.


aaime:
Meh, does not look like it helps speed wise. Probably overkill then.

jodygarnett:
+1 I would like to merge for API consistency

aaime:
Done :-p

PR #2079, opened 2017-01-31
GEOS-7960: Factoring out GeoServerNodeData class from GeoServerNodeInfo.

jdeolive:
Issue link - https://osgeo-org.atlassian.net/browse/GEOS-7960

### DIFF
--- a/src/main/src/main/java/org/geoserver/GeoServerNodeData.java
+++ b/src/main/src/main/java/org/geoserver/GeoServerNodeData.java
@@ -1,0 +1,170 @@
+/* (c) 2017 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver;
+
+import org.geoserver.ows.util.KvpUtils;
+import org.geoserver.platform.GeoServerExtensions;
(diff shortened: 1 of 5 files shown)


src/main/src/main/java/org/geoserver/GeoServerNodeData.java
```diff
@@ -0,0 +1,168 @@
+/* (c) 2017 Open Source Geospatial Foundation - all rights reserved
+ */
```
jodygarnett: Header is usually longer - http://docs.geoserver.org/latest/en/developer/policies/committing.html

jdeolive: Oops, my bad, will fix.

jdeolive:
@jodygarnett Are you happy with the changes?

jodygarnett:
Go for it; I should ask if there is a need to backport this for your customer, or are you content with the 2.11 release timeframe.

jdeolive:
Thanks @jodygarnett. If possible I would like to backport this one but it is technically an api change... so not sure what the policies are there.

jodygarnett:
Ask on the email list, usually backport is okay for GeoServer (after a one month cooling off period).

PR #2080, opened 2017-02-01
[GEOS-7962] Add a WMS FORMAT vendor parameter

nmco:
This pull request implements the support for the WMS ``format`` vendor parameter. WMS 1.1.1 capabilities document supported mime types were extend to include the ``text/xml`` mime type. Tests and documentation were also updated.

Associated issue: 
- https://osgeo-org.atlassian.net/browse/GEOS-7962

Mailing list discussion: 
- http://osgeo-org.1560.x6.nabble.com/Add-a-WMS-vendor-FORMAT-parameter-to-force-the-return-of-WMS-1-1-1-capabilities-in-text-xml-td5305506.html

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformer.java
+++ b/src/wms/src/main/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformer.java
@@ -84,8 +84,6 @@ import org.xml.sax.Attributes;
 import org.xml.sax.ContentHandler;
 import org.xml.sax.helpers.AttributesImpl;
 
-import sun.security.action.GetBooleanAction;
-
 import com.google.common.collect.Iterables;
 import com.vividsolutions.jts.geom.Envelope;
 import org.geoserver.wfs.json.JSONType;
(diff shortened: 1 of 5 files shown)


src/wms/src/main/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformer.java
```diff
@@ -100,8 +98,13 @@
  *      org.geoserver.platform.Operation)
  */
 public class GetCapabilitiesTransformer extends TransformerBase {
-    /** fixed MIME type for the returned capabilities document */
-    public static final String WMS_CAPS_MIME = "application/vnd.ogc.wms_xml";
+    /** default MIME type for the returned capabilities document */
+    public static final String WMS_CAPS_DEFAULT_MIME = "application/vnd.ogc.wms_xml";
```
jodygarnett: I was under the impression the different WMS specifications have different defaults (example WMS 1.3.0 has text/xml" as a default ... 

Okay I see there is a Capabilities_1_3_0_Transformer (which does not have this format option).

jodygarnett:
So this pull requests includes docs, test case, etc ...

I was wondering if this format option could be done once up a level in the request / response framework allowing the mime type to be overridden for all requests (GetCapabilities, GetFeatureInfo, DescribeFeatureType others...). The point would to be to do things consistently.

I also wanted to ask about the relationship between this setting and the gzip compression settings in web.xml?

aaime:
Hi @jodygarnett doing it at a higher level makes little sense, as you'd need first to establish a relationship between the target Response object mime type and "text/xml" so that you know they are indeed the same thing before choosing which Response object to pick (in other words, it would be significant work and risk breaking the OWS Dispatcher).

About gzip compression, there is no relationship, here we are talking about format, a OGC concept, while gzip compression is a HTTP protocol concept that is travelling over HTTP headers in both advertisemebt of support, request for encoding, and confirmation that gzip compression happened. The only thing to make sure is that "text/xml" is part of the gzip compressible whitelist in web.xml (Nuno, is it?)

nmco:
Is ``text/xml`` is part of GZIP white list:
```
<param-value>text/.*,.*xml.*,application/json,application/x-javascript</param-value>
```

PR #2084, opened 2017-02-03
[GEOS-7873] Workaround for JSON array serialization of nulls

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-7873

Alternative to https://github.com/geoserver/geoserver/pull/2072 that doesn't alter the rest API output, but instead uses reflection to hack around the jettison 1.0.1 bug.

### DIFF
--- a/src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
+++ b/src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
@@ -10,6 +10,9 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


jodygarnett:
Thanks for talking this through on the mailing list, this is indeed a terrible (temporary) workaround.

jodygarnett:
Do you need to backport this?

tbarsballe:
This should be backported back to 2.10.x, to match up with the GEOS-7771 fix. I will backport it.

src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
```diff
@@ -1164,7 +1172,26 @@ protected void writeItem(Object item, MarshallingContext context,
...
+                    Field outField = StaxWriter.class.getDeclaredField("out");
+                    Field nodesField = MappedXMLStreamWriter.class.getDeclaredField("nodes");
+                    outField.setAccessible(true);
+                    nodesField.setAccessible(true);
+                    FastStack nodes = (FastStack) nodesField.get(outField.get(writer.underlyingWriter()));
+                    if (nodes.peek() instanceof JSONArray) {
+                        nodes.pop();
+                    }
+                } catch (IllegalAccessException e) {
+                    e.printStackTrace();
```
aaime: Ouch, did we really merge something writing stack traces to the standard error instead of using logging? Checked the code, indeed we did :-(

tbarsballe: Oops. I'll fix that...

tbarsballe: Fix pending: https://github.com/geoserver/geoserver/pull/2395

PR #2092, opened 2017-02-07
Fixed display info of memory usage inside server status page

lavoiej:
When you start GeoServer with -Xmx higher than 1g (for e.g: 2g), the memory usage inside the "Server status" page displays the wrong number.

With -Xmx2g before my fix :
before

With -Xmx2g after my fix :
after

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/admin/StatusPanel.java
+++ b/src/web/core/src/main/java/org/geoserver/web/admin/StatusPanel.java
@@ -267,7 +267,7 @@ public class StatusPanel extends Panel {
 
         String formattedUsedMemory;
         if (bytes > GB) {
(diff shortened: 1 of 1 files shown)


lavoiej:
No, you don't need to do that, unless you want to be really precise (for e.g.: 219.20 MB / 500.24 MB).

tbarsballe:
Hello @lavoiej , thanks for your contribution.

Have you signed a [Contributor License Agreement](https://github.com/geoserver/geoserver/blob/master/CONTRIBUTING.md#contributing) for GeoServer yet? If not, could you please do so? 

This fix is small enough that we can accept it on your behalf without a CLA, but it would still be appreciated. Thanks.

aaime:
Yep, for this kind of changes a CLA is not actually needed

PR #2094, opened 2017-02-07
Pdf optimization for tile patterns

aaime:


jodygarnett:
Is this suitable for backport?

aaime:
@jodygarnett yep, I can backport flipping the default value of the "org.geoserver.pdf.encodeTilingPatterns" variable so that it's available on 2.10.2 for those that want to try it, but not on by default and affecting everybody.

src/wms/src/main/java/org/geoserver/wms/WMSMapContent.java
```diff
@@ -418,4 +418,12 @@ public String getRendererScaleMethod() {
         return metadata;
     }
     
+    @Override
+    public void dispose() {
+        this.request = null;
+        this.callbacks = null;
+        this.metadata = null;
```
jodygarnett: looks great

src/wms/src/main/java/org/geoserver/wms/map/PDFMapResponse.java
```diff
@@ -47,6 +70,11 @@
     /** A logger for this class. */
     private static final Logger LOGGER = org.geotools.util.logging.Logging
             .getLogger("org.vfny.geoserver.responses.wms.map.pdf");
+    
+    /**
+     * Whether to apply the new vector hatch fill optimization, or not (on by default, this is just a safeguard)
+     */
```
jodygarnett: Can we document the system property in the docs; or even in the javadoc?

src/wms/src/main/java/org/geoserver/wms/map/PDFMapResponse.java
```diff
@@ -225,4 +258,159 @@ public void write(Object value, OutputStream output, Operation operation) throws
             throw new ServiceException("Error setting up the PDF", t, "internalError");
         }
     }
+    
+    private static class PDFStreamingRenderer extends StreamingRenderer {
+        
+        public PDFStreamingRenderer() {
+            this.painter = new PDFStyledPainter(labelCache);
```
jodygarnett: all of this to set a field, setPainter is not avaialble?

src/wms/src/main/java/org/geoserver/wms/map/PDFMapResponse.java
```diff
@@ -225,4 +258,159 @@ public void write(Object value, OutputStream output, Operation operation) throws
...
+    
+    /**
+     * Optimized StyledShapePainter that can optimize painting repeated external graphics
+     * by encoding them in a native PDF pattern painter, with no repetition 
+     */
+    private static class PDFStyledPainter extends StyledShapePainter {
+        
+        /**
+         * Re-creating the patters is expensive and would increase the overall
+         * document size, keep a cache of them instead
```
jodygarnett: typo above

src/gwc/src/main/java/org/geoserver/gwc/layer/DefaultTileLayerCatalog.java
```diff
@@ -109,22 +109,22 @@ public boolean accept(Resource res) {
...
         LOGGER.info("Loading tile layers from " + baseDir.path());
-        for (Resource res : tileLayerFiles) {
+        tileLayerFiles.parallelStream().forEach(res -> {
             GeoServerTileLayerInfoImpl info;
             try {
                 info = depersist(res);
             } catch (Exception e) {
                 LOGGER.log(Level.SEVERE, "Error depersisting tile layer information from file "
                         + res.name(), e);
-                continue;
```
jodygarnett: So was this just a programming mistake?

src/wms/src/main/java/org/geoserver/wms/map/PDFMapResponse.java
aaime: That is javadoc already no?

aaime: About the docs... unsure, the property is meant only as a safeguard, and there are a lot similarly undocumented.

src/wms/src/main/java/org/geoserver/wms/map/PDFMapResponse.java
aaime: Not available, not should not be made available. The painter is not supposed to be replaced, as you can see from the code this is really an advanced change (e.g., you exactly know what you're doing, or boom). So best if it's hard and not very visible, I don't want to give users ideas about it.

src/wms/src/main/java/org/geoserver/wms/map/PDFMapResponse.java
aaime: Ah, "patters". Yes, will fix.

src/gwc/src/main/java/org/geoserver/gwc/layer/DefaultTileLayerCatalog.java
aaime: Gah sorry, this really needs to be removed (long builds, I was playing with parallelizing the DefaultTileLayerCatalog loading with streams, and this change is a leftover. Will remove.

aaime:
Rebased and cleaned up the commits.

src/wms/src/main/java/org/geoserver/wms/map/PDFMapResponse.java
jodygarnett: the system property is in the method, not the javadoc.

You are correct that there is a lot of undocumented system properties in the codebase!  Tempted to make a foss4g presentation "geoserver terrors" to cover them all.

Please go ahead I will stop complaining about magic system properties.

jodygarnett:
Excellent I will merge after checks pass; and you are going to backport with default flipped before the stable release right?

aaime:
Yep, I'll backport with flipped system property. About system properties, I'd really keep them non advertised, our user base is often not very technical and they have more than enough rope to hang themselves already.

jodygarnett:
merged to master, backport when ready

src/wms/src/main/java/org/geoserver/wms/map/PDFMapResponse.java
bencaradocdavies: @jodygarnett "GeoServer is Dark and Full of Terrors"!

PR #2100, opened 2017-02-09
[GEOS-7984] Updated pom to jaiext 1.0.13

ugomos:


aaime:
Restarted the build. Please also make sure that xerces is not getting into the "core" web application.

ugomos:
Yes, I checked the web-inf/lib folder in the war of geoserver and xerces is not there.

PR #2101, opened 2017-02-09
GSIP-157 / GEOS-7987 - Track modified properties in CatalogPostModifyEvent

tbarsballe:
GSIP Here: https://github.com/geoserver/geoserver/wiki/GSIP-157

API Changes:

* Public methods `getPropertyNames`,`getOldValues`,`getNewValues` added to **CatalogPostModifyEvent** interface.
* Public methods `getPropertyNames`, `setPropertyNames`, `getNewValues`, `setNewValues`, `getOldValues`, `setOldValues` added to **CatalogPostModifyEventImpl**.
* The arguments `List<String> propertyNames`, `List oldValues`, `List newValues` have been added to the package private method `firePostModified` of **AbstractCatalogDecorator**, **AbstractFilteredCatalog**, **CatalogImpl**, and **SecureCatalogImpl**, and to the protected methods `beforeSaved` and `afterSaved` of **AbstractCatalogFacade**.

All invocations of above methods and classes have been updated accordingly.

~Given that all the changes are internal (protected or package private) to the Catalog, this seems likely to be a relatively safe change (Although I think it should not be backported, as it is still an API change)~

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/impl/DefaultCatalogFacade.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/DefaultCatalogFacade.java
@@ -200,10 +200,17 @@ public class DefaultCatalogFacade extends AbstractCatalogFacade implements Catal
     }
     
     public void save(StoreInfo store) {
-        beforeSaved(store);
+        ModificationProxy h = (ModificationProxy) Proxy.getInvocationHandler(store);
+
+        // figure out what changed
+        List<String> propertyNames = h.getPropertyNames();
+        List<Object> newValues = h.getNewValues();
+        List<Object> oldValues = h.getOldValues();
+
+        beforeSaved(store, propertyNames, oldValues, newValues);
         stores.update(store);
         commitProxy(store);
-        afterSaved(store);
+        afterSaved(store, propertyNames, oldValues, newValues);
     }
     
     public <T extends StoreInfo> T detach(T store) {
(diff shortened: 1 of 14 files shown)


aaime:
Big API break Batman! There is no such a thing as a package protected method for an interface, they are all public.
http://stackoverflow.com/questions/5376970/protected-in-interfaces

In this current form the change is proposal worthy and non backportable. Any chance you can use default methods to fix the backwards compatibility issues?

tbarsballe:
> There is no such a thing as a package protected method for an interface, they are all public.

Huh, learned something new today. Seems like that should have been obvious in retrospect.

GSIP incoming.

There are a few ways I could see of maintaining API backwards-compatibility. Unfortunately, they all run into the same problem:

By necessity, anything using these methods would not set the changed values. This means that anything listening on CatalogPostModifiedEvent can no longer rely on an empty list of properties meaning nothing has changed.
If we can't rely on the list of changed values being valid, then this substantially reduces the utility of this change.

aaime:
Master is ready for breaking changes :-)

tbarsballe:
Thanks for the reminder.
I've updated this so it is mergeable again, and sent out a notification to
the dev list.

src/main/src/main/java/org/geoserver/catalog/impl/DefaultCatalogFacade.java
```diff
@@ -200,10 +200,17 @@ public void remove(StoreInfo store) {
     }
     
     public void save(StoreInfo store) {
-        beforeSaved(store);
+        ModificationProxy h = (ModificationProxy) Proxy.getInvocationHandler(store);
```
aaime: I see this patter repeated over and over and over... how about a method taking a lambda to do the bit in the middle, after the notication and before commitProxy? Just thinking out loud.

tbarsballe: That sounds like a good idea, although I don't know how to implement it (I am still not especially familiar with lambdas in java). If you've got an example of what you are suggesting, I could certainly apply it to the relevant places.

PR #2102, opened 2017-02-10
[GEOS-7900] HttpServletRequestAware: store request in ThreadLocal

tbarsballe:
See: https://osgeo-org.atlassian.net/browse/GEOS-7900


aaime:
+1

PR #2103, opened 2017-02-10
[GEOS-7991] Re-initialize null LegendInfo after Style Page “apply”

mattkrusz:
Fix for https://osgeo-org.atlassian.net/browse/GEOS-7991 -- Re-initialize LegendInfo when it is null after an "apply".

### DIFF
--- a/src/web/wms/src/main/java/org/geoserver/wms/web/data/AbstractStylePage.java
+++ b/src/web/wms/src/main/java/org/geoserver/wms/web/data/AbstractStylePage.java
@@ -327,6 +327,15 @@ public abstract class AbstractStylePage extends GeoServerSecuredPage {
                     });
                 }
             }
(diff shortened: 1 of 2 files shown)


aaime:
Tried a quick cherry-pick on 2.10.x but failed, could you have a look at the backport?

mattkrusz:
Hi Andrea, it looks like a cherry-pick on the  most recent commit introduces a conflict from an intermediate commit to the StyleEditPage that has not yet been backported. 

Both of these commits are simple bugfixes and they could both be backported:

`git cherry-pick 00a49a64c5`
`git cherry-pick dab4b555bf6c4` 

I went ahead and made a PR for both of these against 2.10.x here: https://github.com/geoserver/geoserver/pull/2114

Thanks!

PR #2104, opened 2017-02-12
[GEOS-7992] Grab exclusive lock on catalog only when actually writing

aaime:


aaime:
Merging only on master at this point.

PR #2105, opened 2017-02-12
[GEOS-7881] Include Marlin by default. 

aaime:
Added support for bin under linux, and documentation for war deploys. Missing bin on Windows and Windows/OSX installers


doc/en/user/source/production/container.rst
```diff
@@ -54,6 +52,17 @@ For more information about JVM configuration, see the article `Performance tunin
...
+Enable the Marlin rasterizer
+----------------------------
+
+The Marlin rasterizer in Java 8 and getting better performance and scalability while rendering vector data in Java 8. 
+In order to enable it add the following among the JVM startup options::
+
+     -Xbootclasspath/a:$MARLIN_JAR 
+     -Dsun.java2d.renderer=org.marlin.pisces.MarlinRenderingEngine 
+
+where ``$MARLIN_JAR`` is the location of the ``marlin*.jar`` file located in the geoserver/WEB-INF/lib directory.
```
jodygarnett: Can I recommend including this in the jetty/lib folder rather than the war?

aaime: Meh... the best position for the Marlin jar is inside the JDK, jre/lib/ext.
By putting it in jetty/lib you're kicking out of the game the .war deployers, which will have to download Marlin separately, and it will be harder to place it there than to place it in the war.
The war position is still problematic, because some containers do not explode the .war.

Anyways, if you feel strongly, go for it, fork this pull request and do the work. I asked for help because I won't be able to add any further help on this one.

jodygarnett: I am not too bent out of shape it is not like this is a standalone jetty.

That is quite the trick with find to locate the marlin jar, I was not going  to be able to do something similar for start.bat and will need the specific version.

aaime: In the start.sh I managed to make it version independent, I think it's important as Marlin gets new versions out with sometimes significant improvements. For example the latest has improvements for area filling, so it would be good to upgrade to it. On windows batch files getting the location might be trickier, but should be doable.

jodygarnett:
Looks like this missed the beta; any interest in getting it in for RC1?

aaime:
Need someone on Windows to modify start.bat and service runner. Not sure about OSX, does it need any extra help?

simboss:
I will ping @randomorder about this.

aaime:
@randomorder just added a windows equivalent of what I did for startup.sh, into startup.bat
Items still missing:

- Windows service support
- Does OSX need any change?

jodygarnett:
OSX will need a change, the java startup parameters are located here:

windows service wrapper settings are here:
- https://github.com/geoserver/geoserver/blob/master/src/release/wrapper/bin/wrapper/wrapper.conf

Mac settings are in Info.plist in the app bundle:
- https://github.com/geoserver/geoserver/blob/master/src/release/installer/mac/build.xml#L106

aaime:
So we lost this one along the road eh? I believe we are still missing changes for:
* Windows service - @randomorder 
* OSX - @jodygarnett

aaime:
I'd really hate to lose this one, can we get some help to complete it for 2.12 merge?

guischulz:
@aaime I made some changes to add Marlin support in Windows installer and services
https://github.com/guischulz/geoserver/commit/846d82568b8c1333d24cc04515fe94e7fdd02656

Should I make a pull request to your marlin branch?

aaime:
@guischulz yes please

jodygarnett:
That looks great @guischulz

aaime:
@guischulz managed to alter the NSI installer so that it's marlin version independent. @jodygarnett I believe we're just missing the OSX bits now.

bourgesl:
Congratulations. 
Hacking the jdk / bootclasspath is tricky except if you provide your own jvm (openjdk ?) like Azul Zulu 8 which is redistribuable.

aaime:
Hi @bourgesl we never bundled a JDK, it's too big.

bourgesl:
Hi, I looked at your code changes:
- marlin jars must be loaded by the JVM (Boot classpath Append) so the jar files must not be in WEB-INF/lib folder to avoid tomcat load this library in an application class loader. You should copy jars in another location ...
- you should consider using the lastest release: 0.7.5_b2 at https://github.com/bourgesl/marlin-renderer/releases/download/v0.7.5_2/marlin-0.7.5-Unsafe.jar

Thanks for your interest 
& enjoy Marlin (made for you)
Laurent

bourgesl:
I could help for OSX if you want. I have macs at work to test it

aaime:
Putting it in the war file is the cheap way to ensure Marlin is available in every form, even if when deploiyng war files one has to activate it by hand, otherwise we're back to pointing to the Marlin web site (most production installation deploy from war, not using the installer). The issue with the classloader, did you verify it, or is it an hypothesis? I am expecting no problems since JDK own code loads Marlin, it should not even go though the Tomcat classloader.

bourgesl:
Andrea, @aaime 
I agree it is simpler to put the marlin jar files into the war file. 
I am almost sure it will not cause any problem (class loader) as it can only be loaded (and working) if the bootclasspath argument is given to the JVM and its classes will live in the JVM system class loader (top level in the class loader hierarchy).

aaime:
My take, since OSX devs do not seem to show an interest, I'd go ahead and merge it after upgrading Marlin.

jodygarnett:
So what is  needed here - a decision to put jar in war file, and then changing the wrapper command line options used by osx?

aaime:
This pull request already puts the jar in the war file, it seems there is no debate left there as to whether it will work or not (we think it will). What's missing is changing the wrapper command line options used by osx in a way similar to windows/linux (both env mananged to get independent of the marlin version included, that simplifies maintenance).

bourgesl:
Seems good. If you need, I can test installing geoserver on macos (dmg installer). 
Just give me the package.

jodygarnett:
I do not think the osx wrapper all that smart:

So what do we have in the codebase:
- https://github.com/geoserver/geoserver/blob/master/src/release/wrapper/bin/wrapper/wrapper.conf
- https://github.com/geoserver/geoserver/blob/master/src/release/installer/win/wrapper.conf

In a OSX Distribution we have:
- Contents/Info.plist <-- has an array of JVMOptions including -DGEOSERvER_DATA_DIR and JVMDefaultOptions which is empty.
- Contents/Java/start.ini <-- looks to be jetty config
- Contents/Plugins/jdk1.8.0_73.jdk

The installer [build](https://github.com/geoserver/geoserver/blob/master/src/release/installer/mac/build.sh#L18) does some sed hacking of Info.plist above to change CFBundleVersion

The ant [build.xml](https://github.com/geoserver/geoserver/blob/master/src/release/installer/mac/build.xml#L104) looks to be the original author of the file.

aaime:
The build.xml seems to be the right place to do version recognition and injection of the JVM options, it's already setting some other opts:
https://github.com/geoserver/geoserver/blob/master/src/release/installer/mac/build.xml

guischulz:
I mad some changes to build.sh and build.xml to enable the marlin renderer on OSX.
https://github.com/aaime/geoserver/pull/5/commits/39949bef8feed6c950b1232d86f0fa64cd73e63d

aaime:
@guischulz merged. @jodygarnett can you give it a try?

jodygarnett:
I was trying to think how to try it, other than going through the release process.

bourgesl:
Seems good.
On MacOS it seems really small heap: Xms128m/Xmx512m !
You should increase to 512m/1g at least.

bourgesl:
Please use Marlin release 0.7.5 or 0.7.4 at least.

aaime:
Marlin upgraded to 0.7.5 and merge conflicts fixed, still haven't had time to try out a build.

bourgesl:
Congratulations

PR #2110, opened 2017-02-14
[GEOS-7994] Force transformation of ROI and NoData to transparent/bg color in direct raster rendering path

aaime:
Note, it might not build until https://github.com/geotools/geotools/pull/1475 gets merged and built by Ares


src/wms/src/main/java/org/geoserver/wms/map/RenderedImageMapOutputFormat.java
```diff
@@ -1277,28 +1278,38 @@ protected GridCoverage2D readCoverage(GridCoverage2DReader reader,
...
+                || transparencyType != Transparency.OPAQUE
+                || iw.getNoData() != null || roiCandidate instanceof ROI) {
+            ROI roi;
+            if(roiCandidate instanceof ROI) {
+                ROI imageROI = (ROI) roiCandidate;
+                roi = imageROI.intersect(new ROIShape(mapRasterArea));
+            } else {
+                Rectangle intersection = imageBounds.intersection(mapRasterArea);
+                roi = new ROIShape(!intersection.isEmpty() ? intersection : mapRasterArea);
+            }
```
dromagnoli: Should we use ROIGeometry here instead of ROIShape? 
I also see the added ROIGeometry import at the beginning of this class.
Do we already have the ROIGeometry JAI-EXT version available instead of the jai-tools one?

src/wms/src/main/java/org/geoserver/wms/map/RenderedImageMapOutputFormat.java
```diff
@@ -1277,28 +1278,38 @@ protected GridCoverage2D readCoverage(GridCoverage2DReader reader,
         
         //
         // If we need to add a collar use mosaic or if we need to blend/apply a bkg color
-        if(!(imageBounds.contains(mapRasterArea) || imageBounds.equals(mapRasterArea))||transparencyType!=Transparency.OPAQUE) {
-            Rectangle roi = imageBounds.intersection(mapRasterArea);
-            ROI[] rois = new ROI[] { new ROIShape(!roi.isEmpty() ? roi : mapRasterArea) };
+        ImageWorker iw = new ImageWorker(image);
+        Object roiCandidate = image.getProperty("roi");
+        if(!(imageBounds.contains(mapRasterArea) || imageBounds.equals(mapRasterArea))
```
aaime: Will fix

src/wms/src/main/java/org/geoserver/wms/map/RenderedImageMapOutputFormat.java
aaime: I was tempted, but since I want to backport, decided not to. ROIGeometry general usage also got some recent fixes which needs some more widespread testing before we go and replace all ROIShape usag with ROIGeometry imho.

dromagnoli: Makes sense.

PR #2111, opened 2017-02-14
[GEOS-7995] Make JMS not serialize properties of type catalog

nmco:
Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-7995

JMS plugin will not serialize properties of type catalog when propagating modify events. Catalogs are instance specific and most of is fields are transient, so no collateral effects are expected.

A test case for this was also added.

### DIFF
--- a/src/community/jms-cluster/jms-geoserver/src/main/java/org/geoserver/cluster/impl/handlers/catalog/JMSCatalogEventHandler.java
+++ b/src/community/jms-cluster/jms-geoserver/src/main/java/org/geoserver/cluster/impl/handlers/catalog/JMSCatalogEventHandler.java
@@ -5,17 +5,20 @@
  */
 package org.geoserver.cluster.impl.handlers.catalog;
 
-import java.util.logging.Level;
-
-import javax.jms.JMSException;
-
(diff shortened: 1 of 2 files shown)


src/community/jms-cluster/jms-geoserver/src/main/java/org/geoserver/cluster/impl/handlers/catalog/JMSCatalogEventHandler.java
```diff
@@ -67,4 +71,49 @@ public CatalogEvent deserialize(String s) throws Exception {
...
+	 * for catalog modified events.
+	 */
+    private CatalogEvent removeCatalogProperties(CatalogEvent event) {
+        if (!(event instanceof CatalogModifyEvent)) {
+            // not a modify event so nothing to do
+            return event;
+        }
+        CatalogModifyEvent modifyEvent = (CatalogModifyEvent) event;
+        // index all the properties that are not of catalog type
+		List<Integer> indexes = new ArrayList<>();
```
aaime: Tabs?

nmco:
Thanks, formatting fixed (no idea how the tabs ended there).

PR #2116, opened 2017-02-19
GSIP-155 follow up: parallelizing catalog loading on the IO side

aaime:


aaime:
This build is also failing due to the AbstractDataStore removal, in NetCDF WCS output format module

PR #2117, opened 2017-02-20
GEOS-7989 WMS/WCS supports only String typed custom dimensions

NielsCharlier:
https://osgeo-org.atlassian.net/browse/GEOS-7989


NielsCharlier:
failure unrelated

PR #2118, opened 2017-02-20
Convert deprecated AbstractDataStore usage to ContentDataStore

tbarsballe:
A test case in the netcdf-out module uses AbstractDataStore. This PR modifies this to use ContentDataStore instead.

### DIFF
--- a/src/extension/netcdf-out/src/test/java/org/geoserver/wcs2_0/MultiDimDataStore.java
+++ b/src/extension/netcdf-out/src/test/java/org/geoserver/wcs2_0/MultiDimDataStore.java
@@ -5,33 +5,17 @@
  */
 package org.geoserver.wcs2_0;
 
-import java.io.IOException;
-import java.text.SimpleDateFormat;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Date;
-import java.util.List;
-import java.util.TimeZone;
-import java.util.logging.Level;
-import java.util.logging.Logger;
-
-import org.geotools.data.AbstractDataStore;
-import org.geotools.data.AbstractFeatureSource;
+import com.vividsolutions.jts.geom.Envelope;
+import com.vividsolutions.jts.geom.Geometry;
+import com.vividsolutions.jts.geom.GeometryFactory;
 import org.geotools.data.CollectionFeatureReader;
-import org.geotools.data.DataSourceException;
-import org.geotools.data.DataStore;
-import org.geotools.data.DataUtilities;
-import org.geotools.data.EmptyFeatureReader;
-import org.geotools.data.FeatureListener;
 import org.geotools.data.FeatureReader;
-import org.geotools.data.MaxFeatureReader;
 import org.geotools.data.Query;
 import org.geotools.data.QueryCapabilities;
-import org.geotools.data.ReTypeFeatureReader;
-import org.geotools.data.Transaction;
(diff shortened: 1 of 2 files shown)


tbarsballe:
So, looks like:
a) These tests ought to run fine without netcdf-out native libs, as the Travis failures are the same ones I see on my machine (which makes sense)
b) The test failures _appear_ to be unrelated to the AbstractDataStore -> ContentDataStore change, which is ... odd.

src/extension/netcdf-out/src/test/java/org/geoserver/wcs2_0/MultiDimDataStore.java
```diff
@@ -94,377 +83,304 @@
...
+                } catch (Exception e) {
+
+                    e.printStackTrace();
+                    throw new RuntimeException(e);
+                }
+            }
 
+            @Override
+            protected SimpleFeatureType buildFeatureType() throws IOException {
+                return null;
```
aaime: Null? This method is the one that builds the feature type on behalf of the content store, shouldn't it return FEATURE_TYPE?

tbarsballe: Whoops, missed that one when autogenerating methods. Definitely should return feature type here.

tbarsballe:
The remaining two failures appear to be caused by a difference of a few milliseconds in the time bounds / filter. Looking into it...

tbarsballe:
Date issues should be fixed. I'm going to wait and verify the Travis Build actually passes (works on my machine), then squash the excess commits together and merge.

src/extension/netcdf-out/src/test/java/org/geoserver/wcs2_0/MultiDimDataStoreFactory.java
```diff
@@ -5,24 +5,38 @@
...
 import java.io.Serializable;
+import java.util.Collections;
 import java.util.Map;
 
 import org.geotools.data.AbstractDataStoreFactory;
 import org.geotools.data.DataStore;
 import org.geotools.data.DataStoreFactorySpi;
 
-public class MultiDimDataStoreFactory extends AbstractDataStoreFactory {
+public class MultiDimDataStoreFactory implements DataStoreFactorySpi {
```
jodygarnett: I think AbstractDataStoreFactory still exists ...

src/extension/netcdf-out/src/test/java/org/geoserver/wcs2_0/MultiDimDataStoreFactory.java
```diff
@@ -35,4 +49,19 @@ public String getDescription() {
         return new DataStoreFactorySpi.Param[] { new Param("ParentLocation") };
     }
 
+    @Override
+    public boolean canProcess(Map<String, Serializable> map) {
+        File dir = new File((String) map.get("ParentLocation"));
+        return dir.exists();
```
jodygarnett: There is a utility method to change the map against parameter info you could use.

src/extension/netcdf-out/src/test/java/org/geoserver/wcs2_0/MultiDimDataStoreFactory.java
tbarsballe: It does, but is deprecated.

PR #2125, opened 2017-02-21
[GEOS-7990] Layer sorting index

PascalLike:
Hi,
some changes for supporting the layer sort by index. For details:
https://osgeo-org.atlassian.net/browse/GEOS-7990


aaime:
Hum... the timing of these pull requests could not be worse... we just entered feature freeze, no new features/improvements can be added. On the bright side, the freeze is not long, two weeks (but I'd expect the core developers to start actually paying attention to new features when the final release is out, in a month)

PascalLike:
Ok thanks, see you then!

src/main/src/main/java/org/geoserver/catalog/impl/ResourceInfoImpl.java
```diff
@@ -79,6 +80,8 @@
     protected StoreInfo store;
     
     protected transient Catalog catalog;
+    
+    protected Integer sortIndex = ResourceInfo.DEFAULT_INDEX_FOR_NOT_SORTED;
```
aaime: Mind, XStream deserializing will likely ignore this and set the variable to null. If you want to make sure it has a non null value you can update the CatalogImpl.resolve(...) methods to force a particular value, they are called when an object is added to the catalog (so, also during startup).

aaime:
Hi, the freeze has been over for a while now, will there be progress on these pull requests? :-)

PascalLike:
Excuse me if I didn't answered you before. I'll work on it as soon as possible ;-)

aaime:
Hi, any progress? :-)

PascalLike:
Hi aaime, excuse me, I didn't give the right attention to the issue. I found that the changes doesn't work with REST API of Geoserver. I'm trying to solve this issue, but basically I'm a little bit stuck. Do you have any suggestion? Maybe it's simpler than I imagine.

aaime:
Can you describe what "does not work with the REST API" mean? :-)

aaime:
Any chance of resurrecting this work? :)

PascalLike:
Sorry that it takes so long @aaime . I'm closing it for now, it's not nice to have this PR pending for so long. I will keep the branch and I will open a new PR when I'll have elements to decide what to do with it. I apologize

PR #2128, opened 2017-02-22
[GEOS-8003] Remove BeanFactoryPostProcessor callbacks beans invocations (backport 2.9.x)

nmco:
Backport of pull request: https://github.com/geoserver/geoserver/pull/2126

Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-8003

The error that originate this issue is not easy to reproduce (actually not idea how to do it) since it depends on Sping beans initialization race conditions. The normal beans instantiation is already tested by the current tests, so no new tests were added.

This whole situation happens because we don't have a standard way in Spring and GeoServer of marking a bean as the default one and allow extensions to provide an override. The closes mechanism we have right now is the beans priority interface, but this mechanism implies that who is using the beans should expected several beans and use the priority one and this is always not the case.

### DIFF
--- a/src/extension/control-flow/src/main/java/org/geoserver/flow/ControlFlowCallback.java
+++ b/src/extension/control-flow/src/main/java/org/geoserver/flow/ControlFlowCallback.java
@@ -26,6 +26,7 @@ import org.geoserver.platform.GeoServerExtensions;
 import org.geoserver.platform.Operation;
 import org.geotools.util.logging.Logging;
 import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.beans.factory.config.ConstructorArgumentValues;
@@ -34,6 +35,7 @@ import org.springframework.beans.factory.support.BeanDefinitionRegistryPostProce
 import org.springframework.beans.factory.support.RootBeanDefinition;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
+import org.springframework.context.ConfigurableApplicationContext;
 
 /**
  * Callback that controls the flow of OWS requests based on user specified rules and makes sure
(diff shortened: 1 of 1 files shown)


src/extension/control-flow/src/main/java/org/geoserver/flow/ControlFlowCallback.java
```diff
@@ -187,51 +188,43 @@ public Operation operationDispatched(Request request, Operation operation) {
...
         if (provider == null) {
             provider = new DefaultFlowControllerProvider(applicationContext);
         }
     }
 
-    @Override
-    public void postProcessBeanDefinitionRegistry(BeanDefinitionRegistry registry)
-            throws BeansException {
-        // Inject DefaultFlowControllerProvider definition it not defined
-        if (GeoServerExtensions.bean(FlowControllerProvider.class, applicationContext) == null) {
```
aaime: Skipping GeoServerExtensions.bean breaks the ability to use ExtensionFilter to disable a certain plugin in favor of another. Not that bad here, only one FlowControllerProvider is meant to be used in the whole application regardless, so I'm +0 on this change.

nmco: Thanks for the feedback.

I'm not removing the use of ``GeoServerExtensions.bean(...)``, the beans are still retrieved by ``GeoServerExtensions.bean(...)`` see line 198.

However is TRUE that when I check if the beans exist in the application context I don't use the  ``GeoServerExtensions`` methods, long story short ``GeoServerExtensions`` caches the lookup result and will not update itself if the application context is updated (new beans added) or use the new updated context if provided:
https://github.com/geoserver/geoserver/blob/2.10.x/src/platform/src/main/java/org/geoserver/platform/GeoServerExtensions.java#L202-L202

so I used the application context to test if any bean exists, anyway since the find bean will no be used for anything except the existence test this should not be a problem.

Ideally the extension filter mechanism should be used, i.e. declare the default beans and if anyone wants to override them the extensions filter mechanism should be used. But since this was not used in the first lace adding it now will break backwards compatibility.

PR #2131, opened 2017-02-23
[GEOS-8004] Adds back the possibility to externalize GWC configuration using GEOWEBCACHE_CACHE_DIR

nmco:
Mailing list discussion: http://osgeo-org.1560.x6.nabble.com/Add-back-the-possibility-to-externalize-GWC-configuration-using-GEOWEBCACHE-CACHE-DIR-td5309333.html

Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-8004

This [commit](https://github.com/geoserver/geoserver/commit/a72682bc5f86114dc56ba9d68f19b670da76cd99#diff-5b32e54953013fe49c4405ac132f9d3eR18) removed the possibility to externalize GWC configuration using Java property ``GEOWEBCACHE_CACHE_DIR``.

This pull request fix the issue above and adds a new test case. Code that assumed only the default location was also updated.

This changes were manually tested and everything seems to be working as expected. The ``back-restore`` module was also tested.

Since this restores a functionality that is already documented no documentation was updated.

This pull request makes it also possible to use ``GEOWEBCACHE_CONFIG_DIR`` to set GWC configuration directory.

### DIFF
--- a/src/gwc/src/main/java/org/geoserver/gwc/config/GeoserverXMLResourceProvider.java
+++ b/src/gwc/src/main/java/org/geoserver/gwc/config/GeoserverXMLResourceProvider.java
@@ -1,18 +1,12 @@
+/* (c) 2017 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
 package org.geoserver.gwc.config;
-
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.text.SimpleDateFormat;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.Date;
-import java.util.List;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import org.geoserver.platform.resource.Files;
+import org.geoserver.platform.GeoServerExtensions;
 import org.geoserver.platform.resource.Resource;
 import org.geoserver.platform.resource.ResourceStore;
 import org.geoserver.platform.resource.Resources;
@@ -20,11 +14,29 @@ import org.geoserver.util.Filter;
 import org.geoserver.util.IOUtils;
 import org.geowebcache.config.ConfigurationException;
 import org.geowebcache.config.ConfigurationResourceProvider;
+import org.geowebcache.config.XMLFileResourceProvider;
+import org.geowebcache.storage.DefaultStorageFinder;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.text.SimpleDateFormat;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.Date;
+import java.util.List;
+import java.util.Optional;
 
 public class GeoserverXMLResourceProvider implements ConfigurationResourceProvider {
 
-    private static Log log = LogFactory.getLog(org.geowebcache.config.XMLFileResourceProvider.class);
-    
+    private static Log LOGGER = LogFactory.getLog(GeoserverXMLResourceProvider.class);
+
+    static final String GEOWEBCACHE_CONFIG_DIR_PROPERTY = XMLFileResourceProvider.GWC_CONFIG_DIR_VAR;
+
+    static final String GEOWEBCACHE_CACHE_DIR_PROPERTY = DefaultStorageFinder.GWC_CACHE_DIR;
+
     public static final String DEFAULT_CONFIGURATION_DIR_NAME = "gwc";
     
     /**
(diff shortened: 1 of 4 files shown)


src/gwc/src/main/java/org/geoserver/gwc/config/GeoserverXMLResourceProvider.java
```diff
@@ -39,33 +46,64 @@
...
-   
+
+    /**
+     * Helper method that infers the directory that will contain or contains GWC configuration. First we will
+     * check if a specific location was set using property GEOWEBCACHE_CACHE_DIR, then we will check if a
+     * location was provided and then fallback on the default location.
+     */
+    private static Resource inferConfigDirectory(ResourceStore resourceStore, String providedConfigDirectory) {
+        // first let's see if a specific location was defined using GEOWEBCACHE_CONFIG_DIR property
+        String configDirectoryPath = System.getProperty(GEOWEBCACHE_CONFIG_DIR_PROPERTY);
```
jodygarnett: Something seems odd here; we have a utility method for these kind of settings that respects looking things up as an environmental variable, system property or context parameter.

```
String proxyBase = GeoServerExtensions.getProperty("PROXY_BASE_URL")
```

And the similar logic here - https://github.com/geoserver/geoserver/blob/master/src/platform/src/main/java/org/geoserver/platform/GeoServerResourceLoader.java#L601

src/gwc/src/main/java/org/geoserver/gwc/config/GeoserverXMLResourceProvider.java
```diff
@@ -23,8 +26,12 @@
 
 public class GeoserverXMLResourceProvider implements ConfigurationResourceProvider {
 
-    private static Log log = LogFactory.getLog(org.geowebcache.config.XMLFileResourceProvider.class);
-    
+    private static Log LOGGER = LogFactory.getLog(org.geowebcache.config.XMLFileResourceProvider.class);
```
smithkm: I know that this was already like this already so it's not your fault, but this looks like a good opportunity for confusing logs (Error occurs in `GeoserverXMLResourceProvider` but is logged as occurring in `XMLFileResourceProvider`)

Not a required change, just a comment.

src/gwc/src/main/java/org/geoserver/gwc/config/GeoserverXMLResourceProvider.java
smithkm: Yes, this really should be done using `GeoServerExtensions.getProperty` or `GWCVars.findEnvVar`.

`XMLFileResourceProvider` uses`GWCVars.findEnvVar` to look up the config dir attribute and delegates to the `DefaultStorageLocator` if that was undefined which then uses its own slightly different implementation of the lookup algorithm to find the cache dir.

https://github.com/GeoWebCache/geowebcache/blob/master/geowebcache/core/src/main/java/org/geowebcache/config/XMLFileResourceProvider.java#L82-L98
https://github.com/GeoWebCache/geowebcache/blob/master/geowebcache/core/src/main/java/org/geowebcache/config/XMLFileResourceProvider.java#L160-L162

src/gwc/src/main/java/org/geoserver/gwc/config/GeoserverXMLResourceProvider.java
```diff
@@ -23,8 +26,12 @@
...
+
+    public static final String GEOWEBCACHE_CONFIG_DIR_PROPERTY = "GEOWEBCACHE_CONFIG_DIR";
+
+    public static final String GEOWEBCACHE_CACHE_DIR_PROPERTY = "GEOWEBCACHE_CACHE_DIR";
+
     public static final String DEFAULT_CONFIGURATION_DIR_NAME = "gwc";
```
smithkm: https://github.com/GeoWebCache/geowebcache/pull/491

src/gwc/src/main/java/org/geoserver/gwc/config/GeoserverXMLResourceProvider.java
nmco: Agree my bad I will fix this.

src/gwc/src/main/java/org/geoserver/gwc/config/GeoserverXMLResourceProvider.java
nmco: I didn't notice it, I will fix it.

src/gwc/src/main/java/org/geoserver/gwc/config/GeoserverXMLResourceProvider.java
nmco: Waiting for the change on ``XMLFileResourceProvider.GWC_CONFIG_DIR_VAR`` to be merged.

src/gwc/src/main/java/org/geoserver/gwc/config/GeoserverXMLResourceProvider.java
nmco: Fixed.

nmco:
PR updated, waiting for the change on XMLFileResourceProvider.GWC_CONFIG_DIR_VAR to be merged.

nmco:
Pull request amended with all suggestions.

PR #2137, opened 2017-03-03
Simple NPE Checks being added

dromagnoli:
Adding a simple bunch of Null Pointer checks to the cascaded
a.get.get.get.get


aaime:
I see this is a problem for WMS cascading, where the reader has no source. Please add a jira ticket and refer to it in the commit?

aaime:
I'm going to add a test case and merge manually

PR #2142, opened 2017-03-07
[GEOS-7925] Fix wicket multipart upload

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-7925

Fix wicket multipart upload:

* Upgrade from Wicket 7.1.0 to 7.2.0; this issue was fixed by WICKET-6002.
* Remove CommonsMultipartResolver from gs-gwc-sqlite context

### DIFF
--- a/src/community/gwc-sqlite/src/main/java/org/geoserver/gwc/web/blob/SqliteMultipartFilter.java
+++ b/src/community/gwc-sqlite/src/main/java/org/geoserver/gwc/web/blob/SqliteMultipartFilter.java
@@ -1,0 +1,23 @@
+package org.geoserver.gwc.web.blob;
+
+import org.geoserver.filters.GeoServerFilter;
+import org.springframework.web.multipart.support.MultipartFilter;
+
+import javax.servlet.FilterChain;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+public class SqliteMultipartFilter extends MultipartFilter implements GeoServerFilter {
+
+    @Override
+    protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain) throws ServletException, IOException {
+        String path = request.getRequestURI();
+        if (path.contains("gwc/rest/sqlite/")) {
+            super.doFilterInternal(request, response, filterChain);
+        } else {
+            filterChain.doFilter(request, response);
+        }
+    }
+}
--- a/src/community/gwc-sqlite/src/main/resources/applicationContext.xml
+++ b/src/community/gwc-sqlite/src/main/resources/applicationContext.xml
@@ -12,7 +12,8 @@
     <mvc:annotation-driven/>
     <context:annotation-config/>
     <context:component-scan base-package="org.geowebcache.sqlite"/>
-    <bean id="multipartResolver" class="org.springframework.web.multipart.commons.CommonsMultipartResolver"/>
+    <bean id="filterMultipartResolver" class="org.springframework.web.multipart.commons.CommonsMultipartResolver"/>
+    <bean id="sqliteMultipartFilter" class="org.geoserver.gwc.web.blob.SqliteMultipartFilter"/>
     <bean id="MbtilesConfigurationProvider" class="org.geowebcache.sqlite.MbtilesConfigurationProvider"/>
     <bean class="org.geoserver.gwc.web.blob.MbtilesBlobStoreType"/>
 </beans>
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -1791,7 +1791,7 @@
   <jetty.version>9.2.13.v20150730</jetty.version>
   <jetty.servlet-api.version>3.1.0</jetty.servlet-api.version>
   <poi.version>3.14</poi.version>
-  <wicket.version>7.1.0</wicket.version>
+  <wicket.version>7.6.0</wicket.version>
   <ant.version>1.8.4</ant.version>
   <imageio-ext.version>1.1.17</imageio-ext.version>
   <jaiext.version>1.0.13</jaiext.version>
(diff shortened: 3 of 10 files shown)


src/community/gwc-sqlite/src/main/resources/applicationContext.xml
```diff
@@ -12,7 +12,6 @@
     <mvc:annotation-driven/>
     <context:annotation-config/>
     <context:component-scan base-package="org.geowebcache.sqlite"/>
-    <bean id="multipartResolver" class="org.springframework.web.multipart.commons.CommonsMultipartResolver"/>
```
tbarsballe: @nmco - Do you know what this is used for in gwc-sqlite?

src/pom.xml
```diff
@@ -1791,7 +1791,7 @@
   <jetty.version>9.2.13.v20150730</jetty.version>
   <jetty.servlet-api.version>3.1.0</jetty.servlet-api.version>
   <poi.version>3.14</poi.version>
-  <wicket.version>7.1.0</wicket.version>
+  <wicket.version>7.2.0</wicket.version>
```
tbarsballe: We should probably discuss changing wicket versions at this stage in the release process.

tbarsballe:
I've updated this with a proper filter for the CommonsMultipartResolver such that it only gets applied to the sqlite endpoint.

I've done some basic manual tests, and it seems to be working fine.

src/community/gwc-sqlite/src/main/java/org/geoserver/gwc/web/blob/SqliteMultipartFilter.java
```diff
@@ -0,0 +1,22 @@
...
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+public class SqliteMultipartFilter extends MultipartFilter implements GeoServerFilter {
+    @Override
+    protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain) throws ServletException, IOException {
+        String path = ((HttpServletRequest) request).getRequestURI();
+        if (path.contains("/sqlite/")) {
```
tbarsballe: This is less robust than I would like, but the gwc sqlite OperationsRest configuration seems to be fairly permisive with paths (`@RequestMapping("**/sqlite")`) so I'm not quite sure what I can savely do that is more restricted.

Fortunately, the MultipartResolver doesn't seem to do anything when there is not multipart content, so we should be safe even if this gets applied where it shouldn't (for instance, if someone has a resource directory named `sqlite`)

nmco: Honestly I cannot remember the exact reason why I make this so permissive but I know he has something to do with the GeoServer integration: https://github.com/nmco/geoserver/blob/c435cf87926c6b2bae9349b19702b752ff1e98f4/src/gwc/src/main/resources/geowebcache-rest-context.xml

src/community/gwc-sqlite/src/main/resources/applicationContext.xml
nmco: Awnsered on the issue https://osgeo-org.atlassian.net/browse/GEOS-7925

src/community/gwc-sqlite/src/main/java/org/geoserver/gwc/web/blob/SqliteMultipartFilter.java
tbarsballe: My guess would be to support both the standalonge gwc (`/geowebcache/rest/sqlite`) and embedded gwc (`/geoserver/gwc/rest/sqlite`) rest endpoints?
If it is just that, then I can be much more restrictive, as this workaround/fix is only for the embedded geowebcache.

nmco: Ok I have investigate this and yes this is just for the integration. I have manually tested with path ``gwc/rest/sqlite/`` and it works has expected.

src/community/gwc-sqlite/src/main/java/org/geoserver/gwc/web/blob/SqliteMultipartFilter.java
```diff
@@ -0,0 +1,22 @@
...
+import javax.servlet.FilterChain;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+public class SqliteMultipartFilter extends MultipartFilter implements GeoServerFilter {
+    @Override
+    protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain) throws ServletException, IOException {
+        String path = ((HttpServletRequest) request).getRequestURI();
```
nmco: This cats is unnecessary.

src/community/gwc-sqlite/src/main/java/org/geoserver/gwc/web/blob/SqliteMultipartFilter.java
```diff
@@ -0,0 +1,22 @@
...
+import org.springframework.web.multipart.support.MultipartFilter;
+
+import javax.servlet.FilterChain;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+public class SqliteMultipartFilter extends MultipartFilter implements GeoServerFilter {
+    @Override
```
nmco: I would add a white line between the class and the @Override.

nmco:
I have manually tested this and everything works as expected. @tbarsballe thanks for finding a fix for the gwc-sqlite multipart problem :bowing_man:.

tbarsballe:
Updated based on feedback. Now uses path `gwc/rest/sqlite/`.

PR #2143, opened 2017-03-07
GEOS-7883 Return empty transaction result when no properties are given

dvntucker:
When there's no properties given to update in the transaction there's nothing to do, so update should just return immediately.


src/wfs/src/main/java/org/geoserver/wfs/UpdateElementHandler.java
```diff
@@ -163,6 +163,11 @@ public void execute(TransactionElement element, TransactionRequest request,
             String[] names = new String[properties.size()];
             Object[] values = new Object[properties.size()];
 
+            //If no properties are defined for an update, there's nothing to do
+            if (properties.isEmpty()) {
+                return;
```
jodygarnett: that seems fair

jodygarnett:
@dvntucker this is a bug fix, do you want to backport?

PR #2144, opened 2017-03-07
GEOS-8022: moved SLD and SLD_BODY disabling flag from GeoFence to WMS core

mbarto:


aaime:
Had a look, what's there looks good, two observations on missing bits:

- There is no documentation updates
- Disabling dynamic styling should also imply disabling GetMap POST requests using an inline user style (but not those referring to a named style instead)

mbarto:
@aaime 
I pushed a new commit adding both

PR #2145, opened 2017-03-07
added product_thumb table in opensearch db

Damianofds:
This PR adds to the postgis.sql script the DDL for the thumbs table of the opensearch db


aaime:
Yep, thanks

PR #2146, opened 2017-03-07
[GEOS-7963] Fixed invalid elements being added to the KML IconStyle.

sikeoka:
Attempting to use the technique found here:
http://blog.bdoughan.com/2012/11/using-jaxb-with-xslt-to-produce-html.html

### DIFF
--- a/src/kml/src/main/resources/org/geoserver/kml/icon_style_patch.xsl
+++ b/src/kml/src/main/resources/org/geoserver/kml/icon_style_patch.xsl
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Copyright (C) 2017 - Open Source Geospatial Foundation. All rights reserved.
+ This code is licensed under the GPL 2.0 license, available at the root
+ application directory.
+ -->
+<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:xalan="http://xml.apache.org/xalan" xmlns:kml="http://www.opengis.net/kml/2.2">
+    <xsl:output indent="yes" method="xml" standalone="yes" xalan:indent-amount="4" />
+    <xsl:template match="node()|@*">
+        <xsl:copy>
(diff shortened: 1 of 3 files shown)


src/kml/src/main/resources/org/geoserver/kml/icon_style_patch.xsl
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Copyright (C) 2017 - Open Source Geospatial Foundation. All rights reserved.
+ This code is licensed under the GPL 2.0 license, available at the root
+ application directory.
+ -->
+<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:xalan="http://xml.apache.org/xalan" xmlns:kml="http://www.opengis.net/kml/2.2">
+    <xsl:output indent="yes" method="xml" standalone="yes" xalan:indent-amount="4" />
```
aaime: Explicit xalan reference... wondering if it might cause issues in specific web containers forcing their own XML technology. Do you have any clue?
No change required, just wondering.

sikeoka:
- I have not done that yet.  I will look into getting it done soon.

- I don't really have a set up for doing thorough performance testing but repeatedly calling the encode method in my unit test appears to be about 10% slower with the XSLT than without it.  I could add a field to KMLEncodingContext that keeps track of whether there are any IconStyle elements, modify PlacemarkStyleDecoratorFactory to set that flag when it creates an IconStyle and then modify KMLEncoder to check that flag to determine whether to run the XSLT.

- Processors are supposed to ignore extension attributes that aren't in a namespace that it recognizes.  I checked that the unit tests run fine with both Xalan and Saxon, but I can only test the WAR file deployed in WildFly using Xalan.  Saxon defaults to 3 spaces for indentation but I could override that in the XSL file like I did for Xalan for more consistent indentation.

aaime:
Ok, merging. Will have to wait for the main build server to double check it. Can you remind me to backport in case I forget to do so?

PR #2148, opened 2017-03-09
Update compress-lzf library version

emerkle826:
Conflict between geogig and jdbcconfig depending on different versions of `compress-lzf.jar`

### DIFF
--- a/src/community/jdbcconfig/pom.xml
+++ b/src/community/jdbcconfig/pom.xml
@@ -52,7 +52,6 @@
       <!-- blazing fast LZW compression/decompression -->
       <groupId>com.ning</groupId>
       <artifactId>compress-lzf</artifactId>
-      <version>0.9.1</version>
     </dependency>
 
     <!-- Testing -->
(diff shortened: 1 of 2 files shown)


src/community/jdbcconfig/pom.xml
```diff
@@ -52,7 +52,7 @@
       <!-- blazing fast LZW compression/decompression -->
       <groupId>com.ning</groupId>
       <artifactId>compress-lzf</artifactId>
-      <version>0.9.1</version>
+      <version>1.0.3</version>
```
jodygarnett: I am used to this being handled in root pom.xml dependency management section (so we do not get conflicts between community modules).

Is there a reason why this is hard coded ?

emerkle826: Good point Jody. I was simply updating the version where it was defined, but I will move it to the root pom.

emerkle826: @jodygarnett I've updated the PR to move the library version to the root pom.

jodygarnett:
perfect thanks, I will backport to 2.10.x.

PR #2149, opened 2017-03-09
Import Existing GeoGig repo via web API

goudine:
This commit adds the functionality of importing existing PostGres and directory
backed GeoGig repositories into GeoServer via the web API.


src/community/geogig/src/main/java/org/geogig/geoserver/rest/AddRepoRequestHandler.java
```diff
@@ -27,18 +23,26 @@
...
+import static org.restlet.data.Status.CLIENT_ERROR_BAD_REQUEST;
 
 /**
  * Utility for handling GeoGIG repository init requests. This class will pull repository creation
  * details (like parent directory, or PostgreSQL database connection parameters) from the Request
  * and build a GeoGIG repository form them, by converting the request into a
  * {@link org.locationtech.geogig.repository.Hints Hints}.
  */
-class InitRequestHandler {
+class AddRepoRequestHandler {
```
jdgarrett: This class was refactored out of the plugin, so the updates to support specifying a leaf directory will have to be made on the geogig side while the importGeogig function can be moved.  (See the recent changes in the plugin and model it after the init changes).

goudine:
New PR was made to account for the code being refactored.

PR #2150, opened 2017-03-11
Import Existing GeoGig repo via web API

goudine:
This commit adds the functionality of importing existing PostGres and directory
backed GeoGig repositories into GeoServer via the web API.

### DIFF
--- a/src/community/geogig/src/main/java/org/geogig/geoserver/rest/GeoServerRepositoryProvider.java
+++ b/src/community/geogig/src/main/java/org/geogig/geoserver/rest/GeoServerRepositoryProvider.java
@@ -8,6 +8,7 @@ import static org.locationtech.geogig.web.api.RESTUtils.getStringAttribute;
 
 import java.io.IOException;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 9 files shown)


src/community/geogig/src/main/java/org/geogig/geoserver/rest/GeoServerRepositoryProvider.java
```diff
@@ -194,5 +221,36 @@ private Repository findRepository(String repositoryName) {
             }
             return Optional.absent();
         }
+
+        static Optional<Repository> importGeogig(Request request) {
```
emerkle826: This can be private, it's in a private inner class, not getting exposed outside this class.

src/community/geogig/src/test/java/org/geogig/geoserver/functional/PluginWebAPICucumberHooks.java
```diff
@@ -0,0 +1,74 @@
...
+     */
+    @Inject
+    public PluginWebAPICucumberHooks(WebAPICucumberHooks parent) {
+        if (GeoServerFunctionalTestContext.class.isAssignableFrom(parent.context.getClass())) {
+            this.context = GeoServerFunctionalTestContext.class.cast(parent.context);
+        }
+    }
+
+    String systemTempPath() throws IOException {
+        return context.tempFolder.getRoot().getCanonicalPath().replace("\\", "/");
```
emerkle826: you can replace `context.tempFolder.getRoot()` with simply `context.getTempFolder()`, and thus not have to expose the tempFolder in GeoGig's FunctionalContext class.

emerkle826:
👍

PR #2152, opened 2017-03-12
[GEOS-8027] JUnit ResourceTest.testSpecialCharacterNames fails on Windows

guischulz:
https://osgeo-org.atlassian.net/browse/GEOS-8027

### DIFF
--- a/src/restconfig/src/test/java/org/geoserver/rest/resources/ResourceControllerTest.java
+++ b/src/restconfig/src/test/java/org/geoserver/rest/resources/ResourceControllerTest.java
@@ -7,6 +7,7 @@ package org.geoserver.rest.resources;
 import net.sf.json.JSON;
 import net.sf.json.JSONObject;
 import net.sf.json.test.JSONAssert;
+import org.apache.commons.lang.SystemUtils;
 import org.custommonkey.xmlunit.NamespaceContext;
 import org.custommonkey.xmlunit.SimpleNamespaceContext;
 import org.custommonkey.xmlunit.XMLAssert;
@@ -171,7 +172,7 @@ public class ResourceControllerTest extends GeoServerSystemTestSupport {
     @Test
     public void testSpecialCharacterNames() throws Exception {
         // if the file system encoded the file with a ? we need to skip this test
-        Assume.assumeTrue(getDataDirectory().get("po?zie").getType() == Type.UNDEFINED);
+        Assume.assumeTrue(SystemUtils.IS_OS_WINDOWS || getDataDirectory().get("po?zie").getType() == Type.UNDEFINED);
         Assert.assertEquals(Type.DIRECTORY, getDataDirectory().get("poëzie").getType());
         XMLUnit.setXpathNamespaceContext(NS_XML);
         Document doc = getAsDOM(RestBaseController.ROOT_PATH+"/resource/po%c3%abzie?format=xml");


aaime:
Given the area of the code it would be best is someone from Boundless reviewed this. @jodygarnett or @tbarsballe maybe?

tbarsballe:
I am not sure if Resource.UNDEFINED is really the right way to go here, or if the test case should be changed instead.

While the javadoc is unclear, I was generally under the impression that Resource.UNDEFINED is simply a resource that doesn't exist on the file system.

In a case where we would get an invalid path, the Resource would never be able to exist, so throwing an exception may be more correct. This is also a bit of an interesting case, since the path in question seems valid on UNIX but is invalid on Windows.

Given this, it is probably better to change the test case (possibly adding a special one for windows?).

aaime:
Hum... the original issue was that the particular test in the report used to fail on Linux under the POSIX locale, and non ascii chars were replaced by ? by the file system itself.
Now... not sure what happens in Windows under similar conditions, or if the conditions can be even replicated somehow. According to this, the filenames are stored in UTF-16, but there are still some possible snags: http://stackoverflow.com/questions/2050973/what-encoding-are-filenames-in-ntfs-stored-as

Uh, under these conditions, I'd probably just rather skip them "assume" if the OS is windows.

jodygarnett:
UNDEFINED means the resource does not exist (in whatever the storage mechanism is).

Resource api uses relative paths, following unix "/" conventions (so we should not run into windows vs linux differences).

tbarsballe:
> UNDEFINED means the resource does not exist

Given this, if the path is invalid (and thus the resource can _never_ exist), using UNDEFINED is probably the wrong choice.

Based on Andrea's comment, this seems to be an encoding issue rather than an OS-specific issue, so the test case likely needs to be changed.

bencaradocdavies:
@NielsCharlier the problem is that this Assume fails on Windows because "?" is an illegal character in paths. Can you suggest an alternative way of disabling this test?
https://github.com/geoserver/geoserver/blob/master/src/restconfig/src/test/java/org/geoserver/rest/resources/ResourceTest.java#L165

Note that on Windows with UTF, this test *should* be run and the filesystem should work fine. Only "?" is forbidden. I think this Assume needs to be adjusted. The test is otherwise good and should be run on Windows, where the platform encoding supports it.

NielsCharlier:
@bencaradocdavies  The line appears to have been added by Andrea
@aaime can the assume be replaced by something like this: https://github.com/geoserver/geoserver/blob/master/src/restconfig/src/test/java/org/geoserver/rest/resources/ResourceTest.java#L57

aaime:
@NielsCharlier, I've tried your suggestion replacing the first line with:

{code}
        Assume.assumeTrue(Charset.defaultCharset().newEncoder().canEncode('ë'));
{code}

It makes the test fail with the following error:

{code}
java.nio.file.InvalidPathException: Malformed input or input contains unmappable characters: ./target/default7817122953820465728data/poëzie
	at sun.nio.fs.UnixPath.encode(UnixPath.java:147)
	at sun.nio.fs.UnixPath.<init>(UnixPath.java:71)
	at sun.nio.fs.UnixFileSystem.getPath(UnixFileSystem.java:281)
	at java.io.File.toPath(File.java:2234)
	at org.geoserver.platform.resource.FileSystemResourceStore$FileSystemResource.getType(FileSystemResourceStore.java:431)
	at org.geoserver.rest.resources.ResourceResource.handleGet(ResourceResource.java:331)
{code}

aaime:
@guischulz what about changing the first line in the failing test to be:

````
Assume.assumeTrue(!SystemUtils.IS_OS_WINDOWS && getDataDirectory().get("po?zie").getType() == Type.UNDEFINED);
````

Does that work fine on your Windows installation? 
Can you think of cases in which Windows would not be able to encode "ë" and replace it with a different character?

**Comment by aaime** [comment]
@guischulz did you try my latest suggestion? Does it work?
Sorry for the situation, none of the core developers has a windows dev box, so we're "flying blind" here.

**Comment by ridethepenguin** [comment]
@aaime I'm unable to reproduce on my Windows 10 machine (localized in italian): even after removing the `Assume` statement, the test passes both under Eclipse (well, after I remembered to switch the text editor's encoding to UTF-8 😉 ) and Maven.

In fact, looks like Windows is properly handling the file name with special characters:

image

@guischulz What version of Windows are you running?

**Comment by guischulz** [comment]
@ridethepenguin I am running Windows 10 with German localization and yes Windows is correctly handling the filename, but `getDataDirectory().get("po?zie")` is raising an InvalidPathException

@aaime Sorry for my late reply, your suggestion works here and the test does not fail anymore. As `?` is never a valid character in Windows filenames it should be fine. Windows should always be able to create a "ë" character.

**Comment by ridethepenguin** [comment]
@guischulz Right, sorry I missed the point entirely, I thought the purpose of the `Assume` statement was to have the test skipped on Windows.

@aaime I tried with
```
Assume.assumeTrue(Charset.defaultCharset().newEncoder().canEncode('ë'));
```
and it works for me on both Windows and Linux, wondering why it didn't for you?

Last comment, wouldn't it be enough to reverse the `Assume` statement like this:
```
Assume.assumeTrue(getDataDirectory().get("poëzie").getType() == Type.DIRECTORY);
```
Works for me on both Windows and Linux, but the assumption is always satisfied, so I cannot say whether the test is correctly skipped when it should.

**Comment by bencaradocdavies** [comment]
There are two separate issues: Windows, which always forbids "?" in a path, and a platform encoding that does not support some characters in the path. Although modern Windows typically uses UTF-8, there are many non-Unicode platforms out there; often seen are ```ISO-8859-15``` (ISO Latin No. 9 (contains "€")) but also non Latin code pages such as ```windows-1251``` (Windows Cyrillic). See [Supported Encodings](https://docs.oracle.com/javase/8/docs/technotes/guides/intl/encoding.doc.html). There was some discussion in the original PR: https://github.com/geoserver/geoserver/pull/1487

To test for failures with a non-Unicode encoding, you can for example, in ```src/restconfig``` run:
```
mvn -o -Djvm.opts="-Dfile.encoding=US-ASCII" -Dtest=ResourceTest test
```
or just use ```-Dfile.encoding=US-ASCII``` in Eclipse. You can also test ```ISO-8859-15``` which should run tests with "ë" in the path and ```windows-1251``` which should not.

**Comment by guischulz** [comment]
I have updated my pull request to fix the JUnit test on Windows.  I tested on Windows and Linux with different file encodings. The change is minimal and only avoids `getDataDirectory().get("po?zie")` beeing called on windows. The test is not skipped.
Can anyone merge it?

**Comment by bencaradocdavies** [comment]
@guischulz your pull request now has conflicts because the epic [REST API Code Sprint](http://blog.geoserver.org/2017/04/11/rest-api-code-sprint-results/) changes just got merged onto master. The change that affects you is that ResourceTest has been renamed [ResourceControllerTest](https://github.com/geoserver/geoserver/blob/master/src/restconfig/src/test/java/org/geoserver/rest/resources/ResourceControllerTest.java). Please revise your pull request (in your case it will be easiest to reset to master, pull, and re-apply your change.).

**Comment by guischulz** [comment]
@bencaradocdavies I have resolved the conflicts and updated the pull request

**Comment by bencaradocdavies** [comment]
Thanks @guischulz for your perseverance. I think that the ```Assume``` condition is still incomplete and that the test will likely fail on Windows a locale such as ```US-ASCII``` that cannot encode all characters in "```/poëzie/café```". At least the test is no longer broken for all Windows developers.

@guischulz, what happens if you run in:
```
mvn -o -Djvm.opts="-Dfile.encoding=US-ASCII" -Dtest=ResourceControllerTest test
```
in ```src/restconfig``` on Windows with this most recent fix?

Can you see what names are used for ```/poëzie/café``` in Windows with ```US-ASCII``` encoding? What placeholder is used on Windows for unsupported characters?

guischulz:
@bencaradocdavies I have done tests with different settings for the file encoding and the result was good. The test was always succesful and the file name shows up correctly

Result when running:
`mvn -o -Djvm.opts="-Dfile.encoding=US-ASCII" -Dtest=ResourceControllerTest test`
us-ascii
Overall I have no idea how to force an unsupported character.

bencaradocdavies:
Thanks for testing @guischulz . Perhaps the Windows implementation ignores the file.encoding property and uses the filesystem encoding? Anyway, your test suggests that this test will pass even for Windows Cyrrilic and Greek encodings. Thanks again for the fix.

PR #2153, opened 2017-03-15
[GEOS-5869] Fix REST handling of H2 and SpatiaLite datastores

nmco:
This PR fix the REST handling of H2 and SpatiaLite data stores and add tests cases for both. Both cases requires special handling that cannot be generalized \ abstracted in an elegant way.

SpatiLite tests will only run if the appropriated native libraries are available otherwise they will be skipped and a warning will be logged.

Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-5869


aaime:
Looks good, merged. When the build on ares passes can you also merge the backports?

PR #2158, opened 2017-03-17
Make tile layer use new legend info structure

nmco:
Stop using deprecated ``TileLayer.LegendInfo`` class.

GWC associated PR: https://github.com/GeoWebCache/geowebcache/pull/502


nmco:
Fail due is dependency on the GWC PR. Once the GWC is merged I will restart the build.

PR #2162, opened 2017-03-17
[GEOS-8037] Save style file before saving the style info

nmco:
This PR makes sure that the style file is saved before the style info is stored in the catalog, this way someone listen on the catalog modified event can have access to the correct style file. This PR cotnains also a test for this behavior.

Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-8037
Mailing list discussion: http://osgeo-org.1560.x6.nabble.com/Style-modify-event-and-changed-resource-td5311789.html


jodygarnett:
This looks good: because this is an "internal change" (not visible to end-users) I am comfortable back porting.

I am not sure how others feel?

aaime:
Merged. +1 for backporting

PR #2170, opened 2017-03-19
[GEOS-8028] Exception when saving a layer group in GeoServer UI

aaime:


nmco:
Thanks Andrea, did a review (and didn't find anything to comment :P) and manual tested this and the issue is fixed. Everything else seems to be working as expected.

PR #2174, opened 2017-03-21
[GEOS-8040] Fixed accessing resources in subdirectories of the styles folder.

sikeoka:
External access to resources in subdirectories of the styles folder was broken in GeoServer 2.9.2 and 2.10.0 by the fix for https://osgeo-org.atlassian.net/browse/GEOS-7635.  The ability to access resources in styles subdirectories goes back to at least GeoServer 1.7.x.

This pull request removes the restriction of the number of tokens in the request path and that the first subdirectory must be a valid workspace.

### DIFF
--- a/src/main/src/main/java/org/geoserver/ows/StylePublisher.java
+++ b/src/main/src/main/java/org/geoserver/ows/StylePublisher.java
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2017 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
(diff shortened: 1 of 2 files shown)


jodygarnett:
Very cool, can I ask: How do you resolve conflict if workspace and icons folder have the same name?

We cam up with some ideas in today's meeting, just reporting here for reference:

* styles/folder#foo.png (treat a folder as a bulk icon factory - similar to a zip of icons?)
* styles/folder-foo.png - handle via naming convention
* styles/foo.png - flatten out the subfolders when searching for icon
* styles/conflict/foo.png - allow the conflict and just search both?
* styles/workspace/foo.png (work around is make a workspace with the expected “sub folder name”)

If I read your comment above correct you are doing:

* styles/workspace/icons/foo.png - only allows icons folder in workspace styles directory to prevent conflict

jodygarnett:
I reran the build, given that this partially fixes a regression I would like to merge and back port for the 2.11 release.

aaime:
" I would like to merge and back port for the 2.11 release." --> too late I'm afraid. But we can merge it in time for the 2.10.x release next month I believe?

sikeoka:
Some clarification about this pull request.  If there are no subdirectories, then the behavior doesn't change at all.  If there are subdirectories where the first subdirectory is not a workspace, then it reverts to pre-2.9.2 behavior of looking up the whole path in the global styles directory rather than returning an error.

If the first subdirectory is a workspace (e.g., styles/ws/a/b.png), it first looks up the path in the workspace.  If that resource doesn't exist, then it will look up the path without the workspace (e.g., a/b.png) in the global styles directory.  I'm not exactly sure what this is for but it is part of the current behavior so I left it in there.  These two lookups don't change from the current behavior except for adding support for additional subdirectories after the workspace name.  If neither of these paths find the resource, an additional lookup is performed on the entire path (e.g., ws/a/b.png) in the global styles directory.

There are no cases where a valid pre-2.9.2 URL would return an error response but it will be possible for a different resource to be returned only when the URL contains a workspace.  For the path like "styles/ws/a/b.png", both "workspaces/ws/styles/a/b.png" and "styles/a/b.png" would be returned instead of "styles/ws/a/b.png" if either path exists.

sikeoka:
FYI, this pull request can be backported to 2.10.x and 2.11.x.

aaime:
@jodygarnett are you satisfied with @sikeoka explanations?

jodygarnett:
I am satisfied, but nervous about this design due to the opportunity for conflict. The priory has to be fixing a regression, though so as indicated I would like to merge and backport.

jodygarnett:
backported to 2.11.x

PR #2175, opened 2017-03-22
Fix WFS-T Transaction InsertResults to return FeatureIDs in inserted order [GEOS-8042]

goudine:


src/wfs/src/main/java/org/geoserver/wfs/InsertElementHandler.java
```diff
@@ -271,7 +271,7 @@ public Class getElementClass() {
                 typeNames.add(new QName(namespaceURI, name));
             }
         } else {
-            LOGGER.finer("Insert was empty - does not need a FeatuerSoruce");
+            LOGGER.finer("Insert was empty - does not need a FeatureSource");
         }
```
jodygarnett: I do not see a fix here - only changes to logging, and test cases confirming the other commit worked.

jodygarnett:
This looks good, I assume you need to wait for https://github.com/geotools/geotools/pull/1518 to be merged

src/wfs/src/test/java/org/geoserver/wfs/TransactionTest.java
```diff
@@ -438,4 +438,88 @@ public void testUpdateWithDifferentPrefix() throws Exception {
...
+                + "</cgf:Lines>" + "</wfs:Insert>"
+                + "<wfs:Insert > " + "<cgf:Lines>" + "<cgf:lineStringProperty>"
+                + "<gml:LineString>" + "<gml:coordinates decimal=\".\" cs=\",\" ts=\" \">"
+                + "494475.710564120,5433016.8189323 494982.70115655,5435041.95096670"
+                + "</gml:coordinates>" + "</gml:LineString>"
+                + "</cgf:lineStringProperty>" + "<cgf:id>t0002</cgf:id>"
+                + "</cgf:Lines>" + "</wfs:Insert>"
+                + "</wfs:Transaction>";
+        Document dom = postAsDOM("wfs", features);
+        print(dom);
```
aaime: Remove the printouts when done

src/wfs/src/test/java/org/geoserver/wfs/TransactionTest.java
```diff
@@ -438,4 +438,88 @@ public void testUpdateWithDifferentPrefix() throws Exception {
...
+        XMLAssert.assertXpathEvaluatesTo("new11", "//wfs:InsertResult/ogc:FeatureId[last()]/@fid", dom);
```
aaime: Asserting a single element does not make for order guarantee

aaime:
Targets the wrong branch (should be master, not 2.9.x), Travis reports failures in WFS-T, which _may_ be caused by this change, but I'm not sure.

src/wfs/src/test/java/org/geoserver/wfs/TransactionTest.java
jodygarnett: In this case it did because of [last()], default feature collection was sorting new1, new11, new2, ...

aaime: It just proves a different order, does not prove the right one :-)

src/wfs/src/main/java/org/geoserver/wfs/InsertElementHandler.java
goudine: Just a spelling correction in the logging message

mtCarto:
@aaime Thank you for the review, we will change this to target master. This doesn't need to hold up the release, as we plan on backporting to satisfy our requirements.

src/wfs/src/test/java/org/geoserver/wfs/TransactionTest.java
jodygarnett: I guess to do that we would need a GetFeature using the last fid, and double check the attribute values.

If the datastore used for this test would respect the provided FIDs this would be much easier. Do we know what is behind this test?

goudine:
New PR was submitted to master, with the comments addressed.
Link can be found here:
https://github.com/geoserver/geoserver/pull/2179

PR #2177, opened 2017-03-22
[GEOS-8046] Fixed ProcessParameterIO performing extension lookups with a null ApplicationContext.

sikeoka:
Fix with unit test for:
https://osgeo-org.atlassian.net/browse/GEOS-8046

This bug only causes annoying but harmless warning messages in the logs if the environment does not contain any PPIOFactory implementations.  I've found two, one in gs-gdal-wps and another in gs-ogr-wps, but I don't use either plugin.

### DIFF
--- a/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ppio/ProcessParameterIO.java
+++ b/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ppio/ProcessParameterIO.java
@@ -1,4 +1,4 @@
-/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2017 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
@@ -184,8 +184,12 @@ public abstract class ProcessParameterIO {
         }
 
         // load by factory
-        List<PPIOFactory> ppioFactories = GeoServerExtensions
-                .extensions(PPIOFactory.class, context);
+        List<PPIOFactory> ppioFactories;
+        if (context != null) {
+            ppioFactories = GeoServerExtensions.extensions(PPIOFactory.class, context);
+        } else {
+            ppioFactories = GeoServerExtensions.extensions(PPIOFactory.class);
+        }
         for (PPIOFactory factory : ppioFactories) {
             l.addAll(factory.getProcessParameterIO());
         }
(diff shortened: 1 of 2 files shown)


src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ppio/ProcessParameterIO.java
```diff
@@ -184,8 +184,12 @@ public int compare(ProcessParameterIO o1, ProcessParameterIO o2) {
         }
 
         // load by factory
-        List<PPIOFactory> ppioFactories = GeoServerExtensions
-                .extensions(PPIOFactory.class, context);
+        List<PPIOFactory> ppioFactories;
+        if (context != null) {
+            ppioFactories = GeoServerExtensions.extensions(PPIOFactory.class, context);
```
jodygarnett: Is it worth making this method more relaxed, ie if null context is passed in avoid producing a warning? 

What category of bugs is this warning catching to justify having two methods (and this kind of extra logic required to "Safely" call it?)

aaime: Are you talking about GeoServerExtensions.extensions? The warning happens here:

https://github.com/geoserver/geoserver-history/blob/master/src/platform/src/main/java/org/geoserver/platform/GeoServerExtensions.java#L296

Check how you get there, from what I see the context checked is null only if the caller did pass a null context, or if the caller did not pass a context _and_ GeoSErverExtensions bean has not yet been initialized. Generally speaking, beans depending on a context-less call should declare a dependency on GeoServerExtensions (not all do, not all can). Ideally, it would be nice if GeoServerExtensions was the very first bean initialized in the spring context. There is an article about this here:
http://rachitskillisaurus.blogspot.it/2013/10/spring-force-bean-to-be-first-to.html

Maybe the above is a better option than this fix, along with GeoServerExtensions automatically using its own default context if null is passed.

sikeoka: I try to make pull requests small and isolated and changing the behavior or a core method that is probably called hundreds of times throughout the code in order to suppress warning messages in an extension that isn't breaking anything didn't seem like a good idea.  Also, GeoServerExtensions.extensions is called twice within this ProcessParameterIO method (about 5 lines above this change) and all this pull request does is make both calls handle null contexts in the same way.

I agree that GeoServerExtensions should try to use its own context if null is passed to it.  Even if you guys did decide to change GeoServerExtensions.extensions, it would have to be done in a way that doesn't break this pull request since I suspect that this isn't the only place in the code with this extra logic for handling null contexts.

sikeoka:
FYI, this pull request can be backported to 2.10.x and 2.11.x.

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ppio/ProcessParameterIO.java
aaime: @jodygarnett we are waiting for your reaction

jodygarnett:
I have taking the API question out to GEOS-8146, just rerunning travis to ensure this fix can still be merged cleanly.

@sikeoka I see you have made contributions to geotools and geoserver before; going to assume you have a CLA on file with OSGeo.

aaime:
@jodygarnett @sikeoka I can confirm there is a CLA on file now.

PR #2191, opened 2017-03-24
Include plugin for GeoGig in Travis builds

emerkle826:
GeoGig and the GeoServer plugin for GeoGig are currently under pretty heavy development. As such, API changes tend to cause breaks in the builds, which is particularly disruptive to GeoServer. For now, the plugin for GeoGig is being removed from the `communityRelease` module to keep GeoGig build issues from breaking GeoServer. However, it is desirable to have Travis build the GeoGig plugin when PRs are submitted (to catch build issues prior to merging things). This PR adds the geogig module to the Travis build to retain that functionality when PRs are created.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ language: java
 before_install:
   - rm ~/.m2/settings.xml
 script:
-  - mvn -f src/pom.xml -B -U -T2 -fae -Prelease clean install && mvn -f src/community/pom.xml -B -U -T2 -fae -DskipTests -Prelease -PcommunityRelease clean install
+  - mvn -f src/pom.xml -B -U -T2 -fae -Prelease clean install && mvn -f src/community/pom.xml -B -U -T2 -fae -DskipTests -Prelease -PcommunityRelease -Pgeogig clean install
 notifications:
   email:
     on_success: never


jodygarnett:
thanks, happy to merge when LT Build is restored. Please advise :)

emerkle826:
@jodygarnett Will do.

emerkle826:
@jodygarnett Looks like the LocationTech builds of GeoGig are running again. I re-triggered this PR Travis build. Let me know if anything else needs to be done.

bencaradocdavies:
@emerkle826, from [mailing list discussion](http://osgeo-org.1560.x6.nabble.com/GeoServer-builds-broken-by-failing-geogig-master-build-td5313969.html):

I am unhappy with this pull request because it makes Travis CI and ares Jenkins inconsistent. This affects all pull requests. I would rather have the ```geogig``` module back in the ```communityRelease``` profile and closely monitored by GeoGig developers. The recent failure was caused by an infrastructure problem not an API change (there was an API change but the corresponding fixes were immediately merged into GeoServer). I would rather have the ares build fail for a short time every now and then than break the consistency between Travis CI and ares Jenkins.

Another solution could be to make a stable GeoGig branch for GeoServer to consume and backport GeoGig API changes to this branch at your convenience. This will allow you to choose when to deal with API changes on the GeoServer side. You could achieve the same effect with GeoGig milestones or changing the GeoGig pom version. How much GeoGig API change is coming, and which approach do you think would be easiest?

I have no problem with including the ```geogig``` module in the ```communityRelease``` profile right now if it compiles and will in the future receive your usual close attention.

emerkle826:
As per Ben's and Ian's comments, closing this PR to avoid Travis and Jenkins build differences. I'll submit a PR to add the GeoGig module back into communityRelease once the GeoGig/LocationTech build infrastructure has had a few more days to ensure it's settled down.

PR #2192, opened 2017-03-27
[Rest API Refresh] Added doc generation pom at doc/en/en.xml

mattkrusz:


jodygarnett:
merged by hand, welcome to geoserver commit @mattkrusz

PR #2200, opened 2017-04-03
WMS - fix WMS GetCapabilities 1.3.0 compliance

pmauduit:
See https://osgeo-org.atlassian.net/browse/GEOS-8063 for the details

This fix proposal aims to add a Title element in the Style's GetCapabilities response.

Tests: At runtime, the testsuite onto the gs-wms module was also ok

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
+++ b/src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
@@ -1038,7 +1038,14 @@ public class Capabilities_1_3_0_Transformer extends TransformerBase {
                 }
                 element("Name", defaultStyle.prefixedName());
                 if (ftStyle.getDescription() != null) {
-                    element("Title", ftStyle.getDescription().getTitle());
+                    // PMT: WMS capabilities requires at least a title,
+                    // if description's title is null, use the name
+                    // for title.
+                    if (ftStyle.getDescription().getTitle() != null) {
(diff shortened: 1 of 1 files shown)


aaime:
Half a second check reveals two issues in the pull request:

- Should target master, not 2.11.x
- Must have a test

pmauduit:
Ok for rebasing, I will do it after having introduced a test. Anyway, I've got some questions around testing in GeoServer. The fastest way to test this for me will be to add an integration test (or else I would have used some reflection to access the modified private method). But I could not manage to find a way to test it using this kind of code snippet:

```java
    @Test
    public void testStyleWithoutTitle() throws Exception {
    	StyleInfo st = new StyleInfoImpl(getCatalog());
    	st.setFilename("styleWithoutTitle.sld");
    	st.setName("styleWithoutTitle");
    	st.getStyle().getDescription().setTitle((String) null);
    	getCatalog().save(st);

    	getCatalog().add(st);
        Document doc = getAsDOM("wms?service=WMS&request=getCapabilities&version=1.3.0", true);
        print(doc);
    }
```
sorry if the question sounds like a "do my homework", but do you have some advices where I can get some inspiration (another similar test) to do it in a geoserver fashion ?

aaime:
Normally one adds a style in the setup method using the TestData class, like in this example:

https://github.com/geoserver/geoserver/blob/master/src/wms/src/test/java/org/geoserver/wms/capabilities/GetCapabilitiesLegendURLTest.java#L121

aaime:
Closing and waiting for fixed pull request. The ticket links to it, so in case you don't make it, other interested people can resume the work from here.

PR #2201, opened 2017-04-03
[GEOS-8068] Add GeoJson encoder for complex features

nmco:
Adds a GeoJSON encoder for complex features. The workflow is similar to the GML one, if the response result only contain simple features the actual simple features GeoJSON encoder is used otherwise the complex features GeoJSON encoder will be used.

Tests for the complex GeoJSON encoder were added in the app-schema module since this is the only place were complex features can actually be produced.

Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-8068

### DIFF
--- a/src/extension/app-schema/app-schema-test/pom.xml
+++ b/src/extension/app-schema/app-schema-test/pom.xml
@@ -217,6 +217,7 @@
 							<include>**/*.sql</include>
 							<include>**/*.sld</include>
 			   				<include>**/*.png</include>
+                  <include>**/*.json</include>
 					    </includes>
 			        </testResource>
 		        </testResources>
@@ -269,6 +270,7 @@
 				    	    <include>**/*.sql</include>
 					    	<include>**/*.sld</include>
 					    	<include>**/*.png</include>
+									<include>**/*.json</include>
 				        </includes>
 			        </testResource>
 		        </testResources>
--- a/src/extension/app-schema/app-schema-test/src/test/resources/test-data/Observation.json
+++ b/src/extension/app-schema/app-schema-test/src/test/resources/test-data/Observation.json
@@ -1,0 +1,757 @@
+{
+  "crs": null,
+  "features": [
+    {
+      "geometry": null,
+      "id": "observation:mf1",
+      "properties": {
+        "metadata": {
+          "CGI_NumericValue": {
+            "principalValue": 651
+          }
+        },
+        "result": [
+          {
+            "name": "GUNTHORPE FORMATION",
+            "observationMethod": {
+              "CGI_TermValue": {
+                "value": "urn:ogc:def:nil:OGC::missing"
+              }
+            },
+            "positionalAccuracy": {
+              "CGI_NumericValue": {
+                "principalValue": 200
+              }
+            },
+            "samplingFrame": {},
+            "shape": {
+              "coordinates": [
+                [
+                  [
+                    52.5,
+                    -1.2
+                  ],
+                  [
+                    52.6,
+                    -1.2
+                  ],
+                  [
+                    52.6,
+                    -1.1
+                  ],
+                  [
+                    52.5,
+                    -1.1
+                  ],
+                  [
+                    52.5,
+                    -1.2
+                  ]
+                ]
+              ],
+              "type": "Polygon"
+            },
+            "specification": {
+              "GeologicUnit": {
+                "composition": [
+                  {
+                    "lithology": [
+                      {
+                        "name": [
+                          "name_cc_5",
+                          "name_cc_5"
+                        ],
+                        "vocabulary": {}
+                      }
+                    ],
+                    "proportion": {
+                      "CGI_TermValue": {
+                        "value": "nonexistent"
+                      }
+                    },
+                    "role": "fictitious component"
+                  }
+                ],
+                "description": "Olivine basalt, tuff, microgabbro, minor sedimentary rocks",
+                "exposureColor": {
+                  "CGI_TermValue": {
+                    "value": "Blue"
+                  }
+                },
+                "geologicUnitType": {},
+                "name": [
+                  "Yaugher Volcanic Group",
+                  "-Py"
+                ],
+                "observationMethod": {
+                  "CGI_TermValue": {
+                    "value": "urn:ogc:def:nil:OGC::missing"
+                  }
+                },
+                "occurrence": {},
+                "outcropCharacter": {
+                  "CGI_TermValue": {
+                    "value": "x"
+                  }
+                },
+                "purpose": "instance"
+              }
+            }
+          }
+        ],
+        "resultQuality": {
+          "GeologicUnit": {
+            "composition": [
+              {
+                "lithology": [
+                  {
+                    "name": [
+                      "name_cc_5",
+                      "name_cc_5"
+                    ],
+                    "vocabulary": {}
+                  }
+                ],
+                "proportion": {
+                  "CGI_TermValue": {
+                    "value": "nonexistent"
+                  }
+                },
+                "role": "fictitious component"
+              }
+            ],
+            "description": "Olivine basalt, tuff, microgabbro, minor sedimentary rocks",
+            "exposureColor": {
+              "CGI_TermValue": {
+                "value": "Blue"
+              }
+            },
+            "geologicUnitType": {},
+            "name": [
+              "Yaugher Volcanic Group",
+              "-Py"
+            ],
+            "observationMethod": {
+              "CGI_TermValue": {
+                "value": "urn:ogc:def:nil:OGC::missing"
+              }
+            },
+            "occurrence": {},
+            "outcropCharacter": {
+              "CGI_TermValue": {
+                "value": "x"
+              }
+            },
+            "purpose": "instance"
+          }
+        }
+      },
+      "type": "Feature"
+    },
+    {
+      "geometry": null,
+      "id": "observation:mf2",
+      "properties": {
+        "metadata": {
+          "CGI_NumericValue": {
+            "principalValue": 269
+          }
+        },
+        "result": [
+          {
+            "name": "MERCIA MUDSTONE GROUP",
+            "observationMethod": {
+              "CGI_TermValue": {
+                "value": "urn:ogc:def:nil:OGC::missing"
+              }
+            },
+            "positionalAccuracy": {
+              "CGI_NumericValue": {
+                "principalValue": 100
+              }
+            },
+            "samplingFrame": {},
+            "shape": {
+              "coordinates": [
+                [
+                  [
+                    52.5,
+                    -1.3
+                  ],
+                  [
+                    52.6,
+                    -1.3
+                  ],
+                  [
+                    52.6,
+                    -1.2
+                  ],
+                  [
+                    52.5,
+                    -1.2
+                  ],
+                  [
+                    52.5,
+                    -1.3
+                  ]
+                ]
+              ],
+              "type": "Polygon"
+            },
+            "specification": {
+              "GeologicUnit": {
+                "composition": [
+                  {
+                    "lithology": [
+                      {
+                        "name": [
+                          "name_cc_3",
+                          "name_cc_3"
+                        ],
+                        "vocabulary": {}
+                      }
+                    ],
+                    "proportion": {
+                      "CGI_TermValue": {
+                        "value": "significant"
+                      }
+                    },
+                    "role": "interbedded component"
+                  },
+                  {
+                    "lithology": [
+                      {
+                        "name": [
+                          "name_cc_4",
+                          "name_cc_4"
+                        ],
+                        "vocabulary": {}
+                      }
+                    ],
+                    "proportion": {
+                      "CGI_TermValue": {
+                        "value": "minor"
+                      }
+                    },
+                    "role": "interbedded component"
+                  }
+                ],
+                "description": "Olivine basalt, tuff, microgabbro, minor sedimentary rocks",
+                "exposureColor": [
+                  {
+                    "CGI_TermValue": {
+                      "value": "Blue"
+                    }
+                  },
+                  {
+                    "CGI_TermValue": {
+                      "value": "Yellow"
+                    }
+                  }
+                ],
+                "geologicUnitType": {},
+                "name": [
+                  "Yaugher Volcanic Group 1",
+                  "Yaugher Volcanic Group 2",
+                  "-Py"
+                ],
+                "observationMethod": {
+                  "CGI_TermValue": {
+                    "value": "urn:ogc:def:nil:OGC::missing"
+                  }
+                },
+                "occurrence": [
+                  {},
+                  {}
+                ],
+                "outcropCharacter": [
+                  {
+                    "CGI_TermValue": {
+                      "value": "x"
+                    }
+                  },
+                  {
+                    "CGI_TermValue": {
+                      "value": "y"
+                    }
+                  }
+                ],
+                "purpose": "instance"
+              }
+            }
+          }
+        ],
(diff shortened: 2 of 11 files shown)


bencaradocdavies:
@nmco several tests fail [on Travis CI](https://s3.amazonaws.com/archive.travis-ci.org/jobs/218160786/log.txt):
```
Failed tests:   testAnyTypeAndAnyElementJSON(org.geoserver.test.FeatureChainingWfsTest): (..)
  testNonValidNestedJSON(org.geoserver.test.FeatureChainingWfsTest): (..)
  testComplexTypeWithSimpleContentJSON(org.geoserver.test.FeatureChainingWfsTest): (..)
  testEncodeFeatureMembersJSON(org.geoserver.test.FeatureChainingWfsTest): (..)
  testGetFeatureJSON(org.geoserver.test.FeatureChainingWfsTest): (..)
```

nmco:
Thanks Ben, yeah basically the tests I add for the GeoJSON encoder are all failing which is not happening locally ... investigating.

nmco:
Found the issue I added the pattern to include JSON resources only in the online profile and forgot the offline profile :scream: since I was using the online profile I didn't got the tests errors, the build should pass now.

bencaradocdavies:
@nmco online tests fail against postgis (and I suspect would also fail against oracle):
```
Failed tests:   testAnyTypeAndAnyElementJSON(org.geoserver.test.FeatureChainingWfsTest): (..)
  testNonValidNestedJSON(org.geoserver.test.FeatureChainingWfsTest): (..)
  testComplexTypeWithSimpleContentJSON(org.geoserver.test.FeatureChainingWfsTest): (..)
  testEncodeFeatureMembersJSON(org.geoserver.test.FeatureChainingWfsTest): (..)
  testGetFeatureJSON(org.geoserver.test.FeatureChainingWfsTest): (..)
```
I guess that you need to include the ```*.json``` pattern ```testResources``` ```includes``` in:
```
extension/app-schema/app-schema-postgis-test/pom.xml
extension/app-schema/app-schema-oracle-test/pom.xml
```

Also, I apologies for the formatting of these poms. They are a mess of tabs and spaces. I promise to reformat all ```app-schema-test``` poms on master and 2.11.x after your PR is merged.

src/extension/app-schema/app-schema-test/src/test/resources/test-data/Observation.json
```diff
@@ -0,0 +1,332 @@
...
...Impl:value<CodeWithAuthorityType>=[AttributeImpl:simpleContent<string>=nonexistent]]], ComplexAttributeImpl:lithology<ControlledConceptPropertyType>=[FeatureImpl:ControlledConcept<ControlledConceptType id=cc.5>=[ComplexAttributeImpl:name<CodeType>=[AttributeImpl:simpleContent<string>=name_cc_5], ComplexAttributeImpl:name<LocalizedGenericNameType>=[AttributeImpl:simpleContent<string>=name_cc_5], ComplexAttributeImpl:vocabulary<GeologicVocabularyPropertyType>=[]]]]], ComplexAttributeImpl:exposureColor<CGI_TermValuePropertyType>=[FeatureImpl:CGI_TermValue<CGI_TermValueType id=Blue>=[ComplexAttributeImpl:value<CodeWithAuthorityType>=[AttributeImpl:simpleContent<string>=Blue]]], ComplexAttributeImpl:outcropCharacter<CGI_TermValuePropertyType>=[FeatureImpl:CGI_TermValue<CGI_TermValueType id=x>=[ComplexAttributeImpl:value<CodeWithAuthorityType>=[AttributeImpl:simpleContent<string>=x]]], ComplexAttributeImpl:occurrence<MappedFeaturePropertyType>=[]]], ComplexAttributeImpl:observationMethod<CGI_TermValuePropertyType>=[ComplexAttributeImpl:CGI_TermValue<CGI_TermValueType>=[ComplexAttributeImpl:value<CodeWithAuthorityType>=[AttributeImpl:simpleContent<string>=urn:ogc:def:nil:OGC::missing]]]]"
```
bencaradocdavies: @nmco there are four instances of ```result``` in ```Observation.json``` that are being encoded as the Java ```toString``` of the nested feature. This must be wrong.

src/extension/app-schema/app-schema-test/pom.xml
```diff
@@ -217,6 +217,7 @@
 							<include>**/*.sql</include>
 							<include>**/*.sld</include>
 			   				<include>**/*.png</include>
+                  <include>**/*.json</include>
```
nmco: Ok thanks !

src/wfs/src/main/java/org/geoserver/wfs/json/ComplexGeoJsonWriter.java
```diff
@@ -0,0 +1,377 @@
...
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * GeoJSON writer capable of handling complex features.
+ */
+final class ComplexGeoJsonWriter {
```
nmco: True I would add the public modifier (apparently when writing code I'm a very conservative person :wink:).

src/wfs/src/main/java/org/geoserver/wfs/json/ComplexGeoJsonWriter.java
```diff
@@ -0,0 +1,377 @@
...
+     * Encode a simple attribute, this means that this property
+     * will be encoded as a simple JSON attribute.
+     */
+    private void encodeSimpleAttribute(String name, Object value) {
+        // add a simple JSON attribute to the current object
+        jsonWriter.key(name).value(value);
+    }
+
+    /**
+     * Return TRUE fi a geometry was found during the features collections encoding.
```
nmco: Fixed, thanks !

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
```diff
@@ -125,6 +138,13 @@ protected void write(FeatureCollectionResponse featureCollection, OutputStream o
             if (jsonp) {
                 outWriter.write(getCallbackFunction() + "(");
             }
+
+            // currently complex features count always return zero
+            boolean isComplex = isComplexFeature(featureCollection);
+            if (featureCount != null && isComplex && featureCount.compareTo(BigInteger.ZERO) == 0) {
```
nmco: Agree.

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
```diff
@@ -290,6 +225,123 @@ else if ( id_option.length() != 0){
         }
     }
 
+    public class FeaturesInfo {
```
nmco: So let's make it private ! 
Jokes apart this class should be private and I will add some Java doc anyway, thanks.

src/extension/app-schema/app-schema-test/pom.xml
nmco: Yeah I didn't respect the boy scout rule here :disappointed: fixing it.

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
```diff
@@ -290,6 +225,123 @@ else if ( id_option.length() != 0){
...
+            this.crs = crs;
+            this.hasGeometry = hasGeometry;
+        }
+    }
+
+    private FeaturesInfo encodeSimpleFeatures(GeoJSONBuilder jsonWriter, List<FeatureCollection> resultsList,
+                                                           String id_option, boolean featureBounding) {
+        CoordinateReferenceSystem crs = null;
+        boolean hasGeom = false;
+        for (int i = 0; i < resultsList.size(); i++) {
```
nmco: Yeah I didn't respect the boy scout rule here :disappointed: fixing it.

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
```diff
@@ -290,6 +225,123 @@ else if ( id_option.length() != 0){
...
+            FeatureCollection collection = resultsList.get(i);
+            FeatureIterator iterator = collection.features();
+
```
nmco: Fixed.

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
```diff
@@ -290,6 +225,123 @@ else if ( id_option.length() != 0){
...
+
+            try {
```
nmco: Fixed.

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
```diff
@@ -290,6 +225,123 @@ else if ( id_option.length() != 0){
...
+                while (iterator.hasNext()) {
+
+                    SimpleFeature simpleFeature = (SimpleFeature) iterator.next();
+                    jsonWriter.object();
+                    jsonWriter.key("type").value("Feature");
+
+                    fType = simpleFeature.getFeatureType();
+                    types = fType.getAttributeDescriptors();
+
+                    if( id_option == null ){
```
nmco: Fixed.

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
```diff
@@ -290,6 +225,123 @@ else if ( id_option.length() != 0){
...
+                        jsonWriter.key("id").value(simpleFeature.getID());
+                    }
+                    else if ( id_option.length() != 0){
```
nmco: Fixed.

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
```diff
@@ -290,6 +225,123 @@ else if ( id_option.length() != 0){
...
+                    }
+
+                    GeometryDescriptor defaultGeomType = fType.getGeometryDescriptor();
+                    if(defaultGeomType != null) {
+                        CoordinateReferenceSystem featureCrs =
+                                defaultGeomType.getCoordinateReferenceSystem();
+
+                        jsonWriter.setAxisOrder(CRS.getAxisOrder(featureCrs));
+
+                        if (crs == null)
```
nmco: Fixed.

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
```diff
@@ -290,6 +225,123 @@ else if ( id_option.length() != 0){
...
+                            // decided to 'let convention evolve',
+                            // that is how to handle multiple
+                            // geometries. My take is to print the
+                            // geometry here if it's not the default.
+                            // If it's the default that you already
+                            // printed above, so you don't need it here.
+                            if (ad.equals(defaultGeomType)) {
+                                // Do nothing, we wrote it above
+                                // jsonWriter.value("geometry_name");
+                            } else if(value == null){
```
nmco: Fixed.

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
```diff
@@ -290,6 +225,123 @@ else if ( id_option.length() != 0){
...
+                    }
+                    // Bounding box for feature in properties
+                    ReferencedEnvelope refenv = ReferencedEnvelope.reference(simpleFeature.getBounds());
+                    if (featureBounding && !refenv.isEmpty())
+                        jsonWriter.writeBoundingBox(refenv);
+
+                    jsonWriter.endObject(); // end the properties
+                    jsonWriter.endObject(); // end the feature
+                }
+            } // catch an exception here?
```
nmco: Fixed.

src/extension/app-schema/app-schema-test/src/test/resources/test-data/Observation.json
nmco: Ah! I remember this one, the XML contains the same thing I was not sure if I should consider this a issue.

The XML result for the same query: [observation.txt](https://github.com/geoserver/geoserver/files/907592/observation.txt).

nmco: I case you are wondering why I upload the XML file with a TXT extension, it was because GIT refused the XML extension.

nmco:
Thanks for the detailed review @bencaradocdavies, I amend has requested and runned online tests with PostGIS.

src/wfs/src/main/java/org/geoserver/wfs/json/ComplexGeoJsonWriter.java
aaime: I actually use package private any time a class does not really need to be public. In java that's not so easy, but imho best not to make a class reusable when it's not meant to be reused. It may just be that I feel I'm trying to carry to much on my shoulders already and just trying to reduce damage ;-)

nmco: Agree. I prefer having to make a class public because someone needs to use it \ extend it than worrying about possible strange usages \ extensions of that class when doing a refactoring and actually nobody is using \ extending it. But indeed this kind of class are not very common in GS \ GT.

src/extension/app-schema/app-schema-test/src/test/resources/test-data/Observation.json
bencaradocdavies: @nmco I have just fixed this bug ([GEOT-5696](https://osgeo-org.atlassian.net/browse/GEOT-5696)) on GeoTools master and added test coverage on GeoServer master. You will need to update your reference JSON text.

bencaradocdavies:
@nmco I have restarted the Travis CI build because it should now fail with the [GEOT-5696](https://osgeo-org.atlassian.net/browse/GEOT-5696) fix applied to GeoTools master.

bencaradocdavies:
@nmco the Travis CI build is still passing. I think your JSON writer implementation  needs to be updated to remove the FeatureImpl list toString.

src/wfs/src/main/java/org/geoserver/wfs/json/ComplexGeoJsonWriter.java
bencaradocdavies: Yes, there is nothing wrong with a package-private class, and it can be a good idea to limit the visibility of code that could be considered new API, but implementation helper classes are often implemented as nested classes in GeoTools/GeoServer. In my view, if a class merits a separate file, it might be useful enough to be public.  :-)

aaime: @bencaradocdavies that works fine if the containing class is small, or if the contained class is small, lately I started getting a bit allergic to very large files (yes yes, StreamingRenderer, I know, I know...)

src/extension/app-schema/app-schema-test/src/test/resources/test-data/Observation.json
nmco: Thanks, JSON updated and GeoJSON encoder updated.

nmco:
Tests case and GeoJSON encoder updated.

PR #2203, opened 2017-04-04
[GEOS-8072] [GEOS-5176] WMS 1.3.0 capabilities global bounding box computation doesn't use layer groups

nmco:
Makes WMS 1.3.0 capabilities transform use layer groups to compute the global bounding box. This pull request also adds a test case.

Makes WMS capabilities also filter layer groups by namespaces like layers a test case for this is also added.

Associated issues: 
- https://osgeo-org.atlassian.net/browse/GEOS-8072
- https://osgeo-org.atlassian.net/browse/GEOS-5176

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
+++ b/src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
@@ -658,13 +658,7 @@ public class Capabilities_1_3_0_Transformer extends TransformerBase {
             }
             
             // filter the layers if a namespace filter has been set
-            if (request.getNamespace() != null) {
-                //build a query predicate for the namespace prefix
-                final String nsPrefix = request.getNamespace();
-                final String nsProp = "resource.namespace.prefix";
-                Filter equals = equal(nsProp, nsPrefix);
-                filter = and(filter, equals);
-            }
+            filter = addNameSpaceFilterIfNeed(filter, "resource.namespace.prefix");
 
             final Catalog catalog = wmsConfig.getCatalog();
                         
@@ -682,10 +676,17 @@ public class Capabilities_1_3_0_Transformer extends TransformerBase {
             }
             handleRootCrsList(srs);
 
-            CloseableIterator<LayerInfo> layers;
-            layers = catalog.list(LayerInfo.class, filter);
+            // create layer groups filter
+            Filter lgFilter = Predicates.acceptAll();
+
+            // filter layer groups by namespace if needed
+            lgFilter = addNameSpaceFilterIfNeed(lgFilter, "workspace.name");
+
+            // handle root bounding box
+            CloseableIterator<LayerInfo> layers = catalog.list(LayerInfo.class, filter);
+            CloseableIterator<LayerGroupInfo> layerGroups = catalog.list(LayerGroupInfo.class, lgFilter);
             try{
-                handleRootBbox(layers);
+                handleRootBbox(layers, layerGroups);
             }finally{
                 layers.close();
             }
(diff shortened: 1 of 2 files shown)


src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
```diff
@@ -682,10 +682,14 @@ private void handleLayers() {
             }
             handleRootCrsList(srs);
 
-            CloseableIterator<LayerInfo> layers;
-            layers = catalog.list(LayerInfo.class, filter);
+            // create layer groups filter
+            final Filter lgFilter = Predicates.acceptAll();
```
aaime: A layer group can be part of a workspace, and a "namespace" filter actually filters on the workspace name, so, I'd factor in at least this bit.

src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
```diff
@@ -785,32 +787,49 @@ private String qualifySRS(String srs) {
...
-                if(latlonBbox.contains(world)){
+            // handle layer groups
+            while (layersGroups.hasNext()) {
+                LayerGroupInfo layerGroup = layersGroups.next();
+                ReferencedEnvelope referencedEnvelope = layerGroup.getBounds();
+                if (referencedEnvelope == null) {
+                    // no bounds available move on
+                    continue;
+                }
+                if (!referencedEnvelope.getCoordinateReferenceSystem().equals(DefaultGeographicCRS.WGS84)) {
```
aaime: Use CRS.equalsIgnoreMetadata instead

aaime:
One thing, just thinking out loud here, no need for changes, but: to achieve the testing goal, wouldn't it have been easier to just loop over all layers, make them un-advertised, run the test (I believe there is already a layer group), and then make them advertised again?

nmco:
That was my first idea, then I convince myself that I wanted to make sure that I used that exact layer group with that exact bounds and didn't want to depend on other tests changing the catalog (yeah I know ti should not happen).

src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
nmco: I'm a little lost here, basically you mean using the same filter that is use for layers to filter the namespace ? This will also change the current behavior of which layer groups are shown in the capabilities is that something we want ?

src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
nmco: Ah! Thanks didn't know about this.

src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
aaime: I believe so, yes, the filtering should be uniform. When originally implemented workspace specific groups did not exist, and when they were added, this bit was forgotten about.

aaime:
As said, I was just thinking out loud, no problems if the code just stays as is (and less waste of time for you).

nmco:
Thanks for the review Andrea, I applied the requested changes. Layer groups will also filter by name space like layers, a test case was added for this (some of the code I wrote for the other test was useful for this :wink:).

src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
```diff
@@ -682,10 +676,17 @@ private void handleLayers() {
             }
             handleRootCrsList(srs);
 
-            CloseableIterator<LayerInfo> layers;
-            layers = catalog.list(LayerInfo.class, filter);
+            // create layer groups filter
+            Filter lgFilter = Predicates.acceptAll();
+
+            // filter layer groups by namespace if needed
+            addNameSpaceFilterIfNeed(filter, "workspace.name");
```
aaime: Shouldn't it have been "addNameSpaceFilterIfNeed(lgFilter, "workspace.name");"

nmco: Thanks Andrea, indeed this was wrong (like really wrong) and I had an error in my test case that was not detecting this issue :disappointed:.

nmco:
Pull request amended thanks Andrea for the review.

aaime:
+1 for merge and backport

PR #2204, opened 2017-04-05
Change to use of AbstractHttpMessageConverter for mimeType handling

jodygarnett:


jodygarnett:
This pull request is not ready to merge (not all classes compile). Placing it here for review and feedback.

What is good about AbstractHttpMessageConverter:
- protected methods to allow for fine grain control of canRead and canWrite
- base implementations of read and write that do a consistent job of the headers
- writeInternal and readInternal methods that we can move our code to

What is difficult about AbstractHttpMessageConverter:
- The class is set up with generics to convert a single kind of object (we have a few general purpose converters)
- Where we break our converters into subclasses for different mimeTypes, AbstractHttpMessageConverter expects subclasses for different objects. This results in more classes.
- ImportContextJSONConverterReader are acting as both a builder parsing json, and a converter

aaime:
If that's the case, could we steal only part of the abstract class and place it in BaseMessageConverter? The part that sets the header in particular... thinking out loud:
- mark the write method as final
- roll a method to set the content type, receiving the same params as the write method, and make it overridable for cases in which there is more than one content type or complex logic
- add an abstract method to actual do the write (writeInternal?) that takes the same params as write

src/rest-ng/src/main/java/org/geoserver/rest/converters/XStreamJSONMessageConverter.java
```diff
@@ -26,7 +26,7 @@
     static final MediaType TEXT_JSON = MediaType.valueOf("text/json");
 
     public XStreamJSONMessageConverter() {
-        super();
+        super(MediaType.APPLICATION_JSON);
```
tbarsballe: Shouldn't we also be passing text/json here?

src/restconfig-ng/src/main/java/org/geoserver/catalog/rest/CatalogController.java
```diff
@@ -36,7 +36,13 @@
     /**
      * Not an official MIME type, but GeoServer used to support it
      */
+    public static final MediaType MEDIATYPE_TEXT_JSON = MediaType.valueOf(CatalogController.TEXT_JSON);
```
tbarsballe: We should probably actually move all of these into their own class, and match the MediaType naming scheme, e.g:
```
TEXT_JSON
TEXT_JSON_VALUE
TEXT_XML
TEXT_XML_VALUE
```

jodygarnett:
This is now ready to merge to rest-api-refresh

PR #2212, opened 2017-04-07
[GEOS-8050] Preserve external graphic URL #fragments during GeoServerDataDirectory.parsedStyle(..)

mattkrusz:
This PR adds support for preservation of external graphic URL fragments during the load of a parsedStyle.

When GeoServerDataDirectory loads a parsedStyle, it locates external graphics (while preserving Cql templates) using only the URL base path, ignoring query parameters and URL #fragments. It then attempts to preserve the original URL by re-adding query parameters, but fails to re-add the fragment.

Supporting URL fragments would be generally useful to designate subordinate resources, e.g., an individual icon inside a sprite sheet. This pattern has already been established for [bulk WKT shapes](http://docs.geoserver.org/stable/en/user/styling/sld/extensions/pointsymbols.html#bulk-wkt-shapes), for which you can write: `<OnlineResource xlink:type="simple" xlink:href="example.properties#zig" />`

### DIFF
--- a/src/main/src/main/java/org/geoserver/config/GeoServerDataDirectory.java
+++ b/src/main/src/main/java/org/geoserver/config/GeoServerDataDirectory.java
@@ -1211,14 +1211,27 @@ public class GeoServerDataDirectory {
                         //GEOS-7025: Just get the path; don't try to create the file
                         file = Paths.toFile(root(), resource.path());
                     }
+                    
                     URL u = fileToUrlPreservingCqlTemplates(file);
+                    
                     if (url.getQuery() != null) {
                         try {
-                            return new URL(u.toString() + "?" + url.getQuery());
+                            u = new URL(u.toString() + "?" + url.getQuery());
                         } catch (MalformedURLException ex) {
+                            GeoServerPersister.LOGGER.log(Level.WARNING, "Error processing query string for resource with uri: " + uri, ex);
                             return null;
                         }
                     }
+                    
+                    if (url.getRef() != null) {
+                        try {
+                            u = new URL(u.toString() + "#" + url.getRef());
+                        } catch (MalformedURLException ex) {
+                            GeoServerPersister.LOGGER.log(Level.WARNING, "Error processing # fragment for resource with uri: " + uri, ex);
+                            return null;
+                        }
+                    }
+                    
                     return u;
                 } else {
                     return url;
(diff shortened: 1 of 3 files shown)


dvntucker:
It looks sensible enough to me. Where along the way is the URL fragment being lost in the first place?

mattkrusz:
Hi Devon, thanks for taking a look It's being lost inside that `Resource resource = resourceLoader.fromURL(url);` method [here](https://github.com/mattkrusz/geoserver/blob/018cc4ca8989e1de2b17739d351630311c4c8060/src/main/src/main/java/org/geoserver/config/GeoServerDataDirectory.java#L1205). That method is only using the "base" of the url to find the resource. I didn't want to make a change there, (1) because I thought there was a greater chance of unintended consequences and (2) because there was already similar code that re-added the query parameters, so I thought I would build on that to re-add the #ref part.

src/main/src/main/java/org/geoserver/config/GeoServerDataDirectory.java
```diff
@@ -1211,14 +1211,27 @@ public URL locateResource(String uri) {
...
+                            GeoServerPersister.LOGGER.log(Level.WARNING, "Error processing query string for resource with uri: " + uri, ex);
                             return null;
                         }
                     }
+                    
+                    if (url.getRef() != null) {
+                        try {
+                            u = new URL(u.toString() + "#" + url.getRef());
+                        } catch (MalformedURLException ex) {
+                            GeoServerPersister.LOGGER.log(Level.WARNING, "Error processing # fragment for resource with uri: " + uri, ex);
```
jodygarnett: I never quite know what level to log these warnings as, since this is a problem based on styling (under user control) I agree warning is appropriate.

src/main/src/test/resources/org/geoserver/config/external_with_params_and_fragment.sld
```diff
@@ -0,0 +1,30 @@
...
+      <Name>external</Name> 
+      <Title>An external graphic</Title>
+      <Abstract>A sample point symbolizer with an external graphic with a URL containing both query parameters and a # fragment</Abstract>
+      <FeatureTypeStyle>
+        <Rule>
+          <Title>Red flag</Title>
+          <PointSymbolizer>
+            <Graphic>
+              <ExternalGraphic>
+                <OnlineResource xlink:type="simple" xlink:href="icon.png?param1=1#textAfterHash" />
```
jodygarnett: Looks good, I assume you got this example from the sprite case. The wkt.properties#symbol from the docs is the one we are trying to fix.

Suggestion only no need to change.

jodygarnett:
@mattkrusz I would like to merge this fix, and backport to 2.11.x and 2.10.x.

mattkrusz:
Thanks, Jody. Thumbs up from my end.

PR #2213, opened 2017-04-07
[GEOS-8077] Fixed custom dimension values being dropped from GetMap URLs in the response KML document.

sikeoka:


src/wms/src/main/java/org/geoserver/wms/WMSRequests.java
```diff
@@ -1,4 +1,4 @@
-/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
```
aaime: Our copyright handling procedures do not demand anymore update of years in the header, just that a header needs to be there. No need for change, just letting you know for the future.

sikeoka:
I actually emailed it to info@osgeo.org about a month ago.

aaime:
I did not get any updates from OSGeo, @jodygarnett do you know who's managing the contribution agreements these days? Still Jachim?

PR #2214, opened 2017-04-07
[GEOS-8023] Fixed Virtual Service layer name lookup not using workspace qualified layer names.

sikeoka:

Also, removes a redundant layer group lookup.


sikeoka:
It appears to be failing on this unit test:
https://github.com/geoserver/geoserver/blob/master/src/wfs/src/test/java/org/geoserver/wfs/GetFeatureTest.java#L313

The layer sf:Fifteen doesn't exist.  I could change it to a layer that actually exists in the "sf" workspace.

PR #2215, opened 2017-04-07
Remove FeatureTypeName tag from default raster SLD template

travislbrundage:
This PR is meant to address [GEOS-8079](https://osgeo-org.atlassian.net/browse/GEOS-8079)

The problem is caused by the `<FeatureTypeName>` tag being set arbitrarily to "Raster" which GeoServer does not understand for any raster layer.

My instinct was just to remove the line, in order to make the default template simpler and more generic. Custom styles can still be defined with their own `<FeatureTypeName>` such that they'll work and render correctly, but the default style should work generically.

### DIFF
--- a/src/main/src/main/resources/org/geoserver/catalog/template_raster.sld
+++ b/src/main/src/main/resources/org/geoserver/catalog/template_raster.sld
@@ -6,7 +6,6 @@ http://schemas.opengis.net/sld/1.0.0/StyledLayerDescriptor.xsd" version="1.0.0">
     <UserStyle>
       <Title>A raster style</Title>
       <FeatureTypeStyle>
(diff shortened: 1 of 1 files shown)


aaime:
+1

aaime:
Btw, ideally this should come with a test, to avoid regressions

aaime:
I believe I just merged a pull request by @tbarsballe fixing the same thing?

travislbrundage:
Yeah, this looks like it addresses this issue, I assume you mean this PR: https://github.com/geoserver/geoserver/pull/2279

aaime:
Yep, that's the one.

travislbrundage:
That addresses it more fully, so I will close this PR - thanks aaime!

PR #2217, opened 2017-04-09
[GEOS-7946] Using java.nio.file.Files.move() with ATOMIC_MOVE option …

ridethepenguin:
…to avoid case sensitivity issues


aaime:
@ridethepenguin so how is this pull request related to the @cmangeat  one? Does it have to be applied on top of it? Or should I revert the former and apply just this one?

ridethepenguin:
@aaime Let's say it supersedes it :-) I rebased on master, so you can merged it without reverting the previous one.

aaime:
Awesome. +1 then. Can you then merge and take care of the backports  If memory serves you do have commit rights.

ridethepenguin:
Merged on master, backports will hopefully come by the end of the day.

PR #2216, opened 2017-04-08
Rest API Conversion to Spring MVC with API Documentation [GEOS-7931]

tbarsballe:
See https://github.com/geoserver/geoserver/wiki/REST-API-Refresh for more details.

* https://osgeo-org.atlassian.net/browse/GEOS-7931
Improved REST API Documentation

PR #2218, opened 2017-04-10
Re-enable GeoGig plugin in communityRelease module

emerkle826:
The LocationTech builds for GeoGig seem to be stable again.

### DIFF
--- a/src/community/pom.xml
+++ b/src/community/pom.xml
@@ -72,7 +72,7 @@
           <descriptor>release/ext-jdbcstore.xml</descriptor>
           <descriptor>release/ext-ncwms.xml</descriptor>
           <descriptor>release/ext-gwc-sqlite.xml</descriptor> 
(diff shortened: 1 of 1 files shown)


emerkle826:
Closing this as per #2216 discussion. The GeoGig module will not compile until changes are made to get it working with SpringMVC. I'll re-open a PR to get GeoGig back into the community modules once it is aligned with the Restlet -> Spring changes

bencaradocdavies:
@emerkle826 would it not be better to include geogig on master so that it is handled en-masse with all the non-compiling community modules when https://github.com/geoserver/geoserver/pull/2216 is merged? And we do not know when https://github.com/geoserver/geoserver/pull/2216 will be merged. Until then we can include geogig to detect any *other* breakage caused by changes on master. I am tempted to reopen but will not do so without your agreement.

emerkle826:
@bencaradocdavies Very good point. I believe I over-thought this a bit, trying to be a little too proactive.

PR #2222, opened 2017-04-10
[GEOS-8025] Fixed missing slash in external legend graphic URLs if the proxy base URL does not end in a slash.

sikeoka:


tbarsballe:
Fix looks good, merged.

Have you signed the CLA to OSGeo?
While not necessary for this PR (as it is a small change to a single non-test file), you seem to be getting more active in the community, so it would be a good idea to do so.

tbarsballe:
Backported to 2.11.x and 2.10.x

PR #2224, opened 2017-04-10
Change the default X-Frame-Options header to be DENY

dvntucker:
Set the default X-Frame-Options header to be DENY, to disallow embedding GeoServer in iframes.


aaime:
Mind, deny might be too much, wicket might be depending on iframes for popups (this needs to be confirmed). I believe there was an option to allow the website itself to use iframes, but not others, or something like that? Anyways, did not research this much.

aaime:
http://apache-wicket.1842946.n4.nabble.com/FYI-X-Frame-Options-deny-Header-td4654731.html

aaime:
https://geekflare.com/secure-apache-from-clickjacking-with-x-frame-options/

PR #2236, opened 2017-04-11
Restrict restconfig dependency on hamcrest-library to test scope

bencaradocdavies:


bencaradocdavies:
gs-monitor-core relies on its transitive dependency on hamcrest-library through gs-restconfig. I will make a new PR to fix both.

bencaradocdavies:
This PR was replaced by https://github.com/geoserver/geoserver/pull/2238 .

PR #2249, opened 2017-04-12
Importer REST API Docs

tbarsballe:
Importer API docs, plus some cosmetic code changes discovered while writing the docs.

### DIFF
--- a/doc/en/api/1.0.0/importer.yaml
+++ b/doc/en/api/1.0.0/importer.yaml
@@ -22,7 +22,6 @@ paths:
                    "id": 0,
                    "state": "COMPLETE",
                    "href": "http://localhost:8080/geoserver/rest/imports/0"
-
                  }, {
                    "id": 1,
                    "state": "PENDING",
@@ -36,13 +35,15 @@ paths:
       produces:
         $ref: "#/definitions/Produces"
       parameters:
-        $ref: "#/definitions/PostParameters"
+        - $ref: "#/parameters/AsyncParameter"
+        - $ref: "#/parameters/ExecParameter"
       responses:
         201:
           description: Created
           headers:
             Location:
               description: href to the newly created import
+              type: string
     delete:
       operationId: deleteImport
       summary: Delete all imports
@@ -66,7 +67,8 @@ paths:
       produces:
         $ref: "#/definitions/Produces"
       parameters:
-        $ref: "#/definitions/PostParameters"
+        - $ref: "#/parameters/AsyncParameter"
+        - $ref: "#/parameters/ExecParameter"
       responses:
         201:
           description: Created. Returned if the import was created.
@@ -84,13 +86,15 @@ paths:
       produces:
         $ref: "#/definitions/Produces"
       parameters:
-        $ref: "#/definitions/PostParameters"
+        - $ref: "#/parameters/AsyncParameter"
+        - $ref: "#/parameters/ExecParameter"
       responses:
         201:
           description: Created
           headers:
             Location:
               description: href to the newly created import
+              type: string
     delete:
       operationId: deleteImport
       summary: Delete an import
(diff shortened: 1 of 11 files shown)


jodygarnett:
Looks great, goal was to document stealing text from rst, javadocs as required.

Feedback on description:

```
   description: The Importer extension gives a GeoServer administrator an alternate, more-streamlined method for uploading and configuring new layers. The main endpoint manages individual import jobs.
```

Suggestion:
```
description: End-point for uploading and configuring new layers. Each import results in a task, tracking the import process as data is loaded, database tables created, dataStores connected,  featureTypes and layers published, and styles generated. The importer extension is an optional install and may not be available on all deployments of GeoServer.
```

doc/en/api/1.0.0/importer.yaml
```diff
@@ -99,14 +127,585 @@ paths:
         204:
           description: No Content.
 
+parameters:
+  AsyncParameter:
+    name: async
+    description: Run the import asyncronously. Defaults to false.
```
jodygarnett: you can document a default value, no need to waste description on it.

And that way the request generator can move faster.

tbarsballe:
I just ripped that description from the existing importer API docs. Your revised version is better in some respects but inaccurate in others (among other things, and import will result in any number of tasks). I can add more detail if you want, but thats kind of what the general importer docs are for.

doc/en/api/1.0.0/importer.yaml
```diff
@@ -99,14 +127,585 @@ paths:
...
+    name: exec
+    description: Run the import when it is created. Defaults to false.
+    in: path
+    type: boolean
+    required: false
+  ExpandParameter:
+    name: expand
+    type: string
+    in: path
+    description: 'What level to expand the response object to. Can be "self" (expand only the response object and its immediate children), "all" (expand all children), "none" (don''t include any children), or a nonnegative integer, indicating the depth of children to expand to. Defaults to "self" or "none", depending on the request.'
```
jodygarnett: document as a string enum type, that way you can describe "self", "all", "none" etc...

doc/en/api/1.0.0/importer.yaml
```diff
@@ -99,14 +127,585 @@ paths:
...
+    properties:
+      id:
+        type: string
+        description: The import id
+      href:
+        type: string
+        description: URL to the import context endpoint
+      state: 
+        type: string
+        description: 'State of the import. One of: "INIT", "INIT_ERROR", "PENDING", "RUNNING", "COMPLETE"'
```
jodygarnett: use enum type

doc/en/api/1.0.0/importer.yaml
```diff
@@ -99,14 +127,585 @@ paths:
...
+      updateMode:
+        type: string
+        description: 'Update mode of the task. One of: "CREATE", "REPLACE", "APPEND", "UPDATE"'
+      data:
+        $ref: "#/definitions/Data"
+      target:
+        $ref: "#/definitions/Store"
+      progress:
+        type: string
+        description: URL to the progress endpoint for this task
```
jodygarnett: are these actually a string? or are they a link / href / url as per our usual xstream encoder?

doc/en/api/1.0.0/importer.yaml
```diff
@@ -99,14 +127,585 @@ paths:
...
+      transformChain:
+        $ref: "#/definitions/TransformChain"
+      messages:
+        $ref: "#/definitions/Messages"
+  Store:
+    title: store
+    type: object
+    description: A store
+    properties:
+      href:
```
jodygarnett: can we double check how this appears in json and xml? think we may of used a different word here.

doc/en/api/1.0.0/importer.yaml
```diff
@@ -99,14 +127,585 @@ paths:
...
+        description: '"DataFormatTransform"'
+      href: 
+        type: string
+        description: URL to the transform endpoint
+      field:
+        type: string
+        description: The field to apply the transform to
+      format: 
+        type: string
+        description: A format string used to construct the transformed date
```
jodygarnett: can we get more examples here?

src/extension/importer/rest/src/main/java/org/geoserver/importer/rest/ImportController.java
```diff
@@ -81,13 +81,13 @@ public ImportController(Importer importer) {
     }
 
     @GetMapping
-    public ImportWrapper getImports() {
+    public ImportWrapper getImports(@RequestParam(required=false) String expand) {
```
tbarsballe: Not a bug fix; A cosmetic change.
The expand parameter was implemented by some well-hidden code in ImportJSONWriter (which I discovered, and subsequently documented when making the api docs). This makes it a bit more clear that this parameter exists.

doc/en/api/1.0.0/importer.yaml
tbarsballe: This doesn't work as a string enum, as it can have an infinite number of values. The rest do though.

src/extension/importer/rest/src/main/java/org/geoserver/importer/rest/converters/ImportJSONWriter.java
```diff
@@ -90,89 +90,21 @@
...
-        }
-        // insert new objects here
-
-        if (contentType.equals(MediaType.TEXT_HTML)) {
-            writer.write("</pre></body></html>");
-        }
-        writer.flush();
-    }
 
+    //TODO: Document this behavior
```
tbarsballe: Yes, we are now. Will remove.

jodygarnett:
That is fine, stealing existing description works.

doc/en/api/1.0.0/importer.yaml
tbarsballe: None of the importer stuff goes through XStream, Importer uses its own JSON writer instead. Encoded as a string, but does follow URL syntax.

doc/en/api/1.0.0/importer.yaml
tbarsballe: Importer does not support XML output.

tbarsballe: Additionally, the importer writer manually hacks the `href` property into the store json (all these schema docs were constructed by going through the ImporterJSONWriter code.

doc/en/api/1.0.0/importer.yaml
tbarsballe: Doesn't look like it. The pattern is some custom regex-based pattern written by justin, and seems to be largely undocumented.

I found this in the old importer rest docs, which I will add, but still no good examples:
> A date parsing pattern, setup using the Java SimpleDateFormat syntax. In case it’s missing, a number of built-in formats will be tried instead (short and full ISO date formats, dates without any separators).

tbarsballe:
Updated based on feedback

PR #2251, opened 2017-04-13
[GEOS-8090] Prevent broken DXF on empty input

awaterme:
Pull request for [GEOS-8090](https://osgeo-org.atlassian.net/browse/GEOS-8090)

### DIFF
--- a/src/extension/dxf/core/src/main/java/org/geoserver/wfs/response/dxf/AbstractDXFWriter.java
+++ b/src/extension/dxf/core/src/main/java/org/geoserver/wfs/response/dxf/AbstractDXFWriter.java
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2017 Open Source Geospatial Foundation - all rights reserved
  * (c) 2001 - 2013 OpenPlans
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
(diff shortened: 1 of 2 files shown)


src/extension/dxf/core/src/main/java/org/geoserver/wfs/response/dxf/AbstractDXFWriter.java
```diff
@@ -1,4 +1,4 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
```
aaime: Our latest procedures for copyrights only demand for a header to be there, but we do not care for the dates anymore (research showed they are not really legally binding). No need to change the pull request, just remember it's not a requirement anymore for the future).

src/extension/dxf/core/src/main/java/org/geoserver/wfs/response/dxf/AbstractDXFWriter.java
```diff
@@ -195,14 +195,19 @@ protected ReferencedEnvelope getEnvelope(List featureList) {
...
+    private ReferencedEnvelope normalizeEnvelope(ReferencedEnvelope pEnv) {
+        if (pEnv != null) {
             // if it's empty, get a 1 meter envelope around it
-            if (e.getWidth() == 0 && e.getHeight() == 0) {
-                e.expandBy(1);
+            if (pEnv.getWidth() == 0 && pEnv.getHeight() == 0) {
+                // no features or no geom, enable creation of valid empty file at least
+                pEnv.init(0d, 1d, 0d, 1d);
             }
+        }else{
```
aaime: Please be mindful of coding conventions, it should be "} else {". I'm not going to demand this change, but it would be nice if you could fix it.

awaterme: Hi Andrea,
thanks for the review. I fixed the formatting.
Best regards,
Andreas

aaime:
Manually merged to squash the commits and allow easier backporting

aaime:
Backported to 2.11.x and 2.10.x too

PR #2253, opened 2017-04-13
Allow login form autocomplete to be disabled via System Property

dvntucker:
This PR disables autocomplete for the default login form, with a system property available to reenable it.

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/LoginFormHTMLInclude.java
+++ b/src/web/core/src/main/java/org/geoserver/web/LoginFormHTMLInclude.java
@@ -4,9 +4,9 @@
  */
 package org.geoserver.web;
 
-import java.io.BufferedReader;
-import java.io.InputStream;
-import java.io.InputStreamReader;
+import java.io.StringWriter;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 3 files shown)


aaime:
Looking at the pull request, it seems to be doing the opposite, allowing to turn off autocomplete via a configuration while keeping it on by default (which seems to be an ok setup to me).
Would have been nicer to have a UI tunable, but a sysvar is good too.

src/web/core/src/main/java/org/geoserver/web/LoginFormHTMLInclude.java
```diff
@@ -48,17 +56,23 @@ public void onComponentTagBody(final MarkupStream markupStream, final ComponentT
...
         try {
-            InputStream inputStream = this.resourceReference.getResource().getResourceStream()
-                    .getInputStream();
-            BufferedReader br = new BufferedReader(new InputStreamReader(inputStream));
-            StringBuilder sb = new StringBuilder();
-            String line;
-            while ((line = br.readLine()) != null) {
-                sb.append(line + "\n");
+
+            Configuration cfg = new Configuration( );
```
aaime: Creating the configuration and loading the templates is expensive, you should consider following up this patter (without copying the KML comment maybe :-p ):
https://github.com/geoserver/geoserver/blob/master/src/wms/src/main/java/org/geoserver/wms/featureinfo/HTMLFeatureInfoOutputFormat.java#L51

dvntucker:
Ah yes, I screwed up the wording in the PR. It's enabled by default be able to be turned off. I will look at making the configuration static.

dvntucker:
Made the template configuration static to enable caching.

mprins:
To actually turn off autocomplete for username and password fields the password should be set to 
"new-password" and not to "off" (because that will generally be ignored by browsers)

see: https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

PR #2254, opened 2017-04-14
Property to control global layer group visibility in virtual services, GEOS-8091

smithkm:
https://osgeo-org.atlassian.net/browse/GEOS-8091
Revised version of abandoned PR #1362

Still uses `GEOSERVER_GLOBAL_LAYER_GROUP_INHERIT` as I can't think of a better term for this than inherit even though it's admittedly not entirely intuitive.

### DIFF
--- a/src/platform/src/main/java/org/geoserver/platform/util/GeoServerPropertyFactoryBean.java
+++ b/src/platform/src/main/java/org/geoserver/platform/util/GeoServerPropertyFactoryBean.java
@@ -1,0 +1,97 @@
+/* (c) 2014 - 2015 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2013 OpenPlans
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+
+package org.geoserver.platform.util;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import org.geoserver.platform.GeoServerExtensions;
+import org.geotools.util.logging.Logging;
+import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.config.AbstractFactoryBean;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+
+/**
+ * Spring FactoryBean that can create a bean based on the value of a system property, context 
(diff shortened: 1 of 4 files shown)


src/platform/src/main/java/org/geoserver/platform/util/GeoServerPropertyFactoryBean.java
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Spring FactoryBean that can create a bean based on the value of a system property, context 
+ * parameter, or environment variable
+ * 
+ * @author Kevin Smith, Boundless
+ *
+ * @param <T>
+ */
+public abstract class GeoServerPropertyFactoryBean<T> extends AbstractFactoryBean<T> implements ApplicationContextAware {
```
tbarsballe: Why is this necessary? It doesn't seem to be used anywhere?

aaime:
I think @nmco wanted to comment on this one, something about global layer groups not being visible anymore on master?

smithkm:
That was a utility I created and ended up not needing, but it still seemed generally useful and worth having so I left it in.  I can move it to another PR.

smithkm:
@aaime I tested master before resurrecting this and I was still seeing the global groups in the virtual service capabilities.

tbarsballe:
I've got no strong feelings either way.
It certainly looks like it could be a useful feature, but I am a bit concerned that it will just be committed and then, since there are no examples of it in-use, it will be forgotten and never actually used.

smithkm:
@tbarsballe Yes that worried me, but it is even less likely to get used if it's not committed.

tbarsballe:
Indeed. I am overall happy with the PR, so feel free to merge.

nmco:
Hummm that's strange just give it a try and global layers groups are not showing up for virtual services capabilities and there is a test for this behavior:
https://github.com/geoserver/geoserver/blob/master/src/wms/src/test/java/org/geoserver/wms/wms_1_3/CapabilitiesIntegrationTest.java#L779-L799

Maybe I misunderstood something.

nmco:
Ah wait my bad :man_facepalming: I had in mind that you wanted to removed global layers groups from **capabilities documents**, so forget about what I said.

PR #2256, opened 2017-04-14
GEOS-8083 jdbcconfig: Don't use jdbcstore for logging.xml

NielsCharlier:


jodygarnett:
While this looks fine I am I not in position to test this week.

jodygarnett:
Assume you need to back port.

NielsCharlier:
Thanks for approval Jody, will backport indeed

PR #2257, opened 2017-04-14
Monitoring: remove obsolete caching of config files

NielsCharlier:
Trivial change.


src/extension/monitor/core/src/main/java/org/geoserver/monitor/auditlog/AuditLogger.java
```diff
@@ -76,8 +73,6 @@
 
     public AuditLogger(MonitorConfig config, GeoServerResourceLoader loader) throws IOException {
         this.config = config;
-        Resource monitoring = loader.get("monitoring");        
-        defaultPath = monitoring.dir().getAbsolutePath();
```
jodygarnett: So no need to checkout dir() contents here - we are just going after the single file below.

NielsCharlier:
Thanks Jody, I will backport indeed

PR #2259, opened 2017-04-14
Fix for Issue 8093: Align Backup Restore plugin REST interface with new GeoServer MVC pattern

afabiani:


aaime:
Pull request builds, reviews are not really needed for community modules, so I'd say you can go ahead and merge.

PR #2261, opened 2017-04-16
Fix a wrong use of an indexer in a loop

lifove:
This might be a trivial change. (I'm sorry that I could not issue a ticket first.) The indexer of the outer loop was used in the inner loop. I have no idea if this is intended but wanted to issue this pull request just in case. Thanks!


aaime:
The changes seems to be good, thanks for sharing.
Our contribution procedures demand a junit test to go with it, any chance you can setup one?
See https://github.com/geoserver/geoserver/blob/master/CONTRIBUTING.md

jodygarnett:
@tbarsballe can you combine with #2289

aaime:
If no test shows up, I'll eventually have to close this pull request instead of merging it. 
That would be a real shame, any chance you can add the test?

bencaradocdavies:
This request is now obsolete. The target code has substantially changed. We reviewed it and the current code seems to be correct.

PR #2264, opened 2017-04-17
GEOS-607 SLD files not deleted with Style (backport)

NielsCharlier:


NielsCharlier:
Assume this is ok to backport?

PR #2266, opened 2017-04-18
[geos-7946], prefer @Rule to @Before end @After as done elsewhere.

cmangeat:
@Before and @After were there because I can't manage to make work @Rule with power mockito runner. As no need to power mockito anymore...

Refactor.


aaime:
Thanks!

PR #2225, opened 2017-04-10
Change the default X-Frame-Options header to be SAMEORIGIN

dvntucker:
Set the default X-Frame-Options header to be SAMEORIGIN

PR #2267, opened 2017-04-18
Allow X-Frame-Options header to be set (2.10.x)

dvntucker:
A backport of 

https://github.com/geoserver/geoserver/pull/2225

with the minor caveat that instead of defaulting to SAMEORIGIN, I have turned the header OFF by default, to retain back-compat and not break existing builds.


tbarsballe:
Looks good, seems to be a straightforward backport. No obvious issues.

bencaradocdavies:
No Jira issue but this security fix will receive special mention in the release announcement.

PR #2277, opened 2017-04-20
WMS interpolation by layer

ridethepenguin:
This PR implements what is described in [GEOS-8100](https://osgeo-org.atlassian.net/browse/GEOS-8100).

Tests for GUI, WMS and also REST are included. Docs are updated in a separate commit (I can squash it if deemed more appropriate).


mbarto:
Looks good to me.
+1

src/main/src/main/java/org/geoserver/catalog/LayerInfo.java
```diff
@@ -168,4 +168,28 @@
...
     void setAdvertised(boolean advertised);
 
+    /**
+     * The default interpolation method.
+     *
+     * <p>
+     * If not specifed (i.e. {@code null}), the service default will be used.
+     * </p>
+     */
+    String getDefaultInterpolationMethod();
```
aaime: This method applies to WMS only, it's not a default interpolation in WCS for example, better be explicit about it.
Also, since it can use only three values, best to use an enumeration, like WMSInfo.WMSInterpolation.
Hopefully this will also make it throw an exception is someone sends an invalid value over REST

src/wms/src/test/java/org/geoserver/wms/map/RenderedImageMapOutputFormatTest.java
```diff
@@ -350,6 +350,154 @@ public void testInterpolations() throws IOException, IllegalFilterException, Exc
...
+
+        layerInfo.setDefaultInterpolationMethod("Nearest");
+        assertEquals("Nearest",
+                request.getLayers().get(0).getLayerInfo().getDefaultInterpolationMethod());
+        assertTrue(request.getInterpolations().isEmpty());
+
+        WMSMapContent map = createWMSMap(env);
+        map.setRequest(request);
+
+        RenderedImageMap imageMap = this.rasterMapProducer.produceMap(map);
```
aaime: One non binding observation (no need for change, it's a nice to have). You could extend WebMap a bit to allow access to the WebMapContent, and then use it to check if the right interpolation has been set in the layer, seems a bit better than just chechking that "something changed" (which is fair by the way, one cannot expect  particular value in the output pixels as JAI could be pure java, native, jai-ext, or OSX one, and they would return slight different values)

src/main/src/main/java/org/geoserver/catalog/LayerInfo.java
aaime: Btw, noticed WMSInfo.WMSInterpolation is available only in gs-wms, a similar enum could be created inside LayerInfo (and/or be stand alone and used in both cases, but it would get trickier to ensure backwards compatibility)

ridethepenguin: That's why I opted for a String property, I didn't want to either 1) create two identical enums or 2) run the risk of breaking backwards compatibility by refactoring WMSInfo into using a standalone WMSInterpolation.

Should I move on with 1), 2), or leave things as they are?

aaime: I'd say, go for 1) at the moment.

src/wms/src/test/java/org/geoserver/wms/map/RenderedImageMapOutputFormatTest.java
ridethepenguin: You mean something along these lines?
```
Layer mLayer = imageMap.getMapContext().layers().get(0);
Interpolation interpolationSet = (Interpolation)mLayer.getUserData().get(StreamingRenderer.BYLAYER_INTERPOLATION);
assertNotNull(interpolationSet);
assertEquals(Interpolation.getInstance(Interpolation.INTERP_NEAREST), interpolationSet);
```

I would have to disable the direct raster render path (active by default), otherwise nothing is set on the layer... is my understanding correct?

ridethepenguin:
@aaime I refactored `LayerInfo` and tried to improve `RenderedImageMapOutputFormatTest` as you suggested... is this good to go now?

src/wms/src/test/java/org/geoserver/wms/map/RenderedImageMapOutputFormatTest.java
aaime: Ah, you have a point there, right. Never mind!

src/restconfig/src/test/java/org/geoserver/rest/catalog/LayerControllerTest.java
```diff
@@ -268,4 +269,36 @@ public void testPutWorkspaceAlternateStyle() throws Exception {
...
+        MockHttpServletResponse response =
+            putAsServletResponse(ROOT_PATH + "/layers/cite:Buildings", xml, "application/xml");
+        assertEquals(200, response.getStatus());
+
+        l = cat.getLayerByName("cite:Buildings");
+        assertNotNull(l.getDefaultWMSInterpolationMethod());
+        assertEquals(LayerInfo.WMSInterpolation.Nearest, l.getDefaultWMSInterpolationMethod());
+
+        dom = getAsDOM(ROOT_PATH + "/layers/cite:Buildings.xml",200);
+        print(dom);
```
aaime: Leftover print

aaime:
Ciao Stefano, yep, good to go. I saw a leftover print to standard output in the test, but no big deal really, fix as you see fit (and I believe you can merge yourself right?)

src/restconfig/src/test/java/org/geoserver/rest/catalog/LayerControllerTest.java
ridethepenguin: Ha! As I was writing that line, I thought to myself: "You are going to forget to remove this"; and of course, that's exactly what happened :wink:

ridethepenguin:
Leftover print removed, I'll merge this as soon as Travis is done.

PR #2278, opened 2017-04-21
[GEOS-8105] WCS 2.0 GetCoverage ignores service level overview policy and disallow reader subsampling

aaime:


src/wcs2_0/src/main/java/org/geoserver/wcs2_0/GetCoverage.java
```diff
@@ -932,37 +932,62 @@ private GridCoverage2D readCoverage(CoverageInfo cinfo, GridCoverageRequest requ
         // === read
         // check limits
         WCSUtils.checkInputLimits(wcs,cinfo,reader,readGG);
+        Hints readHints = new Hints();
+        readHints.putAll(hints);
+        if(request.getOverviewPolicy() != null) {
+            hints.add(new Hints(Hints.OVERVIEW_POLICY, request.getOverviewPolicy()));
```
aaime: Yep, good catch, this needs fixing indeed.

aaime:
Restarting build, it was delayed by the Boundless repo availability issues to the point that Travis killed it.

PR #2279, opened 2017-04-21
[GEOS-8101] Remove FeatureTypeName from raster_template.sld

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-8101

Best place I could fit the test case is gs-wms, as that is the only straightforward way I can find to actually render a style.


tbarsballe:
Backported  to 2.11.x and 2.10.x

PR #2282, opened 2017-04-25
Indirection avoiding trouble with turbojpeg and tiff.

cmangeat:
Raster rendering bug, trouble with jai native, jai ext, jpeg turbo and alpha channel removed via jai BandSelect operation.


src/extension/libjpeg-turbo/pom.xml
```diff
@@ -61,4 +61,47 @@
...
+                -Dorg.geotools.image.test.enabled=${image.tests}
+                -Dorg.geotools.image.test.interactive=${interactive.image} -Duser.timezone=${user.timezone}
+                -Dwindows.leniency=${windows.leniency} -XX:+TieredCompilation -XX:TieredStopAtLevel=1
+              </argLine>
+              <enableAssertions>true</enableAssertions>
+              <printSummary>true</printSummary>
+              <testFailureIgnore>${allow.test.failure.ignore}</testFailureIgnore>
+              <excludedGroups>${test.excludedGroups}</excludedGroups>
+              <additionalClasspathElements>
+                <additionalClasspathElement>/usr/share/java/mlibwrapper_jai.jar</additionalClasspathElement>
```
aaime: This assume a specific *nix filesystem, what happens if one tries to run it on windows?

src/extension/libjpeg-turbo/src/main/java/org/geoserver/map/turbojpeg/TurboJpegImageWorker.java
```diff
@@ -103,7 +104,7 @@ public final void writeTurboJPEG(final OutputStream destination, final float com
         final int numBands = image.getSampleModel().getNumBands();
         if (hasAlpha) {
             final int requestedBands = numBands - 1;
-            if (ImageUtilities.isMediaLibAvailable()) {
+            if (ImageUtilities.isMediaLibAvailable() && !JAIExt.isJAIExtOperation("BandSelect")) {
```
aaime: @dromagnoli what was the reason for the medialib check?

cmangeat:
Goog morning,

Thank you for the time spent. I plan to have a scrupulous look at the end
of this a week (rebase, think about file layout, modify the test, maybe
open a ticket where "client value" could be discussed).

Best regards.

cmangeat:
replaced by "libJegTurbo, jai native (mlib) and jai-ext, seems there is a trouble when removing alpha with… #2309"

PR #2283, opened 2017-04-25
Codemirror css complete

aaime:


src/web/core/src/main/java/org/geoserver/web/wicket/CodeMirrorEditor.js
```diff
@@ -3,7 +3,8 @@ var editor = CodeMirror.fromTextArea(textarea, {
     mode: '$mode',
     theme: 'default',
     lineWrapping: true,
-    lineNumbers: true
+    lineNumbers: true,
+    extraKeys: {"Ctrl-Space": "autocomplete"}
```
tbarsballe: Hardcoding Ctrl-Space here brings up a couple issues:
* For optimal compatibility, we should be using Cmd instead of Ctrl on OS X. A little bit of testing shows that Codemirror does indeed use Cmd-Z for undo (etc...) on OS X, so this is supported at some level.
* Cmd-Space is already an OS-level shortcut (Spotlight Search) on OS X, so an alternative shortcut key may be advisable (personally, I have used Ctrl/Cmd-Enter in the past). However, given this issue, I could see just sticking with Ctrl-Space for both OSes

aaime: I am wondering if codemirror might be playing some magic already... like, all the codemirror site demos hard-code CTRL-Space, like this one:
https://codemirror.net/demo/html5complete.html
Do they not work on OSX?

tbarsballe: Looks like Ctrl-Space is still used (an works fine) on OS X in those examples, so I guess that is fine.

aaime:
The Travis build keeps on hitting the 50 minutes mark...

PR #2285, opened 2017-04-26
[GEOS-8112] Fix GeoPackageGetMapOutputFormat tile order

tbarsballe:
GeoPackage tile order along the y-axis is inverted relative to what it should be; the 0,0 tile corresponds to the lower-left corner. This PR fixes this issue for GeoServer WMS GeoPackage output.

According to the [OGC Spec](http://www.geopackage.org/spec/#tile_matrix):

> The tile coordinate (0,0) always refers to the tile in the upper left corner of the tile matrix at any zoom level, regardless of the actual availability of that tile

For more details:
https://osgeo-org.atlassian.net/browse/GEOT-5715
https://osgeo-org.atlassian.net/browse/GEOS-8112
https://trac.osgeo.org/gdal/ticket/6871

This PR is in conjunction with https://github.com/geotools/geotools/pull/1567
While the two fixes are technically independent, they should be merged in relatively short order of each other.


tbarsballe:
A couple of additional changes, technically unrelated to GEOS-8112, were required to get GeoPackage tests working:

* WMSTestSupport now uses the WORLD QName from MockData
* GeoPackageGetFeatureOutputFormatTest no longer uses the SEVEN or FIFTEEN datasets, as they have multiple geometry columns, which is incompatible with the GeoPackage specification.

tbarsballe:
Backported to 2.10.x and 2.11.x

PR #2286, opened 2017-04-28
WIP: gwc-bugfix

spike83:
enabled check for bbox on tile requests from wmts and gmaps

### DIFF
--- a/src/gwc/src/main/java/org/geoserver/gwc/GWC.java
+++ b/src/gwc/src/main/java/org/geoserver/gwc/GWC.java
@@ -82,6 +82,7 @@ import org.geotools.ows.ServiceException;
 import org.geotools.referencing.CRS;
 import org.geotools.referencing.CRS.AxisOrder;
 import org.geotools.util.logging.Logging;
+import org.geowebcache.GeoWebCacheDispatcher;
 import org.geowebcache.GeoWebCacheEnvironment;
 import org.geowebcache.GeoWebCacheException;
 import org.geowebcache.GeoWebCacheExtensions;
@@ -2186,6 +2187,34 @@ public class GWC implements DisposableBean, InitializingBean, ApplicationContext
                 }
             }
         }
+    }
+    
+    /**
+     * Verify that a layer is accessible within a certain bounding box (calculated from tile) using the (secured) catalog
+     * 
+     * @param layerName name of the layer
+     * @param srs reference system
+     * @param level zoom level
+     * @param tileColumn column of tile in tile grid
+     * @param tileRow row  of tile in tile grid
+     * @throws ServiceException
+     */
+    public void verifyAccessTiledLayer(String layerName, String srs, int level, long tileColumn,
+            long tileRow) throws ServiceException {
+        // get bounding box of requested tile
+        GeoServerTileLayer layer = (GeoServerTileLayer) getTileLayerByName(layerName);
+        GridSubset gridSubset = layer.getGridSubset(srs);
+        if (gridSubset == null) {
+            throw new ServiceException(
+                    "The specified grid " + srs + " is not defined on layer " + layerName,
+                    "AccessDenied");
+        }
+        long[] tileIndex = { tileColumn, tileRow, level };
+        BoundingBox bounds = gridSubset.boundsFromIndex(tileIndex);
+        double[] coords = bounds.getCoords();
+        ReferencedEnvelope envelope = new ReferencedEnvelope(coords[0], coords[1], coords[2],
+                coords[3], this.getDeclaredCrs(layerName));
+        this.verifyAccessLayer(layerName, envelope);
     }
 
     public CoordinateReferenceSystem getDeclaredCrs(final String geoServerTileLayerName) {
--- a/src/gwc/src/main/java/org/geoserver/gwc/dispatch/GwcServiceProxy.java
+++ b/src/gwc/src/main/java/org/geoserver/gwc/dispatch/GwcServiceProxy.java
@@ -19,6 +19,7 @@ import org.geoserver.config.ServiceInfo;
 import org.geoserver.config.impl.ServiceInfoImpl;
 import org.geoserver.gwc.GWC;
 import org.geoserver.gwc.config.GWCServiceEnablementInterceptor;
+import org.geoserver.gwc.layer.GeoServerTileLayer;
 import org.geoserver.ows.DisabledServiceCheck;
 import org.geoserver.ows.Dispatcher;
 import org.geoserver.ows.Response;
@@ -30,6 +31,8 @@ import org.geotools.referencing.CRS;
 import org.geotools.util.Version;
 import org.geowebcache.GeoWebCacheDispatcher;
 import org.geowebcache.GeoWebCacheExtensions;
+import org.geowebcache.grid.GridSubset;
+import org.geowebcache.service.gmaps.GMapsConverter;
 
 import com.google.common.collect.ImmutableList;
 
@@ -148,11 +151,16 @@ public class GwcServiceProxy {
             
         } else if (rawRequest.getPathInfo().toLowerCase().startsWith("/service/wmts")) {
             String layer = (String) parameters.get("LAYER");
-                        
+
             if (layer != null) {
-                GWC.get().verifyAccessLayer(layer, null);
+                String tileMatrix = (String) parameters.get("TileMatrix");
+                String srs = tileMatrix.substring(0, tileMatrix.lastIndexOf(":"));
+                int level = Integer.parseInt(tileMatrix.substring(tileMatrix.lastIndexOf(":") + 1));
+                long col = Long.parseLong((String) parameters.get("TileCol"));
+                long row = Long.parseLong((String) parameters.get("TileRow"));
+                GWC.get().verifyAccessTiledLayer(layer, srs, level, col, row);
             }
-            
+
         } else if (rawRequest.getPathInfo().toLowerCase().startsWith("/service/tms/1.0.0/")) {
             String layer = rawRequest.getPathInfo().toLowerCase().substring("/service/tms/1.0.0/".length());
             if (layer.indexOf('/') >= 0) {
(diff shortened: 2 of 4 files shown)


src/gwc/src/main/java/org/geoserver/gwc/dispatch/GwcServiceProxy.java
```diff
@@ -150,7 +153,12 @@ public void verifyAccess(HttpServletRequest rawRequest) throws org.geotools.ows.
             String layer = (String) parameters.get("LAYER");
                         
             if (layer != null) {
-                GWC.get().verifyAccessLayer(layer, null);
+            	String tileMatrix = (String) parameters.get("TileMatrix");
+            	String srs = tileMatrix.substring(0, tileMatrix.lastIndexOf(":"));
+            	int level = Integer.parseInt(tileMatrix.substring(tileMatrix.lastIndexOf(":")+ 1));
+            	long col = Long.parseLong((String) parameters.get("TileCol"));
+            	long row = Long.parseLong((String) parameters.get("TileRow"));
+                GWC.get().verifyAccessTiledLayer(layer,srs, level, col, row);
```
jodygarnett: This look correct to me

src/gwc/src/main/java/org/geoserver/gwc/dispatch/GwcServiceProxy.java
```diff
@@ -173,11 +181,28 @@ public void verifyAccess(HttpServletRequest rawRequest) throws org.geotools.ows.
...
+            	int level = Integer.parseInt((String)parameters.get("zoom"));
+            	long col = Long.parseLong((String) parameters.get("x"));
+            	long row = Long.parseLong((String) parameters.get("y"));
+        		try {
+					long [] converted = GMapsConverter.convert(level, col, row);
+					col = converted[0];
+					row = converted[1];
+				} catch (org.geowebcache.service.ServiceException e) {
+					throw new ServiceException(e);
+				}
```
jodygarnett: the logic looks correct, please use our code formatter (spaces not tabs).

src/gwc/src/main/java/org/geoserver/gwc/dispatch/GwcServiceProxy.java
```diff
@@ -173,11 +181,28 @@ public void verifyAccess(HttpServletRequest rawRequest) throws org.geotools.ows.
...
                 for (String layerName: layers) {
                     layerName = layerName.trim();
-                    GWC.get().verifyAccessLayer(layerName, null);
+                	GeoServerTileLayer layer = (GeoServerTileLayer) GWC.get().getTileLayerByName(layerName);
+                	String srs = "EPSG:3857";
+                	GridSubset gridSubset =  layer.getGridSubset(srs);
+                	if(gridSubset == null){
+                		srs = "EPSG:900913";
+                	}
+                    GWC.get().verifyAccessTiledLayer(layerName ,srs, level, col, row);
```
jodygarnett: This looks correct, are these the only two places we need to verify access?

aaime:
Would also be great if the commit message was in the form of "[TICKET-ID] Title title":

- It ensures that a ticket has been created (so that we have release notes, they are generated out of Jira)
- Points to eventual source of more info

spike83:
Fixed format and commit title

src/gwc/src/main/java/org/geoserver/gwc/dispatch/GwcServiceProxy.java
spike83: @jodygarnett I know about 2 other places for the kml and tms requests direct above which I'll try to fix. I'll have a look for more places to fix. I think a check on the call sources of the ```verifyAccessLayer``` should point that out?

src/gwc/src/main/java/org/geoserver/gwc/dispatch/GwcServiceProxy.java
```diff
@@ -148,11 +151,16 @@ public void verifyAccess(HttpServletRequest rawRequest) throws org.geotools.ows.
...
             String layer = (String) parameters.get("LAYER");
-                        
+
             if (layer != null) {
-                GWC.get().verifyAccessLayer(layer, null);
+                String tileMatrix = (String) parameters.get("TileMatrix");
+                String srs = tileMatrix.substring(0, tileMatrix.lastIndexOf(":"));
+                int level = Integer.parseInt(tileMatrix.substring(tileMatrix.lastIndexOf(":") + 1));
+                long col = Long.parseLong((String) parameters.get("TileCol"));
+                long row = Long.parseLong((String) parameters.get("TileRow"));
```
smithkm: WMTS rows and GWC's internal rows run in opposite directions.  See: https://github.com/GeoWebCache/geowebcache/blob/master/geowebcache/wmts/src/main/java/org/geowebcache/service/wmts/WMTSService.java#L253

src/gwc/src/main/java/org/geoserver/gwc/dispatch/GwcServiceProxy.java
```diff
@@ -148,11 +151,16 @@ public void verifyAccess(HttpServletRequest rawRequest) throws org.geotools.ows.
             
         } else if (rawRequest.getPathInfo().toLowerCase().startsWith("/service/wmts")) {
             String layer = (String) parameters.get("LAYER");
-                        
+
             if (layer != null) {
-                GWC.get().verifyAccessLayer(layer, null);
+                String tileMatrix = (String) parameters.get("TileMatrix");
+                String srs = tileMatrix.substring(0, tileMatrix.lastIndexOf(":"));
+                int level = Integer.parseInt(tileMatrix.substring(tileMatrix.lastIndexOf(":") + 1));
```
smithkm: Rather than parsing out the level yourself, maybe try using the same logic as the WMTS implementation here: https://github.com/GeoWebCache/geowebcache/blob/master/geowebcache/wmts/src/main/java/org/geowebcache/service/wmts/WMTSService.java#L238

src/gwc/src/main/java/org/geoserver/gwc/dispatch/GwcServiceProxy.java
```diff
@@ -173,13 +181,31 @@ public void verifyAccess(HttpServletRequest rawRequest) throws org.geotools.ows.
...
                         
-            if (layerstr != null) {                
+            if (layerstr != null) {
+                int level = Integer.parseInt((String) parameters.get("zoom"));
+                long col = Long.parseLong((String) parameters.get("x"));
+                long row = Long.parseLong((String) parameters.get("y"));
+                try {
+                    long[] converted = GMapsConverter.convert(level, col, row);
+                    col = converted[0];
+                    row = converted[1];
```
smithkm: Admittedly the converter doesn't do anything to the level here, but for consistency it would be best to set the zoom level based on the returned value along with `row` and `col`.  In particular that would make it clearer when adapting the code for the mgmaps protocol

I would probably have used new variables to keep the GMaps and GWC coordinates distinct for readability.

src/gwc/src/main/java/org/geoserver/gwc/GWC.java
```diff
@@ -2187,6 +2188,34 @@ public void verifyAccessLayer(String layerName, ReferencedEnvelope boundingBox)
...
+     * @param level zoom level
+     * @param tileColumn column of tile in tile grid
+     * @param tileRow row  of tile in tile grid
+     * @throws ServiceException
+     */
+    public void verifyAccessTiledLayer(String layerName, String srs, int level, long tileColumn,
+            long tileRow) throws ServiceException {
+        // get bounding box of requested tile
+        GeoServerTileLayer layer = (GeoServerTileLayer) getTileLayerByName(layerName);
+        GridSubset gridSubset = layer.getGridSubset(srs);
```
smithkm: Gridsets, and coordinate systems are different things, do not confuse them.  I think you should be calling this parameter "gridset" or "gridset name"  A CRS can have any number of gridsets using it.  GeoWebCache, for legacy reasons, has several gridsets that are simply named for the CRSes they use, but this is not necessarily the case.

src/gwc/src/test/java/org/geoserver/gwc/GWCDataSecurityTest.java
```diff
@@ -149,6 +163,76 @@ public void testNoMosaic() throws Exception {
...
+        String path = "gwc/service/wmts?LAYER=sf:mosaic&FORMAT=image/png&SERVICE=WMTS&VERSION=1.0.0" +
+        "&REQUEST=GetTile&TILEMATRIXSET=EPSG:4326&TILEMATRIX=EPSG:4326:0&TILECOL=0&TILEROW=0";
+        MockHttpServletResponse response = getAsServletResponse(path);
+        assertEquals("image/png", response.getContentType() );
+        
+        // try again, now should be cached
+        response = getAsServletResponse(path);
+        assertEquals("image/png", response.getContentType());
+
+        // permission must be denied to cite user
```
smithkm: Shouldn't this check that permission is denied after getting the tile cached?  You've cached it for sf:mosaic, but not for sf:mosaic2 .  If nothing's cached already we should be safe as another deeper error will occur.  Your test will still notice the difference between the errors, but I'd feel better if the dangerous case were the one that's tested.

smithkm:
I extended your patch to apply to TMS  including a bit of refactoring upstream in GWC in order to share some parsing logic.  https://github.com/smithkm/geoserver/tree/gwc-optimization

jodygarnett:
There are a couple of different ways of working on this together:
- @spike83, when making a pull request you can invite collaborators such as @smithkm.
- Kevin may also be able to issue a pull request against https://github.com/sbnet-ch/geoserver.git gwc-optimization

spike83:
@smithkm  I've created a pull request on your changes and merged.
FYI: Allow edits from maintainers is set on

spike83:
@smithkm I used your function for all tests and implemented the kml part. In the kml part I've found two other problems and solved them. One was that the toLowerCase on the whole path was changing the layer name when it had upper case letters in it. Second was that the layer was handled correctly for `/service/kml/<namespace>:<layername>.png.kml` but `/service/kml/<namespace>:<layername>/x123y123z16.png.kml` was converted wrong to the layer name `<namespace>:<layername>/x123y123z16` which resulted in 
`org.geotools.ows.ServiceException: Could not find layer <namespace>:<layername>/x123y123z16`.

It would be great if you could take another look on it.

smithkm:
OK, I've also been working on it, including some tests to see if the resulting coordinates are correct.  Not all protocols have the same idea of which direction is up, not all implementations of bounding boxes use the same order for their parameters, and the advertised CRS of the back end layer is a very different thing from the CRS of the gridset of the request.  I'm about to go to sleep (it's way late here)  if you have time to look into merging our work together what I have done is here https://github.com/smithkm/geoserver/commits/gwc-optimization

Otherwise I'll take a look in the morning.

smithkm:
Superseded by #2329

PR #2287, opened 2017-04-28
[GEOS-8110] Fix IndexOutOfBounds in WMSWorkspaceQualifier

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-8110

### DIFF
--- a/src/wms/src/test/java/org/geoserver/wms/wms_1_3/GetMapIntegrationTest.java
+++ b/src/wms/src/test/java/org/geoserver/wms/wms_1_3/GetMapIntegrationTest.java
@@ -373,7 +373,16 @@ public class GetMapIntegrationTest extends WMSTestSupport {
         } finally {
             catalog.remove(group);
         }
-    }   
+    }
+
+    @Test
+    public void testWorkspaceQualifiedNoLayersError() throws Exception {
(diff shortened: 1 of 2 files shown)


dvntucker:
Looks fine, but can we use something like this instead?

http://docs.oracle.com/javase/8/docs/api/java/lang/String.html#join-java.lang.CharSequence-java.lang.Iterable-

tbarsballe:
Yes, that's a much better way of doing this. Updated.

src/wms/src/test/java/org/geoserver/wms/wms_1_3/GetMapIntegrationTest.java
```diff
@@ -373,7 +373,17 @@ public void testLayerGroupModeEo() throws Exception {
...
         }
-    }   
+    }
+
+    @Test
+    public void testWorkspaceQualifiedNoLayersError() throws Exception {
+        Document dom = getAsDOM("sf/wms?bbox=" + bbox + "&styles="
+                + "&layers=" + "&format=image/png" + "&request=GetMap" + "&width=550"
+                + "&height=250" + "&srs=EPSG:4326", Charset.defaultCharset().displayName());
+        print(dom);
```
aaime: Leftover print

tbarsballe:
Backported to 2.11.x and 2.10.x

PR #2288, opened 2017-04-28
User manual section on REST API examples now based on endpoint

bmmpxf:
Previously, section was organized by tool (curl, PHP, etc.)


jodygarnett:
This looks great @bmmpxf - I like the reduction in duplication.

PR #2289, opened 2017-04-28
[GEOS-8117] Improve support for null service/version/request in KvpUtils

tbarsballe:
Fixes support for null service / version / request in KvpUtils.purgeParsers and KvpUtils.findParser (according to the javadoc, this was already supported, but it appears the implementation was bugged).

KvpUtils.purgeParsers will now no longer purge parsers based on service / version / request if the applicable parameter is null in the original Kvp map.

KvpUtils.findParser now properly handles cases where both the kvp map and kvp parser have a null service / version / parser.

The beans "wfsFormatOptionsKvpParser" and "wmsFormatOptionsKvpParser" are functionally identical and have been renamed to "formatOptionsKvpParser" - in the case that both WFS and WMS are enabled, one will override the other. This prevents us throwing an IllegalStateException due to multiple identical KvpParsers (this didn't happen before because null service / version / request was not implemented properly).

See https://osgeo-org.atlassian.net/browse/GEOS-8117 for more details

### DIFF
--- a/src/ows/src/main/java/org/geoserver/ows/util/KvpUtils.java
+++ b/src/ows/src/main/java/org/geoserver/ows/util/KvpUtils.java
@@ -453,12 +453,12 @@ public class KvpUtils {
         for (Iterator<KvpParser> p = parsers.iterator(); p.hasNext();) {
             KvpParser parser = p.next();
 
-            if (parser.getService() != null && !parser.getService().equalsIgnoreCase(service)) {
+            if (service != null && parser.getService() != null && !parser.getService().equalsIgnoreCase(service)) {
                 p.remove();
-            } else if (parser.getVersion() != null
+            } else if (version != null && parser.getVersion() != null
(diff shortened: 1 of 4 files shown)


tbarsballe:
Travis CI failure occurring in INSPIRE - I seem to be seeing the same failure on master locally.

tbarsballe:
CI failure fixed - added support to `findParser` for cases where `service` is `null` for both `candidate` and `parser`.

tbarsballe:
Improved test case - Now tests without service, without version, and without service & version.

jodygarnett:
Looks good, assume this is a fix you want to backport.

tbarsballe:
Yes, this is something I want to backport.

bencaradocdavies:
Please update [GEOS-8117](https://osgeo-org.atlassian.net/browse/GEOS-8117).

This change might be responsible for the CITE WFS 1.1 failures. Please see the mailing list for discussion.

PR #2291, opened 2017-04-28
[GEOS-7757] Fix empty row during legend creation.

goudine:
This commit stops the generation of an empty row on a vertical non-raster
legend when there is a text symbolizer used in the style SLD.

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/legendgraphic/LegendMerger.java
+++ b/src/wms/src/main/java/org/geoserver/wms/legendgraphic/LegendMerger.java
@@ -244,7 +244,7 @@ public class LegendMerger {
         }
 
         if (layout == LegendLayout.VERTICAL) {
-            Column[] columns = createColumns(nodes, mergeOptions.getColumnHeight(), mergeOptions.getColumns());
+            Column[] columns = createColumns(nodes, mergeOptions.getColumnHeight(), mergeOptions.getColumns(), null, false);
             finalLegend = buildFinalVLegend(columns, mergeOptions);
         }
 
@@ -263,7 +263,8 @@ public class LegendMerger {
      * @return the image with all the images on the argument list.
      * 
      */
-    public static BufferedImage mergeLegends(Rule[] rules, GetLegendGraphicRequest req, MergeOptions mergeOptions) {
+    public static BufferedImage mergeLegends(Rule[] rules, GetLegendGraphicRequest req,
+            MergeOptions mergeOptions) {
         List<RenderedImage> imageStack= mergeOptions.getImageStack();
         
         // Builds legend nodes (graphics + label)
@@ -297,7 +298,7 @@ public class LegendMerger {
         }
 
         if (layout == LegendLayout.VERTICAL) {
-            Column[] columns = createColumns(nodes, mergeOptions.getColumnHeight(), mergeOptions.getColumns());
+            Column[] columns = createColumns(nodes, mergeOptions.getColumnHeight(), mergeOptions.getColumns(), req, true);
             finalLegend = buildFinalVLegend(columns, mergeOptions);
         }
 
@@ -343,7 +344,7 @@ public class LegendMerger {
         }
 
         if (layout == LegendLayout.VERTICAL) {
-            Column[] columns = createColumns(nodes, 0, 0);
+            Column[] columns = createColumns(nodes, 0, 0, null, false);
             finalLegend = buildFinalVLegend(columns, mergeOptions);
         }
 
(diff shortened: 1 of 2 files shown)


src/wms/src/main/java/org/geoserver/wms/legendgraphic/LegendMerger.java
```diff
@@ -299,7 +300,7 @@ public static BufferedImage mergeLegends(Rule[] rules, GetLegendGraphicRequest r
         }
 
         if (layout == LegendLayout.VERTICAL) {
-            Column[] columns = createColumns(nodes, mergeOptions.getColumnHeight(), mergeOptions.getColumns());
+            Column[] columns = createNRColumns(nodes, mergeOptions.getColumnHeight(), mergeOptions.getColumns(), req);
```
tbarsballe: To better match the pre-existing naming convention, this should probably be called createColumns, and the other method should be called createRasterColumns (since we are in the mergeLegends method, and the other one is called from mergeRasterLegends)

src/wms/src/main/java/org/geoserver/wms/legendgraphic/LegendMerger.java
```diff
@@ -418,13 +419,111 @@ public int getHeight() {
...
+                    i--;
+                    cn++;
+                    rc = 0;
+                    legendMatrix[cn] = new Column();
+                }
+            }
+        }
+
+        return legendMatrix;
+    }
```
tbarsballe: The vast majority of this method is identical to createColumns. Please refactor to avoid duplicated code.

src/wms/src/main/java/org/geoserver/wms/legendgraphic/LegendMerger.java
```diff
@@ -418,13 +419,111 @@ public int getHeight() {
...
+     * Checks the pixels for presence of colour against legend background
+     *
+     * @param img given row of the legend
+     * @param req general mechanism for acquiring legend symbols
+     * @return false if no colours are detected
+     *
+     */
+    public static boolean checkColour(BufferedImage img, GetLegendGraphicRequest req) {
+        int w = img.getWidth(null);
+        int h = img.getHeight(null);
```
tbarsballe: Just use `getWidth()` and `getHeight()` here.

src/wms/src/main/java/org/geoserver/wms/legendgraphic/LegendMerger.java
```diff
@@ -418,13 +419,111 @@ public int getHeight() {
...
+
+    /**
+     * Checks the pixels for presence of colour against legend background
+     *
+     * @param img given row of the legend
+     * @param req general mechanism for acquiring legend symbols
+     * @return false if no colours are detected
+     *
+     */
+    public static boolean checkColour(BufferedImage img, GetLegendGraphicRequest req) {
```
tbarsballe: Please match the java spelling of "Color"

aaime:
Looks like this one got stuck. @tbarsballe wondering if you are in a position to complete this work?

goudine:
Sorry for having this left alone for this long. Every month during the bug stomp I have been away on either Training or team meetup trips. Been meaning to clear this out of the way at next opportunity to work on Geoserver

src/wms/src/main/java/org/geoserver/wms/legendgraphic/LegendMerger.java
```diff
@@ -416,14 +417,20 @@ public int getHeight() {
...
     /**
-     * Creates legends columns for vertical layout according to max height and max columns limits
+     * Creates non-raster legend columns for vertical layout according to max height and max columns
+     * limits
      * 
      * @param nodes legend images
      * @param maxHeight maximum height of legend
      * @param maxColumns maximum number of columns
+     * @param req general mechanism for acquiring legend symbols
      *
```
tbarsballe: Also need to add checkColor to the javadocs

src/wms/src/main/java/org/geoserver/wms/legendgraphic/LegendMerger.java
```diff
@@ -343,7 +344,7 @@ public static BufferedImage mergeGroups(Rule[] rules, MergeOptions mergeOptions)
         }
 
         if (layout == LegendLayout.VERTICAL) {
-            Column[] columns = createColumns(nodes, 0, 0);
+            Column[] columns = createColumns(nodes, 0, 0, null, false);
```
tbarsballe: Do you know when this gets called, and why it doesn't need to truncate the legend?

src/wms/src/main/java/org/geoserver/wms/legendgraphic/LegendMerger.java
```diff
@@ -244,7 +244,7 @@ public static BufferedImage mergeRasterLegends(MergeOptions mergeOptions) {
         }
 
         if (layout == LegendLayout.VERTICAL) {
-            Column[] columns = createColumns(nodes, mergeOptions.getColumnHeight(), mergeOptions.getColumns());
+            Column[] columns = createColumns(nodes, mergeOptions.getColumnHeight(), mergeOptions.getColumns(), null, false);
```
tbarsballe: Do you know when this gets called, and why it doesn't need to truncate the legend?

goudine: It would only apply to the one legend type

tbarsballe:

PR #2292, opened 2017-04-30
[GEOS-8119] REST API migration broke ability to extract non SLD style in its native format

aaime:


aaime:
Note, the default output format cannot be SLD 1.1 as there is nothing able to write it (yep, xsd-sld cannot write SLD 1.1)

src/rest/src/main/java/org/geoserver/rest/converters/StyleWriterConverter.java
```diff
@@ -40,7 +40,7 @@ public StyleWriterConverter(String mimeType, Version version, StyleHandler handl
 
     @Override
     protected boolean supports(Class<?> clazz) {
-        return Style.class.isAssignableFrom(clazz) || StyleInfo.class.isAssignableFrom(clazz);
+        return RestWrapper.class.isAssignableFrom(clazz) || Style.class.isAssignableFrom(clazz) || StyleInfo.class.isAssignableFrom(clazz);
```
tbarsballe: This writer was only intended to write out style contents. I believe adding RestWrapper here will cause it to also try writing StyleInfo objects as well (which should be handled by the standard XStream writers instead).

src/restconfig/src/main/java/org/geoserver/rest/catalog/StyleController.java
```diff
@@ -225,7 +225,7 @@ private UriComponents getUriComponents(String name, String workspace,
         return uriComponents;
     }
 
-    @GetMapping(path = {"/styles/{styleName}", "/workspaces/{workspaceName}/styles/{styleName}"})
+    @GetMapping(path = {"/styles/{styleName}", "/workspaces/{workspaceName}/styles/{styleName}"}, produces = {MediaType.ALL_VALUE})
```
tbarsballe: Again, this should be handled by the standard XStream writers. There is a separate GetMapping for Style output. See styleSLDGet bellow - this is probably where the fix should be applied.

@jodygarnett - You picked up this from @smithkm at the end of the sprint - anything else you can add here? Do you remember how non-SLD styles were handled; it looks like styleSLDGet does not handle them.

src/extension/css/pom.xml
```diff
@@ -51,6 +51,23 @@
       <artifactId>easymockclassextension</artifactId>
       <scope>test</scope>
     </dependency>
+        <dependency>
+      <groupId>org.geoserver</groupId>
+      <artifactId>gs-rest</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
```
jodygarnett: This is odd, I thought we made a test case for css and ysld to confirm this endpoint would work with different style formats :(

src/extension/css/src/test/java/org/geoserver/rest/catalog/CssStyleControllerTest.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+    
+    @Test
+    public void getGetAsSLD10() throws Exception {
+        MockHttpServletResponse response = getAsServletResponse(
+                RestBaseController.ROOT_PATH + "/styles/test.sld");
+        assertEquals(200, response.getStatus());
+        assertEquals(SLDHandler.MIMETYPE_10, response.getContentType());
+        Document dom = dom(new ByteArrayInputStream(response.getContentAsByteArray()));
+        print(dom);
```
jodygarnett: please don't include print(dom) in final test

src/extension/css/src/test/java/org/geoserver/rest/catalog/CssStyleControllerTest.java
```diff
@@ -0,0 +1,167 @@
...
+        SLDHandler handler = new SLDHandler();
+        handler.encode(Styles.sld(s), SLDHandler.VERSION_10, false, out);
+        content = new String(out.toByteArray());
+        assertTrue(content.contains("<sld:Name>foo</sld:Name>"));
+        catalog.remove(styleInfo);
+    }
+
+    @Test
+    public void testPutCSS() throws Exception {
+        // step 1 create style info with correct format
```
jodygarnett: there is no step 2

src/extension/css/src/test/java/org/geoserver/rest/catalog/CssStyleControllerTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+        StyleInfo styleInfo = catalog.getStyleByName("bar");
+        Style s = styleInfo.getStyle();
+        ByteArrayOutputStream out = new ByteArrayOutputStream();
+
+        SLDHandler handler = new SLDHandler();
+        handler.encode(Styles.sld(s), SLDHandler.VERSION_10, false, out);
+        content = new String(out.toByteArray());
+        assertTrue(content.contains("<sld:Name>bar</sld:Name>"));
+
```
jodygarnett: Thanks for confirming this work, there is a new feature where CSS "validation" is performed, it should catch and reject case where a CSS file is uploaded that cannot generate a correct Style.

src/restconfig/src/main/java/org/geoserver/rest/catalog/StyleController.java
jodygarnett: We wrote a test case for ysld, but I am starting to fear we did not test for format conversion.

src/web/core/src/main/java/org/geoserver/web/wicket/js/codemirror/js/css.js
```diff
@@ -729,7 +729,8 @@ CodeMirror.defineMode("css", function(config, parserConfig) {
     "raster-color-map-type","raster-contrast-enhancement","raster-gamma","raster-geometry","raster-opacity",
     "raster-z-index","rotation","shield","shield-mime","size","sort-by","sort-by-group","stroke","stroke-dasharray",
     "stroke-dashoffset","stroke-geometry","stroke-linecap","stroke-linejoin","stroke-mime","stroke-offset",
-    "stroke-opacity","stroke-repeat","stroke-rotation","stroke-size","stroke-width","transform", "z-index" ],
+    "stroke-opacity","stroke-repeat","stroke-rotation","stroke-size","stroke-width","transform", "z-index", "composite", 
+    "composite-base" ],
```
jodygarnett: This appears to be an unrelated change; you may wish to isolate if you intend to backport the codemirror work.

src/restconfig/src/main/java/org/geoserver/rest/catalog/StyleController.java
aaime: Nope, you did not. The test was not doing GET requests. I added some in this pull request.

src/rest/src/main/java/org/geoserver/rest/converters/StyleWriterConverter.java
aaime: The code ended up here with a RestWrapper, I did not like it either but... we need an API that does the following:

- /styles/foo.html -> html, so rest wrapper
- /styles/foo.xml -> xml, so StyleInfo
- /styles/foo.sld -> sld, for native sld, should be a dump of the file, but for non sld, should be a conversion
- /styles/foo.xyz -> fow a dynamic xyz that a style handler handles, should be dumping the original style

Maybe have a get path that explicitly handles html, one for xml, and a final one returning a StyleInfo object (unwrapped) for everything else? Assuming MVC allows this setup that is.

src/extension/css/pom.xml
aaime: Did not find any for CSS, only for ysld and was only doing writes. Maybe an uncommitted bit?

src/web/core/src/main/java/org/geoserver/web/wicket/js/codemirror/js/css.js
aaime: It's already part of another commit, not sure why it's visible here too...

jodygarnett:
Merged, added a commit to confirm validation worked for CSS, and remove the print statement mentioned above.

Torben I tried removing the RESTWrapper (I remember being annoyed by handling it here during the sprint) but did not get anywhere.

PR #2294, opened 2017-05-01
[GEOS-7908] REST API: Add support for list=all when listing feature types

mattkrusz:
Adds support for a request to `/workspaces/<ws>/datastores/<ds>/featuretypes[.<format>]` with the query parameter `list=all`.

This parameter value is documented in the [REST API docs](http://docs.geoserver.org/stable/en/user/rest/api/featuretypes.html), but the code ignored the value until now, as discussed in [GEOS-7908](https://osgeo-org.atlassian.net/browse/GEOS-7908). The documented behavior is to return “The union of configured and available” feature types, so this commit updates the FeatureTypeController to return a list of all feature type names for the data source when “list=all”.


tbarsballe:
Looks good, has a test case. +1. Merged

jodygarnett:
Thanks @mattkrusz

PR #2295, opened 2017-05-01
Clean up MBStyleHandler, add test case

tbarsballe:


mattkrusz:
Changes look good to me! Thanks for the test case.

PR #2296, opened 2017-05-02
[GEOS-8114] Add a GetLegendGraphics option to return feature counts at the end of rule labels

aaime:


src/wms/src/main/java/org/geoserver/wms/legendgraphic/NoOpGraphics2D.java
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Fake Graphics2D, pretends to draw but actually does nothing (Very dishonest!)
+ * 
+ * Implementation note, the base class provides a lot of state management ensuring clients
+ * do not go NPE when asking for current clip, color, stroke and the like 
+ *
+ * @author Andrea Aime - GeoSolutions
+ */
+class NoOpGraphics2D extends AbstractGraphics2D {
```
jodygarnett: Is this an unrelated change?

jodygarnett:
This is a new feature, do you intend to backport in May?

aaime:
In May... not possible, not before June 2nd I guess no?

PR #2304, opened 2017-05-04
[GEOS-8125] making featureCount processor more robust for missing params

mlechner:
This PR imroves the featureCount option in GetLegendGraphics requests introduced with GEOS-8114 - ticket for this improvement is GEOS-8125 https://osgeo-org.atlassian.net/browse/GEOS-8125

This PR sets default values from LegendRequest class for WIDTH and HEIGHT if not yet set from request and sets BBOX as well from the layers envelope: legend.getLayerInfo().getResource().boundingBox()

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/legendgraphic/FeatureCountProcessor.java
+++ b/src/wms/src/main/java/org/geoserver/wms/legendgraphic/FeatureCountProcessor.java
@@ -290,6 +290,11 @@ class FeatureCountProcessor {
         // ... and bbox as well
         rawKvp.putIfAbsent("BBOX", legend.getLayerInfo().getResource().boundingBox().toString());
         kvp.putIfAbsent("BBOX", legend.getLayerInfo().getResource().boundingBox());
+        // ... and dont forget srs
+        rawKvp.putIfAbsent("SRS", legend.getLayerInfo().getResource().getSRS());
+        kvp.putIfAbsent("SRS", legend.getLayerInfo().getResource().getSRS());
+
+        
 
         // remove decoration to avoid infinite recursion
         final Map formatOptions = (Map) kvp.get("FORMAT_OPTIONS");
--- a/src/wms/src/main/java/org/geoserver/wms/legendgraphic/FeatureCountProcessor.java
+++ b/src/wms/src/main/java/org/geoserver/wms/legendgraphic/FeatureCountProcessor.java
@@ -16,6 +16,7 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
+import com.vividsolutions.jts.geom.Envelope;
 import org.apache.commons.lang.StringUtils;
 import org.geoserver.catalog.NamespaceInfo;
 import org.geoserver.ows.util.CaseInsensitiveMap;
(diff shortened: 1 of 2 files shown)


aaime:
Thank you! By [project rules](https://github.com/geoserver/geoserver/blob/master/CONTRIBUTING.md), that needs to have a test. I believe you can just take one of the existing ones and remove the parameters you're defaulting.

src/wms/src/test/java/org/geoserver/wms/legendgraphic/FeatureCountLegendGraphicTest.java
```diff
@@ -162,6 +162,110 @@ public void testStatesFull() throws Exception {
...
+        assertEquals(4, rules.length);
+        assertLabel("2M - 4M (10)", rules[0]);
+        assertLabel("< 2M (16)", rules[1]);
+        assertLabel("> 4M (23)", rules[2]);
+        // this is the rule for outline and text symbolizer, Alaska and Hawaii are not in the
+        // map but Washington DC is and it's not a state (50 - 2 + 1)
+        assertLabel("(49)", rules[3]);
+    }
+
+    /*@Test
```
aaime: What's up with this one?

mlechner: testStatesMissingHeightWidthBboxSrs() fails when activated returning different values for the featureCounts than expected - it does not crash.

The curious thing is that if I run the request used in the test manually (on a jetty:run geoserver web-app instance with GESERVER_DATA_DIR from geoserver/data/release) the request works as expected and returns the values set in asserts in the Test (which are the same than all the test expect). Tested manually with
http://localhost:8080/geoserver/wms?REQUEST=GetLegendGraphic&VERSION=1.1.1&FORMAT=image/png&LAYER=topp:states&style=&legend_options=countMatched:true

The only difference is that I do not have the style param set on manual test, but that can not be the thing. Any idea what may cause this curious behaviour?

src/wms/src/main/java/org/geoserver/wms/legendgraphic/FeatureCountProcessor.java
```diff
@@ -280,8 +281,16 @@ private GetMapRequest parseAssociatedGetMap(LegendRequest legend, Rule[] rules)
...
-        
+        rawKvp.put("HEIGHT", rawKvp.get("SRCHEIGHT"));
+        // ... set default values if not yet set
+        rawKvp.putIfAbsent("HEIGHT", String.valueOf(GetLegendGraphicRequest.DEFAULT_HEIGHT));
+        rawKvp.putIfAbsent("WIDTH", String.valueOf(GetLegendGraphicRequest.DEFAULT_WIDTH));
+        kvp.putIfAbsent("HEIGHT", String.valueOf(GetLegendGraphicRequest.DEFAULT_HEIGHT));
+        kvp.putIfAbsent("WIDTH", String.valueOf(GetLegendGraphicRequest.DEFAULT_WIDTH));
+        // ... and bbox as well
+        rawKvp.putIfAbsent("BBOX", legend.getLayerInfo().getResource().boundingBox().toString());
+        kvp.putIfAbsent("BBOX", legend.getLayerInfo().getResource().boundingBox());
```
aaime: I believe this will not work correctly in WMS 1.3 due to axis flip, at a minimum I'd add a test for WMS 1.3 without bbox and see if it works or not.

src/wms/src/test/java/org/geoserver/wms/legendgraphic/FeatureCountLegendGraphicTest.java
aaime: Ah, this is due to map generalization, some features become so small that they are not going to be painted. You can fix this by adding 
renderer.setGeneralizationDistance(0); 
in FeatureCountProcessor at line 112 (right after the super.onBeforeRender(renderer);

src/wms/src/main/java/org/geoserver/wms/legendgraphic/FeatureCountProcessor.java
```diff
@@ -280,8 +281,16 @@ private GetMapRequest parseAssociatedGetMap(LegendRequest legend, Rule[] rules)
         rawKvp.put("STYLES", "");
         // ... width and height
         rawKvp.put("WIDTH", rawKvp.get("SRCWIDTH"));
-        rawKvp.put("HEIGTH", rawKvp.get("SRCHEIGHT"));
-        
+        rawKvp.put("HEIGHT", rawKvp.get("SRCHEIGHT"));
+        // ... set default values if not yet set
+        rawKvp.putIfAbsent("HEIGHT", String.valueOf(GetLegendGraphicRequest.DEFAULT_HEIGHT));
```
aaime: Having a value of "20" is unreasonable, most maps with scale dependencies will simply not show at that scale. Also, the sizing is not proportional to the map own bounding box, creating a different scale denominator in south/north east/west directions. See https://github.com/geoserver/geoserver/blob/master/src/wms/src/main/java/org/geoserver/wms/DefaultWebMapService.java#L445 (maybe it's possible to just call that method, something to try out)

src/wms/src/test/java/org/geoserver/wms/legendgraphic/FeatureCountLegendGraphicTest.java
mlechner: added. thnx.

src/wms/src/main/java/org/geoserver/wms/legendgraphic/FeatureCountProcessor.java
mlechner: will look at that later - thnx for the comments

aaime:
Superseded by https://github.com/geoserver/geoserver/pull/2400

PR #2312, opened 2017-05-08
Fix for GEOS-8128 - SLD static legend dimensions not respected

afabiani:


src/wms/src/main/java/org/geoserver/wms/legendgraphic/GetLegendGraphicKvpReader.java
```diff
@@ -248,6 +246,12 @@ private LegendRequest addLayer(LayerInfo layerInfo, GetLegendGraphicRequest requ
             LegendInfo legendInfo = resolveLegendInfo(layerInfo.getLegend(),request);
             if(legendInfo != null ){
```
aaime: This block of code is now repeated 3 times in different places, could it be refactored into a method call?

src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/GetLegendGraphicTest.java
```diff
@@ -108,13 +108,23 @@ public void testCustomLegend() throws Exception {
         BufferedImage expected = ImageIO.read( resource.file() );
         
         assertEquals( getPixelColor(expected,10,2).getRGB(), getPixelColor(image,10,2).getRGB() );
+
+        // test external image dimensions
```
aaime: So the test checks the image size and not the value set in the setup... I suppose they are the same? Could be useful to have a test where they are not.

afabiani: @aaime Done!

src/wms/src/main/java/org/geoserver/wms/legendgraphic/GetLegendGraphicKvpReader.java
afabiani: @aaime Done!

PR #2313, opened 2017-05-08
Add style generation support for MBStyle

tbarsballe:
Add style generation support for MBStyle, plus some improvements to StyleGenerator error messages.


mattkrusz:
Looks good to me!

If it doesn't make sense to have a "generic" style yet, is it possible to remove the option from the dropdown? --  (instead of causing a red "Errors occurred generating the style" message at the top)

tbarsballe:
At present, the drop-down is generated entirely independently of the current style format.
It would be possible to change this, but likely nontrivial, and require API changes to StyleHandler (and consequently require a GSIP).

tbarsballe:
Backported to 2.11.x

PR #2321, opened 2017-05-09
MBStyle - add 'Generic' stlye generation

tbarsballe:
Depends upon https://github.com/geotools/geotools/pull/1580


tbarsballe:
Backported to 2.11.x

PR #2328, opened 2017-05-10
[GEOS-8117] Fix CITE WFS compliance

tbarsballe:
Preliminary fix for WFS CITE failures caused by https://github.com/geoserver/geoserver/pull/2289

While the new test case passes, [CITE tests still fail](http://ares.boundlessgeo.com/jenkins/job/cite-wfs-1.1-troubleshooting/3/consoleFull), so this is not yet mergable.

I'm also not quite happy with the fix itself, as it seems quite hacky, but I can't see a better way of doing this so far.

### DIFF
--- a/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
+++ b/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
@@ -1151,6 +1151,10 @@ public class Dispatcher extends AbstractController {
         if (matches.size() > 1) {
             List vmatches = new ArrayList(matches);
 
+            if (isCiteCompliant() && version == null && "WFS".equals(id)) {
+                version = new Version("1.1.0");
+            }
+
             //match up the version
             if (version != null) {
                 //version specified, look for a match
--- a/src/wfs/src/test/java/org/geoserver/wfs/GetFeatureTest.java
+++ b/src/wfs/src/test/java/org/geoserver/wfs/GetFeatureTest.java
@@ -398,6 +398,24 @@ public class GetFeatureTest extends WFSTestSupport {
     }
 
     @Test
+    public void testStrictComplianceVersionNegotiation() throws Exception {
+        GeoServer geoServer = getGeoServer();
+        WFSInfo service = geoServer.getService(WFSInfo.class);
+        try {
+            service.setCiteCompliant(true);
+            geoServer.save(service);
+
+            Document doc = getAsDOM("wfs?service=WFS&request=GetCapabilities");
+            XMLAssert.assertXpathEvaluatesTo("true", "//wfs:WFS_Capabilities/@version='1.1.0'", doc);
+        } finally {
+            service.setCiteCompliant(false);
+            geoServer.save(service);
+        }
+        Document doc = getAsDOM("wfs?service=WFS&request=GetCapabilities");
+        XMLAssert.assertXpathEvaluatesTo("false", "//wfs:WFS_Capabilities/@version='1.1.0'", doc);
+    }
+
+    @Test
     public void testRequestDisabledResource() throws Exception {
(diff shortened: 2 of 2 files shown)


src/wfs/src/test/java/org/geoserver/wfs/GetFeatureTest.java
```diff
@@ -398,6 +398,24 @@ public void testStrictComplianceBBoxValidator() throws Exception {
     }
 
     @Test
+    public void testStrictComplianceVersionNegotiation() throws Exception {
```
bencaradocdavies: This method should be in `GetCapabilitiesTest`.

src/wfs/src/test/java/org/geoserver/wfs/GetFeatureTest.java
```diff
@@ -398,6 +398,24 @@ public void testStrictComplianceBBoxValidator() throws Exception {
...
+        GeoServer geoServer = getGeoServer();
+        WFSInfo service = geoServer.getService(WFSInfo.class);
+        try {
+            service.setCiteCompliant(true);
+            geoServer.save(service);
+
+            Document doc = getAsDOM("wfs?service=WFS&request=GetCapabilities");
```
bencaradocdavies: Oh, the horror. This test fails if the `ows` endpoint is used. I am pretty sure all the CITE tests use the `ows` endpoint.

src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
```diff
@@ -1151,6 +1151,10 @@ Service findService(String id, String ver, String namespace) throws ServiceExcep
         if (matches.size() > 1) {
             List vmatches = new ArrayList(matches);
 
+            if (isCiteCompliant() && version == null && "WFS".equals(id)) {
```
bencaradocdavies: The problem with this implementation is that CITE tests run with the `ows` endpoint, for which `isCiteCompliant()` is `false`. You can find the services and examine them for this setting, but you can never access their content because `gs-wfs` in which `WFSInfo` is defined depends on `gs-ows` and is out of scope.

I think an alternate solution is required, perhaps disabling WFS 2.0.0 either in code or in the `wfs.xml` settings. I do not think the latter will be honoured. Perhaps we need something to disable WFS 2.0.0 when WFS CITE compliance is enabled?

aaime: @nmco is just working on making WFS 2.0 pass CITE tests... so maybe avoid disabling it fully? Maybe we'll need some way to selectively disable versions (would be useful for WCS CITE testing as well)

nmco: I agree with @aaime. The way I see it since the dispatcher needs to be aware of CITE compliance we need to make that awarenesses more configurable than a simple flag.

src/wfs/src/test/java/org/geoserver/wfs/GetFeatureTest.java
tbarsballe: Yep, I started with a GetFeature test, then noticed it was supposed to be getCapabilities and forgot to move it. Thanks for the reminder.

src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
tbarsballe: I agree as well. That said, I feel like the way we have things set up right now, implementing this sort of thing is going to be a bit tricky.

If `isCiteCompliant()` is `false` for `ows`, is there any check we can do for running cite tests?

tbarsballe:
Ongoing discussion on email list, closing this.

PR #2329, opened 2017-05-11
[GEOS-8041] Do bounding box access checks for all GWC services

smithkm:
Supersedes #2286


dvntucker:
Don't see anything technically wrong with it. I don't know too much about the issue/approach, but don't see anything obviously wrong from that perspective either.

PR #2330, opened 2017-05-11
[GEOS-5151] DataAccess full type name listing called multiple times in OGC request against SQL views

aaime:


dromagnoli:
Looks good to me.
+1

aaime:
Build failure is due to breaking API changes in GWC, merging.

PR #2331, opened 2017-05-11
[GEOS-8136] Adds composite and composite-base properties in CSS doc

Gnafu:


Gnafu:
Misses stroke-composite, fill-composite, etc..

PR #2338, opened 2017-05-11
[GEOS-8138] Prevent StyleFinder from overwriting existing MediaType mappings

mattkrusz:
Updates StyleFinder to only create MediaType mappings for StyleHandler extensions IF the extension is not already being used. This will prevent styles that use general extensions (e.g., “style.json”) from interfering with the existing Restlet content negotiation (which allows .json to be used as a shortcut for an “application/json” accept header).

This effectively preserves the current behavior of the REST API for styles: using *.json will still always map to "application/json" and thus return the style info object in json format. If a user specifically wants the style definition body from a StyleHandler with a conflicting extension, they must use the StyleHandler's mime type (e.g., "application/vnd.geoserver.mbstyle+json").

Issue: https://osgeo-org.atlassian.net/browse/GEOS-8138

### DIFF
--- a/src/community/mbstyle/src/test/java/org/geoserver/community/mbstyle/web/MBStyleHandlerTest.java
+++ b/src/community/mbstyle/src/test/java/org/geoserver/community/mbstyle/web/MBStyleHandlerTest.java
@@ -2,13 +2,15 @@ package org.geoserver.community.mbstyle.web;
 
 import org.geoserver.catalog.Styles;
 import org.geoserver.community.mbstyle.MBStyleHandler;
+import org.geoserver.rest.format.MediaTypes;
 import org.geoserver.test.GeoServerSystemTestSupport;
 import org.geotools.styling.*;
 import org.junit.Test;
+import org.restlet.data.MediaType;
 
 import java.io.IOException;
+import static org.junit.Assert.*;
 
-import static org.junit.Assert.assertNotNull;
 
 public class MBStyleHandlerTest extends GeoServerSystemTestSupport {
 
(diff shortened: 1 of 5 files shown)


src/community/mbstyle/src/test/java/org/geoserver/community/mbstyle/web/MBStyleHandlerTest.java
```diff
@@ -2,13 +2,15 @@
...
 import org.geoserver.test.GeoServerSystemTestSupport;
 import org.geotools.styling.*;
 import org.junit.Test;
+import org.restlet.data.MediaType;
 
-import java.io.IOException;
+import static org.junit.Assert.*;
 
-import static org.junit.Assert.assertNotNull;
+import java.io.IOException;
```
jodygarnett: sort order changing?

mattkrusz: Fixed, repushed

mattkrusz:
Fixed & squashed the import shuffle; everything else is the same.

src/restconfig/src/main/java/org/geoserver/catalog/rest/StyleFinder.java
```diff
@@ -99,8 +99,12 @@ public void onApplicationEvent(ApplicationEvent event) {
         if (event instanceof ContextLoadedEvent) {
             // register style format mime types
             for (StyleHandler sh : Styles.handlers()) {
-                Version ver = sh.getVersions().iterator().next();
-                MediaTypes.registerExtension(sh.getFileExtension(), new MediaType(sh.mimeType(ver)));
+                // Only map the StyleHandler's extension to its mime type 
+                // if the extension does not conflict with existing mappings.
+                if (MediaTypes.getMediaTypeForExtension(sh.getFileExtension()) == null) {
```
tbarsballe: Very minor, but a class-level javadoc (or other sort of documentation that is more visible than an inline comment) mentioning this behavior might be a good idea. I could imagine in the future someone trying to add a new style format and it "inexplicibly" not getting picked up.

mattkrusz: Good call -- added a class-level comment (squashed).

PR #2339, opened 2017-05-11
[GEOS-2749] Reproject geometries in WMS GetFeatureInfo responses when info_format is GML

nmco:
This pull request make WMS GetFeatureInfo request return the features reprojected in the requested SRS. The reprojection will work for normal vector layers, complex features and cascaded WMS layers.

New tests cases where added and app-schema tests updated.

Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-2749


aaime:
I don't see updates to the app-schema tests, am I overlooking them or they were not needed in the end?
Also, this is a change to a (wrong) behavior that GeoServer had... since it has GetFeatureInfo. Is there a way to turn it off, I bet there is a lot of applications out there relying on the current behavior (e.g., they might end up reprojecting data again now).

nmco:
Regarding the app-schema tests: is the first edited file ``WmsGetFeatureInfoTest.java`` :smile:, the test already existed but there was no check for use SRS.

nmco:
Regarding the option to turn this off, I will add it as a flag in the WMS service.

nmco:
And thanks for the review :bowing_man: (for some reason I cannot introduce with lines).

nmco:
Flag to deactivate features reprojection for WMS GetFeatureInfo requests added. Tests cases were updated to in account the possibility to disable the reprojection. Documentation was also added:

imagem

Note that all the changes to public APIs were additive, backward compatibility was not broke.

aaime:
@nmco fix, test, UI, docs... yep, looks good to me!

PR #2348, opened 2017-05-12
[GEOS-8140] GeoServer uses invalid mime type for GML 3.2 responses

nmco:
This PR just switches the preference order of GML 3.2 mime types. Both will still be supported by GeoServer when parsing but ``application/gml+xml; version=3.2`` will be used when encoding a GML 3.2 response.

Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-8140


bencaradocdavies:
@nmco the title of this PR seems to be the opposite of what it does.

bencaradocdavies:
Full build plus app-schema online tests pass with this change. See mailing list for discussion.

nmco:
It has not been done yet, better it was done but I still need to update the tests and deug why InteliJ keeps messing with properties files: https://github.com/nmco/geoserver/commit/6406e281b6c7b0ad5351b1f4a070d8ac6c9d601f

nmco:
A flag has been added to override to override the GML 3.2 MIME type:
image

Documentations and tests were updated to take this into account. Note that right now the MIME type overriding is only available for GML 3.2 but the code was made generic enough to make with work with any other GML version if needed (only some configuration in the ``WFSAdminPage`` would be required).

bencaradocdavies:
@nmco that looks great to me. I can confirm that a full build passes with this change cherry-picked onto master. I can confirm that the MIME type is changed when the setting is changed (tested with "curl -i") and that this setting is serialised to wfs.xml. @aaime I trust that this addresses your concerns.

bencaradocdavies:
@nmco that was a solid piece of work. Tell me that was Google Translate, or do you also speak French and Portuguese?

aaime:
He's native in both languages :-p

PR #2349, opened 2017-05-15
added reference docs for mapbox styles.

vickdw:


tbarsballe:
This seems to duplicate https://github.com/geotools/geotools/pull/1584. We should just have the spec in one place (GeoServer or GeoTools - probably GeoTools), and link from the other as necessary.

tbarsballe:
This can still just be a link to the geotools docs.

tbarsballe:
Travis build failed:

```
     [exec] Warning, treated as error:
     [exec] /home/travis/build/geotools/geotools/docs/user/unsupported/mbstyle/index.rst:44: SEVERE: Unexpected section title.

     [exec] Color
     [exec] '''''
```

tbarsballe:
In the future, we should try and pull the spec .rst directly from geotools, rather than trying to keep both up-to-date.

PR #2356, opened 2017-05-18
Adding javadocs to vector tiles

smithkm:
Added JavaDocs while I was figuring out how the Vectortiles module works.


DBlasby:
ok to merge

PR #2362, opened 2017-05-26
[GEOS-7519] Allow the definition of keywords in layer groups

nmco:
Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-7519

This pull request adds the possibility to associate keywords with a layer group. The keywords can be configured through GeoServer web UI and will be encoded in WMS 1.1.1 and WMS 1.3.0 capabilities documents.

Tests cases were added in the web, wms and rest modules.

GeoServer documentation and swagger REST API documentation were also updated.

Changes in ``LayerGoupInfo`` interface are additive and default implementations of the new added methods ensure backward compatibility.


src/main/src/main/java/org/geoserver/catalog/impl/LayerGroupInfoImpl.java
```diff
@@ -75,9 +77,22 @@
...
+    public List<KeywordInfo> getKeywords() {
+        return keywords;
+    }
+
+    @Override
+    public void setKeywords(List<KeywordInfo> keywords) {
+        if (keywords == null) {
+            this.keywords = new ArrayList<>();
+        }
+        this.keywords = keywords;
```
aaime: So you first set this.keywords to a non null value, and then set it to null? I guess you wanted something like:

````
if (keywords == null) {
     keywords = new ArrayList<>();
}
this.keywords = keywords;

File: src/main/src/main/java/org/geoserver/catalog/LayerGroupInfo.java [hunk 2]
```diff
@@ -198,6 +199,26 @@ public Integer getCode() {
...
+        return Collections.emptyList();
+    }
+
+    /**
+     * Set the keywords of this layer group. The provided keywords will override any existing
+     * keywords no merge will be done.
+     *
+     * @param keywords new keywords of this layer group
+     */
+    default void setKeywords(List<KeywordInfo> keywords) {
```
aaime: Nope, in the GeoServer catalog lists cannot be set, they are "live" list, you get them and modify them. Setters are often found in Impl classes to help XStream.
Check ResourceInfo:

````
    /**
     * A collection of keywords associated with the resource.
     * 
     * @uml.property name="keywords"
     */
    List<KeywordInfo> getKeywords();

    /**
     * List of keyword values, derived from {@link #getKeywords()}.
     */
    List<String> keywordValues();
````

File: src/restconfig/src/test/java/org/geoserver/rest/catalog/LayerGroupControllerTest.java [hunk 3]
```diff
@@ -23,12 +18,28 @@
...
-import org.w3c.dom.Document;
-
 import org.springframework.mock.web.MockHttpServletResponse;
+import org.w3c.dom.Document;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
+import static org.custommonkey.xmlunit.XMLAssert.assertXpathEvaluatesTo;
```
aaime: IntelliJ vs Eclipse clash detected :-p (Eclipse puts static imports at top, I guess IntelliJ does the opposite?)

src/restconfig/src/main/java/org/geoserver/rest/catalog/LayerGroupController.java
```diff
@@ -198,6 +203,11 @@ public boolean supports(MethodParameter methodParameter, Type targetType, Class<
 
     @Override
     public void configurePersister(XStreamPersister persister, XStreamMessageConverter converter) {
+        // add keyword object configuration
+        XStream xstream = persister.getXStream();
```
aaime: Suspicious, why doesn't ServiceInfo and ResourceInfo need the same treatment? They also have a keywords field. Also  check XStreamPersister, a custom converter is registered in that class for ResourceInfo and ServiceInfo (even with the same output, it would be best to maintain consistency)

aaime:
Oh, forgot a thing, keywords are also used in CSW, since layer groups are searchable in CSW they should also expose keywords there.

src/main/src/main/java/org/geoserver/catalog/impl/LayerGroupInfoImpl.java
nmco: :man_facepalming:

src/main/src/main/java/org/geoserver/catalog/LayerGroupInfo.java
nmco: I see indeed not sure how I missed that, thanks.

src/restconfig/src/test/java/org/geoserver/rest/catalog/LayerGroupControllerTest.java
nmco: I missed this one thanks !

src/restconfig/src/main/java/org/geoserver/rest/catalog/LayerGroupController.java
nmco: ServiceInfo and ResourceInfo will return something like this:

```
<keywords>
	<string>WFS</string>
	<string>WMS</string>
	<string>GEOSERVER</string>
</keywords>
```

but with the default configuration, layer goups controller return this instead:

```
<keywords>
	<org.geoserver.catalog.Keyword>keyword</org.geoserver.catalog.Keyword>
</keywords>
```
and could not understand why, anyway I agree with you all should use the same tag name.

aaime: Set them up to use the same converter in XStreamPersister and this configuration won't be necessary

nmco:
Layer groups keywords exposed in CSW, a test that checks it was added.

src/restconfig/src/main/java/org/geoserver/rest/catalog/LayerGroupController.java
nmco: The same converter is used and he correctly encodes the keywords content is just the tag of the element that is not like ``<string>``. Anyway I'm explicitly forcing the tag to be ``<string>``.

aaime: Yes, but if you set it up in XStreamPersister instead of a REST controller, you'll get consistency between disk format and network format...

nmco:
I finally understood why the correct tag was not used to serialize the keywords list ... specific configuration was removed from layers groups controller.

PR #2363, opened 2017-05-26
[GEOS-7893] GeoServerTabelPanel shows the correct number of results

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-7893

When updateMatch is called, dataView has not yet been updated with the filtered results (dataProvider has). This fix makes update rely more upon the dataProvider info for computing values.

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/wicket/GeoServerTablePanel.java
+++ b/src/web/core/src/main/java/org/geoserver/web/wicket/GeoServerTablePanel.java
@@ -580,12 +580,10 @@ public abstract class GeoServerTablePanel<T> extends Panel {
          * Updates the label given the current page and filtering status
          */
         void updateMatched() {
+            size = dataProvider.size();
             fullSize = dataProvider.fullSize();
-            first = first(fullSize);
-            last = last(fullSize);
-            if (dataProvider.getKeywords() != null) {
-                size = dataProvider.size();
-            }
+            first = first(size);
+            last = last(size);
         }
         
         public IModel<String> model() {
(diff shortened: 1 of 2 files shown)


aaime:
+0, seems good but I don't quite understand some of the changes :)

src/web/core/src/main/java/org/geoserver/web/wicket/GeoServerTablePanel.java
```diff
@@ -612,16 +611,17 @@ long first(long fullSize) {
...
         /**
          * User oriented index of the last item in the current page
+         *
+         * @param size The total number of items matched by the current filter
          */
-        long last(long fullSize) {
+        long last(long size) {
             
-            long count = dataProvider.getKeywords() != null ? 
-                    dataView.getPageCount() : optGetPageCount(fullSize);
```
tbarsballe: dataView is not accurate here, rely only on dataProvider (via optGetPageCount)

src/web/core/src/main/java/org/geoserver/web/wicket/GeoServerTablePanel.java
```diff
@@ -580,12 +580,10 @@ public PagerDelegate() {
          * Updates the label given the current page and filtering status
          */
         void updateMatched() {
+            size = dataProvider.size();
             fullSize = dataProvider.fullSize();
-            first = first(fullSize);
-            last = last(fullSize);
-            if (dataProvider.getKeywords() != null) {
-                size = dataProvider.size();
-            }
```
tbarsballe: `dataProvider.getKeywords() != null` checks if there is anything in the filter. Removed the check so that the count gets reset properly when the filter is cleared. Use size (size with filter) instead of fullSize (size without) for calculations (both these are global)

tbarsballe:
The main idea is to rely mainly on the state of dataProvider to calculate total list size and number of items on the current page, rather than on dataView.

I've added a few comments inline to explain the changes, but its certainly still a bit opaque. Took me a few trips through the debugger to figure out how it was supposed to work.

PR #2366, opened 2017-05-26
[GEOS-8155] Allow deletion of workspaced styles that share names with default styles

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-8155


aaime:
Hum... looks good but I have a question, what happens if someone upgrades and has moved the point style to a workspace? Is GeoServer just re-creating it? (I believe so, but it's worth making sure).
Also, will this change allow the user to move a workspaced style named "point" to global? Or will it cause the error messages to come from the catalog instead of coming from the UI (difference being i18n, and possibly readability)

tbarsballe:
After a quick test, it looks like GeoServer does not recreate the style if it has been workspaced, only if it has been renamed.

If the user has a workspaced style named point, the will be able to move it to global as long as a global style named point does not already exist (otherwise, moving the style is prevented by preexisting checks). If the style does exist globally, the error is a regular UI failure message: `java.lang.IllegalArgumentException: Style named 'point' already exists`.

PR #2367, opened 2017-05-26
[GEOS-7958] Disallow workspacing default styles

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-7958


dvntucker:
Looks good to me.

PR #2368, opened 2017-05-26
[GEOS-7959] Disallow changing default styles

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-7959

This prevents the global default styles (point, line, polygon, raster, generic) from being renamed, assigned to a workspace, or deleted. This change is implemented at the CatalogImpl level.

Several changes have already been made to the UI to support this; those changes will remain as they provide useful behaviour (displaying dialogs and error messages)

This change will automatically affect the REST API. REST tests have been added as a separate commit to make any future backporting easier (given the resent REST rewrite).

Since this is a change to a core catalog class, care should be taken when deciding whether or not to backport. I would be happiest waiting a month on master before backporting, even though this is technically a bugfix.


dvntucker:
Looks good to me.

PR #2369, opened 2017-05-29
Geos 7923: Concurrent GeoRSS requests contain malformed template content

Tomboyo:
Patch for [GEOS-7923](https://osgeo-org.atlassian.net/browse/GEOS-7923
)

The AtomUtils class held a static reference to a FeatureTemplate instance, which caused issues in a concurrent context (there was a TODO asking if they were threadsafe). This patch transfers ownership of the feature template from the utility class to the transformer instances handling the request. I've added a test case that checks for concurrency violations by comparing the output of WMS requests made by multiple threads, but it's not a foolproof test, as one could imagine. It should help catch some such violations, though.

Corporate contributor license agreement is pending.

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/georss/AtomGeoRSSTransformer.java
+++ b/src/wms/src/main/java/org/geoserver/wms/georss/AtomGeoRSSTransformer.java
@@ -24,25 +24,25 @@ import com.vividsolutions.jts.geom.Geometry;
 import com.vividsolutions.jts.geom.GeometryCollection;
 
 public class AtomGeoRSSTransformer extends GeoRSSTransformerBase {
-    
+
     private WMS wms;
 
-    public AtomGeoRSSTransformer(WMS wms){
+    public AtomGeoRSSTransformer(WMS wms) {
         this.wms = wms;
     }
-    
+
     public Translator createTranslator(ContentHandler handler) {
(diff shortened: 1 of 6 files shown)


aaime:
Do you know if the contribution agreement was ever filed?

Tomboyo:
The paperwork is going through our legal department now. I think it's been
on the back burner for a while, but it should be done soon-ish.

On Jul 28, 2017 5:16 AM, "Andrea Aime" <notifications@github.com> wrote:

Do you know if the contribution agreement was ever filed?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/geoserver/geoserver/pull/2369#issuecomment-318603469>,
or mute the thread

tbarsballe:
Hi @Tomboyo , do you have an update on the status of the contribution agreement?

Tomboyo:
I have followed up on this with my employer:

My understanding is that our legal department has declined to sign the contributor agreement. For context, my immediate supervisors thought it would be wise to submit this PR, but once we found out about the corporate contributor agreement different departments became involved. They have decided for their own reasons that this PR is not something they will consent to.

Since I worked on this code on behalf of my company, I think I am obligated to retract this submission. I am not sure if I am correct in that assumption, however.

tbarsballe:
Well, that's unfortunate.

I believe you are correct, I'll close the PR.

PR #2375, opened 2017-05-30
WFSPPIO fails to remove MetaDataProperty attributes

ianturton:
Fix for [GEOS-8158](https://osgeo-org.atlassian.net/browse/GEOS-8158)

### DIFF
--- a/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ppio/WFSPPIO.java
+++ b/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ppio/WFSPPIO.java
@@ -26,6 +26,7 @@ import org.geotools.data.simple.SimpleFeatureIterator;
 import org.geotools.data.store.ReprojectingFeatureCollection;
 import org.geotools.feature.FeatureCollection;
 import org.geotools.feature.simple.SimpleFeatureTypeBuilder;
+import org.geotools.gml3.GML;
 import org.geotools.referencing.CRS;
 import org.geotools.util.logging.Logging;
 import org.geotools.wfs.v1_0.WFSConfiguration;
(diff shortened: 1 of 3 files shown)


src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ppio/WFSPPIO.java
```diff
@@ -49,7 +50,9 @@
 
     Configuration configuration;
     private static final Logger LOGGER = Logging.getLogger(WFSPPIO.class);
-    
+    private static final String metaData = GML.metaDataProperty.getLocalPart();
```
aaime: Code conventions, constants should be uppercase

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ppio/WFSPPIO.java
```diff
@@ -129,14 +132,16 @@ public Object decode(Object input) throws Exception {
...
     private SimpleFeatureCollection eliminateFeatureBounds(SimpleFeatureCollection fc) {
         final SimpleFeatureType original = fc.getSchema();
+        
         List<String> names = new ArrayList<String>();
         boolean alternateGeometry = false;
         for(AttributeDescriptor ad : original.getAttributeDescriptors()) {
             final String name = ad.getLocalName();
-            if(!"boundedBy".equals(name) && !"metadataProperty".equals(name)) {
+
+            if(!boundedBy.equals(name) && !metaData.equals(name)) {
```
aaime: So... was it just a typo in the name of "metadataProperty"?

ianturton: It was just a typo, but I thought moving to use constants would reduce the "magic"

PR #2376, opened 2017-05-30
[GEOS-8163] Add display name mappings for GeoPackage output formats

tbarsballe:
Fix for https://osgeo-org.atlassian.net/browse/GEOS-8163

Before:

image

After:

image


tbarsballe:
Configuration for all the other formats added by various extensions (e.g. MBTiles, VectorTiles, etc.) Should probably also be configured in those extensions rather than in the main demo GeoServerApplication.properties file.

aaime:
+1

PR #2381, opened 2017-05-31
[GEOS-8171] adding a generic method to clear caches for external graphic factorie…

vickdw:
…s.  This references work done in geotools PR 1553 found here. https://github.com/geotools/geotools/pull/1553

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/WMSLifecycleHandler.java
+++ b/src/wms/src/main/java/org/geoserver/wms/WMSLifecycleHandler.java
@@ -5,11 +5,19 @@
  */
 package org.geoserver.wms;
 
-import java.awt.Font;
-import java.io.File;
-import java.io.IOException;
+import org.geoserver.config.GeoServerDataDirectory;
+import org.geoserver.config.impl.GeoServerLifecycleHandler;
+import org.geoserver.platform.resource.Resource;
+import org.geoserver.platform.resource.Resources;
+import org.geotools.renderer.style.*;
+import org.geotools.util.logging.Logging;
+import org.springframework.context.ApplicationEvent;
+import org.springframework.context.ApplicationListener;
+import org.springframework.context.event.ContextRefreshedEvent;
+
+import java.awt.*;
 import java.util.ArrayList;
-import java.util.Collection;
+import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.LinkedBlockingQueue;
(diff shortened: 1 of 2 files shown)


src/wms/src/test/java/org/geoserver/wms/WMSTest.java
```diff
@@ -44,6 +50,16 @@
...
+        Catalog catalog = getCatalog();
+        testData.addStyle("burg-fill", "./map/burg-fill.sld", WMSTest.class, catalog);
+
+        // copy over the svg icon
+        File styles = new File(testData.getDataDirectoryRoot(), "styles");
+        File burg = new File(styles, "burg02.svg");
+        try(InputStream is = getClass().getResourceAsStream("./map/burg02.svg"); FileOutputStream fos = new FileOutputStream(burg)) {
+            IOUtils.copy(is, fos);
+        }
+
```
tbarsballe: This doesn't seem to be referenced anywhere in the test case that was added. Why is this here?

src/wms/src/main/java/org/geoserver/wms/WMSLifecycleHandler.java
```diff
@@ -18,18 +26,7 @@
...
-import org.geoserver.config.impl.GeoServerLifecycleHandler;
-import org.geoserver.platform.resource.Resource;
-import org.geoserver.platform.resource.Resources;
-import org.geotools.renderer.style.FontCache;
-import org.geotools.renderer.style.ImageGraphicFactory;
-import org.geotools.renderer.style.SVGGraphicFactory;
-import org.geotools.util.logging.Logging;
-import org.springframework.context.ApplicationEvent;
-import org.springframework.context.ApplicationListener;
-import org.springframework.context.event.ContextRefreshedEvent;
```
tbarsballe: There seem to be a disproportionate number of import changes compared to the actual code change. Please try to limit changes to functional changes only - any reformatting should be done as a separate commit.

vickdw:
Made the changes you suggested, but it must be the IDE that reordered some of the imports, I have no other explanation for that.  Otherwise unused code and unneeded comments have been pulled out.

tbarsballe:
Ok, code looks good, CI build passes.

With respect to the commit message, you don't need / shouldn't have the full URL to the JIRA ticket / PR, just the ticket id. e.g.

> [GEOS-8171] adding a generic method to clear caches for external graphic factories

tbarsballe:
Manually merged: https://github.com/geoserver/geoserver/commit/47b20aca44032aeb599c175055f84cd35ed830f0

One other note - it is generally a bad idea to make a PR from your master branch, as a master branch tends to be subject to change. Create a feature branch instead.

PR #2387, opened 2017-06-01
[GEOS-8166] Allow image mosaic to refer a GeoServer configured store

aaime:
This pull request passes a Repository reference to readers, allowing them to refer to GeoServer datastore/dataaccess.

Working on it I noticed code duplication and oddities in the way readers are created and cached in the resource pool, cleaning it up and ended with one significant change to ResourcePool (one public method removed), but eliminating in the process the need to have two separate coverage reader caches and making sure readers get the hints they need no matter what the context is. 

ResourcePool is not meant to be used by "client" code, and its actual usage is limited to catalog classes and configuration related bits, such as REST config and Web UI, so I'm not too much concerned about the removal of that one public method to get the non hinted coverage cache reference (nothing was using it as far as I could tell, a full build passed locally, although I did a few cleanups after it).

That said, given the nature of the change, it's probably best if either @jodygarnett or @tbarsballe has a look at the pull request (or any PSC member, really).

### DIFF
--- a/doc/en/user/source/data/raster/imagemosaic/configuration.rst
+++ b/doc/en/user/source/data/raster/imagemosaic/configuration.rst
@@ -119,6 +119,18 @@ If needed, different storage can be used for the index — like a spatial DBMS, 
    * - Parameter
      - Mandatory?
      - Description
+   * - StoreName
+     - N
+     - Can be used to refer to a GeoServer registered store, using a "workspace:storeName" syntax. When this is used,
+       the no other connection parameters need to be provided. The SPI can still be provided to inform the mosaic of
+       the resulting type of store (e.g., Oracle) in case specific behavior need to be enacted for it (e.g., in the
+       case of Oracle the attributes are all uppercase and cannot be longer than 30 chars, the mosaic will respect
+       the limits but the `SPI` parameter needs to be explicitly set to `org.geotools.data.oracle.OracleNGDataStoreFactory`
+       as the actual store type is hidden when it reaches the mosaic code). 
+       Also, as a reminder, the code is picking up a Store reference, not a layer one, meaning that security restrictions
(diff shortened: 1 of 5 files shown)


src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -1461,42 +1442,26 @@ private GridCoverageReader getGridCoverageReader(CoverageStoreInfo info, Coverag
...
-                }
-            }
-            
-            key = new CoverageHintReaderKey(info.getId(), hints);
-            reader = hintCoverageReaderCache.get( key );
+        // expand the hints if necessary
+        if (hints != null) {
+            // do not modify the caller hints
+            hints = new Hints(hints);
+            hints.add(new RenderingHints(Hints.REPOSITORY, repository));
```
jodygarnett: suggestion only: you could take this out of the conditional and remove line 1451 below.

doc/en/user/source/data/raster/imagemosaic/configuration.rst
```diff
@@ -119,6 +119,18 @@ If needed, different storage can be used for the index — like a spatial DBMS,
    * - Parameter
      - Mandatory?
      - Description
+   * - StoreName
+     - N
+     - Can be used to refer to a GeoServer registered store, using a "workspace:storeName" syntax. When this is used,
```
aaime: Indeed

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
aaime: Not really, the 1451 creates a new hints object in case it's null. So I would have to leave it there, but replace it with the creation of empty hints. Line count is the same, but I guess having the add out of the conditional makes it clearer that we are going to add a repository anyways.

PR #2388, opened 2017-06-01
Initial commit of s3-geotiff GeoServer module

dvntucker:
This PR creates a community module that uses the corresponding GeoTools unsupported module, as well as documentation for it.


tbarsballe:
See also: https://github.com/geotools/geotools/pull/1610

PR #2389, opened 2017-06-03
Documentation for changes in GEOT-5746, GEOT-5747

aaime:


doc/en/user/source/styling/css/filters.rst
```diff
@@ -152,9 +152,19 @@ The context details that are provided are as follows:
...
+    * - @sd
       - The scale denominator for the current rendering.  More explicitly, this
-        is the ratio of real-world distance to screen/rendered distance.
+        is the ratio of real-world distance to screen/rendered distance. 
+    * - @scale
+      - Same as above, the scale denominator (not scale) for the current rendering. 
+        Supported for backwards compatibility 
+
+The scale value can be expressed as a plain number, for for brevity and readability,
+usage of prefixes such as k (kilo), M (mega), G (giga), for example::
```
tbarsballe: While most often used as prefixes, in this case these characters are coming at the end of the number, so this should be "usage of suffixes"

Looking at https://github.com/geotools/geotools/pull/1615, these three are the only ones supported, so "such as" should not be used, as it implies there may be other suffixes.

How about:

> The scale value can be expressed as a plain number. For brevity and readability the suffixes k (kilo), M (mega), or G (giga) can be used, for example::

aaime:
Was manually merged

PR #2405, opened 2017-06-06
[GEOS-8170] Allow CatalogBuilder to setup a new CoverageInfo even if the test read return null

aaime:


src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java
```diff
@@ -1077,13 +1035,88 @@ private CoverageInfo buildCoverageInternal(GridCoverage2DReader reader, String n
...
+            }
+            final int numBands = sm.getNumBands();
+            sampleDimensions = new GridSampleDimension[numBands];
+            // setting bands names.
+            for (int i = 0; i < numBands; i++) {
+                final ColorInterpretation colorInterpretation = TypeMap.getColorInterpretation(cm, i);
+                if (colorInterpretation == null)
+                    throw new IOException("Unrecognized sample dimension type for band number " + (i + 1));
+                sampleDimensions[i] = new GridSampleDimension(colorInterpretation.name());
+            }
```
dromagnoli: some "if(condition)" bad formatting and an "if(colorInterpretation == null) body " without enclosing bracket.

src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java
```diff
@@ -1077,13 +1035,88 @@ private CoverageInfo buildCoverageInternal(GridCoverage2DReader reader, String n
...
+        }
+
+        parameters.put(AbstractGridFormat.READ_GRIDGEOMETRY2D.getName().toString(), new GridGeometry2D(testRange, testEnvelope));
+
+        // try to read this coverage
+        final GridCoverage2D gc = reader.read(CoverageUtils.getParameters(readParams, parameters, true));
+        final GridSampleDimension[] sampleDimensions;
+        if (gc != null) {
+            // remove read grid geometry since it is request specific
+            parameters.remove(AbstractGridFormat.READ_GRIDGEOMETRY2D.getName().toString());
```
dromagnoli: This was already part of the previous code but I'm wondering why it is needed.

aaime: Indeed, makes no sense, parameters map is built in the method and does not get used outside.

aaime:
@dromagnoli feedback addressed

aaime:
Manually merged

PR #2412, opened 2017-06-08
[GEOS-8176] RetypingFeatureCollection doesn't retype subCollections

ianturton:
Fix for [GEOS-8176](https://osgeo-org.atlassian.net/browse/GEOS-8176) which forces the subCollections returned by the RetypeingFeatureCollection to be retyped.


src/main/src/main/java/org/geoserver/feature/RetypingFeatureCollection.java
```diff
@@ -56,6 +59,13 @@ protected boolean canDelegate(FeatureVisitor visitor) {
         return ReTypingFeatureCollection.isTypeCompatible(visitor, target);
     }
 
+    @Override
+    public SimpleFeatureCollection subCollection(Filter filter) {
+        
+        SimpleFeatureCollection delegateCollection = delegate.subCollection(filter);
+        
+        return new ReTypingFeatureCollection(delegateCollection, target);
```
aaime: This gave me pause for a moment, but retype can really only shave off attributes, so ti should be ok I guess.

PR #2415, opened 2017-06-12
[GEOS-8216] Fix MenuPageInfo Generics

NielsCharlier:
It is impossible to create a MenuPageInfo programmatically, because of a wrong usage of generics. 
It only works through Spring, because spring ignores generics.

MenuPageInfo has an associated class, which should be a subclass of GeoServerBasePage.

However. MenuPageInfo extends ComponentInfo<GeoServerBasePage>, rather than ComponentInfo<?extends GeoServerBasePage> that means that the associated class   MUST be GeoServerBasePage.class itself, which doesn't make sense.

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/MenuPageInfo.java
+++ b/src/web/core/src/main/java/org/geoserver/web/MenuPageInfo.java
@@ -22,7 +22,7 @@ import org.geoserver.web.services.ServiceMenuPageInfo;
  * @author David Winslow <dwinslow@opengeo.org>
  */
 @SuppressWarnings("serial")
-public class MenuPageInfo extends ComponentInfo<GeoServerBasePage> implements Comparable<MenuPageInfo> {
+public class MenuPageInfo<T extends GeoServerBasePage> extends ComponentInfo<T> implements Comparable<MenuPageInfo<T>> {
     Category category;
     int order;
     String icon;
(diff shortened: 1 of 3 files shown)


src/web/core/src/main/java/org/geoserver/web/MenuPageInfo.java
```diff
@@ -22,7 +22,7 @@
  * @author David Winslow <dwinslow@opengeo.org>
  */
 @SuppressWarnings("serial")
-public class MenuPageInfo extends ComponentInfo<GeoServerBasePage> implements Comparable<MenuPageInfo> {
+public class MenuPageInfo<T extends GeoServerBasePage> extends ComponentInfo<T> implements Comparable<MenuPageInfo> {
```
tbarsballe: ServiceMenuPageInfo below ` extends MenuPageInfo<T>`. Should we also be using `implements  Comparable<MenuPageInfo<T>>` here?

NielsCharlier: Well spotted, you are right.

NielsCharlier: Generics, eh.

NielsCharlier:
https://osgeo-org.atlassian.net/browse/GEOS-8216

PR #2424, opened 2017-06-15
[GEOS-7979] Add buffer to vector tiles so wide symbols are not clipped at tile boundaries

smithkm:
Mimics the behaviour of StreamingRenderer by using MetaTileBufferEstimator to estimate the size of the buffer from the style if it is not provided by the `buffer` parameter.

Addresses one of the points of GEOS-7979


smithkm:
Travis failure appears to be a case of Travis itself killing the build.  Possibly due to memory usage or run time. The same build command works locally.  I am re-running now.

DBlasby:
looks good!

PR #2441, opened 2017-06-28
Add instructions for installing LaTeX (and dependencies) to the docguide

tbarsballe:
While working on the new build.geoserver.org build server, it was discovered that installing the necessary dependencies for LaTeX is nontrivial for CentOS.

In order to document the steps taken to get the LaTeX doc build working, I have added a new section to the docguide detailing  how to install LaTeX, and any other required extensions, for a selection of OSes.

@bmmpxf - would you be able to look over the Windows section, and verify that matches your experience.

If anyone else has tried building the docs on linux, I would appreciate their review as well.

### DIFF
--- a/doc/en/docguide/source/index.rst
+++ b/doc/en/docguide/source/index.rst
@@ -10,6 +10,7 @@ This guide covers creating and maintaining the User Manual and Developer Manual.
    quickfix
    contributing
    install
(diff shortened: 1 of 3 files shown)


jodygarnett:
I think I used https://www.tug.org/mactex/

PR #2442, opened 2017-06-29
GEOS-8177 support monitoring of WCS 2.x request

jdries:
Dear all, I'm trying to add support for WCS 2.x GetCoverage requests to the monitoring extension.
However, parsing WCS 2.x requests is fairly complex, so I'd like to reuse the parsing that already exists in the WCS implementation itself.
To do this without requiring a large refactoring, I implemented the monitoring callback in the operationExecuted method. As a result, my WCS 2.0 handler deviates from the existing ones.
I also had to add an extension on the gs-wcs2_0 module, because I'm checking against the 'GranuleStack' interface. Moving this interface into a more general module could be a way to avoid this dependency. 

What do you think?

### DIFF
--- a/src/extension/monitor/core/pom.xml
+++ b/src/extension/monitor/core/pom.xml
@@ -70,6 +70,10 @@
     <artifactId>hamcrest-library</artifactId>
     <scope>test</scope>
   </dependency>
+     <dependency>
+         <groupId>org.geoserver</groupId>
+         <artifactId>gs-wcs2_0</artifactId>
+     </dependency>
  </dependencies>
 
  <build>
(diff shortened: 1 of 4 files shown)


aaime:
I believe the approach is problematic, but not because of the dependency per se, more because of how it's implemented. If one makes a GetCoverage with no bounding box, the code will still log the bbox of the returned coverage, which is out of line compared with how the other requests are managed (the bbox is the requested one, not the returned one).

tbarsballe:
@jdries Any update based on the provided feedback? 
Some issues were identified with the approach, so the PR is not mergeable as-is, and there haven't been any changes in several months. If the feedback is not addressed, then this PR will be closed.

bencaradocdavies:
Please revise and resubmit. Note that we are reformatting the code base this weekend.

PR #2443, opened 2017-06-29
GSIP 160 - Enhancements to KML placemark placement (GEOS-8033)

jdeolive:


src/kml/src/main/java/org/geoserver/kml/utils/KmlCentroidBuilder.java
```diff
@@ -26,6 +36,38 @@
...
+    public Coordinate geometryCentroid(Geometry g, Envelope bbox, KmlCentroidOptions opts) {
+        if (opts == null) {
+            opts = KmlCentroidOptions.DEFAULT;
+        }
+
+        // clip?
+        if (opts.isClip()) {
+            if (bbox != null) {
+                g = clipGeometry(g, bbox);
+            }
```
aaime: Formatting :-)

src/kml/src/main/java/org/geoserver/kml/utils/KmlCentroidBuilder.java
```diff
@@ -40,10 +82,21 @@ public Coordinate geometryCentroid(Geometry g) {
...
-        } else {
-            // return the actual centroid
-            return g.getCentroid().getCoordinate();
-        }
+        } else if (g instanceof Polygon) {
+            if (opts.isContain()) {
+                try {
+                    Point p = RendererUtilities.sampleForInternalPoint(
+                        (Polygon)g, null, null, null, -1, opts.getSamples());
+                    if (p != null && !p.isEmpty()) return p.getCoordinate();
```
aaime: Formatting :-)

jdeolive:
Thanks for the review @aaime. I'm a little bit out of the loop with regards to formatting... not quite sure what rules I am breaking here. Based on what I could find in the docs:

- Avoid the breaking the lines and just have an entire function call, nested calls, etc... one one line?
- Avoid single line if statements

?

src/kml/src/main/java/org/geoserver/kml/utils/KmlCentroidBuilder.java
aaime: Else should be on the same line as closing curly brace. Our conventions are the Java ones minus a few amendments. See: http://www.oracle.com/technetwork/java/javase/documentation/codeconventions-142311.html

src/kml/src/main/java/org/geoserver/kml/utils/KmlCentroidBuilder.java
aaime: Same link as above, avoid single line ifs.

jdeolive: Thanks @aaime , that clears it up. Will make those changes.

PR #2444, opened 2017-06-30
[GEOS-8203] Fixed deleting a workspace when using JDBC Configuration returning an error response.

sikeoka:
https://osgeo-org.atlassian.net/browse/GEOS-8203

Updated to catch EmptyResultDataAccessException which is what gets thrown by Spring instead of returning null.  Added unit test and also added a missing @Test annotation to another unit test.

Can be backported to 2.11.x an 2.10.x.


aaime:
Looks good to me but I'm not super-familiar with the jdbcconfig (not enough to just review at a glance at least). @tbarsballe are you in charge of that module, or do you know who is?

tbarsballe:
I am not aware of who is in charge of this module (I don't think it has an active maintainer). Also not familiar enough with it to confidently review it.
Looking at recent commit history, maybe @NielsCharlier or @smithkm ?

NielsCharlier:
I'm not officially in charge but I do know the module and I approve it. The problem was well analysed, the solution makes sense and there is a good unit test.

PR #2445, opened 2017-06-30
[GEOS-8134] Move style file when moving style from workspace to global

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-8134

This turned out to be a bug in the GeoServerPersister, and was exposed by https://github.com/geoserver/geoserver/pull/2162

It turns out, the persister didn't handle moving style files to the global 'workspace' from a workspace, only moving styles to an actual workspace.

### DIFF
--- a/src/main/src/main/java/org/geoserver/config/GeoServerPersister.java
+++ b/src/main/src/main/java/org/geoserver/config/GeoServerPersister.java
@@ -171,7 +171,12 @@ public class GeoServerPersister implements CatalogListener, ConfigurationListene
                 i = event.getPropertyNames().indexOf("workspace");
                 if (i > -1) {
                     WorkspaceInfo newWorkspace = (WorkspaceInfo) event.getNewValues().get( i );
-                    Resource newDir = dd.getStyles(newWorkspace);
+                    Resource newDir;
+                    if (newWorkspace == null) {
+                        newDir = dd.getStyles();
+                    } else {
+                        newDir = dd.getStyles(newWorkspace);
+                    }
 
                     // look for any resource files (image, etc...) and copy them over, don't move 
                     // since they could be shared among other styles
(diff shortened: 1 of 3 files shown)


aaime:
Added @nmco as a reviewer, since he was involved in #2162

nmco:
I was a bit perplex that ``dd.getStyles()`` and ``dd.getStyles(newWorkspace)`` return a different result when the workspace is ``NULL``. I would expect ``dd.getStyles(newWorkspace)`` to throw an exception or to return the same result as ``dd.getStyles()``.

I debug a bit the code a this is were the the ``NULL`` workspace case is handled:
https://github.com/geoserver/geoserver/blob/master/src/main/src/main/java/org/geoserver/config/GeoServerDataDirectory.java#L819-L828

So when the workspace is ``NULL`` this is what is performed:
```
if (ws == null) {
    return resourceLoader.get("");
}
```

This will return the root of GeoServer data directory, this is a bit dangerous and inconsistent no ?

I would change this method to actually ignore the workspace if it is ``NULL``, this should fix the current issue and maybe other similar issue we didn't discover yet.

nmco:
Apart from my comment above that suggest an alternative fix the PR looks good with good tests coverage (for the WEB UI and for the persister).

nmco:
I have also manually tested the fix and it seems to be working (even in a JMS cluster).

tbarsballe:
Yes, I was surprised about that behaviour as well. This affects more than just dd.getStyles(ws); dd.getResource(ws) has similar behaviour.

I [brought it up on the mailing list](http://osgeo-org.1560.x6.nabble.com/Odd-GeoServerDataDirectory-behaviour-tt5326295.html), but didn't get much of a response...

Changing that GeoServerDataDirectory behaviour would be a better fix I think, but I am a bit concerned that it might have unintended side effects - I would like to see a bit more input from others before going ahead with it; namely is the current behaviour deliberate or a legitimate bug in the GeoServerDataDirectory implementation.

nmco:
I comment on your ML thread.

tbarsballe:
Superseded by #2460

PR #2447, opened 2017-06-30
[GEOS-8062] Include links to nested LayerGroups in REST LayerGroup representation

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-8062

### DIFF
--- a/src/restconfig/src/main/java/org/geoserver/rest/catalog/LayerGroupController.java
+++ b/src/restconfig/src/main/java/org/geoserver/rest/catalog/LayerGroupController.java
@@ -229,6 +229,13 @@ public class LayerGroupController extends AbstractCatalogController {
                 }
                 if ( obj instanceof LayerInfo ) {
                     converter.encodeLink("/layers/" + converter.encode(ref), writer);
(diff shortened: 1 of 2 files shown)


src/restconfig/src/main/java/org/geoserver/rest/catalog/LayerGroupController.java
```diff
@@ -229,6 +229,8 @@ protected void postEncodeReference(Object obj, String ref, String prefix,
                 }
                 if ( obj instanceof LayerInfo ) {
                     converter.encodeLink("/layers/" + converter.encode(ref), writer);
+                } else if ( obj instanceof LayerGroupInfo) {
+                    converter.encodeLink("/layergroups/" + converter.encode(ref), writer);
```
aaime: What happens if the group is workspace specific?

tbarsballe: Good question - I'd forgotten that layer groups behave differently from layers in this case, so that currently is not handled. Will fix.

PR #2449, opened 2017-06-30
[GEOS-8010] dbtype and filetype store arguments should not be visible

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-8010

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/data/store/DefaultDataStoreEditPanel.java
+++ b/src/web/core/src/main/java/org/geoserver/web/data/store/DefaultDataStoreEditPanel.java
@@ -160,13 +160,15 @@ public class DefaultDataStoreEditPanel extends StoreEditPanel {
         final Class<?> binding = paramMetadata.getBinding();
         final List<Serializable> options = paramMetadata.getOptions();
 
+        boolean visible = true;
+
         Panel parameterPanel;
         if("dbtype".equals(paramName) || "filetype".equals(paramName)) {
             // skip the two well known discriminators
             IModel model = new MapModel(paramsModel, paramName);
             TextParamPanel tp = new TextParamPanel(componentId,
                     model, new ResourceModel(paramLabel, paramLabel), required);
-            tp.setVisible(false);
+            visible = false;
             parameterPanel = tp;
         } else  if ("namespace".equals(paramName)) {
             IModel namespaceModel = new NamespaceParamModel(paramsModel, paramName);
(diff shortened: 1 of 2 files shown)


src/web/core/src/main/java/org/geoserver/web/data/store/DefaultDataStoreEditPanel.java
```diff
@@ -160,13 +160,15 @@ private Panel getInputComponent(final String componentId, final IModel paramsMod
         final Class<?> binding = paramMetadata.getBinding();
         final List<Serializable> options = paramMetadata.getOptions();
 
+        boolean visible = true;
+
         Panel parameterPanel;
         if("dbtype".equals(paramName) || "filetype".equals(paramName)) {
```
jodygarnett: can you check for level "program" here, rather than do a bunch of custom logic? it is a more scalable approach using the geotools api

aaime: That also requires changing the GeoTools datastore factory parameters, as they are not using said flag:
public static final Param DBTYPE = new Param("dbtype", String.class, "Type", true);
public static final Param FILE_TYPE = new Param("filetype", String.class,
            "Discriminator for directory stores", false, "shapefile", new KVP(Param.LEVEL,
                    "advanced"));

tbarsballe: Given that adding that flag to the parameters in GeoTools could affect the behaviour of arbitrary downstream applications, that would count as an API change, correct?

aaime: Hum... strictly it would, but I doubt anything but uDig ever used those params.

jodygarnett: And our gt-swing module :)

aaime:
This one got stuck, I recommend merging.

src/web/core/src/main/java/org/geoserver/web/data/store/DefaultDataStoreEditPanel.java
tbarsballe: I've attempted an approach using the GeoTools parameters, See https://github.com/geotools/geotools/pull/1659 and https://github.com/geoserver/geoserver/pull/2472

tbarsballe:
Closing in favour of https://github.com/geoserver/geoserver/pull/2472

PR #2450, opened 2017-07-02
[GEOS-8108] Several WFS requests on complex features are returning null namespace prefixes

nmco:
Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-8108

A test for app-schema that tests features chaining with features types from multiple namespaces and for both WFS 1.1 \ GML 3.1 and WFS 2.0 \ GML 3.2 was added.


src/wfs/src/main/java/org/geoserver/wfs/xml/ApplicationSchemaXSD1.java
```diff
@@ -146,6 +147,8 @@ private static XSDSchema importWfsSchema(XSDSchema schema) throws IOException {
         synchronized (wfsSchema.eAdapters()) {
             ((XSDSchemaImpl) wfsSchema).imported(wfsImport);
         }
+        // make sure that GML 3.2 namespace is used
```
aaime: Given the import, is this really 3.2?

src/wfs/src/main/java/org/geoserver/wfs/xml/FeatureTypeSchemaBuilder.java
```diff
@@ -209,12 +211,23 @@ public final XSDSchema buildSchemaInternal(FeatureTypeInfo[] featureTypeInfos, S
             }
 
             if (!simple) {
-                // add secondary namespaces from catalog
-                for (NamespaceInfo nameSpaceinfo : catalog.getNamespaces()) {
-                    if (!schema.getQNamePrefixToNamespaceMap().containsKey(nameSpaceinfo.getPrefix())) {
-                        schema.getQNamePrefixToNamespaceMap().put(nameSpaceinfo.getPrefix(),
-                                nameSpaceinfo.getURI());
+                // complex features may belong to different workspaces
+                WorkspaceInfo localWorkspace = LocalWorkspace.get();
```
aaime: Aaah, tricky... but effective indeed, cannot think of a better way.

src/wfs/src/main/java/org/geoserver/wfs/xml/ApplicationSchemaXSD1.java
nmco: Nope is the 3.1 the comment is wrong :( ... fixing it.

src/wfs/src/main/java/org/geoserver/wfs/xml/FeatureTypeSchemaBuilder.java
nmco: Yeah, in this case we are only liking the namespace so security wise it is not a problem.

PR #2456, opened 2017-07-06
[GEOS-8212]: minMax range excluding nodata category value

dromagnoli:


dromagnoli:
This PR depends on:
https://github.com/geotools/geotools/pull/1643

PR #2457, opened 2017-07-07
[GEOS-8212]: Do not report nodata value in CoverageDimensionInfo min/…

dromagnoli:
…Max range (use dedicate nullValues field)
Backport to 2.11.x


dromagnoli:
Depends on https://github.com/geotools/geotools/pull/1644

PR #2458, opened 2017-07-10
Input Output Pipe for CSV for WPS

ianturton:
Allows WPS processes to import and export CSV files.

Need to check if there is a way to prevent CSV datastore showing up in Stores or fix that store to have a frontend.


ianturton:
replaced by latter cleaner PR

PR #2459, opened 2017-07-10
[GEOS-8212]: Do not report nodata value in CoverageDimensionInfo min/…

dromagnoli:
…Max range (use dedicate nullValues field) - bkp to 2.10.x


dromagnoli:
The travis build failed after 1 hour of running job with a timeout. (the 2nd in a row).
This morning I had the same time with the related GT PR (needed 4 travis restarts)
Merging it anyway.

PR #2466, opened 2017-07-18
[GEOS-8226] Get stores in the right workspace when listing resources via REST (2.11.x)

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-8226

This is the 2.11.x version of the fix, the issue does not appear to affect master.


aaime:
+1

PR #2469, opened 2017-07-24
GEOS-7989 -part2: wms/wcs custom dimensions with Number

dromagnoli:


src/main/src/main/java/org/geoserver/catalog/util/ReaderDimensionsAccessor.java
```diff
@@ -565,6 +565,8 @@ public boolean hasResolution(String domain) {
                 Class<?> type = Class.forName(typeName);
                 if (type == java.util.Date.class) {
                     result.addAll(new TimeParser().parse(value));
+                } else if ((type == java.lang.Double.class || type == java.lang.Float.class) && !value.contains(",")) {
```
aaime: What if it's an integer, or short, or ... whatever? Can it check Number.class.isAssignableFrom(...) instead?

src/main/src/test/java/org/geoserver/catalog/impl/ReaderDimensionAccessorTest.java
```diff
@@ -119,4 +119,18 @@ public void testCustomDimensionConvertion() throws IOException, ParseException {
...
+    @Test
+    public void testCustomDepthDimensionConvertion() throws IOException, ParseException {
+        MockDimensionReader reader = new MockDimensionReader();
+        reader.metadata.put("HAS_MYDIM_DOMAIN", "true");
+        reader.metadata.put("MYDIM_DOMAIN_DATATYPE", "java.lang.Double");
+        ReaderDimensionsAccessor accessor = new ReaderDimensionsAccessor(reader);
+        List<Object> converted = accessor.convertDimensionValue("MYDIM",
+                Arrays.asList("10/20"));
+        assertEquals(1, converted.size());
+        NumberRange<Double> range = new NumberRange<Double>(Double.class, 10d, 20d);
```
aaime: "expected" instead of range?

src/main/src/main/java/org/geoserver/catalog/util/ReaderDimensionsAccessor.java
dromagnoli: Right, I was thinking about something similar indeed... Fixing it.

src/main/src/test/java/org/geoserver/catalog/impl/ReaderDimensionAccessorTest.java
dromagnoli: Fixing it.

dromagnoli:
@aaime, Thanks for the feedbacks. I have updated the code based on your suggestions.

PR #2472, opened 2017-07-28
[GEOS-8010] dbtype and filetype store arguments should not be visible

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-8010

Depends upon https://github.com/geotools/geotools/pull/1659. Travis CI will fail until that is merged.

Supersedes https://github.com/geoserver/geoserver/pull/2449; only one of these PRs should be merged.

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/data/store/ParamInfo.java
+++ b/src/web/core/src/main/java/org/geoserver/web/data/store/ParamInfo.java
@@ -34,6 +34,8 @@ public class ParamInfo implements Serializable {
     
     private boolean largeText;
 
+    private String level;
+
     private Class<?> binding;
 
     private boolean required;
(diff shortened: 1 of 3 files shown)


src/web/core/src/test/java/org/geoserver/web/data/store/DataAccessNewPageDatabaseTest.java
```diff
@@ -0,0 +1,48 @@
+package org.geoserver.web.data.store;
```
jodygarnett: missing header

src/web/core/src/main/java/org/geoserver/web/data/store/ParamInfo.java
```diff
@@ -105,7 +108,11 @@ public String getTitle() {
     public boolean isPassword() {
         return password;
     }
-    
+
+    public String getLevel() {
```
jodygarnett: thanks, working around ParamInfo is frustrating, do we need javadocs for these public methods (say to define the levels).

tbarsballe: ParamInfo more or less directly mirrors Param / Parameter, and is only used by the UI code.
Javadocs defining the level are accessible from the Parameter javadocs for the equivalent methods.
Ultimately, the information already exists in a more useful location, although we could maybe reference those classes?

src/web/core/src/test/java/org/geoserver/web/data/store/DataAccessNewPageDatabaseTest.java
tbarsballe: Indeed. Fixed.

tbarsballe:
@jodygarnett - Are you happy with the changes here? Do you want to mark this as approved.

PR #2475, opened 2017-07-28
[GEOS-8178] check if necessary to prepend EPSG to CRS identifier for OL output

mtCarto:


tbarsballe:
@mtCarto - are you able to add a test case? 

It would be nice to get this in before the 2.12-beta.

PR #2478, opened 2017-07-31
update with query parameter documetation

jodygarnett:
noticed on user list that recalculate options are not documented.

Migrated parameter documentation from here http://docs.geoserver.org/stable/en/user/rest/api/featuretypes.html#rest-api-featuretypes-recalculate


doc/en/api/1.0.0/featuretypes.yaml
```diff
@@ -477,18 +477,37 @@ paths:
...
...   description: Specifies whether to recalculate any bounding boxes for a feature type. Some properties of feature types are automatically recalculated when necessary. In particular, the native bounding box is recalculated when the projection or projection policy are changed, and the lat/long bounding box is recalculated when the native bounding box is recalculated, or when a new native bounding box is explicitly provided in the request. (The native and lat/long bounding boxes are not automatically recalculated when they are explicitly included in the request.) In addition, the client may explicitly request a fixed set of fields to calculate, by including a comma-separated list of their names in the recalculate parameter.  For example 'recalculate=' (empty parameter) does not calculate any fields, regardless of any changes to projection, projection policy, etc. This is useful avoid slow recalculation when operating against large datasets. For example 'recalculate=nativebbox' recalculates the native bounding box, but does not recalculate the lat/long bounding box. For example 'recalculate=nativebbox,latlonbbox' recalculates both the native bounding box and the lat/long bounding box.
```
bmmpxf: The triple "For example" is a little awkward here, but not a showstopper.

PR #2479, opened 2017-08-01
Add required namespaces to SLD InlineFeature example

tbarsballe:
I just tried out the InlineFeature example at http://docs.geoserver.org/latest/en/user/styling/sld/reference/layers.html#example using a WMS request with the SLD parameter.

GeoServer gave an error response with SAXParseExceptions indicating that the "gml" and "ogc" namespaces were not defined in the SLD.

After adding these namespaces the SLD rendered fine.

I have updated the doc example accordingly.


aaime:
+1, merged

PR #2489, opened 2017-08-22
GSIP-159 Add SecurityDispatcher to GWC context

smithkm:
Updates to not break when https://github.com/GeoWebCache/geowebcache/pull/526 is applied upstream

This is not the complete GSIP implementation.


src/gwc/src/main/java/org/geoserver/gwc/dispatch/GeoServerSecurityFilter.java
```diff
@@ -0,0 +1,41 @@
+package org.geoserver.gwc.dispatch;
```
jodygarnett: header

src/gwc/src/main/resources/geowebcache-geoserver-context.xml
```diff
@@ -91,6 +91,11 @@
     <description>
     </description>
   </bean>
+  
+  <bean id="GWCSecurityFilter" class="org.geoserver.gwc.dispatch.GeoServerSecurityFilter">
+    <description>
```
jodygarnett: this descriptions is strangely unsatisfying

PR #2490, opened 2017-08-23
[GSIP-161] - Style Groups

tbarsballe:
See https://github.com/geoserver/geoserver/wiki/GSIP-161 for reference.

**Summary of changes** (27 core files, 33 test files, and 8 doc files):

* Created a `SLDVisitor.java` class based off of the logic in `GetMapKvpRequestReader.processStandaloneSLD()`, intended to provide a more useful, extensible framework for dealing with standalone sld documents (such as style groups).
* Added support for reading and writing `StyledLayerDescriptor` objects from styles to ResourcePool and GeoServerDataDirectory (formerly, only `Style` objects were supported).
* Added logic to interpret a layer group containing a null layer and a not-null style as a style group / standalone SLD document and handle it accordingly:
  * `LayerGroupHelper` changes to support retrieving the full list of layers and styles, calculating bounds, and checking for recursive loops (This also made WMS support style groups)
  * Rest API and UI support
  * Misc catalog, serialization, and deserialization support.
* Added documentation describing style groups
* Added test cases:
  * Unit tests for the new classes
  * Various integration tests: Catalog, WMS, UI, and Rest API tests.
  * Also updated a number of test cases to add the right number of styles to any layer groups they create. For a portion of this branches lifetime, this was necessary for the tests to pass, although it isn't any more. However, no point in reverting the change, as it makes the test data "more correct". (previously, most tests just didn't add any styles to the layer groups, regardless of the number of layers).

Note: The first bullet under item 5 of the proposal section did not get completed, and is not included. This was due to a few factors, mainly time and some unforeseen complications:
 * For the REST API portion, the REST API does not really handle renames in a consistent fashion - They are not supported for layer groups, sort of supported for layers, and seem to be supported for styles. This needs to be looked at and made consistent before it makes sense to add a consistent set of validation options across these endpoints.
 * For the UI portion, implementing a dialog menu that pops up during a submit and affects the outcome of that submit will be complicated and time consuming, due to Wicket.
A SLDNamedLayerRenameHelper was added to facilitate this functionality in the future.

Validation when saving a style group, and when saving a layer group containing a style group,*is* in place.


src/main/src/main/java/org/geoserver/catalog/SLDVisitor.java
```diff
@@ -0,0 +1,383 @@
...
+ * Visitor for standalone {@link org.geotools.styling.StyledLayerDescriptor}s
+ *
+ * Visits each {@link StyledLayer}s of an SLD.
+ * Resolves LayerGroups, Remote OWS references, and inline features into individual layers
+ * Visits each {@link Style} in each{@link StyledLayer}, alongside each of these resolved layers
+ *
+ * Intended to provide a definitive, extensible approach to parsing standalone
+ * {@link org.geotools.styling.StyledLayerDescriptor}s, including style groups and external SLDs.
+ */
+public abstract class SLDVisitor {
```
jodygarnett: Please use geotools StyleVisitor rather than introduce a new abstract class (http://docs.geotools.org/stable/javadocs/org/opengis/style/StyleVisitor.html).

src/main/src/main/java/org/geoserver/catalog/SLDVisitor.java
```diff
@@ -0,0 +1,383 @@
...
+     * Construct temporary stores for inline features and remote OWS layers.
+     * After visiting a layer, visit each style in that layer. In cases where the styled layer references multiple
+     * layers, such as a named layer referencing a layer group, or remote OWS user layer exposing multiple layers, visit
+     * each style for each of these sublayers instead.
+     *
+     * @param sld The sld the visitor is applied to
+     * @throws ServiceException If the SLD document is invalid
+     */
+    public void apply(StyledLayerDescriptor sld) throws IOException {
+        final StyledLayer[] styledLayers = sld.getStyledLayers();
```
jodygarnett: This is where you can use GeoTools StyleVisitor?

src/main/src/main/java/org/geoserver/catalog/SLDVisitor.java
tbarsballe: That interface is never used anywhere in GeoTools, and has no implementations  (plus, its intended to be applied to `Style`s, not `StyledLayerDescriptor`s).

I assume you mean the similar, but different one here, and will use that one: http://docs.geotools.org/stable/javadocs/org/geotools/styling/StyleVisitor.html

jodygarnett: Yes, sorry the crew that was focused on opengis interfaces forked off.

tbarsballe: Updated to extend StyleVisitor

doc/en/user/source/styling/sld/working.rst
```diff
@@ -20,12 +20,17 @@ Catalog Styles
 Styles in the catalog can be viewed, edited and validated via the :ref:`styling_webadmin_styles` menu of the :ref:`web_admin`. 
 They may also be created and accessed via the REST :ref:`rest_api_styles` API.
 
-Catalog styles consist of a :ref:`sld_reference_sld` document 
+There are two types of Catalog Styles: Layer Styles (the default) and Style Groups.
+
+Layer Styles
+~~~~~~~~~~~~
```
jodygarnett: Consider rewriting to focus on symbology encoding document, or how an SLD can be used in this role (short listing only the first named layer lement).

> Symbology Encoding Style
> 
> Symbology Encoding style consists of an :file:`Symbology Encoding` document used to specify the styling of a single layer.
> 
> GeoServer supports the use of a :ref:`sld_reference_sld` document containing a single ``<NamedLayer>`` element, which contains a single ``<UserStyle>`` element to specify the styling. 
> 
> * When used in this fashion the layer name is ignored, since the style may be applied to many different layers.
> * When using an an :ref:`sld_reference_sld` generated by another application keep in mind only the first ``<NamedLayer>`` is used, any subsequent content is ignored.

doc/en/user/source/styling/sld/working.rst
```diff
@@ -35,7 +40,22 @@ using the :ref:`data_webadmin_layers` page of the :ref:`web_admin`.
 .. note:: When adding a layer and a style for it to GeoServer at the same time, the style should be added first, 
           so that the new layer can be associated with the style immediately. 
 
-          
+Style Groups
```
jodygarnett: Consider rewriting to make this section focus on style layer descriptor, used to define the operation of a wms.

> Style Layer Descriptor Style
> 
> A :ref:`sld_reference_sld` document containing any number of ``<NamedLayer>`` and ``<UserLayer>`` elements,   each of which may contain any number of ``<UserStyle>`` or ``<NamedStyle>`` elements.
>
> Within a style layer descriptor document the name of any ``<NamedLayer>`` elements should match a layer (or layer group) in the catalog. Likewise,  ``<NamedStyle>`` elements should refer to a style in the catalog.
>
> etc...

doc/en/user/source/styling/sld/working.rst
```diff
@@ -35,7 +40,22 @@ using the :ref:`data_webadmin_layers` page of the :ref:`web_admin`.
...
+
+Style groups can define new layers of styled data, by using the SLD :ref:`sld_reference_inlinefeature` element to provide feature data.
+
+Style Groups can be added to Layer Groups as an alternative way of defining a collection of styled layers, using either the :ref:`Web Administration interface <data_webadmin_layergroups>` or the :api:`REST API <layergroups.yaml>`.
+
+Style groups can still be assigned to layers and used like a layer style, in which case only the first ``<NamedLayer>`` will be used.
+Style groups can also be used as an External Style, via the geoserver styles endpoint (``/geoserver/styles``) or the geoserver REST api.
+
 External Styles
 ^^^^^^^^^^^^^^^
```
jodygarnett: Do you have any interaction with external styles to sort out?

tbarsballe: I'm not quite sure what you mean?

There is no direct interaction, but any valid external style could be added to the catalog as a style group, as described here.

doc/en/user/source/styling/sld/working.rst
tbarsballe: While that is definitely more correct, and clearer if you understand what it means, I am a bit concerned that it is more confusing from a user perspective.

doc/en/user/source/styling/sld/working.rst
tbarsballe: Same comment as above; not sure if this is clearer from a user perspective, especially since we refer to the styling language as Style Layer Descriptor.

tbarsballe:
Headers added.

src/main/src/main/java/org/geoserver/catalog/GeoServerSLDVisitor.java
```diff
@@ -0,0 +1,469 @@
...
+    protected final Catalog catalog;
+    protected final CoordinateReferenceSystem fallbackCrs;
+
+    //Current state
+    protected PublishedInfo info;
+    protected StyledLayer layer;
+    protected int styleCount = 0;
+
+
+    public GeoServerSLDVisitor(Catalog catalog, CoordinateReferenceSystem fallbackCrs) {
```
jodygarnett: javadoc for public constructor

(or the fields, what does fallbackCrs do?)

src/main/src/main/java/org/geoserver/catalog/GeoServerSLDVisitor.java
```diff
@@ -0,0 +1,469 @@
...
+     * @return The resolved catalog layer
+     */
+    public abstract PublishedInfo visitNamedLayerInternal(StyledLayer namedLayer);
+
+    /**
+     * Called on each remote OWS user layer in the style
+     *
+     * @param userLayer The user layer
+     */
+    public abstract void visitUserLayerRemoteOWS(UserLayer userLayer);
```
jodygarnett: Are you in position to provide a default method? Or do you want to force subclasses to think about it.

src/main/src/main/java/org/geoserver/catalog/GeoServerSLDVisitorAdapter.java
```diff
@@ -0,0 +1,41 @@
+package org.geoserver.catalog;
```
jodygarnett: header

src/main/src/main/java/org/geoserver/catalog/LayerGroupHelper.java
```diff
@@ -284,21 +321,82 @@ public String getLoopAsString(Stack<LayerGroupInfo> path) {
...
+                    }
                 }
             }
         }
         
         path.pop();
         return false;
     }
+
+    private static boolean checkStyleGroupLoops(StyleInfo styleGroup, LayerGroupInfo group, Stack<LayerGroupInfo> path) {
```
jodygarnett: looks like you are checking for recursion, can we have a java doc just to say what true and false mean?

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -207,6 +208,7 @@
     Map<String, WebMapServer> wmsCache;
     Map<String, WebMapTileServer> wmtsCache;
     Map<CoverageHintReaderKey, GridCoverageReader> hintCoverageReaderCache;
+    Map<StyleInfo,StyledLayerDescriptor> sldCache;
```
jodygarnett: So this is the more generic version of styleCache, can we implement styleCache in terms of sldCache and avoid duplication?

src/main/src/main/java/org/geoserver/catalog/SLDNamedLayerRenameHelper.java
```diff
@@ -0,0 +1,219 @@
+package org.geoserver.catalog;
```
jodygarnett: header

src/main/src/main/java/org/geoserver/catalog/SLDNamedLayerValidator.java
```diff
@@ -0,0 +1,61 @@
+package org.geoserver.catalog;
```
jodygarnett: header

src/main/src/main/java/org/geoserver/catalog/GeoServerSLDVisitor.java
tbarsballe: I've got an adapter subclass; I suppose I could have done that using default methods instead.

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
tbarsballe: Since getting a "Style" involves changing the name directly in the style object, we probably want to avoid it, although I am honestly not sure.

src/main/src/main/java/org/geoserver/catalog/LayerGroupHelper.java
tbarsballe: Sure. In that case I'll update `checkLoop` as well (the preexisting version of this)

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
tbarsballe: Come to think of it, I need to fix the current implementation so we don't do that.

If we were to delegate, we would need to do a bit of post-processing each time we get the style - get the first namedLayer, then change the name. No disk lookups, so that might be okay.
I'll change it for now to delegate, and we can test during the beta to see if anything weird happens.

tbarsballe: Ok, I tried out integrating them, and ran into some issues (bug in StyleImpl.clone() - normally i'd fix it, but with the code freeze this close I don't really want to be messing around in core GeoTools), so for now I am just going to fully separate styleCache and sldCache. styleCache is likely to be much more frequently used, so this shouldn't be much of an issue from a memory / efficiency perspective.

jodygarnett: Use duplicating style visitor, rather than clone()

src/main/src/main/java/org/geoserver/catalog/LayerGroupHelper.java
```diff
@@ -280,25 +322,101 @@ public String getLoopAsString(Stack<LayerGroupInfo> path) {
...
+     * Check if a style group contains the enclosing layer group, or other recursive structures.
+     *
+     * @param styleGroup The style group
+     * @param group The current layer group
+     * @param path Stack of each visited/parent LayerGroup
+     * @return true if the style group contains itself, or another LayerGroup contains itself
+     */
+    private static boolean checkStyleGroupLoops(StyleInfo styleGroup, LayerGroupInfo group, Stack<LayerGroupInfo> path) {
+        try {
+            Catalog catalog = getCatalogFromStyle(styleGroup);
```
jodygarnett: Do you need to look this up via the passed in StyleInfo to facilitate test cases? Or can you use GeoServerExtensions like some of the other utility classes:

```
return (Catalog) GeoServerExtensions.bean("catalog");
```

src/main/src/main/java/org/geoserver/catalog/LayerGroupHelper.java
```diff
@@ -280,25 +322,101 @@ public String getLoopAsString(Stack<LayerGroupInfo> path) {
...
+            StyledLayerDescriptor sld = new GeoServerDataDirectory(catalog.getResourceLoader()).parsedSld(styleGroup);
```
jodygarnett: If you are just using the catalog here to make a flyweight GeoServerDataDirectory wrapper around a resource loader .... can I suggest making a StyleInfo.getSLD() method to avoid all this violation of encapsulation (and make everybody more testable).

src/main/src/main/java/org/geoserver/catalog/LayerGroupHelper.java
```diff
@@ -309,4 +427,79 @@ private static boolean isGroupInStack(LayerGroupInfo group, Stack<LayerGroupInfo
...
+            StyleGroupHelper helper = new StyleGroupHelper(catalog, crs);
+            sld.accept(helper);
+            layers.addAll(helper.getLayers());
+            styles.addAll(helper.getStyles());
+        } catch (IllegalStateException | IOException | ServiceException | UncheckedIOException | UnsupportedOperationException e) {
+            LOGGER.log(Level.WARNING, "Error extracting styles from Style Group '" + styleGroup.getName() + "'. Skipping...", e);
+        }
+    }
+
+    protected static class StyleGroupHelper extends GeoServerSLDVisitorAdapter {
```
jodygarnett: want a javadoc if you are going to make a protected class (or lock it down to private until you get a subclass)

src/wms/src/test/resources/org/geoserver/wms/map/BasicStyleGroup.sld
```diff
@@ -0,0 +1,84 @@
...
+    <!-- FeatureTypeStyles describe how to render different features -->
+    <!-- a feature type for polygons -->
+    <FeatureTypeStyle>
+      <Rule>
+        <!-- like a linesymbolizer but with a fill too -->
+        <PolygonSymbolizer>
+		<Fill>
+			<CssParameter name="fill">
+				<ogc:Literal>#0000C0</ogc:Literal>
+			</CssParameter>
```
jodygarnett: tabs vs spaces, @johngschulz :)

src/main/src/main/java/org/geoserver/catalog/LayerGroupHelper.java
tbarsballe: I would much prefer modifying StyleInfo, so I will do that. Was initially a bit hesitant to make unplanned API changes to _Info classes.

src/wms/src/test/resources/org/geoserver/wms/map/BasicStyleGroup.sld
tbarsballe: Fixed.

PR #2492, opened 2017-08-23
Add packaging config for gs-mbstyle community module

tbarsballe:
Added assembly config file and dependency to community/release module.

Also added to the "communityRelease" profile.


tbarsballe:
Depends upon https://github.com/geotools/geotools/pull/1684

src/community/release/ext-mbstyle.xml
```diff
@@ -0,0 +1,18 @@
...
+  </formats>
+  <includeBaseDirectory>false</includeBaseDirectory>
+  <fileSets>
+    <fileSet>
+      <directory>release/target/dependency</directory>
+      <outputDirectory></outputDirectory>
+      <includes>
+        <include>gs-mbstyle-*.jar</include>
+        <include>gt-mbstyle-*.jar</include>
+        <include>hamcrest-all-*.jar</include>
```
jodygarnett: I am not used to seeing hamcrest included, that is usually for test cases

tbarsballe: Looks like this was incorrectly scoped as compile in gt-mbstyle: https://github.com/geotools/geotools/blob/master/modules/unsupported/mbstyle/pom.xml#L130-L134

I'll update here and in geotools...

PR #2493, opened 2017-08-23
Add troubleshooting information for installing geoserver on a Mac with an old version of JAI

tbarsballe:
This is stemming from some feedback we got at the code sprint at FOSS4G 2017.

Note: The line endings in the changed files have been converted to standard unix for sanity, only actual changes are adding `    .. include:: ./osx_jaierror.txt`.


jodygarnett:
If we include the JRE in the mac bundle we should be able to avoid the system classpath and extensions (and thus this problem).

doc/en/user/source/installation/osx_jaierror.txt
```diff
@@ -0,0 +1,15 @@
+.. warning:: If you encounter the following error during startup, you may have some invalid JAI jars from the default Mac Java install:
+
+    .. code-block:: bash
+
+        java.lang.NoClassDefFoundError: Could not initialize class javax.media.jai.JAI
+    
+    To fix this error, locate your Java installation folder, and delete the following jars:
```
jodygarnett: These can also be found in the ~/Library/Java/Extensions folder (and are automatically installed in this location by uDig).

jai-extensions

doc/en/user/source/installation/osx_installer.rst
```diff
@@ -1 +1,38 @@
...
       Starting GeoServer

#. The console window will be populated with log entries showing the GeoServer loading process. Once GeoServer is completely started, a browser window will open at ``http://localhost:8080/geoserver``, which is the :ref:`web_admin` for GeoServer.
\ No newline at end of file
+.. _installation_osx_installer:
+
+Mac OS X installer
+==================
+
+The Windows installer provides an easy way to set up GeoServer on your system, as it requires no configuration files to be edited or command line settings.
```
ianturton: Should that be the Mac installer not Windows?

tbarsballe: Indeed it should. Wonder how long that has been there.

tbarsballe:
Updated based on feedback.

PR #2494, opened 2017-08-24
[GEOS-8263] Allow in layer preview filtering (CQL, OGC and FeatureID) for rasters formats that support filtering

nmco:
Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-8263

The PR changes the layer preview behavior to activate filtering if a coverage layer reader supports filtering (e.g. image mosaic).

Testing this was bit tricky, depending on ImageMosaic for the tests was not a good idea so I created a static reader that always return a static image but more importantly allowed me to define the read parameters I need.

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/map/OpenLayersMapOutputFormat.java
+++ b/src/wms/src/main/java/org/geoserver/wms/map/OpenLayersMapOutputFormat.java
@@ -35,6 +35,7 @@ import org.geoserver.wms.MapProducerCapabilities;
 import org.geoserver.wms.WMS;
 import org.geoserver.wms.WMSMapContent;
 import org.geotools.geometry.jts.ReferencedEnvelope;
+import org.geotools.map.GridReaderLayer;
 import org.geotools.map.Layer;
 import org.geotools.map.WMSLayer;
 import org.geotools.referencing.CRS;
@@ -46,6 +47,7 @@ import org.geotools.util.Converters;
 import org.geotools.util.logging.Logging;
 import org.opengis.feature.type.FeatureType;
 import org.opengis.geometry.MismatchedDimensionException;
+import org.opengis.parameter.GeneralParameterValue;
 import org.opengis.referencing.FactoryException;
 import org.opengis.referencing.crs.CoordinateReferenceSystem;
 import org.opengis.referencing.crs.ProjectedCRS;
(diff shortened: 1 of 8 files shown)


src/wms/src/main/java/org/geoserver/wms/map/OpenLayersMapOutputFormat.java
```diff
@@ -287,6 +291,31 @@ private boolean hasOnlyCoverages(WMSMapContent mapContent) {
...
 
+    /**
+     * Helper method that checks if filtering support should be activated. If the map
+     * is not composed only by coverages then filtering should be activated. If the
+     * map is composed only of coverages but at least one of the coverages supports
+     * filtering, then filtering should be activated. Otherwise filtering capabilities
+     * will be deactivated.
+     */
+    private boolean supportsFiltering(WMSMapContent mapContent, boolean hasOnlyCoverages) {
+        return !hasOnlyCoverages || mapContent.layers().stream().anyMatch(layer -> {
```
aaime: Filtering against non supporting layers is just ignored. What about activating if there is at least one vector layer or one supporting raster layer instead?

nmco: That's what the first check does ``!hasOnlyCoverages`` (at least it was my goal maybe I'm wrong). So if there at least one vector layer (non raster layer) the filtering will be activated, if there is only coverages (raster) we check if there is at least one reader that supports filtering.

aaime: Ah ok, I see it now. Might warrant a comment.

aaime:
Looks good. The Travis build failure seems unrelated, but it's weird...

src/wms/src/main/java/org/geoserver/wms/map/OpenLayersMapOutputFormat.java
nmco: True adding it.

nmco:
Build restarted, lets see.

nmco:
I have a lot of errors and they all seem to be unrelated with this PR, for example:
```
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.431 sec <<< FAILURE!
testNoOutline(org.geoserver.wms.map.RasterWithoutOutlineMapStyleTest)  Time elapsed: 413 sec  <<< FAILURE!
java.lang.AssertionError: expected:<java.awt.Color[r=255,g=165,b=0]> but was:<java.awt.Color[r=255,g=255,b=255]>
```

And every time I restart the Travis build I get a different set of errors, no idea what is going on ... I'm merging anyway. I will monitoring GS build and if the build fails I will revert right away.

PR #2499, opened 2017-08-26
[GEOS-8265] Move GeoPackage support into core

aaime:
This moves the read part of the GeoPackage module into core.
Vector data works fine, but is slow extracting small portions because the spatial index is not used.
Raster data unfortunately does not work, seems like there is an axis flipping issue.

### DIFF
--- a/src/community/release/ext-geopkg.xml
+++ b/src/community/release/ext-geopkg.xml
@@ -11,9 +11,7 @@
       <includes>	
         <include>gt-mbtiles-*.jar</include>
         <include>gs-mbtiles-*.jar</include>
-        <include>gt-geopkg-*.jar</include>
         <include>gs-geopkg-*.jar</include>
-        <include>sqlite*.jar</include>
       </includes>
     </fileSet>
   </fileSets>
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -468,6 +468,12 @@
     <artifactId>gt-process-feature</artifactId>
     <version>${gt.version}</version>
    </dependency>
+   <dependency>
+    <groupId>org.geotools</groupId>
+    <artifactId>gt-geopkg</artifactId>
+    <version>${gt.version}</version>
+   </dependency>
+
 
    <dependency>
      <groupId>com.google.code.gson</groupId>
--- a/src/web/app/pom.xml
+++ b/src/web/app/pom.xml
@@ -143,6 +143,10 @@
     </dependency>
     <dependency>
       <groupId>org.geotools</groupId>
+      <artifactId>gt-geopkg</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.geotools</groupId>
       <artifactId>gt-wfs-ng</artifactId>
     </dependency>
     <dependency>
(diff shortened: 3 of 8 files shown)


src/community/release/ext-geopkg.xml
```diff
@@ -11,9 +11,7 @@
       <includes>	
         <include>gt-mbtiles-*.jar</include>
         <include>gs-mbtiles-*.jar</include>
-        <include>gt-geopkg-*.jar</include>
```
nmco: I had too look twice at the ``gt-mbtiles`` vs ``gs-mbtiles`` 😛

nmco:
By looking at the Travis build these modules failed:
```
[INFO] GeoWebCache (GWC) Module ........................... FAILURE [01:12 min]
[INFO] ArcSDE DataStore Extension ......................... FAILURE [ 30.718 s]
[INFO] GRIB Coverage Format ............................... FAILURE [ 44.346 s]
[INFO] NetCDF Coverage format ............................. FAILURE [ 18.162 s]
```

The ```GRIB Coverage Format`` format modules fails because the ``GeoPackageFormat`` class is missing:
Tests run: 2, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 3.319 sec <<< FAILURE!
testGribEdit(org.geoserver.web.data.store.raster.GribRasterEditPanelTest)  Time elapsed: 3130 sec  <<< ERROR!
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'geopkgStorePanel' defined in URL [jar:file:/home/travis/build/geoserver/geoserver/src/web/core/target/gs-web-core-2.12-SNAPSHOT.jar!/applicationContext.xml]: Initialization of bean failed; nested exception is org.springframework.beans.TypeMismatchException: Failed to convert property value of type 'java.lang.String' to required type 'java.lang.Class' for property 'factoryClass'; nested exception is java.lang.IllegalArgumentException: Cannot find class [org.geotools.geopkg.mosaic.GeoPackageFormat]
```

The other modules integration tests fail because they got exception report intead of the expected image, the underling error is probably the same (i.e. missing ``GeoPackageFormat`` class).

File: src/pom.xml [hunk 2]
```diff
@@ -468,6 +468,12 @@
     <artifactId>gt-process-feature</artifactId>
     <version>${gt.version}</version>
    </dependency>
+   <dependency>
+    <groupId>org.geotools</groupId>
+    <artifactId>gt-geopkg</artifactId>
+    <version>${gt.version}</version>
+   </dependency>
+
```
bradh: I had to add this to the gs-main pom.xml as well, or some of the extension modules fail tests.

bradh: The breakage is actually WCS 1.0, and its because we end up with "GeoPackage (mosaic)" in the supported formats: `<wcs:formats>GeoPackage (mosaic)</wcs:formats>`.

aaime:
The GeoTools merge should have fixed this one, let's try

aaime:
And it's finally green. Mergef

jodygarnett:
Thanks Andrea

PR #2500, opened 2017-08-27
[GEOS-8063] Fix WMS1.3.0 SLD definition can break getcapabilities

geojs:
This fix proposal aims to add a Title element in the Style's GetCapabilities response.
It resumes the work from https://github.com/geoserver/geoserver/pull/2200
See https://osgeo-org.atlassian.net/browse/GEOS-8063

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
+++ b/src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
@@ -1061,7 +1061,7 @@ public class Capabilities_1_3_0_Transformer extends TransformerBase {
             }
         }
 
-		private void handleStyles(final LayerInfo layer) {
+        private void handleStyles(final LayerInfo layer) {
             if (layer.getResource() instanceof WMSLayerInfo) {
                 // do nothing for the moment, we may want to list the set of cascaded named styles
                 // in the future (when we add support for that)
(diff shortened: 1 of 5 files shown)


src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
```diff
@@ -1082,7 +1082,14 @@ private void handleStyles(final LayerInfo layer) {
...
                 if (ftStyle.getDescription() != null) {
-                    element("Title", ftStyle.getDescription().getTitle());
+                    // PMT: WMS capabilities requires at least a title,
+                    // if description's title is null, use the name
+                    // for title.
+                    if (ftStyle.getDescription().getTitle() != null) {
+                        element("Title", ftStyle.getDescription().getTitle());
+                    } else {
+                        element("Title", defaultStyle.prefixedName());
+                    }
```
tbarsballe: If the description is null, the title will be missing. Probably want another else clause outside of `if (ftStyle.getDescription() != null)`

src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
```diff
@@ -1100,7 +1107,11 @@ private void handleStyles(final LayerInfo layer) {
                         start("Style");
                         element("Name", styleInfo.prefixedName());
                         if (ftStyle.getDescription() != null) {
-                            element("Title", ftStyle.getDescription().getTitle());
+                            if (ftStyle.getDescription().getTitle() != null) {
+                                element("Title", ftStyle.getDescription().getTitle());
+                            } else {
+                                element("Title", styleInfo.prefixedName());
+                            }
```
tbarsballe: Same as above; If the description is null, the title will be missing.

PR #2502, opened 2017-08-28
Intellij Quickstart

tbarsballe:
Another result of some feedback at the FOSS4G 2017 code sprint.

Split out the Eclipse section of the quickstart into its own page, and added an Intellij quickstart to go alongside it.

Also added a docguide tag to the extlinks, and extlinks to the developer guide, to simplify some linking.


nmco:
Interesting, I always use the ``Search for projects recursively`` option to retrieve the nested projects like the web ones, I guess this option is not needed anymore.

nmco:
An option I find useful when configuring the GeoServer start is the ``Single instance only`` option, we can only have one GeoServer instance running on the same port, this option will stop any running instance on that port and restart it automatically.

tbarsballe:
> Interesting, I always use the `Search for projects recursively` option to retrieve the nested projects like the web ones, I guess this option is not needed anymore.

I haven't needed to no, I think it can read the nested submodules out of the pom.

> An option I find useful when configuring the GeoServer start is the Single instance only option, we can only have one GeoServer instance running on the same port, this option will stop any running instance on that port and restart it automatically.

That seems super usefull; I might add that as a note.

jodygarnett:
If you folks are happy this can be merged.

nmco:
It looks good to me.

PR #2505, opened 2017-08-30
JTS 1.14.0 upgrade

jodygarnett:
small pom change to protect against transitive dependencies


jodygarnett:
This build requires https://github.com/geotools/geotools/pull/1665 to be merged first.

PR #2506, opened 2017-09-01
Js escape param names in OL preview template

jnewmoyer:


jodygarnett:
Patch has applied cherry-picked to master, 2.11.x and 2.10.x. Testing now before manual merge (manual merge to avoid the fun and games from your origional 2.11.2 tag starting point).

jodygarnett:
merged manually

PR #2510, opened 2017-09-07
[GEOS-8277] fix filtering

etj:
This PR
- fixes the NPE at runtime
- improves the coverage identification by excluding wmts cascading layers
- provides the testing logic.

A test is still @ignored since it requires a fix on the logic that detects when the filtering should be available.


src/wms/src/main/java/org/geoserver/wms/map/OpenLayersMapOutputFormat.java
```diff
@@ -311,10 +313,12 @@ private boolean supportsFiltering(WMSMapContent mapContent, boolean hasOnlyCover
                 return false;
             }
             GeneralParameterValue[] readParams = ((GridReaderLayer) layer).getParams();
-            for (GeneralParameterValue readParam : readParams) {
```
nmco: I would substitute ``supportsFiltering`` method by:
```
    /**
     * Helper method that checks if filtering support should be activated. If the map contains
     * at least one layer that is queryable, filtering should be activated.
     */
    private boolean supportsFiltering(WMSMapContent mapContent) {
        // returns TRUE if at least one layer supports filtering
        return mapContent.layers().stream().anyMatch(layer -> {
            if (layer instanceof FeatureLayer) {
                // vector layers support filtering
                return true;
            }
            if (!(layer instanceof GridReaderLayer)) {
                // filtering is not support for the remaining types
                return false;
            }
            // let's see if this coverage type supports filtering
            GeneralParameterValue[] readParams = ((GridReaderLayer) layer).getParams();
            if (readParams == null || readParams.length == 0) {
                // filtering is not supported
                return false;
            }
            for (GeneralParameterValue readParam : readParams) {
                if (readParam.getDescriptor().getName().getCode().equalsIgnoreCase("FILTER")) {
                    // the reader of this layer supports filtering
                    return true;
                }
            }
            // filtering is not supported
            return false;
        });
    }
```

This will also fix the behavior for the WMTS layers.

PR #2512, opened 2017-09-08
review of rest api docs

jodygarnett:
Mostly looking for missing query parameters and examples, we will get a short list of missing documentation as well.

### DIFF
--- a/src/restconfig/src/main/java/org/geoserver/rest/catalog/WorkspaceController.java
+++ b/src/restconfig/src/main/java/org/geoserver/rest/catalog/WorkspaceController.java
@@ -88,7 +88,7 @@ public class WorkspaceController extends AbstractCatalogController {
     }
 
     @GetMapping(value = "/{workspaceName}")
-    public RestWrapper<WorkspaceInfo> workspaceG(@PathVariable String workspaceName) {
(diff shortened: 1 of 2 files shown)


jodygarnett:
@ianturton you did the initial workspace swagger docs can you review my small change here?

src/restconfig/src/main/java/org/geoserver/rest/catalog/WorkspaceController.java
```diff
@@ -88,7 +88,7 @@ public RestWrapper workspacesGet() {
     }
 
     @GetMapping(value = "/{workspaceName}")
-    public RestWrapper<WorkspaceInfo> getWorkspace(@PathVariable String workspaceName) {
+    public RestWrapper<WorkspaceInfo> workspaceG(@PathVariable String workspaceName) {
```
tbarsballe: Looks like there is a mistake here?

jodygarnett: Yeah should be workspaceGet to agree with the naming convention; fixing now

jodygarnett:
corrected, will wait on travis build

PR #2516, opened 2017-09-12
fixed the test for the gwc reload controller.

vickdw:


tbarsballe:
Depends upon https://github.com/GeoWebCache/geowebcache/pull/540

PR #2521, opened 2017-09-14
[GEOS-8289] AbstractCatalogFacade.resolve(StyleInfo) resolves the workspace

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-8289

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/impl/AbstractCatalogFacade.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/AbstractCatalogFacade.java
@@ -11,6 +11,8 @@ import java.lang.reflect.Proxy;
 import java.rmi.server.UID;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 import org.geoserver.catalog.CatalogFacade;
 import org.geoserver.catalog.CatalogInfo;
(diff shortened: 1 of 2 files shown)


src/main/src/main/java/org/geoserver/catalog/impl/AbstractCatalogFacade.java
```diff
@@ -169,6 +169,16 @@ protected void resolve(LayerGroupInfo layerGroup) {
...
         setId(style);
+
+        // resolve the workspace
+        WorkspaceInfo resolved = ResolvingProxy.resolve(getCatalog(), style.getWorkspace());
+        if (resolved != null) {
+            resolved = unwrap(resolved);
+            style.setWorkspace(resolved);
+        } else {
+            // this means the workspace has not yet been added to the catalog, keep the proxy around
+            System.out.println("this means the workspace has not yet been added to the catalog, keep the proxy around");
```
aaime: Well, no, we cannot have a System.out here. A log should replace it instead.

tbarsballe: Indeed. I'll fix the place I copied it from too ( https://github.com/tbarsballe/geoserver/blob/1b4ef13ba465b82b732377ab2661590ab08e3234/src/main/src/main/java/org/geoserver/catalog/impl/AbstractCatalogFacade.java#L180 )

tbarsballe:
System.out call(s) replaced with proper logging statement.

PR #2522, opened 2017-09-15
GEOS-8286 Move store parts of geopackage to core docs.

bradh:


aaime:
Looks good, thank you!

PR #2525, opened 2017-09-16
Rest API Swagger doc QA - coverages and coveragestores

tbarsballe:


tbarsballe:
Two potentially outstanding issues with this one:

* For `/workspaces/{workspace}/coveragestores/{store}/{method}.{format}` POST (coverage store file upload / granule harvest), the filename parameter appears in the docs, but does not show up anywhere in the code.

*  The granule delete endpoint (`/workspaces/{workspace}/coveragestores/{store}/granules/{granuleId:.+}/{format}`) has an optional format path variable that is declared in the code but never appears to be used, and is largely undocumented.

tbarsballe:
Both outstanding issues now resolved.

tbarsballe:
Defaults and enum usage updated based on feedback.

PR #2526, opened 2017-09-18
Rest API Swagger doc QA - datastores and featuretypes

tbarsballe:
Also includes some code changes to bring the endpoints inline with documented functionality (POST to `/workspaces/{workspaceName}/featureTypes` was throwing a NPE)

### DIFF
--- a/doc/en/api/1.0.0/datastores.yaml
+++ b/doc/en/api/1.0.0/datastores.yaml
@@ -181,15 +181,6 @@ paths:
           required: true
           description: The name of the data store to delete.
           type: string
-        - name: purge
-          in: query
-          required: false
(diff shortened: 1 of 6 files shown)


tbarsballe:
Found one outstanding issue: The "purge" parameter for datastore delete is included in the parameter list in the code, but never actually used. Not sure if this is a code issue or a documentation issue.

tbarsballe:
> Found one outstanding issue: The "purge" parameter for datastore delete is included in the parameter list in the code, but never actually used. Not sure if this is a code issue or a documentation issue.

I have removed the unimplemented parameter.
However, based on discussion on the mailing lists, the functionality represented could be useful, so we should consider implementing this in the future. More details here: https://osgeo-org.atlassian.net/browse/GEOS-8296

PR #2527, opened 2017-09-18
Rest API Swagger doc QA - wms and wmts stores and layers

tbarsballe:


tbarsballe:
As with https://github.com/geoserver/geoserver/pull/2526, both wmsstore and wmtsstore delete has a "purge" parameter that is included in the parameter list in the code, but never actually used.

From what I can tell, in both cases this is the result of blindly copying from coveragestore delete, which uses this parameter to determine whether or not to delete the underlying store. This definatly does not make any sense for wms and wmts stores, which are just referencing a remote service.

PR #2530, opened 2017-09-20
Rest API Swagger doc QA - styles

tbarsballe:
As part of this, ran into an issue showing XML/SLD examples in Swagger Markdown.
Using an image for now, should try and figure out a workaround as soon as possible.


doc/en/api/1.0.0/styles.yaml
```diff
@@ -85,36 +102,23 @@ paths:
...
         - name: style
           in: path
           required: true
           description: Name of the style to retrieve.
           type: string
-        - name: purge
-          in: query
-          required: false
-          description: Specifies whether the underlying file containing the style should be deleted on disk. Allowable values are "true" or "false" (default).
-          type: boolean
```
jodygarnett: what was this doing here?

doc/en/api/1.0.0/styles.yaml
```diff
@@ -85,36 +102,23 @@ paths:
...
-        - name: quietOnNotFound 
-          in: query
-          required: false
-          description: When set to "true", will not log an exception when the style is not present. The 404 status code will still be returned. Allowable values are "true" or "false" (default).
-          type: boolean
```
jodygarnett: I thought this was handled by the dispatcher, so is valid everywhere ...

doc/en/api/1.0.0/styles.yaml
```diff
@@ -85,36 +102,23 @@ paths:
...
-        - name: pretty
-          in: query
-          required: false
-          description: Returns the style in a human-readable format, with proper whitespace and indentation. This parameter has no effect if you request a style in its native format. The HTML, XML, and JSON formats do not support this parameter.
-          type: boolean
```
jodygarnett: does pretty not work?

doc/en/api/1.0.0/styles.yaml
```diff
@@ -574,30 +605,35 @@ paths:
...
+        - name: default
+          type: boolean
+          in: query
+          description: Whether to make this the default style for the layer.
+          default: false
+          required: false
       consumes:
         - application/xml
         - application/json
-        - application/zip
```
jodygarnett: was was zip formally supported and then no longer works? or was the functionality never there ...

jodygarnett:
While I had some "huh?" questions there are no issues with this pull request.

doc/en/api/1.0.0/styles.yaml
tbarsballe: Excellent question. No idea. Seems like it came from a bulk copy-pasting of parameters.

doc/en/api/1.0.0/styles.yaml
tbarsballe: Yep, but we've been super inconsistent documenting it. I'm going to do a pass where I add it everywhere at the end. One thing I might do is add it as a definition in a separate file, and ref it everywhere.

doc/en/api/1.0.0/styles.yaml
tbarsballe: Nope, this got broke during the upgrade, see https://osgeo-org.atlassian.net/browse/GEOS-8294

doc/en/api/1.0.0/styles.yaml
tbarsballe: Never there

tbarsballe:
All you're 'huh' questions are sensible, I've answered them inline. There were/are a bunch of weird things with this endpoint.

doc/en/api/1.0.0/styles.yaml
jodygarnett: Can we remove it from everywhere because it does not add value to the client / server relationship. It is kind of a detail added for the automated test cases of interest to nobody.

So if you have to we can add a "note" when we introduce the REST API and take it out of the swagger docs.

PR #2532, opened 2017-09-21
Swagger API Docs - proper style example

tbarsballe:
Replaces 3c5093ca319243301e84d70e4bfcf4f48216562e with a proper style example
Errors described in that commit were erroneous, and only affected the output displayed by swagger-editor, not the actual swagger-ui used by the API docs.


doc/en/api/1.0.0/styles.yaml
```diff
@@ -192,12 +187,7 @@ paths:
         Using PUT with any other format will modify the content of the style. You can also use PUT with a ZIP file to upload a SLD 1.0 (`application/vnd.ogc.sld+xml`) file and any associated icon files
 
       parameters:
-        - name: styleBody
-          description: Style body information to modify.
-          in: body
-          required: true
-          schema:
-            $ref: "#/definitions/StyleInfoPost"
+        - $ref: "#/definitions/StyleInfoPost"
```
jodygarnett: you are giving up "to upload" vs "to modify"; I expect that is fine.

doc/en/api/1.0.0/styles.yaml
```diff
@@ -617,7 +597,7 @@ paths:
           in: body
           required: true
           schema:
-            $ref: "#/definitions/StyleInfoPost"
+            $ref: "#/definitions/StyleInfoPost/schema"
```
jodygarnett: you are tricky

doc/en/api/1.0.0/styles.yaml
```diff
@@ -660,33 +640,81 @@ paths:
...
-    properties:
-      name:
-        type: string
-        description: Name of style
-        example: roads_style
-      filename:
-        type: string
-        description: Name of filename containing SLD code
-        example: roads.sld
+          ```
```
jodygarnett: do we need an sld 1.1 example also?

doc/en/api/1.0.0/styles.yaml
```diff
@@ -660,33 +640,81 @@ paths:
...
-      JSON or XML style uploads follow the schema below.
+
+      The style body of a request
+      JSON or XML style uploads follow this schema. For example:
+
+        - application/xml: 
+          
+          ```
+          <style><name>roads_style</name><filename>roads.sld</filename></style>
+          ```
```
jodygarnett: These are both complete examples, rather than partial data structure to update ...

Examples to consider:
- minimal example with name and filename; it is complete enough for a POST? What format would it assume ...
- complete example with name, filename, format, legend, ...

What was our approach going to be?
- POST: create using a minimal viable example
- PUT: update using a minimal example
- GET: providing a maximal example showing all the stuff, even those things we cannot edit ...

doc/en/api/1.0.0/styles.yaml
tbarsballe: That would probably be useful, I suppose we could generate one using the REST API.

doc/en/api/1.0.0/styles.yaml
tbarsballe: Yeah, the description of the PUT / POST itself should describe the distinction fine. Not really needed in the description of the payload as well.

doc/en/api/1.0.0/styles.yaml
tbarsballe: _Technically_, a minimal PUT is just an empty object; the minimum useful PUT is any one attribute, so that is still not all that helpful.

I do like using GET for the maximal example, although there may be attributes you can GET that you can't PUT, depending on the object in question. I'm not entirely certain about this though.

jodygarnett: Well a minimal put with an empty object would not do anything right?

If you wanted to we could do a maximal PUT example that shows all the editable things; there is a good chance we could use the same example for GET and PUT for the majority of end points.

How about:

* POST: example create using a minimal viable example
* PUT: example showing all editable information
* GET: Only provide an example if required to show extra non-editable information

tbarsballe: I like this approach, sounds good. 
Although in the cases where GET and PUT are the same, we may still want to ref the PUT example from the GET, since they do show up under seperate headings, and some people may only look at one or the other.

jodygarnett: yes, well played and good reduction of duplication

PR #2533, opened 2017-09-22
Rest API Swagger doc - GWC swagger

johngschulz:
Part of first pass for the GWC swagger documentation

### DIFF
--- a/doc/en/api/1.0.0/gwcbounds.yaml
+++ b/doc/en/api/1.0.0/gwcbounds.yaml
@@ -1,0 +1,41 @@
+---
+swagger: '2.0'
+info:
+  version: 1.0.0
+  title: GeoWebCache Bounds
+  description: A RESTful interface through which get the bounds of a given layer based on srs.
+schemes:
+- http
+host: localhost:8080
+basePath: /geoserver/gwc/rest
+paths:
+  /bounds/{layer}/{srs}/{type}:
+    get:
+      produces:
+        - text/plain
+      operationId: boundsGet
+      summary: Returns the bounds for a layer based on srs.
+      description: Returns the bounds for a layer based on srs.
+      parameters:
+        - name: layer
+          in: path
+          required: true
+          description: The name of the layer to update.
+          type: string
+        - name: srs
+          in: path
+          required: true
+          description: The srs projection used against the layer to find the bounds such as EPSG:4326
+          type: string
+        - name: type
+          in: path
+          description: Accepts java as an extension.
+          required: true
+          type: string
+          enum: [java]
+      responses:
+        200:
+          description: OK
+          examples:
+            text/plain: |
+              {{0, 0, 1, 0, 0}, {0, 0, 3, 1, 1}, {0, 0, 7, 3, 2}, {0, 0, 15, 7, 3}, {0, 0, 31, 15, 4}, {0, 0, 63, 30, 5}, {0, 0, 127, 61, 6}, {0, 0, 255, 123, 7}, {0, 0, 511, 246, 8}, {0, 0, 1023, 493, 9}, {0, 0, 2047, 987, 10}, {0, 0, 4095, 1975, 11}, {0, 0, 8191, 3950, 12}, {0, 0, 16383, 7900, 13}, {0, 0, 32767, 15801, 14}, {0, 0, 65535, 31602, 15}, {0, 0, 131071, 63205, 16}, {0, 0, 262143, 126411, 17}, {0, 1, 524287, 252822, 18}, {0, 3, 1048575, 505645, 19}, {0, 6, 2097151, 1011291, 20}, {0, 12, 4194303, 2022582, 21}}
--- a/doc/en/api/1.0.0/gwcdiskquota.yaml
+++ b/doc/en/api/1.0.0/gwcdiskquota.yaml
@@ -1,0 +1,127 @@
+---
+swagger: '2.0'
+info:
+  version: 1.0.0
+  title: GeoWebCache DiskQuota
+  description: The REST API for Disk Quota management provides a RESTful interface through which clients can configure the disk usage limits and expiration policies for a GeoWebCache instance through simple HTTP calls.  Since disk usage increases geometrically by zoom level, one single seeding task could fill up an entire storage device. Because of this, GeoWebCache employs a disk quota system where one can specify the maximum amount of disk space to use for a particular layer, as well as logic on how to proceed when that quota is reached. There are two different policies for managing the disk quotas - Least Frequently Used (LFU) and Least Recently Used (LRU).
+schemes:
+- http
+host: localhost:8080
+basePath: /geoserver/gwc/rest
+paths:
+  /diskquota:
+    get:
+      produces:
+        - application/xml
+        - application/json
+      operationId: diskQuotaGet
+      summary: Returns the gwc Quota configuration
+      description: Returns the GeoWebCache DiskQuota configurations which include disk usage limits and expritation policies for a gwc instance.
+      responses:
+        200:
+          description: OK
+          examples:
+            application/xml: |
+              <gwcQuotaConfiguration>
+                <enabled>true</enabled>
+                <diskBlockSize>2048</diskBlockSize>
+                <cacheCleanUpFrequency>5</cacheCleanUpFrequency>
+                <cacheCleanUpUnits>SECONDS</cacheCleanUpUnits>
+                <maxConcurrentCleanUps>5</maxConcurrentCleanUps>
+                <globalExpirationPolicyName>LRU</globalExpirationPolicyName>
+                <globalQuota>
+                  <value>100</value>
+                  <units>MiB</units>
+                </globalQuota>
+                <layerQuotas/>
+              </gwcQuotaConfiguration>
+            appliction/json: |
+               {
+                 "org.geowebcache.diskquota.DiskQuotaConfig": {
+                   "enabled": false,
+                   "cacheCleanUpFrequency": 10,
+                   "cacheCleanUpUnits": "SECONDS",
+                   "maxConcurrentCleanUps": 2,
+                   "globalExpirationPolicyName": "LFU",
+                   "globalQuota": {
+                     "id": 0,
+                     "bytes": 524288000
+                   }
+                 }
+               }
+    put:
+      consumes:
+        - application/xml
+        - application/json
+      operationId: diskQuotaPut
+      summary: Modify properties of a gwc instance disk quota configuration.
+      description:  The request body for PUT should contain desired properties to be modified.
+      parameters:
+        - name: requestBody
+          in: body
+          description: The diskquota configuration with modified property values.
+          schema:
+            $ref: "#/definitions/DiskQuota"
+      responses:
+        200:
+          description: OK
+
+
+definitions:
+  DiskQuota:
+    title: gwcQuotaConfiguration
+    type: object
+    properties:
+        enabled:
+          type: boolean
+          description: Determines if Disk Quota is to be used.
+        diskBlockSize:
+          type: number
+          description: The number of bytes allocated to Disk Quota.
+        cacheCleanUpFrequency:
+          type: number
+          description: GeoWebCache will not truncate the cache as soon as the disk quota is exceeded. Instead, it polls the store at given intervals, with this time interval set.
+        cacheCleanUpUnits:
+          type: string
+          enum: [SECONDS, MINUTES, HOURS, DAYS]
+          description: The time unit used to describe the number in cacheCleanUpFrequency.
+        maxConcurrentCleanUps:
+          type: number
+          description: The amount of threads to use when processing the disk quota.
+        globalExpirationPolicyName:
+          type: string
+          enum: [LRU, LSU]
+          description: When a disk quota is reached, further tiles will be saved at the expense of other tiles which will be truncated. The Least Frequently Used (LFU) policy will analyze the disk quota page store and delete the pages of tiles that have been accessed the least often. The Least Recently Used (LRU) policy will analyze the diskquota page store and delete the tiles that haven’t been accessed in the longest amount of time.
+        globalQuota:
+          type: object
+          description: When the global quota is exceeded, first any explicitly configured quota is enforced, and then the global quota is enforced acting upon all the remaining layers until the global quota is reached back.
+          properties:
+            value:
+              type: number
+              description: Numeric value for global quota.
+            units:
+              type: string
+              enum: [B, KiB, MiB, GiB, TiB]
+              description: The type of unit quantified by value.
+        layerQuota:
+          type: object
+          description: When a layer's disk quota is reached, further tiles will be saved at the expense of other tiles which will be truncated.
+          properties:
+            layer:
+              type: string
+              description: The layer name.
+            expirationPolicyName:
+              type: string
+              enum: [LRU, LSU]
+              description: When a disk quota is reached, further tiles will be saved at the expense of other tiles which will be truncated. The Least Frequently Used (LFU) policy will analyze the disk quota page store and delete the pages of tiles that have been accessed the least often. The Least Recently Used (LRU) policy will analyze the diskquota page store and delete the tiles that haven’t been accessed in the longest amount of time.
+            quota:
+              type: object
+              description: Container for quota value and units
+              properties:
+                value:
+                  type: number
+                  description: Numeric value for global quota.
+                units:
+                  type: string
+                  enum: [B, KiB, MiB, GiB, TiB]
+                  description: The type of unit quantified by value.
--- a/doc/en/api/1.0.0/gwcfilterupdate.yaml
+++ b/doc/en/api/1.0.0/gwcfilterupdate.yaml
@@ -1,0 +1,57 @@
+---
+swagger: '2.0'
+info:
+  version: 1.0.0
+  title: GeoWebCache Filter Update
+  description: The REST API for Updating GWC filter
+schemes:
+- http
+host: localhost:8080
+basePath: /geoserver/gwc/rest
+paths:
+  /filter/{filterName}/update/{updateType}:
+    post:
+      consumes:
+        - application/xml
+        - application/zip
+      operationId: filterUpdatePost
+      summary: Updates a given layer filter by way of xml or zip file.
+      description: Restfully updates the given filterName with parameters provided in the xml or zip
+      parameters:
+        - name: filterName
+          in: path
+          required: true
+          description: The name of the filter to update.
+          type: string
+        - name: updateType
+          in: path
+          required: true
+          description: One of 'zip' or 'xml'
+          type: string
+          enum: [zip, xml]
+        - name: requestBody
+          in: body
+          required: true
+          description: The parameters that are accepted by a given filter.
+          schema:
+            $ref: "#/definitions/requestFilterUpdate"
+      responses:
+        200:
+          description: OK
+        400:
+          description: Bad Request
+
+definitions:
+  requestFilterUpdate:
+    title: requestFilterUpdate
+    type: object
+    properties:
+      gridSet:
+        type: string
+        description: The projection used in the request filter.
+      zoomStart:
+        type: integer
+        description: This is the minimum zoom level for which the filter is applied.
+      zoomStop:
+        type: integer
+        description: The maximum zoom level for which the filter is applied.
--- a/doc/en/api/1.0.0/gwclayers.yaml
+++ b/doc/en/api/1.0.0/gwclayers.yaml
@@ -1,0 +1,231 @@
+---
+swagger: '2.0'
+info:
+  version: 1.0.0
+  title: GeoWebCache Layers
+  description: A layer is a published resource (feature type or coverage).
+schemes:
+- http
+host: localhost:8080
+basePath: /geoserver/gwc/rest
+paths:
+  /layers:
+    get:
+      operationId: layersGet
+      summary: Get a list of cached layers
+      description: Displays a list of all cached layers on the server.
+      produces:
+        - application/xml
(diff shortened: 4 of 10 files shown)


doc/en/api/1.0.0/gwclayers.yaml
```diff
@@ -0,0 +1,236 @@
...
+      produces:
+        - application/xml
+      parameters:
+        - name: layerName
+          in: path
+          required: true
+          description: The name of the layer to retrieve.
+          type: string
+        - name: quietOnNotFound
+          in: query
```
tbarsballe: We've been ignoring / removing the quietOnNotFound parameter everywhere, since it technically applies to all endpoints. I'll be adding a note to the top-level api docs, but that is about it.

See discussion here: https://github.com/geoserver/geoserver/pull/2530

doc/en/api/1.0.0/gwclayers.yaml
```diff
@@ -0,0 +1,236 @@
...
+                <parameterFilters>
+                  <styleParameterFilter>
+                    <key>STYLES</key>
+                    <defaultValue/>
+                  </styleParameterFilter>
+                </parameterFilters>
+                <gutter>0</gutter>
+              </GeoServerLayer>
+        404:
+          description: Rest Exception 404 Unknown layer
```
tbarsballe: Given that you alreay have 404 in the parent tag, just "Unknown Layer" is probably fine.

doc/en/api/1.0.0/gwclayers.yaml
```diff
@@ -0,0 +1,236 @@
...
+      responses:
+        201:
+          description: The layer was successfully created.
+        400:
+          description: Layer with requested name already exists, use POST if you want to replace it.
+
+    post:
+      operationId: layersNamePost
+      summary: Modify a cached layer.
+      description: Modifies an existing cached layer on the server. Use the "Accept:" header to specify format or append an extension to the endpoint (example "/layers/{layer}.xml" for XML).
```
tbarsballe: Since this is just text, probably want to add the base path too it, so no one gets confused with the geoserver layers endpoint. e.g. `/gwc/rest/layers/{layer}.xml`.

While that does make this a bit less applicable for standalone geowebcache, this is in the geoserver docs, so should probably be favouring embedded.

doc/en/api/1.0.0/gwcmasstruncate.yaml
```diff
@@ -0,0 +1,48 @@
...
+      parameters:
+        - name: requestType
+          in: query
+          required: true
+          description: The type of mass truncate request to use.
+          type: string
+
+        - name: layer
+          in: query
+          required: false
```
tbarsballe: Does this have a default value? If so, can you add it. If not, describe what happens if layer is not included

doc/en/api/1.0.0/gwcmasstruncate.yaml
```diff
@@ -0,0 +1,48 @@
...
+      operationId: masstruncatePost
+      summary: Issue a mass truncate request
+      description: Issues a mass truncate request based on the request type parameter. truncateLayer, will clear all caches associated with a named layer, including all permutations of gridset, parameter filter values, and image formats.
+      parameters:
+        - name: requestType
+          in: query
+          required: true
+          description: The type of mass truncate request to use.
+          type: string
+
```
tbarsballe: Presumably this has a finite set of available types. Can you add an enum param which iterates those. See here for an example: https://github.com/geoserver/geoserver/blob/master/doc/en/api/1.0.0/featuretypes.yaml#L42

doc/en/api/1.0.0/gwcreload.yaml
```diff
@@ -0,0 +1,19 @@
...
+schemes:
+- http
+host: localhost:8080
+basePath: /geoserver/gwc/rest
+paths:
+  /reload:
+    post:
+      operationId: reloadPost
+      summary: Reloads GWC settings/
+      description: Reloads the GeoWebCache settings after making changes to the configuration. Can include the the data parameter in request to reload the new settings from data.
```
tbarsballe: You mention a data parameter, but don't document it as a parameter. Can you add it?

doc/en/api/1.0.0/gwcseed.yaml
```diff
@@ -0,0 +1,100 @@
...
+        200:
+          description: OK
+          examples:
+            text/json: |
+              {"long-array-array":[
+                [364832,2761733810,998564,16,1],
+                [373472,2761733810,975448,17,1],
+                [363760,2761733810,1001509,18,1]
+              ]}
+  /seed/{layer}.json:
```
tbarsballe: Are the xml and json endpoints really different enough to need seperate documentation? Can't you use a single endpoint and have two entries under produces / consumes?

doc/en/api/1.0.0/gwcmasstruncate.yaml
```diff
@@ -0,0 +1,48 @@
...
+host: localhost:8080
+basePath: /geoserver/gwc/rest
+paths:
+  /masstruncate:
+    get:
+      operationId: masstruncateGet
+      summary: Returns available request types for truncation
+      description: Returns xml containing the request type capabilities for mass truncation
+      produces:
+        - application/xml
```
tbarsballe: Most other endpoints support both json and xml from what I can see. Is this an ommission, or does this endpoint really only support XML

johngschulz: This will accept a json extension in the request, but it will still respond/produce xml.
```curl -v -u admin:geoserver -X GET "http://localhost:8080/geoserver/gwc/rest/masstruncate.json"
Note: Unnecessary use of -X or --request, GET is already inferred.
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
* Server auth using Basic with user 'admin'
> GET /geoserver/gwc/rest/masstruncate.json HTTP/1.1
> Host: localhost:8080
> Authorization: Basic YWRtaW46Z2Vvc2VydmVy
> User-Agent: curl/7.54.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Date: Mon, 25 Sep 2017 17:20:19 GMT
< X-Frame-Options: SAMEORIGIN
< Content-Type: application/json;charset=ISO-8859-1
< Content-Length: 283
< Server: Jetty(9.2.13.v20150730)
< 
* Connection #0 to host localhost left intact
<massTruncateRequests href="http://localhost:8080/geoserver/gwc/rest/masstruncate.json"> <requestType>truncateLayer</requestType> <requestType>truncateParameters</requestType> <requestType>truncateOrphans</requestType> <requestType>truncateExtent</requestType></massTruncateRequests>```

File: doc/en/api/1.0.0/gwcbounds.yaml [hunk 9]
```diff
@@ -0,0 +1,41 @@
+---
+swagger: '2.0'
+info:
+  version: 1.0.0
+  title: GeoWebCache Bounds
+  description: The REST API for Disk Quota management provides a RESTful interface through which get the bounds of a given layer based on srs.
```
jodygarnett: what does disk quota management have to do with this one?

doc/en/api/1.0.0/gwcbounds.yaml
```diff
@@ -0,0 +1,41 @@
...
+        - name: type
+          in: path
+          description: Accepts java as an extension.
+          required: true
+          type: string
+      responses:
+        200:
+          description: OK
+          examples:
+            application/xml: |
```
jodygarnett: this appears to be a json example

doc/en/api/1.0.0/gwcfilterupdate.yaml
```diff
@@ -0,0 +1,46 @@
...
+      parameters:
+        - name: filterName
+          in: path
+          required: true
+          description: The name of the filter to update.
+          type: string
+        - name: updateType
+          in: path
+          required: true
+          description: One of 'zip' or 'xml'
```
jodygarnett: We can make this an enum of 'zip' or 'xml'.

See: https://swagger.io/docs/specification/data-models/enums/

```
schema:
  type: string
  enum: [zip, xml]
```

johngschulz: Updated along with other similar parameters to use enums.

doc/en/api/1.0.0/gwcbounds.yaml
johngschulz: Corrected an updated

doc/en/api/1.0.0/gwcbounds.yaml
johngschulz: This was clumsy of me - fixed.

doc/en/api/1.0.0/gwcreload.yaml
johngschulz: Added in some clarification.  Examples I saw used -d 'data' to reload the configuration based on that data setting.

doc/en/api/1.0.0/gwcseed.yaml
johngschulz: I consolidated where I could - the /rest/seed.json is the only endpoint that queries running task for all layers and it only accepts a json extension.

doc/en/api/1.0.0/gwcbounds.yaml
```diff
@@ -0,0 +1,41 @@
...
+          description: Accepts java as an extension.
+          required: true
+          type: string
+          enum: [java]
+      responses:
+        200:
+          description: OK
+          examples:
+            text/json: |
+              {{0, 0, 1, 0, 0}, {0, 0, 3, 1, 1}, {0, 0, 7, 3, 2}, {0, 0, 15, 7, 3}, {0, 0, 31, 15, 4}, {0, 0, 63, 30, 5}, {0, 0, 127, 61, 6}, {0, 0, 255, 123, 7}, {0, 0, 511, 246, 8}, {0, 0, 1023, 493, 9}, {0, 0, 2047, 987, 10}, {0, 0, 4095, 1975, 11}, {0, 0, 8191, 3950, 12}, {0, 0, 16383, 7900, 13}, {0, 0, 32767, 15801, 14}, {0, 0, 65535, 31602, 15}, {0, 0, 131071, 63205, 16}, {0, 0, 262143, 126411, 17}, {0, 1, 524287, 252822, 18}, {0, 3, 1048575, 505645, 19}, {0, 6, 2097151, 1011291, 20}, {0, 12, 4194303, 2022582, 21}
```
tbarsballe: This doesn't look like valid JSON; lists should use square brackets; curly brackets are specifically for objects.
If this is actually what GWC returns, then it is violating the JSON spec.

doc/en/api/1.0.0/gwcdiskquota.yaml
```diff
@@ -0,0 +1,50 @@
...
+                </globalQuota>
+                <layerQuotas/>
+              </gwcQuotaConfiguration>
+    put:
+      consumes:
+        - application/xml
+        - application/json
+      operationId: diskQuotaPut
+      summary: Modify properties of a gwc instance disk quota configuration.
+      description:  The request body for PUT should contain desired properties to be modified.
```
tbarsballe: Add a parameter for the request body

doc/en/api/1.0.0/gwcdiskquota.yaml
```diff
@@ -0,0 +1,50 @@
...
+      responses:
+        200:
+          description: OK
+          examples:
+            application/json: |
+              {"gwcQuotaConfiguration":{"diskBlockSize":2048,"enabled":true,"globalQuota":{"value":"100","units":"MiB"}}
```
tbarsballe: Is this the content of the response, or the request? If the request, it belongs in the request body parameter, not in the response examples.
If the response, then you should add a produces section since the response has content.

doc/en/api/1.0.0/gwcdiskquota.yaml
```diff
@@ -0,0 +1,50 @@
...
+                <cacheCleanUpFrequency>5</cacheCleanUpFrequency>
+                <cacheCleanUpUnits>SECONDS</cacheCleanUpUnits>
+                <maxConcurrentCleanUps>5</maxConcurrentCleanUps>
+                <globalExpirationPolicyName>LRU</globalExpirationPolicyName>
+                <globalQuota>
+                  <value>100</value>
+                  <units>MiB</units>
+                </globalQuota>
+                <layerQuotas/>
+              </gwcQuotaConfiguration>
```
tbarsballe: In cases where we support multiple media types (e.g. XML and JSON) it is useful to have an example of both

doc/en/api/1.0.0/gwcfilterupdate.yaml
```diff
@@ -0,0 +1,47 @@
...
+  /filter/{filterName}/update/{updateType}:
+    post:
+      consumes:
+        - application/xml
+        - application/zip
+      operationId: filterUpdatePost
+      summary: Updates a given layer filter by way of xml or zip file.
+      description: Restfully updates the given filterName with parameters provided in the xml or zip
+      parameters:
+        - name: filterName
```
tbarsballe: request body should be one of the parameters

doc/en/api/1.0.0/gwcfilterupdate.yaml
```diff
@@ -0,0 +1,47 @@
...
+          required: true
+          description: One of 'zip' or 'xml'
+          type: string
+          enum: [zip, xml]
+      responses:
+        200:
+          description: OK
+          examples:
+            application/xml: |
+              <parameterFilters>
```
tbarsballe: As above - is this the request body or the response?

doc/en/api/1.0.0/gwclayers.yaml
```diff
@@ -0,0 +1,231 @@
...
+schemes:
+- http
+host: localhost:8080
+basePath: /geoserver/gwc/rest
+paths:
+  /layers:
+    get:
+      operationId: layersGet
+      summary: Get a list of cached layers
+      description: Displays a list of all cached layers on the server. You must use the "Accept:" header to specify format or append an extension to the endpoint (example "/layers.xml" for XML)
```
tbarsballe: > You must use the "Accept:" header to specify format or append an extension to the endpoint (example "/layers.xml" for XML)

This seems redundant, if the only supported format is XML

doc/en/api/1.0.0/gwcmemorycachestatistics.yaml
```diff
@@ -0,0 +1,36 @@
...
+                <hitCount>0</hitCount>
+                <missCount>0</missCount>
+                <evictionCount>0</evictionCount>
+                <totalCount>0</totalCount>
+                <hitRate>100.0</hitRate>
+                <missRate>0.0</missRate>
+                <currentMemoryOccupation>0.0</currentMemoryOccupation>
+                <totalSize>67108864</totalSize>
+                <actualSize>0</actualSize>
+              </gwcInMemoryCacheStatistics>
```
tbarsballe: Since this supports XML and JSON, it would be useful to also have a json example

doc/en/api/1.0.0/gwcseed.yaml
```diff
@@ -0,0 +1,138 @@
...
+basePath: /geoserver/gwc/rest
+paths:
+  /seed.json:
+    get:
+      operationId: seedGet
+      summary: Query's currently running GWC task
+      description: Query's and returns a json array of the status for all currently running task. Requires .json extension in the request. The array contains a set of long in the following order:[tiles processed, total number of tiles to process, number of remaining tiles, Task ID, Task status].
+        The returned task status will be one of -1 = ABORTED, 0 = PENDING, 1 = RUNNING, 2 = DONE
+      produces:
+        - text/json
```
tbarsballe: You have `text/json` here, but `application/json` elsewhere. Can you confirm that GWC really uses different media types for JSON for different endpoints?

doc/en/api/1.0.0/gwcseed.yaml
```diff
@@ -0,0 +1,138 @@
...
+                [363760,2761733810,1001509,18,1]
+              ]}
+  /seed/{layer}.{format}:
+    get:
+      operationId: layerSeedGet
+      summary: Query's the currently running GWC task for a given layer
+      description: Returns HTML UI page of running GWC task for a given layer and available parameters and tasks to execute orQuery's and returns a json array of the status for currently running task of a given layer if using json extension. The array contains a set of long in the following order:[tiles processed, total number of tiles to process, number of remaining tiles, Task ID, Task status]
+      produces:
+        - text/json
+        - application/html
```
tbarsballe: You have `application/html` here, but `text/html` elsewhere. Can you confirm that GWC really uses different media types for HTML for different endpoints?

doc/en/api/1.0.0/gwcseed.yaml
```diff
@@ -0,0 +1,138 @@
...
+              </html>
+        400:
+          description: Bad Request
+    post:
+      operationId: layerSeedPost
+      summary: Issue a seed, reseed or truncate task request
+      description: Executes an issue to seed, reseed or truncate task request for a layer and returns HTML UI page of running GWC tasks and tasks to execute.
+      consumes:
+        - text/json
+        - text/xml
```
tbarsballe: You have `text/xml` here, but `application/xml` elsewhere. Can you confirm that GWC really uses different media types for XML for different endpoints?

doc/en/api/1.0.0/gwcseed.yaml
```diff
@@ -0,0 +1,138 @@
...
+                  'srs':{
+                    'number':4326
+                  },
+                'zoomStart':1,
+                'zoomStop':12,
+                'format':'image\/png',
+                'type':'truncate',
+                'threadCount':4
+                }
+              }
```
tbarsballe: These examples look good, nice to see examples for both output formats

tbarsballe:
One more thing: You'll want to add links to the api docs to https://github.com/tbarsballe/geoserver/blob/master/doc/en/user/source/rest/index.rst

Probably add links in a subsection right above the importer subsection.

doc/en/api/1.0.0/gwcbounds.yaml
johngschulz: In the code this is actually a built up string, it only accepts "java" as the type in the path and simply seems to build this string in the code.  I'm not too sure what the desired format is intended to be now that I look at it again.

tbarsballe: What does the "Content-type" header of the response say?

johngschulz: text/plain; updated to such.

doc/en/api/1.0.0/gwcmemorycachestatistics.yaml
johngschulz: Added json example.

johngschulz:
I think I've addressed most of the concerns - I still need to see clarification for the gwcfilterupdate from someone.

doc/en/api/1.0.0/gwcdiskquota.yaml
johngschulz: Addressed

doc/en/api/1.0.0/gwcdiskquota.yaml
johngschulz: Added.

doc/en/api/1.0.0/gwcfilterupdate.yaml
johngschulz: Seeking clarification on this, I'm now not sure if I'm using the correct parameters for this endpoint - could be one of:
http://geowebcache.org/docs/1.11.0/configuration/layers/parameterfilters.html
or
http://geowebcache.org/docs/1.11.0/configuration/layers/requestfilters.html

The latter has limited documentation and neither has much documentation for REST endpoints.

johngschulz: I've left the request body out of this endpoint, it is looking for the requestfilters - perhaps @smithkm can help me out with this one at some point.

tbarsballe: Looking at [FilterUpdateController](https://github.com/GeoWebCache/geowebcache/blob/master/geowebcache/rest/src/main/java/org/geowebcache/rest/controller/FilterUpdateController.java#L83), I believe it is dealing with RequestFilters.

As for the structure of the body, I can't find anything useful; I agree we probably need help from @smithkm

tbarsballe:
Alright, looks like all feedback has been addressed. Merging.

doc/en/api/1.0.0/gwcseed.yaml
wenyu-start: @tbarsballe 
微信图片_20260302171108_111_31
I would like to ask where to find the corresponding API development documentation for calling Kill Task. Currently, I have not found it in the documentation.

tbarsballe: If you can't find it in the GeoServer documentation, it's probably only documented in GeoWebCache.
See: https://github.com/geowebcache/geowebcache/blob/master/documentation/en/user/source/rest/seed.rst#terminating-running-tasks

That could be a nice addition to the GeoServer REST docs.

wenyu-start: Thank you very much for your reply. I have found a way to terminate slicing of a certain layer or all layers, but I still cannot find a way to terminate a certain task. For example, I passed in the task ID and then killed this process.

tbarsballe: In that case, I'd suggest asking on the [GeoServer User Forum](https://discourse.osgeo.org/c/geoserver/user/51). I haven't touched this part of the API in quite some time, so I don't have any more info than what's in the docs.

PR #2534, opened 2017-09-22
Rest API Swagger doc QA - layers

tbarsballe:


johngschulz:
Stylistically everything looks good, from my experience in swagger and everything appears to work fine in the swagger editor.  
I assume in your discussions it was decided to remove `quietOnNotFound` and throw an exception.  Looks good.

tbarsballe:
ResourceNotFoundException gets captured and ignored by a top-level filter that applies to all requests if quietOnNotFound is set. Consequently, the previous implementation was redundant.

johngschulz:
Sounds good

PR #2536, opened 2017-09-25
Rest API Swagger doc QA - settings

tbarsballe:


doc/en/api/1.0.0/owsservices.yaml
```diff
@@ -691,10 +691,10 @@ paths:
 
     delete:
       operationId: deleteWFSWorkspaceSettings
-      description: Invalid. Use PUT to edit settings.
+      description: Deletes a workspace-specific WFS setting.
       responses:
-        405:
-          description: Method Not Allowed
+        200:
```
jodygarnett: This should be `204` right? Assume we cannot make such a change.

tbarsballe: 204 has a specific enough use-case that it doesn't necessarily apply here, and we already use 200 for all similar requests throughout the rest API, so no, I don't think we want to make such a change.

PR #2537, opened 2017-09-25
Rest API Swagger doc QA - templates

tbarsballe:


doc/en/api/1.0.0/templates.yaml
```diff
@@ -114,6 +209,289 @@ paths:
...
+          description: Method Not Allowed
+    put:
+      operationId: templatesCoveragePut
+      description: Invalid. PUT to /{template} to edit a template.
+      responses:
+        405:
+          description: Method Not Allowed
+    delete:
+      operationId: templatesCoverageDelete
+      description: Invalid. Delete from /{template} to remove a template.
```
jodygarnett: Is the description markdown? Do you need to use:

```
Invalid. Delete from `/{template}` to remove a template.
```

doc/en/api/1.0.0/templates.yaml
```diff
@@ -114,6 +209,289 @@ paths:
...
+      responses:
+        200:
+          description: OK
+        404: 
+          description: Template not found
+
+  /workspaces/{workspace}/templates/{template}.ftl:
+    parameters:
+      - $ref: "#/parameters/Workspace"
+      - $ref: "#/parameters/Template"
```
jodygarnett: nicely done reusing parameter definitions

doc/en/api/1.0.0/templates.yaml
tbarsballe: That would probably be better, yes.

PR #2538, opened 2017-09-25
Rest API Swagger doc QA - resources

tbarsballe:
Adds swagger docs for resources

Also includes a fix for https://osgeo-org.atlassian.net/browse/GEOS-8300 as a separate commit (discovered during doc QA)

### DIFF
--- a/doc/en/api/1.0.0/resource.yaml
+++ b/doc/en/api/1.0.0/resource.yaml
@@ -1,0 +1,337 @@
+---
+swagger: '2.0'
+info:
+  version: 1.0.0
+  title: GeoServer Resources
+  description: A resource is any item in the data directory that does not represent configuration. Typical resources include styles and icons.
+
+paths:
+  /resource/{pathToResource}:
+    parameters:
+      - name: pathToResource
+        in: path
+        type: string
+        required: true
+        description: The full path to the resource. Required, but may be empty; a request to `/resource` references the top level resource directory.
+    get:
+      operationId: resourceGet
+      description: | 
+        Download a resource, list contents of directory, or show formatted resource metadata.
+
+        Response content depends upon parameters.
+
+        With `operation=default`, if the request is made against a non-directory resource, the content of the resource is returned.
+
+        For example, `/resource/styles/default_point.sld?operation=default`
+
+        ```
+        <?xml version="1.0" encoding="UTF-8"?>
+        <StyledLayerDescriptor version="1.0.0" 
+         xsi:schemaLocation="http://www.opengis.net/sld StyledLayerDescriptor.xsd" 
+         xmlns="http://www.opengis.net/sld" 
+         xmlns:ogc="http://www.opengis.net/ogc" 
+         xmlns:xlink="http://www.w3.org/1999/xlink" 
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+          <!-- a Named Layer is the basic building block of an SLD document -->
+          <NamedLayer>
+            <Name>default_point</Name>
+            <UserStyle>
+            <!-- Styles can have names, titles and abstracts -->
+              <Title>Default Point</Title>
+              <Abstract>A sample style that draws a point</Abstract>
+              <!-- FeatureTypeStyles describe how to render different features -->
+              <!-- A FeatureTypeStyle for rendering points -->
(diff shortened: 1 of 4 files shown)


doc/en/api/1.0.0/resource.yaml
```diff
@@ -0,0 +1,210 @@
...
+  description: A resource is any item in the data directory that does not represent configuration. Typical resources include styles and icons.
+
+paths:
+  /resource/{pathToResource}:
+    parameters:
+      - name: pathToResource
+        in: path
+        type: string
+        required: true
+        description: The full path to the resource. Required, but may be empty; a request to `resource` references the top level resource directory.
```
jodygarnett: If it is empty does that make `required: true` above incorrect? Or is this a complicated way of saying that the `/` is required....

doc/en/api/1.0.0/resource.yaml
```diff
@@ -0,0 +1,210 @@
...
+        - name: operation
+          in: query
+          type: string
+          description: The type of GET operation. `default` returns a list of the contained resources in the case of a directory resource, or the actual resource contents in the case of a resource resource.`metadata` requests a metadata summary of the resource.
+          required: false
+          default: default
+          enum:
+            - default
+            - metadata
+            - move
```
jodygarnett: Does get really perform a move?

doc/en/api/1.0.0/resource.yaml
```diff
@@ -0,0 +1,210 @@
...
+          required: false
+          default: html
+          enum:
+            - html
+            - xml
+            - json
+      responses:
+        200:
+          description: OK
+          headers:
```
jodygarnett: while we document the header here, do we also need to document the response body?

doc/en/api/1.0.0/resource.yaml
```diff
@@ -0,0 +1,210 @@
...
+        - name: operation
+          in: query
+          type: string
+          description: The type of PUT operation. `default` creates a new resource or alters an existing resource. `move` moves the resource to a new location. `copy` duplicates the resource to a new location
+          required: false
+          default: default
+          enum:
+            - default
+            - move
+            - copy
```
jodygarnett: It makes more sense for move and copy to be here, not sure why they were on `get`

doc/en/api/1.0.0/resource.yaml
tbarsballe: The slash isn't required, no (its not part of the path parameter anyways).
I considered required=false, that just seemed a bit weird for a path parameter. If you think that would be clearer, I am happy to change it.

doc/en/api/1.0.0/resource.yaml
tbarsballe: Nope, thats an error, good catch

doc/en/api/1.0.0/resource.yaml
tbarsballe: We should, yes. Unfortunately, that is impossible within the constraints of the Swagger 2.0 API, due to the variety in possible bodies. I tried a bunch of approached within the constraints of the API, but didn't get anything that actually worked.

I think I'm going to just add a bunch of examples to the description

jodygarnett: Could we document the metadata response (the only structured response) and use the description to document the `*/*` case?

tbarsballe: Its a bit more complicated than that unfortunately.
A regular GET against a resource will return the actual resource
A regular GET against a directory will return a "ResourceDirectory", which is a structured response listing information about the directory and its children.
A GET with operation=METADATA will return a "ResourceMetadata", which is a structured response listing information about the resource. This will be returned even if the resource is a directory, and does not list the children.

I have schemas for both these responses in the definitions section, but wasn't able to construct a suitable response definition for all the possible cases.

doc/en/api/1.0.0/resource.yaml
tbarsballe: Just tried out with required=false, and apparently the swagger rules don't like it if you have a path variable that isn't required, so I guess that will have to stay as-is

tbarsballe:
Fixed extra move and copy parameters, added examples to get description.

PR #2542, opened 2017-09-27
ignore user java extensions [GEOS-8304]

jodygarnett:
See https://osgeo-org.atlassian.net/browse/GEOS-8304

Signed-off-by: Jody Garnett <jody.garnett@gmail.com>

### DIFF
--- a/src/release/installer/mac/build.xml
+++ b/src/release/installer/mac/build.xml
@@ -110,6 +110,10 @@
       <option value="-Xmx512m"/>
       <option value="-Dapple.laf.useScreenMenuBar=true"/>
 
+      <!-- Work around conflict with local JAI extensions installed in ~/Library/Java/Extensions -->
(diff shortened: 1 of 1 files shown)


src/release/installer/mac/build.xml
```diff
@@ -110,6 +110,10 @@
       <option value="-Xmx512m"/>
       <option value="-Dapple.laf.useScreenMenuBar=true"/>
 
+      <!-- Work around conflict with local JAI extensions installed in ~/Library/Java/Extensions -->
+      <!-- (system extension  MRJToolkit used by launcher to open the default browser) -->
+      <option value="-Djava.ext.dirs=${env.JAVA_HOME}/jre/lib/ext:/System/Library/Java/Extensions"/>
```
tbarsballe: I've had JAI show up in `/System/Library/Java/Extensions`, especially on older OS X machines that have been upgraded a few times, so this is a partial fix at best.

PR #2552, opened 2017-09-29
[GEOS-8230] - Fix login box not working on the demo request page

dvntucker:
So for whatever reason the PropertyModel for the password does not work correctly, I chalk it up to Wicket quirks. Similarly with the test I could not for the life of me to get the test to get the password correct. Tried a few different ways as well. It works for a text field, but not a password field. I chalk it up to wicket weirdness.

### DIFF
--- a/src/web/demo/src/main/java/org/geoserver/web/demo/DemoRequestResponse.java
+++ b/src/web/demo/src/main/java/org/geoserver/web/demo/DemoRequestResponse.java
@@ -11,6 +11,7 @@ import org.apache.wicket.markup.html.form.Form;
 import org.apache.wicket.markup.html.form.HiddenField;
 import org.apache.wicket.markup.html.form.TextArea;
 import org.apache.wicket.model.IModel;
+import org.apache.wicket.model.Model;
 import org.apache.wicket.model.PropertyModel;
 import org.geoserver.web.GeoServerBasePage;
 import org.vfny.geoserver.wfs.servlets.TestWfsPost;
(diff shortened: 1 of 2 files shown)


dvntucker:
I can't imagine there would be any security issues (beyond what already might be there). From the web side nothing has changed.

src/web/demo/src/main/java/org/geoserver/web/demo/DemoRequestResponse.java
```diff
@@ -51,7 +52,7 @@ public DemoRequestResponse(final IModel model) {
         form.add(new HiddenField("url", new PropertyModel(model, "requestUrl")));
         form.add(new TextArea("body", new PropertyModel(model, "requestBody")));
         form.add(new HiddenField("username", new PropertyModel(model, "userName")));
-        form.add(new HiddenField("password", new PropertyModel(model, "password")));
+        form.add(new HiddenField("password", new Model(((DemoRequest)model.getObject()).getPassword())));
```
aaime: Confused as to why username interpolation works, but password one needs to dodge the model instead?

dvntucker: I honestly don't know, and didn't dig into Wicket enough to find out.

dvntucker: >Note that by default the model object is nullified after each request to prevent the entered password to be serialized along with the containing page, see{@link#setResetPassword(boolean)}. Hmm, maybe something to do with that? I'll check.

Which... now that I think about it makes perfect sense, and also means that this page is broken as is and really should be fixed properly. This page sends the submitted password back to fill the modal window that pops up which we really shouldn't be doing. Wicket was trying to be helpful the whole time I guess.

PR #2556, opened 2017-10-01
[GEOS-8257] Correct Y order for geopackage WPS tiles output.

bradh:
GeoPackage first tile is top left (flipped from mbtiles). This is already handled for
Map output, just incorrect for WPS tiles.


src/community/geopkg/src/test/java/org/geoserver/geopkg/wps/gs/GeoPackageProcessTileTest.java
```diff
@@ -0,0 +1,88 @@
+package org.geoserver.geopkg.wps.gs;
```
jodygarnett: header needed

src/community/geopkg/src/main/java/org/geoserver/geopkg/wps/GeoPackageProcess.java
```diff
@@ -314,6 +314,7 @@ else if (tiles.getSldBody() != null) {
                }
                request.setFormat("none");
                Map formatOptions = new HashMap();
+               formatOptions.put("flipy", "true");
```
jodygarnett: I am wondering if we are fixing this issue (https://osgeo-org.atlassian.net/browse/GEOT-5582) and then running into trouble downstream here in geoserver

bradh: I don't think so. It isn't CRS dependent, more about how tile indices (in a TMS z/x/y sense) are used. GeoPackage does it one way, mbtiles the other.

bradh:
I redid this as #2570 after the usual joy of rebase and github.

bencaradocdavies:
@bradh AFAIK pull requests work by source branch name, so it should be possible to keep the same PR and use a force push onto your source branch.

PR #2557, opened 2017-10-01
[GEOS-8315] Show possible axis orders in SRS list demo page

aaime:


src/web/demo/src/main/resources/GeoServerApplication.properties
```diff
@@ -34,6 +34,10 @@ SRSDescriptionPage.name=Name
 SRSDescriptionPage.scope=Scope
 SRSDescriptionPage.remarks=Remarks
 SRSDescriptionPage.areaOfValidity=Area of validity
+SRSDescriptionPage.epsgOrder=EPSG description
+SRSDescriptionPage.epsgOrderDescription=This is the axis order by the official EPSG database, used by the latest OGC specifications. Geographic CRSs along with some projected ones are in North/East order.
```
bradh: Probably would read better as: "...order from the official EPSG database..." (i.e. "from" instead of "by"

src/web/demo/src/main/resources/GeoServerApplication.properties
```diff
@@ -34,6 +34,10 @@ SRSDescriptionPage.name=Name
...
+SRSDescriptionPage.internalOrder=Internal description
+SRSDescriptionPage.internalOrderDescription=This is the axis order used internally by GeoServer, early OGC specifications, and most vector data sources. All coordinates systems are forced in East/North order.
```
bradh: Perhaps "earlier" instead of "early". Also "forced in East/North" would be better as "forced to East/North" or "forced into East/North".

aaime:
@bradh feedback applied, merging

PR #2559, opened 2017-10-02
[GEOS-8307] RenderedImageMapOutputFormat won't try to render requests outside of the reader declared envelope

aaime:
Follow up on https://github.com/geotools/geotools/pull/1709 , actually needs those jars to be deployed to pass


aaime:
Well actually they are new to me... I will investigate

aaime:
Fixed (and merged)

PR #2564, opened 2017-10-04
[GEOS-8318] Updated virtual service global layer group look up to not load workspaced layer groups

sikeoka:
This pull request modifies LocalWorkspaceCallback and OWSHandlerMapping to only look up layer groups without a workspace.  It also delays looking up the global layer group until after checking if a workspace exists to reduce database queries with JDBC Configuration.


sikeoka:
The test failure seems to be unrelated to this pull request:
> testGetNoAuthHide(org.geoserver.wfs.SecuredGetFeatureTest)  Time elapsed: 19 sec  <<< ERROR!
org.xml.sax.SAXParseException; lineNumber: 1; columnNumber: 1; Premature end of file.

aaime:
Looks good, merging

PR #2568, opened 2017-10-05
[GEOS-8320] Added system property for Hazelcast cache expiration time

sikeoka:
Updated HzCacheProvider to support the same system property to control the cache eviction time that JDBCCacheProvider uses.  This is pretty much a copy and paste from:
https://github.com/geoserver/geoserver/blob/master/src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/JDBCCacheProvider.java


tbarsballe:

PR #2570, opened 2017-10-06
[GEOS-8257] Correct Y order for geopackage WPS tiles output.

bradh:
GeoPackage first tile is top left (flipped from mbtiles). This is already handled for
Map output, just incorrect for WPS tiles.

Andrea, Justin, Neils: Also adds missing headers for tests you wrote. Any objections?


jodygarnett:
You make squeek in under the deadline if we backport the fix

PR #2571, opened 2017-10-06
[GEOS-8249] CoverageView: relax constraint on bbox and resolution

aaime:


doc/en/user/source/data/raster/coverageview.rst
```diff
@@ -59,6 +59,31 @@ Once all the properties of the layer have been configured, by selecting the :gui
 .. figure:: images/coverageviewavailablelayers.png
    :align: center
 
+Heterogeenous coverage views
```
dromagnoli: small typo here:
should be heterogeneous

src/main/src/main/java/org/geoserver/catalog/CoverageViewReader.java
```diff
@@ -236,12 +151,58 @@ public CoverageViewReader(GridCoverage2DReader delegate, CoverageView coverageVi
...
+                        (int) range.getMinX() - HETEROGENEOUS_RASTER_GUTTER,
+                        (int) range.getMinY() - HETEROGENEOUS_RASTER_GUTTER,
+                        (int) range.getWidth() + HETEROGENEOUS_RASTER_GUTTER * 2,
+                        (int) range.getHeight() + HETEROGENEOUS_RASTER_GUTTER * 2);
+                GridGeometry2D expandedGG = new GridGeometry2D(expandedRange,
+                        requestedGridGeometry.getGridToCRS(),
+                        requestedGridGeometry.getCoordinateReferenceSystem());
+                value.setValue(expandedGG);
+            }
+        }
```
dromagnoli: small note:
there are 4 if(condition) here not respecting the template 
if (condition)

src/main/src/main/java/org/geoserver/catalog/CoverageViewReader.java
```diff
@@ -309,10 +276,35 @@ public GridCoverage2D read(GeneralParameterValue[] parameters) throws IllegalArg
...
-        if (inputCoverages.isEmpty()) {
+        if (nonNullCoverages == 0 || inputCoverages.isEmpty()) {
             return null;
         }
         
+        // some returned null?
+        if(nonNullCoverages < inputCoverages.size()) {
+            float width, height;
+            if(requestedGridGeometry != null) {
+                GridEnvelope2D range = requestedGridGeometry.getGridRange2D();
```
dromagnoli: 2 if(condition) without space in between

src/web/core/src/main/java/org/geoserver/web/data/layer/CoverageViewAbstractPage.java
```diff
@@ -160,7 +179,25 @@ public void onClick() {
...
+            view.setSelectedResolutionIndex(referenceCoverageIndex);
+        }
+        
+        return view;
+    }
+
+    private int getReferenceCoverageIndex() {
+        for (int i = 0; i < coverageEditor.currentOutputBands.size(); i++) {
+            CoverageBand band = coverageEditor.currentOutputBands.get(i);
+            if(this.resolutionReferenceCoverage.equals(band.getInputCoverageBands().get(0).getCoverageName())) {
```
dromagnoli: if(condition)

src/web/core/src/main/java/org/geoserver/web/data/layer/CoverageViewAbstractPage.java
```diff
@@ -160,7 +179,25 @@ public void onClick() {
 
 
     protected CoverageView buildCoverageView() throws IOException {
-        return new CoverageView(name, coverageEditor.currentOutputBands);
+        CoverageView view = new CoverageView(name, coverageEditor.currentOutputBands);
+        view.setEnvelopeCompositionType(envelopeCompositionType);
+        view.setSelectedResolution(selectedResolution);
+        if(resolutionReferenceCoverage != null) {
```
dromagnoli: if(condition)

src/web/core/src/main/java/org/geoserver/web/data/layer/CoverageViewEditor.java
```diff
@@ -40,23 +47,34 @@
     List<CoverageBand> currentOutputBands;
     ListMultipleChoice<String> coveragesChoice;
     CompositionType compositionType;
+    IModel<EnvelopeCompositionType> envelopeCompositionType;
+    IModel<SelectedResolution> selectedResolution;
+    IModel<String> resolutionReferenceCoverage;
+
+//    List<EnvelopeCompositionType> envelopeCompositionModes;
+//    private DropDownChoice<EnvelopeCompositionType> envelopeComposition;
```
dromagnoli: I think we can get rid of these commented pieces since they are created on the fly in the method below

src/main/src/main/java/org/geoserver/catalog/CoverageViewReader.java
aaime: Heh,there were a ton in this file, fixed them all.

PR #2574, opened 2017-10-06
[GEOS-8324] Avoid recursion in FeatureTypeInfoImpl.equals

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-8324

My solution has been to duplicate the behaviour from `AttributeTypeInfoImpl.equals()` inside of `FeatureTypeInfoImpl.equals()`, excluding the check for whether the FeatureTypeInfos are equal to avoid recursion. This is a bit brittle to API changes, but is at least a complete equality check.

An alternative solution would be to remove the equality check on FeatureTypeInfo from `AttributeTypeInfoImpl.equals()`

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/impl/FeatureTypeInfoImpl.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/FeatureTypeInfoImpl.java
@@ -8,6 +8,7 @@ package org.geoserver.catalog.impl;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.ListIterator;
 
 import org.geoserver.catalog.AttributeTypeInfo;
 import org.geoserver.catalog.Catalog;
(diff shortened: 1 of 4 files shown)


src/main/src/main/java/org/geoserver/catalog/impl/FeatureTypeInfoImpl.java
```diff
@@ -185,8 +186,73 @@ public boolean equals(Object obj) {
...
+                if (attr == null) {
+                    if (otherAttr != null)
+                        return false;
+                }
+
+                /*
+                 * Compare attributes, excluding the feature type to avoid recursion.
+                 * This should be updated whenever AttributeTypeInfoImpl.equals() changes.
+                 */
+                if (attr instanceof AttributeTypeInfoImpl && otherAttr instanceof AttributeTypeInfoImpl) {
```
aaime: One suggestion, given that AttributeTypeInfo backlink to FeatureTypeInfo is "structural", exposed even by that interface, does it make sense to have a specialized equals method in AttributeTypeInfo whose implementation is basically the code below? Would make things clearer for other potential implementors, avoid knowing about the Impl class, and keep the code comparing AttributeTypeInfo close to it.

tbarsballe: Thinking about it, is there ever a case where we would want to use one over the other? Anything I can think of required already knowing something about the natures of the objects being compared (e.g. FeatureType being null).

From a code cleanliness and maintainability standpoint, I agree that adding a second equals method would be better.

As an alternative, does simply removing the check on FeatureTypeInfo from `AttributeTypeInfoImpl.equals()` (moving the contents of this PR into that equals method instead) cause a problem anywhere? It does stop is from doing an exact interface comparison, but the interface already has a build-in recursive structure.

tbarsballe: Thinking further, I suppose adding a second equals to AttributeTypeInfo would make sense if `FeatureTypeInfoImpl.equals()` was the main thing that made use of it. So basically the same thing that is going on here, but actually supported by the interface.

In that case, yes, seems like a good idea. I can make that change to this PR.

aaime:
+1 
I did not merge since I guess you also want to backport, so I'll let you handle it all

tbarsballe:
Ok, will do.

PR #2575, opened 2017-10-08
[GEOS-8325] Allow requesting both OL3 and OL2 from the client side

aaime:


bradh:
Looks fine, led to thinking about whether we should also support OL4....

aaime:
I think it would be nice, just a matter of finding the developer time to do it. I would see it more of as an upgrade of the standard "openlayers" preview, like, what is OL3 providing that OL4 does not? If nothing, I'd just remove 3 and replace it with 4.

tbarsballe:
OL basically changed their approach to versioning with OL4, such that OL4 is a pretty clean upgrade  from OL3, with only minor API changes. For our purposes, OL4 should provide everything OL3 does.
So if we were to support OL4, we should just replace OL3.

src/wms/src/main/java/org/geoserver/wms/map/OpenLayers2MapOutputFormat.java
```diff
@@ -0,0 +1,54 @@
...
+        return OUTPUT_FORMATS;
+    }
+
+    @Override
+    public String getMimeType() {
+        return MIME_TYPE;
+    }
+
+    @Override
+    protected boolean isOL3Enabled(WMSMapContent mapContent) {
```
tbarsballe: This is a bit weird now.
Is the property to disable OL3 still in place, and if so, would that result in two openlayers map output formats that produce OL2 output?

A more consistent approach that still maintains backwards compatibility would be to have an ol2 format, and ol3 format and a default ol format that delegates to one of ol2 or ol3 based on the value of the parameter.

aaime: I can go for that, but there is a couple of gotchas
* The OL3 format is not supported on all browsers, currently the code automatically falls back on OL2. But does it make sense when someone is explicitly asking for 3?
* OL2 is the odd one out and needs a special name because it offers some unique advantages but say we start advertising OL3 and then upgrade to OL4... does that mean we also keep 3 and add a new explicit OL4 format? From your comment above it seems there would be no reason to keep OL3 around, but removing it would break backwards compatibility.

tbarsballe: * If someone explicitly asks for 3, but is on a browser that doesn't support it, some sort of failure message seems appropriate (probably just a standard ServiceException).
* With respect to OL3 vs Ol4, good point. I want to say we should count OL3 and OL4 as the same thing, but that doesn't seem especially viable here. Perhaps support both? The minified js files are pretty small relative to GeoServer.

aaime: Ok, so you're proposing to have one more subclass specific to OL3, and maybe make the base class a little bit more generic (so that adding OL4 would just be a matter of overriding a few template methods)?

tbarsballe: Yes, something like that seems the best option if we are going to have version-specific subclasses.

aaime:
@tbarsballe pull request updated following our discussion

jodygarnett:
I cannot see any reason to support OL2+OL3+OL4, supporting OL2+(Latest) seems appropriate.

tbarsballe:
@jodygarnett - As per the discussion above, this is mostly a semantic issue, since we need different version names / mime types for:
* Latest OL supported by your machine (this will delegate to one of the actual formats)
* OL2 
* OL3 and/or OL4 (Latest OL release). If we just support OL4, there is a weird gap in our supported formats.

jodygarnett:
It looks like OpenLayers has adopted the same kind of versioning as us, so OL4 is about to be replaced with OL5 shortly (and is retaining backwards compatibility a lot better).

I think we should just keep two speeds of open layers preview.

aaime:
So I already applied changes based on @tbarsballe feedback and now @jodygarnett wants them undone? Could you please reach to an agreement? Currently the pull request is blocked because of it...

tbarsballe:
@jodygarnett - My only concern with having only two versions of openlayers is sensible response types:
* `application/openlayers` - Newest version of OpenLayers that works (Could be 2 or 4 or something else)
* `application/openlayers2` - OpenLayers 2
* `apllication/???` - Latest OpenLayers (Currently 4).

If you've got something that makes sense, then just having two versions seems fine, but otherwise I fear it leads to more confusion.

aaime:
I'm merging this one, if @jodygarnett wants it to behave differently he can go down the same path I used, list discussion and a pull request.

PR #2576, opened 2017-10-10
[GEOS-8328] Updated GeoServerResourceLoader to stop setting its logging level.

sikeoka:
Also, changed the logger name to the correct package name.  This is a simple logging change and does not have any unit tests.


aaime:
+1
According to git blame this was introduced by @jodygarnett in a5bb6fbda38caee2b8f84ab71cedcd145e3d2e2e , seems quite likely to have been a leftover debugging aid.

PR #2577, opened 2017-10-10
Swagger REST API Examples - Datastores

tbarsballe:
POST examples are minimal required to be valid.
PUT examples are maximal (except for connection parameters, which are described inline), and exclude values which cannot be changed via REST, such as `name`, `workspace`, or the `dbtype` connection parameter.

The current examples just include stores that are included in GeoServer by default, not extensions or community modules.

Misc swagger errors were also fixed.

### DIFF
--- a/doc/en/api/1.0.0/datastores.yaml
+++ b/doc/en/api/1.0.0/datastores.yaml
@@ -153,12 +153,7 @@ paths:
           required: true
           description: The name of the data store to modify.
           type: string
-        - name: dataStoreBody
-          in: body
-          description: The updated data store definition.
-          required: true
-          schema:
-            $ref: "#/definitions/datastore"
+        - $ref: "#/parameters/dataStorePut"
       consumes:
         - application/xml
         - application/json
@@ -439,6 +434,7 @@ parameters:
             <name>nyc</name>
             <connectionParameters>
               <url>file:/path/to/nyc.shp</database>
+            </connectionParameters>
           </dataStore>
           ```
 
@@ -482,6 +478,7 @@ parameters:
             <name>nyc</name>
             <connectionParameters>
               <url>file:/path/to/directory</database>
+            </connectionParameters>
           </dataStore>
           ```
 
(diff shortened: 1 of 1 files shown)


jodygarnett:
First up this is amazing and it will be great to see it go live; too bad it is not in the 2.12.0 release :)

doc/en/api/1.0.0/datastores.yaml
```diff
@@ -290,22 +285,609 @@ paths:
...
+              }
+            }
+          }
+          ```
+
+        Connection Parameters:
+
+        | key | description | level | type | required | default |
+        | --- | ----------- | ----- | ---- | -------- | ------- |
+        | Primary key metadata table | The optional table containing primary key structure and sequence associations. Can be expressed as 'schema.name' or just 'name' | user | String | False | ` ` |
```
jodygarnett: Looks amazing, want to double check the use of spaces in the key:

Checking [JDBCDataStoreFactory](https://github.com/geotools/geotools/blob/master/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCDataStoreFactory.java#L131):

```
  /** Metadata table providing information about primary keys **/
    public static final Param PK_METADATA_TABLE = new Param("Primary key metadata table", String.class,
            "The optional table containing primary key structure and sequence associations. Can be expressed as 'schema.name' or just 'name'", false);
```

Yep, spaces.

doc/en/api/1.0.0/datastores.yaml
```diff
@@ -290,22 +285,609 @@ paths:
...
+
+    in: body
+    required: true
+    schema:
+      $ref: "#/definitions/datastore"
+  dataStorePut:
+    name: dataStoreBody
+    description: |
+      The updated data store definition.
+      For a PUT, only values which should be changed need to be included.
```
jodygarnett: idea: clarification that updating an individual connection parameter is not possible, the entire connection parameter block would need to be changed?

doc/en/api/1.0.0/datastores.yaml
```diff
@@ -290,22 +285,609 @@ paths:
...
+
+      The contents of the connection parameters will differ depending on the type of data store being added.
+
```
jodygarnett: Understandable duplication of connection parameter information.

doc/en/api/1.0.0/datastores.yaml
tbarsballe: I did not actually know that was the case, but verified in the code. I suppose that makes sense, as otherwise there is no way to remove connection parameters.

Will add clarification.

PR #2578, opened 2017-10-11
[GEOS-8334] Reduced the number of LocalWorkspaceCatalog database queries when using JDBC Configuration.

sikeoka:
No unit tests have been added or modified but the behavior from the caller's point of view should not be changed so the existing tests in LocalWorkspaceCatalogTest and LocalWorkspaceLayersTest should be sufficient.

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/impl/LocalWorkspaceCatalog.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/LocalWorkspaceCatalog.java
@@ -338,8 +338,8 @@ public class LocalWorkspaceCatalog extends AbstractCatalogDecorator implements C
 
         CloseableIterator<T> iterator = delegate.list(of, filter, offset, count, sortBy);
         if (iterator.hasNext() && useNameDequalifyingProxy()) {
-            return CloseableIteratorAdapter.transform(iterator,
-                obj -> NameDequalifyingProxy.create(obj, of));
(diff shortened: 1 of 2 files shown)


sikeoka:
FYI, reviewing this pull request does not require any specific knowledge of JDBC Configuration besides understanding that calling geoServer.getSettings() may require multiple database queries.  The method outputs should be completely unchanged by this pull request regardless of whether or not JDBC Configuration is enabled.

aaime:
Had a quick look in spare time but cannot figure out fit he code is actually equivalent (I have a hunch that it might not be for null values but might be wrong).

sikeoka:
I reviewed my code again and you are correct.  If delegate.list returns an iterator containing a null value, then this modified code will throw an NPE where the existing code wouldn't.  JDBCCatalogFacade actually applies a notNull() filter to the iterator in ConfigDatabase so it won't contain null values and DefaultCatalogFacade will throw an NPE if it tries to return an iterator with null values but I suppose that it's possible for one of the various Catalog delegates to cause null values in the iterator.  There are many cases that I've found where Catalog.list() is called in the GeoServer code that actually assumes that the iterator can never contain null values and will throw an NPE if there is one.

Since the documentation doesn't explicitly specify this, I will look into fixing it so there's no regression and adding some unit tests for the various null/notnull cases.

sikeoka:
The pull request and unit test has been updated.

aaime:
Yep, looks good, merging

PR #2579, opened 2017-10-11
Swagger REST API Examples - Workspaces and Namespaces

tbarsballe:


tbarsballe:
While styles and stores are necessary, since they have a complex PUT structure not readily representable by the standard swagger examples, we already have sufficient namespace and workspace examples via the model.

PR #2580, opened 2017-10-11
[GEOS-8335] WMS 1.3.0 GetCapabilities Optimizations For JDBC Configuration

sikeoka:
Updated WMS 1.3.0 GetCapabilities not reload the service configuration from the database multiple times per layer and layer group and to only load the layers and layer groups from the catalog once per request.

No unit tests have been added or modified but the behavior from the caller's point of view should not be changed so the numerous existing GetCapabilities tests in the org.geoserver.wms.wms_1_3 package should be sufficient.

This pull request should be able to be backported to 2.11.x and 2.12.x.  It does change the method signature of a protected method but that is in a nested private class.

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
+++ b/src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
@@ -89,6 +89,7 @@ import org.xml.sax.ContentHandler;
 import org.xml.sax.helpers.AttributesImpl;
 
 import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
 import com.vividsolutions.jts.geom.Envelope;
 import org.geoserver.wfs.json.JSONType;
 
@@ -191,6 +192,8 @@ public class Capabilities_1_3_0_Transformer extends TransformerBase {
         DimensionHelper dimensionHelper;
 
         private boolean skipping;
+
+        private WMSInfo serviceInfo;
         
         private LegendSample legendSample;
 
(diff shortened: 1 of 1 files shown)


sikeoka:
Any thoughts on this pull request?  In my environment running JDBC Configuration and Hazelcast Clustering, workspaced WMS 1.3.0 GetCapabilities with hundreds of multi-dimensional coverages appeared to be processed several times faster without having to constantly re-query the database for the WMS service configuration.

WCS 1.0, WCS 1.1, WCS 2.0, WFS 1.0, WFS 1.1, WFS 2.0 and WMS 1.1 already all use a local instance of their service configuration throughout the GetCapabilities request.

aaime:
Hi, today during the PSC meeting the topic came out of whether we should just drop JDBCConfig, since:
* mails related to it are not really getting answered
* pull requests related to it are sitting there not getting reviewed
We still have to see if there is some interested, but we need someone to champion the module
just enough to make it worth keeping around (I'm not talking about immediate answers, it's
a community/unsupported module, but at least some presence).

Would you be interested, maybe in coordination with others? (Niels Charlier is also showing interest in the module).

NielsCharlier:
This PR is useful to me and I am willing to review it.

sikeoka:
FYI, reviewing this pull request does not require any specific knowledge of JDBC Configuration besides understanding that calling wmsConfig.getServiceInfo() and catalog.list() may require many database queries.  The method outputs should be completely unchanged by this pull request regardless of whether or not JDBC Configuration is enabled.

aaime:
@NielsCharlier then please do so :)

aaime:
Looks ok to me. Merging. Won't take responsibility of backports, sorry (in other words, merging and then making sure that the builds does not break, or if a few weeks down the line a regression is discovered, since you are not core developers the blame would fall on me). If you want to get this on stable raise your participation to a level where you can care for these changes yourself.

NielsCharlier:
see https://github.com/geoserver/geoserver/pull/2714 for changes I mentioned

PR #2581, opened 2017-10-11
Swagger REST API Examples - Coverage Stores

tbarsballe:
Unlike data stores, all coverage stores have the same PUT/POST structure - the location of the data is determined solely by an URL.

Consequently, individual examples for each coverage type seem unnecessary (perhaps a second example for imagemosaic would be useful, but that seems to already be covered better in the ImageMosac extension docs).

### DIFF
--- a/doc/en/api/1.0.0/coveragestores.yaml
+++ b/doc/en/api/1.0.0/coveragestores.yaml
@@ -68,12 +68,7 @@ paths:
           required: true
           description: The name of the worskpace containing the coverage stores.
           type: string
-        - name: coverageStoreBody
-          description: The coverage store body information to upload.
-          in: body
-          required: true          
-          schema:
-            $ref: "#/definitions/CoverageStoreInfo"
+        - $ref: "#/parameters/CoverageStorePost"
       consumes:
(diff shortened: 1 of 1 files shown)


jodygarnett:
Thanks Torben, for some reason I thought there were more settings that could be configured, but perhaps those are all defined at the layer level.

I agree image moasic configuration is already handled - I am not sure if anyone has tried using the resource api to modify that configuration file remotely.

doc/en/api/1.0.0/coveragestores.yaml
```diff
@@ -359,6 +349,72 @@ paths:
...
+        ```
+        {
+          "coverageStore": {
+            "description": "A coverage store",
+            "enabled": "true",
+            "_default": "true",
+            "url": "file:/path/to/file.tiff"
+          }
+        }
+        ```
```
jodygarnett: This looks good covering the 90% case.

It can be left to the docs for individual formats to explore relative paths, path variables or external URLs.

tbarsballe:
I was surprised at how few settings coverage store had as well.

PR #2582, opened 2017-10-11
Swagger REST API Examples - Style PUT

tbarsballe:
Improve style docs with a PUT example showing legend creation.

### DIFF
--- a/doc/en/api/1.0.0/styles.yaml
+++ b/doc/en/api/1.0.0/styles.yaml
@@ -187,7 +187,7 @@ paths:
         Using PUT with any other format will modify the content of the style. You can also use PUT with a ZIP file to upload a SLD 1.0 (`application/vnd.ogc.sld+xml`) file and any associated icon files
 
       parameters:
-        - $ref: "#/parameters/StyleInfoPost"
+        - $ref: "#/parameters/StyleInfoPut"
         - name: style
           in: path
           required: true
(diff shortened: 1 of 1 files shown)


doc/en/api/1.0.0/styles.yaml
```diff
@@ -643,19 +643,121 @@ parameters:
...
+            }
+          }
+          ```
+
+      Otherwise, the style body is an actual style:
+
+        - application/zip: 
+        
+          ```
+          <ZIP file containing SLD and icons>
```
jodygarnett: Is it worth listing the directory structure of the zip as an example?

```
point_of_interest.sld
landmark.png
park.png
OpenSans.ttf
...
```

Icons can also be managed with the resource endpoint.

doc/en/api/1.0.0/styles.yaml
```diff
@@ -643,19 +643,121 @@ parameters:
...
+            <languageVersion>
+              <version>1.0.0</version>
+            </languageVersion>
+            <filename>roads.sld</filename>
+            <legend>
+              <width>32</width>
+              <height>32</height>
+              <format>image/png; charset=UTF-8</format>
+              <onlineResource>legend.png</onlineResource>
+            </legend>
```
jodygarnett: thanks for including optional legend example.

PR #2592, opened 2017-10-12
[GEOS-8332] - DataStoreFileController wrong management of FeatureTypes when dealing with DB-based DataStores

afabiani:


src/restconfig/src/main/java/org/geoserver/rest/catalog/DataStoreFileController.java
```diff
@@ -132,6 +161,31 @@ public static String lookupDataStoreFactoryFormat(String type) {
         return null;
     }
 
+    public static List<String> lookupUploadedFeatureTypeNames(List<Resource> files, String format) {
+        List<String> uploadedFeatureTypeNames = new ArrayList<>();
+        
+        for (Resource file : files) {
```
aaime: This won't work for uploads that are not shapefiles

PR #2593, opened 2017-10-12
[GEOS-8321] Monitoring plugin Path Exception

afabiani:


src/extension/monitor/core/src/main/java/org/geoserver/monitor/rest/RESTMonitorCallback.java
```diff
@@ -43,9 +46,14 @@ public void init(HttpServletRequest request, HttpServletResponse response) {
         
         data.setCategory(Category.REST);
         if (request.getPathInfo() != null) {
-            String resource = Paths.get(request.getPathInfo()).getFileName().toString();
-            resource = FilenameUtils.getBaseName(resource);
-            data.getResources().add(resource);
+        	try {
```
aaime: Needs a test (like any other bug fix).

afabiani: Added.

afabiani:
@aaime done.

aaime:
Looks ok. I was wondering about the "lenght - 1", but cannot think of a case in which the result of split would be zero length.

PR #2594, opened 2017-10-12
[GEOS-8338] Update Community Module 'sldExtension' to new Spring MVC

afabiani:


src/community/sldService/pom.xml
```diff
@@ -16,85 +16,103 @@
...
-   <artifactId>xmlunit</artifactId>
-   <scope>test</scope>
-  </dependency>
-  <dependency>
-     <groupId>org.mockito</groupId>
-     <artifactId>mockito-core</artifactId>
-     <scope>test</scope>
-  </dependency>    
- </dependencies>
+	<dependencies>
```
aaime: Mass reformat, using tabs instead of spaces.

src/community/sldService/src/main/java/applicationContext.xml
```diff
@@ -1,33 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- 
- Copyright (C) 2014 - Open Source Geospatial Foundation. All rights reserved.
- This code is licensed under the GPL 2.0 license, available at the root
- application directory.
- -->
-<!DOCTYPE beans PUBLIC "-//SPRING//DTD BEAN//EN" "http://www.springframework.org/dtd/spring-beans.dtd">
-<beans>
+<beans xmlns="http://www.springframework.org/schema/beans"
```
aaime: The copyright header should not be removed.

src/community/sldService/src/main/java/org/geoserver/sldservice/rest/ClassifierController.java
```diff
@@ -0,0 +1,443 @@
...
+import net.sf.json.JSONObject;
+import net.sf.json.xml.XMLSerializer;
+
+/**
+ * ClassifierController.
+ */
+@RestController
+@ControllerAdvice
+@RequestMapping(path = RestBaseController.ROOT_PATH + "/sldservice")
+public class ClassifierController extends AbstractCatalogController {
```
aaime: Formatted with tabs. Please check all files that have been basically written anew, as well as any new code inline with existing files, and make sure the formatting is correct. Being a community module and considering the REST port basically broke any attempt at backporting changes easily and track history of changes, a mass reformat of the whole module would not be a bad idea imho.

src/community/sldService/src/test/java/org/geoserver/sldservice/rest/RasterizerTest.java
```diff
@@ -0,0 +1,166 @@
...
+	        Document dom = getAsDOM(restPath, 200);
+	        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+	        // print(dom);
+	        print(dom, baos);
+			assertTrue(baos.toString().indexOf("<featureTypeStyles>")>0);
+			// checkColorMap(baos.toString(), 5);
+        }
+	}
+
+//	public void testRasterizeWithFeatureType() throws IOException {
```
aaime: Did the code coverage go down here? It's community module, stricly speaking it could have no test and be jolly good anyways, but seeing commented out tests is a potential sign of regressions (which again might be fine, a community module is unsupported in any and every way). Still... could you comment on this?

src/community/sldService/src/test/java/org/geoserver/sldservice/rest/ClassifierTestSupport.java
```diff
@@ -0,0 +1,273 @@
...
+
+	protected SimpleFeatureCollection pointCollection, lineCollection;
+
+	FilterFactory2 ff = CommonFactoryFinder.getFilterFactory2(GeoTools.getDefaultHints());
+
+	protected SimpleFeatureType dataType;
+
+	protected SimpleFeature[] testFeatures;
+
+//	public void testClassifyForFeatureDefault() throws IOException {
```
aaime: Please do not leave commented out code around. See also my other comment about test coverage.

src/community/sldService/src/main/java/org/geoserver/sldservice/rest/ClassifierController.java
afabiani: Done

src/community/sldService/pom.xml
afabiani: Done

src/community/sldService/src/test/java/org/geoserver/sldservice/rest/RasterizerTest.java
afabiani: Currently since this is a community module very old, and those methods other than "attributes" have not been used since a long time, I would prefer to open a specific PR for re-adding again tests to the module, since they must be rewritten from scratch basically.

src/community/sldService/pom.xml
```diff
@@ -5,8 +5,8 @@
  application directory.
  -->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
- <modelVersion>4.0.0</modelVersion>
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
aaime: Still tabs here :-)

src/community/sldService/pom.xml
```diff
@@ -5,8 +5,8 @@
...
+	<modelVersion>4.0.0</modelVersion>
```
aaime: And here

src/community/sldService/pom.xml
aaime: And here too, still using tabs

PR #2595, opened 2017-10-12
[GEOS-8336][GEOS-8337] - Monitoring Hibernate obsolete dependencies / Monitoring Hibernate properties causing issues with the latest spring mvc changes

afabiani:


src/extension/monitor/hibernate/pom.xml
```diff
@@ -11,50 +11,71 @@
...
-   <scope>test</scope>
-  </dependency>
+	</dependencies>
 
- </dependencies>
-
- <properties>
-   <test.exclude.pattern>**/HibernateMonitorDAOTest.*</test.exclude.pattern>
- </properties>
+	<properties>
```
aaime: This exclusion can be removed, there is a single test class and it has a different name: HibernateMonitorDAO2Test

src/extension/monitor/hibernate/pom.xml
```diff
@@ -11,50 +11,71 @@
...
+	<groupId>org.geoserver.extension</groupId>
+	<artifactId>gs-monitor-hibernate</artifactId>
+	<packaging>jar</packaging>
+	<name>Monitor Hibernate Extension</name>
 
- <groupId>org.geoserver.extension</groupId>
- <artifactId>gs-monitor-hibernate</artifactId>
- <packaging>jar</packaging>
- <name>Monitor Hibernate Extension</name>
+	<dependencies>
```
aaime: Files should be indented with spaces (as the original version was)

src/extension/monitor/hibernate/src/main/resources/org/geoserver/monitor/hib/db.properties
```diff
@@ -20,3 +20,5 @@ url=jdbc:h2:file:${GEOSERVER_DATA_DIR}/monitoring/monitoring
 #url=jdbc:oracle:thin:@[host]:[port]\:[instance]
 #username=[username]
 #password=[password]
+
+gwc.context.suffix=gwc
```
aaime: For this one I'd rather fix the source issue (more on this later, let me have a look)

src/extension/monitor/hibernate/src/main/resources/applicationContext-hib2.xml
```diff
@@ -1,27 +1,36 @@
...
-        </props>
-      </property>
+   <bean id="monitoring-hibPropertyConfigurer" class="org.geoserver.config.GeoServerPropertyConfigurer">
+
+        <property name="order" value="100"/>
+        
+        <constructor-arg ref="dataDirectory"/>
+
+        <property name="ignoreResourceNotFound" value="true" />
+        <property name="ignoreUnresolvablePlaceholders" value="true" />
```
aaime: I was thinking exactly of these properties to solve the GWC property configurer nuke... did you already try it by any chance?

aaime:
Hi Alessio, I've tested the change and I got it working (full startup) without any other changes, and my property config files do not contain the gwc property. 
I am wondering if I just go lucky with classpath, or does it match your experience?

afabiani:
@aaime deps here are quite old and for sure conflicting with ones from other plugins, like GeoFence and GeoFence Server.

src/extension/monitor/hibernate/src/main/resources/org/geoserver/monitor/hib/db.properties
aaime: Can we remove this property? Should still work

PR #2598, opened 2017-10-13
Update antrun plugin in docs to use same version of ant as the root pom

smithkm:
Update antrun plugin in docs to use same version of ant as the root pom.  Docs is not a submodule of root so `${ant.version}` is not available.

### DIFF
--- a/doc/en/pom.xml
+++ b/doc/en/pom.xml
@@ -345,14 +345,20 @@
         <artifactId>maven-antrun-plugin</artifactId>
         <dependencies>
           <dependency>
-            <groupId>ant</groupId>
+            <groupId>org.apache.ant</groupId>
             <artifactId>ant</artifactId>
-            <version>1.6.5</version>
+            <version>1.8.4</version>
           </dependency>
           <dependency>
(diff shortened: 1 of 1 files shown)


tbarsballe:
A similar change caused problems on certain systems for GeoTools. Is this likely to cause similar issues for some people?

aaime:
@smithkm seems like we're stuck on @tbarsballe question not being answered

aaime:
@tbarsballe what should we do about this pull request?

tbarsballe:
Good question. Not sure; it would be nice to hear a bit more about the GeoTools issues. As-is, I would hesitate to merge until we have a bit more info.
@smithkm - do you have any comment on whether this has the potential to cause problems for different build systems?

Alternatively, we could just merge it and wait and see if any developers complain, and potentially revery if necessary?

smithkm:
Unfortunately I don't remember why I made this PR. I think it had to do with a problem with the Boundless Server build that I think we worked around some other way.

tbarsballe:
Given that, I think we should just close it.

PR #2602, opened 2017-10-17
Updating Script REST API from Restlet to Spring MVC enabling this mod…

vickdw:
…ule to work with GeoServer's REST API. [GEOS-8345]


src/community/script/core/src/main/java/org/geoserver/script/rest/controller/SessionController.java
```diff
@@ -0,0 +1,86 @@
...
+        sessionService.getSessions(response, null);
+    }
+
+    /**
+     * Get a list of Scripting Sessions
+     * @return
+     */
+    @GetMapping(path = "/sessions/{language}", produces = {MediaType.APPLICATION_JSON_VALUE,
+            MediaType.TEXT_PLAIN_VALUE})
+    public void getList(HttpServletResponse response, @PathVariable String language) {
```
tbarsballe: Instead of having two methods that do effectively the same thing, you can do the following:

```
    @GetMapping(path = {"/sessions", "/sessions/{language}"}, produces = {MediaType.APPLICATION_JSON_VALUE,
            MediaType.TEXT_PLAIN_VALUE})
    public void getList(HttpServletResponse response, @PathVariable(required = false) String language) {
```

This services both endpoints, if the former is used language will be null.

src/community/script/core/src/main/java/org/geoserver/script/rest/converter/ScriptConverter.java
```diff
@@ -0,0 +1,172 @@
...
+
+        @Override
+        protected XStream createXStreamInstance() {
+            return new SecureXStream();
+        }
+
+        @Override
+        public void encodeLink(String link, HierarchicalStreamWriter writer) {
+            String path = request.getRequestURI();
+            if (path.contains("script")) {
```
tbarsballe: Need to do a stricter check then that. What if there is a layer or store named "script" for some reason? Then doing a GET for that layer may use this converter.

src/community/script/core/src/main/java/org/geoserver/script/rest/converter/ScriptConverter.java
```diff
@@ -0,0 +1,172 @@
...
+                    link = URLDecoder.decode(link, "UTF-8");
+                    encodeAlternateAtomLinkNoExt(link, writer);
+                    writer.addAttribute("type", getMediaType());
+
+                    writer.endNode();
+                } catch (UnsupportedEncodingException e) {
+
+                }
+            } else {
+                encodeAlternateAtomLink(link, writer);
```
tbarsballe: Should this be delegating to a different method than `encodeLink`, above?

src/community/script/core/src/main/java/org/geoserver/script/rest/converter/ScriptConverter.java
```diff
@@ -0,0 +1,172 @@
...
+            }
+        }
+
+        @Override
+        public void encodeCollectionLink(String link, HierarchicalStreamWriter writer) {
+            String path = request.getRequestURI();
+            if (path.contains("script")) {
```
tbarsballe: Need a stricter check than this.

src/community/script/core/src/main/java/org/geoserver/script/rest/converter/ScriptConverter.java
```diff
@@ -0,0 +1,172 @@
...
+     */
+    @Component
+    public static class JSONXStreamScriptListConverter extends JSONXStreamListConverter {
+        @Autowired
+        HttpServletRequest request;
+
+        @Override
+        public void encodeLink(String link, HierarchicalStreamWriter writer) {
+            String path = request.getRequestURI();
+            if (path.contains("script")) {
```
tbarsballe: Stricter check. Given that this is in 3 places, is there a better method that we can override that would only require a single path check?

src/community/script/core/src/main/java/org/geoserver/script/rest/service/AppService.java
```diff
@@ -0,0 +1,107 @@
...
+
+@Service
+public class AppService {
+    static Logger LOGGER = Logging.getLogger(AppService.class);
+
+    @Autowired
+    ScriptManager scriptManager;
+
+    public ResponseEntity<?> getAppList() {
+        org.geoserver.platform.resource.Resource appRoot = scriptManager.app();
```
tbarsballe: I don't think you need the full package anymore, since we aren't importing the Restlet Resource anymore.

src/community/script/core/src/main/java/org/geoserver/script/rest/service/AppService.java
```diff
@@ -0,0 +1,107 @@
...
+
+            ScriptEngine eng = scriptManager.createNewEngine(script);
+            if (eng == null) {
+                throw new RestException(format("Script engine for %s not found", script.name()), HttpStatus.BAD_REQUEST);
+            }
+
+            //look up the app hook
+            AppHook hook = scriptManager.lookupAppHook(script);
+            if (hook == null) {
+                //TODO: fall back on default
```
tbarsballe: Has this TODO been resolved?

src/community/script/core/src/main/java/org/geoserver/script/rest/service/AppService.java
```diff
@@ -0,0 +1,107 @@
...
+                throw new RestException(format("No hook found for %s", script.path()),
+                        HttpStatus.INTERNAL_SERVER_ERROR);
+            }
+
+            Reader in = new BufferedReader(new InputStreamReader(script.in()));
+            try {
+                eng.eval(in);
+                hook.run(request, response, eng);
+            }
+            finally {
```
tbarsballe: Minor style issue - finally should be on the same line as the closing brace of the try

src/community/script/core/src/main/java/org/geoserver/script/rest/service/AppService.java
```diff
@@ -0,0 +1,107 @@
...
+                in.close();
+            }
+
+        }
+        catch (Exception e) {
```
tbarsballe: Same here, catch should be on the same line as the closing brace

src/community/script/core/src/main/java/org/geoserver/script/rest/service/ScriptService.java
```diff
@@ -0,0 +1,214 @@
...
+    ScriptManager scriptMgr;
+
+    String path;
+
+    public List<Script> getScriptList(HttpServletRequest request) {
+        final String type = (String) request.getAttribute("type");
+        this.path = getPath(request);
+        this.path = stripExtension(this.path);
+
+        org.geoserver.platform.resource.Resource dir = scriptMgr.script(path);
```
tbarsballe: Don't think you need the full package name anymore?

src/community/script/core/src/main/java/org/geoserver/script/rest/service/ScriptService.java
```diff
@@ -0,0 +1,214 @@
...
+                } else {
+                    String name = f.name();
+                    scripts.add(new Script(name));
+                }
+            }
+        } else {
+            // return empty array, perhaps we should return a 404?
+            // throw new RestletException(format("Could not find script dir %s",
+            // path),
+            // Status.CLIENT_ERROR_NOT_FOUND);
```
tbarsballe: Why is this commented out? Either fix and uncomment, or remove the else block.

src/community/script/core/src/main/java/org/geoserver/script/rest/service/ScriptService.java
```diff
@@ -0,0 +1,214 @@
...
+                    HttpStatus.NOT_FOUND);
+        }
+
+        String fileContents;
+
+        try {
+            fileContents = FileUtils.readFileToString(script.file());
+            response.getWriter().write(fileContents);
+            response.setStatus(200);
+//            return new ResponseEntity<String>(fileContents, HttpStatus.OK);
```
tbarsballe: Again, why is this commented out? If it is not necessary, remove it.

src/community/script/core/src/main/java/org/geoserver/script/rest/service/ScriptService.java
```diff
@@ -0,0 +1,214 @@
...
+        } catch (IOException e) {
+//            return new ResponseEntity<String>(HttpStatus.INTERNAL_SERVER_ERROR);
+            response.setStatus(500);
+        }
+    }
+
+    public void doPut(HttpServletRequest request, HttpServletResponse response) {
+        this.path = getPath(request);
+
+        org.geoserver.platform.resource.Resource script;
```
tbarsballe: Probably don't need package name here

src/community/script/core/src/main/java/org/geoserver/script/rest/service/SessionService.java
```diff
@@ -0,0 +1,143 @@
...
+public class SessionService {
+    static Logger LOGGER = Logger.getLogger("org.geoserver.script.rest");
+
+    @Autowired
+    ScriptManager scriptMgr;
+
+    /**
+     * Get a list of Scripting Sessions
+     * @return
+     */
```
tbarsballe: This javadoc is kind of useless; either improve it or remove it. Perhaps just add a top-level description to the class?

src/community/script/core/src/test/java/org/geoserver/script/app/AppTest.java
```diff
@@ -33,7 +32,7 @@ public void testSimple() throws Exception {
         FileUtils.copyURLToFile(
             getClass().getResource("main-helloWorld."+ext), new File(app, "main."+ext));
     
-        MockHttpServletResponse resp = getAsServletResponse("/script/apps/foo/main."+ext);
+        MockHttpServletResponse resp = getAsServletResponse("/rest/apps/foo/main."+ext);
```
tbarsballe: I think this should be ``/rest/script/apps``, in which case the app controller is also wrong. Did you test in an actual running geoserver with script installed (from before the REST update) before changing this?

src/community/script/core/src/test/java/org/geoserver/script/rest/SessionTest.java
```diff
@@ -7,16 +7,13 @@
...
 import org.geoserver.script.ScriptIntTestSupport;
-import org.geoserver.test.GeoServerTestSupport;
-
 import org.springframework.mock.web.MockHttpServletResponse;
 
 public class SessionTest extends ScriptIntTestSupport {
 
     public void testPost() throws Exception {
-        MockHttpServletResponse response = postAsServletResponse("/script/sessions/js", "");
+        MockHttpServletResponse response = postAsServletResponse("/rest/sessions/js", "");
```
tbarsballe: Likewise, this should probably be `/rest/script/sessions`, and the sessions controller should be updated.

The restlet framework handled servlet paths differently from the MVC framework I believe.
Either way, you should verify in a 2.11 geoserver with script installed to be sure of the correct path.

src/community/script/js/src/main/java/org/geoserver/script/js/JsgiRequest.java
```diff
@@ -36,41 +30,41 @@
...
+//            LOGGER.warning(e.getMessage());
+//        }
+
+//        Scriptable headers = cx.newObject(scope);
+//        Form requestHeaders = (Form) request.getAttributes().get("org.restlet.http.headers");
+//        for (String name : requestHeaders.getNames()) {
+//            String value = requestHeaders.getFirstValue(name, true);
+//            headers.put(name.toLowerCase(), headers, value);
+//        }
+//        this.put("headers", this, headers);
```
tbarsballe: Why is this commented out? If it is unnecessary, remove it.

src/community/script/js/src/main/java/org/geoserver/script/js/JsgiResponse.java
```diff
@@ -84,55 +74,55 @@ public JsgiResponse(Scriptable obj, Scriptable scope) {
...
+//        }
+//        if (headers != null) {
+//            for (Object id : headers.getIds()) {
+//                String name = id.toString();
+//                String value = headers.get(name, headers).toString();
+//                if (!name.equalsIgnoreCase("content-type")) {
+//                    responseHeaders.add(name, value);
+//                }
+//            }
+//        }
```
tbarsballe: Again, why is this commented out?

src/community/script/js/src/main/java/org/geoserver/script/js/JsgiResponse.java
```diff
@@ -84,55 +74,55 @@ public JsgiResponse(Scriptable obj, Scriptable scope) {
...
-            mediaType = MediaType.TEXT_PLAIN;
-        } else {
-            mediaType = new MediaType(type);
-        }
+//        String type = responseHeaders.getFirstValue("content-type", true);
+//        if (type == null) {
+//            mediaType = MediaType.TEXT_PLAIN;
+//        } else {
+//            mediaType = new MediaType(type);
+//        }
```
tbarsballe: Again, why is this commented out?

src/community/script/js/src/main/java/org/geoserver/script/js/JsgiResponse.java
```diff
@@ -84,55 +74,55 @@ public JsgiResponse(Scriptable obj, Scriptable scope) {
...
+//                Object[] args = {boundWrite};
+//                try {
+//                    forEach.call(cx, scope, body, args);
+//                } finally {
+//                    Context.exit();
+//                    outputStream.close();
+//                }
+//            }
+//
+//        });
```
tbarsballe: Again, why is this commented out?

src/community/script/py/src/main/java/org/geoserver/script/py/PyAppHook.java
```diff
@@ -239,30 +201,30 @@ public void write(OutputStream outputStream) throws IOException {
...
+//        seg.add(0, "");
+//
+//        environ.put("PATH_INFO", StringUtils.join(seg.toArray(), "/"));
+//        //environ.put("PATH_INFO", );
+//
+//        environ.put("QUERY_STRING", request.getResourceRef().getQuery());
+//
+//        environ.put("wsgi.version", new PyTuple(new PyInteger(0), new PyInteger(1)));
+//        environ.put("wsgi.url_scheme", ref.getScheme());
+//        environ.put("wsgi.input", new PyFile(request.getEntity().getStream()));
```
tbarsballe: Why is this commented out? If it is unnecessary, remove it.

src/community/script/py/src/test/java/org/geoserver/script/py/PyAppTest.java
```diff
@@ -30,7 +26,7 @@ public void testSimple() throws Exception {
         FileUtils.copyURLToFile(
             getClass().getResource("main-helloWorld.py"), new File(app, "main.py"));
 
-        MockHttpServletResponse resp = getAsServletResponse("/script/apps/foo/main.py");
+        MockHttpServletResponse resp = getAsServletResponse("rest/apps/foo/main.py");
```
tbarsballe: See above, this path is probably wrong

src/community/script/core/src/main/java/org/geoserver/script/rest/service/AppService.java
vickdw: The TODO is an artifact from the original version, not sure what default was in this context.

src/community/script/core/src/main/java/org/geoserver/script/rest/service/ScriptService.java
vickdw: fixed.

src/community/script/core/src/main/java/org/geoserver/script/rest/service/ScriptService.java
vickdw: fixed.

src/community/script/core/src/main/java/org/geoserver/script/rest/service/ScriptService.java
vickdw: fixed

src/community/script/core/src/main/java/org/geoserver/script/rest/service/SessionService.java
vickdw: removed.

src/community/script/core/src/test/java/org/geoserver/script/app/AppTest.java
vickdw: Yes. I tested with 2.9.x which is why this needed to be changed.  If not it interferes with the Script Controller at the same endpoint.

src/community/script/core/src/test/java/org/geoserver/script/rest/SessionTest.java
vickdw: I tested in 2.9.x, will test with 2.11 if you think that would be a better test.

tbarsballe: 2.9 is fine.

src/community/script/core/src/test/java/org/geoserver/script/app/AppTest.java
tbarsballe: To be clear, what was the endpoint this used in 2.9?

src/community/script/js/src/main/java/org/geoserver/script/js/JsgiResponse.java
vickdw: fixed

src/community/script/js/src/main/java/org/geoserver/script/js/JsgiRequest.java
vickdw: fixed.

src/community/script/js/src/main/java/org/geoserver/script/js/JsgiResponse.java
vickdw: fixed

src/community/script/js/src/main/java/org/geoserver/script/js/JsgiResponse.java
vickdw: fixed

src/community/script/py/src/test/java/org/geoserver/script/py/PyAppTest.java
vickdw: See response to above.

src/community/script/py/src/main/java/org/geoserver/script/py/PyAppHook.java
vickdw: fixed

src/community/script/js/src/main/java/org/geoserver/script/js/JsgiResponse.java
vickdw: fixed

src/community/script/js/src/main/java/org/geoserver/script/js/JsgiResponse.java
vickdw: fixed

src/community/script/js/src/main/java/org/geoserver/script/js/JsgiResponse.java
vickdw: fixed

src/community/script/core/src/main/java/org/geoserver/script/rest/controller/SessionController.java
vickdw: Changed.

src/community/script/core/src/main/java/org/geoserver/script/rest/converter/ScriptConverter.java
vickdw: Created the following method
```
protected static boolean checkPath(HttpServletRequest request) {
        String path = request.getRequestURI();
        if (path.contains("script/apps") || (path.contains("script/function"))
                || (path.contains("script/wfs/tx")) || (path.contains("script/wps"))) {
            return true;
        } else {
            return false;
        }
    }
```

src/community/script/core/src/main/java/org/geoserver/script/rest/converter/ScriptConverter.java
vickdw: See reply above.

src/community/script/core/src/main/java/org/geoserver/script/rest/converter/ScriptConverter.java
vickdw: No it uses the un-overridden method defined by the parent.

src/community/script/core/src/main/java/org/geoserver/script/rest/converter/ScriptConverter.java
vickdw: See comment containing new method above to facilitate this check.

src/community/script/core/src/main/java/org/geoserver/script/rest/service/AppService.java
vickdw: fixed.

src/community/script/core/src/main/java/org/geoserver/script/rest/service/AppService.java
vickdw: This is an artifact from the original code.  Not sure what was meant here.  Can remove TODO if you think that is better.

src/community/script/core/src/main/java/org/geoserver/script/rest/service/AppService.java
vickdw: fixed.

src/community/script/core/src/main/java/org/geoserver/script/rest/service/AppService.java
vickdw: fixed.

src/community/script/core/src/main/java/org/geoserver/script/rest/service/ScriptService.java
vickdw: fixed.

src/community/script/core/src/main/java/org/geoserver/script/rest/service/ScriptService.java
vickdw: fixed.

src/community/script/core/src/main/java/org/geoserver/script/rest/service/ScriptService.java
vickdw: fixed.

src/community/script/core/src/main/java/org/geoserver/script/rest/service/ScriptService.java
vickdw: fixed

src/community/script/core/src/main/java/org/geoserver/script/rest/service/SessionService.java
vickdw: fixed.

src/community/script/core/src/test/java/org/geoserver/script/app/AppTest.java
vickdw: Tested this with 2.9.x and needed to use this path to avoid conflicting with the Script endpoints.

src/community/script/core/src/test/java/org/geoserver/script/rest/SessionTest.java
vickdw: Tested with 2.9.x and this is the expected behavior and the correct path.

src/community/script/py/src/test/java/org/geoserver/script/py/PyAppTest.java
vickdw: See above comments regarding path.

src/community/script/core/src/test/java/org/geoserver/script/app/AppTest.java
tbarsballe: If this is a change from 2.9/2.10/2.11 then it is an unacceptable API break. What is the actual issue that required you to change this path, in a bit more detail?

src/community/script/core/src/main/java/org/geoserver/script/rest/service/AppService.java
tbarsballe: If it is left over from the original code, I guess it is safer to keep it. Probably won't make much difference either way.

tbarsballe:
One other note - when squashing, you should retain your original commit message, and add any additional info in the description. As-is "code cleanup after review" doesn't make much sense as the only commit message (The actual commit message appears to be in the description).

src/community/script/core/src/test/java/org/geoserver/script/rest/SessionTest.java
vickdw: From the Restlet versions applicationContext, the mapping for SessionFinder, which is now SessionController is as follows.
```
<!-- Scripting Sessions -->
        <entry>
          <key><value>/sessions</value></key>
          <value>sessionFinder</value>
        </entry>
        <entry>
          <key><value>/sessions/{ext}</value></key>
          <value>sessionFinder</value>
        </entry>
        <entry>
          <key><value>/sessions/{ext}/{session}</value></key>
          <value>sessionFinder</value>
        </entry>
```
Therefore this is the correct path.

src/community/script/core/src/test/java/org/geoserver/script/app/AppTest.java
vickdw: AppFinder has become AppController and this was the path in the Restlet version.
```
<!-- Scripting Apps -->
        <entry>
            <key><value>/apps/{app}</value></key>
            <value>appFinder</value>
        </entry>
        <entry>
            <key><value>/apps/{app}/</value></key>
            <value>appFinder</value>
        </entry>
```

src/community/script/core/src/test/java/org/geoserver/script/rest/SessionTest.java
tbarsballe: After some investigation, it turns out the scripts extension proxies the REST api at /script, such that both `"/script/sessions/js"` and `"/rest/sessions/js"` are valid paths:

```
<bean id="scriptDispatcherMapping" 
        class="org.springframework.web.servlet.handler.SimpleUrlHandlerMapping">
    <property name="alwaysUseFullPath" value="true"/>
    <property name="mappings">
        <props>
            <prop key="/script/**">restWrapper</prop>
        </props>
    </property>
  </bean>
```

All the controllers will need to be updated to support both paths.

src/community/script/core/src/main/java/org/geoserver/script/rest/converter/ScriptConverter.java
```diff
@@ -0,0 +1,179 @@
...
+        @Override
+        public void encodeLink(String link, HierarchicalStreamWriter writer) {
+            if (checkPath(request)) {
+                try {
+                    link = URLDecoder.decode(link, "UTF-8");
+                    writer.startNode("href");
+                    writer.setValue(hrefNoExt(link));
+                    writer.endNode();
+                } catch (UnsupportedEncodingException e) {
+
```
tbarsballe: You should at least log a WARNING or DEBUG message here, swallowing exceptions makes debugging unnecessarily difficult

src/community/script/core/src/main/java/org/geoserver/script/rest/service/SessionService.java
```diff
@@ -0,0 +1,139 @@
...
+                if (LOGGER.isLoggable(Level.FINE)) {
+                    LOGGER.log(Level.FINE, "Script error", e);
+                }
+            }
+            w.flush();
+        }
+        catch(IOException e) {
+            throw new RestException("i/o error", HttpStatus.INTERNAL_SERVER_ERROR, e);
+        }
+        try {
```
tbarsballe: Some more dangling braces

src/community/script/py/pom.xml
```diff
@@ -78,6 +78,10 @@
    <version>${project.version}</version>
    <scope>test</scope>
   </dependency>
+   <dependency>
+     <groupId>org.springframework</groupId>
+     <artifactId>spring-web</artifactId>
+   </dependency>
```
tbarsballe: Another new dependency. Any idea why this got added?

src/community/script/core/pom.xml
```diff
@@ -31,6 +31,11 @@
      <artifactId>cglib-nodep</artifactId>
      <scope>test</scope>
    </dependency>
+   <dependency>
```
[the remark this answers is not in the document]
vickdw: removed this and is now using net-sf-json from GeoServer.

tbarsballe:
Looks like there are some build failures caused by this change:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:2.3.2:compile (default-compile) on project gs-xslt: Compilation failure: Compilation failure: 
[ERROR] /home/travis/build/geoserver/geoserver/src/extension/xslt/src/main/java/org/geoserver/wfs/xslt/rest/TransformationsJsonConverter.java:[17,4] error: method does not override or implement a method from a supertype
[ERROR] /home/travis/build/geoserver/geoserver/src/extension/xslt/src/main/java/org/geoserver/wfs/xslt/rest/TransformationsXmlConverter.java:[17,4] error: method does not override or implement a method from a supertype
```

Probably caused by the change to the XmlConverter. Should be an easy fix.

tbarsballe:
Merged, and backported to 2.12.x.

PR #2605, opened 2017-10-18
[GEOS-8353] Ensure KML validates without being online.

bradh:
If not online, this happens:

```
Tests in error: 
  testValidKML(org.geoserver.kml.KMLReflectorTest): src-resolve: Cannot resolve the name 'xal:AddressDetails' to a(n) 'element declaration' component.

```
This adds the xAL.xsd file. The alternative would be to change this Reflector test to be an Online one.


aaime:
Looks good to me, merging.

aaime:
Backported to 2.12.x too

PR #2608, opened 2017-10-22
Minor fix in OSEO REST Doc

simboss:


doc/en/api/1.0.0/opensearch-eo.yaml
```diff
@@ -848,7 +848,7 @@ paths:
         404:
           description: The specified collection or product cannot be found
     put:
-      description: Updates/creates a products granules
+      description: Updates/creates a product granules
```
bradh: This probably should be "product's granules", possessive form because the granules "belong to" the product.

aaime:
Will backport manually.

PR #2611, opened 2017-10-24
[GEOS-8364] Fixed "Failed to resolve workspace" messages for non-workspaced styles.

sikeoka:
This modifies the fix for https://osgeo-org.atlassian.net/browse/GEOS-8289 to only apply to styles that have a workspace.

This pull request can be backported to 2.12.x only.


aaime:
Looks good to me, merging

PR #2613, opened 2017-10-28
Repair 2.12.x after someone forced pushed

bencaradocdavies:
See mailing list discussion:

[Geoserver-devel] Who force pushed to 2.12.x?
http://osgeo-org.1560.x6.nabble.com/Who-force-pushed-to-2-12-x-td5340313.html


bencaradocdavies:
Thanks @tbarsballe for the review.

PR #2618, opened 2017-10-31
[GEOS-8371] Querying the granules of a StructuredCoverageViewReader fails with a 'coverage not known' exception

aaime:


src/main/src/main/java/org/geoserver/feature/CompositeFeatureCollection.java
```diff
@@ -14,9 +14,12 @@
 import org.geotools.data.store.DataFeatureCollection;
 import org.geotools.feature.FeatureCollection;
 import org.geotools.feature.FeatureIterator;
+import org.geotools.feature.visitor.FeatureCalc;
 import org.geotools.geometry.jts.ReferencedEnvelope;
+import org.opengis.feature.FeatureVisitor;
 import org.opengis.feature.simple.SimpleFeatureType;
 import org.opengis.filter.identity.FeatureId;
+import org.opengis.util.ProgressListener;
```
dromagnoli: Are these imports needed?

aaime: Nope, changes to this file are leftovers. Removed!

PR #2619, opened 2017-11-02
[GEOS-6745] Copy style needs to copy content and format

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-6745 seems to have reoccured after the Style Editor update in 2.11.x

I have fixed it. I attempted to construct a viable test case, but was unsuccessful (given this functionality involves Wicket + Ajax, this is unsurprising).


tbarsballe:
Backported to 2.11 and 2.12

PR #2622, opened 2017-11-03
[GEOS-8225] Allow workspace to be set on new styles if not admin

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-8225

This looks like it was another issue introduced by the style page rewrite; prior to that the workspace drop-down was disabled only when editing styles, not when creating new ones.
This behaviour has been restored.


tbarsballe:
Backported to 2.11 and 2.12

PR #836, opened 2014-11-28
Various fixes for GeoServer GUI + JDBCConfig

n-lagomarsini:
This pull request contains various fixes for the GeoServer performances using JDBCConfig plugin.

PR #2623, opened 2017-11-03
[GEOS-8242] Use separate keys when caching different queries

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-8242

The size of the results when there was no filter was being cached, and that value was getting used in place of the actual number of results returned by the filter.

An alternative fix would be to just not cache the size value.

This behaviour seems to have been introduced by https://github.com/geoserver/geoserver/pull/836

### DIFF
--- a/src/web/demo/src/main/java/org/geoserver/web/demo/PreviewLayerProvider.java
+++ b/src/web/demo/src/main/java/org/geoserver/web/demo/PreviewLayerProvider.java
@@ -129,6 +129,10 @@ public class PreviewLayerProvider extends GeoServerDataProvider<PreviewLayer> {
     @Override
     public long size() {
         try {
+            if (getKeywords() != null && getKeywords().length > 0) {
+                // Use a unique key for different queries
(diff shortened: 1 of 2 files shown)


src/web/demo/src/main/java/org/geoserver/web/demo/PreviewLayerProvider.java
```diff
@@ -129,6 +129,10 @@ public Object getPropertyValue(PreviewLayer item) {
     @Override
     public long size() {
         try {
+            if (getKeywords() != null && getKeywords().length > 0) {
+                //Use a unique key for different queries
```
aaime: Space after "//" ;)

tbarsballe:
Style issue fixed.
I don't have a JDBCConfig setup either.
Fortunately this seems like an innocuous enough change that it shouldn't be an issue. More importantly, it affects core / non-JDBCConfig geoserver, so even if this does cause a minor reduction in GUI performance for JDBCConfig, that is probably an acceptable trade-off.

I'll merge (+ backport to 2.11 and 2.12) once the CI build is done.

PR #2624, opened 2017-11-03
[GEOS-7622] Automatic relase scripts should copy PDF docs, shouldn't copy mac and win prebuild files

tbarsballe:
See here: https://osgeo-org.atlassian.net/browse/GEOS-7622


aaime:
Yes please!

PR #2625, opened 2017-11-03
[GEOS-7343] Add jetty-servlets to system-independent binary to support CORS filter

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-7343

Tested locally by building the bin.zip and uncommenting the CORS filter, seems to work. Should try and confirm with the next release artifacts.

### DIFF
--- a/src/release/jetty/modules/servlets.mod
+++ b/src/release/jetty/modules/servlets.mod
@@ -1,0 +1,9 @@
+#
+# Jetty Servlets Module
+#
+
+[depend]
+server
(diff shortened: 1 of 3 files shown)


aaime:
I'd say let's merge it then? I did not test it personally, but this seems to be a sore point in deploys, a fix is much appreciated.

tbarsballe:
Yep. I'll also make sure and test this during the 2.12.1 release this week.

PR #2626, opened 2017-11-08
Remove duplicate dependency gs-wps-core (en)

chrismayer:
This removes the duplicate dependency ``gs-wps-core`` in the English WPS-implementation documentation. Otherwise the duplicate leads to an error when building the WPS with Maven:

``duplicate declaration of version ${gs.version}``


aaime:
Thank you!

PR #2630, opened 2017-11-10
[GEOS-8332] DataStoreFileController wrong management of FeatureTypes when dealing with DB-based DataStores

afabiani:


src/rest/src/main/java/org/geoserver/rest/util/IOUtils.java
```diff
@@ -769,7 +769,7 @@ public static void inflate(ZipFile archive, Resource outputDirectory, String fil
      */
     public static void inflate(ZipFile archive, Resource outputDirectory, String fileName,
             String workspace, String store, List<Resource> files,
-            boolean external) throws IOException {
+            boolean external, boolean saveFile) throws IOException {
```
aaime: Missing javadoc for the new parameter

aaime: Also, this is public API, there should be a method without the new parameter and a default with the same behavior as before (for backwards compatibility)

src/restconfig/src/main/java/org/geoserver/rest/catalog/DataStoreFileController.java
```diff
@@ -492,6 +519,14 @@ public void dataStorePut(
                     if (!uploadedFile.parent().delete()) {
                         LOGGER.info("Unable to delete " + uploadedFile.path());
                     }
+                } else if (uploadedFile.getType() == Resource.Type.DIRECTORY) {
+                    for (Resource file : files) {
```
aaime: Resource.delete removes files recursively when called, it should be enough. You can check the return value to see if the contents have been fully removed, or not. Or you did experience single file locking issues? In any case, a warning is issued by Files.emtpyDirectory if that happens.

afabiani: Hi @aaime well the problem here is a bit different. Basically "uploadedFile" for some type of DataStore it is the folder containing all the uploaded files. "files" contains the list of the files uploaded on the current session and the flag "canDelete" tells me if I need to remove them or not.

I have chosen this approach in order to delete just the uploaded files selectively instead of cleaning the whole parent folder. Do you think that this is correct?

src/rest/src/main/java/org/geoserver/rest/util/IOUtils.java
afabiani: @aaime ok thanks, will fix ASAP.

src/restconfig/src/main/java/org/geoserver/rest/catalog/DataStoreFileController.java
aaime: Ah ok, sorry, misread the code :-)

PR #2631, opened 2017-11-12
editing english swagger for workspaces

ho3einjb:


jodygarnett:
@tbarsballe this is a small change/correction, no need for CLA. Is it correct?

tbarsballe:
Not quite sure. `workspaceBody` -> `workspace` seems correct based on the schema, but possibly inconsistent with our other swagger docs, e.g. https://github.com/geoserver/geoserver/blob/master/doc/en/api/1.0.0/layers.yaml#L176

So I don't know if that naming scheme is just by preference, or has some functional meaning.

Looking at the code, 401 -> 500 is wrong: https://github.com/geoserver/geoserver/blob/master/src/restconfig/src/main/java/org/geoserver/rest/catalog/WorkspaceController.java#L116

We do in fact return a 401 Unauthorized if the workspace already exist (That is probably the wrong error code to be returning, but it is the current behaviour)

So overal, I would say these changes are not correct, no.

aaime:
The original author does not seem to be anymore on github (check https://github.com/ghost, it's for deleted users). Closing the pull request.

PR #2637, opened 2017-11-14
NSG community module, timeout support

aaime:


src/community/nsg-profile/src/test/java/org/geoserver/nsg/timeout/TimeoutTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+public class TimeoutTest extends WFS20TestSupport {
+
+    @Override
+    protected void onSetUp(SystemTestData testData) throws Exception {
+        super.onSetUp(testData);
+
+        SimpleFeatureStore fs = (SimpleFeatureStore) getCatalog().getFeatureTypeByName
+                ("Fifteen").getFeatureSource(null, null);
+        // double the number of features to get a large amount of them, breaking
```
nmco: We are actually multiplying them by 5 right ?

src/ows/src/main/java/org/geoserver/ows/Request.java
```diff
@@ -21,7 +21,7 @@
  * {@link Dispatcher} while doing its dispatching work. In case of dispatching
  * exceptions some fields may be left blank, depending how far the dispatching
  * went.
- * 
+ *
  * @author Justin DeOliveira
```
nmco: Not sure if all the reformat change sin this file were intentional.

src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
```diff
@@ -68,6 +34,40 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
```
nmco: InteliJ messing up with the imports ? No an issue ...

src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
```diff
@@ -1771,6 +1771,18 @@ void handleServiceException( ServiceException se, Service service, Request reque
             handler = new SOAPServiceExceptionHandler(handler);
         }
 
+        // remove content disposition if set (to allow browsers inline display of exception)
+        HttpServletResponse httpResponse = request.getHttpResponse();
+        if (httpResponse.containsHeader(HttpHeaders.CONTENT_DISPOSITION)) {
```
nmco: Looks like a legit change, but not sure were this is coming from.

nmco:
Some minor comments but apart from that looks good.

src/community/nsg-profile/src/test/java/org/geoserver/nsg/timeout/TimeoutTest.java
aaime: Nope, each time the loop iterates, the number of items doubles, so you get 30 at the end of the first loop, 60 and the second, and so on: 30, 60, 120, 240, 480. I'll explain it better in the comment.

src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
aaime: When requesting a downloadable file (like shapefile) the content type and content disposition are set before giving the output stream to the respnse. On exception, they are still there, and the browser proposes to download what looks like a zip file, while it's really an exception. This causes some confusion on Linux, but much more so on OSs uses the extension to determine the type of application needed to open the file. So this change tries to reset the content disposition if set to avoid that.

src/ows/src/main/java/org/geoserver/ows/Request.java
aaime: It was not, I'll remove this change.

aaime:
Pull request updated based on feedback

PR #2639, opened 2017-11-14
[GEOS-8391,GEOS-8392,GEOS-8397] Fix GeoServer GWC integration issues related with WMTS

nmco:
**Depends on GeoWebCache PR: https://github.com/GeoWebCache/geowebcache/pull/565**

Associated issues: 
* https://osgeo-org.atlassian.net/browse/GEOS-8391
* https://osgeo-org.atlassian.net/browse/GEOS-8392
* https://osgeo-org.atlassian.net/browse/GEOS-8397

**Makes GeoServer default styles be advertised in WMTS capabilities document.**

Before this fix WMTS capabilities document will contain this:

```xml
<Contents>
    <Layer>
      <ows:Title>Manhattan (NY) points of interest</ows:Title>
      <ows:Abstract>Points of interest in New York, New York (on Manhattan). One of the attributes contains the name of a file with a picture of the point of interest.</ows:Abstract>
      <ows:WGS84BoundingBox>
        <ows:LowerCorner>-74.0118315772888 40.70754683896324</ows:LowerCorner>
        <ows:UpperCorner>-74.00153046439813 40.719885123828675</ows:UpperCorner>
      </ows:WGS84BoundingBox>
      <ows:Identifier>tiger:poi</ows:Identifier>
      <Style>
        <ows:Identifier>burg</ows:Identifier>
        <LegendURL format="image/png" xlink:href="http://localhost:8081/geoserver/ows?service=WMS&amp;request=GetLegendGraphic&amp;format=image%2Fpng&amp;width=20&amp;height=20&amp;layer=tiger%3Apoi&amp;style=burg" width="20" height="20"/>
      </Style>
```

After this fix:

```xml
<Contents>
    <Layer>
      <ows:Title>Manhattan (NY) points of interest</ows:Title>
      <ows:Abstract>Points of interest in New York, New York (on Manhattan). One of the attributes contains the name of a file with a picture of the point of interest.</ows:Abstract>
      <ows:WGS84BoundingBox>
        <ows:LowerCorner>-74.0118315772888 40.70754683896324</ows:LowerCorner>
        <ows:UpperCorner>-74.00153046439813 40.719885123828675</ows:UpperCorner>
      </ows:WGS84BoundingBox>
      <ows:Identifier>tiger:poi</ows:Identifier>
      <Style isDefault="true">
        <ows:Identifier>poi</ows:Identifier>
[...]

PR #2649, opened 2017-11-19
[GEOS-8402] Redirect users experiencing an internal UI error to the user list

aaime:


aaime:
Here is how it looks:

selezione_057

aaime:
Ouch, we have a corrupted repo breaking the build on Travis, restarting the build is not helping:

````
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:2.3.2:compile (default-compile) on project gs-gdal: Compilation failure
[ERROR] error: error reading /home/travis/.m2/repository/org/geotools/gt-coverage/19-SNAPSHOT/gt-coverage-19-SNAPSHOT.jar; zip file is empty
[ERROR] -> [Help 1]
````

File: src/web/core/src/main/java/org/geoserver/web/GeoServerErrorPage.html [hunk 1]
```diff
@@ -5,9 +5,7 @@
             <wicket:message key="GeoServerErrorPage.loginRequired"></wicket:message>
         </p>
         <p wicket:id="trace-explanation">
-            <wicket:message key="GeoServerErrorPage.preamble"></wicket:message> 
-            <a wicket:id="jiraLink"></a>
-            <wicket:message key="GeoServerErrorPage.conclusion"></wicket:message> 
+            <a wicket:id="userListLink"></a>
```
bencaradocdavies: Can we set `target="_blank"` on the link to request a new browser window? I usually avoid these but in this case it could stop the user from losing the stack trace.

src/web/core/src/main/resources/GeoServerApplication.properties
```diff
@@ -269,11 +269,11 @@ GeoServerHomePage.description         = Welcome
 GeoServerErrorPage.title              = Oops, something went wrong...
 GeoServerErrorPage.description        = Sorry, something unexpected happened on the server.
 GeoServerErrorPage.loginRequired      = If you were logged in, you could see an error report.
-GeoServerErrorPage.preamble           = Here's an error report you can include in a
-GeoServerErrorPage.jiraName           = JIRA bug report
+GeoServerErrorPage.userListLink       = http://geoserver.org/comm/
+GeoServerErrorPage.userListText       = Please follow this link, subscribe to the user list, and report this issue. Include the stack trace below and the steps to reproduce the problem.
```
bencaradocdavies: Perhaps the infamous "click here" to make it even more blindingly obvious. And perhaps a more general mention of help as it might be their problem not a problem with GeoServer? Something like: "Click here to subscribe to the GeoServer users mailing list. When asking for help, please include the stack trace below and the steps to reproduce the problem."

aaime:
@bencaradocdavies updated according to your review, the text message I came up with is not verbatim what you suggestion but I hope it carries the same information/spirit

aaime:
Build is failing in GWC, maybe because of pull requests recently merged on GWC, see  discussion here: https://github.com/geoserver/geoserver/pull/2639

bencaradocdavies:
@aaime that text looks good to me. I also tried restarting the Travis build, but as you suggest, the failure may be a legitimate GWC problem.

aaime:
Finally green. And merged!

PR #2651, opened 2017-11-20
[GEOS-8386]: Make UUID a visible field in DescribeFeatureType

dromagnoli:


src/wfs/src/main/java/org/geoserver/wfs/xml/FeatureTypeSchemaBuilder.java
```diff
@@ -780,8 +780,14 @@ private XSDComplexTypeDefinition buildComplexSchemaContent(ComplexType complexTy
                         Class binding = attribute.getType().getBinding();
                         typeName = findTypeName(binding);
                         if (typeName == null) {
-                            throw new NullPointerException("Could not find a type for property: "
-                                    + attribute.getName() + " of type: " + binding.getName());
+                            // Fallback on String
+                            logger.finer("Fallback mapping: attribute " + attribute.getName() + " to String" );
```
aaime: Guard logging :-)

src/wfs/src/main/java/org/geoserver/wfs/xml/FeatureTypeSchemaBuilder.java
```diff
@@ -780,8 +780,14 @@ private XSDComplexTypeDefinition buildComplexSchemaContent(ComplexType complexTy
...
+                            typeName = findTypeName(String.class);
+                            if (typeName == null) {
+                                throw new NullPointerException(
```
aaime: Maybe a WFSException or a ServiceException instead?

src/wfs/src/main/java/org/geoserver/wfs/xml/FeatureTypeSchemaBuilder.java
dromagnoli: Yep. At the very beginning I was thinking to add guard but I have noticed that the class has several other logs without any guard so I didn't introduce that there.

src/wfs/src/main/java/org/geoserver/wfs/xml/FeatureTypeSchemaBuilder.java
dromagnoli: I have preferred to preserve the original exception (see old line 783) in case any caller is catching that specific exception.

PR #2656, opened 2017-11-22
[GEOS-8409] GZIPResponseStream should not throw Exception on flushing

afabiani:
Fixes JIRA GEOS-8409


aaime:
Yep, thanks for taking the time and writing a test for this one liner (required, but painful when the tests takes much more work than the fix). Merged, don't know if you want to backport or not (and remember to close the jira ticket)

PR #2659, opened 2017-11-22
[GEOS-8418] GeoServer Wicket multipart requests fail with gs-geogig

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-8418

Issue caused by https://github.com/geoserver/geoserver/pull/2543
Issue (and solution) similar to [GEOS-7925](https://osgeo-org.atlassian.net/browse/GEOS-7925) (https://github.com/geoserver/geoserver/pull/2142)

Summary: The CommonsMultipartResolver interferes with other multipart handlers within GeoServer, in particular the Wicket Form handler.
Consequently, anything that uses it should wrap it with a filter so that it is only applied to the applicable endpoints.


tbarsballe:
This fix is a bit hard to test reliabley, due to how the GeoServer MockRequest framework works, it mostly disregards the filter (and the multipart resolver, before and after this change).
Manual testing shows that this does fix the issue in GeoServer.

@emerkle826 - is there a good manual test I can to to verify this does not break the fix introduced by #2543 ?

aaime:
@tbarsballe don't know if this is useful or not, but you can programmatically add the servlet filters you want in tests, see:
https://github.com/geoserver/geoserver/blob/b772809a14681d7838cd6d4af2c9891809e6e174/src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java#L1970

tbarsballe:
That might help, although is at the point where what is being tested doesn't really represent the actual code/configuration anymore, especially since the whole issue here is with the wrong stuff picking up the resolver (so if the filter needs to be manually added to the test, it doesn't make an especially good regression test for if something similar adds a new filter that would conflict).

However, I should at least be able to add a minimal functionality test with that, which is better than nothing.

DBlasby:
I verified this was working in QGIS (which does import/export of geopackage files)

tbarsballe:
I was able to add a test case based on a GeoPackage Multipart import, should be adequate to test this.

mrcnc:
@tbarsballe is this something we need to backport to the 2.12.x branch?

tbarsballe:
I've already backported it

PR #2661, opened 2017-11-23
[GEOS-8420] Remove unused RESTLET dependencies from Backup-Restore community modules

afabiani:
https://osgeo-org.atlassian.net/browse/GEOS-8420


bencaradocdavies:
Thanks, @afabiani. It is also worth noting that this PR makes a bunch of packaging changes and fixes some tests so was a bit more work than suggested by the title.  :-)

bencaradocdavies:
@afabiani this fix should probably be backported to 2.12.x where backup-restore has also been re-enabled. Do you plan to backport?

dalbani:
Backport to 2.12.x proposed in https://github.com/geoserver/geoserver/pull/2675.

PR #2662, opened 2017-11-23
Fixes for GEOS-8427: rename scriptMgr to scriptManager, fix ruby test  and upgrade groovy

jericks:
The script community module can't find the scriptMgr bean because it was renamed scriptManager. Also there is a failing ruby application test because of the restlet to spring migration.

JIRA:
https://osgeo-org.atlassian.net/browse/GEOS-8427

### DIFF
--- a/src/community/script/groovy/pom.xml
+++ b/src/community/script/groovy/pom.xml
@@ -24,7 +24,7 @@
   <dependency>
    <groupId>org.geoscript</groupId>
    <artifactId>geoscript-groovy</artifactId>
-   <version>1.7-SNAPSHOT</version>
+   <version>1.11-SNAPSHOT</version>
   </dependency>
   <dependency>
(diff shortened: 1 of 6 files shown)


aaime:
I was wondering when the rename happened, after a bit of searching I believe it was @vickdw REST API upgrade, switching from XML beans declaration to component scan that had it happen as a side effect (the bean was named scriptMgr in the XML file, but with @Component not having a name I guess it's just using the class name).

@tbarsballe wondering, did you try the python extension as part of the last Suite release?

aaime:
Backported to 2.12.x too

tbarsballe:
Yes, we did test the python extension, although I think only the script functionality was tested, I don't know if anyone tried out the REST API/Web UI part. And yes, if a name is not specifically provided, `@Component` will just use the class name, so that explains the name change.

PR #2664, opened 2017-11-25
[GEOS-8430] Cache small amount of features in memory to avoid repeated data scans in GetFeature requests

aaime:
https://osgeo-org.atlassian.net/browse/GEOS-8430

Replaces #1355
Rebased the pull request on current master to make sure everything builds against today's code, and:

- Improve use of generic and use try with resources when possible
- Some fixes to docs, cut down on advertisement and augment code behavior description
- Made the class public (sigh) because subclasses of GetFeature in GeoServer extensions might actually need it (since the code is right in the middle of an extension point)
- Added test to make sure GetFeature works with both caching on and off
- Created a GeoServer ticket and made the commits refer to it

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/FeatureSizeFeatureCollection.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/FeatureSizeFeatureCollection.java
@@ -1,0 +1,157 @@
+/* (c) 2017 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.wfs;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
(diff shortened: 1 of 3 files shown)


aaime:
@ahuarte47 can you review?

ahuarte47:
@aaime, it is right for me, thank you very much for your review and improvements!

aaime:
@ianturton @jodygarnett @bencaradocdavies opinions?

aaime:
@ianturton given how narrow this is and the mounting charge against the shapefile format (see http://switchfromshapefile.org/ ) I'd rather keep it in WFS for the time being. We can always move the file to GeoTools later.

bencaradocdavies:
@aaime no objection from me. This is small increase in code base and thus complexity, but if the performance increase is noticeable, it may be worthwhile.

ahuarte47:
Thanks to all!

PR #2670, opened 2017-11-30
[GEOS-8459] Handle case where principal is instance of String

davisc:
This PR adds a condition to account for the principal being an instance of String. This occurs when sending wfs transactions while supplying the oauth access token, anonymous is stored as the VersioningCommitAuthor.

https://osgeo-org.atlassian.net/browse/GEOS-8459


tbarsballe:
Please add a test case and a GeoServer JIRA issue, per https://github.com/geoserver/geoserver/blob/master/CONTRIBUTING.md

With respect to the CLA, you will be covered under the Boundless corporate CLA

tbarsballe:
Looks good to me, thanks for the test case. Merged, and I have also backported to 2.12.x and 2.11.x.

PR #2671, opened 2017-11-30
[GEOS-8403] Gets SLD style for non-sld style updates

johngschulz:


src/web/wms/src/main/java/org/geoserver/wms/web/data/StyleEditPage.java
```diff
@@ -122,6 +123,10 @@ protected void onStyleFormSubmit() {
             getCatalog().save(style);
             // provide feedback to the user
             styleForm.info("Style saved");
+            // retrieve sld style for non-sld formatted styles on update
+            if ((!SLDHandler.FORMAT.equals(format))) {
+              getCatalog().getResourcePool().getStyle(stylePath);
```
tbarsballe: This fix seems reasonable, although I am a little surprised it is using getStyle as opposed to some sort of write method - I assume requesting the file in SLD automatically causes it to be rewritten?

johngschulz: That is correct.  It won't do the transformation until it attempts to retrieve the style, and writes over the saved sld.

src/community/mbstyle/pom.xml
```diff
@@ -62,6 +62,24 @@
 	      <artifactId>gs-restconfig</artifactId>
 	      <version>${project.version}</version>
 	      <scope>test</scope>
-	    </dependency>        
+	    </dependency>
+  	    <dependency>
+          <groupId>org.geoserver.wms</groupId>
```
tbarsballe: This dependency looks like a malformed duplicate duplicate of the other gs-web-wms dependency below (The correct groupId is `org.geoserver.web`). Try removing this one and see if everything still works.

src/community/mbstyle/src/test/java/org/geoserver/community/mbstyle/web/MBStyleEditPageTest.java
```diff
@@ -0,0 +1,96 @@
...
+import org.geotools.styling.PointSymbolizer;
+import org.geotools.styling.PointSymbolizerImpl;
+import org.geotools.styling.SLD;
+import org.geotools.styling.StyledLayerDescriptor;
+import org.geoserver.wms.web.data.OpenLayersPreviewPanel;
+
+
+public class MBStyleEditPageTest extends GeoServerWicketTestSupport {
+    
+	StyleInfo mbstyle;
```
tbarsballe: I see some tabs here, should be using spaces.

src/web/wms/src/main/java/org/geoserver/wms/web/data/OpenLayersPreviewPanel.java
```diff
@@ -77,7 +77,7 @@
     final Random rand = new Random();
     final Component olPreview;
 
-    boolean isPreviewStyleGroup;
+    public boolean isPreviewStyleGroup;
```
tbarsballe: This is technically an api change, and shouldn't be necessary.

src/community/mbstyle/src/test/java/org/geoserver/community/mbstyle/web/MBStyleEditPageTest.java
```diff
@@ -0,0 +1,96 @@
...
+                      " }\n";
+      
+      FormTester form = tester.newFormTester("styleForm");
+      form.setValue("context:panel:format", MBStyleHandler.FORMAT);
+      form.setValue("context:panel:name", "mbstyleTest");
+      tester.executeAjaxEvent("apply", "click");
+      tester.executeAjaxEvent("styleForm:context:tabs-container:tabs:2:link", "click");
+      tester.assertComponent("styleForm:context:panel", OpenLayersPreviewPanel.class);
+      OpenLayersPreviewPanel previewPanel = (OpenLayersPreviewPanel) tester.getComponentFromLastRenderedPage("styleForm:context:panel");
+      assertFalse(previewPanel.isPreviewStyleGroup);
```
tbarsballe: You can just use the value of the checkbox instead; something like: 
```
tester.assertModelValue("styleForm:context:panel:previewStyleGroup", false);
```
(This gets around needing to make `isPreviewStyleGroup a public field)

johngschulz:
The dependency issues I was having are resolved and I've addressed all of the other issues raised in the review.

src/community/mbstyle/src/test/java/org/geoserver/community/mbstyle/web/MBStyleEditPageTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+public class MBStyleEditPageTest extends GeoServerWicketTestSupport {
+    
+    StyleInfo mbstyle;
+    StyleEditPage edit;
+
+    @Before
+    public void setUp() throws Exception {
+        Catalog catalog = getCatalog();
+    	    login();
```
tbarsballe: I still see tabs here - you might want to take a look at your IDE settings and verify tabs to spaces and visible whitespace are turned on.

johngschulz: Indeed, I haven't adjusted my settings since upgrading to the new version of Eclipse. Took some time to set up my preferences properly.

tbarsballe:
Backported to 2.12.x

PR #2674, opened 2017-12-04
GEOS-8464 Geofence: complex rules GUI

NielsCharlier:


src/community/geofence-server/src/main/java/org/geoserver/geofence/web/LayerAttributeModel.java
```diff
@@ -0,0 +1,35 @@
+package org.geoserver.geofence.web;
```
nmco: Missing copyright header.

nmco:
Just had a very quick look, the documentation update seems to be missing, will this be contributed in another PR ?

etj:
In the tab "Layer details" the field "Layer type" should be automatically detected from the selected layer:
image

etj:
In the PR implementation the layer attributes should be typed, while they could be automatically got from the selected featuretype and selected through a dropdown menu, but this can be considered a "nice to have":
image

etj:
Bad error message:
image

NielsCharlier:
okay, changes were made and docs added. I changed the approach to the layer attributes quite a bit, rather than a drop-down box all attributes are shown in the table with the option to specify access mode.

NielsCharlier:
@etj can we merge this?

NielsCharlier:
I have merged since all the reviewer's requests were fulfilled.

PR #2676, opened 2017-12-04
[GEOS-8250] WFS 2.0 DescribeFeatureType responses for app-schema type s contain a spurious WFS 2.0 jar import

nmco:
Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-8250

Like in WFS 1.0, WFS 2.0 schema import is unnecessary when doing a describe feature type. According to WFS 2.0 standard the describe feature type should only describe the feature type not the full response:

> The DescribeFeatureType operation returns a schema description of feature types offered by a WFS instance. The schema descriptions define how a WFS expects feature instances to be encoded on input (via Insert, Update and Replace actions) and how feature instances shall be encoded on output (in response to a GetPropertyValue, GetFeature or GetFeatureWithLock operation).

This PR removes the uncessary and wrong import adn adds several tests cases for this, including app-schema integration tests.

### DIFF
--- a/src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/FeatureChainingWfsTest.java
+++ b/src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/FeatureChainingWfsTest.java
@@ -6,6 +6,9 @@
 
 package org.geoserver.test;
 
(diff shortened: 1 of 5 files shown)


bencaradocdavies:
@nmco that looks great. Tidy and thorough. A full local build + app-schema online tests against postgis passed for me. Thanks!

PR #2677, opened 2017-12-05
GEOS-8468: allow setting alternative LabelCache implementation in Ren…

mbarto:
…deredImageMapOutputFormat


src/wms/src/main/java/org/geoserver/wms/map/RenderedImageMapOutputFormat.java
```diff
@@ -174,6 +175,8 @@
      * The lookup table used for data type transformation (it's really the identity one)
      */
     private static LookupTableJAI IDENTITY_TABLE = new LookupTableJAI(getTable());
+    
+    private Class<? extends LabelCache> labelCache = null;
```
aaime: Rather than assuming the LabelCache has a empty constructor I'd use a Supplier<LabelCache>

aaime: Or even better a Function<WebMapContent>, so that it gets some context to build the label cache

aaime:
Yep, thanks for the changes

PR #2681, opened 2017-12-11
[GEOS-8485] GetLegendGraphic request not working with layer groups in workspace

mwerlen:
https://osgeo-org.atlassian.net/browse/GEOS-8485

LocalWorkspaceCatalog is called with a "workspace:layergroup" string wich was not well handled.
This commit aligns getLayerGroupByName on getLayerByName method.

To reproduce, configure a LayerGroup in a Workspace then call a GetLegendGraphic. Without this PR, call fails.

Contributor License Agreement sent for Sogelink.

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/impl/LocalWorkspaceCatalog.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/LocalWorkspaceCatalog.java
@@ -165,15 +165,14 @@ public class LocalWorkspaceCatalog extends AbstractCatalogDecorator implements C
             //prefix the unqualified name
             if (name.contains(":")) {
                 //name already prefixed, ensure it is prefixed with the correct one
-                if (name.startsWith(wsName+":")) {
+                if (name.startsWith(wsName + ":")) {
                     //good to go, just pass call through
                     LayerGroupInfo layerGroup = super.getLayerGroupByName(name);
                     if (layerGroup != null) {
                         return wrap(layerGroup);
                     }
                     // else fall back on unqualified lookup
-                }
-                else {
+                } else {
                     //JD: perhaps strip of existing prefix?
                 }
             }
--- a/src/main/src/main/java/org/geoserver/catalog/impl/LocalWorkspaceCatalog.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/LocalWorkspaceCatalog.java
@@ -160,11 +160,28 @@ public class LocalWorkspaceCatalog extends AbstractCatalogDecorator implements C
     @Override
     public LayerGroupInfo getLayerGroupByName(String name) {
         if (LocalWorkspace.get() != null) {
-            LayerGroupInfo layerGroup = super.getLayerGroupByName(LocalWorkspace.get(), name);
+            String wsName = LocalWorkspace.get().getName();
+
+            //prefix the unqualified name
+            if (name.contains(":")) {
+                //name already prefixed, ensure it is prefixed with the correct one
+                if (name.startsWith(wsName+":")) {
+                    //good to go, just pass call through
+                    LayerGroupInfo layerGroup = super.getLayerGroupByName(name);
+                    if (layerGroup != null) {
+                        return wrap(layerGroup);
+                    }
+                    // else fall back on unqualified lookup
+                }
+                else {
+                    //JD: perhaps strip of existing prefix?
+                }
+            }
+            //prefix it explicitly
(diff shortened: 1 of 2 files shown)


bencaradocdavies:
@mwerlen the Travis build will likely fail because GeoServer master is currently broken by [a change in GeoTools](https://github.com/geotools/geotools/pull/1768). I will restart your build when the problematic GeoTools change has been reverted.

bencaradocdavies:
@mwerlen the Travis build failed as expected. The [problematic change in GeoTools](https://github.com/geotools/geotools/pull/1768) has now been reverted and a new GeoTools snapshot has been deployed, so I have restarted the Travis build for this pull request.

src/main/src/main/java/org/geoserver/catalog/impl/LocalWorkspaceCatalog.java
```diff
@@ -160,11 +160,28 @@ public LayerGroupInfo getLayerGroup(String id) {
     @Override
     public LayerGroupInfo getLayerGroupByName(String name) {
         if (LocalWorkspace.get() != null) {
-            LayerGroupInfo layerGroup = super.getLayerGroupByName(LocalWorkspace.get(), name);
+            String wsName = LocalWorkspace.get().getName();
+
+            //prefix the unqualified name
+            if (name.contains(":")) {
+                //name already prefixed, ensure it is prefixed with the correct one
+                if (name.startsWith(wsName+":")) {
```
bencaradocdavies: Formatting: need spaces around "+".

src/main/src/main/java/org/geoserver/catalog/impl/LocalWorkspaceCatalog.java
```diff
@@ -160,11 +160,28 @@ public LayerGroupInfo getLayerGroup(String id) {
...
+                    //good to go, just pass call through
+                    LayerGroupInfo layerGroup = super.getLayerGroupByName(name);
+                    if (layerGroup != null) {
+                        return wrap(layerGroup);
+                    }
+                    // else fall back on unqualified lookup
+                }
+                else {
+                    //JD: perhaps strip of existing prefix?
```
bencaradocdavies: Formatting: should be "} else {" all on one line.

Also, who is JD? @jdeolive? Is this the restoration of removed code or code copied from elsewhere? I am not familiar enough with this to know whether it will cause problems. There have been a bunch of recent changes, but the last change to this method was from @groldan back in May 2012. @aaime, any opinion?

bencaradocdavies:
The Travis build passed ([eventually](https://osgeo-org.atlassian.net/browse/GEOS-8038)).

src/main/src/main/java/org/geoserver/catalog/impl/LocalWorkspaceCatalog.java
mwerlen: It's an adaptation from the getLayerByName method: 
```
    @Override
    public LayerInfo getLayerByName(String name) {
        if (LocalWorkspace.get() != null) {
            String wsName = LocalWorkspace.get().getName();

            //prefix the unqualified name
            if (name.contains(":")) {
                //name already prefixed, ensure it is prefixed with the correct one
                if (name.startsWith(wsName+":")) {
                    //good to go, just pass call through
                    return wrap(super.getLayerByName(name));
                }
                else {
                    //JD: perhaps strip of existing prefix?
                }
            }

            //prefix it explicitly
            NamespaceInfo ns = super.getNamespaceByPrefix(LocalWorkspace.get().getName());
            LayerInfo layer = super.getLayerByName(new NameImpl(ns.getURI(), name));
            return wrap(layer);
        }
        return super.getLayerByName(name);
    }
```

The commit seems to be from jdeolive based on work from Juan Marin: https://github.com/sogelink/geoserver/commit/dbedaaea132ca88b2371107132ee7fd9271ca938 with commit message : "GEOS-5016, dropping layer name prefixes when accessing through local workspace / virtual service, original patch by Juan Marin"

src/main/src/main/java/org/geoserver/catalog/impl/LocalWorkspaceCatalog.java
mwerlen: Done

mwerlen:
Travis failed on a test unrelated to the PR, could you restart it ?

bencaradocdavies:
@mwerlen I restarted the Travis build and it passed.

mwerlen:
Thank you @bencaradocdavies

bencaradocdavies:
@mwerlen thanks for the fix.

PR #2693, opened 2017-12-19
[GEOS-8250] WFS 2.0 DescribeFeatureType responses for app-schema types contain a spurious WFS 2.0 jar import

ridethepenguin:
Link to related JIRA issue:
https://osgeo-org.atlassian.net/browse/GEOS-8250

This is the backport of #2676 to GeoServer 2.12.x (stable).


ridethepenguin:
Same as #2694, build failure was caused by the usage of `typeNames` instead of `typeName` in WFS 2.0 DescribeFeatureType requests (it works with both on `master` thanks to [this commit](https://github.com/geoserver/geoserver/commit/025c34b3bdd626af3ecd96392abc2a136312ccf8)). All should be well now.

PR #2696, opened 2017-12-20
[GEOS-8498] WMTS multi dimensional community module throws NULL pointer exception instead of an exception report

nmco:
Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-8498

This PR fix the issue and adds a new test case, only code from **``wmts-multi-dimensional`` community module** was changed.


aaime:
Failed, maybe caught in @dromagnoli geotools changes with the corresponding geoserver pull request not yet merged?

nmco:
I will rekuick the build later, thanks for the heads up Andrea.

PR #2698, opened 2017-12-23
[GEOS-8507] Make it possible to run XML related tests in IDE when both GeoTools and GeoServer sources are loaded

aaime:


aaime:
Tried to make minimal changes sufficient to have tests work in IDE

src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java
```diff
@@ -112,10 +112,12 @@
 import org.geoserver.security.password.GeoServerDigestPasswordEncoder;
 import org.geoserver.security.password.GeoServerPBEPasswordEncoder;
 import org.geoserver.security.password.GeoServerPlainTextPasswordEncoder;
+import org.geoserver.util.EntityResolverProvider;
 import org.geotools.data.DataUtilities;
 import org.geotools.data.simple.SimpleFeatureSource;
 import org.geotools.util.logging.Log4JLoggerFactory;
 import org.geotools.util.logging.Logging;
+import org.geotools.xml.PreventLocalEntityResolver;
```
bradh: Is this needed? Or left over from an earlier version?

src/main/src/main/java/org/geoserver/util/EntityResolverProvider.java
```diff
@@ -23,6 +28,54 @@
     public static final EntityResolverProvider RESOLVE_DISABLED_PROVIDER = new EntityResolverProvider(
             null);
 
+    /**
+     * In IDEs during development GeoTools sources can be in the classpath of GeoServer tests, this
+     * resolver allows them to be resolved while blocking the rest
+     */
+    public static final EntityResolver RESOLVE_DISABLED_PROVIDER_DEVMODE = new PreventLocalEntityResolver() {
```
bradh: This seems very test-specific behaviour. Is there a way it could live in the `test/` side without (big) rework on the `main/` side? If not, could it be made more general so that you can just pass in a map of local locations => schemas?

src/wfs/src/test/java/org/geoserver/wfs/v2_0/GetFeatureTest.java
```diff
@@ -464,14 +461,13 @@ public void testPostWithLessThanOnBoundedBy() throws Exception {
                     + "</wfs:Query>"
                     + "</wfs:GetFeature>";
             MockHttpServletResponse response = postAsServletResponse("wfs", xml);
+            System.out.println(response.getContentAsString());
```
bradh: Left over?

src/main/src/main/java/org/geoserver/util/EntityResolverProvider.java
aaime: As you saw in other comments this is the indeed third iteration of this patch, and the least intrusive one of the three variations.
I've first stared by allowing all entities in the test, which required a big rework in existing tests checking external entities.

Then tried to only allow schema entities (anything ending with .xsd), that also required significant rework as various tests explicitly check for entity expansions on stuff ending with .xsd.

This version is the one that requires the least changes for a case that's indeed quite specific (developer working in Maven aware IDE with both GeoTools and GeoServer sources loaded).

Passing "paths" is more work and would not help, what would be needed are regular expressions matching paths, and eventually a different one for systemId and baseURI. It would also require to make changes down in Geotools.

If test specific code in this class is bothering you maybe the code could be reworked to have a DEFAULT_RESOLVER static non final field that defaults to ``PreventLocalEntityResolver.INSTANCE`` but that test code could replace with its own version, and move the test specific code to SystemTestSupport. 
Would that be better? Is it something you're interested in doing (a pull request can be made against this one, targetting the right branch on my clone)

src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java
aaime: Indeed leftover, removing it

src/wfs/src/test/java/org/geoserver/wfs/v2_0/GetFeatureTest.java
aaime: Yep, removing it

PR #2701, opened 2018-01-02
[GEOS-8510] Show feedback messages both at top and bottom of the page

aaime:
Enough said:

selezione_051

### DIFF
--- a/src/web/wms/src/main/java/org/geoserver/wms/web/data/OpenLayersPreviewPanel.html
+++ b/src/web/wms/src/main/java/org/geoserver/wms/web/data/OpenLayersPreviewPanel.html
@@ -19,6 +19,7 @@
         </li>
       </ul>
       <div wicket:id="olPreview" style="margin-bottom: 1em"/>
+      <div wicket:id="dialog"/>
     </wicket:panel>
   </body>
 </html>
--- a/src/community/backup-restore/web/src/main/java/org/geoserver/backuprestore/web/BackupRestoreDataPage.java
+++ b/src/community/backup-restore/web/src/main/java/org/geoserver/backuprestore/web/BackupRestoreDataPage.java
@@ -182,7 +182,7 @@ public class BackupRestoreDataPage extends GeoServerSecuredPage implements GeoSe
             }
 
             protected void onError(AjaxRequestTarget target, Form<?> form) {
-                target.add(feedbackPanel);
+                addFeedbackPanels(target);
             }
             
             protected void onSubmit(AjaxRequestTarget target, final Form<?> form) {
@@ -213,7 +213,7 @@ public class BackupRestoreDataPage extends GeoServerSecuredPage implements GeoSe
                     //update the button back to original state
                     resetButtons(form, target, "newBackupStart");
 
-                    target.add(feedbackPanel);
+                    addFeedbackPanels(target);
                 }
 
                 cancel.setDefaultModelObject(jobid);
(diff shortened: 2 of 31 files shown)


src/main/src/main/java/org/geoserver/catalog/impl/ResolvingProxy.java
```diff
@@ -22,7 +23,7 @@
  * @author Justin Deoliveira, The Open Planning Project
  *
  */
-public class ResolvingProxy extends ProxyBase {
+public class ResolvingProxy extends ProxyBase implements Serializable {
```
bradh: Is this really part of the same (feedback panel) change?

src/web/wms/src/main/java/org/geoserver/wms/web/data/OpenLayersPreviewPanel.html
```diff
@@ -19,6 +19,7 @@
         </li>
       </ul>
       <div wicket:id="olPreview" style="margin-bottom: 1em"/>
+      <div wicket:id="dialog"/>
```
bradh: Part of same change?

aaime:
Import wise, it's different behavior between IntelliJ and Eclipse (so no project wide change, everyone is free to use whatever IDE they like). I'm still a noob IntelliJ wise so maybe I did it unadvertedly, but I have a hunch the sucker is doing changes along with automatic refactors. Will check it.

src/main/src/main/java/org/geoserver/catalog/impl/ResolvingProxy.java
aaime: No, it's a fix for an issue I've found while developing this that was preventing me from moving on (it's related to dangling references in configuration). 
Will revert and someone else with time and interest will do it the "right" way.

src/web/wms/src/main/java/org/geoserver/wms/web/data/OpenLayersPreviewPanel.html
aaime: Unsure, but probably, the preview page in the style editor was just not working, and this is structural, it cannot have been failing just for me. The code is referencing a dialog element that's just not there in the panel, it's available in the page but the code cannot find it. Might be a side effect of changes in the base page, but could not confirm it. Leaving it to be on the safe side.
@tbarsballe git tells me you added this dialog reference in the code in August along with the style groups, do you know how it was working since the HTML did not have one?

aaime:
Checked the imports, all I could see  was removal of unused imports during the refactor. No harm done there.

bradh:
On the imports thing, I was just suggesting that that be done project wide at some stage. 

I like to see them removed, it just makes real changes somewhat longer to review.

aaime:
Build failed due to repo communication issue, restarting:

``
[ERROR] Failed to execute goal on project geoserver: Could not resolve dependencies for project org.geoserver:geoserver:pom:2.13-SNAPSHOT: Failed to collect dependencies at org.geotools:gt-api:jar:19-SNAPSHOT: Failed to read artifact descriptor for org.geotools:gt-api:jar:19-SNAPSHOT: Could not transfer artifact org.geotools:library:pom:19-20180103.083430-158 from/to boundless (https://repo.boundlessgeo.com/main/): Connect to repo.boundlessgeo.com:443 [repo.boundlessgeo.com/52.22.156.115] failed: Connection timed out (Connection timed out) -> [Help 1]
``

aaime:
Ok, this is interesting, now it failed in the community build part because the code has been updated to use a new method introduced in this pull request, and the build runs with -U forcing the download of jars from the repo. However, it's a "screwed if you do, screwed if you don't" case, cause removing the -U would fix this one, but may break the build in a community module that depends on a GeoTools jar change which was not downloaded in the main build. Merging and moving on, if anything else breaks I'll be on it.

PR #2702, opened 2018-01-03
[GEOS-8511] Allow file chooser text editor to auto-complete typed paths

aaime:
schermata del 2018-01-03 11-08-48

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/data/store/panel/FileParamPanel.java
+++ b/src/web/core/src/main/java/org/geoserver/web/data/store/panel/FileParamPanel.java
@@ -5,12 +5,10 @@
  */
 package org.geoserver.web.data.store.panel;
 
-import java.io.File;
-import java.io.FileFilter;
-
 import org.apache.wicket.Component;
 import org.apache.wicket.ajax.AjaxRequestTarget;
 import org.apache.wicket.ajax.markup.html.form.AjaxSubmitLink;
+import org.apache.wicket.extensions.ajax.markup.html.autocomplete.AutoCompleteTextField;
 import org.apache.wicket.extensions.ajax.markup.html.modal.ModalWindow;
 import org.apache.wicket.markup.html.basic.Label;
 import org.apache.wicket.markup.html.form.Form;
@@ -21,7 +19,16 @@ import org.apache.wicket.markup.html.panel.Panel;
 import org.apache.wicket.model.IModel;
 import org.apache.wicket.model.Model;
 import org.apache.wicket.validation.IValidator;
+import org.geoserver.web.wicket.browser.FileRootsFinder;
 import org.geoserver.web.wicket.browser.GeoServerFileChooser;
+import org.geotools.util.logging.Logging;
+
+import java.io.File;
+import java.io.FileFilter;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * A label, a text field, a file chooser
@@ -29,7 +36,7 @@ import org.geoserver.web.wicket.browser.GeoServerFileChooser;
  * @author Andrea Aime - GeoSolutions
  */
 public class FileParamPanel extends Panel implements ParamPanel {
-
+    private static final Logger LOGGER = Logging.getLogger(FileParamPanel.class);
 	private static final long serialVersionUID = 2630421795437249103L;
 	TextField<String> textField;
     ModalWindow dialog;
@@ -60,7 +67,23 @@ public class FileParamPanel extends Panel implements ParamPanel {
         add(label);
 
         // the text field, with a decorator for validations
-        textField = new TextField<String>("paramValue", new FileModel(paramValue));
+        FileRootsFinder rootsFinder = new FileRootsFinder(false);
+        textField = new AutoCompleteTextField<String>("paramValue", new FileModel(paramValue)) {
+            @Override
+            protected Iterator<String> getChoices(String input) {
+                try {
+                    // do we need to filter files?
+                    FileFilter fileFilter = FileParamPanel.this.fileFilter != null ?
+                            FileParamPanel.this.fileFilter.getObject() : null;
+
+                    return rootsFinder.getMatches(input, fileFilter).iterator();
+                } catch(Exception e) {
+                    // this is a helper, don't let it break the UI at runtime but log errors instead
+                    LOGGER.log(Level.SEVERE, "Failed to provide autocomplete for path " + input, e);
+                    return Collections.emptyIterator();
+                }
+            }
+        };
         textField.setRequired(required);
         textField.setOutputMarkupId(true);
         // set the label to be the paramLabelModel otherwise a validation error would look like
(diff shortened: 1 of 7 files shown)


jodygarnett:
Okay this did not read quite as expected, sorry you are running into trouble using Paths; it was intended for referencing content inside the data directory - not for referencing Files.

src/platform/src/main/java/org/geoserver/platform/resource/FileSystemResourceStore.java
```diff
@@ -162,7 +162,7 @@ public String toString() {
         File file;
 
         public FileSystemResource(String path) {
-            this.path = path;
+            this.path = Paths.convert(path);
```
jodygarnett: So adding this logic here is not quite what was intended, paths are supposed to be within the data directory.

For the rare case, mostly test cases, that you need to use a file as a resource there is the [Files.asResource( file )](https://github.com/geoserver/geoserver/blob/master/src/platform/src/main/java/org/geoserver/platform/resource/Files.java#L308) method.

The [Paths.convert( file, file )](https://github.com/geoserver/geoserver/blob/55c4740060eb2b8809ae935a510ac83eeb923313/src/platform/src/main/java/org/geoserver/platform/resource/Paths.java#L257) should be used rather than insert this logic?

aaime: I disagree, if you don't do the above the file system resource will stop working on Windows, e.g., most methods stop working (think parent for example, it won't find the parent because it's looking for "/" while the path above in the constructor contains "\").

My guess is that the work was not tested against Windows at all, and probably not though out to work there too (the data directory does not live in the clouds, it normally comes from a real file system, if you get the path from File on Windows you end up with "\" in the path).

Can you try the whole thing on Windows so that you can check what works, what does not, adjust your expectations accordingly, and change the code so that it actually works? This is not my mess, someone else did it, I'm just trying to get a auto-complete to work.

aaime: The code was definitely pushing into the constructor a "\" but I cannot replicate that anymore, and I've rewritten the file chooser lookup code so many times now that I don't remember which version managed to do that (I never called the constructor directly, there is some code path doing that by just using FileSystemResourceStore and Resource, but cannot find it right now). Removing it.

aaime: I've also pushed a change to clarify in the javadocs that unix like paths need to be used, previously only one method did that.

jodygarnett: Thanks Andrea, the javadocs change is wise, I tried to use "path" consistently when a string is used, and File when a file is expected.

Your original stack trace, to check path validity, was intended to catch code that was passing in windows file system paths, or the old macos paths with ":" file system paths.

src/web/core/src/main/java/org/geoserver/web/data/store/panel/FileParamPanel.java
```diff
@@ -60,7 +67,23 @@ public FileParamPanel(final String id, final IModel<String> paramValue, final IM
...
+            protected Iterator<String> getChoices(String input) {
+                try {
+                    // do we need to filter files?
+                    FileFilter fileFilter = FileParamPanel.this.fileFilter != null ?
+                            FileParamPanel.this.fileFilter.getObject() : null;
+
+                    return rootsFinder.getMatches(input, fileFilter).iterator();
+                } catch(Exception e) {
+                    // this is a helper, don't let it break the UI at runtime but log errors instead
+                    LOGGER.log(Level.SEVERE, "Failed to provide autocomplete for path " + input, e);
```
jodygarnett: Not sure SEVERE is required here, you are not going to shut down the application... can we tone this down to FINE before merge? If not I can see the logs filling up ...

src/web/core/src/main/java/org/geoserver/web/wicket/browser/FileRootsFinder.java
```diff
@@ -0,0 +1,176 @@
...
+        // build the roots
+        roots = new ArrayList<File>();
+        if (!hideFileSystem) {
+            roots.addAll(Arrays.asList(File.listRoots()));
+        }
+        Collections.sort(roots);
+
+        // TODO: find a better way to deal with the data dir
+        GeoServerResourceLoader loader = getLoader();
+        dataDirectory = loader.getBaseDirectory();
```
jodygarnett: Q: What would a better way look like? That class has the responsibility of knowing the base directory location, which is better than half a dozen classes as previous ...

src/web/core/src/main/java/org/geoserver/web/wicket/browser/FileRootsFinder.java
```diff
@@ -0,0 +1,176 @@
...
+
+        if (includeDataDir) {
+            roots.add(0, dataDirectory);
+        }
+
+        // add the home directory as well if it was possible to determine it at all
+        if(!hideFileSystem && GeoServerFileChooser.USER_HOME != null) {
+            roots.add(1, GeoServerFileChooser.USER_HOME);
+        }
```
jodygarnett: I would *love* to have the https://github.com/geoserver/geoserver/wiki/GSIP-145 environmental variables added to this list of well-known paths to browse from. Not sure if it is possible to do though ...

src/web/core/src/main/java/org/geoserver/web/wicket/browser/FileRootsFinder.java
aaime: I've already spent way more spare time than I wanted on this one, if you care about changing it make a pull request yourself. If you think this is not acceptable I'll just close the pull request and forget about it.

src/web/core/src/main/java/org/geoserver/web/wicket/browser/FileRootsFinder.java
aaime: See above, have fun doing it yourself.

src/web/core/src/main/java/org/geoserver/web/wicket/browser/FileRootsFinder.java
jodygarnett: Andrea no need to be harsh, I was asking for feedback on the API, I am not aware of a better way to get the base directory location, am I missing something?

src/web/core/src/main/java/org/geoserver/web/wicket/browser/FileRootsFinder.java
jodygarnett: For sure, I will ask on the email list if it is possible

src/web/core/src/main/java/org/geoserver/web/data/store/panel/FileParamPanel.java
jodygarnett: +1

aaime: I've merged in the meantime, will direct commit the change to INFO

PR #2704, opened 2018-01-08
Fix Documentation Typos

bgsmase:


bgsmase:
I don't think my typo changes can have had any effect on the build failing

tbarsballe:
No, I think the build may have just been unstable. Its passing now.
Thanks for the fixes.

bencaradocdavies:
Yes, it looked like an infrastructure outage so I kicked the build and it passed. Thanks @bgsmase for the fixes.

PR #2705, opened 2018-01-08
Remove out-of-date note on INSPIRE WCS guidance

bgsmase:
INSPIRE has now published guidance on using a WCS for a download service at https://inspire.ec.europa.eu/id/document/tg/download-wcs


aaime:
Yep, thanks!

PR #2706, opened 2018-01-08
Replace cobertura with jacoco for Java 9 compatibility

tbarsballe:
Replaces all usages of cobertura with jacoco.
- Added jacoco profile for generating coverage reports (previously, you just had to know what command to use to run with cobertura; a profile makes this a bit easier).

Tested on Java 9 locally, jacoco works fine.


src/pom.xml
```diff
@@ -1438,7 +1427,6 @@
       <exclude>${test.exclude.pattern}</exclude>
      </excludes>
      <forkMode>once</forkMode>
-     <argLine>-Xmx${test.maxHeapSize} -enableassertions ${jvm.opts} -Djava.awt.headless=${java.awt.headless} -Dsun.java2d.d3d=${sun.java2d.d3d} -DremoteOwsTests=${remoteOwsTests} -DquietTests=${quietTests} -Dorg.geotools.image.test.enabled=${image.tests} -Dorg.geotools.image.test.interactive=${interactive.image} -Duser.timezone=${user.timezone} -Dwindows.leniency=${windows.leniency} -XX:+TieredCompilation -XX:TieredStopAtLevel=1</argLine>
```
aaime: This is certainly bad?

tbarsballe: That got moved to a property, so that both jacoco and surefire can use it without clobbering over each other (the jacoco setup needs to append some stuff to argline during setup, but was getting overwritten by the surefire plugin config).
Since that property is only used by those plugins, this should not have an effect when they are not being used.
As far as I can tell, the surefire tests are still correctly picking up the argline (I would expect there to be at least one test failure if this were not the case)

tbarsballe:
You can run tests with coverage using `mvn test -Pjacoco`. I've been looking through the dev docs trying to find a good place to document this, but there doesn't seem to be any section on code coverage, so I think I'll end up having to add a new section somewhere, probably under maven setup.

src/pom.xml
```diff
@@ -1896,6 +1871,7 @@
   <compress-lzf.version>1.0.3</compress-lzf.version>
   <marlin.version>0.7.5-Unsafe</marlin.version>
   <postgresql.jdbc.version>42.1.1</postgresql.jdbc.version>
+  <argLine>-Xmx${test.maxHeapSize} -enableassertions ${jvm.opts} -Djava.awt.headless=${java.awt.headless} -Dsun.java2d.d3d=${sun.java2d.d3d} -DremoteOwsTests=${remoteOwsTests} -DquietTests=${quietTests} -Dorg.geotools.image.test.enabled=${image.tests} -Dorg.geotools.image.test.interactive=${interactive.image} -Duser.timezone=${user.timezone} -Dwindows.leniency=${windows.leniency} -XX:+TieredCompilation -XX:TieredStopAtLevel=1</argLine>
```
tbarsballe: argLine moved to here.

tbarsballe:
Basic javadocs added.

src/pom.xml
aaime: Ah ha!

aaime:
+1

PR #2708, opened 2018-01-11
[GEOS-8528] Add decorations info on wps map/animation download doc

Gnafu:


aaime:
Merged, thanks for the clarifications!

PR #2710, opened 2018-01-12
Update release guide with new docbuild instructions

tbarsballe:
With the new changes to the nightly doc build to get it to work with the new build server (see discussion on mailing list), the procedure for cutting a new RC has changed slightly.

Since we are directly pushing to the relevant folders, it is no longer necessary to update the ProxyPass on the doc server.


aaime:
Thanks for the docs update... and much more for the change that made for a simpler release process! ;-)

tbarsballe:
You're welcome.

PR #2717, opened 2018-01-17
[GEOS-8536] Support HeaderContribution for SRSListPage

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-8536


tbarsballe:

PR #2723, opened 2018-01-22
[GEOS-8541] Fix ground overlay east/west swap.

bradh:


aaime:
Looks good here, thanks!

PR #2724, opened 2018-01-22
[GEOS-8543] [GEOS-8544] Demo page fixes

tbarsballe:
This one wound up turning into two different bugs partway through fixing it. Still submitting as one PR, since the bugs are closely related, but included the fixes as separate commits.

[GEOS-8543](https://osgeo-org.atlassian.net/browse/GEOS-8543) - Error handling just used a PrintWriter for output, so if an error occurred during error handling, then everything already written would be included in the new error response, resulting in malformed output. Fixed, and added some (online) tests (not optimal, but there isn't really any other way to test the full servlet functionality)

[GEOS-8544](https://osgeo-org.atlassian.net/browse/GEOS-8544) - When sending a request, the demo page sends the form content to a new form in the modal window page, which then sends that to GeoServer and handled the response. Upgrading to Wicket 7.6 broke authentication for this, due to [WICKET-6211](https://issues.apache.org/jira/browse/WICKET-6211). Fixed, and added equivalent password-clearing behaviour to the fix.

### DIFF
--- a/src/web/demo/src/main/java/org/geoserver/web/demo/DemoRequestResponse.java
+++ b/src/web/demo/src/main/java/org/geoserver/web/demo/DemoRequestResponse.java
@@ -11,6 +11,7 @@ import org.apache.wicket.markup.html.form.Form;
 import org.apache.wicket.markup.html.form.HiddenField;
 import org.apache.wicket.markup.html.form.TextArea;
 import org.apache.wicket.model.IModel;
+import org.apache.wicket.model.Model;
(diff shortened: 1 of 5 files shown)


dvntucker:
Looks good to me, except for the security issues still in the page, but that's minor.

tbarsballe:
Yeah, the way the demo page is structured it is always going to some amount have security issues unless we do a full rewrite. I've mitigated as much as I could while still maintaining functionality.

tbarsballe:

PR #2725, opened 2018-01-22
[GEOS-8545] Fixed incorrect GetFeatureInfo for coverage using reproject to declared

sikeoka:
This pull request fixes incorrect WMS GetFeatureInfo responses for raster layers using "Reproject native to declared" with a different native and declared CRS.  RasterLayerIdentifier.identify makes multiple calls to the GridCoverage2DReader getOriginalEnvelope, getOriginalGridToWorld and getCoordinateReferenceSystem methods which should all be returning responses using the native CRS instead of the declared CRS when the projection policy is reproject to declared.

This pull request can be backported to 2.12.x and 2.11.x.


aaime:
Yep, that's pretty good testing, thank you!

PR #2726, opened 2018-01-22
Fix serialization issue with tile layers with empty style parameter filter

dvntucker:
This fixes an issue with tile layers that have an empty "Allowed Styles" parameter filter when saving. The issue has to do with Collections.unmodifiableSet not being serializable after the tile layer catalog switched to the SecureXstream implementation.

The issue mainly affected the backup/restore community module, which was failing when a tile layer had now styles selected except for the default.

### DIFF
--- a/src/gwc/src/test/java/org/geoserver/gwc/layer/DefaultTileLayerCatalogTest.java
+++ b/src/gwc/src/test/java/org/geoserver/gwc/layer/DefaultTileLayerCatalogTest.java
@@ -10,6 +10,7 @@ import static junit.framework.Assert.assertNotNull;
 import static junit.framework.Assert.assertNull;
 
 import java.io.File;
+import java.util.Collections;
 
 import org.apache.commons.io.FileUtils;
 import org.geoserver.catalog.impl.ModificationProxy;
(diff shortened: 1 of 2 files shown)


src/gwc/src/test/java/org/geoserver/gwc/layer/DefaultTileLayerCatalogTest.java
```diff
@@ -93,6 +94,44 @@ public void tearDown() throws Exception {
             info.setName("name1");
             info.getMimeFormats().add("image/png");
             info.getMimeFormats().add("image/jpeg");
+
+            assertNull(catalog.save(info));
+
+            original = catalog.getLayerById("id1");
+            assertEquals(info.getMimeFormats(), original.getMimeFormats());
+        }
```
tbarsballe: These scope blocks seem odd, what is the reason for them?

dvntucker: Not sure? Just copied and pasted from the previous test to be honest.

tbarsballe: Fair enough, seems fine to merge.

smithkm:
I suppose it's possible with something as broad as `Set` that it's conceivable it might extend it in a way that could be dangerous.  The same was true of `SortedSet` though  so it's not much worse than what we had.  Still, I'd feel a bit safer if our whitelist were getting tighter rather than looser.

dvntucker:
Fair enough. Changed to use UnmodifiableSet rather than Set.

PR #2727, opened 2018-01-23
[GEOS-8493] Support POST for non-SLD styles

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-8493

In order to make this fix a bit more sane, this PR also includes heavy refactoring on the StyleController class, moving most common code snippets into methods, and making each endpoint implementation more consistent with respect to behaviour, error messages, and variable names. As part of this, some methods were moved to different places in the file, so it might be a little difficult to follow the changes (the file structure was a mess before, it should be more organized now).

Functionality-wise, the only change should be the Support for POSTing non-SLD files (StyleController has pretty good test cases, and they are all happy, so I am reasonably confident in this). Some error messages _may_ have changed to be more consistent between different methods.


dvntucker:
Seems reasonable to me.

tbarsballe:
Backported to 2.12.x

PR #2728, opened 2018-01-24
GEOS-8548: Output bbox on each CRS on LayerGroups too

dromagnoli:


aaime:
Looks good but.. why does it cover WMS 1.3 only? :)

dromagnoli:
Because I'm bad :)
Fixing it.

dromagnoli:
@aaime Updated for WMS 1.1.1 too. Waiting for travis build

PR #2729, opened 2018-01-25
[JIRA GEOS-8551] - [AuthKey - WebService Key Mapper] - Add the possibility of retrieving principal Authorities from the WebService Response

afabiani:
Closes JIRA GEOS-8551


src/community/authkey/src/main/java/org/geoserver/security/GeoServerAuthenticationKeyFilter.java
```diff
@@ -24,99 +23,95 @@
...
  * 
  * @author christian
  *
  */
-public class GeoServerAuthenticationKeyFilter extends  GeoServerSecurityFilter 
-    implements AuthenticationCachingFilter, GeoServerAuthenticationFilter  
-    {
-    
-    private String authKeyMapperName,authKeyParamName;
+public class GeoServerAuthenticationKeyFilter extends GeoServerSecurityFilter
```
mbarto: Is there anything changed in this file? I can see many reformats, but it is not clear if any line of code was really changed

afabiani: None, just reformat accordingly to geoserver code conventions.

PR #2737, opened 2018-01-26
[GEOS-7233] Fix `totalFeatures` count for WFS 2.0 GetFeature requests

emerkle826:
Prevent optimizations when the combination of `count` and `startIndex` result in fewer than `count` features matching the request.

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java
@@ -251,8 +251,7 @@ public class GetFeature {
         }
         
         // WFS 2.0 validation, with locks "hits" is not allowed
-        if (WFSInfo.Version.V_20.compareTo(request.getVersion()) >= 0
-                && request.isLockRequest() && request.isResultTypeHits()) {
+        if (WFSInfo.Version.V_20.compareTo(request.getVersion()) >= 0 && request.isLockRequest() && request.isResultTypeHits()) {
             throw new WFSException("GetFeatureWithLock cannot be used with result type 'hits'", 
                     ServiceException.INVALID_PARAMETER_VALUE, "resultType");
         }
@@ -608,6 +607,23 @@ public class GetFeature {
                 
                 //update the count
                 count += size;
+
+                // collect queries required to return numberMatched/totalSize
+                // check maxFeatures and offset, if they are unset we can use the size we 
+                // calculated above
+                isNumberMatchedSkipped = meta.getSkipNumberMatched() && !request.isResultTypeHits();
+                if (!isNumberMatchedSkipped) {
+                    if (calculateSize
+                            && (queryMaxFeatures == Integer.MAX_VALUE || size < queryMaxFeatures)
+                            && offset <= 0) {
+                        totalCountExecutors.add(new CountExecutor(size));
+                    } else {
+                        org.geotools.data.Query qTotal = toDataQuery(query, filter, 0,
+                                Integer.MAX_VALUE, source, request, allPropNames.get(0), viewParam,
+                                joins, primaryTypeName, primaryAlias);
+                        totalCountExecutors.add(new CountExecutor(source, qTotal));
+                    }
+                }
                 
                 //if offset is present we need to check the size of this returned feature collection
                 // and adjust the offset for the next feature collection accordingly
@@ -629,24 +645,6 @@ public class GetFeature {
                             //adjust the offset for the next query
                             offset = Math.max(0, offset - size2);
                         }
-                    }
-                }
-
-                // collect queries required to return numberMatched/totalSize
-                // check maxFeatures and offset, if they are unset we can use the size we 
-                // calculated above
-                isNumberMatchedSkipped = meta.getSkipNumberMatched()
-                        && !request.isResultTypeHits();
-                if (!isNumberMatchedSkipped) {
-                        if (calculateSize
-                                && (queryMaxFeatures == Integer.MAX_VALUE || size < queryMaxFeatures)
-                                && offset <= 0) {
-                        totalCountExecutors.add(new CountExecutor(size));
-                    } else {
-                        org.geotools.data.Query qTotal = toDataQuery(query, filter, 0,
-                                Integer.MAX_VALUE, source, request, allPropNames.get(0), viewParam,
-                                joins, primaryTypeName, primaryAlias);
-                        totalCountExecutors.add(new CountExecutor(source, qTotal));
                     }
                 }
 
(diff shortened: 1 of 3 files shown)


src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java
```diff
@@ -251,8 +251,8 @@ public FeatureCollectionResponse run(GetFeatureRequest request)
         }
         
         // WFS 2.0 validation, with locks "hits" is not allowed
-        if (WFSInfo.Version.V_20.compareTo(request.getVersion()) >= 0
-                && request.isLockRequest() && request.isResultTypeHits()) {
+        final boolean isV_20 = WFSInfo.Version.V_20.compareTo(request.getVersion()) >= 0;
```
jodygarnett: We usually use longer local variable names, using '_' for constants.

```
final boolean WFS_2.0_SEMANTICS = WFSInfo.Version.V_20.compareTo(request.getVersion()) >= 0;
```

GeoTools [Version|http://docs.geotools.org/stable/javadocs/org/geotools/util/Version.html] also has a compareTo( version, limit) which may be appropriate depending on the new WFS 1.3.0 expectations.

aaime:
Careful here, in the history of WFS we got a mountain of tickets and complaints about needless counts that were killing performance. I'm all for correctness, but sick of getting reports about GetFeature being slow due to count queries (of course, to feel this one way has to be on the user list daily, people that have not been there will probably shrug).

Can you provide some explanation about the changes and eventually show what the overall effect might be?

src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java
```diff
@@ -640,7 +640,7 @@ public FeatureCollectionResponse run(GetFeatureRequest request)
                 if (!isNumberMatchedSkipped) {
                         if (calculateSize
                                 && (queryMaxFeatures == Integer.MAX_VALUE || size < queryMaxFeatures)
-                                && offset <= 0) {
+                                && offset <= 0 && !isV_20) {
```
aaime: Can you elaborate what in WFS 2.0 makes things different than in WFS 1.1? I cannot remember off the top of my head, maybe it's best to also clarify it in the comment (here and below)

emerkle826:
@aaime Thanks for the review. I can appreciate the concerns about performance.

As for the changes here, what I found when stepping through a GetFeature request is that, for 2.0, the `numberMatched` value was getting set to the number of features returned when the `startIndex` in the request is fewer results from the last result than the requested `count`. For example, if the total matched results of the query is 20, but the request specifies a startIndex of 15, and a count of 10, only the last 5 features are returned (15 through 19, inclusive), and the `numberMatched` in the response is `5` instead of `20`.

The spot where I added a check for 2.0 is where processing builds CountExectors for the query. The logic was checking if the size of the results (5 in this example) is less than the max requested (`count`, 10 in this example), as well as if the offset is <= 0. In this request the offset should be 15, but it is set to 0 prior because the size of the returned features is >0 (here: https://github.com/geoserver/geoserver/blob/master/src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java#L612-L617). So, based on the conditions at this point, the code only builds CountExectuors with `5`, which results in an incorrect total feature count in this case.

The second spot where I added a check for 2.0 is when the total count is calculated, and a check to see if the queries even need to be run. The logic checks to see if the feature count (number of features returned in the query, `5` in this case) is less than the maxFeatures (the max number to return, `10` in this case). Since it is, the logic determines the CountExecutors don't need to be collected and just sets the totalFeature count to `size` (`5` in this case).

Both changes result in a full query, which is what happens when the `startIndex` is further from the last result than the `count` in the request (i.e. 20 matches, startIndex of 5, count of 10). My proposal attempts to make sure the same thing happens for results on that _last page_, where the total count seems to get messed up.

To be honest, the logic in here is quite complex and I don't have the intimate history of count issues. I understand wanting to avoid full queries when not necessary. I also admit that I'm not sure if this change might break other 2.0 queries (i.e. multiple queries in a single request).

src/wfs/src/test/java/org/geoserver/wfs/v2_0/GetFeatureTest.java
```diff
@@ -157,6 +157,33 @@ public void testGetWithCount() throws Exception {
         assertEquals("15", dom.getDocumentElement().getAttribute("numberMatched"));
     }
 
+    @Test
+    public void testGetWithCountAndStartIndex0() throws Exception {
```
aaime: Removing the changes in GetFeature this test does not fail

src/wfs/src/test/java/org/geoserver/wfs/v2_0/GetFeatureTest.java
```diff
@@ -157,6 +157,33 @@ public void testGetWithCount() throws Exception {
...
+        Document dom = getAsDOM(
+            "wfs?request=GetFeature&typenames=cdf:Fifteen&version=2.0.0&service=wfs&count=5&startIndex=0");
+        XMLAssert.assertXpathEvaluatesTo("5", "count(//cdf:Fifteen)", dom);
+        assertEquals("5", dom.getDocumentElement().getAttribute("numberReturned"));
+        assertEquals("15", dom.getDocumentElement().getAttribute("numberMatched"));
+    }
+
+    @Test
+    public void testGetWithCountAndStartIndexMiddle() throws Exception {
```
aaime: Removing the changes in GetFeature this test does not fail

src/wfs/src/test/java/org/geoserver/wfs/v2_0/GetFeatureTest.java
```diff
@@ -157,6 +157,33 @@ public void testGetWithCount() throws Exception {
...
+        Document dom = getAsDOM(
+            "wfs?request=GetFeature&typenames=cdf:Fifteen&version=2.0.0&service=wfs&count=5&startIndex=7");
+        XMLAssert.assertXpathEvaluatesTo("5", "count(//cdf:Fifteen)", dom);
+        assertEquals("5", dom.getDocumentElement().getAttribute("numberReturned"));
+        assertEquals("15", dom.getDocumentElement().getAttribute("numberMatched"));
+    }
+
+    @Test
+    public void testGetWithCountAndStartIndexEnd() throws Exception {
```
aaime: But this one does :-) 
The extra test coverage of the others does not hurt, either.

aaime:
Hi @emerkle826 I've worked out a WFS version independent variation of this and made a pull request against your branch: https://github.com/emerkle826/geoserver/pull/1
I could not really find a moment of quiet and peace, wife sick and kids roaming around, so please take it with a grain of salt ;-)

aaime:
Cough cough :-p

aaime:
I would like to move on with this fix, should I do a pull request of my own?

emerkle826:
@aaime Sorry for the delay, I've been off the grid for a while. I'll get to the changes you made and apply them to this PR today.

aaime:
Awesome, thank you @emerkle826 !!!

emerkle826:
@aaime Your changes look good. I've merged them locally and tested again, everything looks good. I didn't squash the commits here (didn't want to take full credit for the work you did), but I can if a cleaner commit history would be preferable.

emerkle826:
Squashed and rebased on top of latest master.

PR #2738, opened 2018-01-26
[GEOS-8127] Improved validation to WorkspaceNewPage for duplicate names

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-8127


tbarsballe:

PR #2740, opened 2018-01-27
Fix serialization of brute force whitelist mask

emerkle826:
When using the UI to add IP addresses to the `Excluded network masks`, the serialized config that is written looks like this:
```xml
  <bruteForcePrevention>
    <enabled>true</enabled>
    <minDelaySeconds>1</minDelaySeconds>
    <maxDelaySeconds>5</maxDelaySeconds>
    <maxBlockedThreads>100</maxBlockedThreads>
    <whitelistedMasks  class="java.util.Arrays$ArrayList">
        <a class="string-array">
            <string>127.0.0.1</string>
            <string>192.168.0.1</string>
        </a>
    </whitelistedMasks>
  </bruteForcePrevention>
```

This is because the UI page (`AuthenticationPage`) that contains the brute force protection config uses `CommaSeparatedListConverter` to convert the user entered list of IPs into a `List<String>`. The problem is, the implementation uses `java.util.Arrays.asList()` which returns a private concrete class: `java.util.Arrays.ArrayList`.

There doesn't seem to be a way to use any of the `allowTypes` methods within `SecureXStream.java` to add this Class type to the allowed list, so when the config is deserialized on restarts, it throws `com.thoughtworks.xstream.security.ForbiddenClassException`.

Changing the UI implementation to create an `ArrayList<String>` changes how the config is serialized to this:
```xml
  <bruteForcePrevention>
    <enabled>true</enabled>
    <minDelaySeconds>1</minDelaySeconds>
    <maxDelaySeconds>5</maxDelaySeconds>
    <maxBlockedThreads>100</maxBlockedThreads>
    <whitelistedMasks>
        <string>127.0.0.1</string>
        <string>192.168.0.1</string>
    </whitelistedMasks>
  </bruteForcePrevention>
```
which can be deserialized correctly over restarts.

### DIFF
--- a/src/web/security/core/src/main/java/org/geoserver/security/web/auth/CommaSeparatedListConverter.java
+++ b/src/web/security/core/src/main/java/org/geoserver/security/web/auth/CommaSeparatedListConverter.java
@@ -4,7 +4,7 @@
  */
 package org.geoserver.security.web.auth;
 
-import java.util.Arrays;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


aaime:
+1

PR #2744, opened 2018-02-04
[GEOS-8571] Autocomplete support for dropdowns with long lists of values

aaime:


src/web/core/src/test/java/org/geoserver/web/GeoServerWicketTestSupport.java
```diff
@@ -118,7 +123,7 @@ public Component findComponentByContent(MarkupContainer root, Object content, Cl
         root.visitChildren(componentClass, finder);
         return finder.candidate;
     }
-    
+
```
tbarsballe: This file seems to contain only non-functional changes

aaime: Yep, will remove it from changeset

aaime:
Ah yep, the layer dropdowns for vector/raster layers are indeed a good candidate for switch, I'll look into that.

aaime:
Added support for WPS builder usage, minor CSS tweaks (in WPS there is a native and a select2 side by side, they were misaligned), and... merged!

PR #2745, opened 2018-02-05
[GEOS-8564]: Add support for writing parameters to WPS-Download

dromagnoli:


src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ppio/BinaryPPIO.java
```diff
@@ -14,6 +14,11 @@
  */
 public abstract class BinaryPPIO extends ComplexPPIO {
 
+    protected static final String TILE_WIDTH_KEY = "tilewidth";
```
aaime: BinaryPPIO has no changes related to the parameters, I'd move these out. In WCS they are specific to GeoTiff, if other formats have their own params they could be different. Suggestion: can we put them into GeoTiffPPIO for the time being, and then move them up as needed if more formats actually start needing them?

doc/en/user/source/community/wps-download/rawDownload.rst
```diff
@@ -378,3 +379,42 @@ This time, when issued (and process has finished on the server), the GET request
...
+        </wps:Data>
+    </wps:Input>
+
+GeoTIFF/TIFF supported writing parameters
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+The supported writing parameters are:
+
+ * ``tilewidth`` : Width of internal tiles, in pixels
+ * ``tileheight`` : Height of internal tiles, in pixels
+ * ``compression`` : Compression type used to store internal tiles. Most commonly used values are:
```
aaime: I'd write down the actual supported list (you can check the GeoTiff WCS 2.0 support for a more complete one)

doc/en/user/source/community/wps-download/rawDownload.rst
```diff
@@ -34,6 +34,7 @@ The parameters to set are
  * ``targetSizeX`` : size X in pixels of the output (optional, applies for raster input only)
  * ``targetSizeY`` : size Y in pixels of the output (optional, applies for raster input only)
  * ``selectedBands`` : a set of the band indices of the original raster that will be used for producing the final result (optional, applies for raster input only)
+ * ``writeParameters`` : a set of writing parameters (optional, applies for raster input only)
```
aaime: Since this is a complex input, I'd refer people to the example at the bottom of document to show how they are encoded

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ppio/BinaryPPIO.java
dromagnoli: Well. at the very beginning I have put that on GeoTIFF PPIO indeed. Then I have moved up to BinaryPPIO thinking about more formats starting supporting them in the future. However, we can move up them afterwards, so I'm going to put them back down on GeoTIFF only.

doc/en/user/source/community/wps-download/rawDownload.rst
dromagnoli: Done

PR #2746, opened 2018-02-07
[GEOS-8382] Add isolated workspaces concept to GeoServer

nmco:
Associated issue:
- https://osgeo-org.atlassian.net/browse/GEOS-8382

Associated mailing list discussion:
- https://sourceforge.net/p/geoserver/mailman/message/35658017
- https://sourceforge.net/p/geoserver/mailman/message/36097736

This pull request add isolated workspaces concept to GeoServer, this brings two mains features:
- possibility to create an workspace whose content will be _hidden_ in global services and only available to the virtual services bind to the isolated workspace
- possibility to reuse a namespace URI, **which allows the publication of the same complex feature multiple times when using App-Schema**

Must of the added \ changed code is related with testing and **all the changes are backward compatible, i.e. no existing functionality are affected by this new functionality**.
The main changes happen in:
- ``CatalogImpl`` the allow the catalog to handle multiple workspaces\namespaces that use the same URI
- ``DefaultCatalogFacade`` to allow the handling of namespaces with the same UI
- ``IsolatedCatalogFacade`` catalog facade wrapper that implements all the behavior necessary to handle isolated workspaces, having all this behavior in a single place makes the code cleaner and focused
- ``WorkspaceNewPage`` to handle isolated funcionality

This pull request adds documentation and updates existing deprecated screen-shoots.

Is also worth mentioning that this pull request updates the demo page links generation for complex feature types to include the workspace in the URL path, this was not done because of the NULL namespaces bug that has been fixed.


src/main/src/main/java/org/geoserver/catalog/impl/CatalogImpl.java
```diff
@@ -1118,10 +1129,13 @@ public ValidationResult validate(NamespaceInfo namespace, boolean isNew) {
...
             throw new IllegalArgumentException( "Namespace with prefix '" + namespace.getPrefix() + "' already exists.");
         }
-        
-        existing = getNamespaceByURI( namespace.getURI() );
-        if ( existing != null && !existing.getId().equals( namespace.getId() ) ) {
-            throw new IllegalArgumentException( "Namespace with URI '" + namespace.getURI() + "' already exists.");
+
+        if (!namespace.isIsolated()) {
+            // not an isolated namespace \ workplace so we need to check for duplicates
+            existing = getNamespaceByURI(namespace.getURI());
```
aaime: Say we are adding a non isolated workspace that "conflicts" with an isolated one, it should be accepted no?. What prevents getNamespaceByURI from returning the isolated one? I checked quickly in IsolatedCatalogFacade and did not see anything that would prevent this issue.

aaime: Ok, found it in DefaultCatalogFacade... a bit surprising though, wouldn't it make more sense to keep that logic in the IsolatedCatalogFacade wrapper instead? Ideally, having no changes in DefaultCatalogFacade at all, leaving all in the wrapper, would ease plugging other catalog facade implementation by simple wrapping

src/main/src/main/java/org/geoserver/catalog/util/LegacyCatalogImporter.java
```diff
@@ -282,7 +282,8 @@ void importCatalog(File catalogFile) throws FileNotFoundException,
         reader.read(Files.asResource(catalogFile));
 
         // build all the catalog objects that can be read from the catalog.xml file
-        importNamespaces(factory, reader.namespaces());
+        importNamespaces(factory, reader.namespaces(), false);
+        importNamespaces(factory, reader.isolatedNamespaces(), true);
```
aaime: A true legacy catalog won't have isolated workspaces. I'm guessing this support is there due to tests (they are still writing catalogs using the old structure, aren't they?)

src/main/src/test/java/org/geoserver/data/test/SystemTestData.java
```diff
@@ -346,28 +346,53 @@ XStreamPersister createXStreamPersister() {
...
+
+        // let's see if the workspace already exists
         WorkspaceInfo ws = catalog.getWorkspaceByName(name);
         if (ws == null) {
+            // new workspace, we need to create a new one
             ws = catalog.getFactory().createWorkspace();
             ws.setName(name);
+            ws.setIsolated(isolated);
             catalog.add(ws);
+        } else {
```
aaime: Given the method name, I would have preferred to see an exception (trying to add an already existing workspace should result in an error?), but the old code was confusing too, was ignoring the add and was eventually aligning the namespace.
So I guess it's all right :)

src/restconfig/src/main/java/org/geoserver/rest/catalog/NamespaceController.java
```diff
@@ -114,6 +94,7 @@ public RestWrapper getNamespaces() {
         if ( catalog.getWorkspaceByName( namespace.getPrefix() ) == null ) {
             WorkspaceInfo ws = catalog.getFactory().createWorkspace();
             ws.setName( namespace.getPrefix() );
+            ws.setIsolated(namespace.isIsolated());
```
aaime: The name of the method thru me off balance, was wondering what was going on here... namespaceGet, but it's actually a namespacePost :-)
I'd fix the method name, but I understand it's not your fault for it being there. @tbarsballe git blame points at you, probably some mass refactor side effect by the commit message.

src/restconfig/src/main/java/org/geoserver/rest/catalog/NamespaceController.java
```diff
@@ -152,6 +133,17 @@ public void namespacePut(
     
             new CatalogBuilder(catalog).updateNamespace(nsi, namespace);
             catalog.save(nsi);
+
+            // make sure workspace isolation value is updated too
+            WorkspaceInfo workspace = catalog.getWorkspaceByName(namespace.getPrefix());
```
aaime: In the catalog there is already a NamespaceWorkspaceConsistencyListener, wondering if code like this may be better put there?

src/restconfig/src/main/java/org/geoserver/rest/catalog/WorkspaceController.java
```diff
@@ -170,6 +171,17 @@ public void workspacePut(
             new CatalogBuilder(catalog).updateWorkspace(wks, workspace);
             catalog.save(wks);
         }
+
+        // make sure namespace isolation value is updated too
+        NamespaceInfo namespace = catalog.getNamespaceByPrefix(workspace.getName());
```
aaime: Same comments as with the namespace controller

src/web/core/src/main/java/org/geoserver/web/data/resource/ResourceConfigurationPage.java
```diff
@@ -68,8 +68,7 @@ public ResourceConfigurationPage(String workspaceName, String layerName) {
                 // with a layer page open in another tab/window
                 throw new RuntimeException("Could not find workspace " + workspaceName);
             }
-            String nsURI = ns.getURI();
-            layer = getCatalog().getLayerByName(new NameImpl(nsURI, layerName));
+            layer = getCatalog().getLayerByName(ns.getName() + ":" + layerName);
```
aaime: This change is interesting.. so the other method would not work properly in the context of a isolated workspace? Ideally the two should behave the same, if not, some copius javadoc warning people should be added (and maybe some migration guide for code sitting outside GeoServer)

src/web/core/src/main/java/org/geoserver/web/data/workspace/WorkspaceEditPage.java
```diff
@@ -113,6 +113,13 @@ private void init(WorkspaceInfo ws) {
         wsModel = new WorkspaceDetachableModel(ws);
 
         NamespaceInfo ns = getCatalog().getNamespaceByPrefix( ws.getName() );
+
+        if (ns == null) {
+            // unfortunately this may happen if the namespace associated to the workspace was deleted or never created
+            throw new RuntimeException(String.format(
+                    "Workspace '%s' associated namespace doesn't exists.", ws.getName()));
```
aaime: This will show up as the error message for the user... is there any suggestion we can provide to fix this also? Like I don't know "Go to the workspace page and save again, assuming that will fix the misalignment, or "use the rest api to fix this")

src/web/core/src/main/java/org/geoserver/web/data/workspace/WorkspaceEditPage.java
```diff
@@ -171,10 +182,16 @@ private void saveWorkspace() {
...
-        
+
+        namespaceInfo.setIsolated(workspaceInfo.isIsolated());
+
         // sync up workspace name with namespace prefix, temp measure until the two become separate
         namespaceInfo.setPrefix(workspaceInfo.getName());
-        
+
+        // validate workspace and namespace before updating them
+        catalog.validate(workspaceInfo, false).throwIfInvalid();
```
aaime: Better to turn this into a validator and generate a "nice" error message instead of making the page fail fully and display a stack trace?

aaime:
I don't see a proposal, but there were two deep discussion threads and it seems all concerns were addressed there already, to everyone's satisfaction.

src/restconfig/src/main/java/org/geoserver/rest/catalog/NamespaceController.java
tbarsballe: Yeah, that commit was most likely part of a mass refactor, probably moving the module from rest-ng to rest.
As for the method name, I assume it was some copy-paste that got missed during the code sprint. I've found and fixed at least one other similar issue elsewhere in the REST api. At least the MVC annotations make that sort of issue a bit easier to spot.

nmco:
A proposal has been writing for this:
https://github.com/geoserver/geoserver/wiki/GSIP-165---Add-isolated-workspaces-concept-to-GeoServer

src/web/core/src/main/java/org/geoserver/web/data/resource/ResourceConfigurationPage.java
nmco: I don't think there is a need for a migration guide, I made this change because I want to support both the isolated and non isolated workspaces behavior.

Javadoc of ``getLayerByName`` method updated.

src/web/core/src/main/java/org/geoserver/web/data/workspace/WorkspaceEditPage.java
nmco: If this situation happens it means that the user manually edited the data directory or that the system went in corrupted state. The is nor much the user can do except:

1. Manually try to fix the data directory.
2. Remove the workspace and recreate it.

src/web/core/src/main/java/org/geoserver/web/data/workspace/WorkspaceEditPage.java
nmco: This is already done, the exception message will be catch and show to the user:
image

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/web/Gml311LinksTest.java
```diff
@@ -14,12 +14,12 @@
...
-                "http://localhost:80/context/ows?service=WFS&amp;version=1.1.0&amp;request=GetFeature&amp;typeName=gsml:GeologicUnit&amp;outputFormat=gml3&amp;maxFeatures=50",
-                "http://localhost:80/context/ows?service=WFS&amp;version=1.1.0&amp;request=GetFeature&amp;typeName=ex:FirstParentFeature&amp;outputFormat=gml3&amp;maxFeatures=50",
-                "http://localhost:80/context/ows?service=WFS&amp;version=1.1.0&amp;request=GetFeature&amp;typeName=ex:SecondParentFeature&amp;outputFormat=gml3&amp;maxFeatures=50",
-                "http://localhost:80/context/ows?service=WFS&amp;version=1.1.0&amp;request=GetFeature&amp;typeName=ex:ParentFeature&amp;outputFormat=gml3&amp;maxFeatures=50",
-                "http://localhost:80/context/ows?service=WFS&amp;version=1.1.0&amp;request=GetFeature&amp;typeName=om:Observation&amp;outputFormat=gml3&amp;maxFeatures=50"));
+        super(Arrays.asList("http://localhost:80/context/gsml/ows?service=WFS&amp;version=1.1.0&amp;request=GetFeature&amp;typeName=gsml:MappedFeature&amp;outputFormat=gml3&amp;maxFeatures=50",
```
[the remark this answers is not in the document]
aaime: It does, thank you

src/main/src/main/java/org/geoserver/catalog/impl/CatalogImpl.java
aaime: I see... however other facades should be modified as well, it would be a change in expected behavior (so javadocs, migration guide). The other possibility would be to iterate over all namespaces in the wrapper, but I'm not sure what performance effect it would have.

aaime: Oh, another possibility could be to extend the Facade with a getNamespacesByURI method whose default implementation is a simple iterator, but DefaultCatalogFacade could have a smarter version of it (there should not be many matching the same URI, right?)

src/main/src/main/java/org/geoserver/catalog/util/LegacyCatalogImporter.java
aaime: I understand

nmco:
``DefaultCatalogFacade`` changes reverted, and the filtering of isolated namespaces moved to ``IsolatedCatalogFacade`` :smiley:.

aaime:
Thank you!

nmco:
After chatting with Andrea, it turns out that handling the synchronization of isolation between workspaces and namespaces after a modify event is not that hard to handle in the ``NamespaceWorkspaceConsistencyListener``  if plugged into the post modify event. The code has been updated accordingly.

aaime:
Awesome, thanks for that update. Looks good to me :)

PR #2748, opened 2018-02-09
[GEOS-8607] Fix warning when parsing rules on a non global group

mwerlen:
When parsing security rules, class DataAccessRuleDao log a warn when a rule doesn't match an existing layer. This check doesn't handle layergroups in workspaces, only workspaces, layers with or without
workspaces and layergroups without workspace.
This fix adds a check on layergroups in workspace in order not to log a false warning.

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/impl/DataAccessRuleDAO.java
+++ b/src/main/src/main/java/org/geoserver/security/impl/DataAccessRuleDAO.java
@@ -24,6 +24,7 @@ import org.geoserver.platform.GeoServerExtensions;
 import org.geoserver.platform.resource.Resource;
 import org.geoserver.security.AccessMode;
 import org.geoserver.security.CatalogMode;
+import org.geotools.feature.NameImpl;
 import org.geotools.util.logging.Logging;
 
 /**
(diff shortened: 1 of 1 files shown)


tbarsballe:
Thanks for the fix. Per [CONTRIBUTING.md](https://github.com/geoserver/geoserver/blob/master/CONTRIBUTING.md), could you please add a test case?

mwerlen:
Hi Torben,

I don't know how to test this modification as there is no side effects except for an absence of log. Logger is not injectable as it uses a static class. I have no access to underlying log framework from the test class.
To test it efficiently I would need to extract my modification in a separate method. It looks like over-engineering to me.

Do you have any hint (or example) about how to test such a modification ?

src/main/src/main/java/org/geoserver/security/impl/DataAccessRuleDAO.java
```diff
@@ -163,7 +164,9 @@ DataAccessRule parseDataAccessRule(String ruleKey, String ruleValue) {
         if(layerName != null) {
             if (!ANY.equals(root) && rawCatalog.getWorkspaceByName(root) == null)
                 LOGGER.warning("Namespace/Workspace " + root + " is unknown in rule " + rule);
-            if (!ANY.equals(layerName) && rawCatalog.getLayerByName(layerName) == null)
+            if (!ANY.equals(layerName) &&
+                    !(   rawCatalog.getLayerByName(new NameImpl(root, layerName)) != null
+                    || rawCatalog.getLayerGroupByName(root, layerName) != null ))
```
tbarsballe: Given the negation, wouldn't this be clearer? The double negation makes is a bit awkward to read at a glance.
```
if (!ANY.equals(layerName) &&
        rawCatalog.getLayerByName(new NameImpl(root, layerName)) == null
        && rawCatalog.getLayerGroupByName(root, layerName) == null )

```

mwerlen:
Thanks Torben,

I've updated the MR with your proposal. It's effectivly clearer.

tbarsballe:
Backported to 2.13.x and 2.12.x

mwerlen:
Thanks a lot Torben for merging and backporting.

PR #2749, opened 2018-02-09
[OAuth2] Minor refactoring of OAuth Filters

afabiani:


src/community/security/oauth2-geonode/src/main/java/org/geoserver/security/oauth2/GeoNodeOAuthAuthenticationFilter.java
```diff
@@ -21,4 +26,31 @@ public GeoNodeOAuthAuthenticationFilter(SecurityNamedServiceConfig config,
         super(config, tokenServices, oauth2SecurityConfiguration, oauth2RestTemplate);
     }
 
+    @Override
+    protected String getCustomSessionCookieValue(HttpServletRequest request)  {
```
mbarto: Cannot we put this in the extended class? It looks like it's repeated in every specialized class.

afabiani: @mbarto it seems like this but it is not. GeoNode requires special check to avoid Web UI session.

In general this method should be used by different implementations in order to check whether special UI/UX session should be taken into account or not.

afabiani: @mbarto you are right, there should be a difference at line 40 on each file, but it looks like I lost it during the commit.

afabiani: @mbarto changes committed now.

PR #2751, opened 2018-02-11
[GEOS-8580] Use appropriate editors for coverage parameters list

aaime:


aaime:
Note to self, there is a screenshot to update here: http://docs.geoserver.org/latest/en/user/data/raster/imagemosaic/configuration.html

aaime:
Screenshot updated, and.. merging!

PR #2760, opened 2018-02-19
GWC Configuration API change compatibility

tbarsballe:
GeoServer side of https://github.com/GeoWebCache/geowebcache/pull/593


smithkm:
Triggering Travis after the upstream merge

PR #2763, opened 2018-02-22
Changing release process - create new branch during beta instead of RC

tbarsballe:
For an upcoming GSIP.
Also includes some other minor release process changes discovered during the 2.13-beta release.

See also: https://github.com/geotools/geotools/pull/1811


aaime:
Looks good to me!

tbarsballe:
Updated, based on mailing list discussions. The beta has been replaced with the release candidate,
which now occurs a month before the release (functionally this the same as the initially proposed change, and is just a matter of different terminology).

PR #2771, opened 2018-02-27
[GEOS-8603] Fix StackOverflowError for nesting decorated catalog Infos

emerkle826:


smithkm:
CI failure looks like it may have been a glitch, restarting

src/main/src/main/java/org/geoserver/security/decorators/DefaultSecureCatalogFactory.java
```diff
@@ -29,32 +37,145 @@ public boolean canSecure(Class clazz) {
...
+        String msg = String.format("Double securing %s: proxies: %,d, SecuredDataStoreInfos: %,d",
+                store.getName(), proxies, secured);
+        LOGGER.warning(msg);
+    }
+
+    private WMTSLayerInfo unwrap(WMTSLayerInfo object) {
+        WMTSLayerInfo unwrapped = ModificationProxy.unwrap(object);
+        boolean isProxy = object != unwrapped;
+        if(unwrapped instanceof SecuredWMTSLayerInfo) {
+            logDoubleWrap(unwrapped, object);
```
smithkm: This confused me for a bit but if I understand what's going on this is to ensure it's not wrapped by a security proxy before wrapping it with one.  So the warning about being double wrapped is that this is about to be wrapped, not the combination of modification and security wrappers?

src/main/src/main/java/org/geoserver/security/decorators/DefaultSecureCatalogFactory.java
```diff
@@ -29,32 +37,145 @@ public boolean canSecure(Class clazz) {
...
+            return unwrap(((SecuredWMTSLayerInfo)unwrapped).unwrap(WMTSLayerInfo.class));
+        }
+        if (isProxy) {
+            object = ModificationProxy.create(unwrapped, WMTSLayerInfo.class);
```
smithkm: Is it possible for the modification proxy to have had modifications made to it? when this happens.  They would be lost in this process.

src/main/src/main/java/org/geoserver/security/decorators/DefaultSecureCatalogFactory.java
emerkle826: The warning is more to indicate that somewhere, a secured Object is being doubly secured. I have not seen this for some of the Info Object types handled in this class, but it has been observed for at least DataStoreInfo objects. In theory, we shouldn't see this log generated, as I would think we wouldn't want to try to secure an already secured Object.

src/main/src/main/java/org/geoserver/security/decorators/DefaultSecureCatalogFactory.java
emerkle826: Looking at ModificationProxy more closely, I see your point.

emerkle826: @smithkm 
I've retested this without the unwrapping in DefaultSecurityCatalogFactory and my issue seems to still be resolved. I'll leave the logging in, in case a similar issue is hit in the future. It seems the bulk of what I need is in making sure SecuredFeatureStoreInfo *undecorates* DataStoreInfo objects on **setStore()**, to make it the reciprocal of the secure decorating done in **getStore()**. This is consistent with how SecuredLayerInfo handles things here: https://github.com/geoserver/geoserver/blob/master/src/main/src/main/java/org/geoserver/security/decorators/SecuredLayerInfo.java#L30-L58

smithkm:
So I had a thought on how to handle the modification proxies safely: https://github.com/smithkm/geoserver/commit/75a5a0a2392ee7387e748240205009bf14eb5f46

emerkle826:
@smithkm That looks good to me, however I've changed the code here so that it doesn't unwrap ModificationProxy instances, outside of logging things if DefaultSecureCatalogFactory tries to wrap an already secured object. The main reason I chose to do it this way is that unwrapping proxies in that class would just be covering up an issue somewhere else, and unless someone is paying attention to GeoServer logs, it would go undetected. The way I have it now, if a secured object slips through and gets re-wrapped, it will eventually stack overflow, but we'll also have logs prior to that indicating which data type, which should make it easier to track down where things need to be fixed.

That said, I see value in the changes you propose. I can include it in this PR, but the functionality wouldn't necessarily be used at the moment.

src/main/src/main/java/org/geoserver/security/decorators/DefaultSecureCatalogFactory.java
```diff
@@ -29,32 +38,125 @@ public boolean canSecure(Class clazz) {
...
+            }
+            object = unwrapped;
+        }
+
+        String msg = String.format("Double securing %s: proxies: %,d, SecuredDataStoreInfos: %,d",
+                store.getName(), proxies, secured);
+        LOGGER.warning(msg);
+    }
+
+    private WMTSLayerInfo unwrap(WMTSLayerInfo object) {
```
smithkm: `unwrap` methods that just return what they are given are a bit confusing.  If they are intended to always just be logging then maybe a different name would help.  Otherwise if they are more hooks to be used to deal with future bugs, maybe some documentation explaining your thinking.

emerkle826: I've renamed the `unwrap` methods to `logIfSecured`, as that is what they do now.

emerkle826:
Unit tests added.

smithkm:
This looks like it should be a fairly easy backport to 2.13.x and 2.12.x.

emerkle826:
@smithkm Should I open separate PRs to backport this to 2.13.x and 2.12.x?

smithkm:
That's generally how I do backports.  I like to give Travis a chance to build before merging.

PR #2772, opened 2018-02-27
GEOS-8599: WPS download at granule's resolution

dromagnoli:


src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/GridGeometryProvider.java
```diff
@@ -0,0 +1,395 @@
...
+                }
+                return getNativeResolutionGridGeometry();
+            }
+            // Initialize resolution with infinite numbers
+            double[] resolution = new double[] {Double.POSITIVE_INFINITY, Double.POSITIVE_INFINITY};
+            ReferencedEnvelope envelope = provider.getBestResolution(features, resolution);
+            AffineTransform at = new AffineTransform(resolution[0], 0, 0, -resolution[1], envelope.getMinX(), envelope.getMaxY());
+            MathTransform tx = ProjectiveTransform.create(at);
+            
+            return new GridGeometry2D(PixelInCell.CELL_CORNER, tx, envelope, GeoTools.getDefaultHints());
```
aaime: What would happen if there are no granules captured by the query (and the resolutions remain at their infinite value)? I guess at some point an error should be returned of course, I'm just worried of whether the error would be understandable, or would look like a random internal error.

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/GridGeometryProvider.java
```diff
@@ -0,0 +1,395 @@
...
+
+        // Add the filter if specified
+        if (filter != null) {
+            filters.add(filter);
+        }
+
+        // Set the query filter
+        if (!filters.isEmpty()) {
+            query = new Query();
+            query.setFilter(FeatureUtilities.DEFAULT_FILTER_FACTORY.and(filters));
```
aaime: An "and of one" is probably working but looks odd.  Since you have a list of filters, this method will do the right thing for you (you could dodge the if completely): https://github.com/geoserver/geoserver/blob/bab3171ef936628d28d6cbd0377a58f648716e9c/src/main/src/main/java/org/geoserver/catalog/Predicates.java#L240

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/GridGeometryProvider.java
```diff
@@ -0,0 +1,395 @@
...
+     * @throws TransformException
+     * @throws IOException
+     */
+    private GridGeometry2D getNativeResolutionGridGeometry()
+            throws TransformException, IOException {
+        final ReferencedEnvelope roiEnvelope = roiManager != null
+                ? new ReferencedEnvelope(roiManager.getSafeRoiInNativeCRS().getEnvelopeInternal(),
+                        roiManager.getNativeCRS()) : null;
+        ScaleToTarget scaling = new ScaleToTarget(reader, roiEnvelope);
+        scaling.setTargetSize(null, null);
```
aaime: Do we actually need to call this method? I have the impression it does not actually end up changing the state of the ScaleToTarget object.

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/ROIManager.java
```diff
@@ -214,7 +214,11 @@ public CoordinateReferenceSystem getTargetCRS() {
         return targetCRS;
     }
 
-    public boolean isRoiCrsEqualsTargetCrs() {
+    public CoordinateReferenceSystem getNativeCRS() {
+		return nativeCRS;
```
aaime: Tabs

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/RasterDownload.java
```diff
@@ -178,170 +185,441 @@ public Resource execute(String mimeType, final ProgressListener progressListener
...
+                // Since we have imposed a target size, delegate GridCoverageRenderer to do all the dirty job
+                requestedGridGeometry = new GridGeometry2D(
+                        new GridEnvelope2D(0, 0, targetSizeX, targetSizeY), targetEnvelope);
+                if (LOGGER.isLoggable(Level.FINE)) {
+                    LOGGER.log(Level.FINE, "Target size has been specified. Setting up requested GridGeometry: "
+                                    + requestedGridGeometry.toString());
+                }
             }
 
+            readParameters = updateReadParams(readParameters, parameterDescriptors, bandIndices, filter);
```
aaime: A comment explaining "update how" would be nice (e.g., add filter, enforce deferred loading and the like)

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/RasterDownload.java
```diff
@@ -178,170 +185,441 @@ public Resource execute(String mimeType, final ProgressListener progressListener
...
+        final int requestedMaxY = requestedRange.getHigh(1);
+
+        final int imageW = rasterData.getWidth();
+        final int imageH = rasterData.getHeight();
+        final int minX = rasterData.getMinX();
+        final int minY = rasterData.getMinY();
+        final int maxX = minX + imageW - 1;
+        final int maxY = minY + imageH - 1;
+
+        if (((imageW + ((minX >= 0) ? 0 : minX)) >= requestedW)
```
aaime: I don't get while the min values are summed to the width, shouldn't it be a simple width vs width comparison?
I have the impression that building two java.awt.Rectangle and doing a contains test would probably be shorter to write and easier to understand (question, can there be the case where an image is outside the requested area? or one where the image is just translated vs what was requested?)

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/RasterDownload.java
```diff
@@ -178,170 +185,441 @@ public Resource execute(String mimeType, final ProgressListener progressListener
...
+        final RenderedImage mosaic = MosaicDescriptor.create(new RenderedImage[] { rasterData },
+                MosaicDescriptor.MOSAIC_TYPE_BLEND, null, null, null, backgroundValues, jaiHints);
+
+        // Setting up a gridCoverage on top of the new image
+        final Envelope envelope = new GeneralEnvelope(
+                new GridEnvelope2D(0, 0, requestedW, requestedH), PixelInCell.CELL_CENTER,
+                requestedGridGeometry.getGridToCRS(),
+                requestedGridGeometry.getCoordinateReferenceSystem());
+
+        return GC_FACTORY.create("mosaiced", mosaic, envelope);
```
aaime: "mosaiced"?

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/RasterDownload.java
```diff
@@ -178,170 +185,441 @@ public Resource execute(String mimeType, final ProgressListener progressListener
...
+            GridGeometry2D requestedGridGeometry, double[] backgroundValues) {
+
+        final GridEnvelope requestedRange = (GridEnvelope) requestedGridGeometry.getGridRange();
+        final int requestedW = requestedRange.getSpan(0);
+        final int requestedH = requestedRange.getSpan(1);
+
+        ImageLayout layout = new ImageLayout();
+        // Setting output image layout
+        // I can impose 0,0 as the origin since we are working under
+        // that assumption as part of the reader request contract
```
aaime: reader request contract? Afaik there is no such a thing, believe I've modified mosaic to avoid super-large initial offsets just recently to avoid math issues.

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/RasterDownload.java
```diff
@@ -178,170 +185,441 @@ public Resource execute(String mimeType, final ProgressListener progressListener
...
+
+            // Do the reader supports BANDS param?
+            for (GeneralParameterValue readParameter : readParameters) {
+                if (AbstractGridFormat.BANDS.getName()
+                        .equals(readParameter.getDescriptor().getName())) {
+                    Object bands = ((ParameterValue) readParameter).getValue();
+                    if (bands != null && reader.getFormat() != null
+                            && reader.getFormat().getReadParameters().getDescriptor().descriptors()
+                                    .contains(AbstractGridFormat.BANDS)) {
+                        // Band selection will be made at reader's level.
```
aaime: "has been made", at this point the read already happened

src/community/wps-download/src/test/java/org/geoserver/wps/gs/download/DownloadProcessTest.java
```diff
@@ -939,11 +955,11 @@ public void testDownloadRasterSelectedBandsScaledWithROI() throws Exception {
             // check envelope
             Assert.assertEquals(-130.88669845369998,
                     gc.getEnvelope().getLowerCorner().getOrdinate(0), 1E-6);
-            Assert.assertEquals(48.611129008700004, gc.getEnvelope().getLowerCorner()
+            Assert.assertEquals(48.5552612829, gc.getEnvelope().getLowerCorner()
```
aaime: Can you elaborate on these test value changes? I'm guessing it's a different adjustment between having the requested bbox or the requested resolution, when the two are inconsistent, e.g, not leading to an integer number of pixels?

src/community/wps-download/src/test/java/org/geoserver/wps/gs/download/DownloadProcessTest.java
```diff
@@ -1674,6 +1690,185 @@ public void testWrongOutputFormat() throws Exception {
...
+            if (reader != null) {
+                reader.dispose();
+            }
+
+            // clean up process
+            resourceManager.finished(resourceManager.getExecutionId(true));
+        }
+
+        // Download native resolution 2
+        filter = FeatureUtilities.DEFAULT_FILTER_FACTORY.and(FeatureUtilities.DEFAULT_FILTER_FACTORY.lessOrEqual(property,
```
aaime: This is hard to read, could you put the FeatureUtilities.DEFAULT_FILTER_FACTORY into a local variable? (e..g, ``FilterFactory2 ff = FeatureUtilities.DEFAULT_FILTER_FACTORY``)

src/community/wps-download/src/test/java/org/geoserver/wps/gs/download/DownloadProcessTest.java
```diff
@@ -1674,6 +1690,185 @@ public void testWrongOutputFormat() throws Exception {
...
+                FeatureUtilities.DEFAULT_FILTER_FACTORY.literal(10000)),FeatureUtilities.DEFAULT_FILTER_FACTORY.greaterOrEqual(property,
+                        FeatureUtilities.DEFAULT_FILTER_FACTORY.literal(1000)));
+        
+        rasterZip = downloadProcess.execute(getLayerId(MIXED_RES), // layerName
```
aaime: These large blocks of code are copy/pasted, a single method taking the filter and the expected resolution would probably be better.

aaime: And oh, maybe also a check on the output CRS? The input mosaic seems to be heterogeneous, if there is no target CRS specified, check that the resulting CRS is the native one too.

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/GridGeometryProvider.java
dromagnoli: Not sure to understand this comment.
Isn't the 
            if (features == null || features.isEmpty()) 
                return getNativeResolutionGridGeometry();
check above, already dealing with that?

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/GridGeometryProvider.java
dromagnoli: Cool. Thanks for the suggestion. going to update

src/community/wps-download/src/test/java/org/geoserver/wps/gs/download/DownloadProcessTest.java
dromagnoli: Well, right now, once you specify a target size, read and resize are made by GridCoverageRenderer. Before the PR, there was different code paths involved (so a read followed by a resize).

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/RasterDownload.java
dromagnoli: fixed

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/GridGeometryProvider.java
```diff
@@ -0,0 +1,395 @@
...
+
+    private static final Logger LOGGER = Logging.getLogger(GridGeometryProvider.class);
+    
+    /**
+     * A trivial CRS cache in order to avoid several CRS.decode for the same EPSG 
+     */
+    class CRSCache {
+        Map<String, CoordinateReferenceSystem> cache = new HashMap<String, CoordinateReferenceSystem>();
+        
+        public CoordinateReferenceSystem getCrs(String crsId) throws NoSuchAuthorityCodeException, FactoryException {
```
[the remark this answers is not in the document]
dromagnoli: Updated, by having the catalog as a field of the rasterDownload

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/RasterDownload.java
dromagnoli: Fixed

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/RasterDownload.java
dromagnoli: This is a check to deal with some corner cases occurred in a different project using wps.
I would leave it as is.

aaime: Ok

src/community/wps-download/src/test/java/org/geoserver/wps/gs/download/DownloadProcessTest.java
aaime: Ok

dromagnoli:
All the feedbacks have been handled and a new PR has been made
https://github.com/geoserver/geoserver/pull/2786

PR #2775, opened 2018-02-28
[GEOS-8606]: RawDataEncoderDelegate not closing the stream after copy

dromagnoli:


src/extension/wps/wps-core/src/main/java/org/geoserver/wps/RawDataEncoderDelegate.java
```diff
@@ -53,7 +53,13 @@ public void encode(ContentHandler output) throws Exception {
     }
 
     public void encode(OutputStream os) throws IOException {
-        IOUtils.copy(rawData.getInputStream(), os);
+        InputStream stream = null;
```
aaime: try-with-resources anyone? :)

PR #2779, opened 2018-03-01
[GEOS-8611] Allow DescribeDomain to return a selection of dimensions

aaime:


src/community/wmts-multi-dimensional/src/main/java/org/geoserver/gwc/wmts/MultiDimensionalExtension.java
```diff
@@ -162,15 +175,44 @@ private Domains getDomains(SimpleConveyor conveyor) throws Exception {
...
+        if (requestedDomains == ALL_DOMAINS || requestedDomains.contains(SPACE_DIMENSION)) {
+            spatialDomain = DimensionsUtils.getBounds(resource, filter);  
+        } 
         // encode the domains
-        return new Domains(dimensions, layerInfo, boundingBox, SimplifyingFilterVisitor.simplify(filter));
+        return new Domains(dimensions, layerInfo, spatialDomain, SimplifyingFilterVisitor.simplify(filter));
+    }
+
+    
+
```
nmco: Extra white lines.

src/community/wmts-multi-dimensional/src/main/java/org/geoserver/gwc/wmts/MultiDimensionalExtension.java
```diff
@@ -162,15 +175,44 @@ private Domains getDomains(SimpleConveyor conveyor) throws Exception {
...
+        dimensionFilterBuilder.appendFilters(startAttribute, endAttribute, restrictionList);
+        return filterFactory.and(filter, dimensionFilterBuilder.getFilter());
+    }
+
+    private Set<String> getRequestedDomains(Object domains) {
+        if (domains == null) {
+            return ALL_DOMAINS;
+        }
+
+        String[] domainNames = domains.toString().trim().split("\\s*,\\s*");
```
nmco: If ever domains are an empty string ``""``, the split will return an array with an empty string element, is it worth to add also an empty check in the if above ?

src/community/wmts-multi-dimensional/src/main/java/org/geoserver/gwc/wmts/dimensions/CoverageDimensionsReader.java
```diff
@@ -162,6 +162,22 @@ public String getGeometryAttributeName() {
                 throw new RuntimeException("Error reading domain values.", exception);
             }
         }
+
+        @Override
+        public ReferencedEnvelope getBounds(Filter filter) {
+            try {
+                GranuleSource source = reader.getGranules(reader.getGridCoverageNames()[0], true);
+                Query query = new Query(null, filter);
+                if (filter != null) {
```
nmco: I'm not sure, but if we provide the filter in the constructor do we need do set it here ?

doc/en/user/source/community/wmts-multidimensional/index.rst
```diff
@@ -154,14 +154,17 @@ This operation is useful to understand which domains are available in our layer
...
-   * - BBOX=minx,miny,maxx,maxy
+   * - bbox=minx,miny,maxx,maxy
      - No
      - Bounding box corners (lower left, upper right) in CRS units
    * - DimensionIdentifier
      - No
      - At most one per dimension, a range described as min/max, restricting the domain of this dimension
+   * - Domains
+     - No
+     - A comma separated list of domain names to be returned, in case only a subset is required. The space domain is identified by "bbox".
```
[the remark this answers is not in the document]
aaime: Yep, it's what went in the BoundingBox element, which can be removed by explicitly specifying the desired domains

src/community/wmts-multi-dimensional/src/main/java/org/geoserver/gwc/wmts/MultiDimensionalExtension.java
aaime: will remove

src/community/wmts-multi-dimensional/src/main/java/org/geoserver/gwc/wmts/MultiDimensionalExtension.java
aaime: I don't believe a nameless domain is allowed, but the user can ask for it... instead of an empty test, the code should probably have an existence check (which is indeed missing now) so that the client would get back an error when requesting a non existing dimension.

src/community/wmts-multi-dimensional/src/main/java/org/geoserver/gwc/wmts/dimensions/CoverageDimensionsReader.java
aaime: Good catch :)

aaime:
Pushed one more commit addressing your review (and fixing a test failing)

PR #2742, opened 2018-01-29
[GEOS-8556] Add option to copy global attributes from NetCDF/GRIB source files

maka-io:
https://osgeo-org.atlassian.net/browse/GEOS-8556

I renamed the copyAttribute (and the like) to copyVariableAttribute for consistency.  Updated tests.

PR #2780, opened 2018-03-01
[GEOS-8556] Add option to copy global attributes from NetCDF/GRIB source files (2.12.x)

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOS-8556

Backport to 2.12.x of: https://github.com/geoserver/geoserver/pull/2742


aaime:
A month passed, so the backport is fine, but ... isn't it a bit strange to have it in 2.12.x and not in 2.13.x?
However 2.13.x is frozen so... will you remember to port there once 2.13.0 is out?

aaime:
Actually, scratch that, it's probably already in 2.13.x :)

bencaradocdavies:
@aaime yum, already in 2.13.x as it was merged on master before the new branch was made. In fact, before the decision to branch on beta.  :-)

An interesting question is whether 2.12.x counts as stable or maintenance in the window between 2.13-beta and 2.13.0. What rules apply?

aaime:
Unsure, we're in a gray area :)
Personally I'm not too concerned, this patch worked fine when I tried it out and we are talking about a smaller change on an extension module, not a large change in core. But it's just me :-)

PR #2781, opened 2018-03-02
[GEOS-8065] Vector tiles miss support for WMS filters

aaime:
Added also an integration test, cause... http://silveiraneto.net/2017/03/27/2-unit-tests-0-integration-tests/


DBlasby:
looks good!

PR #2782, opened 2018-03-02
Update gs-gwc javadocs

tbarsballe:
Based primarily on changes introduced by https://github.com/geoserver/geoserver/pull/2760

Also fixes a missing license header.

### DIFF
--- a/src/gwc/src/main/java/org/geoserver/gwc/GWC.java
+++ b/src/gwc/src/main/java/org/geoserver/gwc/GWC.java
@@ -74,6 +74,7 @@ import org.geoserver.wfs.kvp.BBoxKvpParser;
 import org.geoserver.wms.GetMapRequest;
 import org.geoserver.wms.WMS;
 import org.geoserver.wms.map.RenderedImageMap;
+import org.geoserver.wms.map.RenderedImageMapResponse;
 import org.geotools.factory.CommonFactoryFinder;
 import org.geotools.filter.visitor.ExtractBoundsFilterVisitor;
 import org.geotools.geometry.GeneralEnvelope;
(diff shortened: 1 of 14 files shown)


src/gwc/src/main/java/org/geoserver/gwc/layer/GeoServerTileLayerInfoImpl.java
```diff
@@ -266,28 +268,50 @@ public void setMetaTilingX(int metaTilingX) {
...
+     */
     @Override
     public int getExpireClients() {
     	return expireClients;
     }
-    
+
+    /**
+     * @see GeoServerTileLayerInfo#setExpireClients(int)
+     */
```
smithkm: I usually consider the `@Override` enough on its own.  Until recently "@see X" javadocs like this would get in the way in Eclipse meaning you had to click through to get to the real docs instead of just inheriting the,  That's fixed now but this is still a bit redundant.

src/gwc/src/main/java/org/geoserver/gwc/GWC.java
```diff
@@ -1563,6 +1593,12 @@ private BlobStoreAggregator getBlobStoreAggregator() {
         return blobStoreAggregator;
     }
 
+    /**
+     * Image encoder for metatile requests
```
smithkm: This phrasing is a bit misleading, what you use further down gets the point across better.  There's a moment of confusion that this says "Encoder" where the return type is `Response`.  Having that "Response that can encode" phrasing up here at the top would head that confusion off.

src/gwc/src/main/java/org/geoserver/gwc/layer/GeoServerTileLayerInfoImpl.java
tbarsballe: Fair enough, that class already had some methods with `@see` and `@Override`, so I was mostly trying to make things consistent. In general I would agree that just `@Override` is sufficient.

src/gwc/src/main/java/org/geoserver/gwc/GWC.java
tbarsballe: I'll update to use that wording.

tbarsballe:
Backported to 2.13.x

PR #2784, opened 2018-03-05
[GEOS-8617] NetCDF 4 output may fail on subsetting when copy source variables is enabled

aaime:


bencaradocdavies:
Nice. I like.

PR #2787, opened 2018-03-06
ModificationProxy.rewrap

smithkm:
Utility method to alter wrappers underneath a modification proxy without losing any staged modifications.

Created to assist with #2771 but it turned out to not be needed there.  Still potentially useful though.


tbarsballe:
This does seem useful, although should probably have an associated JIRA issue for tracking purposes.
Also needs a test case.

aaime:
Has test cases. Inclined to merged even without ticket reference, as long as it does not require a backport.

aaime:
Merged :)

PR #2793, opened 2018-03-12
Add the ability to parameterize outgoing passwords in backups

dvntucker:
This PR adds the ability to have outgoing store passwords in backups replaced with tokens in the backup xml, as well as the ability to replace these tokens on a restore. Functionality is all optional so far, and only REST configuration at this point.

Primary use case is migrating existing environments for which it would be too difficult to parameterize using the existing parameterization functions (because of the size of the catalog).

I had to change the protection level on a few XML converters in order to do this. Originally I wanted to refactor these out into their own class, but because they are implemented as non-static inner classes it proved to be a bit of a hassle.


src/community/backup-restore/core/src/test/java/org/geoserver/backuprestore/BackupTest.java
```diff
@@ -308,8 +379,8 @@ private void checkExtraPropertiesExists() {
         } catch (Exception exception) {
             throw new RuntimeException("Error reading extra properties file.", exception);
         }
-        assertThat(extraProperties.size(), is(2));
-        assertThat(extraProperties.getProperty("property.a"), is("1"));
-        assertThat(extraProperties.getProperty("property.b"), is("2"));
+         assertThat(extraProperties.size(), is(2));
+         assertThat(extraProperties.getProperty("property.a"), is("1"));
+         assertThat(extraProperties.getProperty("property.b"), is("2"));
```
tbarsballe: 9 space indentation here, should be 8 spaces

src/gwc/src/main/java/org/geoserver/gwc/layer/DefaultTileLayerCatalog.java
```diff
@@ -324,7 +324,7 @@ public String getLayerName(String layerId) {
         checkInitialized();
         return layersById.get(layerId).getName();
     }
-    
+
```
tbarsballe: No functional changes in this file...

src/community/backup-restore/core/src/main/java/org/geoserver/backuprestore/BackupRestoreItem.java
```diff
@@ -164,6 +181,28 @@ public void retrieveInterstepData(StepExecution stepExecution) {
...
         JobParameters jobParameters = this.currentJobExecution.getJobParameters();
 
+        boolean parameterizePasswords = Boolean.parseBoolean(
+            jobParameters.getString(Backup.PARAM_PARAMETERIZE_PASSWDS, "false"));
+
+
+        if (parameterizePasswords) {
+
+            //here we set some customized XML handling code. For backups, we add a converter that tokenizes
+            //outgoing passwords. for restores, a handler for those tokenized backups.
```
tbarsballe: Given that one of the converters only marshals, and the other only unmarshals, you could make an argument for combining them, but they are probably fine as-is.

dvntucker: Unfortunately the marshalling/unmarshalling are mutually exclusive, otherwise I'd combine them.

PR #2796, opened 2018-03-13
Control-flow priority support

aaime:


doc/en/user/source/extensions/controlflow/index.rst
```diff
@@ -48,6 +48,24 @@ A few examples::
...
+  
+Request priority support
+........................
+
+Requests controlled by "ows.*" controllers above can be also executed in priority order, in case there are too many
+the request will block and wait, and will we awoken in priority order (highest to lowest).
+
+Currently the only way to specific a priority for a request is to add it to a request HTTP header::
+
+  ows.priority.header=<headerName>,<defaultPriority>
```
nmco: The code is expecting **http** instead of **header**:
```
ows.priority.http=<headerName>,<defaultPriority>
```
... not sure which one is the correct one :stuck_out_tongue:

src/extension/control-flow/src/main/java/org/geoserver/flow/config/DefaultControlFlowConfigurator.java
```diff
@@ -107,19 +112,20 @@ public DefaultControlFlowConfigurator() {
...
 
         Properties p = configFile.getProperties();
-        List<FlowController> newControllers = new ArrayList<FlowController>();
+        List<FlowController> newControllers = new ArrayList<>();
+        PriorityProvider priorityProvider = getPriorityProvider(p);
+        
         for (Object okey : p.keySet()) {
             String key = ((String) okey).trim();
             String value = (String) p.get(okey);
-            LOGGER.info("Loading control-flow configuration: " + key + "=" + value);
```
nmco: So this has moved down to give more evaluable feedback right ?

src/extension/control-flow/src/main/java/org/geoserver/flow/config/DefaultControlFlowConfigurator.java
```diff
@@ -178,22 +185,75 @@ protected KeyGenerator buildKeyGenerator(String[] keys, String value) {
...
+                        String ip = key.substring("ip.".length());
+                        controller = new SingleIpFlowController(queueSize, ip);
+                    }
                 }
             }
+            
             if (controller == null) {
                 LOGGER.severe("Could not parse rule '" + okey + "=" + value);
             } else {
+                LOGGER.info("Loaded flow controller configuration: " + key + "=" + value);
```
nmco: This two logs messages are bit out of sync, I would have put something like:
```
LOGGER.severe("Could not parse control-flow rule '" + okey + "=" + value);
LOGGER.info("Loaded control-flow rule: " + key + "=" + value);
```

src/extension/control-flow/src/test/java/org/geoserver/flow/controller/PriorityFlowControllerTest.java
```diff
@@ -0,0 +1,119 @@
...
+    @Test
+    public void testTimeout() {
+        // create a single item flow controller 
+        HttpHeaderPriorityProvider priorityProvider = new HttpHeaderPriorityProvider
+                (PRIORITY_HEADER_NAME, 0);
+        GlobalFlowController controller = new GlobalFlowController(1, new PriorityThreadBlocker
+                (1, priorityProvider));
+
+        // make two testing threads that will "process" for 400ms, but with a timeout of 200 on the
+        // flow controller
```
nmco: The timeout used is actually 100 not 200.

bradh:
Looks like relevant failure in the test:

```
Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 3.781 sec <<< FAILURE!
testTimeout(org.geoserver.flow.controller.PriorityFlowControllerTest)  Time elapsed: 3574 sec  <<< FAILURE!
java.lang.AssertionError: expected:<TIMED_OUT> but was:<COMPLETE>
```

doc/en/user/source/extensions/controlflow/index.rst
aaime: The code, thanks for spotting this!

src/extension/control-flow/src/main/java/org/geoserver/flow/config/DefaultControlFlowConfigurator.java
aaime: Yes, to avoid it logging loading of a directive that is not a flow controller

src/extension/control-flow/src/main/java/org/geoserver/flow/config/DefaultControlFlowConfigurator.java
aaime: Yup

src/extension/control-flow/src/test/java/org/geoserver/flow/controller/PriorityFlowControllerTest.java
aaime: Correct

aaime:
Yep saw that.. put the method in question in a infinite loop, has run thousands of times without failing... what a pain :)

aaime:
Ok, added a commit and modified the tests so that they should not fail anymore on a noisy/otherwise loaded machine. Tested locally using "stress" with 128 threads computing square roots at full tilt, did not fail (was very occasionally failing locally before fixes)... let's see here.

PR #2803, opened 2018-03-14
[GEOS-8631] JDBC Configuration: Fixed JDBC Configuration not upgrading to write locks before modifying the catalog.

sikeoka:
Pull request for https://osgeo-org.atlassian.net/browse/GEOS-8631 that fixes READ locks not being upgraded to WRITE locks before modifying the catalog when using JDBC Configuration.  Ideally, IsolatedCatalogFacade would have been moved to setFacade() at the same time but this causes ModificationProxy issues in JDBC Configuration so it will have to be done separately.

A commented out unit test that ignored an assertion error was removed since that unit test has been passing for a while now.

Contains unit test and can be backported to 2.13.x.

### DIFF
--- a/src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/catalog/JDBCCatalogFacade.java
+++ b/src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/catalog/JDBCCatalogFacade.java
@@ -26,6 +26,7 @@ import org.geoserver.catalog.CatalogInfo;
 import org.geoserver.catalog.DataStoreInfo;
 import org.geoserver.catalog.LayerGroupInfo;
 import org.geoserver.catalog.LayerInfo;
+import org.geoserver.catalog.LockingCatalogFacade;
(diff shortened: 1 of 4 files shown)


jodygarnett:
Steve this one is unmaintained, please ask on the developer list for assistance. @NielsCharlier has been helpful in the past and may be in position to help.

You two may wish to volunteer for the module and check each others work?

NielsCharlier:
This looks very good to me, I approve of it. There is a core change too, maybe they need separate approval for that?

sikeoka:
Hello Niels, 
could you also review these two JDBC Config pull requests?  I don't have permissions to assign them to you.  2801 is not the most ideal fix but it is kind of a big issue with a clustered GeoServer and 2797 is just a minor annoyance with an easy workaround.
https://github.com/geoserver/geoserver/pull/2801
https://github.com/geoserver/geoserver/pull/2797

Thank you.

NielsCharlier:
I saw @smithkm already reviewed and merged them, thanks for your contributions @sikeoka

PR #2805, opened 2018-03-16
[GEOS-8603] Include missing types in `canSecure` list

emerkle826:
The original GEOS-8603 fix added WMSLayerInfo and WMTSLayerInfo types to the `secure` method of the default factory, but forgot to add them to the list of supported types. Tests added as well.

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/decorators/SecuredObjects.java
+++ b/src/main/src/main/java/org/geoserver/security/decorators/SecuredObjects.java
@@ -5,6 +5,8 @@
  */
 package org.geoserver.security.decorators;
 
+import org.geoserver.catalog.impl.ModificationProxy;
+
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
@@ -15,9 +17,9 @@ import org.geoserver.security.WrapperPolicy;
 /**
  * Utility class that provides easy and fast access to the registered
  * {@link SecuredObjectFactory} implementations
- * 
+ *
  * @author Andrea Aime - TOPP
- * 
+ *
  */
 public class SecuredObjects {
     /**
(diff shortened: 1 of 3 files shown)


src/main/src/main/java/org/geoserver/security/decorators/SecuredObjects.java
```diff
@@ -45,9 +46,13 @@ public static Object secure(Object object, WrapperPolicy policy) {
...
             return null;
 
+        // In case the object to be secured is wrapped in a ModificationProxy, get the proxied Object class
+        // if the object is not wrapped, ModificationProxy.unwrap() will just return the object reference.
+        final Object unwrappedObject = ModificationProxy.unwrap(object);
+
         // if we already know what can handle the wrapping, just do it, don't
         // scan the extension points once more
-        Class<?> clazz = object.getClass();
+        Class<?> clazz = unwrappedObject.getClass();
```
aaime: Seems fine but can you elaborate on this change?

emerkle826: Sure thing.

In debugging the issue, I noticed that the code here was handling dynamic Proxy object instances (`java.lang.reflect.Proxy.$Proxy101` for example) in some cases, and normal Catalog objects in other cases (`org.geoserver.catalog.impl.DataStoreInfoImpl` for example), where the Proxy objects were ModificationProxy objects wrapping _normal_ Catalog objects. While the Proxy objects will still pass the `canSecure()` method as long as they wrap an object instance of a supported type, it seemed possible for the Factory cache to grow larger than necessary over time, if the Proxy instances changed. I felt it was safer to use the unwrapped class type as the key for this cache, to avoid the possibility of multiple Proxy instances wrapping the same underlying type being added to the cache.

I'n debugging through a running instance, I didn't observe multiple Proxy instances wrapping the same type come through this code, however. So I don't know how possible or impossible it actually is over time. But it seems safer to have the actual Catalog class types in the cache.

PR #2808, opened 2018-03-16
[GEOS-8646] Have DescribeDomain expand the list of actual dimension values once the list is small enough

aaime:


src/community/wmts-multi-dimensional/src/main/java/org/geoserver/gwc/wmts/MultiDimensionalExtension.java
```diff
@@ -198,8 +217,48 @@ private Domains getDomains(SimpleConveyor conveyor) throws Exception {
...
+        if (clientExpandLimit != null) {
+            Integer value = Converters.convert(clientExpandLimit, Integer.class);
+            if (value == null) {
+                throw new OWSException(400, "InvalidParameterValue", EXPAND_LIMIT, "Invalid " + EXPAND_LIMIT + " " +
+                        "value " + clientExpandLimit + ", expected an integer value");
+            }
+            if (value < expandLimitMax) {
+                return value;
+            } else {
+                LOGGER.log(Level.INFO, "Client provided a expand value higher than the configured max, using the " +
```
nmco: The spec will need to be explicit about this behavior, since there is now way for the client to know about the max value defined in the server this may cause some misunderstanding. Bu tI'm not sure about the practical consequences of that.

src/community/wmts-multi-dimensional/src/main/java/org/geoserver/gwc/wmts/dimensions/Dimension.java
```diff
@@ -56,18 +57,28 @@ public Dimension(WMS wms, String dimensionName, LayerInfo layerInfo, DimensionIn
...
+     * @param attribute
+     * @param expandLimit
+     * @return
+     */
+    protected DomainSummary getDomainSummary(FeatureCollection features, String attribute, int expandLimit) {
+        // grab domain, but at most expandLimit + 1, to know if there are too many
+        if (expandLimit != 0) {
+            TreeSet uniqueValues = DimensionsUtils.getUniqueValues(features, attribute, expandLimit + 1);
+            if (uniqueValues.size() <= expandLimit || expandLimit < 0) {
+                return new DomainSummary(uniqueValues);
```
nmco: In the case were the retrieved unique values are superior to the expand limit, will not be faster to compute the MIN and MAX than delegate everything again to the db ? Or let's be crazy compute those values on the fly when building the tree ?

src/community/wmts-multi-dimensional/src/test/java/org/geoserver/gwc/wmts/MultiDimensionalExtensionTest.java
```diff
@@ -272,14 +267,44 @@ public void testVectorDescribeDomainsOperationWithBoundingBoxFilter() throws Exc
...
         // check the time domain
         checkXpathCount(result, "/md:Domains/md:DimensionDomain[ows:Identifier='time']", "1");
-        checkXpathCount(result, "/md:Domains/md:DimensionDomain[md:Domain='2012-02-11T00:00:00.000Z,2012-02-12T00:00:00.000Z']", "1");
+        checkXpathCount(result, "/md:Domains/md:DimensionDomain[ows:Identifier = 'time' and md:Size='2']", "1");
+        checkXpathCount(result, "/md:Domains/md:DimensionDomain[ows:Identifier='time' and md:Domain='2012-02-11T00:00:00.000Z,2012-02-12T00:00:00.000Z']", "1");
+    }
+
+    /**
+     * Same as {@link #testVectorDescribeDomainsOperationWithBoundingBoxFilter()} but with a limit of zero, so all domain
+     * descriptions should contract to a
```
nmco: ... min max value ? :stuck_out_tongue:

src/community/wmts-multi-dimensional/src/main/java/org/geoserver/gwc/wmts/dimensions/Dimension.java
aaime: Computing all the unique values might take too much time and would basically ignore the settings configured by the admin on the limit and max limit. The getUniqueValues tries to get (expandLimit + 1) unique values, that is translated down into query, if the limits are properly configured, that will result in a fast query. If (expandLimit +1) are returned then there are too many, but we don't have them all, so delegating again to the db is the only option.

src/community/wmts-multi-dimensional/src/test/java/org/geoserver/gwc/wmts/MultiDimensionalExtensionTest.java
aaime: Yes sir!

src/community/wmts-multi-dimensional/src/main/java/org/geoserver/gwc/wmts/MultiDimensionalExtension.java
aaime: image

src/community/wmts-multi-dimensional/src/main/java/org/geoserver/gwc/wmts/dimensions/Dimension.java
nmco: Touché.

PR #2809, opened 2018-03-20
[GEOS-8652] WMTS multidimensional histogram calculation failures against coverage views, and bucketing has overlaps

aaime:


nmco:
Looks good to me. The [specification](https://demo.geo-solutions.it/share/wmts-multidim/wmts_multidim_geosolutions.html#truehistogram-generation) would improve in clarity in having in having a few examples that show the corner cases.  The specification \ documentation should also be explicit that the build histogram is not inclusive for the last value, only for the start one.

aaime:
The specification linked is stale and not version controlled, we'll have to eventually move it into the GeoServer documentation, but for the moment, I've updated the module docs.

PR #2810, opened 2018-03-22
[GEOS-8656] Correct wrong encoding for French translation

yjacolin:


nmco:
Looks good to me, thank you @yjacolin ! I don't think changes on a single file require a cotnribution agreement (http://docs.geoserver.org/latest/en/developer/policies/committing.html#getting-commit-access) if nobody complains I will merge.

nmco:
Wait, actually this commit is targeting 2.13.x a pull should be made against 2.14.x (master) and then backported.

aaime:
Indeed afaik single file changes do not require a CLA

aaime:
In other words, keep this one, but also make one for master (2.14.x) :-)

yjacolin:
I am on it!

yjacolin:
done for 2.14 (master): https://github.com/geoserver/geoserver/pull/2811

bencaradocdavies:
It is worth noting the Java 9 changes the default encoding of properties files to UTF-8, with fallback to ISO-8859-1 if invalid byte sequences are detected:
https://docs.oracle.com/javase/9/intl/internationalization-enhancements-jdk-9.htm

nmco:
Good point, thanks @bencaradocdavies.

bencaradocdavies:
The encoding change might in theory cause misinterpretation of some bytes, but I much prefer UTF-8. In other projects I have used XML property files in UTF-8 so I can cut and paste from web pages including Google Translate. Not to mention [我的氣墊船裝滿了鱔魚](http://www.omniglot.com/language/phrases/hovercraft.htm) ([context](https://en.wikipedia.org/wiki/Dirty_Hungarian_Phrasebook)).

nmco:
I'm a native speaker in both French (Switzerland) and Portuguese, so no copy paste ... actually if the copy paste happen it was the other way around from PT or FR to English. I actually kind of remember that IntelliJ told me something about the file encoding ... anyway I should have pay more attention.

And yes this one _English As She Is Spoke_ is quite interesting :yum:

PR #2811, opened 2018-03-22
[GEOS-8656] Correct wrong encoding for French translation

yjacolin:


nmco:
Thank you @yjacolin !

PR #2813, opened 2018-03-22
Fix incorrect log message

dbaston:
Hopefully such a small PR isn't annoying, but I noticed this in my logs and thought it was worth fixing.


aaime:
Yep good one. I can apply it directly to master, with no backport, and close an eye, or if you could create a ticket for it and make the commit message be "[GEOS-XYWZ] Title of the message" do also a backport?

dbaston:
@aaime If going through the trouble to ticket and backport, should I go a bit futher standardize the format of all of these log messages?

aaime:
As you prefer :-)

aaime:
Opened [ticket](https://osgeo-org.atlassian.net/browse/GEOS-8685), rewrote commit message and manually merged on 2.13.x, 2.12.x

PR #2817, opened 2018-03-28
[GEOS-8667] WMTS Multidimensional GetHistogram is memory bound and can cause OOM over large data sets

aaime:
Requires the changes in https://github.com/geotools/geotools/pull/1843 to be build and get into the M2 repos for tests to pass

### DIFF
--- a/src/community/wmts-multi-dimensional/src/main/java/org/geoserver/gwc/wmts/dimensions/Dimension.java
+++ b/src/community/wmts-multi-dimensional/src/main/java/org/geoserver/gwc/wmts/dimensions/Dimension.java
@@ -10,18 +10,31 @@ import org.geoserver.catalog.FeatureTypeInfo;
 import org.geoserver.catalog.LayerInfo;
 import org.geoserver.catalog.ResourceInfo;
 import org.geoserver.gwc.wmts.Tuple;
+import org.geoserver.platform.GeoServerExtensions;
 import org.geoserver.wms.WMS;
 import org.geoserver.wms.dimension.DimensionDefaultValueSelectionStrategy;
+import org.geotools.data.Query;
 import org.geotools.feature.FeatureCollection;
 import org.geotools.feature.visitor.Aggregate;
+import org.geotools.feature.visitor.GroupByVisitor;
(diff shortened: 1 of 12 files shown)


src/community/wmts-multi-dimensional/src/main/java/org/geoserver/gwc/wmts/dimensions/Dimension.java
```diff
@@ -81,27 +94,164 @@ protected DomainSummary getDomainSummary(FeatureCollection features, String attr
...
+     * Small easter egg to allow testing, we might want to extend it to a per layer configuration
+     * in case there are large shapefile layers involved in this (computing min/max/groupby is
+     * three full data scans in that case). Or have a way to figure out if a collection can optimize
+     * out a visit (completely missing right now, that would be a significant API change).
+     * @return
+     */
+    private boolean loadDataInMemory() {
+        String value = Optional.ofNullable(GeoServerExtensions.getProperty("WMTS_HISTOGRAM_IN_MEMORY"))
+                .orElse("false");
+        return Boolean.getBoolean(value);
```
nmco: If ``value`` is ``NULL``, ``Boolean.getBoolean(value)`` will return ``FALSE`` anyway, but I understand that this is more explicit.

aaime:
In memory yes, but streaming, instead of making GeoServer go OOM

src/community/wmts-multi-dimensional/src/main/java/org/geoserver/gwc/wmts/dimensions/Dimension.java
aaime: Good point, I can remove that optional

aaime:
Ah, about the test, it can be done using H2, no need to make it an online one, but it would not add much, unless the code is modified in a way that allows to check if the visitor has been optimized out or not. As long as the memory visit works, it's up to GT2 to guarantee a in memory visit and a optimized sql one provide the same result.

nmco:
Ok that makes sense, thanks for the feedback.

PR #2821, opened 2018-04-01
[Fixes JIRA #GEOS-8650] Externalize Importer working directory

afabiani:


src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -109,6 +114,14 @@
 public class Importer implements DisposableBean, ApplicationListener {
 
     static Logger LOGGER = Logging.getLogger(Importer.class);
+    
+    public static final String PROPERTYFILENAME = "geoserver-importer.properties";
```
aaime: Other plugins using simple config files are not prefixing with "geoserver-", like "controlflow.properties".
Probably better to just call it "importer.properties"

src/extension/importer/core/src/test/java/org/geoserver/importer/ImporterDataTest.java
```diff
@@ -82,6 +83,32 @@ public SimpleFeature apply(ImportTask task, DataStore dataStore, SimpleFeature o
...
+        // Read the upload root folder and creates it if does not exists
+        importer.getUploadRoot();
+        
+        // Read the folder again...
+        dirFromProperties = Resources.directory(Resources.fromPath("props_uploads"));
+        // ... and ensure it is the same as defined on the .properties file
+        assertEquals(dirFromProperties, importer.getUploadRoot());
+        
+        // Let's now override the external folder through the Environment variable. This takes precedence on .properties
+        System.setProperty(Importer.UPLOAD_ROOT_KEY, "env_uploads");
```
aaime: A test setting an env variable should be clearing it at the end (finally block here, or @AfterTest method)

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -1645,7 +1678,36 @@ public File getImportRoot() {
...
+        }
+        
+        if (props != null && value == null) {
+            value = props.getProperty(UPLOAD_ROOT_KEY);
+        }
+        
+        try {
+            if(value != null) {
+                Resource uploadsRoot = Resources.fromPath(value);
+                return Resources.directory(uploadsRoot, !Resources.exists(uploadsRoot));
```
aaime: What if someone wants to use an absolute path to keep data outside of the data directory? (quite common)
Suggestion: use a fixed location relative to the data directory by default, and for configured locations, check first if they are absolute paths.

src/extension/importer/core/src/test/java/org/geoserver/importer/ImporterTestSupport.java
```diff
@@ -74,6 +74,11 @@ protected void onSetUp(SystemTestData testData) throws Exception {
         namespaces.put("wms", "http://www.opengis.net/wms");
         
         XMLUnit.setXpathNamespaceContext(new SimpleNamespaceContext(namespaces));
+        
+        // add limits properties file
```
aaime: Limits?

aaime: The test should also check that the imported data actually ends up in that directory (either here or in a REST test)

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
afabiani: Ok. Also might be good to have the importer config directory as you suggested on the other PR.

src/extension/importer/core/src/test/java/org/geoserver/importer/ImporterDataTest.java
afabiani: Right, I will fix it.

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
afabiani: This check is done already by "Resources.fromPath" method.

On my tests it works either with relative and absolute paths.

src/extension/importer/core/src/test/java/org/geoserver/importer/ImporterTestSupport.java
afabiani: The comment is wrong, will change it.

I'll add REST test also.

PR #2822, opened 2018-04-01
[Fixes JIRA #GEOS-8649] Importer does not clean temporary folders aft…

afabiani:
…er imports

### DIFF
--- a/src/extension/importer/core/src/main/java/org/geoserver/importer/Directory.java
+++ b/src/extension/importer/core/src/main/java/org/geoserver/importer/Directory.java
@@ -39,7 +39,7 @@ import com.google.common.collect.Iterables;
 public class Directory extends FileData {
 
     private static final Logger LOGGER = Logging.getLogger(Directory.class);
-    
+
     private static final long serialVersionUID = 1L;
 
     /**
@@ -51,6 +51,7 @@ public class Directory extends FileData {
      * flag controlling whether file look up should recurse into sub directories.
      */
     boolean recursive;
+
     String name;
 
     public Directory(File file) {
@@ -64,7 +65,9 @@ public class Directory extends FileData {
 
     public static Directory createNew(File parent) throws IOException {
         File directory = File.createTempFile("tmp", "", parent);
-        if (!directory.delete() || !directory.mkdir()) throw new IOException("Error creating temp directory at " + directory.getAbsolutePath());
+        if (!directory.delete() || !directory.mkdir())
+            throw new IOException(
+                    "Error creating temp directory at " + directory.getAbsolutePath());
         return new Directory(directory);
     }
 
@@ -93,10 +96,11 @@ public class Directory extends FileData {
     }
 
     public void unpack(File file) throws IOException {
-        //if the file is an archive, unpack it
+        // if the file is an archive, unpack it
         VFSWorker vfs = new VFSWorker();
         if (vfs.canHandle(file)) {
-            LOGGER.fine("unpacking " + file.getAbsolutePath() + " to " + this.file.getAbsolutePath());
+            LOGGER.fine(
+                    "unpacking " + file.getAbsolutePath() + " to " + this.file.getAbsolutePath());
             vfs.extractTo(file, this.file);
 
             LOGGER.fine("deleting " + file.getAbsolutePath());
@@ -105,10 +109,10 @@ public class Directory extends FileData {
             }
         }
     }
-    
+
     public File child(String name) {
         if (name == null) {
-            //create random
+            // create random
             try {
                 return File.createTempFile("child", "tmp", file);
             } catch (IOException e) {
@@ -116,7 +120,7 @@ public class Directory extends FileData {
             }
         }
 
-        return new File(this.file,name);
+        return new File(this.file, name);
     }
 
     public void setName(String name) {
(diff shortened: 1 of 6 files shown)


src/extension/importer/core/src/main/java/org/geoserver/importer/Directory.java
```diff
@@ -39,7 +39,7 @@
 public class Directory extends FileData {
 
     private static final Logger LOGGER = Logging.getLogger(Directory.class);
-    
+
```
aaime: In this file I only see reformats, but might be wrong. Formats changes are not encouraged, and if you have to do them for any reason, place them in a different commit.

src/extension/importer/core/src/main/java/org/geoserver/importer/ImportContext.java
```diff
@@ -278,12 +279,48 @@ public void updated() {
...
+                    }
+                }
+            }
+        }
+        
+        return false;
+    }
+
+    protected boolean canBeRemoved(ImportContext context) {
+        if (context.getState() == ImportContext.State.COMPLETE ||
```
aaime: The formatting is wrong in various places in this method. Also, ifs always need { and } around their blocks (even if one lineers)

src/extension/importer/core/src/main/java/org/geoserver/importer/ImportContext.java
```diff
@@ -278,12 +279,48 @@ public void updated() {
...
+        return false;
+    }
+
+    protected boolean canBeRemoved(ImportContext context) {
+        if (context.getState() == ImportContext.State.COMPLETE || 
+                context.getState() == ImportContext.State.INIT_ERROR)
+            return true;
+        else if (context.getState() == ImportContext.State.PENDING) {
+            for(ImportTask task : context.getTasks()) {
+                if (task.getState() == ImportTask.State.ERROR || task.getError() != null)
```
aaime: Even an errored task can be updated and re-run, imho the files should not be removed. 
However, this might mean some files are not going to to cleaned up in case of shutdown without cleaning the completed and errored imports. Hum... there could be a shutdown hook that cleans the errored ones in case the memory storage is used, maybe? A non clean shutdown will still leave stuff around of course. Thinking out loud, maybe also add a marker file that indicates incomplete imports? (could be used for pending ones too, but not sure)

src/extension/importer/core/src/main/java/org/geoserver/importer/ImportContext.java
```diff
@@ -278,12 +279,48 @@ public void updated() {
...
+                    return true;
+            }
         }
+        return false;
     }
 
+    protected boolean canBePruned(ImportTask task) {
+        if (!task.isDirect() ||
+                task.getState() == ImportTask.State.ERROR || task.getError() != null)
```
aaime: Same as above

src/extension/importer/core/src/main/java/org/geoserver/importer/job/JobQueue.java
```diff
@@ -60,8 +71,19 @@ protected void afterExecute(Runnable r, Throwable t) {
             public void run() {
                 List<Long> toremove = new ArrayList<Long>();
                 for (Map.Entry<Long, Task<?>> e : jobs.entrySet()) {
-                    if (e.getValue().isCancelled() || (e.getValue().isDone() && e.getValue().isRecieved())) {
-                        toremove.add(e.getKey());
+                    if (e.getValue().isCancelled() || (e.getValue().isDone())) {
```
aaime: Why the change in logic condition here?

src/extension/importer/core/src/main/java/org/geoserver/importer/job/JobQueue.java
```diff
@@ -60,8 +71,19 @@ protected void afterExecute(Runnable r, Throwable t) {
...
+                        try {
+                            ImportContext context = (ImportContext) e.getValue().get();
+                            if(context.delete()) {
```
aaime: Not sure this is "valid", what if the task is done but the context is still running (there can be imports with many tasks, hundreds or thousands)?

src/extension/importer/core/src/test/java/org/geoserver/importer/ImporterDataTest.java
```diff
@@ -431,6 +435,10 @@ public void testImportIntoDatabase() throws Exception {
 
         runChecks("archsites");
         runChecks("bugsites");
+        
+        // Test that temporary data is removed also
+        context.delete();
```
aaime: If you're calling delete manually, you're testing the delete cleans up, but not if the delete will be called by the importer machinery automatically. Probably best to add a REST test and do the checks there too, without a manual delete call?
The JobQueue cleanup frequency should be made configurable, and I guess more frequent than "one minute" per loop (or the test will be too long).

aaime:
@afabiani one more thing, I'm working on a pull request that will make it possible to run executable scripts placed in "importer/scripts". Does it make sense to change this one in order to obtain the following complete directory layout:

````
  data_dir
     - importer
          - importer.properties
          - uploads
          - scripts
````

That is, everything in the same folder? That would match more or less the monitoring module approach.

afabiani:
@aaime I'm going to close this one and re-issue a new PR related to this Issue along with:

1. Improved logic (as per your review's suggestions)
2. Alignment with directory structure

PR #2823, opened 2018-04-03
GEOT-5963 support S3 on other servers (docs)

NielsCharlier:
see https://github.com/geotools/geotools/pull/1850


dvntucker:
Hi Niels, it looks like there is only one file changed. Are there commits missing from the PR?

NielsCharlier:
This is only the documentation, there is a link above to the geotools PR.

dvntucker:
Oops, sorry I missed that.

PR #2824, opened 2018-04-03
GEOS-8663 Easy insert images into style

NielsCharlier:


src/web/wms/src/main/java/org/geoserver/wms/web/data/AbstractStylePage.java
```diff
@@ -59,6 +87,130 @@
  */
 @SuppressWarnings("serial")
 public abstract class AbstractStylePage extends GeoServerSecuredPage {
+    
+    private static String[] IMAGE_EXTENSIONS = new String[] {"png", "jpg", "jpeg", "gif", "svg"};
```
tbarsballe: I wonder if this should be configurable, in case there are extensions that add supported image types (I don't know of any currently, but MBStyle may eventually support pbf images).
Probably fine as-is, and if anyone adds a new image format they can be responsible for making this support it

src/web/wms/src/main/java/org/geoserver/wms/web/data/AbstractStylePage.java
```diff
@@ -291,6 +443,75 @@ public void onSubmit(AjaxRequestTarget target, Form<?> form) {
...
+                                            IOUtils.copy(is, os);                                            
+                                        }                                        
+                                    } catch (IOException e) {
+                                        error(e.getMessage());
+                                        target.add(imagePanel.getFeedback());
+                                        return false;
+                                    }
+                                }
+                                
+                                StringBuffer sb = new StringBuffer("<ExternalGraphic>\\n")
```
tbarsballe: This should be pluggable to support different formats (CSS, YSLD, MBStyle, etc.).

Converting SLD snippents isn't really possible, so it will have to have a custom implementation per format. Something similar to StyleGenerator would probably work, with a bit of extra tweaking.

aaime: What about just having the code insert the path to the image, instead of the whole XML block around it? That should work well across formats.

NielsCharlier: But that would also reduce the convenience of the functionality. Perhaps a pluggable option with default (not SLD): only path is the best compromise for now?

src/web/wms/src/main/java/org/geoserver/wms/web/data/AbstractStylePage.java
NielsCharlier: Agreed, unless you have a different suggestion of where to configure it in Spring perhaps

NielsCharlier:
@tbarsballe do you have a suggestion of how to make it clearer to users? The toolbar seems like the most obvious place to put it.

src/web/wms/src/main/java/org/geoserver/wms/web/data/AbstractStylePage.java
tbarsballe: @NielsCharlier That seems like a good option, then module maintainers can implement there own for other formats

tbarsballe:
With respect to the insert, I agree the toolbar seems a logical place to put it. Perhaps to make it a bit more obvious, also include the text "Insert image..." within the button alongside the icon (not just as title text). A bit inelegant, but will make it more obvious.

src/web/wms/src/main/java/org/geoserver/wms/web/data/AbstractStylePage.java
aaime: Agreed

aaime:
In terms of it being obvious or not, a documentation update is also needed, there is a table explaining the various icons in the toolbar here:
http://docs.geoserver.org/stable/en/user/styling/webadmin/index.html#style-definition

NielsCharlier:
I fixed the NPE, moved the code generation to the style handler so it is format specific and be extended, and I added docs. If it's okay I chose not to put the text in the button because it was a bit ugly and I pesonally think the tooltip + docs should be enough.

aaime:
Also agree the text in the button would be been... strange? The other buttons are pure icon. Personally believe the doc update plus tooltip should be enough too.

NielsCharlier:
Thanks everyone

src/web/wms/src/test/java/org/geoserver/wms/web/data/StyleEditPageTest.java
```diff
@@ -197,6 +202,72 @@ public void testLoad() throws Exception {
...
+        Matcher matcher = pattern.matcher(tester.getLastResponse().getDocument());
+        assertTrue(matcher.find());
+        assertEquals("somepicture.png", matcher.group(1));
+        assertEquals("image/png", matcher.group(2));
+        
+        //test uploading
+        tester.clickLink("styleForm:styleEditor:editorContainer:toolbar:custom-buttons:1");
+        formTester = tester.newFormTester("dialog:dialog:content:form");
+        org.apache.wicket.util.file.File file = new org.apache.wicket.util.file.File(
+                getClass().getResource("GeoServer_75.png").getFile());
```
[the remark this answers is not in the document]
NielsCharlier: What is windows? ok, sorry about that.

PR #2825, opened 2018-04-03
[GEOS-8669] Allow importer to call a well known script as post transformation in task

aaime:


src/extension/importer/rest/src/test/java/org/geoserver/importer/rest/PostScriptTransformTest.java
```diff
@@ -0,0 +1,47 @@
...
+        File script = new File(scripts, "test.sh");
+        FileUtils.writeStringToFile(script, "touch test.properties\n");
+        script.setExecutable(true, true);
+    }
+
+    @Override
+    protected void tearDown() throws Exception {
+        super.tearDown();
+        File dir = dd.getRoot().dir();
+        FileUtils.deleteQuietly(dir);
```
bencaradocdavies: This causes:

[GEOS-8797] PostScriptTransformTest recursively deletes entire system temporary directory
https://osgeo-org.atlassian.net/browse/GEOS-8797

PR #2826, opened 2018-04-04
[GEOS-8672]: PNG + JPEG WPS Downloads

dromagnoli:


dromagnoli:
Right. For the moment, I have just mimic what the ImagePPIO was doing for RenderedImages.

PR #2828, opened 2018-04-05
Adapt Geoserver to JSR363 API for units of measure

dispiste:
Geotools is evaluating moving to JSR363 API for units of measure, see https://github.com/geotools/geotools/wiki/Migrate-Units-to-JSR-363

This branch adapts Geoserver to JSR363 API. It correctly compiles and passes unit tests if the right Geotools  branch has been previously compiled locally: https://github.com/dispiste/geotools/tree/test_units_jsr363_resync

Reviews are welcome.

### DIFF
--- a/src/community/netcdf-ghrsst/src/main/java/org/geoserver/wcs/responses/GHRSSTEncoder.java
+++ b/src/community/netcdf-ghrsst/src/main/java/org/geoserver/wcs/responses/GHRSSTEncoder.java
@@ -47,9 +47,9 @@ import ucar.nc2.dataset.NetcdfDataset;
 import ucar.nc2.write.Nc4Chunking;
 import ucar.nc2.write.Nc4ChunkingDefault;
 
-import javax.measure.converter.UnitConverter;
-import javax.measure.unit.NonSI;
-import javax.measure.unit.Unit;
+import javax.measure.UnitConverter;
+import si.uom.NonSI;
+import javax.measure.Unit;
 import javax.media.jai.iterator.RandomIter;
 import javax.media.jai.iterator.RandomIterFactory;
 import java.awt.image.RenderedImage;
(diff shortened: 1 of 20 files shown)


jodygarnett:
Wow @dispiste that was much faster then updating GeoTools!

jodygarnett:
@dispiste So we have a difference between a symbol and a name.

dispiste:
JSR363 API provides some methods in Unit interface:
- getName(): it should get the name
- getSymbol(): it should get the symbol

However, the implementation returns null for  getName() and/or  getSymbol() for most units. Some examples:
SI.METRE.getName() returns null
SI.METRE.getSymbol() returns "m"
SI.LITRE.getName() returns "Litre"
SI.LITRE.getSymbol() returns null
NonSI.DEGREE_ANGLE.getName() returns "Degree Angle"
NonSI.DEGREE_ANGLE.getSymbol() returns null
USCustomary.FOOT_SURVEY.getName() returns "US Survey foot"
USCustomary.FOOT_SURVEY.getSymbol() returns null

The toString() method relies on  SimpleUnitFormat.getInstance().format() method.

The SimpleUnitFormat.getInstance().format() method is based on the labels that have been registered using the label() method. If no labels have been registered, format uses getSymbol() for `BaseUnit`s and `AlternateUnit`s. For `TransformedUnit`s, a name is composed based on the BaseUnit format and the conversion factor.

Therefore, from a practical point of view, I would say we only have format() method, which we can configure in the best way we decide, but that it will fallback to getSymbol() to "unconfigured" units. Therefore, I'd consider it a replacement for getSymbol() and not for getName().

Note that there are 2 main use cases for names and symbols: formatting and parsing. We can only set one label for formatting, but we can use aliases to be able to parse alternate symbols or names.

jodygarnett:
@dispiste thanks again for you hard work, build.geoserver.org is reporting success across geotools and geoserver branches.

PR #2840, opened 2018-04-17
[GEOS-8687] WCS multidimensional requests (NetCDF/Grib outputs) may lead to un-closed readers and eventually OOMs

aaime:


aaime:
Depends on https://github.com/geotools/geotools/pull/1858

PR #2842, opened 2018-04-18
Multiple Windows Installer issues + developer documentation

ronaldhoek:
Fixes regarding the following issue:

[GEOS-8683] Windows Installation issue - upgrading GeoServer results in corrupt data_dir
[GEOS-8697] Remove language warnings during Windows setup compilation
[GEOS-8698] Windows Installer: Remove 'work' folder when uninstalling
[GEOS-8712] GeoServer Developer documentation contains invalid links to contributor agreements

### DIFF
--- a/doc/en/developer/source/policies/committing.rst
+++ b/doc/en/developer/source/policies/committing.rst
@@ -36,8 +36,8 @@ The process of getting community commit access is as follows:
 
 #. **Print, sign, scan and send the contributor agreement**
    
-   * `corporate_contributor.txt <http://www.osgeo.org/sites/osgeo.org/files/Page/corporate_contributor.txt>`_
-   * `individual_contributor.txt <http://www.osgeo.org/sites/osgeo.org/files/Page/individual_contributor.txt>`__
+   * `corporate_contributor <https://www.osgeo.org/resources/corporate-contributor-license/>`_
+   * `individual_contributor <https://www.osgeo.org/resources/individual-contributor-license/>`__
    
    Scanned assignment agreement can be emailed `info@osgeo.org <mailto:info@osgeo.org>`_ at OSGeo.
 
--- a/doc/en/developer/source/policies/committing.rst
+++ b/doc/en/developer/source/policies/committing.rst
@@ -8,8 +8,8 @@ Getting commit access
 
 All contributors are asked to provide an assignment agreement for working on the project:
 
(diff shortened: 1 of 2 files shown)


ianturton:
It's preferable to not mix solutions to multiple issues in one PR. I would have merged the doc changes just now but would need to review all the windows changes now.

aaime:
@ianturton that might be my fault, I suggested that he could have added more commits in a pull request. I meant it for the windows installer changes, so that whoever reviews and tries it out, has to boot into windows only one time :-p

aaime:
Hi, I had a look at the changes, they seem ok, and then tried the generated installer.
By your initial description of the problem I thought you were trying to run the installer on top of an existing installation, so I've tried the same (without uninstalling the previous version, trying to do a straight upgrade)... the result is that GeoServer does not start anymore, because in ``geoserver/web-inf/lib/`` there are now libraries of both the old and new version of GeoServer.
However, if I choose to uninstall the previous version, the internal data dir would be removed by the old installer, which I guess would make the issue non reproducible.
Could you clarify?

aaime:
Btw, in the meantime, I realized maybe what you were after was a "uninstall previous and install new" manually, but without losing the data directory in the process, which seems to be working fine.

ronaldhoek:
To clarify things a bit:

The initial problem I encountered was the corruption of the data_dir.
Looking into that specific problem, I also noticed a probleem regarding a duplicate service installation and the need to uninstall.

> Btw, in the meantime, I realized maybe what you were after was a "uninstall previous and install new" manually, but without losing the data directory in the process, which seems to be working fine.

I think this is what would be something the installer should do eventually, but since I first needed too learn the NSIS scripting, I thought it would be best to first pick up some smaller issues (starting with the data_dir issue).

Therefor I fixed the data_dir issue, by making sure a data_dir would not get corrupted/deleted/modified by the installer. And two small issues I encountered while testing.

Thanks for looking into the changes @aaime .

I would have made a separate PR, but I noticed the commits were automatically added to my precious PR. This is because my knowledge of Git(Hub) is limited at this time. So some pointers on how to a more commits to my repo without adding these to a previous PR would be nice..

aaime:
@ronaldhoek I just noticed you made pull requests out of your master branch, that's considered "bad practice".  When preparing a pull request, create a branch from master, do your commits, make a pull request from it. If you want to make a separate pull request, go back on master, create a new branch, and make a separate pull request from it.

About update installers, not sure if they are feasible on NSIS, but found this which could be of interest: http://nsis.sourceforge.net/Auto-uninstall_old_before_installing_new

PR #2868, opened 2018-05-07
Update automation.rst

daRasmussen:
Typo

### DIFF
--- a/doc/en/user/source/community/opensearch-eo/automation.rst
+++ b/doc/en/user/source/community/opensearch-eo/automation.rst
@@ -3,7 +3,7 @@
 Automation with the administration REST API
 ============================================
 
-The OpenSearch module sports full automation REST API that can be used to
+The OpenSearch module supports full automation REST API that can be used to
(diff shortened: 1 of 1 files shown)


bradh:
I think "sports" could be valid here (http://www.dictionary.com/browse/sport defn. 26), but supports definitely reads better.

aaime:
Yes, was not a typo, written as intended. But if we go with the notion that using it is unclear, there is a couple other places where the term is used:

````
>  git grep sports
en/developer/source/programming-guide/wps-services/design-guide.rst:The GoeServer code base sports by default two implementation of a ``ProcessFilter``:
fr/user/source/styling/sld-extensions/index.rst:GeoServer sports a number of vendor specific extensions to SLD 1.0.
````

(cleaned up the output to report only the actual "sports" matches)

bradh:
I will address those (plus that GoeServer thing too, and the other four places it occurs).

bradh:
The follow-up PR is at https://github.com/geoserver/geoserver/pull/2869

PR #2871, opened 2018-05-07
[GEOS-8724] Intermittent build failure in NamespacesWfsTest (backport 2.13.x)

nmco:
Associated issue:
https://osgeo-org.atlassian.net/browse/GEOS-8724


nmco:
@aaime you were correct this one was repeated indeed.

PR #2884, opened 2018-05-16
[GEOS-6233] app-schema gives error response to WFS 2 GetFeatureById s…

danieluct:
https://osgeo-org.atlassian.net/browse/GEOS-6233

Basically, if the current feature type detection from id mechanism fails (the likely case for app-schema layers), it replicates the getfeaturebyid query for each app-schema published feature type and returns the first hit (if any).

### DIFF
--- a/src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/FeatureGML32Test.java
+++ b/src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/FeatureGML32Test.java
@@ -11,7 +11,6 @@ import static org.junit.Assert.assertTrue;
 
 import java.util.ArrayList;
 import java.util.List;
-
 import org.geotools.filter.v2_0.FES;
 import org.geotools.wfs.v2_0.WFS;
 import org.junit.Test;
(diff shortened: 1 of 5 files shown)


bencaradocdavies:
@danieluct this pull request does not conform to the Google AOSP formatting used by this project (and detected that our format checking was broken!). I will now close and reopen this PR to retrigger Travis CI (now fixed) to demonstrate this. Maven will reformat it for you when you build it.

bencaradocdavies:
@danieluct these changes make sense to me but I would like an opinion from @nmco (and good, I see you have requested his review). The build and app-schema online tests against postgis both pass with these changes

nmco:
Ah! @bencaradocdavies I actually request this review for myself and forgot about it :(

bencaradocdavies:
Thanks @nmco, your opinion on LocalWorkspace usage would be appreciated.

nmco:
I actually remember stepping on this issue a few months ago, and at the time my investigation was reduced to the question: **what is a valid ID, or better what should be the format of a feature identifier ?** According to what other servers do and what GeoServer does for simple features the answer is:

``FEATURE_TYPE_NAME.ID``

For example querying ``tiger_roads`` layer using WFS ``GetFeatureById`` operation should be done like this:

```
https://<host>:<port>/geoserver/tiger/ows?service=wfs&version=2.0&request=GetFeature&storedquery_id=urn:ogc:def:query:OGC-WFS::GetFeatureById&id=tiger_roads.1
```

Note the usage of ``id=tiger_roads.1``. For simple features the identification of the layer \ feature type is done by this visitor:

```
https://github.com/geoserver/geoserver/blob/6058e38a0c74247ee6fea3db81c3aecbca1f40d2/src/main/src/main/java/org/geoserver/feature/TypeNameExtractingVisitor.java#L40-L45
```

When things start to get a bit weird is that if we use the same approach to query a complex feature type, i.e. a feature type defined through App-Schema, we get an feature not found exception. This happens because App-Schema does take into account that the feature id may contain the feature type and do a simple ``toString()`` when unrolling the query:

```
https://github.com/geotools/geotools/blob/865cc72a90b07841a4c871d1b07fcc8e5defd35b/modules/extension/app-schema/app-schema/src/main/java/org/geotools/data/complex/filter/UnmappingFilterVisitor.java#L329-L333
```

So summarizing, if we agree that a Feature ID should have the format ``FEATURE_TYPE_NAME.ID`` then we just need to update the App-Schema unrolling query behavior to make the behavior consistent with simple feature types. IMHO this is what makes sense, unfortunately (at least in my understanding) [OGC ``Feature ID`` definition](http://docs.opengeospatial.org/is/09-026r2/09-026r2.html#71) is a bit abstract ... but querying all the available complex types is quite an _wild_ thing to do and may very, very, very ... inefficient.

nmco:
Sorry for jumping in the middle this PR discussion and jeopardizing it a bit, but I really think that using the ``FEATURE_TYPE_NAME.ID`` may be the best long term option ... now time to sleep is getting later for me :sleepy:.

bencaradocdavies:
app-schema types usually set the id to something else. There are also workspaces and namespaces so names might need to be prefixed, and the requirement that ids are unique in the scope of each service.

danieluct:
I see the point of prefixing IDs with the feature type name, though I think there should still be an option to use prefix-less IDs in app-schema if explicitly requested in the config file, to account for database joins when changing the underlying DB schema/data is not possible . This PR, while not efficient, would then still be of use when prefix-less IDs are used. In spite of the brute force approach, when using PostGIS as store and IDs directly mapped to a(n indexed) DB column, this provides reasonable response times (less than 5 seconds on my dev machine, serving about 10,000 INSPIRE complex features); higher response times are expected when using non-DB stores or IDs mapped to expression. 

In case this PR is deemed useful, I'll ask to postpone its merge (or temporarily close it), as I have to check what happens when no feature is found in the auto-detected (simple feature) store (my intuition is that right now it continues to search for that ID in the complex features stores, which probably is unnecessary).

bencaradocdavies:
@danieluct I think this PR is useful, and even an inefficient algorithm is better than broken behaviour, as long as the brute force method does not make any currently working behaviour broken or slower. At your request, I will temporarily close this PR. Please comment here when you would like it reopened.

bencaradocdavies:
@danieluct other things to consider are deployments with thousands of layers or stores, and whether a brute force implementation could be exploited in a denial of service attack. I think we have some protection at the service level. I do not think it adds any vulnerability because clients can already make requests with queries that require brute force to evaluate.

nmco:
I agree with @bencaradocdavies, this is useful. The brute force approach could also be made optional, i.e. allow an administrator to enable or disable it, by default it should be disabled.

PR #2886, opened 2018-05-17
[GEOS-8750] Improve GeoJSON output format with WFS3 suggested properties

aaime:


src/community/wfs3/src/main/java/org/geoserver/wfs3/response/RFCGeoJSONFeaturesResponse.java
```diff
@@ -0,0 +1,189 @@
...
+
+    @Override
+    public String getMimeType(Object value, Operation operation) throws ServiceException {
+        return MIME;
+    }
+
+    public void write(Object value, OutputStream output, Operation operation) throws IOException {
+        // was it a single feature request?
+        String requestFeatureId = getWFS3FeatureId();
+        if (requestFeatureId != null) {
```
nmco: I guess the difference is that when there is only a single feature no array is used ? I dislike this, it will make client life more difficult for no clear reason, but well we need to respect the standard.

src/community/wfs3/src/main/java/org/geoserver/wfs3/response/RFCGeoJSONFeaturesResponse.java
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * Returns the WFS3 featureId, or null if it's missing or the request is not a WFS3 one
+     *
+     * @return
+     */
+    private String getWFS3FeatureId() {
+        Request dr = Dispatcher.REQUEST.get();
+        String featureId = null;
+        if (dr != null && (new Version(dr.getVersion()).getMajor().equals(3))) {
+            Object featureIdValue = dr.getKvp().get("featureId");
```
nmco: The case sensitive looks weird here, the WFS3 query parameters are case sensitive ?

src/community/wfs3/src/main/java/org/geoserver/wfs3/response/RFCGeoJSONFeaturesResponse.java
```diff
@@ -0,0 +1,189 @@
...
+            if (featureIdValue instanceof List) {
+                featureId = (String) ((List) featureIdValue).get(0);
```
nmco: Interesting, this probably deserves a log, or is just vouching for some weird passing rules ?

nmco:
Looks good to me, refactoring the GeoJSON writer code in a more elegant way, fixi existing tests and adds new ones to cover th new functionalities. App-Schema tests where also take into account, thank you.

src/community/wfs3/src/main/java/org/geoserver/wfs3/response/RFCGeoJSONFeaturesResponse.java
aaime: Go and complain with the spec writers, when there is a single feature, that's what the spec orders to to.

PR #2889, opened 2018-05-22
GEOS-8271 SLD files other catalog implementations

NielsCharlier:
https://osgeo-org.atlassian.net/browse/GEOS-8271


bencaradocdavies:
Close-reopen because Travis.

bencaradocdavies:
Travis CI timed out. I will kick it.

src/main/src/main/java/org/geoserver/config/GeoServerResourcePersister.java
```diff
@@ -0,0 +1,210 @@
+/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2013 OpenPlans
```
jodygarnett: This is a new file right? Does it need a 2018 header...

NielsCharlier:
header fixed & merged with new formatting

PR #2896, opened 2018-06-03
[GEOS-8768] OpenLayers preview do not report exceptions 'in image' anymore

aaime:
We can get at least this, it's not much, but better than a blank:

selezione_221

selezione_223


aaime:
Build failure is spurious file system related one in importer (this build still run on sudo=false). Merging.

PR #2901, opened 2018-06-05
[GEOS-8772] Add documentation about the :where_clause: place holder in SQL views \ virtual tables

nmco:
**Should only be merged after this PR https://github.com/geotools/geotools/pull/1919**

Associated issue:
https://osgeo-org.atlassian.net/browse/GEOS-8772


nmco:
Thank you for the review @smithkm, I have applied your suggestions:
- I activated SQL syntax making ti to read the example
- add an white space and comment to make it easier to spot the place holder
- add an explicit note about the the ``AND`` behavior

PR #2905, opened 2018-06-08
GEOS-8779 Allow for the FeatureSizeFeatureCollection to delegate

jnh5y:
* The FeatureSizeFeatureCollection can now delegate visitors to underlying datastores.

https://osgeo-org.atlassian.net/browse/GEOS-8779


aaime:
+1

PR #2912, opened 2018-06-12
Check filenames inside of archives

sikeoka:
Added filename checks when extracting files from user provided archives.


sikeoka:
The unit test error may be unrelated to this pull request.

src/rest/src/main/java/org/geoserver/rest/util/IOUtils.java
```diff
@@ -764,9 +764,10 @@ public static void inflate(
 
         final Enumeration<? extends ZipEntry> entries = archive.entries();
         try {
+            File destDir = outputDirectory.dir();
             while (entries.hasMoreElements()) {
                 ZipEntry entry = entries.nextElement();
-
+                org.geoserver.util.IOUtils.getZipOutputFile(destDir, entry);
```
smithkm: This is just as a check that it doesn't have a harmful path?  A get method call where the return value isn't used is a bit odd otherwise.  If so maybe add a comment explaining it.

smithkm:
Failed again, in a completely different location.  Kicking Travis again and hoping the third time's the charm.

src/rest/src/main/java/org/geoserver/rest/util/IOUtils.java
sikeoka: I added a comment to that line.

smithkm:
Great Success!

PR #2913, opened 2018-06-12
[GEOS-8751] install/deploy community gs-nsg-profiles for -PcommunityRelease

sikeoka:
https://osgeo-org.atlassian.net/browse/GEOS-8751

This pull request can be backported to 2.13.x

### DIFF
--- a/src/community/nsg-profiles/pom.xml
+++ b/src/community/nsg-profiles/pom.xml
@@ -12,8 +12,18 @@
   <packaging>pom</packaging>
   <version>2.14-SNAPSHOT</version>
   <name>NSG Profiles</name>
-  <modules>
-  	<module>nsg-wfs-profile</module>
-  	<module>nsg-wmts-profile</module>
-  </modules>
+  <profiles>
+    <profile>
+      <id>nsg-wfs-profile</id>
+      <modules>
+        <module>nsg-wfs-profile</module>
+      </modules>
+    </profile>
+    <profile>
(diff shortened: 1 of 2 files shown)


nmco:
Hi @sikeoka thank you for this contribution ! I create a PR proposing a slightly different approach build on top of what you did:
1. the handling of which profile should be activated is delegated to the NSG profiles module
2. a maven command executed inside NSG profiles module directory will hit all the profiles

If you agree with my approach  we can close this PR and merge these two:
- https://github.com/geoserver/geoserver/pull/2919
- https://github.com/geoserver/geoserver/pull/2920

sikeoka:
I tried to make gs-nsg-profiles consistent with how gs-security works.  I don't care too much about how it's done as long as everything gets installed correctly but wouldn't having two different ways to handle multi-module community projects be confusing for future developers?  I'll leave it up to you unless someone else wants to comment.

nmco:
Yes I know, security community module is doing it that way, actually recent modules have all their own way of handling this ... but if you have a look at older modules, like jms and script for example, they are all using the way I propose.

I agree that we should make consistent the way community plugins with multi-modules are handled. I will take that into account when reviewing new community modules contribution, maybe adding this to the documentation too ?

sikeoka:
That sounds good to me.

nmco:
Cool, merged my PRs closing this one, thank you for the discussion.

PR #2915, opened 2018-06-13
[GEOS-8785] Cannot access or delete single granule via REST when using multi-coverage views

aaime:


aaime:
Yeah, the failure is indeed unrelated

PR #2924, opened 2018-06-19
[GEOS-8789] FileWrapperResourceTheoryTest fails if java.io.tmpdir is relative

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOS-8789


bencaradocdavies:
Updated with a more general implementation that handles both cases.

PR #2926, opened 2018-06-19
GEOS-8779 Disable FeatureSizeFeatureCollection by default to avoid un…

elahrvivaz:
…optimized counts

* FeatureSizeFeatureCollection is disabled by default, to prevent unoptimized counts when a feature collection returns -1 for size()

https://osgeo-org.atlassian.net/browse/GEOS-8790

Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>


aaime:
@ahuarte47 please have a look?

elahrvivaz:
@bencaradocdavies thanks, could we also back-port to 2.13.x?

PR #2933, opened 2018-06-22
[GEOS-8797] PostScriptTransformTest recursively deletes entire system temporary directory

aaime:


bencaradocdavies:
Thanks @aaime, that looks good. I can confirm that this fixes the problem. Thanks also @smithkm for merging.

PR #2934, opened 2018-06-22
[GEOS-8550] Fix GWC Secrutiy NPE for layer group request

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-8550

NPE only occurs when you make a workspace-virtual WMS request against a workspaced layer group and all of the following are true:
- Caching is enabled
- GWC Data Security is enabled
- Direct WMS integration is enabled

Caused because LocalWorkspace has a NameDequalifyingProxy which will strip off the workspace if we just call prefixedName


jodygarnett:
This looks good thanks @tbarsballe

tbarsballe:
I'll backport, yes.

tbarsballe:
Backported to 2.13.x and 2.12.x.

PR #2935, opened 2018-06-22
[GEOS-8764] Allow LayerGroups in default worksaces to be found when workspace isn't specified

tbarsballe:
https://osgeo-org.atlassian.net/browse/GEOS-8764


jodygarnett:
This is a functionality change and cannot be back ported.

PR #2936, opened 2018-06-23
[GEOS-8563] Importer should correctly report INIT_ERROR state

tbarsballe:
When an error occurs during import initialization, the error is caught and the context state is set to INIT_ERROR.
However, later in initialization, Importer.changed() is called, which resets the state, overwriting this value with PENDING. This has been fixed, and Importer will now correctly report the INIT_ERROR state.

For some background info, see https://osgeo-org.atlassian.net/browse/GEOS-8563
Note - This doesn't really fix GEOS-8563, but does make it more clear that importer is in an error state when the issue occurs.

### DIFF
--- a/src/extension/importer/rest/src/test/java/org/geoserver/importer/rest/ImporterIntegrationTest.java
+++ b/src/extension/importer/rest/src/test/java/org/geoserver/importer/rest/ImporterIntegrationTest.java
@@ -91,6 +91,126 @@ public class ImporterIntegrationTest extends ImporterTestSupport {
     }
 
     @Test
+    public void testDirectWrongFile() throws Exception {
+
+        boolean async = false;
+        // set a callback to check that the request spring context is passed to the job thread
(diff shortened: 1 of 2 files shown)


src/extension/importer/rest/src/test/java/org/geoserver/importer/rest/ImporterIntegrationTest.java
```diff
@@ -90,6 +90,126 @@ public void createH2Store() throws Exception {
...
+                        + "        }\n"
+                        + "      }\n"
+                        + "   }\n"
+                        + "}";
+        // @formatter:on
+
+        // initialize the import
+        MockHttpServletResponse servletResponse =
+                postAsServletResponse(
+                        "/rest/imports?async=true", contextDefinition, "application/json");
```
bencaradocdavies: Use of `async=true` causes [GEOS-8818](https://osgeo-org.atlassian.net/browse/GEOS-8818), fixed in https://github.com/geoserver/geoserver/pull/2962.

PR #2939, opened 2018-06-25
Fix Dutch translation: submit/apply

NielsCharlier:


src/web/core/src/main/resources/GeoServerApplication_nl.properties
```diff
@@ -789,7 +789,8 @@ showingAllRecords = Resultaat {0} t/m {1} (totaal\: {2})
 
 stores = Bronnen
 
-submit = Toepassen
+submit = Verzenden
```
mprins: I would rather choose "Opslaan" instead of "Verzenden"

NielsCharlier: Well, it does seem to be used that way although that is not the conventional meaning of submit on the internet (which is normally "verzenden"). There is already "Save" which is translated as "Opslaan" but I'm not really sure why some pages in geoserver use "submit" instead.

NielsCharlier: To give some background, currently the page to create/edit styles has a button "Apply" and a button "Toepassen" and this is obviously very confusing.

mprins: what about "Indienen" ?

mprins: > To give some background, currently the page to create/edit styles has a button "Apply" and a button "Toepassen" and this is obviously very confusing.

yeah, I always hit the wrong button too :confused:

NielsCharlier: Yeah OK, that is fine too. It sounds a bit like giving your homework to the teacher, but it is actually the most accurate translation :)

NielsCharlier:
New version with "indienen"

PR #2940, opened 2018-06-25
[GEOS-8778, GEOS-8803] Keep track of whether gridsubset extent is dynamic

smithkm:
This retains the feature of being able to override the dynamic extent with a static one as from reading the code this appears to be intentional, and is a useful feature, if not one accessible by the UI.  

I did update the UI to indicate that this is happening  (GEOS-8803).  The mechanism for doing this is a bit crude as it uses an if statement when building the Wicket UI rather than a more proper Converter object.  If we ever make this editable via the UI, this will need to change, but it's good enough for now.

 Depends on https://github.com/GeoWebCache/geowebcache/pull/664


bencaradocdavies:
Travis CI bogus CoverageViewTest failure.

PR #2945, opened 2018-06-29
Update to JTS 1.15.1

jnh5y:
This depends on https://github.com/GeoWebCache/geowebcache/pull/663

For details see https://github.com/geotools/geotools/wiki/Upgrade-to-JTS-1.15

See also:

* https://osgeo-org.atlassian.net/browse/GEOS-8761


src/community/geomesa/pom.xml
```diff
@@ -1,178 +1,178 @@
...
-
-    <repositories>
-        <repository>
-            <id>conjars.org</id>
-            <url>http://conjars.org/repo</url>
-        </repository>
-    </repositories>
-
-</project>
+<?xml version="1.0" encoding="ISO-8859-1"?>
```
jodygarnett: this appears to be a reformat of a pom.xml file rather than an actual change

src/gwc/pom.xml
```diff
@@ -46,12 +46,6 @@
     <dependency>
       <groupId>org.geowebcache</groupId>
       <artifactId>gwc-georss</artifactId>
-      <exclusions>
-        <exclusion>
-           <groupId>com.vividsolutions</groupId>
-           <artifactId>jts</artifactId>
-        </exclusion>
-      </exclusions>
```
jodygarnett: you may want to keep the exclusion if geowebcache gets on a different version of jts then geoserver?

src/pom.xml
```diff
@@ -1884,7 +1884,7 @@
   <wicket.version>7.6.0</wicket.version>
   <ant.version>1.8.4</ant.version>
   <imageio-ext.version>1.1.24</imageio-ext.version>
-  <jaiext.version>1.0.23</jaiext.version>
+  <jaiext.version>1.0-SNAPSHOT</jaiext.version>
```
jodygarnett: should be 1.0.1

src/restconfig/src/test/java/org/geoserver/rest/catalog/FeatureTypeControllerTest.java
```diff
@@ -634,7 +634,7 @@ public void testCreateFeatureType() throws Exception {
                         + "  <attributes>\n"
                         + "    <attribute>\n"
                         + "      <name>the_geom</name>\n"
-                        + "      <binding>com.vividsolutions.jts.geom.MultiPolygon</binding>\n"
+                        + "      <binding>org.locationtech.jts.geom.MultiPolygon</binding>\n"
```
jodygarnett: It looks like the REST API describes attributes interns of their JTS class. While this does not amount to a REST API change, it is a change that is visible through the REST API and I wonder if downstream projects like gsconfig will be affected.

Checking now...

* https://github.com/boundlessgeo/gsconfig - no references
* https://github.com/geosolutions-it/geoserver-manager - one reference in a javadoc

PR #2951, opened 2018-06-30
[GEOS-8810] Upgrade commons-io dependency to 2.6

nmco:
Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-8810
**Depends on https://github.com/GeoWebCache/geowebcache/pull/666**


aaime:
+1

PR #2952, opened 2018-06-30
[GEOS-8814] Intermittent build failure caused by H2 log race with fixture data directory deletion

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOS-8814


aaime:
Yesssir, +1

bencaradocdavies:
Thanks @aaime.

PR #2953, opened 2018-06-30
 [GEOS-8815] Add profile for dependency checking

bradh:
https://osgeo-org.atlassian.net/browse/GEOS-8815

### DIFF
--- a/src/dependency-check-suppression.xml
+++ b/src/dependency-check-suppression.xml
@@ -1,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.1.xsd">
+    <suppress>
+        <notes><![CDATA[file name: gt-swing]]></notes>
+        <gav regex="true">^org\.geotools:gt-swing:.*$</gav>
+        <cve>CVE-2015-6737</cve>
+    </suppress>
+    <suppress>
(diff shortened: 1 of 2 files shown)


bencaradocdavies:
I like Jira tags in commit messages because I am lazy and and it is easy to search for them.

bradh:
@bencaradocdavies : Amended to include the JIRA ticket number.

aaime:
Tried it, good stuff! Couple of feedbacks:

- Running it from a single module does not seem to work fine, it fails with the following error:
``Warn initializing the suppression analyzer: Failed to load dependency-check-suppression.xml, caused by Suppression file 'dependency-check-suppression.xml' does not exist.``
- Running it from root, as it seems to be designed to, does not deliver an aggregate report, but this can be fixed by switching the goal from "check" to "aggregate"

I've also tried lowering the failure level, but it fails at the first issue it finds, instead of failing at the end. 
I was interested in it, because we could make a daily build with this profile. I guess a failure anywhere is good enough. And then have this build report to the PSC.

bradh:
I added support for dependency suppression that should work anywhere in the tree, and switched the goal to `aggregate`.

There are likely more suppressions that we'd need before going to automated runs - the "detection" code is somewhat prone to false alarms (especially on some older CVEs).

aaime:
Tested, looks good. Seems it won't survive a parallel build, pity cause the tests are slow, but I guess it's not a big deal, eventually we'll just need to run it once a day or something like that.

PR #2954, opened 2018-06-30
[GEOS-8816] Update Apache POI dependency to 3.17.

bradh:
Tested this by deploying an updated geoserver war + extension to tomcat; hitting the two standard URLs (per extension README, one is .xls and the other .xlsx), and opening them in LibreOffice. Data looked reasonable, headings were bolded as expected.


aaime:
Yep, +1

aaime:
Merged, backported to 2.13.x and 2.12.x

PR #2955, opened 2018-06-30
[GEOS-8240] Support X-Forwarded-Proto header in ResponseUtils.baseURL

lmoran:
Use of request headers in composing the Proxy URL, (see [GEOS-8240](https://osgeo-org.atlassian.net/browse/GEOS-8240)). NOTE: one validator in GlobalSettingsPage.java ha to be dropped.


src/main/src/main/java/org/geoserver/ows/QuickTemplate.java
```diff
@@ -0,0 +1,47 @@
+/* (c) 2017 Open Source Geospatial Foundation - all rights reserved
```
bencaradocdavies: If this is a new file, it needs the correct date in the copyright header. @lmoran I think you have signed a contributor agreement; please confirm.

src/main/src/main/java/org/geoserver/ows/QuickTemplate.java
```diff
@@ -0,0 +1,47 @@
...
+package org.geoserver.ows;
+
+import java.util.Map;
+
+/**
+ * Simpleton but fast template engine, replaces variables following the "${varName}" syntax into a
+ * string. No escaping, no extras, but avoids building lots of strings to do its work and the
+ * overhead of template instantiation of a true template engine.
+ *
+ * @author Andrea Aime - GeoSolutions
```
bencaradocdavies: @lmoran if this is a new file, please check the author tag. Was this copied from elsewhere?

src/main/src/main/java/org/geoserver/ows/ProxifyingURLMangler.java
```diff
@@ -5,33 +5,129 @@
...
+
+        // If the proxy base URL does not contain templates, fall back to
+        // the fixed URL cse
+        if (!proxyBase.contains(TEMPLATE_PREFIX)) {
+            return this.mangleURLFixedURL(baseURL, proxyBase);
+        }
+
+        // Cycles through templates in the proxy base URL until one is competely matched
+        Map<String, String> headers = this.compileHeadersMap();
+        for (String template : Arrays.asList(proxyBase.split(TEMPLATE_SEPARATOR))) {
```
bencaradocdavies: Did you consider using a `Pattern` to implement template expansion? See `InterpolationProperties` in gs-app-schema.

Note to GeoTools developers: we should really move `InterpolationProperties` to gt-metadata for wider use as this is the second thing I have seen this year that could have used it.

src/wms/src/main/java/org/geoserver/wms/legendgraphic/ColorMapLegendCreator.java
```diff
@@ -275,9 +275,10 @@ public void setBackgroundColor(final Color backGroundColor) {
         }
 
         public void setBand(final SelectedChannelType grayChannel) {
-            if (grayChannel != null)
-                this.grayChannelName = grayChannel.getChannelName().evaluate(null, String.class);
-
+            /* XXX FIXME: to savoid a compilation error
```
bencaradocdavies: savoid -> avoid. And is this something that has been fixed or something to be fixed? If required for compilation, perhaps just remove?

src/main/src/test/java/org/vfny/geoserver/util/URLProxifyingTest.java
```diff
@@ -0,0 +1,270 @@
+/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
```
bencaradocdavies: Incorrect header date in this new file.

src/main/src/main/java/org/geoserver/ows/QuickTemplate.java
aaime: It's a copy of one class I made for the OpenSearch for EO.

src/main/src/main/java/org/geoserver/ows/ProxifyingURLMangler.java
aaime: The implementation was meant, as described in the javadoc, to be as fast as possible (Freemarker was too slow for the job, given that we potentially have a different template for each record). I did not know about InterpolationProperties, performance should be comparable (relative to large Freemarker init time for new templates, at the very least).

src/main/src/main/java/org/geoserver/ows/QuickTemplate.java
bencaradocdavies: If the class is a copy from elsewhere in the codebase, then it best to leave the copyright date and author tag the same.

src/wms/src/main/java/org/geoserver/wms/legendgraphic/ColorMapLegendCreator.java
lmoran: My bad: his shouldn't have been added to the commit, as it is an unrelated issue. I will return it to its original state in a another commit.

src/main/src/main/java/org/geoserver/ows/QuickTemplate.java
lmoran: Yes, copied and, later, adapted.

src/main/src/main/java/org/geoserver/ows/ProxifyingURLMangler.java
lmoran: IIUC InterpolationProperties is case-sensitive in its substitutions, which clashes with the case-in-sensitiveness of HTTP headers: what about adding this option to it?

bencaradocdavies:
Thanks @lmoran.

aaime:
Ouch, was this tested interactively? http://cloudsdi.geo-solutions.it/geoserver/ is completely bricked by this change, capabilities documents, preview page, services building backlinks, they all fail with a stack trace like:

{code}
Caused by: java.lang.NullPointerException
	at org.geoserver.ows.ProxifyingURLMangler.mangleURL(ProxifyingURLMangler.java:93)
	at org.geoserver.ows.util.ResponseUtils.buildURL(ResponseUtils.java:372)
	at org.geoserver.ows.util.ResponseUtils.buildSchemaURL(ResponseUtils.java:403)
	at org.geoserver.wms.WMSServiceExceptionHandler.handleXmlException(WMSServiceExceptionHandler.java:321)
	at org.geoserver.wms.WMSServiceExceptionHandler.handleServiceException(WMSServiceExceptionHandler.java:124)
	at org.geoserver.ows.Dispatcher.handleServiceException(Dispatcher.java:1799)
{code}

src/main/src/main/java/org/geoserver/ows/ProxifyingURLMangler.java
```diff
@@ -1,37 +1,193 @@
...
+        String proxyBase =
+                (GeoServerExtensions.getProperty(Requests.PROXY_PARAM) != null)
+                        ? GeoServerExtensions.getProperty(Requests.PROXY_PARAM)
+                        : this.geoServer.getSettings().getProxyBaseUrl();
+
+        // Mangles the URL base in different ways based on a flag
+        // (for two reasons: a) speed; b) to make the admin aware of
+        // possible security liabilities)
+        baseURL =
+                (this.geoServer.getGlobal().isUseHeadersProxyURL() == true && proxyBase != null)
```
aaime: Believe the NPE is here, isUseHeadersProxyURL() return a Boolean, which can be null

lmoran: Yes, I tested it interactively, but not the on the latest GeoTools.

src/main/src/main/java/org/geoserver/config/impl/GeoServerInfoImpl.java
```diff
@@ -40,6 +40,8 @@
 
     protected Boolean globalServices = true;
 
+    protected Boolean useHeadersProxyURL = false;
```
aaime: This is guaranteed to be null when XStream deserializes from XML, if the element is missing, as it won't go through normal class construction (so forget about constructors and field initializers, that are also invoked as part of normal constructor invocation).

src/main/src/main/java/org/geoserver/ows/ProxifyingURLMangler.java
aaime: Yes, the geoserver field is always not null, this class is instantiated by Spring

src/main/src/main/java/org/geoserver/config/impl/GeoServerInfoImpl.java
lmoran: OK, I will solve this tomorrow.

src/main/src/main/java/org/geoserver/ows/ProxifyingURLMangler.java
bencaradocdavies: The NPE has also broken all CITE tests with `java.lang.NullPointerException at org.geoserver.ows.ProxifyingURLMangler.mangleURL(ProxifyingURLMangler.java:93)`
https://build.geoserver.org/job/cite-wms-1.3/1735/consoleFull

src/main/src/main/java/org/geoserver/config/impl/GeoServerInfoImpl.java
lmoran: Hopefully this is solved: please, refer to this other PR: #2968

PR #2960, opened 2018-07-01
[GEOS-8814] Fix loop failure condition off-by-one error

bencaradocdavies:
https://osgeo-org.atlassian.net/browse/GEOS-8814

Fix for a pre-existing loop failure condition off-by-one error found while backporting.


bencaradocdavies:
@aaime an interesting consequence is that this error was masking permanent directory removal failures on Windows.

PR #2967, opened 2018-07-04
[GEOS-8821] Upgrade spring-security to 4.2.7-RELEASE.

bradh:
Tested by booting on tomcat8, logged in OK.

Made sure the CAS extension unpacked OK, tomcat8 restarted afterwards, and that the CAS authentication filter showed up - couldn't test whether it worked correctly, since I don't have a CAS server.

I did see one problem where  geoserver (presumably spring-security) complained about a URL with a ";" character being potentially malicious, but couldn't reproduce and nothing in the logs. More testing would be appreciated.

### DIFF
--- a/src/extension/security/cas/pom.xml
+++ b/src/extension/security/cas/pom.xml
@@ -47,5 +47,11 @@
       <artifactId>easymockclassextension</artifactId>
       <scope>test</scope>
     </dependency>
-   </dependencies>
+    <dependency>
+      <groupId>org.jasig.cas.client</groupId>
+      <artifactId>cas-client-core</artifactId>
+      <version>3.4.1</version>
+      <type>jar</type>
+    </dependency>
+  </dependencies>
 </project>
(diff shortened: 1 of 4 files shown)


mcrmcr:
The CAS setup is cumbersome. The test simulates the client and the service requiring authentication.
It is necessary to install and start CAS and to distribute self signed certificates to enable HTTPS connections from the test case to the CAS server and from the CAS server to the test case (the test case starts the Java built in HTTP server to simulate the service).

IMHO opinion the best solution would be to start a prepared Docker Container running a preconfigured CAS Server. There is a also  nice Java API for Docker handling. But this would require a public GeoServer/Geotools Docker registry and I think there is none.

aaime:
I've merged the pull request on master, and tried to get some attention internally at GeoSolutions to do some extra tests on CAS and OAuth. Can we hold up the backport a few days? 
The interactive tests should be done end of this week/beginning of next (tentatively).

bradh:
Thanks. More than happy to hold the backport.

PR #2969, opened 2018-07-04
[GEOS-8800] Reverting style changes to legend span. No longer blocks …

travislbrundage:
…certain menu items.

Resolves [GEOS-8800](https://osgeo-org.atlassian.net/projects/GEOS/issues/GEOS-8800)

This change reverts a previous change introduced in tbarsballe#6, which was to resolve [GEOS-8746](https://osgeo-org.atlassian.net/projects/GEOS/issues/GEOS-8746). As far as I can tell, the change to the `legend span` element's CSS did not contribute to fixing the fullscreen style editor UI, only the `fieldset` element's CSS was needed. Please correct me if I'm wrong, @aaime .

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/css/geoserver.css
+++ b/src/web/core/src/main/java/org/geoserver/web/css/geoserver.css
@@ -376,9 +376,8 @@ border: 0 none transparent;
 legend span { 
 position: absolute;
 width: 29.2em;
-top: -3em;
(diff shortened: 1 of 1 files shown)


aaime:
Ah yeah, I did notice the problem this week, it became hard to select the style format in the "new style" page. Seems to work fine (I just tried to change the CSS on the fly in a browser, same changes as this PR)

PR #2970, opened 2018-07-05
Add osgeo dependency to implementing wps tutorial

lubojr:
Added osgeo dependency to fix Maven build. 
These were not downloaded before: 
*The following artifacts could not be resolved: jgridshift:jgridshift:jar:1.0, javax.media:jai_imageio:jar:1.1, javax.media:jai_codec:jar:1.1.3, javax.media:jai_core:jar:1.1.3: Could not find artifact jgridshift:jgridshift:jar:1.0 in boundless*

geoserver-users mail list message linked to this:
https://sourceforge.net/p/geoserver/mailman/message/36360921/

### DIFF
--- a/doc/en/developer/source/programming-guide/wps-services/implementing.rst
+++ b/doc/en/developer/source/programming-guide/wps-services/implementing.rst
@@ -102,12 +102,12 @@ For this example the project will be called "hello_wps".
             <repository>
                 <id>boundless</id>
                 <name>Boundless Maven Repository</name>
-                <url>http://repo.boundlessgeo.com/main</url>
+                <url>https://repo.boundlessgeo.com/main</url>
             </repository>
 	    <repository>
                 <id>osgeo</id>
                 <name>Open Source Geospatial Foundation Repository</name>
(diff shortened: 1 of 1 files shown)


lubojr:
I think it is not required either for osgeo or boundlessgeo actually, because it went through without both having snapshots = true. Should I remove it then?

aaime:
Probably best to remove them

Gnafu:
Since osgeo is already redirecting to https I suggest to use that instead of plain http

aaime:
@Gnafu if I follow http://download.osgeo.org/webdav/geotools/ with the browser I'm getting no redirection, but both http and https seem to work.

aaime:
Doh weird... now it does... maybe I had a cached page.

Gnafu:
With curl the server does not redirect. Maybe its a FF/Chrome feature: once you reach a resource with https, they use it for http links automatically.

lubojr:
Changed to https :)

aaime:
Thanks!

PR #2972, opened 2018-07-05
[GEOS-8824] Refactor duplicated code in OAuth modules

aaime:


aaime:
I have tested interactively gihub and google logins post refactor

PR #2973, opened 2018-07-05
GEOS-8779 Disable FeatureSizeFeatureCollection by default to avoid un…

elahrvivaz:
…optimized counts

* FeatureSizeFeatureCollection is disabled by default, to prevent unoptimized counts when a feature collection returns -1 for size()

https://osgeo-org.atlassian.net/browse/GEOS-8790

Signed-off-by: Emilio Lahr-Vivaz <elahrvivaz@ccri.com>

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/FeatureSizeFeatureCollection.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/FeatureSizeFeatureCollection.java
@@ -30,8 +30,8 @@ import org.opengis.feature.type.FeatureType;
  */
 public class FeatureSizeFeatureCollection extends DecoratingSimpleFeatureCollection {
 
-    /** The default feature cache size */
-    public static final int DEFAULT_CACHE_SIZE = 16;
+    /** The default feature cache size - disabled by default */
+    public static final int DEFAULT_CACHE_SIZE = 0;
 
     /** The original feature source. */
     protected SimpleFeatureSource featureSource;
--- a/src/wfs/src/test/java/org/geoserver/wfs/GetFeatureCachingTest.java
+++ b/src/wfs/src/test/java/org/geoserver/wfs/GetFeatureCachingTest.java
@@ -6,21 +6,23 @@ package org.geoserver.wfs;
 
 import org.geoserver.wfs.v2_0.GetFeatureTest;
 import org.junit.AfterClass;
+import org.junit.Assert;
 import org.junit.BeforeClass;
 
 /**
  * Test making sure GetFeature still works when the caching machinery is off (since most test
  * datasets will actually be cached in memory)
  */
-public class GetFeatureNoCachingTest extends GetFeatureTest {
+public class GetFeatureCachingTest extends GetFeatureTest {
 
     @BeforeClass
-    public static void disableCaching() {
-        FeatureSizeFeatureCollection.setFeatureCacheLimit(0);
+    public static void enableCaching() {
+        Assert.assertEquals(0, FeatureSizeFeatureCollection.DEFAULT_CACHE_SIZE);
+        FeatureSizeFeatureCollection.setFeatureCacheLimit(16);
     }
 
     @AfterClass
-    public static void enableCaching() {
+    public static void disableCaching() {
         FeatureSizeFeatureCollection.setFeatureCacheLimit(
                 FeatureSizeFeatureCollection.DEFAULT_CACHE_SIZE);
     }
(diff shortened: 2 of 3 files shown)


elahrvivaz:
This is a back-port to 2.13, original PR was #2926

src/wfs/src/main/java/org/geoserver/wfs/FeatureSizeFeatureCollection.java
```diff
@@ -30,8 +30,8 @@
  */
 public class FeatureSizeFeatureCollection extends DecoratingSimpleFeatureCollection {
 
-    /** The default feature cache size */
-    public static final int DEFAULT_CACHE_SIZE = 16;
+    /** The default feature cache size - disabled by default */
+    public static final int DEFAULT_CACHE_SIZE = 0;
```
jodygarnett: So this results in a change of behaviour, do we need a documentation note saying how to enable the behaviour?

src/wfs/src/test/java/org/geoserver/wfs/GetFeatureCachingTest.java
```diff
@@ -6,21 +6,23 @@
...
-        FeatureSizeFeatureCollection.setFeatureCacheLimit(0);
+    public static void enableCaching() {
+        Assert.assertEquals(0, FeatureSizeFeatureCollection.DEFAULT_CACHE_SIZE);
+        FeatureSizeFeatureCollection.setFeatureCacheLimit(16);
     }
 
     @AfterClass
-    public static void enableCaching() {
+    public static void disableCaching() {
         FeatureSizeFeatureCollection.setFeatureCacheLimit(
```
jodygarnett: While this is a programmatic way to enable, GeoServer often uses a system property (to control internal functionality changes)

src/wfs/src/main/java/org/geoserver/wfs/FeatureSizeFeatureCollection.java
elahrvivaz: I couldn't find any documentation indicating the original change in behavior... but I'm happy to add some. Do you have a recommended spot it should go?

src/wfs/src/test/java/org/geoserver/wfs/GetFeatureCachingTest.java
jnh5y: I believe this controlled by a system property: https://github.com/ccri/geoserver/blob/a850eba6036d39255f74c82995dd89f5d07ff215/src/wfs/src/main/java/org/geoserver/wfs/FeatureSizeFeatureCollection.java#L55-L57.

elahrvivaz: I think this method was just used for testing - the initial value is controlled by the system property here: https://github.com/ccri/geoserver/blob/a850eba6036d39255f74c82995dd89f5d07ff215/src/wfs/src/main/java/org/geoserver/wfs/FeatureSizeFeatureCollection.java#L55-L57

jnh5y: Jinx!  I said it first!

jodygarnett: Do you know if "org.geoserver.wfs.getfeature.cachelimit" is documented somewhere?

However since this a back port I do not wish to hold it up on docs.

src/wfs/src/main/java/org/geoserver/wfs/FeatureSizeFeatureCollection.java
elahrvivaz: maybe https://github.com/geoserver/geoserver/blob/master/doc/en/user/source/production/config.rst ?

jodygarnett: Either on the WFS page or on the store page?

Actually there is a cache your data heading here - http://docs.geoserver.org/latest/en/user/production/config.html

src/wfs/src/test/java/org/geoserver/wfs/GetFeatureCachingTest.java
elahrvivaz: it does not appear to be documented anywhere, but I don't mind adding it, I'm just not sure where would be appropriate.

elahrvivaz:
Added documentation note on enabling caching, and opened #2975 for the same change against master

elahrvivaz:
thanks @jodygarnett !

PR #2983, opened 2018-07-08
Assorted fixes to make build pass on Windows (INCOMPLETE, DO NOT MERGE!)

aaime:
Assorted fixes to improve the ability to build on Windows. Themes:

* Make sure streams and readers are closed
* When possible instantiate a reader instead of a input stream, at least from Strings it picks the right charset.
* On Windows newline are all over the place, Git checkout can contain "\n" (there is no developer rule on how newlines should be handled in checkouts, git offers a variety of options). Make tests work regardless.
* Do not try to read a file and write on it at the same time.
* Do not assume file separators are "/", but consider that in some cases they can still be "/" even on Windows (e.g. from the classpath).
* Bash may be available on Windows too, but does not mean that .sh files are considered executable.

The work is incoplete but wanted to share because:
- Hopefully someone else will chime in and help
- Need to check if it also builds on Linux


src/wfs/src/test/java/org/geoserver/wfs/v2_0/StoredQueryTest.java
```diff
@@ -479,6 +479,7 @@ public void testCreateWithLocalNamespaceDeclaration() throws Exception {
 
         // create
         Document dom = postAsDOM("wfs", xml);
+        print(dom);
```
tbarsballe: Unnecessary print?

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/DefaultGeometryTest.java
```diff
@@ -427,6 +427,7 @@ public void testWmsGetFeatureInfo(
                         x,
                         y);
         Document document = getAsDOM(getFeatureUrl);
+        print(document);
```
tbarsballe: Unnecessary print?

PR #2991, opened 2018-07-09
Set the EntityResolver in the WPS Request Builder when building the Execute XML.

sikeoka:
This pull request ensures that the EntityResolver is set on the WPSExecuteTransformer object before using the transformer to builder the Execute XML and updates WPSExecuteTransformer to use the EntityResolver when internally parsing text as XML.  WPSRequestBuilderPanel contained unused code to build the Execute XML that was removed.

This pull request can be backported to 2.13.x and probably 2.12.x.


jodygarnett:
tested locally, merging and back porting

PR #2997, opened 2018-07-10
GEOS-8837 get rid of hazelcast cache

NielsCharlier:


sikeoka:
Shouldn't the HzExtensionFilter bean in the applicationContext.xml also be commented out since its purpose is to suppress the JDBCCacheProvider?  The DefaultCacheProvider and JDBCCacheProvider are very similar.  The JDBC cache uses soft values instead of weak values and allows the maximum size and expiration time to be configured through system properties.

NielsCharlier:
yes, you are right, I have adjusted it.

PR #2998, opened 2018-07-10
GEOS-8822 publishing tab cannot be submitted with hazelcast session

NielsCharlier:
https://osgeo-org.atlassian.net/browse/GEOS-8822

The bug itself cannot really be tested in unit test, but existing unit tests should cover that this does nothing bad.

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/data/resource/LayerModel.java
+++ b/src/web/core/src/main/java/org/geoserver/web/data/resource/LayerModel.java
@@ -16,20 +16,20 @@ import org.geoserver.web.GeoServerApplication;
  * @author Andrea Aime - OpenGeo
  */
 @SuppressWarnings("serial")
-public class LayerModel implements IModel {
+public class LayerModel implements IModel<LayerInfo> {
     LayerInfo layerInfo;
 
     public LayerModel(LayerInfo layerInfo) {
         setObject(layerInfo);
     }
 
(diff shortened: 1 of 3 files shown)


aaime:
@bencaradocdavies the info will also be a LayerGroupInfo (the page has subclasses), but there is no parallel LayerGroupModel that performs the detaching like the layer model does. There are detachable versions of both layer and group, where detach is to be interpreted as "keep an id and reload from catalog after session restore" but don't think they are suitable for an editing page, only for read only references (might be wrong).

I see there is no test... forcing session serialization by just hitting elements in the UI will likely not work (otherwise I guess we'd see the issue without the Hazelcast clustering plugin in the mix), but maybe this approach could be used, with adaptations:
https://github.com/apache/wicket/blob/master/wicket-core/src/test/java/org/apache/wicket/serialize/java/JavaSerializerTest.java#L54

NielsCharlier:
Note that an IF statement ensures that the LayerModel is only used for LayerInfo, and not for LayerGroupInfo. The sync problem doesn't exist for LayerGroupInfo because there is no reference to the catalog object in there.

@bencaradocdavies will make the change

@aime ok, I will try to make a test with your tip.

NielsCharlier:
@bencaradocdavies the annotation were added.

A test was added which proves expected behaviour and I confirmed it didn't work before the patch and did work after.

PR #3002, opened 2018-07-11
[GEOS-8843] Env function in RasterSymbolizer's ChannelSelection doesn't evaluate in layer groups or layer lists

aaime:
Won't compile until https://github.com/geotools/geotools/pull/1973 is merged


src/main/src/main/java/org/geoserver/threadlocals/EnvVarThreadLocalTransfer.java
```diff
@@ -0,0 +1,30 @@
+/* (c) 2018Open Source Geospatial Foundation - all rights reserved
```
dromagnoli: Typo here (not a deadly problem :) ).
I would add a whitespace between year and Organization

PR #3003, opened 2018-07-12
GEOS-8839: Removing any reference to GTOPO30 modules/support

dromagnoli:


aaime:
Yep, looks ok to me. Did you try starting a GeoServer on this data dir? Any suspicious error message?

dromagnoli:
Yes, I did it.
The only error message I got was an xstream StreamException occurring during the logging initialization, but the same error message occurs from a clean datadir on the original code (I think it's related to missing logging.xml file being generated afterwards)

PR #3007, opened 2018-07-13
[GEOS-8844]: access wrapped GridCoverage

dromagnoli:


dromagnoli:
Updated as per suggestion (thanks).

aaime:
+1

PR #3009, opened 2018-07-13
Set the EntityResolver in OWS Dispatcher before parsing SOAP requests.

sikeoka:
This pull request updates the OWS Dispatcher to set the EntityResolver on the DocumentBuilder before using it to parse SOAP requests.  Reflection is used here because the EntityResolverProvider class is in the gs-main module.  This pull request also fixes a race condition parsing SOAP requests since DocumentBuilder instances are not thread-safe.

This pull request can be backported to 2.13.x and 2.12.x


aaime:
Looks good, will attempt backport.

PR #3016, opened 2018-07-20
Broken nodata handling of native JAI

eLvErDe:
Hello,

Native JAI has some serious issues handling nodata when granule are overlapping. For instance, I haven't been able to get it considering FFFFFFF has nodata color.
It has been confirmed by GeoSolutions.it people that it's kinda broken:
https://sourceforge.net/p/geoserver/mailman/message/36372215/

Regards, Adam.


aaime:
Comment was too specific to image mosaics, merged a more general version of the same statement instead. Thanks for the fix anyways :-)

PR #3024, opened 2018-07-30
GEOS-8861 - [Community GeoFence] Update GeoFence Plugin to JTS 1.15.1 and put it back to the build

afabiani:
Depends on https://github.com/geoserver/geofence/pull/114

### DIFF
--- a/src/community/geofence-server/pom.xml
+++ b/src/community/geofence-server/pom.xml
@@ -58,7 +58,7 @@
         <dependency>
             <groupId>org.opengeo</groupId>
             <artifactId>geodb</artifactId>
-            <version>0.9</version>
+            <version>${opengeo-geodb.version}</version>
         </dependency>
 
         <dependency>
--- a/src/community/geofence-server/pom.xml
+++ b/src/community/geofence-server/pom.xml
@@ -58,7 +58,7 @@
         <dependency>
             <groupId>org.opengeo</groupId>
             <artifactId>geodb</artifactId>
-            <version>0.8</version>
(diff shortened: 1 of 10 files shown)


src/community/geofence-server/pom.xml
```diff
@@ -58,7 +58,7 @@
         <dependency>
             <groupId>org.opengeo</groupId>
             <artifactId>geodb</artifactId>
-            <version>0.8</version>
+            <version>0.9</version>
```
jodygarnett: is it worth putting this in the root pom dependencies management section?

src/community/geofence/src/main/java/org/geoserver/geofence/rest/xml/JaxbRule.java
```diff
@@ -299,7 +299,11 @@ public JaxbRule(Rule rule) {
             layerDetails = new LayerDetails();
             layerDetails.setAllowedArea(rule.getLayerDetails().getArea());
             layerDetails.getAllowedStyles().addAll(rule.getLayerDetails().getAllowedStyles());
-            layerDetails.setCatalogMode(rule.getLayerDetails().getCatalogMode().toString());
+            if (rule.getLayerDetails().getCatalogMode() != null) {
```
jodygarnett: this appears to be a fix unrelated to upgrade?

afabiani:
@jodygarnett @jnh5y I did changes accordingly to your feedbacks. Many thanks for reviewing it.

PR #3025, opened 2018-08-01
GEOS-8827: introduce caching for remote styles in WMS GetMap

mbarto:


src/pom.xml
```diff
@@ -738,6 +738,16 @@
     <artifactId>commons-httpclient</artifactId>
     <version>3.1</version>
    </dependency>
+   <dependency>
+    <groupId>org.apache.httpcomponents</groupId>
+    <artifactId>httpclient</artifactId>
+    <version>4.5.5</version>
```
aaime: This was already declared below in the same file.
Mind, the httpcomponents are used by some community modules (authkey, geogig, onelogin) this could cause packaging and installation issues (duplicate jars with potentially different version). Maybe it's best to remove the version from the community modules now that v4 is used by core, so that they are all aligned.

src/web/wms/src/main/resources/GeoServerApplication.properties
```diff
@@ -154,6 +154,11 @@ WMSAdminPage.dynamicStylingDisabled = Disable usage of SLD and SLD_BODY paramete
 WMSAdminPage.disableFeaturesReprojectTitle = GetFeatureInfo results reprojection
 WMSAdminPage.disableFeaturesReproject = Disable the reprojection of GetFeatureInfo results
 
+WMSAdminPage.remoteStylesCache = Remote Styles Cache
+WMSAdminPage.cacheEnabled = Enabled
+WMSAdminPage.cacheEntries = Max entries
+WMSAdminPage.cacheEntrySize = Max entry size
```
aaime: Probably best to specify what is the unit of measure (bytes, kb, mb?)

src/wms/src/main/java/org/geoserver/wms/map/GetMapKvpRequestReader.java
```diff
@@ -96,16 +108,44 @@
...
+        this(wms, null);
+    }
+
+    public GetMapKvpRequestReader(WMS wms, HttpClientConnectionManager manager) {
         super(GetMapRequest.class);
         this.wms = wms;
         this.entityResolverProvider = new EntityResolverProvider(wms.getGeoServer());
+
+        // configure the http client used to fetch remote styles
+        RequestConfig requestConfig = RequestConfig.custom().build();
```
aaime: Does it make any difference between this an using RequestConfig.DEFAULT directly?

src/wms/src/main/java/org/geoserver/wms/map/GetMapKvpRequestReader.java
```diff
@@ -325,51 +365,110 @@ public GetMapRequest read(Object request, Map kvp, Map rawKvp) throws Exception
 
             URL styleUrl = getMap.getStyleUrl();
 
-            if (getMap.getValidateSchema().booleanValue()) {
-                InputStream input = Requests.getInputStream(styleUrl);
-                List errors = null;
-
+            InputStream input = null;
+            if (styleUrl.getProtocol().toLowerCase().indexOf("http") == 0) {
```
aaime: This block has become big enough that I'd suggest factoring it out in its own method if possible (e.g., getStyleStream(styleURL))

src/wms/src/main/java/org/geoserver/wms/WMS.java
```diff
@@ -144,6 +144,8 @@
 
     public static final String FEATURES_REPROJECTION_DISABLED = "featuresReprojectionDisabled";
 
+    public static final String CACHE_REMOTE_STYLES = "cacheRemoteStyles";
```
aaime: Unused constant

src/wms/src/main/java/org/geoserver/wms/WMSXStreamLoader.java
```diff
@@ -63,7 +63,10 @@ public static void initXStreamPersister(XStreamPersister xp) {
         xs.alias("wms", WMSInfo.class, WMSInfoImpl.class);
         xs.registerConverter(new WMSInfoConverter(xp));
         xs.addDefaultImplementation(WatermarkInfoImpl.class, WatermarkInfo.class);
```
aaime: Unsure here, but how does the generated look like, does it conform to the rest, or do we need to add an alias for CacheConfiguration? While we should not worry too much about the disk storage, the config is also visible in the REST API, both XML and JSON representations.

src/wms/src/main/java/org/geoserver/wms/map/GetMapKvpRequestReader.java
```diff
@@ -96,16 +108,44 @@
     /** EntityResolver provider, used in SLD parsing */
     EntityResolverProvider entityResolverProvider;
 
+    /** HTTP client used to fetch remote styles * */
+    CloseableHttpClient httpClient;
```
aaime: I don't see attempts at closing it though (if it's kept as a field, should be done when the spring app-context goes down, making the bean implement DisposableBean)

src/wms/src/main/java/org/geoserver/wms/map/GetMapKvpRequestReader.java
```diff
@@ -96,16 +108,44 @@
...
      * This flags allows the kvp reader to go beyond the SLD library mode specification and match
      * the first style that can be applied to a given layer. This is for backwards compatibility
      */
     private boolean laxStyleMatchAllowed = true;
 
     public GetMapKvpRequestReader(WMS wms) {
+        this(wms, null);
+    }
+
+    public GetMapKvpRequestReader(WMS wms, HttpClientConnectionManager manager) {
```
aaime: I think this is the main problem of the current approach, the GetMapKvpRequestReader is only initialized once at startup, past that it won't be re-created and thus changes to the chaching configuration will be ignored (at least, I believe that what will happen). If the client needs to be preserved in order to allow caching, then it should probably be managed externally and rebuilt every time the config changes.

src/wms/src/main/java/org/geoserver/wms/WMSXStreamLoader.java
mbarto: in the storage it looks like this:
```xml
<cacheConfiguration>
    <enabled>false</enabled>
    <maxEntries>1000</maxEntries>
    <maxEntrySize>51200</maxEntrySize>
</cacheConfiguration>
```
Do you think it should work also for the others?

aaime: Yep, looks good to me

aaime:
Hi Mauro, yep, looks good. I see there is one file in conflict, this prevents merge.

PR #3026, opened 2018-08-01
GEOS-8871: adding missing return to proxy url lookup

jdeolive:


src/wfs/src/main/java/org/vfny/geoserver/wfs/servlets/TestWfsPost.java
```diff
@@ -316,13 +316,20 @@ public String errorResponse(Exception e) {
...
-            geoServer.getGlobal().getSettings().getProxyBaseUrl();
+            return geoServer.getGlobal().getSettings().getProxyBaseUrl();
         }
         return null;
     }
 
+    /*
+     * Protected for testing purposes.
+     */
+    protected GeoServer getGeoServer() {
```
smithkm: Minor suggestion, it might be worth using `@VisibleForTesting` annotation instead of a comment.

jratike80:
Before merging ask Ben or Andrea if it is OK to make this change.  From https://sourceforge.net/p/geoserver/mailman/message/35692680/

> support for demo requests against URLs other than
> http://localhost:8080/geoserver was removed in 2.10.1 because it is a
> vulnerability both in that it allows access to otherwise inaccessible
> internal hosts using GeoServer as a proxy, and because the proxy allows
> circumvention of browser cross-domain security restrictions

aaime:
I think it's fine, the change allows to use GeoServer demo requests on proxied GeoServers, but only from the known external address, should not be turning it back to an open proxy like before.

jdeolive:
Thanks for the review guys!

smithkm:
Hey Justin,  looks like you merged without fixing the formatting errors.  I've got formatting changes appearing in these files when I run unit tests against master now.

jdeolive:
Whoops, my bad, sorry about that guys. Thanks for bailing me out Kevin. I'll take the time now to go read the docs on how I need to set up my environment so that things are formatted properly.

bencaradocdavies:
@jdeolive a Maven build will now reformat by default (fmt-maven-plugin magic added by @aaime). Travis CI runs fmt-maven-plugin in check mode; if you have reformatted by building with Maven, Travis CI fmt-maven-plugin should be happy.

jdeolive:
Thanks @bencaradocdavies! I eventually found the info in the geotools docs, I didn't find anything in the geoserver docs though... might be worth adding there too.

PR #3029, opened 2018-08-02
GEOS-8866 jdbcconfig: handle advertised field properly

NielsCharlier:
https://osgeo-org.atlassian.net/browse/GEOS-8866

### DIFF
--- a/src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
+++ b/src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
@@ -818,6 +818,9 @@ public class ConfigDatabase {
         final boolean updateResouceLayersName =
                 info instanceof ResourceInfo
                         && modificationProxy.getPropertyNames().contains("name");
+        final boolean updateResouceLayersAdvertised =
+                info instanceof ResourceInfo
+                        && modificationProxy.getPropertyNames().contains("advertised");
         final boolean updateResourceLayersKeywords =
                 CollectionUtils.exists(
                         modificationProxy.getPropertyNames(),
@@ -854,6 +857,9 @@ public class ConfigDatabase {
             if (updateResouceLayersName) {
                 updateResourceLayerName((ResourceInfo) info);
             }
+            if (updateResouceLayersAdvertised) {
+                updateResourceLayerAdvertised((ResourceInfo) info);
+            }
             if (updateResourceLayersKeywords) {
                 updateResourceLayerKeywords((ResourceInfo) info);
             }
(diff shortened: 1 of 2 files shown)


NielsCharlier:
Could you check this one @sikeoka

sikeoka:
The pull request looks good but this is the third time that something like this has had to be fixed which makes me wonder how many other instances exist but just haven't been found yet.

aaime:
Wondering if this could be covered by tests, like, have them fail if a mapping is missing? Or is it just a performance issue?
In any case, as long as the module is unsupported, tests won't be run (it's the "punishment" for unsupported modules, or put another way, the "incentive" to make them supported).

NielsCharlier:
@aaime This is not a performance issue, I did write a test that would fail without the changes

@sikeoka you make a good point. I had a look and 'enabled' should be handled the same. I will add it to this PR.

 On the other hand 'keywords' is also handled this way but there is no derived 'keywords' property in LayerInfo (anymore). Do you have any idea @aaime if perhaps this used to be there but was removed perhaps?

aaime:
@NielsCharlier keywords are in ResourceInfo... I don't remember them being moved from Layer to Resource.

src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
```diff
@@ -872,28 +884,14 @@ public boolean evaluate(Object input) {
...
-        final Object newValue = info.getKeywords();
+    private <T> void updateResourceLayerProperty(ResourceInfo info, String propertyPath) {
+        final Object newValue = info.isAdvertised();
         Filter filter = Predicates.equal("resource.id", info.getId());
         List<LayerInfo> resourceLayers;
         resourceLayers = this.queryAsList(LayerInfo.class, filter, null, null, null);
         for (LayerInfo layer : resourceLayers) {
             Set<PropertyType> propertyTypes =
-                    dbMappings.getPropertyTypes(LayerInfo.class, "resource.keywords.value");
+                    dbMappings.getPropertyTypes(LayerInfo.class, "advertised");
```
sikeoka: Should this use propertyPath instead of "advertised"?

src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
```diff
@@ -872,28 +884,14 @@ public boolean evaluate(Object input) {
...
-            Property changedProperty = new Property(propertyType, newValue);
-            Integer layerOid = findObjectId(layer);
-            updateQueryableProperties(layer, layerOid, ImmutableSet.of(changedProperty));
-        }
-    }
-
-    private <T> void updateResourceLayerKeywords(ResourceInfo info) {
-        final Object newValue = info.getKeywords();
+    private <T> void updateResourceLayerProperty(ResourceInfo info, String propertyPath) {
+        final Object newValue = info.isAdvertised();
```
sikeoka: This uses the isAdvertised() value for all of the properties.  Maybe add newValue as a parameter to this method and pass in the correct value?

src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
```diff
@@ -818,6 +818,12 @@ public void remove(Info info) {
         final boolean updateResouceLayersName =
                 info instanceof ResourceInfo
                         && modificationProxy.getPropertyNames().contains("name");
+        final boolean updateResouceLayersAdvertised =
+                info instanceof ResourceInfo
+                        && modificationProxy.getPropertyNames().contains("advertised");
+        final boolean updateResouceLayersEnabled =
+                info instanceof ResourceInfo
+                        && modificationProxy.getPropertyNames().contains("enabled");
```
sikeoka: Not a big deal but "Resource" is misspelled.

NielsCharlier:
@sikeoka Sorry, that was a bit of a mess. Fixed now.

src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
```diff
@@ -852,10 +858,20 @@ public boolean evaluate(Object input) {
...
+                        ((ResourceInfo) info).isAdvertised());
+            }
+            if (updateResourceLayersEnabled) {
+                updateResourceLayerProperty((ResourceInfo) info, "enabled", 
+                        ((ResourceInfo) info).isEnabled());
             }
             if (updateResourceLayersKeywords) {
-                updateResourceLayerKeywords((ResourceInfo) info);
+                updateResourceLayerProperty((ResourceInfo) info, "resource.keywords.value", 
+                        ((ResourceInfo) info).getName());
```
sikeoka: This should be getKeywords().

NielsCharlier:
oh boy, okay, fixed, thanks for your patience

sikeoka:
It looks good to me.

PR #3030, opened 2018-08-04
GEOS-8870 error trying to rename workspace through rest api

lfelipevc:
[https://osgeo-org.atlassian.net/projects/GEOS/issues/GEOS-8870](https://osgeo-org.atlassian.net/projects/GEOS/issues/GEOS-8870)

Fixed bug  error trying to rename workspace through rest api.

For tests:

curl -v --user admin:geoserver -X put -H "Content-Type: application/json" -d '{"workspace": {"name":"toppNew"}}' http://localhost:8080/geoserver/rest/workspaces/topp 

Before
< HTTP/1.1 403 Can't change name of workspace

After:
< HTTP/1.1 200 OK

### DIFF
--- a/src/restconfig/src/main/java/org/geoserver/rest/catalog/WorkspaceController.java
+++ b/src/restconfig/src/main/java/org/geoserver/rest/catalog/WorkspaceController.java
@@ -175,7 +175,7 @@ public class WorkspaceController extends AbstractCatalogController {
             }
 
             String infoName = workspace.getName();
-            if (infoName != null && workspaceName.equals(infoName)) {
+            if (infoName != null && infoName.isEmpty()) {
                 throw new RestException("Can't change name of workspace", HttpStatus.FORBIDDEN);
             }
 
--- a/src/restconfig/src/main/java/org/geoserver/rest/catalog/WorkspaceController.java
+++ b/src/restconfig/src/main/java/org/geoserver/rest/catalog/WorkspaceController.java
@@ -176,7 +176,7 @@ public class WorkspaceController extends AbstractCatalogController {
 
             String infoName = workspace.getName();
             if (infoName != null && infoName.isEmpty()) {
-                throw new RestException("Can't change name of workspace", HttpStatus.FORBIDDEN);
+                throw new RestException("The workspace name cannot be empty", HttpStatus.FORBIDDEN);
             }
 
             new CatalogBuilder(catalog).updateWorkspace(wks, workspace);
--- a/src/restconfig/src/main/java/org/geoserver/rest/catalog/WorkspaceController.java
+++ b/src/restconfig/src/main/java/org/geoserver/rest/catalog/WorkspaceController.java
@@ -175,7 +175,7 @@ public class WorkspaceController extends AbstractCatalogController {
             }
 
             String infoName = workspace.getName();
-            if (infoName != null && !workspaceName.equals(infoName)) {
+            if (infoName != null && workspaceName.equals(infoName)) {
                 throw new RestException("Can't change name of workspace", HttpStatus.FORBIDDEN);
             }
 
--- a/src/restconfig/src/test/java/org/geoserver/rest/catalog/WorkspaceTest.java
+++ b/src/restconfig/src/test/java/org/geoserver/rest/catalog/WorkspaceTest.java
@@ -304,7 +304,7 @@ public class WorkspaceTest extends CatalogRESTTestSupport {
 
     @Test
     public void testPutNameChangeForbidden() throws Exception {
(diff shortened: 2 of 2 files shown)


aaime:
The patch is causing many test failures:

```
Results :

Failed tests:   testPutNameChangeForbidden(org.geoserver.rest.catalog.WorkspaceTest): expected:<403> but was:<200>

Tests in error: 
  testGetAsHTML(org.geoserver.rest.catalog.WorkspaceTest): Namespace with URI 'http://geoserver.org' already exists.
  testGetAsJSON(org.geoserver.rest.catalog.WorkspaceTest): Namespace with URI 'http://geoserver.org' already exists.
  testPostToResource(org.geoserver.rest.catalog.WorkspaceTest): Namespace with URI 'http://geoserver.org' already exists.
  testDelete(org.geoserver.rest.catalog.WorkspaceTest): Namespace with URI 'http://geoserver.org' already exists.
  testRoundTripJSONSerialization(org.geoserver.rest.catalog.WorkspaceTest): Namespace with URI 'http://geoserver.org' already exists.
  testIsolatedWorkspaceHandling(org.geoserver.rest.catalog.WorkspaceTest): Namespace with URI 'http://geoserver.org' already exists.
  testDeleteRecursive(org.geoserver.rest.catalog.WorkspaceTest): Namespace with URI 'http://geoserver.org' already exists.
  testPutDefaultWorkspace(org.geoserver.rest.catalog.WorkspaceTest): Namespace with URI 'http://geoserver.org' already exists.
```

Please remember to always run a full build after changes, it will verify the changes do not introduce obvious regression, and will reformat the code to match our conversions (non complying code will also make the build fail on Travis)

lfelipevc:
The testPutNameChangeForbidden asserts that the reponse will be 403, if try to put a different workspace name, theres a business rule about this behavior or this is a wrong test ?

src/restconfig/src/main/java/org/geoserver/rest/catalog/WorkspaceController.java
```diff
@@ -175,7 +175,7 @@ public void workspacePut(
             }
 
             String infoName = workspace.getName();
-            if (infoName != null && !workspaceName.equals(infoName)) {
+            if (infoName != null && infoName.isEmpty()) {
                 throw new RestException("Can't change name of workspace", HttpStatus.FORBIDDEN);
```
aaime: The error message should also be updated to something like "the name cannot be empty" or something like that. 
Honestly I don't know why the check was here, as the UI allows for workspace renames.

aaime: Checked the history, looks like this behavior was there since the very beginning of the REST API, 9 years ago:
https://github.com/geoserver/geoserver-history/blame/master/src/restconfig/src/main/java/org/geoserver/catalog/rest/WorkspaceResource.java

aaime:
Overall looks good, I see no valid reason to prevent workspace renames as the UI allows for it too. Did you check that GeoServer still behaves correctly after a rename, like, try to add a store, get a layer from the renamed workspace and the like? All should be handled by usage of ids instead of names, and internal listeners, but just in case, some interactive testing would not hurt.

aaime:
@lfelipevc did you see my message of 21 days ago starting with "overall looks good"?

lfelipevc:
@aaimeI had tested renaming some workspace that has a PostGis,SHP, and WMS datastores, has some layers, a group layer and style. After that, I had included new datastores (postgis and shp), layers, changed some style attributes and had updated a group layer. 
All changes was done sucessfully and i didn't found any missbehavior. 
The workspace global service url has changed as expected. I was able to rename again and everthing had keep working as expected.

lfelipevc:
@aaime  I noticed that renaming workspace doesn't update access rules associated with it (DataSecurity). It not only occours when using rest api to rename but also using web ui. At the data acess rules list you'll find rules with old workspace name. If you try to update these rules, the workspace name selected will be "*", at the combobox, and the new name will be at the list, so you can choose it. If you dont remember the old one, this behavior could be a problem.

Weird behavior was found when i openned a layer at QGIS by WFS(t) service and renamed a workspace associed with this layer. QGIS started to show error that couldn't download this layer features (as expected) but i was able to to write polygon changes. At the Geoserver log i was able to see that QGIS was still calling wfst using a GML request with old workspace name and no error was logged.

aaime:
@lfelipevc I'm merging this one, it looks like it's in a good state. You might want to make further PRs to backport the changes to 2.14.x and 2.13.x (not a requirement, up to you). 
As for the renames not affecting the data security, yep, a catalog listener should likely be added that updates them. Could you open a Jira ticket for that (if not there already)? A PR to fix that too would be welcomed :-)

PR #3032, opened 2018-08-08
[GEOS-8879] Adding a GetDomainValues to get paged unique values of a given domain, in WMTS multidimensional community module

aaime:


src/community/wmts-multi-dimensional/src/main/java/org/geoserver/gwc/wmts/MultiDimensionalExtension.java
```diff
@@ -65,15 +69,25 @@
     public static final Set<String> ALL_DOMAINS = Collections.emptySet();
     /** Configuration key for the maximum allowed value of expand_limit */
     public static final String EXPAND_LIMIT_MAX_KEY = "MD_DOMAIN_EXPAND_LIMIT_MAX";
+
+    public static final String LIMIT_MAX_KEY = "MD_GET_DOMAIN_VALUES_LIMIT_MAX";
```
nmco: Since there is already other ``LIMIT`` constants I would probably put something more meaningful than ``LIMIT* ``, e.g. ``DOMAIN_VALUES_LIMIT_MAX_KEY``.

src/community/wmts-multi-dimensional/src/main/java/org/geoserver/gwc/wmts/MultiDimensionalExtension.java
```diff
@@ -288,6 +417,47 @@ private int getExpandlimit(
...
+            if (value < limitMax) {
+                return value;
+            } else {
+                String message =
+                        "Client provided a limit value higher than the configured max, using the "
+                                + "internal value (provided/max): "
+                                + value
+                                + "/"
+                                + limitMax;
+                LOGGER.log(Level.INFO, message);
```
nmco: Will it make sense to log it as an warning ?

nmco:
Looks good to me, adds the documentation for the new operation and the necessary tests, :+1: .

src/community/wmts-multi-dimensional/src/main/java/org/geoserver/gwc/wmts/MultiDimensionalExtension.java
aaime: It would make it pop up in PRODUCTION Logs, which is not what we want, this is more of a debugging facility. Also, the other limit is handled the same way (INFO when client goes above)

src/community/wmts-multi-dimensional/src/main/java/org/geoserver/gwc/wmts/MultiDimensionalExtension.java
aaime: Fixing

PR #3034, opened 2018-08-09
GEOS-8786 Fix Wicket exception - Test Connection button in LDAP Authentication Provider

lfelipevc:
[GEOS-8786](https://osgeo-org.atlassian.net/browse/GEOS-8786) Change feedback panel name

### DIFF
--- a/src/web/security/ldap/src/main/java/org/geoserver/web/security/ldap/LDAPAuthProviderPanel.java
+++ b/src/web/security/ldap/src/main/java/org/geoserver/web/security/ldap/LDAPAuthProviderPanel.java
@@ -194,7 +194,7 @@ public class LDAPAuthProviderPanel extends AuthenticationProviderPanel<LDAPSecur
                                     (LDAPSecurityServiceConfig) getForm().getModelObject();
                             doTest(ldapConfig, username, password);
 
-                            target.add(getPage().get("feedback"));
+                            target.add(getPage().get("topFeedback"));
                         }
 
                         void doTest(
--- a/src/web/security/ldap/src/test/java/org/geoserver/web/security/ldap/LDAPAuthProviderPanelTest$LDAPFormTestPage.html
+++ b/src/web/security/ldap/src/test/java/org/geoserver/web/security/ldap/LDAPAuthProviderPanelTest$LDAPFormTestPage.html
@@ -1,6 +1,6 @@
 <html xmlns:wicket="http://wicket.apache.org/">
   <body>
-    <div wicket:id="feedback"></div>
+    <div wicket:id="topFeedback"></div>
     <form wicket:id="form">
       <div wicket:id="panel"></div>
     </form>
--- a/src/web/security/ldap/src/test/java/org/geoserver/web/security/ldap/LDAPAuthProviderPanelTest.java
+++ b/src/web/security/ldap/src/test/java/org/geoserver/web/security/ldap/LDAPAuthProviderPanelTest.java
@@ -174,7 +174,7 @@ public class LDAPAuthProviderPanelTest extends AbstractSecurityWicketTestSupport
 
         @Override
         protected void onBeforeRender() {
-            feedbackPanel = new FeedbackPanel("feedback");
+            feedbackPanel = new FeedbackPanel("topFeedback");
             feedbackPanel.setOutputMarkupId(true);
             addOrReplace(feedbackPanel);
             super.onBeforeRender();


rduivenvoorde:
Hi, seems like a trivial ui fix? Haven't tested it though, but is there a reason to not merge this in current RC?

rduivenvoorde:
fyi: https://github.com/geoserver/geoserver/pull/3035 seems like the same sort of fix

@aaime (/me not into wicket, but looking at your comment as an outsider): is this an answer to @mbarto question? Or is it a comment on the proposed code fix of @lfelipevc ?

rduivenvoorde:
@aaime @mbarto I'm looking at (/debugging) this pr now (based on http://docs.geoserver.org/latest/en/user/security/tutorials/ldap/index.html). 
For what I see @lfelipevc adds a 'feedback' div, and puts the msg in that, and it works :-)

I understand that there is a top- and bottom feedback now, and you should be able to bring the message to that via 'info()'? But that does not work (nor getParent.info or getForm.info ).
Looks like that this 'test' panel is a subpanel from the larger form in this page?
Can you maybe point at some code which does about the same. I can try to look into that then, and see if I can make this work...
Thanks!

rduivenvoorde:
Ok, I looked into it a little deeper, these two solutions work:

```
target.add(getPage().get("topFeedback"));
target.add(getPage().get("bottomFeedback"));
```
instead of
`target.add(getPage().get("feedback"));`

but probably best is:

`((GeoServerBasePage)this.getPage()).addFeedbackPanels(target);`

as there are not hardcoded named components then. Not sure about the cast of getPage() though.

if people are ok with the last option, I can create a PR for this.

mbarto:
@rduivenvoorde your proposed solution looks good to me

lfelipevc:
@rduivenvoorde @mbarto. If we look to the JDBCAuthProviderPanel, thats a similar page, it has own feedback panel (JDBCAuthProviderPanel.html at line 9) and the JDBCAuthProviderPanel.java use it to show messages. 

The bug is the same as #3035  at AuthenticationKeyFilterPanel.java

rduivenvoorde:
@lfelipevc ah, clear. You mean, they also make a custom feedback component too at 

https://github.com/geoserver/geoserver/blob/master/src/web/security/jdbc/src/main/java/org/geoserver/security/web/jdbc/JDBCAuthProviderPanel.java#L77

Ok, I let this onto the geoserver dev's, I just wanted to fix my issue, as I'm doing ldap stuff now :-)
But in general I think it is better to reuse the generic 'feedback'-components yes?
But I'm ok with whatever solution is preferred by devs (my PR has failing tests anyway :( )

rduivenvoorde:
@mbarto @bradh my pr #3095 has failing tests, which I'm not able to fix, so I dropped the pr:

```
Caused by: java.lang.ClassCastException: org.geoserver.web.security.ldap.LDAPAuthProviderPanelTest$LDAPFormTestPage cannot be cast to org.geoserver.security.web.SecurityNamedServiceEditPage
	at org.geoserver.web.security.ldap.LDAPAuthProviderPanel$TestLDAPConnectionPanel$1.onSubmit(LDAPAuthProviderPanel.java:198)
```

The cast I did to either `SecurityNamedServiceEditPage` or `GeoServerBasePage` will fail because the 'page' used in the tests is a `LDAPFormTestPage`, so the panel will never be able to use the cast to `geoserverbasepage` (to use the specific `addFeedbackPanels` during tests) ...
OR we should move `addFeedbackPanels` to an interface and implement a stub in the `LDAPFormTestPage` too.
But quickest fix is just to use @lfelipevc fix I think, as even
```
target.add(getPage().get("topFeedback"));
target.add(getPage().get("bottomFeedback"));
```
will not work during tests (UNLESS we create those in the test panel too...)

aaime:
I'm back from my holiday and I finally have more than just a smartphone to look at this (proper computer and IDE). So yeah, the test is using a custom mock test page to run tests that has nothing to do with GeoServerBasePage. A few options:

- Rewrite the test to use the actual page containing the panel instead of a test one (not sure how hard, I'm unfamiliar with this part of GeoServer)
- Add both a "topFeedback" and "bottomFeedback" in the test page container, to mimick what's contained in the actual container page used at runtime (this one looks easier)
- Have the test use GeoServerBasePage instead of the custom mock page, or have the custom mock page extend it (if not too hard to do, that's probably best approach and will allow to cast to GeoServerBasePage and use the GeoServerBasePage.addFeedbackPanels(targer) to do the job in a type safe way)

rduivenvoorde:
@aaime thanks for the feedback! But I fail to make your first and third point work. Too much (Wicket?) magic going on for me.

What @lfelipevc is doing is actually your second point (except that @lfelipevc does not add both topFeedback and bottomFeedback, but only use topFeedback in the test page container).

@mbarto the reason this works and 'feedback' not, is that if you use 'feedback', then the test page (which has a feedback-widget) will be OK and all tests will succeed. But you will have https://osgeo-org.atlassian.net/browse/GEOS-8786 in real world, because 'feedback' as widget is removed in favour of the topFeedback and bottomFeedback widgets from the real page.

My earlier attempts casting to GeoServerBasePage or SecurityNamedServiceEditPage:
`((GeoServerBasePage) this.getPage()).addFeedbackPanels(target);`
will work in real world, but not in Test case because there the case will fail because LDAPAuthProviderPanelTest$LDAPFormTestPage cannot be cast to GeoServerBasePage/SecurityNamedServiceEditPage

I agree that aaime's third bullet is the best solution, but that needs proper knowledge of setting up the Wicket test pages.

PR #3035, opened 2018-08-09
GEOS-8868 Exception Synchronize user/group Authkey

lfelipevc:
[GEOS-8868](https://osgeo-org.atlassian.net/browse/GEOS-8868) Change feedback panel name


petersmythe:
Thanks for fixing this!

aaime:
Works for me, merged. Want to make PR for backports too?

PR #3044, opened 2018-08-13
[GEOS-8888] Cannot add more than 25 layers in layer group

aaime:


aaime:
Merging, Travis failure is towards the end and looks like a repository issue

PR #3046, opened 2018-08-13
[GEOS-8863] [Community GeoFence] GeoFence Embedded REST endpoints cause NPE when Limits have been set on Layers

afabiani:


NielsCharlier:
It looks all pretty good to me, but there seems to be some general re-factoring too apart from the bug fix,or is it somehow related?

afabiani:
@NielsCharlier except for some formatting stuff, which now is automatic on GeoServer during the rebuild, I have refactored and improved a bit the test cases and the test support base class. Tests include also check on limits. So the response is yes, it is somehow related.

PR #3051, opened 2018-08-14
[GEOS-8891] Make it possible to configure coordinates measures encoding per layer

nmco:
Associated issue:
https://osgeo-org.atlassian.net/browse/GEOS-8891

This PR allow the configuration, per feature type, of coordinates measures values encoding. Is up to each output format tot ake advantage of that information (GML2, GML3, GoJSON, etc ...).

### DIFF
--- a/doc/en/user/source/data/webadmin/layers.rst
+++ b/doc/en/user/source/data/webadmin/layers.rst
@@ -262,6 +262,8 @@ WFS Settings
 
      The list will be used only for the capabilities document generation, but will not be used to limit the actual target SRS usage in GetFeature requests.
 
+* **Encode coordinates measures**-By default coordinates measures will not be encoded in WFS output formats, checking this check-box  will make coordinates measures to be encoded in WFS output formats that support them.  
+
 WCS Settings
 ^^^^^^^^^^^^
 
--- a/src/main/src/main/java/org/geoserver/catalog/FeatureTypeInfo.java
+++ b/src/main/src/main/java/org/geoserver/catalog/FeatureTypeInfo.java
@@ -153,4 +153,23 @@ public interface FeatureTypeInfo extends ResourceInfo {
     boolean isCircularArcPresent();
 
     void setCircularArcPresent(boolean arcsPresent);
+
+    /**
+     * Controls if coordinates measures should be included in WFS outputs.
+     *
+     * @return TRUE if measures should be encoded, otherwise FALSE
+     */
+    default boolean getEncodeMeasures() {
+        // by default coordinates measures are not encoded
+        return false;
+    }
+
+    /**
+     * Sets if coordinates measures should be included in WFS outputs.
+     *
+     * @param encodeMeasures TRUE if measures should be encoded, otherwise FALSE
+     */
+    default void setEncodeMeasures(boolean encodeMeasures) {
+        // nothing to do
+    }
 }
--- a/src/main/src/main/java/org/geoserver/catalog/impl/FeatureTypeInfoImpl.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/FeatureTypeInfoImpl.java
@@ -42,6 +42,9 @@ public class FeatureTypeInfoImpl extends ResourceInfoImpl implements FeatureType
     boolean skipNumberMatched = false;
     boolean circularArcPresent;
 
+    // we don't use the primitive because we need to detect the situation where no value was set
+    Boolean encodeMeasures;
+
     public boolean isCircularArcPresent() {
         return circularArcPresent;
     }
@@ -238,5 +241,16 @@ public class FeatureTypeInfoImpl extends ResourceInfoImpl implements FeatureType
     public void setCqlFilter(String cqlFilter) {
         this.cqlFilter = cqlFilter;
         this.filter = null;
+    }
+
+    @Override
+    public boolean getEncodeMeasures() {
+        // by default encoding of coordinates measures is not activated
+        return encodeMeasures == null ? false : encodeMeasures;
+    }
+
+    @Override
+    public void setEncodeMeasures(boolean encodeMeasures) {
+        this.encodeMeasures = encodeMeasures;
     }
 }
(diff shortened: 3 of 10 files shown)


src/main/src/main/java/org/geoserver/catalog/FeatureTypeInfo.java
```diff
@@ -153,4 +153,23 @@
     boolean isCircularArcPresent();
 
     void setCircularArcPresent(boolean arcsPresent);
+
+    /**
+     * Controls if coordinates measures should be included in WFS outputs.
+     *
+     * @return TRUE if measures should be encoded, otherwise FALSE
+     */
+    default boolean getEncodeCoordinatesMeasures() {
```
aaime: Should be "isEncodeCoordinateMeasures" or even better "isCoordinateMeasureEncoded"

src/main/src/main/java/org/geoserver/catalog/impl/FeatureTypeInfoImpl.java
```diff
@@ -239,4 +242,15 @@ public void setCqlFilter(String cqlFilter) {
         this.cqlFilter = cqlFilter;
         this.filter = null;
     }
+
+    @Override
+    public boolean getEncodeCoordinatesMeasures() {
+        // by default encoding of coordinates measures is activated
+        return encodeCoordinatesMeasures == null ? true : encodeCoordinatesMeasures;
```
aaime: Not sure if it should default to true. In GML that will break standard clients. In GeoJSON it might confuse some.

src/wfs/src/main/java/org/geoserver/wfs/WFSGetFeatureOutputFormat.java
```diff
@@ -217,6 +217,34 @@ protected int getNumDecimals(List featureCollections, GeoServer geoServer, Catal
...
+    /**
+     * Helper method that checks if coordinates measured values should be encoded for the provided
+     * feature collections.
+     *
+     * @param featureCollections features collections
+     * @param catalog GeoServer catalog
+     * @return TRUE if coordinates measures should be encoded, otherwise FALSE
+     */
+    protected boolean encodeCoordinatesMeasures(List featureCollections, Catalog catalog) {
+        boolean encodeCoordinatesMeasures = true;
```
aaime: This bit is also affected by the defaults, in case we decided to flip them after review

src/main/src/main/java/org/geoserver/catalog/FeatureTypeInfo.java
nmco: I agree :stuck_out_tongue:, done.

src/main/src/main/java/org/geoserver/catalog/impl/FeatureTypeInfoImpl.java
nmco: Making coordinates measures not encoded by default, done.

nmco:
GeoJSONBuilder is not covered for now. All changes request applied. 

Thanks for the detailed review @aaime (like always), much appreciated.

src/web/wfs/src/test/java/org/geoserver/wfs/web/WFSLayerConfigTest.java
```diff
@@ -0,0 +1,42 @@
+/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2013 OpenPlans
```
jodygarnett: new file?

src/web/wfs/src/main/resources/GeoServerApplication.properties
```diff
@@ -47,6 +47,9 @@ WFSLayerConfig.otherSRS.message=A comma separated list of EPSG codes, e.g. 4326,
    response. The list can be left empty to have no extra SRS declared for this specific type, in \
    override to a list of values specified in the WFS service configuration.
 
+WFSLayerConfig.coordinatesEncodingTitle=Coordinates Encoding
+WFSLayerConfig.coordinatesEncodingHelp.message=lalala
```
jodygarnett: la la la?

jodygarnett:
Is anything needed from the JTS library for this PR? Or is it all handled in geotools ....

(This change does effected the UI so some documentation change is needed)

nmco:
Documentation updated. This doesn't depends on any JTS or GeoTools changes.

src/web/wfs/src/main/resources/GeoServerApplication.properties
aaime: LOL LOL LOL!

src/web/wfs/src/test/java/org/geoserver/wfs/web/WFSLayerConfigTest.java
aaime: So @jodygarnett the GeoTools part is merged but won't do anything until this one is also merged. Should we merge this one as is and fix later?

jodygarnett: The only reason I did not approve was the musical encoding help instructions

doc/en/user/source/data/webadmin/layers.rst
```diff
@@ -262,6 +262,8 @@ WFS Settings
 
      The list will be used only for the capabilities document generation, but will not be used to limit the actual target SRS usage in GetFeature requests.
 
+* **Encode coordinates measures**-By default coordinates measures will not be encoded in WFS output formats, checking this check-box  will make coordinates measures to be encoded in WFS output formats that support them.
```
bradh: * **Encode coordinates measures** - Checking this setting will cause coordinate measures ("M") to be encoded in WFS output formats that support measures. The default (not checked) is to not encode coordinates measures

aaime:
Manually merged on master after local manual testing to include GT changes and subsequent build fixes on the GeoServer side. I've also applied Brad's feedback while at it.

PR #3052, opened 2018-08-14
[JIRA GEOS-8864] [Community GeoFence] GeoFence Embedded REST endpoints paths should be consistent with std GeoServer behavior

afabiani:


src/community/geofence-server/src/test/java/org/geoserver/geofence/rest/RulesRestControllerTest.java
```diff
@@ -57,50 +63,7 @@ public void testInsertUpdateDelete() {
         rule.setLayer("layer");
         rule.setAccess("ALLOW");
 
-        JaxbRuleList rules =
```
aaime: What is going on here?

afabiani: Yes, I have places the cleanup code into the utility method of the same class

protected long prepareGeoFenceTestRules(JaxbRule rule)

That one checking and cleaning up all the previously created rules.

PR #3054, opened 2018-08-14
ImageMosaic GetCapabilities performance

ivanya82:
We are using GeoServer to display a large number of GeoTIFFs using ImageMosaic data stores. The data has several dimensions (time and custom). We are adding new data to these stores periodically and the goal is to be able to display years of data (millions of granules). We have encountered performance problems with the WMS GetCapabilities operation. It is most noticeable when using Oracle database (or PostGIS with parameter WrapStore=true). We implemented changes which improved the performance for our use case significantly. We would like to publish these changes and hopefully merge them into the official repo.

Details:
our performance problems are caused by the queries for values and defaults for dimensions. The currently used queries vary depending on the selected presentation for the dimension:

1. query for values:
  "List":
    * PostGIS (with WrapStore=false): `SELECT distinct("<dim_name>") FROM "<table_name>"`
    * Oracle: `SELECT FID, <DIM_NAME> FROM <table_name>`
    * PostGIS (with WrapStore=true): `SELECT "fid","<dim_name>" FROM "<table_name>"`
    First version is clearly the fastest, because it loads only distinct values.

"Interval and resolution":
  The queries are the same as for "List", but in this case we only really need the min/max values, which could be calculated much faster in the database.

2. query for defaults (when using smallest/biggest domain value):
  * PostGIS (with WrapStore=false): `SELECT <min | max> ("<dim_name>") FROM "<table_name>"`
  * Oracle: `SELECT FID, <DIM_NAME> FROM <table_name>`
  * PostGIS (with WrapStore=true): `SELECT "fid","<dim_name>" FROM "<table_name>"`
  Again the first PostGIS variant is clearly the fastest. Others load all the values to the application where the min/max values are calculated.

Our proposed solution creates optimized queries depending on which presentation and default setting is used for the dimension. The needed additional information from the settings is sent to the lower level functions using new RenderingHints. The goal of the optimized queries is to load only the needed values from the database (usually what the current PostGIS query does). Unit tests were also added.

We would like to get some feedback on this.

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/capabilities/DimensionHelper.java
+++ b/src/wms/src/main/java/org/geoserver/wms/capabilities/DimensionHelper.java
@@ -264,7 +264,26 @@ abstract class DimensionHelper {
     private void handleElevationDimensionRaster(
             CoverageInfo cvInfo, DimensionInfo elevInfo, ReaderDimensionsAccessor dimensions)
             throws IOException {
-        TreeSet<Object> elevations = dimensions.getElevationDomain();
+        TreeSet<Object> elevations = null;
+        try {
+            if (elevInfo.getPresentation() != DimensionPresentation.LIST) {
+                Double minValue = dimensions.getMinElevation();
+                if (minValue != null) {
+                    elevations = new TreeSet<>();
+                    elevations.add(minValue);
+                    elevations.add(dimensions.getMaxElevation());
+                }
+            }
+            if (elevations == null) {
+                throw new Exception(
+                        "The \"List\" presentation of the elevation dimension has been selected");
+            }
+        } catch (Exception ex) {
+            if (LOGGER.isLoggable(Level.FINE)) {
+                LOGGER.log(Level.FINE, "Dimension has not been extracted. The reason: ", ex);
+            }
+            elevations = dimensions.getElevationDomain();
+        }
         String elevationMetadata = getZDomainRepresentation(elevInfo, elevations);
         String defaultValue = getDefaultValueRepresentation(cvInfo, ResourceInfo.ELEVATION, "0");
         writeElevationDimension(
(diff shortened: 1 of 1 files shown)


ivanya82:
Thanks to the advises [Andrea](https://github.com/aaime) during discussing on [Pull requests](https://github.com/geotools/geotools/pull/2009) and on [the list](https://sourceforge.net/p/geotools/mailman/geotools-devel/thread/92376c9e-0bf0-39d7-18c8-a25690a97cc9%40microstep-mis.com/#msg36388750) two main problems have been detected and their solutions have been found. The first problem which refers to the optimization broken by `TransformFeatureCollection` was solved by adding the "`accepts`" method to the `TransformFeatureCollection` class (GeoTools). As a result, the performance improvement has been obtained only to get the minimum/maximum defaults of the dimensions and its distinct list by using the visitors.

The second problem refers to extracting the whole domain dimension values even in the case of "Interval and resolution" presentation for the dimension is selected was solved by adding the changes to the `handleTimeDimensionRaster` and `handleElevationDimensionRaster` methods of the `DimensionHelper.java`.

aaime:
The last changes look good, I would just make it fall back on dimension extraction (old code) if the min/max time happen to be "null", just in case some reader in the wild has been coded by just observing what was asked of it and does not provide the min/max metadata.

Testing wise, this is an optimization, not a new functionality, and there are tests already covering representing "interval and resolution" dimensions over rasters here:
https://github.com/geoserver/geoserver/blob/6e9e25c0c7cdda9ada9f33f8255130d3afc76801/src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/DimensionsRasterCapabilitiesTest.java#L98-L97

So, add the fallback behavior that I discussed above, and then this pull request will be ready for merge.

Ah, the GeoServer contribution guidelines ask for you to sign a contribution agreement, see here:
I thought that for a single file change the CLA was not requested, but I don't see the wording here, I'll double check.

Gyuri16:
Regarding the CLA:
Ivan and I work for the same company (MicroStep-MIS). The company has signed a Corporate CLA about two weeks ago.

src/wms/src/main/java/org/geoserver/wms/capabilities/DimensionHelper.java
```diff
@@ -264,7 +264,22 @@ void handleRasterLayerDimensions(final LayerInfo layer) throws RuntimeException,
...
+            if (elevInfo.getPresentation() != DimensionPresentation.LIST) {
+                Double minValue = dimensions.getMinElevation();
+                if (minValue != null) {
+                    elevations = new TreeSet<>();
+                    elevations.add(minValue);
+                    elevations.add(dimensions.getMaxElevation());
+                }
+            }
+            if (elevations == null) {
+                throw new Exception();
```
aaime: The exception should provide some reason, as it goes up and becomes the OGC ServiceException response text

src/wms/src/main/java/org/geoserver/wms/capabilities/DimensionHelper.java
```diff
@@ -264,7 +264,22 @@ void handleRasterLayerDimensions(final LayerInfo layer) throws RuntimeException,
...
+            }
+        } catch (Exception e) {
+            elevations = dimensions.getElevationDomain();
```
aaime: Please log the exception here, maybe at FINE level

src/wms/src/main/java/org/geoserver/wms/capabilities/DimensionHelper.java
```diff
@@ -291,7 +306,22 @@ private String getDefaultValueRepresentation(
...
+                    temporalDomain = new TreeSet<>();
+                    temporalDomain.add(minValue);
+                    temporalDomain.add(dimension.getMaxTime());
+                }
+            }
+            if (temporalDomain == null) {
+                throw new Exception();
+            }
+        } catch (Exception e) {
+            temporalDomain = dimension.getTimeDomain();
```
aaime: Same here

ivanya82:
The last checks were not successful because "The Travis CI build could not complete due to an error". Log file contains message "The job exceeded the maximum time limit for jobs, and has been terminated" but there is another message "BUILD SUCCESS" before terminated error. Could the Travis CI test be repeated or is there another way to fix it? Thank you.

aaime:
I've restarted the Travis build.

aaime:
Merging on master, unsure if 2.14.x is already open for backports, will delay that for a few days.

PR #3055, opened 2018-08-14
[JIRA GEOS-8867] Extra login forms show up in the UI only if the filter chain name matches the LoginFormInfo name

afabiani:


aaime:
I like the direction this is going.
I would suggest to use Class rather than string, this way Spring can validate the value actually matches an existing class during context startup (so, less error prone).
Ah, the build failure is unrelated.

afabiani:
@aaime I guess it should be good now, can you quick check?

aaime:
Looking good!

PR #3060, opened 2018-08-17
[JIRA GEOS-8851] importer uploads cleanup task deletes raster files for existing installs

afabiani:


aaime:
@afabiani I don't understand the change, could you please add some descriptive words about how it's supposed to work?

afabiani:
Sure. This is well described into the https://osgeo-org.atlassian.net/browse/GEOS-8851

This PR toggles the current behavior. Instead of removing anything that does not have a .locking flag into the folder, it will remove "only" the ones having a .clean-me flag instead. Thus resolver two issues:

1. It is backward compatible and no one risks that old uplaoded files are evicted by the new Importer behavior.

2. If for whatever reason an extension of the Importer does not correctly manage the .locking mechanism, you con't risk to lose your data. In the worst case you will have garbage into the upload folder.

afabiani:
@aaime is this fulfill your doubts?

aaime:
@afabiani merged. Backport as you see fit.

PR #3063, opened 2018-08-21
[JIRA GEOS-8899] [Community Security Plugins] Include Keycloak security plugins artifacts to community dependencies and release

afabiani:
closes https://osgeo-org.atlassian.net/browse/GEOS-8899

### DIFF
--- a/src/community/release/ext-geofence-server.xml
+++ b/src/community/release/ext-geofence-server.xml
@@ -21,7 +21,7 @@
         <include>*geofence-services*</include>
         <include>*geofence-persistence*</include>
         <include>hatbox*</include>
-        <include>*hibernate*</include>
+        <include>hibernate*</include>
         <include>jackson*</include>
         <include>javassist*</include>
         <include>jta*</include>
--- a/src/community/security/keycloak/src/main/java/org/geoserver/security/keycloak/AuthResults.java
+++ b/src/community/security/keycloak/src/main/java/org/geoserver/security/keycloak/AuthResults.java
@@ -49,16 +49,24 @@ class AuthResults implements AuthenticationEntryPoint {
      * @param authentication valid credentials
      */
     public AuthResults(Authentication authentication) {
-        assert authentication.getDetails() instanceof SimpleKeycloakAccount;
-        final SimpleKeycloakAccount details = (SimpleKeycloakAccount) authentication.getDetails();
+        Object username = null;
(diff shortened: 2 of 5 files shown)


src/community/release/ext-geofence-server.xml
```diff
@@ -21,7 +21,7 @@
         <include>*geofence-services*</include>
         <include>*geofence-persistence*</include>
         <include>hatbox*</include>
-        <include>*hibernate*</include>
+        <include>hibernate*</include>
```
aaime: Is this unrelated? Or keycloack brings in hibernate jars that should not be in GeoFence?

src/community/security/keycloak/src/main/java/org/geoserver/security/keycloak/AuthResults.java
```diff
@@ -49,16 +49,24 @@ public AuthResults(AuthChallenge challenge) {
      * @param authentication valid credentials
      */
     public AuthResults(Authentication authentication) {
-        assert authentication.getDetails() instanceof SimpleKeycloakAccount;
-        final SimpleKeycloakAccount details = (SimpleKeycloakAccount) authentication.getDetails();
+        Object username = null;
+        Object details = null;
+        if (authentication.getDetails() instanceof SimpleKeycloakAccount) {
```
aaime: Is there a case where we are using a keycloak plugin and not getting a SimpleKeycloakAccount?

src/community/release/ext-geofence-server.xml
afabiani: Exactly. Written like this the geofence will import <evrything>-hibernate which might cause issues to the keycloak plugin if not correctly configured. Typical example is the gs-monitoring-hibernate.jar which is wrongly included into the geofence-server package and therefore causing issues to security plugins too.

src/community/security/keycloak/src/main/java/org/geoserver/security/keycloak/AuthResults.java
afabiani: Yeps, when a user has logged in through the form method and the security context is already available. The plugin checks first if there's a principal already present into the security context with a matching username. In that case it brings the roles from the latter.

afabiani:
@aaime any other question on this? Can I merge it?

aaime:
No objections

PR #3065, opened 2018-08-27
[2.13.x][GEOS-8877] Added test scope to gs-sec-jdbc test dependency.

sikeoka:
2.13.x backport for https://github.com/geoserver/geoserver/pull/3031


sikeoka:
One more try since the build timed out.

aaime:
Hi @sikeoka since these are backports just check that the build does not break (and yes, it can be a bit finicky) and then merge as you see fit.

PR #3072, opened 2018-08-30
[GEOS-8799] NULL namespace prefixes used in AppSchema WFS GetFeature responses

fernandor777:
Fix and tests for [GEOS-8799] "WFS GetFeature response contains NULL namespaces for App-Schema complex features when multiple queries are used".

https://osgeo-org.atlassian.net/browse/GEOS-8799

### DIFF
--- a/src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/NamespacesWfsTest.java
+++ b/src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/NamespacesWfsTest.java
@@ -15,7 +15,10 @@ import java.util.Map;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import net.opengis.wfs20.StoredQueryDescriptionType;
+import org.apache.commons.io.IOUtils;
 import org.custommonkey.xmlunit.XpathEngine;
+import org.geoserver.catalog.impl.NamespaceInfoImpl;
+import org.geoserver.data.test.SystemTestData;
 import org.geoserver.wfs.StoredQuery;
 import org.geoserver.wfs.StoredQueryProvider;
 import org.geotools.wfs.v2_0.WFS;
@@ -103,6 +106,13 @@ public final class NamespacesWfsTest extends AbstractAppSchemaTestSupport {
                         "http://www.opengis.net/wfs/2.0",
                         "gml",
                         "http://www.opengis.net/gml/3.2");
+    }
+
+    @Override
+    protected void onSetUp(SystemTestData testData) throws Exception {
+        super.onSetUp(testData);
+        // inject a test namespace to check on responses
+        addTestNamespaceToCatalog();
     }
 
     /** * GetFeature tests ** */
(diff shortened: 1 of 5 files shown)


src/wfs/src/main/java/org/geoserver/wfs/xml/FeatureTypeSchemaBuilder.java
```diff
@@ -384,6 +366,39 @@ public final XSDSchema buildSchemaInternal(
         return schema;
     }
 
+    private void addAllNamespacesFromCatalog(XSDSchema schema) {
```
nmco: Nice, avoiding repating code is always good karma :)

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/NamespacesWfsTest.java
```diff
@@ -336,4 +346,80 @@ private void checkCount(
...
+                WFS20_XPATH_ENGINE,
+                document,
+                1,
+                "/wfs:FeatureCollection/wfs:member/wfs:FeatureCollection/wfs:member/"
+                        + "ms_gml32:Measurement_gml32");
+        // check prefixes:
+        String output = toString(document);
+        assertTrue(output.indexOf("null:Measurement_gml32") < 0);
+        assertTrue(output.indexOf("null:Station_gml32") < 0);
+        assertTrue(output.indexOf("ms_gml32:Measurement_gml32") > -1);
```
nmco: Why -1 ? It should always be be greater than zero, no ?

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/NamespacesWfsTest.java
```diff
@@ -336,4 +346,80 @@ private void checkCount(
...
+
+        checkCount(
+                WFS11_XPATH_ENGINE,
+                document,
+                1,
+                "/wfs:FeatureCollection/gml:featureMember/" + "ms_gml31:Measurement_gml31");
+        // check prefixes:
+        assertTrue(output.indexOf("null:Measurement_gml31") < 0);
+        assertTrue(output.indexOf("null:Station_gml31") < 0);
+        assertTrue(output.indexOf("ms_gml31:Measurement_gml31") > -1);
```
nmco: Same as above.

nmco:
Apart from some minor comments, looks good and its aligned with what has been done for the situation where we have only a single feature type queried.

src/wfs/src/main/java/org/geoserver/wfs/xml/FeatureTypeSchemaBuilder.java
```diff
@@ -320,6 +300,8 @@ public final XSDSchema buildSchemaInternal(
                                     schemaURIs.get(namespace),
                                     imports,
                                     includes);
+                            // add prefix
+                            schema.getQNamePrefixToNamespaceMap().put(prefix, namespace);
```
nmco: Is this still needed ? If this is a complex feature type all the name-spaces should already have been added in the ``if`` above right ?

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/NamespacesWfsTest.java
fernandor777: String.indexOf returns the index of the first occurrence of the specified substring, or -1 if there is no such occurrence.  So if is == -1 substring not found.

nmco: Ah! right my bad, I thought you were doing a XPATH count.

src/wfs/src/main/java/org/geoserver/wfs/xml/FeatureTypeSchemaBuilder.java
fernandor777: Yes, it sounds sensible, deleting it.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/NamespacesWfsTest.java
fernandor777: -1 == substring not found.

nmco:
The Windows build failure looks unrelated to this PR:
```
[INFO] gs-restconfig ...................................... FAILURE [02:22 min]
```

... and the same build failure happen on other PRs. I'm merging this.

PR #3076, opened 2018-09-01
Fixed tiled layers preview does not use Proxy Base Url - Geos 7653

mirandadam:
Added buidURL call when building the link to the tiled layers preview.
Made during the foss4g2018 code sprint with Ian Turton


ianturton:
superdeeded https://github.com/geoserver/geoserver/pull/3079

PR #3077, opened 2018-09-03
Cannot save new style after expand and collapse on the web

tixvieira:
On the geoserver New Style page,

Clicking the "expand" green button on the top right corner, and then collapse.
After doing it, cannot save the style because the layout of the page has broken.


bencaradocdavies:
@tixvieira this is a request to merge the development branch of GeoServer into the stable branch. Please submit your bug report as a [https://osgeo-org.atlassian.net/projects/GEOS/summary](GeoServer Jira issue).

aaime:
Please also add in the Jira issue details about your system, the browser used, screen resolution and the like. Locally I could not reproduce the problem, so it's likely related to some differences between my setup and yours.

PR #3082, opened 2018-09-04
Remove spring-asm dependency from jms-cluster

tbarsballe:
spring-asm was merged into spring-core with the release of Spring 3.2.0, and should no longer be included as a dependency.


tbarsballe:
The CI failure is the same as a failure on master, will restart after master is stable and a nightly has been run

PR #3088, opened 2018-09-05
[GEOS-8908] Wrong Coverage Parameter Type for OVERVIEW POLICY causing Coverage layer to be deleted from the Catalog upon reload of configuration or restart.

aaime:


aaime:
@dromagnoli I believe this needs backports to 2.13.x too (sent from phone, limited view of what is going on)

PR #3092, opened 2018-09-06
[GEOS-8918] [GEOS-8919] Relax existing constraints on complex feature layer names

fernandor777:
Associated issues:
- https://osgeo-org.atlassian.net/browse/GEOS-8919
- https://osgeo-org.atlassian.net/browse/GEOS-8918

This PR relax some existing constraints on the naming of complex features layers, namely:
- It should not be mandatory to use always use the full resource name when querying a complex feature type;
- It should be possible to name complex features layers with an alias;

Is worth mentioning that this constraints **don't exist on simple features** layer names.


nmco:
I would suggest a more meaningful title than "Fixes" for this pull request. I would suggest something like:
```
[GEOS-8918] [GEOS-8919] Relax existing constraints on complex feature  layers names
```
I saw that you add the link to the associated issues, thank you, this makes things easier to connect.

nmco:
I'm not sure why there is only a single commit for this two issues, they are related but they are still different issues.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaTestSupport.java
```diff
@@ -607,13 +607,12 @@ protected NestedFilterToSQL createNestedFilterEncoder(FeatureTypeMapping mapping
...
+     * Utility method that converts a XML document object to a string.
      *
-     * @param document
-     * @return
-     * @throws TransformerException
+     * @param document Xml Document to parse
+     * @return String representation of xml document
      */
-    protected String toString(Document document) throws TransformerException {
+    public static String toString(Document document) throws TransformerException {
```
nmco: Does it need to be ``public`` is ``protected`` not enough ?

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/LayersNamesAliasingTests.java
```diff
@@ -0,0 +1,229 @@
...
+        featureType.setName("MyStation");
+        featureType.setNativeName("Station");
+        catalog.add(featureType);
+        LayerInfo layer = builder.buildLayer(featureType);
+        layer.setDefaultStyle(catalog.getStyleByName("point"));
+        catalog.add(layer);
+    }
+
+    @Test
+    public void testMyStationGet() throws Exception {
```
nmco: I would suggest as test name ``testAliasedNameWfsGetFeature11`` instead of ``testMyStationGet``.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/LayersNamesAliasingTests.java
```diff
@@ -0,0 +1,229 @@
...
+        // requested with MyStation, must returns Station:
+        checkCount(
+                WFS11_XPATH_ENGINE,
+                document,
+                3,
+                "//wfs:FeatureCollection/gml:featureMembers/st:Station");
+    }
+
+    @Test
+    public void testMyStationPostQuery() throws Exception {
```
nmco: I would suggest the same name pattern as above.

src/extension/app-schema/app-schema-test/src/test/resources/test-data/stations/layerNamesTest/measurementsDefaultGeometry.properties
```diff
@@ -0,0 +1,4 @@
+_=ID:String,NAME:String,UNIT:String,STATION:String,SAMPLEDAREA_CODE:String,SAMPLEDAREA_GEOM:Geometry:srid=4326
```
nmco: What do ``DefaultGeometry`` tests cases have to do with this ? The stations use case base properties should be enough for this no ?

src/wfs/src/main/java/org/geoserver/wfs/kvp/TypeNameKvpParser.java
```diff
@@ -43,8 +42,7 @@ protected Object parseToken(String token) throws Exception {
             if (ftInfo == null) {
                 return new QName(null, token);
             } else {
-                final Name name = ftInfo.getFeatureType().getName();
-                return new QName(name.getNamespaceURI(), name.getLocalPart());
+                return new QName(ftInfo.getNamespace().getURI(), token);
```
nmco: I had to stop for a moment on this change, so indeed before we were always using the layer native name, this was indeed not allowing any type of aliasing. This change looks good to me and seems to be backward compatible, and the fact that tests passed makes believe that indeed this is full backward compatible.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AliasStationsMockData.java
```diff
@@ -0,0 +1,10 @@
+package org.geoserver.test;
+
+public class AliasStationsMockData extends StationsMockData {
```
nmco: This class could be transformed into a sub-private-final-static class ``LayersNamesAliasingTests.java``, or even better: instantiated has an anonymous class. If nobody else is using it, nobody should spend time trying to udnertand why this class exists :)

nmco:
This pull request looks good to me:
- this fix a bug \ limitation, so no documentation update needed
- this PR ads the necessary tests
- looks like the changes are backward compatible, i.e. only people affected by this change will notice a difference

fernandor777:
Thanks, proceeding to rebuild the backport branch with last changes.

bradh:
This is also showing format failures.

aaime:
Indeed as @bradh mentions the build is failing for a legit format failure. Please make sure to run a maven build before pushing on PRs:

```
[INFO] Processed 100 files (1 non-complying).
[INFO] Processed 142 files (1 non-complying).
[ERROR] Non complying file: /home/travis/build/geoserver/geoserver/src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/LayersNamesAliasingTests.java
```

fernandor777:
Indeed, looks like Eclipse git plugin was not able to detect some minor changes (line delete probably) and didn't add changes to commit - push.  Re-done with git diff and add on command line directly. Thanks.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java
```diff
@@ -489,6 +492,11 @@ private static void writeInfoFile(
         }
     }
 
+    private String getLayerName(String typeName) {
```
nmco: This behavior should not be added here. This is a _specific_ behavior only used in particular test case.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java
```diff
@@ -394,7 +396,7 @@ private void copy(String content, String location) {
      * @param featureTypeDir feature type directory
      * @param dataStoreName data store directory name
      */
-    private static void writeInfoFile(
+    private void writeInfoFile(
```
nmco: The received type name as input should already be complete, consequently the method should stay ``public``.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java
```diff
@@ -632,8 +640,8 @@ protected void removeNamespace(String namespacePrefix) {
      * @param typeName local name of the WFS feature type
      * @return name of the data store for the feature type
      */
-    protected static String getDataStoreName(String namespacePrefix, String typeName) {
-        return namespacePrefix + "_" + typeName;
+    protected String getDataStoreName(String namespacePrefix, String typeName) {
+        return namespacePrefix + "_" + getLayerName(typeName);
```
nmco: Same comment as above.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java
```diff
@@ -652,7 +660,7 @@ protected File getFeatureTypesBaseDir() {
      * @param typeName local name of the WFS feature type
      * @return directory that contains the mapping and property files
      */
-    private static File getFeatureTypeDir(
+    private File getFeatureTypeDir(
```
nmco: Any specific reason to make this method no ``static`` anymore ?

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java
```diff
@@ -781,4 +789,17 @@ private String modifyOnlineMappingFileContent(String mappingFileName) throws IOE
         }
         return content.toString();
     }
+
+    /**
+     * Prefix for layer name, to test name vs nativeName
+     *
+     * @return layer name prefix
+     */
+    public String getLayerNamePrefix() {
```
nmco: See my comments above, this specific behavior should not be implemented here.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java
```diff
@@ -42,6 +42,8 @@
 public abstract class AbstractAppSchemaMockData extends SystemTestData
         implements NamespaceTestData {
 
+    private String layerNamePrefix;
```
nmco: See my comments bellow, but long story short, this behavior should not be implement here, it should be up to the test setup or test using this class to set the final type name.

If everyone starts adding its specific bit of business to this class it would end up with 10'000  lines.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaTestSupport.java
```diff
@@ -607,13 +607,12 @@ protected NestedFilterToSQL createNestedFilterEncoder(FeatureTypeMapping mapping
...
+     * Utility method that converts a XML document object to a string.
      *
-     * @param document
-     * @return
-     * @throws TransformerException
+     * @param document Xml Document to parse
+     * @return String representation of xml document
      */
-    protected String toString(Document document) throws TransformerException {
+    protected static String toString(Document document) throws TransformerException {
```
nmco: Why non ``static`` anymore ?

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/LayersNamesAliasingTests.java
```diff
@@ -0,0 +1,233 @@
...
+        info.setEnabled(true);
+        info.setNativeName("Station_gml31");
+        catalog.save(info);
+
+        info = catalog.getFeatureTypeByName("st_gml32", "lyr_Station_gml32");
+        info.setEnabled(true);
+        info.setNativeName("Station_gml32");
+        catalog.save(info);
+    }
+
```
nmco: This is what I would expect to see:

- testAliasedNameWfsGetFeature11
- testAliasedNameWfsGetFeature20
- testAliasedNameWfsGetFeaturePost11
- testAliasedNameWfsGetFeaturePost20
- testLocalWorkspaceNoPrefixWfsGetFeature11
- testLocalWorkspaceNoPrefixWfsGetFeature20
- testLocalWorkspaceNoPrefixWfsGetFeaturePost11
- testLocalWorkspaceNoPrefixWfsGetFeaturePost20

I find that having the same tests that only differ on the WFS version used side by side helpful.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/LayersNamesAliasingTests.java
```diff
@@ -0,0 +1,233 @@
...
+            throw new RuntimeException("Error evaluating xpath.", exception);
+        }
+    }
+
+    public static void copyResource(String resourcePath, Path directoryPath) throws IOException {
+        ClassLoader cl = LayersNamesAliasingTests.class.getClassLoader();
+        Files.copy(cl.getResourceAsStream(resourcePath), directoryPath);
+    }
+
+    /** * GetFeature tests ** */
```
nmco: This comment really made things clear to me.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/LayersNamesAliasingTests.java
```diff
@@ -0,0 +1,233 @@
...
+        mappingsFolderPath = destDir.toPath();
+    }
+
+    @Override
+    protected void onSetUp(SystemTestData testData) throws Exception {
+        super.onSetUp(testData);
+        Catalog catalog = getCatalog();
+
+        FeatureTypeInfo info = catalog.getFeatureTypeByName("st_gml31", "lyr_Station_gml31");
+        info.setEnabled(true);
```
nmco: The method ``createTestData`` bellow doesn't already setup everything ? Is question, is not clear to me.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/LayersNamesAliasingTests.java
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    public static void copyResource(String resourcePath, Path directoryPath) throws IOException {
+        ClassLoader cl = LayersNamesAliasingTests.class.getClassLoader();
+        Files.copy(cl.getResourceAsStream(resourcePath), directoryPath);
+    }
+
+    /** * GetFeature tests ** */
+    @Override
+    protected AliasStationsMockData createTestData() {
```
nmco: I would keep code that ``works on the setup of the tests`` together, at beginning or at the end (IMHO it should at the end).

nmco:
As being in stand by for too long, closing it.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java
fernandor777: Right, done.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java
fernandor777: Done.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java
fernandor777: In this case, due to extended use along the abstract class for this method and the need to change its behavior on extended classes its needed to make the method non-static.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaTestSupport.java
fernandor777: For use the Document to String convert logic without repeating code on other parts of testing code.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/LayersNamesAliasingTests.java
fernandor777: Good, done.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/LayersNamesAliasingTests.java
fernandor777: No, this method does setup for layers needed on current tests.

fernandor777:
Review changes done, thanks.

aaime:
Checked the failed Travis build, it's due to misformatted files in the mapml community module, which have been fixed in the meantime. The contents of this PR should be good.

PR #3096, opened 2018-09-11
[GEOS-7636] GetLegendGraphic online resource is ignored on rasters

fernandor777:
Fix and test for GEOS-7636 "GetLegendGraphic ignores the OnlineResource configured in the style for raster layers"

This PR modify some nested conditional logic that prevented to show the correct OnlineResource and always returned the standard blue square.  Now GetLegendGraphic shows the proper image.

https://osgeo-org.atlassian.net/browse/GEOS-7636

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/GetLegendGraphicRequest.java
+++ b/src/wms/src/main/java/org/geoserver/wms/GetLegendGraphicRequest.java
@@ -134,6 +134,21 @@ public class GetLegendGraphicRequest extends WMSRequest {
             this.layerName = featureType.getName();
         }
 
+        /**
+         * LegendRequest for a feature type, additional details (title and legend graphic) provided
+         * by MapLayerInfo.
+         *
+         * @param featureType
+         * @param layerName layerName distinct to featureType name
+         */
+        public LegendRequest(FeatureType featureType, Name layerName) {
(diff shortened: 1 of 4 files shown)


src/wms/src/main/java/org/geoserver/wms/GetLegendGraphicRequest.java
```diff
@@ -134,6 +134,21 @@ public LegendRequest(FeatureType featureType) {
...
         }
 
+        /**
+         * LegendRequest for a feature type, additional details (title and legend graphic) provided
+         * by MapLayerInfo.
+         *
+         * @param featureType
+         * @param layerName layerName distinct to featureType name
+         */
+        public LegendRequest(FeatureType featureType, Name layerName) {
```
aaime: Is this change actually related to the GetLegendGraphic on raster ignoring the pre-drawn legend?

fernandor777: In case of GridCoverage the featureType Name provided is not the layer name, and without this change the blue square is returned.

aaime: Ah right, it's still pretending that it's a vector... yeah, I remember now

PR #3097, opened 2018-09-12
NetCDF-out: Reworded attribute blacklisting method and added global a…

dromagnoli:
…ttribute


dromagnoli:
Failure is related to random GWC behavior. Merging

PR #3098, opened 2018-09-12
NetCDF-out: Reworded attribute blacklisting method and added global a…

dromagnoli:
…ttribute. bkp to 2.14.x


dromagnoli:
Failure is unrelated. Merging

PR #3106, opened 2018-09-18
[GEOS-8933] Requesting workspaces triggers an NPE when proxy headers

lmoran:
are used for the proxy URL

- Added a guard to avoid proxying an URL when it does not come from the
  OWS Dispatcher.


bradh:
Will close / reopen to try to get a different VM.

PR #3107, opened 2018-09-18
[GEOS-8936] Allow Additional Styles select boxes to scroll

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-8936

This fix modifies the Palette.html template used by wicket, wrapping the each of the `<select>` elements in a div so that a horizontal scroll can be added (the select element does *not* support horizontal scroll on its own). Note: Wicket, and by extension the original Palette.html, is licensed under Apache 2.0, which is compatible with GPL and permits modified files to be relicensed.

Limitations:

* If you have a wide style name such that horizontal scrolling is required, the vertical scroll bar for the select will start off outside the visible region of the `<select>`. This is a limitation of how the select element works.

* If you have an especially large style name, the default style select box higher up on the page will overflow the screen. This is unrelated to this change, but was discovered during testing (this can be seen in the example below).

Example of horizontal scroll in action (On OS X, so the actual scrollbar is hidden by the OS):

styles_scroll_1

I've tested these changes on both Windows and OS X, and on Chrome, Firefox, Safari, Edge, and Internet Explorer to confirm they work on all modern browsers. Older browser versions were not tested.

No test case since the wicket tester doesn't support testing the actual size/rendering of web elements.

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/css/geoserver.css
+++ b/src/web/core/src/main/java/org/geoserver/web/css/geoserver.css
@@ -93,6 +93,39 @@ input[readonly] {
   padding: 0;
 }
 
+
+.palette-header {
+  margin-right: -2px;
+}
+
+.palette-theme-default .palette-choices > div {
+  width: 100%;
+  overflow-x: scroll;
+  border: 1px solid rgb(187, 187, 187);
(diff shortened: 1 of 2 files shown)


aaime:
+1

PR #3120, opened 2018-09-26
[GEOS-8945] Missing Spatial Index in OSEO Setup Instructions

randomorder:
Adding the missing index on the geometry attribute of the "granule". This can have a significant impact on systems with a great number of published granules


aaime:
Thanks!

PR #3121, opened 2018-09-27
Windows batch fixes

ronaldhoek:
Some fixes regarding the Windows batch files

https://osgeo-org.atlassian.net/browse/GEOS-8950

### DIFF
--- a/src/release/bin/shutdown.bat
+++ b/src/release/bin/shutdown.bat
@@ -10,11 +10,17 @@ echo.
 set error=0
 
 rem JAVA_HOME not defined
-if "%JAVA_HOME%" == "" goto noJava
+if "%JAVA_HOME%" == "" goto trySystemJava
 
 rem JAVA_HOME defined incorrectly
 if not exist "%JAVA_HOME%\bin\java.exe" goto badJava
 
+rem Setup the java command and move on
+set RUN_JAVA=%JAVA_HOME%\bin\java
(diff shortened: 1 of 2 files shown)


src/release/bin/shutdown.bat
```diff
@@ -24,7 +30,22 @@ if not exist "%GEOSERVER_HOME%\bin\startup.bat" goto badHome
 rem No errors
 goto shutdown
 
-
+:trySystemJava
+  echo The JAVA_HOME environment variable is not defined, trying to use System Java
```
jodygarnett: Is it worth checking JRE_HOME in addition to JAVA_HOME - geoserver 2 no longer requires a compiler for JSP pages and can run with a JRE.

jodygarnett:
Tested local see https://osgeo-org.atlassian.net/browse/GEOS-8950 for details.

I presume we are interested in seeing these changes backported

src/release/bin/shutdown.bat
ronaldhoek: Hmmm, when JAVA_HOME is not defined the batch will search for any Java contained in the PATH environment variable and then find the JRE-path anyway (unless the JRE is not added to the PATH).

Also, JRE_HOME is not set by the Oracle Windows JRE  Installer, so it's normally not set on Windows platforms (as far as I have seen).

PR #3123, opened 2018-09-27
GEOS-8935: introduced new coordinates formatting options for WFS GML output formats

mbarto:
Enables configuration of new formatting options introduced with GEOT-6119 for WFS GML output formats

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/FeatureTypeInfo.java
+++ b/src/main/src/main/java/org/geoserver/catalog/FeatureTypeInfo.java
@@ -93,8 +93,19 @@ public interface FeatureTypeInfo extends ResourceInfo {
      */
     void setPadWithZeros(boolean padWithZeros);
 
+    /**
+     * True if numbers should always be formatted as decimal (no scientific notation allowed).
+     *
+     * @return
+     */
     boolean getForcedDecimal();
 
+    /**
+     * Set to true if numbers should always be formatted as decimal (no scientific notation
+     * allowed).
+     *
+     * @param forcedDecimal
+     */
     void setForcedDecimal(boolean forcedDecimal);
 
     /**
--- a/src/main/src/main/java/org/geoserver/catalog/FeatureTypeInfo.java
+++ b/src/main/src/main/java/org/geoserver/catalog/FeatureTypeInfo.java
@@ -80,6 +80,24 @@ public interface FeatureTypeInfo extends ResourceInfo {
     void setNumDecimals(int numDecimals);
 
     /**
+     * If numbers float should be formatted right-padding them with zeros.
+     *
+     * @return
+     */
+    boolean getPadWithZeros();
+
+    /**
+     * Sets wether float numbers should be formatted right-padding them with zeros.
+     *
+     * @param padWithZeros
+     */
+    void setPadWithZeros(boolean padWithZeros);
+
+    boolean getForcedDecimal();
+
+    void setForcedDecimal(boolean forcedDecimal);
+
+    /**
      * Tolerance used to linearize this feature type, as an absolute value expressed in the
      * geometries own CRS
      */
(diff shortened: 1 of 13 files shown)


src/main/src/main/java/org/geoserver/catalog/FeatureTypeInfo.java
```diff
@@ -79,6 +79,24 @@
...
+    boolean getPadWithZeros();
+
+    /**
+     * Sets wether float numbers should be formatted right-padding them with zeros.
+     *
+     * @param padWithZeros
+     */
+    void setPadWithZeros(boolean padWithZeros);
+
+    boolean getForcedDecimal();
```
aaime: Interface method without javadocs, please add

src/wfs/src/main/java/org/geoserver/wfs/WFSGetFeatureOutputFormat.java
```diff
@@ -229,6 +238,38 @@ protected int getNumDecimals(List featureCollections, GeoServer geoServer, Catal
...
+    protected boolean getForcedDecimal(
+            List featureCollections, GeoServer geoServer, Catalog catalog) {
+        boolean forcedDecimal = false;
+        for (int i = 0; i < featureCollections.size(); i++) {
+            Boolean forced =
+                    getFeatureTypeInfoProperty(
+                            catalog,
+                            (FeatureCollection) featureCollections.get(i),
+                            fti -> fti.getForcedDecimal());
+            if (forced != null && forced.booleanValue()) {
```
aaime: This is a bit weird way to do the comparison, ``Boolean.TRUE.equals(forced)`` would be the idiomatic approach

src/wfs/src/main/java/org/geoserver/wfs/WFSGetFeatureOutputFormat.java
```diff
@@ -240,18 +281,14 @@ protected int getNumDecimals(List featureCollections, GeoServer geoServer, Catal
...
-                if (!featureTypeInfo.getEncodeMeasures()) {
-                    // no measures should be encoded
-                    encodeMeasures = false;
-                }
+            Boolean measures =
+                    getFeatureTypeInfoProperty(
+                            catalog,
+                            (FeatureCollection) featureCollections.get(i),
+                            fti -> fti.getEncodeMeasures());
+            if (measures != null && !measures.booleanValue()) {
```
aaime: Same as above

src/wfs/src/test/java/org/geoserver/wfs/xml/GMLOutputFormatTest.java
```diff
@@ -60,6 +62,35 @@ public void testGML2() throws Exception {
         assertNull(getFirstElementByTagName(dom, "gml:exterior"));
     }
 
+    @Test
+    public void testGML2CoordinatesFormatting() throws Exception {
+        enableCoordinatesFormatting();
```
aaime: Would be best to restore the config the way it was after the test, you could use a @After method to do that.

mbarto:
@aaime your observations should be addressed now

aaime:
Looks good. The failure on app-veyor is one of the random build failures still left on Windows (btw, I'm looking for devs on windows to help with those).

src/main/src/main/java/org/geoserver/catalog/impl/FeatureTypeInfoImpl.java
```diff
@@ -34,6 +34,8 @@
 
     protected int maxFeatures;
     protected int numDecimals;
+    protected boolean padWithZeros;
+    protected boolean forcedDecimal;
```
tbarsballe: I just noticed this now, but you should probably be including these two new parameters in the equals() and hashCode() implementations

mbarto:
Good catch. I will make a quick-fix PR as soon as possible.
Thanks

PR #3124, opened 2018-09-27
[GEOS-8953] WFS shapefile and GeoJSON output can't handle geometries with measures

nmco:
**DEPENDS ON PULL REQUEST https://github.com/geotools/geotools/pull/2074** 
Associated issue:
https://osgeo-org.atlassian.net/browse/GEOS-8953

Makes GeoJSON handle geometries measures when writing, and add WFS integration tests for both GeoJSON and shapefiles.


nmco:
Done and build restarted, thank you for the heads up @simboss !

src/wfs/src/main/java/org/geoserver/wfs/WFSGetFeatureOutputFormat.java
```diff
@@ -212,8 +212,10 @@ public void write(Object value, OutputStream output, Operation operation)
         return null;
     }
 
-    protected int getNumDecimals(List featureCollections, GeoServer geoServer, Catalog catalog) {
+    protected int[] getNumDecimalsEncodeMeasures(
```
aaime: Ugh... I'd go for two separate methods with clearer names, if the copied code is too much, maybe you can pass in lambdas? It's true that mechanically one has to perform a similar operation, but decimals and measures are conceptually separate and actually orthogonal.

src/wfs/src/test/java/org/geoserver/wfs/response/ShapeZipTest.java
```diff
@@ -740,4 +752,159 @@ private String get4326_ESRI_WKTContent() {
...
+        fb.add(gf.createPoint(new CoordinateXYZM(1, 2, 3, 4)));
+        List<SimpleFeature> features = new ArrayList<SimpleFeature>();
+        features.add(fb.buildFeature("1"));
+        SimpleFeatureCollection featureCollection = DataUtilities.collection(features);
+        // write the zip shapefile bytes
+        byte[] zipBytes = writeOut(featureCollection);
+        // get shp file bytes only
+        byte[] resultBytes = getShpOnlyBytes(zipBytes);
+        // get expected byte array
+        InputStream resource =
```
aaime: The test works in its intention and does a valid test, but if it ever failed, how does one figure out where the difference is? They would have to open the file with some reader and figure out the differences (could there be "cosmetic" differences? I cannot remember, but file contents test seem to test more than just whether M is encoded or not). A different approach could be to read back the SHP, check the geometry type, read the geometry, check it has the expected contents.

src/wfs/src/main/java/org/geoserver/wfs/WFSGetFeatureOutputFormat.java
nmco: _It's true that mechanically one has to perform a similar operation, but decimals and measures are conceptually separate and actually orthogonal._

It was not about what the meaning of the values but about the operation which is the same: **getting meta-information from the feature type info**.

That say, I agree with you when you say _I'd go for two separate methods with clearer names_, I will do this.

src/wfs/src/test/java/org/geoserver/wfs/response/ShapeZipTest.java
nmco: I agree, but reading of geometries from shape files with M measurements is not supported (no funded). I spend a bit of extra time extending ``PropertiesDataStore`` with some base code to read geometries with M coordinates and implement it for line strings. No spare time left for the shape files reading.

aaime: Oh I see, no problem then.

nmco:
Changes applied, thank you for the review @aaime !

PR #3126, opened 2018-09-28
[GEOS-8944] GWC fails to cache tiles if GeoFence with source IP driven rules are used for authentication

aaime:
Associated issue: https://osgeo-org.atlassian.net/browse/GEOS-8944

### DIFF
--- a/src/gwc/src/main/java/org/geoserver/gwc/FakeHttpServletRequest.java
+++ b/src/gwc/src/main/java/org/geoserver/gwc/FakeHttpServletRequest.java
@@ -12,9 +12,9 @@ import java.security.Principal;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Enumeration;
-import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Optional;
 import javax.servlet.AsyncContext;
 import javax.servlet.DispatcherType;
 import javax.servlet.RequestDispatcher;
@@ -28,6 +28,8 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.servlet.http.HttpSession;
 import javax.servlet.http.Part;
+import org.springframework.web.context.request.RequestContextHolder;
+import org.springframework.web.context.request.ServletRequestAttributes;
 
 @SuppressWarnings("rawtypes")
 class FakeHttpServletRequest implements HttpServletRequest {
@@ -49,9 +51,11 @@ class FakeHttpServletRequest implements HttpServletRequest {
 
     private String workspace;
 
-    private Map<String, String> parameterMap = new HashMap<String, String>(10);
+    private Map<String, String> parameterMap;
 
     private Cookie[] cookies;
+
+    private Optional<HttpServletRequest> original;
 
     public FakeHttpServletRequest(Map<String, String> parameterMap, Cookie[] cookies) {
         this(parameterMap, cookies, null);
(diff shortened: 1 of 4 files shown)


src/gwc/src/main/java/org/geoserver/gwc/FakeHttpServletRequest.java
```diff
@@ -62,6 +66,13 @@ public FakeHttpServletRequest(
         this.parameterMap = parameterMap;
         this.cookies = cookies;
         this.workspace = workspace;
+        // grab the original request from Spring to forward security related attributes
+        // such as requests host, ports and headers
+        this.original =
+                Optional.ofNullable(
+                                (ServletRequestAttributes)
+                                        RequestContextHolder.getRequestAttributes())
+                        .map(atts -> atts.getRequest());
```
nmco: Not sure if you notice, a method reference can be used here:
```
        this.original =
                Optional.ofNullable(
                                (ServletRequestAttributes)
                                        RequestContextHolder.getRequestAttributes())
                        .map(ServletRequestAttributes::getRequest);
```

src/gwc/src/main/java/org/geoserver/gwc/FakeHttpServletRequest.java
```diff
@@ -300,35 +312,35 @@ public String getRealPath(String arg0) {
     }
 
     public String getRemoteAddr() {
-        return "127.0.0.1";
+        return original.map(r -> r.getRemoteAddr()).orElse("127.0.0.1");
```
nmco: Wondering, is there a valid situation where the remote address will not be available ? I can't think of any. I would feel more safe throwing a  ``ServletDebugException``. The same opinion for the port (method above).

src/gwc/src/test/java/org/geoserver/gwc/HttpRequestRecorderCallback.java
```diff
@@ -0,0 +1,32 @@
+package org.geoserver.gwc;
```
nmco: Missing header.

src/gwc/src/test/java/org/geoserver/gwc/HttpRequestRecorderCallback.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import javax.servlet.http.HttpServletRequest;
+import org.geoserver.ows.AbstractDispatcherCallback;
+import org.geoserver.ows.Request;
+
+public class HttpRequestRecorderCallback extends AbstractDispatcherCallback {
```
nmco: Could be made ``final``, I know this is a tests support class, but I would add some minor comment explaining what is going on:

```
Integration tests support class that keeps track of the requests send to the dispatcher.
```

nmco:
Looks good, just some minor comments.

src/gwc/src/test/java/org/geoserver/gwc/HttpRequestRecorderCallback.java
aaime: Will add

src/gwc/src/main/java/org/geoserver/gwc/FakeHttpServletRequest.java
aaime: That might break existing code, better to preserve backwards compatibility

src/gwc/src/main/java/org/geoserver/gwc/FakeHttpServletRequest.java
aaime: I did not think about it, but not sure replacing 26 chars with 37 to do the same thing is an improvement ;-)

src/gwc/src/main/java/org/geoserver/gwc/FakeHttpServletRequest.java
nmco: Ok, fair enough.

src/gwc/src/test/java/org/geoserver/gwc/HttpRequestRecorderCallback.java
aaime: No problem

src/gwc/src/main/java/org/geoserver/gwc/FakeHttpServletRequest.java
nmco: :dancer:

PR #3130, opened 2018-09-28
[GEOS-8869] Importer REST should not override default LayerInfo Rest converter

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-8869


src/extension/importer/rest/src/main/java/org/geoserver/importer/rest/ImportLayer.java
```diff
@@ -0,0 +1,19 @@
+package org.geoserver.importer.rest;
```
jodygarnett: header

aaime:
Looks good to me

tbarsballe:
Backported to 2.14.x and 2.13.x

PR #3133, opened 2018-09-30
[GEOS-8955] Disable docs for Resumable REST and REST mapper.

bradh:
The community modules have been disabled for a couple of years, this makes the docs consistent.


aaime:
Yep, thanks!

PR #3134, opened 2018-09-30
Update documentation versions.

bradh:


aaime:
Nice, thanks... there is a part of the release script that updates the version numbers, would be nice to update it to keep the docs current (just an idea, not asking you to do it, don't have bandwidth myself).

bradh:
I think the script should already do that (or tries), maybe it just didn't get committed. There is an override in any case, so release outputs should be fine.

PR #3135, opened 2018-09-30
[GEOS-8956] Update docs to fix sphinx warnings.

bradh:
There are still 20+ warnings for where pygments isn't lexing the samples correctly.


aaime:
Oh good. Merging... @bradh with these changes, does it build without any warning at all?
If so, we could update the docs build setup so that it fails on any issue, like the GeoTools one. @jodygarnett what do you think?

jodygarnett:
It would be great @aaime broken docs links (a common warning) help nobody.

aaime:
Cool, anyone has time to make the docs build stricter?

jodygarnett:
We cannot kill the rest examples yet, sadly the swagger docs are incomplete.

bradh:
I only meant for the community modules that aren't being compiled in any case. See
https://github.com/geoserver/geoserver/blob/master/src/community/pom.xml#L51-L59

We haven't built those since the REST rework.

PR #3137, opened 2018-10-01
[GEOS-8957] Clean up remaining doc warnings, switch config.

bradh:


bradh:
I don't think we're losing anything. Its already not highlighting. See http://docs.geoserver.org/latest/en/user/styling/css/cookbook/point.html#simple-point as an example.

I didn't check other highlighters though. Can take a look tomorrow.

aaime:
Doh, you're right!

bradh:
less and sass didn't work, but it looks like scss can give us some help.

image

The `:nth_image` part isn't great, but its probably OK overall.

aaime:
@bradh I think it looks good (and much better better than what we have now), +1

bradh:
Will close this PR and submit off a rebased / clean branch.

PR #3138, opened 2018-10-01
[GEOS-8875] Implement FIDs for Mapbox vector tiles

victorzinho:
In order to make it work I had to execute `mvn install` for the `java-vector-tile` library locally. I guess we need to add version 1.3.3 to the Boundless Maven repo: https://repo.boundlessgeo.com/main/no/ecc/vectortile/java-vector-tile/

Also, I'm a bit unsure about the fid parsing. Currently it only checks `<anything>.<number>` and uses `-1` (interpreted as no FID by the underlying library) if it cannot be parsed. Are there any other possible FIDs that GeoServer can send and that I need to take into account?

### DIFF
--- a/src/extension/vectortiles/src/main/java/org/geoserver/wms/mapbox/MapBoxTileBuilder.java
+++ b/src/extension/vectortiles/src/main/java/org/geoserver/wms/mapbox/MapBoxTileBuilder.java
@@ -9,16 +9,19 @@ import static org.geoserver.wms.mapbox.MapBoxTileBuilderFactory.MIME_TYPE;
 import java.awt.Rectangle;
 import java.io.IOException;
 import java.util.Map;
+import java.util.logging.Logger;
 import no.ecc.vectortile.VectorTileEncoder;
 import no.ecc.vectortile.VectorTileEncoderNoClip;
 import org.geoserver.wms.WMSMapContent;
 import org.geoserver.wms.map.RawMap;
 import org.geoserver.wms.vector.VectorTileBuilder;
 import org.geotools.geometry.jts.ReferencedEnvelope;
+import org.geotools.util.logging.Logging;
 import org.locationtech.jts.geom.Geometry;
 
 /** @author Niels Charlier */
 public class MapBoxTileBuilder implements VectorTileBuilder {
+    private static final Logger LOGGER = Logging.getLogger(MapBoxTileBuilder.class);
 
     private VectorTileEncoder encoder;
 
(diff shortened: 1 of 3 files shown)


bradh:
So you had to build java-vector-tile from source?

victorzinho:
Yes, there were some changes required in `java-vector-tile` that has been already [merged](https://github.com/ElectronicChartCentre/java-vector-tile/pull/25) and the version (1.3.3) has been [tagged](https://github.com/ElectronicChartCentre/java-vector-tile/tree/v1.3.3), but it has not been deployed to any Maven repo yet. I assume the idea would be to deploy to the Boundless [repo](https://repo.boundlessgeo.com/main/no/ecc/vectortile/java-vector-tile/) together with the previous versions.

jodygarnett:
And geoserver psc member has access to repo.boundlesgeo.com for uploading third-party dependencies like this. If you give me the jar via email I can upload it?

As for other FID formats - when publishing a table you can take more control over generating out a fid based on primary keys. I am not sure if they follow the <anything>.<number> format.

We came up with <anything>.<number> format because feature ids are supposed to be unique in the GML spec and must start with a letter. So starting with typename was chosen.

aaime:
@jodygarnett they all follow the "typename.key" format, otherwise the GetFeature requests using just "&featureid=<id>" without the typename (valid and tested by CITE tests) would fail. In other words, a id not following that format would break WFS compliance anyways.

aaime:
@victorzinho see https://github.com/geoserver/geoserver/pull/3177 for an upgrade
If you could review that PR and then rebase this one on top of it, I guess we'd be both in business.
The jar we need has been deployed, if you check the other PR, you'll see that some other changes were needed to do the upgrade (packaging instructions).

aaime:
I've manually merged this change on master, closing the PR

pere:
hi all, 

 Taking into account that you solved this issue on October 2018 (and my  Geoserver build date is 20-Nov-2018) should I assume your changes are already integrated on my Geoserver? 
Or should I proceed to update something? (java-vector-tile ??).  

Below are the parameters of my Geoserver.
Thanks in advance.

GeoServer Version
2.14.1
Git Revision
499990d642d60e6043f3d44deb26ac6172a4e885
Build Date
20-Nov-2018 02:04
GeoTools Version
20.1 (rev ef377c748a46efd4a84bfa0df939aac71ade0863)
GeoWebCache Version
1.14.1 (rev 1.14.x/4bb10e75710629ad5571e69b75684cf49256ad67)

mprins:
@pere You should update to a supported version (2.15.3 or later) your version is EOL.
The ticket states the version it was fixed for: https://osgeo-org.atlassian.net/browse/GEOS-8875

PR #3146, opened 2018-10-03
[GEOS-8953] WFS shapefile and GeoJSON output can't handle geometries with measures (backport 2.14.x)

nmco:
**Depends on https://github.com/geotools/geotools/pull/2088**

Backport pull request: https://github.com/geoserver/geoserver/pull/3124

Backport discussion:
http://osgeo-org.1560.x6.nabble.com/Backporting-geometries-measures-M-related-fixes-td5380344.html


src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
```diff
@@ -123,8 +126,21 @@ private JSONBuilder writeGeomCollection(GeometryCollection collection) {
         return this.endArray();
     }
 
-    private JSONBuilder writeCoordinates(Coordinate[] coords) throws JSONException {
-        return writeCoordinates(new CoordinateArraySequence(coords));
+    private JSONBuilder writeCoordinate(Point point) throws JSONException {
+        CoordinateSequence seq = point.getCoordinateSequence();
+        if (!seq.hasM() || !encodeMeasures) {
```
jodygarnett: logic like this makes me wish for an EnumSet of XYZ, XYZM.

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
```diff
@@ -139,13 +155,31 @@ private JSONBuilder writeCoordinates(CoordinateSequence coords) throws JSONExcep
...
+            if (measures == 0 || !encodeMeasures) {
+                if (dim > 2) {
+                    writeCoordinate(coords.getX(i), coords.getY(i), coords.getOrdinate(i, 2));
+                } else {
+                    writeCoordinate(coords.getX(i), coords.getY(i));
+                }
             } else {
-                writeCoordinate(coords.getX(i), coords.getY(i));
+                // if is XYZM
+                if (dimension - measures > 2) {
```
jodygarnett: Quick feedback - doing the logic in the loop is not as fast as doing it outside - and having each block contain its own for lop.

This code is valid but not quite what I expected:

- You have access to coords.hasM() and coords.hasZ() methods if you want to make this logic more readable. If taking this approach consider using the coords.getM(index) and coords.getZ( index ) methods.
- The method used, coords.getOrdinate(index, ordinate), was intended for blindly copying out ordinate values without creating temporary Coordinate object. I kind of wanted it to be fast rather than having to do any logic around dimension and measure.

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
nmco: I see your point I will rewrite a more generic method and simplify this syntax.

nmco:
Thank you for the review @jodygarnett. Changes applied, I reduced those method to a few lines only, the elated master PR here: https://github.com/geoserver/geoserver/pull/3155

PR #3152, opened 2018-10-04
Wfs3 tiling scheme description double format locale

fernandor777:
This PR fix the problem on tiling scheme description on no-US Locale JVM setting.  Now doubles format are always on Locale.US


aaime:
Failure in CSW, unrelated

PR #3153, opened 2018-10-04
GEOS-8958 - ClientStreamAbortedException should be logged at debug level

tixvieira:
At ows dispatcher class the level is FINER:
"
if (current instanceof ClientStreamAbortedException) {
logger.log(Level.FINER, "Client has closed stream", t);
"
So i commit with the same FINER Level.


src/wms/src/main/java/org/geoserver/wms/map/png/PNGJWriter.java
```diff
@@ -51,7 +51,7 @@ public RenderedImage writePNG(
         try {
             output = writer.writePNG(image, outStream, quality, filterType);
         } catch (Exception e) {
-            LOGGER.log(Level.SEVERE, "Failed to encode the PNG", e);
+            LOGGER.log(Level.FINER, "Failed to encode the PNG", e);
```
aaime: No... it should check if there is a ClientStreamAbortedException wrapped, or something else. Check how the dispatcher does it.
https://github.com/geoserver/geoserver/blob/5afddea76c9b95c2a9005ebb3957338719f488c3/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java#L1679

aaime:
Ah, also, the PR must start on master, once accepted and merged backports PRs to 2.14.x and 2.13.x are more than welcomed :-)

src/wms/src/main/java/org/geoserver/wms/map/png/PNGJWriter.java
aaime: A simple approach could be to just avoid logging and have the exception go up the chain, it will get logged by the Dispatcher if it's not a client stream aborted exception, e.g.:

```
       } catch (Exception e) {
        }
```

tixvieira:
I'm closing this PR, and open a new one, because i messed up somehow changing from the branch to master, and now it says that i want to merge 47 commits.

tixvieira:
New PR for future reference:
https://github.com/geoserver/geoserver/pull/3158

PR #3156, opened 2018-10-06
[GEOS-8957] Clean up remaining doc warnings and switch configuration

bradh:
This makes use of the scss parser, which seems more tolerant. There are still a couple of cases that aren't parsed and we have no syntax highlighting, but they're minor.

### DIFF
--- a/doc/en/build.xml
+++ b/doc/en/build.xml
@@ -45,7 +45,7 @@
   <target name="user" depends="init"
     description="Generate user html documentation for GeoServer">
     <mkdir dir="${build.directory}"/>
-    <antcall target="sphinx-ignore-warnings">
+    <antcall target="sphinx">
       <param name="id" value="user" />
       <param name="build" value="html" />
     </antcall>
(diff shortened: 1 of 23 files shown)


doc/en/build.xml
```diff
@@ -96,12 +96,6 @@
   <target name="full" depends="docguide,user,developer">
   </target>
 
-  <target name="sphinx-ignore-warnings" if="sphinx.available">
```
aaime: Ouch ouch, this introduced a regression... the build is no more incremental, it takes almost a minute now to build the docs. I believe the reason is due to the lack of the arg line element in the "sphinx" target.

bradh: See if I can work it today.

bradh: @aaime Can't reproduce. 

A full build from clean (`mvn install` - so the HTML outputs) on my VM takes 2 min 14 sec (all of the `en` docs - docguide, user, developer + the swagger bits). A subsequent build takes about 12 seconds. If I just touch the `user/source/index.rst` file, a build takes about 16 seconds. 

It looks like its doing incremental updates to me. 

Do you have something else?

PR #3158, opened 2018-10-06
GEOS-8958 - ClientStreamAbortedException should be logged at debug level

tixvieira:
Linked to this PR: https://github.com/geoserver/geoserver/pull/3153

At ows dispatcher class the level is FINER.

@aaime commented on the old pull request:
A simple approach could be to just avoid logging and have the exception go up the chain, it will get logged by the Dispatcher if it's not a client stream aborted exception, e.g.:
       } catch (Exception e) {
        }

I tested and it works fine.
It logs as a TRACE level as expected.

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/map/png/PNGJWriter.java
+++ b/src/wms/src/main/java/org/geoserver/wms/map/png/PNGJWriter.java
@@ -9,7 +9,6 @@ import ar.com.hjg.pngj.FilterType;
 import it.geosolutions.imageio.plugins.png.PNGWriter;
 import java.awt.image.RenderedImage;
 import java.io.OutputStream;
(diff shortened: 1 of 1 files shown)


aaime:
The build failed due to code formatting violation:

```
[ERROR] Failed to execute goal com.coveo:fmt-maven-plugin:2.4.0:check (default) on project gs-wms: Found 1 non-complying files, failing build -> [Help 1]
```

The easiest way to solve it is to build using maven from command line, either a normal build like "mvn install" or one specific for formatting, "mvn com.coveo:fmt-maven-plugin:format"

tixvieira:
I close this PR, because i can't do a change over the github interface without Format Errors when processing maven.
The new PR: https://github.com/geoserver/geoserver/pull/3165

PR #3161, opened 2018-10-07
[GEOS-8967] Reduce verbosity in WPS tests.

bradh:
One case is just dumping what was read from the file, which is not useful.


aaime:
Yep

PR #3165, opened 2018-10-07
[GEOS-8958] - ClientStreamAbortedException should be logged at debug level

tixvieira:
Old PRs:
https://github.com/geoserver/geoserver/pull/3153
https://github.com/geoserver/geoserver/pull/3158

At ows dispatcher class the level is FINER.

@aaime commented on the old pull request:
A simple approach could be to just avoid logging and have the exception go up the chain, it will get logged by the Dispatcher if it's not a client stream aborted exception, e.g.:
} catch (Exception e) {
}

I tested and it works fine.
It logs as a TRACE level as expected.

I create this new PR, because i can't do a change over the github interface without Format Errors when processing maven.


aaime:
Merging, thanks! If you have time, backports PRs would also be appreciated.

tixvieira:
PR 2.14.x:
https://github.com/geoserver/geoserver/pull/3172

PR #3167, opened 2018-10-08
[GEOS-8967] Reduce test verbosity in WMS module.

bradh:


sennahLessert:
why don't you use something like a switch to enable wms request verbosity, instead of commiting commented source code ^^.

aaime:
@sennahLessert the "print" statements actually already have a switch inside but it's not handy to use. The sysout obviously don't, they could be turned into logging statements, but it takes significantly more time and nobody is getting paid to work on this.

If you want to influence the project, first check on the list about your proposed approach (make sure others agree), and then do a PR using your own time. We need more maintainers here, people just sharing complaints or suggestions are going to change anything I'm afraid (besides killing the enthusiasm of the few that actually do, which is something we cannot afford).

sennahLessert:
ok, next time i will check my complain.  sorry for that

bradh:
While it would be neater (and easier) to just remove the code, and occasionally I do, it usually represents debugging insight from the original author. That insight would be recoverable from git history, but I think reducing developer friction where I can is generally useful.

I do recognise that different people have different views on this.

PR #3175, opened 2018-10-10
[GEOS-8971] Use standard MapBox Vector Tiles mime type

aaime:
Depends on https://github.com/GeoWebCache/geowebcache/pull/694


aaime:
I certainly did before trying to open the PR... but I have this habit of self-reviewing the PR and fix the obvious oustanding bits, and might not have re-run the build after doing that. Easily solved anyways :)

tbarsballe:
I've done that a number of times myself. Thanks for the fix.

PR #3177, opened 2018-10-11
[GEOS-8974] Vector tiles ring orientations do not match the latest MVT spec

aaime:
The ring orientation fix has been done in the java-vector-tiles libraries, we just need to upgrade and handle the changed protobuf dependencies.


aaime:
Tested locally with a OL2 vector tiles client, gonna merge to push a build online where more clients can be tested (including some that are affected by ring orientations)

PR #3180, opened 2018-10-11
[GEOS-8973] add srs lookup to REST coverage handler

mtCarto:
[GEOS-8973](https://osgeo-org.atlassian.net/browse/GEOS-8973)

This adds CRS lookup to the REST coverage handler to avoid ArrayIndexOutOfBounds exceptions when the source files NativeCRS is missing identifiers.

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java
+++ b/src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java
@@ -832,8 +832,16 @@ public class CatalogBuilder {
 
         CoordinateReferenceSystem nativeCRS = cinfo.getNativeCRS();
 
-        if (cinfo.getSRS() == null) {
-            cinfo.setSRS(nativeCRS.getIdentifiers().toArray()[0].toString());
+        if (nativeCRS != null) {
+            try {
+                Integer code = CRS.lookupEpsgCode(nativeCRS, false);
+                if (code != null) {
+                    cinfo.setSRS("EPSG:" + code);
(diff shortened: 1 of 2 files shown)


src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java
```diff
@@ -832,8 +832,16 @@ public void initCoverage(CoverageInfo cinfo, final String coverageName) throws E
...
         CoordinateReferenceSystem nativeCRS = cinfo.getNativeCRS();
 
-        if (cinfo.getSRS() == null) {
-            cinfo.setSRS(nativeCRS.getIdentifiers().toArray()[0].toString());
+        if (nativeCRS != null) {
+            try {
+                Integer code = CRS.lookupEpsgCode(nativeCRS, false);
+                if (code != null) {
+                    cinfo.setSRS("EPSG:" + code);
+                    cinfo.setProjectionPolicy(ProjectionPolicy.REPROJECT_TO_DECLARED);
```
aaime: If a match was found, best to set the policy to "force declared". If there is a match, it means no reprojection is needed.

mtCarto: I was going by this [note](https://github.com/geoserver/geoserver/blob/bbf9b4f4efadf38a3d02bc10028077e34157a9a1/src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java#L987)  in the buildCoverageInternal method that the UI uses. Should that also be FORCE_DECLARED?

aaime:
Looks good thanks, I've re-kicked the windows build, it's likely a random failure but want to make sure.

PR #3181, opened 2018-10-12
2.14.x - fix [GEOS-8968] Geoserver printing plugin (print-lib) leaves opened files

tixvieira:
Geoserver printing plugin (print-lib) leaves opened files

- upgrade to print-lib 2.1.3

merge from master


tixvieira:
The issue can be closed: https://osgeo-org.atlassian.net/browse/GEOS-8968

PR #3193, opened 2018-10-23
Typo in GeoFence doc

Gnafu:


bradh:
Thanks for the fix. Applied.

PR #3194, opened 2018-10-23
Remove JDK internals from rendering status panel

dvntucker:
Providing slightly less information in this page in order to remove dependence on internals.

### DIFF
--- a/src/platform/src/main/java/org/geoserver/platform/RenderingEngineStatus.java
+++ b/src/platform/src/main/java/org/geoserver/platform/RenderingEngineStatus.java
@@ -5,41 +5,16 @@
 package org.geoserver.platform;
 
 import java.util.Optional;
-import sun.java2d.pipe.RenderingEngine;
 
 /** @author Morgan Thompson - Boundless */
 public class RenderingEngineStatus implements ModuleStatus {
 
-    private static final String UNKNOWN = "unknown";
+    private static final String DEFAULT = "PLATFORM DEFAULT";
 
-    private String engine;
     private String provider;
 
-    @SuppressWarnings("unchecked")
     public RenderingEngineStatus() {
-
-        Class<RenderingEngine> renderer;
-        try {
-            renderer =
-                    (Class<RenderingEngine>)
-                            sun.java2d.pipe.RenderingEngine.getInstance().getClass();
-        } catch (Throwable e) {
-            engine = UNKNOWN;
-            provider = UNKNOWN;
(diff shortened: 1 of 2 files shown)


src/platform/src/main/java/org/geoserver/platform/RenderingEngineStatus.java
```diff
@@ -5,41 +5,16 @@
...
-        if (pkg.getName().contains("marlin")) {
-            provider = "Marlin";
-        } else if (pkg.getName().contains("sun.dc")) {
-            provider = "OracleJDK";
-        } else if (pkg.getName().contains("sun.java2d")) {
-            provider = "OpenJDK";
-        } else {
-            provider = pkg.getName();
-        }
+        this.provider = System.getProperty("sun.java2d.renderer", DEFAULT);
```
jodygarnett: Q: So in OpenJDK 8 or Oracle JDK 8 this would just show up as empty right?

src/platform/src/main/java/org/geoserver/platform/RenderingEngineStatus.java
```diff
@@ -5,41 +5,16 @@
 package org.geoserver.platform;
 
 import java.util.Optional;
-import sun.java2d.pipe.RenderingEngine;
 
 /** @author Morgan Thompson - Boundless */
 public class RenderingEngineStatus implements ModuleStatus {
 
-    private static final String UNKNOWN = "unknown";
+    private static final String DEFAULT = "PLATFORM DEFAULT";
```
jodygarnett: Do we need to translate this? probably not ...

src/platform/src/main/java/org/geoserver/platform/RenderingEngineStatus.java
aaime: It would show up a "PLATFORM DEFAULT"

dvntucker:
Woops, I must have forgot to push the test changes I had.

src/platform/src/main/java/org/geoserver/platform/RenderingEngineStatus.java
dvntucker: Yes, it would show up as PLATFORM DEFAULT with no config. Realistically this status is only going to be meaningful running anything pre JDK8 I guess.

dvntucker:
Ok, looks like the test is passing now (because it's trivial)

PR #3195, opened 2018-10-23
Remove internal XMLChar reference, use xalan version instead.

dvntucker:


src/wms/src/main/java/org/geoserver/wms/featureinfo/RasterLayerIdentifier.java
```diff
@@ -17,7 +16,10 @@
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
+
 import javax.media.jai.PlanarImage;
+
+import org.apache.xml.utils.XMLChar;
```
bradh: Suggest removing the blank lines - I've seen the formatter barf on this.

tbarsballe: The formatter did indeed complain; I've pushed up a version with formatting fixes applied directly to jdk11; closing this PR.
https://github.com/geoserver/geoserver/commit/372188a966b038b04fd6b3794fa2dab100bb365a

PR #3196, opened 2018-10-24
[GEOS-8985]: Fixing GeoTransform Attribute's coefficients on NetCDF o…

dromagnoli:
…utput


src/extension/netcdf-out/src/main/java/org/geoserver/wcs/responses/NetCDFCRSWriter.java
```diff
@@ -90,9 +92,14 @@
     public NetCDFCRSWriter(NetcdfFileWriter writer, GridCoverage2D sampleGranule) {
         this.writer = writer;
         this.sampleGranule = sampleGranule;
-        GridGeometry gridGeometry = sampleGranule.getGridGeometry();
-        transform = gridGeometry.getGridToCRS();
         crs = sampleGranule.getCoordinateReferenceSystem();
+        RenderedImage image = sampleGranule.getRenderedImage();
+        GridToEnvelopeMapper geMapper =
```
aaime: So this handles eventual translations in the transform that should not be there in the output? I'd add a comment explaining

dromagnoli:
Explanation added

aaime:
+1

PR #3206, opened 2018-10-25
first pass at EMF upgrade

ianturton:
Should pass once https://github.com/geotools/geotools/pull/2147 is added to GeoTools


aaime:
I assume this PR can be closed?

aaime:
@ianturton closing under the assumption that this has been merged already, otherwise the code in GeoServer would not compile at all after the Java 11 work

PR #3208, opened 2018-10-25
[GEOS-8986] Update documentation \ info related with measures (M) support

nmco:
Associated issue:
https://osgeo-org.atlassian.net/browse/GEOS-8986

Documentation fixes regarding coordinates measures (M) support.


src/web/wfs/src/main/resources/GeoServerApplication.properties
```diff
@@ -50,7 +50,9 @@ WFSLayerConfig.otherSRS.message=A comma separated list of EPSG codes, e.g. 4326,
    override to a list of values specified in the WFS service configuration.
 
 WFSLayerConfig.coordinatesEncodingTitle=Coordinates Encoding
-WFSLayerConfig.coordinatesEncodingHelp.message=lalala
+WFSLayerConfig.coordinatesEncodingHelp.message= This settings control how geometries coordinates will be encoded in \
```
bradh: I think it should be `This setting controls how geometry coordinates will be encoded in`

PR #3220, opened 2018-10-30
[GEOS-8991] WCS 2.0 should return an InvalidSubsetting exception when requested out of time/elevation/custom range values

aaime:


aaime:
Build on appveyor took more than a hour and was killed. On Travis it took 41 minutes, we're getting close to the 50 minutes limit where builds are killed (and some builds are already going over from time to time)... as we graduate more modules, we'll have to look at lowering the build times of existing ones....

dromagnoli:
Ah!. Yep. indeed, while reviewing the PR, I tried a restart of the travis one since it was a "random" failure not related to your changes :)

PR #3223, opened 2018-10-30
[GEOS-8990] - Add an option to scale the vector tile size

dvntucker:
In practice, only MVT is interested in this, but because of how the paint area is set up it needed to be more generic, that's why I added it to the VTMOF.java interface.


src/extension/vectortiles/src/main/java/org/geoserver/wms/mapbox/MapBoxTileBuilderFactory.java
```diff
@@ -33,4 +33,32 @@ public String getMimeType() {
...
+     * Use 16x oversampling to match actual Mapbox tile extent, which is 4096 for 900913 tiles
+     *
+     * @return
+     */
+    @Override
+    public int getOversampleX() {
+        return 16;
+    }
+
+    @Override
```
tbarsballe: If you're going to give a javadoc for X, should add one to Y too

PR #3225, opened 2018-10-31
[GEOS-8996] CoverageView setup fails if one of the source bands has an indexed color model

aaime:


src/main/src/main/java/org/geoserver/catalog/CoverageViewReader.java
```diff
@@ -480,6 +484,60 @@ public GridCoverage2D read(GeneralParameterValue[] parameters)
...
+            Hints localHints = new Hints(hints);
+            localHints.put(JAI.KEY_IMAGE_LAYOUT, layout);
+            RenderedOp formatted = JAI.create("Format", paramBlk, localHints);
+
+            return new GridCoverageFactory()
+                    .create(
+                            coverage.getName(),
+                            formatted,
+                            coverage.getGridGeometry(),
+                            coverage.getSampleDimensions(),
```
simboss: Are you sure this is correct, I mean, are we suere the original coverage has 3 SD and not 1.
Just checking.

src/main/src/main/java/org/geoserver/catalog/CoverageViewReader.java
```diff
@@ -480,6 +484,60 @@ public GridCoverage2D read(GeneralParameterValue[] parameters)
...
+     * The BandMerge operation takes indexed images and expands them, however in the context of
+     * coverage view band merging we don't normally want that, e.g., raster mask bands are
+     * represented as indexed but we really want to keep them in their binary, single band form. To
+     * do so, the IndexColorModel is replaced by a ComponentColorModel
+     *
+     * @param coverage
+     * @return
+     */
+    private GridCoverage2D prepareForBandMerge(GridCoverage2D coverage) {
+        RenderedImage ri = coverage.getRenderedImage();
```
simboss: In ImageWorker there is a method called forceComponentColorModel which should do this work already.
I would use that rather than writing this code here, it is also optimized for using Lookup  rather than format (as lookup should be faster).

You might want to check if forceComponentColorModel has all you need and update it if needed.

src/main/src/main/java/org/geoserver/catalog/CoverageViewReader.java
aaime: The branch starts with "if(sampleModel.getNumBands() == 1" so the source image should have one sample dimension, just like the output one.

src/main/src/main/java/org/geoserver/catalog/CoverageViewReader.java
aaime: I've checked, the forceComponentColorModel uses the palette and expands it, so if the source image has a gray palette, we'll get with different values, e.g., with this:

```
Band 1 Block=1475x44 Type=Byte, ColorInterp=Palette
  Image Structure Metadata:
    NBITS=1
  Color Table (RGB with 2 entries)
    0: 0,0,0,255
    1: 255,255,255,255
```

the image past forceComponentColorModel has value that are 0 and 255 instead of 0 and 1 (or whatever gray color was assigned in the palette).

However, what I did not notice is that format does something similar, and when facing a binary image, it forcefully expands 0 to 0 and 1 to 255. I don't think this is correct either.

What I believe is needed, is to add a new parameter to forceComponentColorModel indicating it to preserve the original values constructing a LUT accordingly (a "identity" LUT)

aaime:
Prepared a variant that uses lookup with a identity LUT. Kept the code inside the CoverageViewReader for the moment, as it's quite different from what ImageWoker.forceComponentColorModel does (could be another if in that method, driven by a different parameter, but it would make the code ignore the other 3 params if the flag to preserve the values is up).
The short identity LUT is gonna use 128KB of memory, static, but seems better than allocating and filling all those values at every call.
Opinions?

aaime:
Method shared back, now depends on https://github.com/geotools/geotools/pull/2164

PR #3228, opened 2018-11-03
Some attempts at build time reduction *DO NOT MERGE*

aaime:
Basic stuff, PR to see how this affects Travis build time


src/main/src/main/java/org/geoserver/config/GeoServerLoader.java
```diff
@@ -319,7 +319,16 @@ protected void reloadInitializers(GeoServer geoServer) throws Exception {
      * Does some post processing on the catalog to ensure that the "well-known" styles are always
      * around.
      */
-    protected void initializeStyles(Catalog catalog, XStreamPersister xp) throws IOException {
+    protected void initializeDefaultStyles(Catalog catalog, XStreamPersister xp)
```
bradh: Travis is failing. Looks like this change also needs to go into https://github.com/geoserver/geoserver/blob/master/src/community/jms-cluster/jms-geoserver/src/main/java/org/geoserver/config/ReadOnlyGeoServerLoader.java#L61

PR #3234, opened 2018-11-08
[GEOS-9003] Fixed exception building KML network links for requests containing a layer group and CQL filter.

sikeoka:
https://osgeo-org.atlassian.net/browse/GEOS-9003


src/wms/src/main/java/org/geoserver/wms/WMSRequests.java
```diff
@@ -434,6 +437,30 @@ private static String getLayerStyles(Layer[] layers) {
         return params;
     }
 
+    @SuppressWarnings("unchecked")
+    private static int getRawLayerIndex(GetMapRequest req, int layerIndex) {
+        List<String> names = KvpUtils.readFlat(req.getRawKvp().get("layers"));
+        if (names.size() != req.getLayers().size()) {
+            // Finds the index in the raw layers KVP parameter that corresponds
```
aaime: This should be the javadoc of this method. Even if not public, it does not hurt to see what the method does without checking its code. Maybe expand a bit the explanation, I know what WMS does but I still had to open the code in the IDE to figure out what was going on.

src/wms/src/main/java/org/geoserver/wms/WMSRequests.java
```diff
@@ -309,6 +311,7 @@ private static String getLayerStyles(Layer[] layers) {
                     }
                 }
             }
+            index = getRawLayerIndex(req, index);
```
aaime: Thinking out loud, would it be better to have the index be -1 by default, and have it throw an exception if the layer is not found?

sikeoka: I'm not sure about the -1 but an exception is now thrown if it can't find the correct index.

src/wms/src/main/java/org/geoserver/wms/WMSRequests.java
sikeoka: I added javadoc and more comments to this method.

aaime:
Merged, not backported. You can do backport PRs in case you need them? If not let me know, the ticket should be closed.

PR #3236, opened 2018-11-09
[GEOS-9007] Fixed the KML output format ignoring the sortBy parameter when querying the records for a vector layer.

sikeoka:
https://osgeo-org.atlassian.net/browse/GEOS-9007


src/kml/src/test/java/org/geoserver/kml/KMLTest.java
```diff
@@ -180,6 +180,48 @@ public void testVectorWithFeatureId() throws Exception {
...
+                        "wms?request=getmap&service=wms&version=1.1.1"
+                                + "&format="
+                                + KMLMapOutputFormat.MIME_TYPE
+                                + "&layers="
+                                + getLayerId(MockData.ROAD_SEGMENTS)
+                                + "&styles=&height=1024&width=1024&bbox=-180,-90,180,90&srs=EPSG:4326"
+                                + "&sortBy=FID A");
+
+        // print(doc);
+        assertXpathEvaluatesTo("5", "count(//kml:Placemark)", doc);
```
aaime: Got stumped a bit here, was surprised to see sorting on a field, FID, but checks on the fature id, the found the fids and the feature ids happen to be following the same order:

```
RoadSegments.1107532045088=MULTILINESTRING ((-0.0042 -0.0006, -0.0032 -0.0003, -0.0026 -0.0001, -0.0014 0.0002, 0.0002 0.0007))|102|Route 5
RoadSegments.1107532045089=MULTILINESTRING ((0.0002 0.0007, 0.0014 0.001, 0.0028 0.0014))|103|Route 5
RoadSegments.1107532045089=MULTILINESTRING ((0.0028 0.0014, 0.003 0.0024))|104|Route 5
RoadSegments.1107532045090=MULTILINESTRING ((0.0002 0.0007, 0.0014 0.001, 0.0028 0.0014, 0.0042 0.0018))|105|Main Street
RoadSegments.1107532045091=MULTILINESTRING ((-0.0014 -0.0024, -0.0014 0.0002))|106|Dirt Road by Green Forest
````

Is FID not available to test against in the KML output (can't remember if we dump the attributes by default or not). I'd add at least a clarification as to why the test is actually checking the right thing

sikeoka: I added comments to clarify what the unit tests are checking.

aaime:
Merged, not backported. You can do backport PRs in case you need them? If not let me know, the ticket should be closed.

PR #3237, opened 2018-11-11
Don't use 10000 points per side to transform envelopes in GWC, the default should do

aaime:


aaime:
Build failed in Travis due to a timeout fetching the GeoGig dependencies, otherwise it would have passed. 
Restarted it to give it another try, but other builds today failed the same way, reported the issue on list.

aaime:
Jira created, fixed commit message, and manually merged, closing

PR #3239, opened 2018-11-15
[GEOS-9013] Delegate filter execution matching multiple same-xpath feature chained attribute names to database: Integration test

fernandor777:
Currently app-schema plugin detect two posible attributes with same xpath and delegate filter execution to JVM runtime. Fix delegates filter execution to database using all available join-fields.
This PR adds integration test.

Depends on:
https://github.com/geotools/geotools/pull/2173

Ticket:
https://osgeo-org.atlassian.net/browse/GEOS-9013


nmco:
As being in stand by for too long, closing it.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/SameNestedAttributeTest.java
```diff
@@ -0,0 +1,63 @@
+/* (c) 2018 Open Source Geospatial Foundation - all rights reserved
```
nmco: This shoudl be 2019.

nmco:
Looks good to me, since this touch SQL generation code, where both PostGIS and Oracle tests run?

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/SameNestedAttributeTest.java
fernandor777: Fixed, thanks.

fernandor777:
Yes, both databases tested.

nmco:
This PR only contributes integration tests merging it.

PR #3240, opened 2018-11-15
[GEOS-9014] KML layergroup styles broken by layer sequence bug

johngschulz:
This PR removes the Sequence interface, and replaces instances with Iterator and fixes double iteration of layers in SequenceList.java


src/kml/src/main/java/org/geoserver/kml/sequence/AbstractFolderSequenceFactory.java
```diff
@@ -34,20 +35,20 @@ public AbstractFolderSequenceFactory(KmlEncodingContext context) {
         this.decorators = context.getDecoratorsForClass(Folder.class);
     }
 
-    public abstract class AbstractFolderGenerator implements Sequence<Feature> {
-        protected int i = 0;
-
+    public abstract class AbstractFolderGenerator implements Iterator<Feature> {
         protected int size;
 
+        Iterator iterator;
```
tbarsballe: layerIterator would be a more clear name

aaime:
I don't remember why Sequence was adopted instead of Iterator, but don't see where the actual bug fix is? Do you have one sequence that would return a legit "null" value, and if so, why does it do that?
Also, the refactor breaks all nomenclature in the module, if you remove Sequence and replace it with Iterator, then everything that contains Sequence in the name should be also modified.

src/kml/src/main/java/org/geoserver/kml/sequence/SequenceList.java
```diff
@@ -19,18 +19,15 @@
...
 
-        private Sequence<T> generator;
+        private Iterator<T> generator;
 
-        private T item;
-
-        public GeneratorIterator(Sequence<T> generator) {
+        public GeneratorIterator(Iterator<T> generator) {
             this.generator = generator;
-            this.item = generator.next();
```
tbarsballe: @aaime - the actual bugfix is here; previously the invocation of next() in the constructor would pre-emptively modify the KML context, leading to off-by-one errors on anything that referenced the current layer via the context.

aaime: Makes sense. As said above, if Sequence goes the way of the dodo, the other classes need to be renamed too.

johngschulz: I refactored everything, including the sequence package, renaming to iterator as appropriate. Confirmed the test passes in this branch and fails in master as well. Thanks for the notes let me know if anything else needs to be addressed.

PR #3242, opened 2018-11-16
Fixed YSLD doc for priority labeling.

sikeoka:
Changed "x-label-priority" and "x-labelPriority" in the YSLD documentation to "priority".  Also, fixed an instance where "x-max-displacement" was being used instead of "x-maxDisplacement".


sikeoka:
This pull request is only minor doc changes so I'm just going to merge it.

PR #3244, opened 2018-11-17
GEOS-9018: Autocomplete support for SLD 1.0 in style editor

aaime:
@tbarsballe I've tried to generate the schema structure for SLD 1.1 using http://q42jaap.github.io/xsd2codemirror/ but hit a few dead ends, the tool does not seem to be able to handle some looping data structure in GML 3.
In the end few people use SLD 1.1, so I suggest we start with SLD 1.0 and see if someone wants to pitch in some working hours/days to support SLD 1.1 too.
In the last commit I've added a bit checking for the version attribute, if present it checks for the value, if not 1.0.0 then completion is not provided (to avoid luring SLD 1.1 users down the wrong path).


aaime:
In it goes then!

PR #3245, opened 2018-11-18
fix REST api urls in the docs

hishamkaram:


aaime:
Thank you @hishamkaram 
Hi @tbarsballe this seems to fix the issue of S3 handling (checked, the change looks very reasonable to me).
However, I'm wondering about the doc versioning, does that mean we need a different URL in there for each branch? Wondering if it would be possible to have a relative URL instead, so that we don't have to fix all conf.py.
It seems to me the issue was there even before the switch to S3, with docs pointing to a single location for the API yamls.

PR #3247, opened 2018-11-20
[GEOS-9024] Expose equal area classification method in SLDService

aaime:
Integrates the equal area classification in SLDService. There are a few more changes fixing minor issues that I've found while adding equal area, I've left out larger changes for other commits.

Here is an example request:

http://localhost:8080/geoserver/rest/sldservice/nurc:world/classify.xml?attribute=POP_CNTRY&method=equalArea&intervals=5&open=true&fullSLD=true&ramp=jet

and the map generated out of it (from the style editor):

selezione_437

Of course antartica is huge and there is zero people, so maybe not the best usage of this classification approach (polygons should have relatively similar area for it to work the best)


src/extension/sldService/src/main/java/org/geoserver/sldservice/rest/ClassifierController.java
```diff
@@ -109,26 +103,6 @@ public void configurePersister(XStreamPersister persister, XStreamMessageConvert
         xstream.allowTypes(new Class[] {RulesList.class, JSONObject.class});
     }
 
-    /**
```
aaime: I've removed commented out code sections here and there

src/extension/sldService/src/main/java/org/geoserver/sldservice/utils/classifier/RulesBuilder.java
```diff
@@ -79,27 +78,18 @@ public void setIncludeStrokeForPoints(boolean includeStrokeForPoints) {
...
     }
 
-    /**
-     * Generate a List of rules using quantile classification Sets up only filter not symbolizer
-     *
-     * @param features
-     * @param property
-     * @param classNumber
-     */
-    public List<Rule> quantileClassification(
```
aaime: Much code was duplicated, I've shared it in a single method

src/extension/sldService/src/test/java/org/geoserver/sldservice/rest/ClassifierTest.java
```diff
@@ -23,37 +23,42 @@
 import org.geotools.filter.function.FilterFunction_parseDouble;
 import org.geotools.styling.PointSymbolizer;
 import org.geotools.styling.Rule;
-import org.junit.Before;
 import org.junit.Test;
 import org.springframework.mock.web.MockHttpServletResponse;
 import org.w3c.dom.Document;
 
-public class ClassifierTestSupport extends SLDServiceBaseTest {
+public class ClassifierTest extends SLDServiceBaseTest {
```
aaime: This test is a true unit one, does not need to be a GeoServerSystemTest subclass

src/extension/sldService/src/test/java/org/geoserver/sldservice/rest/ColorRampTest.java
```diff
@@ -16,7 +16,7 @@
 import org.geoserver.sldservice.utils.classifier.impl.RedColorRamp;
 import org.junit.Test;
 
-public class ColorRampTest extends SLDServiceBaseTest {
+public class ColorRampTest {
```
aaime: This test is a true unit one, does not need to be a GeoServerSystemTest subclass

src/extension/sldService/src/test/java/org/geoserver/sldservice/rest/SLDServiceBaseTest.java
```diff
@@ -33,23 +27,12 @@
 
 public abstract class SLDServiceBaseTest extends CatalogRESTTestSupport {
 
-    protected Map<String, Object> attributes = new HashMap<String, Object>();
```
aaime: Removed dead fields

src/extension/sldService/src/test/java/org/geoserver/sldservice/rest/ClassifierTest.java
aaime: Also, name must end with "Test", otherwise maven would not run this class (this is the reason why removing the XOM dependency did not make any difference)

PR #3248, opened 2018-11-20
[GEOS-9025] When reprojecting a filter in WFS, if no native CRS is available the declared one should be used

nmco:
**Depends on GeoTools PR: https://github.com/geotools/geotools/pull/2177**

Associated issue:
https://osgeo-org.atlassian.net/browse/GEOS-9025

Note, this is different from what the existing ``DefaultCRSFilterVisitor`` does, which sets the default CRS in all geometries of the filter that don't have a CRS defined. This will make the layer declared CRS to be used for the re-projection if no native CRS is available. This changes are backward compatible.


src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java
```diff
@@ -1229,7 +1229,21 @@ void encodeQueryAsKvp(
         Filter transformedFilter = filter;
         if (declaredCRS != null) {
             transformedFilter =
-                    WFSReprojectionUtil.normalizeFilterCRS(filter, source.getSchema(), declaredCRS);
+                    WFSReprojectionUtil.normalizeFilterCRS(
+                            filter, source.getSchema(), declaredCRS, declaredCRS);
+        } else {
+            // let's try to sue the feature type info CRS
```
aaime: Don't think we want to "sue" anyone here? :-p
Maybe "use"

aaime: I'd also add a comment that this happens for example with complex features

src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java
```diff
@@ -1229,7 +1229,21 @@ void encodeQueryAsKvp(
...
+            FeatureTypeInfo featureTypeInfo =
+                    catalog.getFeatureTypeByName(
+                            primaryTypeName.getPrefix(), primaryTypeName.getLocalPart());
+            if (featureTypeInfo != null && featureTypeInfo.getCRS() != null) {
+                // the feature type info has a CRS defined, so let's use it
+                declaredCRS =
```
aaime: So here you're looking at handling axis flipping, but it does not happen in the path above?

src/wfs/src/test/java/org/geoserver/wfs/WFSReprojectionUtilTest.java
```diff
@@ -4,19 +4,64 @@
...
     @Test
     public void testNullCheck() {
         // used to throw an NPE if the feature type was null
         CoordinateReferenceSystem crs =
                 WFSReprojectionUtil.getDeclaredCrs((FeatureType) null, "1.1.0");
         assertNull(crs);
     }
+
+    @Test
+    public void testReprojectFilterWithTargetCrs() throws FactoryException {
```
aaime: This looks good, but would it make sense to also have a test covering app-schema?
Axis flipping wise, would be interesting to see WFS 1.0 vs WFS 2.0, but I'm guessing one cannot do app-schema with WFS 1.0 (all app-schemas being based on GML3... or not?)

src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java
nmco: :P thx.

src/wfs/src/test/java/org/geoserver/wfs/WFSReprojectionUtilTest.java
nmco: You are correct, I should have added a specific App-Schema test for this, doing it. Regarding GML2, well in theory it shoudl work ... but we don't have any test that uses GML2 and I never saw an use case actually using GML2, so I will let it aside.

src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java
aaime: Makes sense

aaime:
Hum... compile failure in Travis?

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:2.3.2:compile (default-compile) on project gs-wfs: Compilation failure
[ERROR] /home/travis/build/geoserver/geoserver/src/wfs/src/main/java/org/geoserver/wfs/WFSReprojectionUtil.java:[99,44] error: constructor ReprojectingFilterVisitor in class ReprojectingFilterVisitor cannot be applied to given types;
[ERROR] -> [Help 1]
```

aaime:
Re-kicked it yesterday so that it would pick the latest geotools build, but if still fails, just in a different way:

````
Tests run: 2, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 1.765 sec <<< FAILURE!
testFullRequest(org.geoserver.test.CSVOutputFormatTest)  Time elapsed: 1655 sec  <<< FAILURE!
org.junit.ComparisonFailure: expected:<[text/csv]> but was:<[application/xml]>
	at org.junit.Assert.assertEquals(Assert.java:115)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at org.geoserver.test.CSVOutputFormatTest.testFullRequest(CSVOutputFormatTest.java:42)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
````

PR #3251, opened 2018-11-22
[GEOS-9028] Expose SLDService classification methods for raster data too

aaime:
Exposes the same services as for vector data on raster layers.
First implementation is simple and uses existing methods in GeoTools, won't work properly on very large rasters, but it's suitable for smaller ones. A subsequent ticket will handle larger cases (by a combination of subsampling, area reduction, and histogram calculations)


src/extension/sldService/src/main/java/org/geoserver/sldservice/rest/ClassifierController.java
```diff
@@ -347,231 +346,272 @@ private JSONObject jsonRule(Object obj) {
...
+        }
+        ColorMap colorMap;
+        try {
+            if (customClasses.isEmpty()) {
+                if ("equalInterval".equals(method)) {
+                    colorMap =
+                            builder.equalIntervalClassification(image, intervals, open, continuous);
+                } else if ("uniqueInterval".equals(method)) {
+                    colorMap = builder.uniqueIntervalClassification(image, intervalsForUnique);
+                } else if ("quantile".equals(method) || "equalArea".equals(method)) {
```
mbarto: So... equalArea for rasters is identical to quantile classification?

aaime: Indeed it is... one pixel, one value, one unit of area, so counting repeated values and their area becomes the same.

mbarto: wonderful

PR #3255, opened 2018-11-23
[2.13.x][JIRA GEOS-9031][Backup & Restore] Add the possibility of filtering against Store and Layer

afabiani:
image

image


mbarto:
When you filter on a specific layer I would:

 * only backup the necessary data to restore that particular layer:
   * layer.xml  + featuretype.xml or coverage.xml
   * datastore.xml (to be used during restore only if the related store does not exist)
   * workspace.xml / namespace.xml (to be used during restore only if the related workspace does not exist)
   * styles referenced by the layer (no assets, unless we can make the backup code smart to read styles and list the assets used by the style)
 * I would not save anything outside of the workspaces folder (unless we can make the backup smart and guess the data used by shapefiles or coverage stores); eventually we can introduce a flag to allow a backup of minimal cfg data, without linked assets
 * during restore, we should guess what we need to restore reading the archive, not by asking the user to set filters (think of the scenario where you want to restore a layer that has been removed in the meantime)

PR #3259, opened 2018-11-24
[GEOS-9032] Use OGC scale computation for vector tiles

aaime:


bradh:
@aaime Do you still need this branch? PR is closed with an unmerged commit.

PR #3260, opened 2018-11-25
[GEOS-9033] Allow configuring services on a per layer basis

fernandor777:
GeoServer currently automatically publishes layers on all suitable services. In order to selectively publish a layer onto specific services, one has to resort to either GeoFence or a custom security manager.

However, deciding a layer service availability is not always a security matter, in those cases setting up a security system just for that is overly complicated. It would be easier to have this configuration directly in the layer "publishing" tab instead.

Full proposal:
https://github.com/geoserver/geoserver/wiki/GSIP-172---Allow-configuring-services-on-a-per-layer-basis

JIRA ticket:
https://osgeo-org.atlassian.net/browse/GEOS-9033


src/extension/wps/wps-core/src/main/java/org/geoserver/wps/WPSResourceVoter.java
```diff
@@ -0,0 +1,21 @@
...
+package org.geoserver.wps;
+
+import org.geoserver.catalog.ResourceInfo;
+import org.geoserver.catalog.ServiceResourceVoter;
+
+public class WPSResourceVoter implements ServiceResourceVoter {
+
+    @Override
+    public boolean hideService(ResourceInfo resource) {
+        return true;
```
aaime: Why so? WPS is a legitimate user of local layers, and it does so by avoiding making other requests, e.g., when the Execute looks like it's doing a WFS request, it's in fact avoiding the OGC machinery and goes directly to the WFS service implementation, without going thought the disptacher

src/extension/csw/core/src/main/java/org/geoserver/csw/CSWResourceVoter.java
```diff
@@ -0,0 +1,21 @@
+/* (c) 2018 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.csw;
+
+import org.geoserver.catalog.ResourceInfo;
+import org.geoserver.catalog.ServiceResourceVoter;
+
+public class CSWResourceVoter implements ServiceResourceVoter {
```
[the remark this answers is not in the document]
aaime: Ah I see now, the implementation of the the service lookup is backwards and would require adding a voter to each an every service, which is not what we want, there might be services developed outside of GeoServer which would be broken by this change. See my comment on the utility class doing the voting.

src/main/src/main/java/org/geoserver/catalog/ServiceResourceUtil.java
```diff
@@ -0,0 +1,36 @@
...
+    public static final String BEAN_ID = "serviceResourceUtil";
+
+    protected ApplicationContext context;
+
+    public ServiceResourceUtil() {}
+
+    public List<String> getLayerVotedServices(ResourceInfo resource) {
+        List<ServiceResourceVoter> voters =
+                GeoServerExtensions.extensions(ServiceResourceVoter.class, context);
+        return voters.stream()
```
aaime: This implementation is a compatibility issue, if a service implementation exists out of GeoServer it will stop working without a voter. Please get a list of all the services from the catalog and have the voters vote on the combination of services and layer (see also comment about ServiceResorceVoter)

src/main/src/main/java/org/geoserver/catalog/ServiceResourceVoter.java
```diff
@@ -0,0 +1,24 @@
...
+package org.geoserver.catalog;
+
+/**
+ * Allows to verify if a given layer is suitable for publishing on a given service. A single
+ * negative vote will make the service disappear from the selective service layer UI
+ *
+ * @author Andrea Aime, Geosolutions
+ * @author Fernando Mino, Geosolutions
+ */
+public interface ServiceResourceVoter {
```
aaime: This interface has been changed compared to the proposal, without a public discussion. This is problematic, as the proposal has been voted. Please use it in the way it has been discussed.

aaime: Also, when going back to the original design, you'll need to add a GWC voter I believe

aaime:
Oh, another important bit, the user documentation should be updated.

src/main/src/main/java/org/geoserver/security/DisabledServiceResourceFilter.java
```diff
@@ -0,0 +1,72 @@
...
+                disabledServices =
+                        CollectionUtils.isEmpty(resource.getDisabledServices())
+                                ? Collections.emptyList()
+                                : resource.getDisabledServices();
+            } else {
+                // service configuration disabled, get global env default disabled services list
+                disabledServices = defaultDisabledServices();
+            }
+            // if any disabled service match with current service -> hide resource
+            String service = request().getService();
```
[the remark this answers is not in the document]
fernandor777: In isFilterSubject() method we already checked that not null values are present on requets() or getService()

src/wfs/src/main/java/org/geoserver/wfs/WFSResourceVoter.java
```diff
@@ -0,0 +1,25 @@
...
+
+import org.geoserver.catalog.FeatureTypeInfo;
+import org.geoserver.catalog.ResourceInfo;
+import org.geoserver.catalog.ServiceResourceVoter;
+
+public class WFSResourceVoter implements ServiceResourceVoter {
+
+    @Override
+    public boolean hideService(ResourceInfo resource) {
+        if (resource instanceof FeatureTypeInfo) {
```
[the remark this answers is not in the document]
fernandor777: Changed to return !(resource instanceof FeatureTypeInfo);

fernandor777:
Thanks Andrea for you review. I did changes in base of it and services filter in layer bases added to Layer Preview Page.

aaime:
Ah, the build is failing due to changes in GeoTools, you should rebase the branch against master

fernandor777:
Service configuration on Layer added to User documentation, fixed WMS/WFS select groups label on service disabled and WMTS omitted from available services.

doc/en/user/source/data/webadmin/layers.rst
```diff
@@ -216,6 +216,19 @@ Cache parameters that apply to the HTTP response from client requests.
 
 * **Response Cache Headers**— If selected, GeoServer will not request the same tile twice within the time specified in :guilabel:`Cache Time`. One hour measured in seconds (3600), is the default value for :guilabel:`Cache Time`.
 
+Services Settings
+^^^^^^^^^^^^^^^^^
+
+Sets services configuration on layer level.
```
aaime: Does not describe the system variable

src/main/src/main/java/org/geoserver/catalog/ServiceResourceProvider.java
```diff
@@ -0,0 +1,64 @@
...
+            ResourceInfo resource, String serviceName, List<ServiceResourceVoter> voters) {
+        return voters.stream().anyMatch(v -> v.hideService(serviceName, resource));
+    }
+
+    private List<String> servicesList() {
+        return geoServer
+                .getServices()
+                .stream()
+                .map(si -> si.getName())
+                .filter(n -> !"WMTS".equalsIgnoreCase(n))
```
aaime: Works, I'd add a comment explaining why. Alternative, make a GWCResourceServiceVoter (but even in that case, explain why it's doing what it's doing).

doc/en/user/source/data/webadmin/layers.rst
fernandor777: Added a note describing system variable

src/main/src/main/java/org/geoserver/catalog/ServiceResourceProvider.java
fernandor777: Added GWCResourceServiceVoter  extension class, thanks

aaime:
Looking good, travis is happy, merging.

PR #3261, opened 2018-11-25
[JIRA GEOS-9031][Backup & Restore] Add the possibility of filtering against Store and Layer

afabiani:


src/security/ldap/pom.xml
```diff
@@ -40,6 +40,16 @@
       <groupId>org.springframework.ldap</groupId>
       <artifactId>spring-ldap-test</artifactId>
       <scope>test</scope>
+      <exclusions>
```
mbarto: Why do we need this?

afabiani: The "ldap-tests" module includes a very old version of Spring-Batch which is incompatible with the version used by the Backup-&-Restore module. You might end up with a method clashing and odd errors like "method not found" on Classes.  In some circumstances, it also prevents GeoServer to start.

mbarto: Good, merging then. I see you already backported to 2.13.x, can you also backport for 2.14.x?

afabiani: Sure

PR #3274, opened 2018-11-29
Assorted fixes for SLDService

aaime:
Fixing three issues in SLDService

### DIFF
--- a/src/extension/sldService/src/main/java/org/geoserver/sldservice/utils/classifier/RulesBuilder.java
+++ b/src/extension/sldService/src/main/java/org/geoserver/sldservice/utils/classifier/RulesBuilder.java
@@ -6,7 +6,7 @@
  */
 package org.geoserver.sldservice.utils.classifier;
 
-import java.awt.Color;
+import java.awt.*;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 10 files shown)


src/extension/sldService/src/main/java/org/geoserver/sldservice/utils/classifier/RulesBuilder.java
```diff
@@ -382,33 +371,22 @@ public StyleFactory getStyleFactory() {
                 } else {
                     f =
                             FF.and(
-                                    FF.greater(att, FF.literal(groups.getMin(i))),
-                                    FF.lessOrEqual(att, FF.literal(groups.getMax(i))));
+                                    FF.greaterOrEqual(att, FF.literal(groups.getMin(i))),
```
mbarto: So you changed the rules from >min AND <=max to >=min AND < max, I suppose is a fix, can you explain quickly?

aaime: Explanation at https://osgeo-org.atlassian.net/browse/GEOS-9023 , basically the previous setup was breaking the allocation of items in quantile/jenks/equal area classifications, moving items one class up and breaking the balancing of the classes. Most visible when there are few records to categorize, difficult to notice when there are many.

mbarto: Understood. +1 to merge

PR #3279, opened 2018-11-30
[GEOS-9044] Allow modifying blobstore without changing name via UI

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-9044


tbarsballe:
Backported to 2.13.x and 2.14.x

PR #3281, opened 2018-12-01
[GEOS-9045] Vector tiles module: avoid using TopologyPreservingSimplifier for lines

aaime:
Topology preservation rules are there for polygons, but I don't see any for lines


DBlasby:
This is fine.

There are some possible topology issues with simplifying lines (like causing them to cross when they didn't cross or visa-versa).  However, I don't think this would actually be a problem for 99.99% of users.

PR #3282, opened 2018-12-04
GEOS-9033 Allow configuring services on a per layer basis - Layer Group preview fix

fernandor777:
This PR fix a bug related to layer groups preview hiding WMS and WFS preview links since Resource doesn't exists for them.


aaime:
Thanks!

PR #3283, opened 2018-12-04
[GEOS-9046] Remove unnecessary check for ready.

mtoonen:
See https://osgeo-org.atlassian.net/browse/GEOS-9046


mtoonen:
Yes, created them for the 2.13.x and 2.14x branch, was that ok?

PR #3288, opened 2018-12-05
Follow up on GEOT-6198

aaime:
Depends on https://github.com/geotools/geotools/pull/2203 and a jai-ext release


src/pom.xml
```diff
@@ -1948,7 +1948,7 @@
   <wicket.version>7.6.0</wicket.version>
   <ant.version>1.8.4</ant.version>
   <imageio-ext.version>1.2.0</imageio-ext.version>
-  <jaiext.version>1.1.4</jaiext.version>
+  <jaiext.version>1.1-SNAPSHOT</jaiext.version>
```
dromagnoli: we may want to update to 1.1.next once new jai-ext is released

aaime: Yep, that's why I said it depends on a jai-ext release :-)

PR #3292, opened 2018-12-10
GEOS-9051 Printing plugin upgrade version of JTS from vividsolutions to locationtech

tixvieira:
Printing plugin upgrade version of JTS from vividsolutions to locationtech.

Mapfish print 2.1.4 released:
https://github.com/mapfish/mapfish-print/pull/772

### DIFF
--- a/src/extension/printing/pom.xml
+++ b/src/extension/printing/pom.xml
@@ -55,7 +55,7 @@
     <dependency>
       <groupId>org.mapfish.print</groupId>
       <artifactId>print-lib</artifactId>
-      <version>2.1.3</version>
+      <version>2.1.4</version>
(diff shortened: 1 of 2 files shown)


tixvieira:
Mapfish 2.1.4 on maven repo:
http://central.maven.org/maven2/org/mapfish/print/print-lib/2.1.4/

aaime:
I don't want to intrude in matters of module maintenance, but the PR should be made against master too :)

tixvieira:
> I don't want to intrude in matters of module maintenance, but the PR should be made against master too :)

PR Created:
https://github.com/geoserver/geoserver/pull/3293

PR #3295, opened 2018-12-12
GEOS-9071 Implemented usage of transformations for vector tiles

ManuelB:
Hi guys,
I wanted to use point stacker for an vector tile layer. I implemented the following code to do so.

How can I make sure that it get integrated into geoserver?

Further is it possible to get it as a backport for version 2.14?

Thanks a lot
    Manuel

### DIFF
--- a/src/extension/vectortiles/src/main/java/org/geotools/renderer/lite/VectorMapRenderUtils.java
+++ b/src/extension/vectortiles/src/main/java/org/geotools/renderer/lite/VectorMapRenderUtils.java
@@ -467,11 +467,14 @@ public class VectorMapRenderUtils {
 
         List<LiteFeatureTypeStyle> featureTypeStyles =
                 VectorMapRenderUtils.getFeatureStyles(layer, paintArea, mapScale, schema);
-        if (!featureTypeStyles.isEmpty() && featureTypeStyles.get(0).transformation != null) {
-            Expression transform = featureTypeStyles.get(0).transformation;
-            return (FeatureCollection<?, ?>) transform.evaluate(features);
-        } else {
-            return features;
+        if (!featureTypeStyles.isEmpty()) {
+            for (LiteFeatureTypeStyle featureTypeStyle : featureTypeStyles) {
+                if (featureTypeStyle.transformation != null) {
+                    Expression transform = featureTypeStyle.transformation;
+                    features = (FeatureCollection<?, ?>) transform.evaluate(features);
+                }
+            }
         }
+        return features;
     }
 }
--- a/src/extension/vectortiles/src/main/java/org/geotools/renderer/lite/VectorMapRenderUtils.java
+++ b/src/extension/vectortiles/src/main/java/org/geotools/renderer/lite/VectorMapRenderUtils.java
@@ -469,9 +469,15 @@ public class VectorMapRenderUtils {
                 VectorMapRenderUtils.getFeatureStyles(layer, paintArea, mapScale, schema);
         if (!featureTypeStyles.isEmpty()) {
             for (LiteFeatureTypeStyle featureTypeStyle : featureTypeStyles) {
-                if (featureTypeStyle.transformation != null) {
-                    Expression transform = featureTypeStyle.transformation;
-                    features = (FeatureCollection<?, ?>) transform.evaluate(features);
+                try {
+                    if (featureTypeStyle.transformation != null) {
+                        Expression transform = featureTypeStyle.transformation;
+                        features = (FeatureCollection<?, ?>) transform.evaluate(features);
+                    }
+                } catch (Exception ex) {
+                    if (LOGGER.isLoggable(Level.WARNING)) {
+                        LOGGER.log(Level.WARNING, "Error while appliying transformation", ex);
+                    }
                 }
             }
         }
--- a/src/extension/vectortiles/src/main/java/org/geotools/renderer/lite/VectorMapRenderUtils.java
+++ b/src/extension/vectortiles/src/main/java/org/geotools/renderer/lite/VectorMapRenderUtils.java
@@ -19,6 +19,8 @@ import org.geotools.data.DataUtilities;
 import org.geotools.data.FeatureSource;
 import org.geotools.data.Query;
 import org.geotools.factory.CommonFactoryFinder;
+import org.geotools.factory.Hints;
+import org.geotools.feature.FeatureCollection;
 import org.geotools.feature.FeatureTypes;
 import org.geotools.filter.IllegalFilterException;
 import org.geotools.filter.spatial.DefaultCRSFilterVisitor;
@@ -41,6 +43,7 @@ import org.opengis.feature.type.FeatureType;
 import org.opengis.feature.type.GeometryDescriptor;
 import org.opengis.filter.Filter;
 import org.opengis.filter.FilterFactory2;
+import org.opengis.filter.expression.Expression;
 import org.opengis.referencing.FactoryException;
 import org.opengis.referencing.crs.CoordinateReferenceSystem;
 import org.opengis.referencing.operation.MathTransform;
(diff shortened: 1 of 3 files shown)


aaime:
Hi, I'm not the module maintainer, the PR should be reviewed by @DBlasby  or @tbarsballe , however, 
for the sake of getting this started, I'm going to add a few generic comments.

First, the [contribution](https://github.com/geoserver/geoserver/blob/master/CONTRIBUTING.md) guidelines require a CLA to be signed and sent to OSGEO, did you already do that, or work for a company that has signed one?

Going to add a quick comment on the code too.

src/extension/vectortiles/src/main/java/org/geotools/renderer/lite/VectorMapRenderUtils.java
```diff
@@ -443,4 +445,33 @@ private static boolean isFeatureTypeStyleActive(FeatureType ftype, FeatureTypeSt
...
+            Layer layer,
+            Rectangle paintArea,
+            double mapScale,
+            FeatureType schema,
+            FeatureCollection<?, ?> features)
+            throws IOException {
+
+        List<LiteFeatureTypeStyle> featureTypeStyles =
+                VectorMapRenderUtils.getFeatureStyles(layer, paintArea, mapScale, schema);
+        if (!featureTypeStyles.isEmpty() && featureTypeStyles.get(0).transformation != null) {
```
aaime: A layer can have multiple FTS attached and one can be transformed while the others are not, or have different transformations per FTS. 
If I understand how the module works (mind, not the original creator nor its maintainer) for this case to be supported the layer should be split in multiple ones, I'm guessing by checking the FTSs and grouping them by uniform transform.

ManuelB: @aaime I agree that it might be possible that there are multiple transformation. Nevertheless the patch already enhances the behavior of geoserver a  lot. Currently the transformations are not applied at all.

aaime: It's not up to me to decide if that's ok or not, that's the responsibility of the module maintainer. If it's ok to be included as is, I'd add a small note in the documentation that the RT must be in the first FTS and others might be ignored (but again, not my call).

ManuelB: @aaime  thanks a lot for reviewing. Let's see what the maintainer says.

src/extension/vectortiles/src/main/java/org/geotools/renderer/lite/VectorMapRenderUtils.java
```diff
@@ -443,4 +445,33 @@ private static boolean isFeatureTypeStyleActive(FeatureType ftype, FeatureTypeSt
...
+            Expression transform = featureTypeStyles.get(0).transformation;
+            return (FeatureCollection<?, ?>) transform.evaluate(features);
```
tbarsballe: This should have a little more validation, so that we aren't throwing ClassCast exceptions if there is an invalid transform (for example: heatmap) in the style; at the very least some sort of exception handling with an error that indicates the transform was invalid.

src/extension/vectortiles/src/main/java/org/geotools/renderer/lite/VectorMapRenderUtils.java
tbarsballe: This should apply transforms to each FeatureTypeStyle, not just the first.

src/extension/vectortiles/src/test/java/org/geotools/renderer/lite/VectorMapRenderUtilsTest.java
```diff
@@ -0,0 +1,65 @@
...
+import org.geotools.feature.FeatureCollection;
+import org.geotools.map.Layer;
+import org.geotools.styling.FeatureTypeStyle;
+import org.geotools.styling.Rule;
+import org.geotools.styling.Style;
+import org.junit.Test;
+import org.opengis.feature.type.FeatureType;
+import org.opengis.filter.expression.Expression;
+
+public class VectorMapRenderUtilsTest {
```
tbarsballe: In addition to this unit test, and integration test using an actual style with a transfrom would be nice, although is not strictly necessary.

ManuelB:
I will open a JIRA issue and integrate the given Feedback. For now I only tested it with point stacker because this is what I need.

src/extension/vectortiles/src/main/java/org/geotools/renderer/lite/VectorMapRenderUtils.java
ManuelB: Added some exception handling

src/extension/vectortiles/src/main/java/org/geotools/renderer/lite/VectorMapRenderUtils.java
```diff
@@ -443,4 +445,42 @@ private static boolean isFeatureTypeStyleActive(FeatureType ftype, FeatureTypeSt
...
+            throws IOException {
+
+        List<LiteFeatureTypeStyle> featureTypeStyles =
+                VectorMapRenderUtils.getFeatureStyles(layer, paintArea, mapScale, schema);
+        if (!featureTypeStyles.isEmpty()) {
+            for (LiteFeatureTypeStyle featureTypeStyle : featureTypeStyles) {
+                try {
+                    if (featureTypeStyle.transformation != null) {
+                        Expression transform = featureTypeStyle.transformation;
+                        features = (FeatureCollection<?, ?>) transform.evaluate(features);
```
tbarsballe: This gets reset after each loop; if you have more then one feature type style you will end up only returning features for the last one.

You'll need to combine all the feature collections from each iteration of the loop and return that, rather than just the last one.

Additionally, this does not account for a mix of transformed and not transformed feature collections - at current, only the transformed features will be returned.

It may make more sense to handle transforms as part of the layer handling in VectorTileMapOutputFormat.java and in cases where there are a mix of transformed and not transformed collections treat each as a separate layer


Given the number of different possible configurations involved, I would also recommend a test case for each (No transformation, only a transformation, a transformation and a normal featuretypestyle, an invalid transformation).

src/extension/vectortiles/src/main/java/org/geotools/renderer/lite/VectorMapRenderUtils.java
```diff
@@ -443,4 +445,42 @@ private static boolean isFeatureTypeStyleActive(FeatureType ftype, FeatureTypeSt
...
+            Layer layer,
+            Rectangle paintArea,
+            double mapScale,
+            FeatureType schema,
+            FeatureCollection<?, ?> features)
+            throws IOException {
+
+        List<LiteFeatureTypeStyle> featureTypeStyles =
+                VectorMapRenderUtils.getFeatureStyles(layer, paintArea, mapScale, schema);
+        if (!featureTypeStyles.isEmpty()) {
```
tbarsballe: This should be unnecessary; if featureTypeStyles is empty it will simply iterate zero times

src/extension/vectortiles/src/main/java/org/geotools/renderer/lite/VectorMapRenderUtils.java
```diff
@@ -443,4 +445,42 @@ private static boolean isFeatureTypeStyleActive(FeatureType ftype, FeatureTypeSt
...
+            for (LiteFeatureTypeStyle featureTypeStyle : featureTypeStyles) {
+                try {
+                    if (featureTypeStyle.transformation != null) {
+                        Expression transform = featureTypeStyle.transformation;
+                        features = (FeatureCollection<?, ?>) transform.evaluate(features);
+                    }
+                } catch (Exception ex) {
+                    if (LOGGER.isLoggable(Level.WARNING)) {
+                        LOGGER.log(Level.WARNING, "Error while appliying transformation", ex);
+                    }
```
tbarsballe: "appliying" is misspelt - should be "applying"

Also, the error message is a little vague; it might be useful to include exactly what transformation it is trying to apply.

src/extension/vectortiles/src/main/java/org/geotools/renderer/lite/VectorMapRenderUtils.java
ManuelB: Hi,
I don't thing that it gets reset. It will just apply the transformations one after another if you do e.g. first a buffer and afterwards a union it will return all the merged polygons.
I have to admit that I did not study the SLD specification on the semantics of how transformations should be applied.

src/extension/vectortiles/src/main/java/org/geotools/renderer/lite/VectorMapRenderUtils.java
ManuelB: @tbarsballe when getting into detail what problem happens another null pointer exception might happen. From my perspective the exception itself should contain detailed information what went wrong.

src/extension/vectortiles/src/main/java/org/geotools/renderer/lite/VectorMapRenderUtils.java
tbarsballe: Thats a good enough reason to add a test case to verify what actually happens. I'm not certain myself, but the fact that you are reassigning the feature collection each iteration of the loop suggests that something is likely to get overwritten if you have multiple featureTypeStyles

src/extension/vectortiles/src/main/java/org/geotools/renderer/lite/VectorMapRenderUtils.java
tbarsballe: Thats what null checks are for. Depending on what the exception is, the error message may not be obvious. 

In particular, if there is a ClassCastException on line 475 because the Transformation did not return a FeatureCollection, then its not going to be obvious that the actual error is that the style is using an incompatible rendering transform.

ManuelB:
Sorry for the current waiting time. I will try to work on this as soon as possible.

ManuelB:
Hi @tbarsballe,
unfortunately it was decided that we are not going to do the clustering as a vector-to-vector transformation in the geoserver. So I am not allowed to spend more time on this work. Maybe somebody else is interested in finishing it.
Sorry

PR #3296, opened 2018-12-13
NetCDF out packed data fix, schema compliance fix and build speedup for WCS output modules

aaime:


src/extension/netcdf-out/src/test/java/org/geoserver/wcs2_0/WCSNetCDFTest.java
```diff
@@ -70,6 +77,10 @@ protected void onSetUp(SystemTestData testData) throws Exception {
         setupRasterDimension(getLayerId(NO2), ResourceInfo.TIME, DimensionPresentation.LIST, null);
         setupRasterDimension(
                 getLayerId(NO2), ResourceInfo.ELEVATION, DimensionPresentation.LIST, null);
+        setupRasterDimension(getLayerId(O3), ResourceInfo.TIME, DimensionPresentation.LIST, null);
+        setupRasterDimension(
+                getLayerId(O3), ResourceInfo.ELEVATION, DimensionPresentation.LIST, null);
+        setupRasterDimension(getLayerId(O3), ResourceInfo.TIME, DimensionPresentation.LIST, null);
```
dromagnoli: maybe a typo here. It looks like it has already been added at line 80

aaime:
The AppVeyor build failed reaching the 1 hour timeout, it was still downloading stuff... probably slow to fetch data from the repositories (or repositories slow to feed bytes to it).

src/extension/netcdf-out/src/test/java/org/geoserver/wcs2_0/WCSNetCDFTest.java
aaime: Yep, will remove

PR #3297, opened 2018-12-13
Fixed Hazelcast Clustering module unit test failures after JDK11 compatibility updates.

sikeoka:
This pull request combines seven changes for the hz-cluster module related to JDK 11 compatibility:
1. Upgraded Hazelcast from 3.11-BETA-1 to 3.11.1.  Someone will have to change this in the geowebcache project.
2. Upgraded Hazelcast Web Manager from 3.3.1 to 3.8.3.  This library hasn't been updated in over a year so I don't know how compatible it will be with Hazelcast 3.11.  Regarding the change to HzSessionShareFilter, the WebFilter will load the Hazelcast instance using the "instance-name" parameter in initDelegate().
3. Upgraded Metrics from 2.1.5 to 4.0.3.  Metrics 4.x.x is documented as being Java 9 compatible.
4. Added WMTS Layer/Store to fix a unit test failure in RemovedObjectProxyTest.
5. Fixed ConfigTest error on Windows and updated it to clean up its temp directories/files.
6. Refactoring to mitigate race conditions that caused unit test failures in ReloadHzSynchronizerRecvTest.  Some of these unit tests are highly dependent on multi-threaded code executing in a very specific order that for some reason worked fine through 2.14.x but stopped working as expected on 2.15.x.  The race conditions still exist but the unit tests should pass most of the time.  This appears to be a flaw in the unit test code rather than the actual ReloadHzSynchronizer code.
7. Adds hz-cluster back to the community release.


aaime:
The module does not have a maintainer, not sure who can review this PR?

sikeoka:
I could volunteer to be the maintainer for this module although it wouldn't help with finding someone else to review PRs?

aaime:
It would help, as the module maintainer can do direct commits and does not need to ask for reviews :-D

sikeoka:
What would be the process for me to become a maintainer for this module?  I could also be a maintainer for jdbcconfig.

aaime:
If you plan to leave the modules in unsupported land, I believe you just need to send a mail to geoserver-devel stating that you'd like to become maintainer of the modules. 
If you want to reap the benefits of a supported module (tests get actually run in the build!) then you'll have to make a proposal like this one and make sure every element in the list is ticked:
https://github.com/geoserver/geoserver/wiki/GSIP-170

sikeoka:
@NielsCharlier , could you take a look at this PR if you have time?  I'd like to merge this as soon as possible to get hz-cluster back in the build.

sikeoka:
I will merge this PR today if there are no additional comments?

NielsCharlier:
Sorry, I was not very available during holidays. Thanks for the fix and the review.

PR #3299, opened 2018-12-14
Update translation in GeoServerApplication_de.properties

michikommader:
fixes incomplete translation


aaime:
@fgdrf would you like to have a look?

fgdrf:
@michikommader Thanks for this fix! I'd like to suggest to have a look at Transifex project to contribute and review translations : https://www.transifex.com/GeoServer/geoserver/

Would you help to update ressources at Transifex? It would help us to maintain it only once. We synchronize ressources back to codebase once awhile.

fgdrf:
@michikommader However, would you allow us to take over this change to transifex ressource?

PR #3300, opened 2018-12-17
Documentation: fix download URL for the community modules and extensions

jgrocha:
This PR replaces references to the old server ares.opengeo.org in the documentation. The protocol is also changed from http to https.

```bash
cd doc/en/user/source
find . -type f -name '*'.rst -exec sed -i 's/http:\/\/ares.opengeo/https:\/\/build.geoserver/g' {} +
```

Afterwards, I've also:
1. changed the pointers to the geofence (server and plugin) from community to extensions,
1. fix some URLs to make them all terminating with a slash.

### DIFF
--- a/doc/en/user/source/extensions/geofence-server/installing.rst
+++ b/doc/en/user/source/extensions/geofence-server/installing.rst
@@ -1,12 +1,25 @@
 .. _geofence_server_install:
 
+GeoFence promoted to extension since version 2.15
+=================================================
+
+Geofence modules were community modules up to Geoserver version 2.14.
+Geofence modules were promoted to extensions since version 2.15
+(see `GSIP 164 - Promote geofence modules from Community to Extension <https://github.com/geoserver/geoserver/wiki/GSIP-164>`_.
+
+Up to Geoserver version 2.14, Geofence can be downloaded from `Geoserver build server <https://build.geoserver.org/geoserver/>`_.
+
 Installing the GeoServer GeoFence Server extension
 ==================================================
 
- #. Download the extension from the `nightly GeoServer community module builds <https://build.geoserver.org/geoserver/master/ext-latest/>`_.
+ #. Navigate to the `GeoServer download page <http://geoserver.org/download>`_.
 
-    .. warning:: Make sure to match the version of the extension to the version of the GeoServer instance!
+ #. Find the page that matches the version of the running GeoServer.
 
- #. Extract the contents of the archive into the ``WEB-INF/lib`` directory of the GeoServer installation.
+ #. Download the GeoFence extension. The download link will be in the :guilabel:`Extensions` section under :guilabel:`Other`.
+
+ #. Extract the files in this archive to the :file:`WEB-INF/lib` directory of your GeoServer installation.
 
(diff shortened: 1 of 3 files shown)


doc/en/user/source/extensions/mongodb/index.rst
```diff
@@ -1,40 +1,40 @@
...
-
-#. The absolute path to a data directory where GeoServer will store the schema produced for the published collections.
-
-.. figure:: images/mongodb_store_2.png
-   :align: center
-
-   *Configuring a MongoDB data store.*
-
-For more details about the usage of this data store please check the `GeoTools MongoDB plugin documentation
+.. _mongodb:
```
nmco: Not sure why this file is all marked as changed, something obvious I'm missing ? That say, it looks good.

doc/en/user/source/extensions/geofence-server/installing.rst
```diff
@@ -3,7 +3,7 @@
 Installing the GeoServer GeoFence Server extension
 ==================================================
 
- #. Download the extension from the `nightly GeoServer community module builds <http://ares.opengeo.org/geoserver/master/community-latest/>`_.
+ #. Download the extension from the `nightly GeoServer community module builds <https://build.geoserver.org/geoserver/master/ext-latest/>`_.
```
nmco: Well since GeoFence is an extension now, the text should be updated too, usually for extensions something like this is used:
```

#. Navigate to the `GeoServer download page <http://geoserver.org/download>`_.
#. Find the page that matches the version of the running GeoServer.
#. Download the GeoFence extension. The download link will be in the :guilabel:`Extensions` 
   section under :guilabel:`Other`.
#. Extract the files in this archive to the :file:`WEB-INF/lib` directory of your GeoServer installation.
#. Restart GeoServer
```

Although since GeoFence is only available in master as an extension I'm not sure where it will appear on the download page,  @afabiani any idea ?

doc/en/user/source/extensions/geofence/installing.rst
```diff
@@ -3,7 +3,7 @@
 Installing the GeoServer GeoFence extension
 ===========================================
 
- #. Download the extension from the `nightly GeoServer community module builds <http://ares.opengeo.org/geoserver/master/community-latest/>`_.
```
nmco: Same comment as above.

nmco:
Just some minor comment, thanks for the contribution @jgrocha !

doc/en/user/source/extensions/mongodb/index.rst
jgrocha: Let me check if it is the annoying LF/CRLF problem.

doc/en/user/source/extensions/geofence-server/installing.rst
jgrocha: @nmco It does not appear on previous versions. We need to point to `master/ext-latest/` right now.

jgrocha:
@nmco Thank you for your comments. I've changed the MongoDB intslattion instructions to follow the same pattern of other extensions.

I've also added a note to Geofence installation (client and server extensions) to highlight the change from community to extension.

PR #3303, opened 2018-12-19
Remove some dead or unnecessary code that causes JDK11 issues.

dvntucker:
Just a couple small changes removing Sun package references from test cases. Removed a method that is unused, and in another case removed the internal NotImplementedException for a simple RuntimeException (in code that should be unreachable anyway).


tbarsballe:
Looks good to me. Appveyor CI is failing, but looking at the failure I think it is erroneous (which is to say, I can't actually see a legitimate failure anywhere)

aaime:
Seem legit to me, since they are both in app-schema land, would be nice to get a comment from @nmco

aaime:
And oh, I've restarted the AppVeyor build.

dvntucker:
The AppVeyor looks to be passing now. The previous failure claimed to be because it took too long.

PR #3305, opened 2018-12-28
The path of delete user was wrong

himyata:


bradh:
This feels more like a bug in the implementation. Do you have a Jira ticket or test to go with this?

himyata:
**Documentation says this:** `/roles/roles/{role}`

image


**On a running app:** `/security/roles/role/{role}`

image

PR #3309, opened 2019-01-03
[GEOS-9084] Made custom dimension names in WCS GetCoverage requests case insensitive.

sikeoka:
See https://osgeo-org.atlassian.net/browse/GEOS-9084

This pull request makes custom dimension names in WCS 1.0.0 and 2.0.1 GetCoverage requests case insensitive to be consistent with time and elevation (WCS 1.1.1 has no custom dimension support).  The WCS 1.0 GetCoverageTest was modified to demonstrate that elevation is already case insensitive.  The WCS 2.0 GetCoverageTest already has tests that use both "time" and "Time" and both "elevation" and "Elevation" so there was no need to modify any tests to demonstrate that those dimension names are already case insensitive.

Rather than rejecting it as an invalid axis label, GeoServer currently returns the coverage using the default custom dimension value instead of the requested value for WCS 2.0.1 GetCoverage requests with a custom dimension name that is a case insensitive match to the actual dimension name but not a case sensitive match.


aaime:
Looks good to me, merging. Backport as you see fit.

PR #3312, opened 2019-01-07
[GEOS-9048] NPE if a Jiffle Rendering Transformation is used with Channel Selection

aaime:


aaime:
@dromagnoli here is a fun one for you... I solved the NPE reported in the bug, caused by a "band select on reader", which with a RT should not happen, as the band select should run after the RT (it's part of the symbolization, which happens later).

However, doing so, broke another test, RenderedImageMapOutputFormatExtendedTest.testRenderingTransformationChannelsSelectionFromCoverageView, because that one was running a band select and a crop coverage RT on a 4 banded view.... the bandmerge in the view reader was coming out with a RGBA color model, and then the band select done in the symbolizer preserved the alpha, resulting in 2 bands in output instead of one.

I've modified the CoverageViewReader so that it won't generate an alpha band when doing band merge... that seems like the 0.1% case, people merging random bands and one happens to be an alpha band in the right position? Nah...

PR #3313, opened 2019-01-08
[GEOS-9088] Appschema OR performance improvement on postgresql

fernandor777:
Is known that OR queries are difficult to optimize for postgresql and are usually slow.

The fix for slow queries with OR involved is using UNION queries, so we need optimize certain complex subqueries. The slow appschema subquery involved is that generated by NestedFilterToSQL class so we need to change the use of OR clause in the root condition for UNION queries.

Since this performance issue affect postgresql, the fix should be isolated for this datasource only, and adding a system property for disable.

We'll need execute online tests with postgresql and geoserver-appschema with new improvement enabled and disabled.

Depends on geotools PR:
https://github.com/geotools/geotools/pull/2220

### DIFF
--- a/doc/en/user/source/data/app-schema/joining.rst
+++ b/doc/en/user/source/data/app-schema/joining.rst
@@ -83,3 +83,27 @@ Much like joining support, native encoding of nested filters is turned on by def
 Or, alternatively, by setting the value of the Java System Property "app-schema.encodeNestedFilters" to "false", for example ::
 
      java -DGEOSERVER_DATA_DIR=... -Dapp-schema.encodeNestedFilters=false Start
+
+UNION performance improvement for OR conditions
+-----------------------------------------------
+
+OR conditions are difficult to optimize for postgresql and are usually slow.  App-Schema improves OR condition performance using UNION clauses instead OR for nested filter subqueries.
+
+With UNION improvement enabled main OR binary operator on nested filter subquery will rebuild normal OR query like::
+
+     SELECT id, name FROM table WHERE name = "A" OR name = "B"
+
+to::
+
+     SELECT id, name FROM table WHERE name = "A" UNION SELECT id, name FROM table WHERE name = "B"
+
+UNION improvement is enabled by default, and it is disabled by adding to your app-schema.properties file the line ::
+
+     app-schema.orUnionReplace = false
+	 
+Or, alternatively, by setting the value of the Java System Property "app-schema.orUnionReplace" to "false", for example ::
+
+     java -DGEOSERVER_DATA_DIR=... -Dapp-schema.orUnionReplace=false Start
+	
+.. note::
+    This optimization will only be applied when a PostgresSQL database is being used.
--- a/src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/SimpleAttributeFeatureChainWfsTest.java
+++ b/src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/SimpleAttributeFeatureChainWfsTest.java
@@ -17,6 +17,7 @@ import org.geotools.appschema.filter.FilterFactoryImplNamespaceAware;
 import org.geotools.appschema.jdbc.NestedFilterToSQL;
 import org.geotools.data.FeatureSource;
 import org.geotools.data.complex.AppSchemaDataAccess;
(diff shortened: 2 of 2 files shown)


doc/en/user/source/data/app-schema/joining.rst
```diff
@@ -83,3 +83,14 @@ Much like joining support, native encoding of nested filters is turned on by def
 Or, alternatively, by setting the value of the Java System Property "app-schema.encodeNestedFilters" to "false", for example ::
 
      java -DGEOSERVER_DATA_DIR=... -Dapp-schema.encodeNestedFilters=false Start
+
+UNION performance improvement for OR conditions
+-----------------------------------------------
+
+Is known that OR queries are difficult to optimize for postgresql and are usually slow.  App-Schema improve OR condition performance using UNION clauses instead OR for nested filter subqueries.
```
nmco: I would reword this section and provide a concrete example, this what a an OR query looks like and this is what a UNION query looks like.

doc/en/user/source/data/app-schema/joining.rst
```diff
@@ -83,3 +83,14 @@ Much like joining support, native encoding of nested filters is turned on by def
 Or, alternatively, by setting the value of the Java System Property "app-schema.encodeNestedFilters" to "false", for example ::
 
      java -DGEOSERVER_DATA_DIR=... -Dapp-schema.encodeNestedFilters=false Start
+
+UNION performance improvement for OR conditions
+-----------------------------------------------
+
+Is known that OR queries are difficult to optimize for postgresql and are usually slow.  App-Schema improve OR condition performance using UNION clauses instead OR for nested filter subqueries.
+
+This improvement is enabled by default but there is a System Property that allows you to enable/disable it: "app-schema.orUnionReplace".  For example if you want to disable the UNION improvements you should use::
```
nmco: Please use the same pattern used for the property above:

```
Much like joining support, native encoding of nested filters is turned on by default, and it is disabled by adding to your app-schema.properties file the line ::
     app-schema.encodeNestedFilters = false
Or, alternatively, by setting the value of the Java System Property "app-schema.encodeNestedFilters" to "false", for example ::

     java -DGEOSERVER_DATA_DIR=... -Dapp-schema.encodeNestedFilters=false Start
```

doc/en/user/source/data/app-schema/joining.rst
```diff
@@ -83,3 +83,14 @@ Much like joining support, native encoding of nested filters is turned on by def
...
+UNION performance improvement for OR conditions
+-----------------------------------------------
+
+Is known that OR queries are difficult to optimize for postgresql and are usually slow.  App-Schema improve OR condition performance using UNION clauses instead OR for nested filter subqueries.
+
+This improvement is enabled by default but there is a System Property that allows you to enable/disable it: "app-schema.orUnionReplace".  For example if you want to disable the UNION improvements you should use::
+
+     app-schema.orUnionReplace = false
+	
+This performance improvement is isolated for Postgresql datasources.
```
nmco: This deserves to be a note \ warning, and I would reword it: ``This optimization will only be applied when a PostgresSQL database is being used.``.

src/extension/app-schema/pom.xml
```diff
@@ -31,6 +31,7 @@
             <modules>
                 <module>app-schema-oracle-test</module>
                 <module>app-schema-postgis-test</module>
+				<module>app-schema-postgis-nounion-test</module>
```
nmco: Bad formating, too much extra spaces.

nmco:
I foreseen two major issues with ``app-schema-postgis-nounion-test`` integration tests:
- they run all the tests, and should only run only the ones that are relevant to the union optimization
- the other issue is what will happen if we runt this tests with ``mvn -T*`` they will be targeting the same database and the same tables

Does my concerns make sense ? At this point I'm thinking that the best solution maybe to select the relevant tests and make sure they run with the optimization on and with the optimization off, this means taking the tests isolate it ina generic method and then create two tests invoking it setting explicit the relevant system property, and get ride of the  ``app-schema-postgis-nounion-test``. What do you think @fernandor777 ?

Besides that I run both online tests version and they work fine :tada:

fernandor777:
Thanks @nmco for your review. I did modifications and did rewrite on testing part.

nmco:
Cool, thanks @fernandor777 !

PR #3294, opened 2018-12-12
[GEOS-9055] JDBCConfig - duplicate database index

dalbani:
https://osgeo-org.atlassian.net/projects/GEOS/issues/GEOS-9055

PR #3316, opened 2019-01-10
[GEOS-9055] JDBCConfig - duplicate database index

sikeoka:
This pull requests removes duplicate database indexes on the jdbcconfig object, type and property_type tables since PostgreSQL automatically creates indexes for primary keys.
https://www.postgresql.org/docs/current/indexes-unique.html

This pull request is related to https://github.com/geoserver/geoserver/pull/3294
https://osgeo-org.atlassian.net/projects/GEOS/issues/GEOS-9055


dalbani:
I think the duplicate index check described on https://wiki.postgresql.org/wiki/Index_Maintenance#Duplicate_indexes is probably more reliable than the one used by PgHero (https://github.com/ankane/pghero/blob/master/lib/pghero/methods/indexes.rb#L171), so I'm all for it.

PR #3318, opened 2019-01-11
GEOS-9095 csw iso metadata tag order

NielsCharlier:
https://osgeo-org.atlassian.net/browse/GEOS-9095

also fixed the wrong use of the AbstractMD_Identification tag (has to be MD_DataIdentification for layers)


tbarsballe:
While I'm not at all familiar with the CSW module, the changes look fine to me.

NielsCharlier:
Thanks!

PR #3319, opened 2019-01-11
[GEOS-9088] Appschema OR performance improvement on postgresql (2.14.x backport)

fernandor777:
Is known that OR queries are difficult to optimize for postgresql and are usually slow.

The fix for slow queries with OR involved is using UNION queries, so we need optimize certain complex subqueries. The slow appschema subquery involved is that generated by NestedFilterToSQL class so we need to change the use of OR clause in the root condition for UNION queries.

Since this performance issue affect postgresql, the fix should be isolated for this datasource only, and adding a system property for disable.

We'll need execute online tests with postgresql and geoserver-appschema with new improvement enabled and disabled.

Depends on Geotools PR:
https://github.com/geotools/geotools/pull/2225


nmco:
looks good to me.

PR #3320, opened 2019-01-11
[GEOS-9094] GetLegendGraphics JSON output generates the same link for…

ianturton:
… stacked symbols in the same rule


aaime:
Looks good, Travis status did not update here but the builds passed

PR #3322, opened 2019-01-11
[GEOS-9084][2.14.x] Made custom dimension names in WCS GetCoverage requests case insensitive.

sikeoka:
2.14.x backport for https://github.com/geoserver/geoserver/pull/3309


sikeoka:
That Travis CI build passed even though the PR status won't update.

PR #3331, opened 2019-01-18
Update Jetty deprecated properties

Gnafu:


Gnafu:
I tested the port change with the 2.15-M0 windows package and I was able to change the port with the new parameter.

jetty.http.port=8888

PR #3334, opened 2019-01-21
GEOS-9105 csw iso profile gml support

NielsCharlier:


NielsCharlier:
I'm afraid now this will only start working properly once #3343 has been merged, as it depends on the improved builder

NielsCharlier:
closed until csw module is split

PR #3335, opened 2019-01-22
[GEOS-9018] Only use Ctrl-Space to trigger autocomplete

tbarsballe:
Follow-up to https://github.com/geoserver/geoserver/pull/3244

Disables automatic autocomplete on keyup - now Ctrl-Space is the only autocomplete method

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/wicket/CodeMirrorEditor.js
+++ b/src/web/core/src/main/java/org/geoserver/web/wicket/CodeMirrorEditor.js
@@ -54,14 +54,6 @@ var editor = CodeMirror.fromTextArea(textarea, {
 });
 editor.getWrapperElement().style.fontSize = "12px"; 
 editor.refresh();
-CodeMirror.on(window, 'keyup', function(event) {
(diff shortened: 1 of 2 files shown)


aaime:
I've tried it out, seems that completion works on a brand new style with ctrl-space, but checking out existing styles, I have several where completion does not work... strange? Like for example population in the existing data dir. I see no errors in the javascript console....

aaime:
And oh, I've also checked the windows build, it works on my windows laptop, and the appveyour output does not show an error (that I can see), seems like it dies just randomly, there, in the middle of the build.

tbarsballe:
Okay, found the bug. Turns out getVersion was using a greedy regex and therefore returning `"1.0.0" xmlns="http://www.opengis.net/sld" xmlns:ogc="http://www.opengis.net/ogc"` instead of `"1.0.0"` for the population style. I've fixed the regex; should work now.

As for the Appveyor failure, I saw a couple similar failures yesterday, where it would just stop in the middle. Not sure whats going on there...

PR #3337, opened 2019-01-23
Mapbox Styles expressions documentation

tbarsballe:
GeoServer port of https://github.com/geotools/geotools/pull/2238 (retaining GeoServer-specific URLs and examples)


tbarsballe:
The appveyor failure appears to be spurious, or at least unrelated. Merging.

PR #3341, opened 2019-01-29
WMTS stylecommunity module and associated fixes

aaime:


src/community/wmts-styles/src/main/java/org/geoserver/gwc/wmts/StylesExtension.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    private final List<OperationMetadata> extraOperations = new ArrayList<>();
+
+    // {
+    //        extraOperations.add(new OperationMetadata("DescribeDomains"));
+    //        extraOperations.add(new OperationMetadata("GetDomainValues"));
+    //        extraOperations.add(new OperationMetadata("GetFeature"));
+    //        extraOperations.add(new OperationMetadata("GetHistogram"));
+    // }
```
dromagnoli: Is this intentionally left commented out?

aaime: Good catch, forgot to remove it

PR #3343, opened 2019-01-30
GEOS-8948 CSW: support Feature Catalogues

NielsCharlier:


NielsCharlier:
As part of this PR, significant improvements were made to the CSW mapping file syntax/generation process
    * to support multidimensional mappings (with multiple % in the mapping file) , since this exists in the feature catalogue
    * fix the ambivalence issue with @ meaning both id and attribute

NielsCharlier:
Some more notes
* support for feature catalogues is particularly useful with upcoming metadata module which will allow to create them using the GUI
* documentation will also be added if this feature is accepted

NielsCharlier:
closed until module split

PR #3346, opened 2019-01-31
GEOS-9117 add wcs and wmts endpoint placeholders

NielsCharlier:


src/extension/csw/core/src/test/resources/org/geoserver/csw/store/internal/MD_Metadata.properties
```diff
@@ -6,6 +6,6 @@ identificationInfo.AbstractMD_Identification.abstract.CharacterString=abstract
 $dateStamp.Date= if_then_else ( isNull("metadata.date") , 'Unknown', "metadata.date")
 hierarchyLevel.MD_ScopeCode.@codeListValue='http://purl.org/dc/dcmitype/Dataset'
 $contact.CI_ResponsibleParty.individualName.CharacterString='John Smith'
-distributionInfo.MD_Distribution.transferOptions.MD_DigitalTransferOptions.onLine.CI_OnlineResource%.linkage.URL=list('${url.wfs}','${url.wms}')
-distributionInfo.MD_Distribution.transferOptions.MD_DigitalTransferOptions.onLine.CI_OnlineResource%.protocol.CharacterString=list('OGC:WFS','OGC:WMS')
+distributionInfo.MD_Distribution.transferOptions.MD_DigitalTransferOptions.onLine.CI_OnlineResource%.linkage.URL=list('${url.wfs}','${url.wms}','${url.wcs}')
```
aaime: How about WMTS here, should it be included?

NielsCharlier: I did not add it to the default config, because it depends on whether the user is using gwc or not if they want that exposed

NielsCharlier: @aaime is this ok then?

aaime: Ah yes

NielsCharlier:
Excellent, thanks

PR #3347, opened 2019-02-01
[GEOS-9120] Assorted issues when using WMTS restful bindings in virtual workspace services

aaime:


src/gwc/src/main/java/org/geoserver/gwc/dispatch/GwcServiceDispatcherCallback.java
```diff
@@ -47,6 +48,15 @@
...
 
+    static final String buildRestPattern(int numPathElements) {
+        return ".*/service/wmts/rest" + Strings.repeat("/([^/]+)", numPathElements);
+    }
+
+    private static final Pattern TILE_1 = Pattern.compile(buildRestPattern(5));
+    private static final Pattern TILE_2 = Pattern.compile(buildRestPattern(6));
+    private static final Pattern FEATUREINFO_1 = Pattern.compile(buildRestPattern(7));
+    private static final Pattern FEATUREINFO_2 = Pattern.compile(buildRestPattern(8));
+
```
nmco: We have 1 and 2 to vouch for virtual services ?

aaime: No it's because the GWC services accept a PATH with or without the style name in the middle, hence it's either a 5 or 6 params path for the tile, and a 7 or 8 params path for the feature info

src/gwc/src/main/java/org/geoserver/gwc/dispatch/GwcServiceDispatcherCallback.java
```diff
@@ -66,8 +76,22 @@ public Request init(Request request) {
             return null;
         }
 
-        // storing the current operation
-        GWC_OPERATION.set((String) request.getKvp().get("REQUEST"));
+        // storing the current operation (bridge
```
nmco: Not sure if something is missing after ``(bridge``.

aaime: Yep, "bridge with RESTful services", which do not have a "request" KVP param

aaime:
Travis/Appveyor builds are green, going to fix that comment and then merge

PR #3348, opened 2019-02-01
[GEOS-9121] Added a fixed sort order to the module status table.

sikeoka:
See https://osgeo-org.atlassian.net/browse/GEOS-9121

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/admin/ModuleStatusPanel.java
+++ b/src/web/core/src/main/java/org/geoserver/web/admin/ModuleStatusPanel.java
@@ -4,8 +4,8 @@
  */
 package org.geoserver.web.admin;
 
+import java.util.Comparator;
 import java.util.List;
-import java.util.ListIterator;
 import java.util.stream.Collectors;
 import org.apache.wicket.ajax.AjaxRequestTarget;
 import org.apache.wicket.ajax.markup.html.AjaxLink;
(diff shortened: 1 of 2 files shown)


src/web/core/src/main/java/org/geoserver/web/admin/ModuleStatusPanel.java
```diff
@@ -51,19 +51,14 @@ public void initUI() {
         add(popup);
 
         // get the list of ModuleStatuses
-        GeoServerExtensions gse = new GeoServerExtensions();
         List<ModuleStatus> applicationStatus =
-                gse.extensions(ModuleStatus.class)
+                GeoServerExtensions.extensions(ModuleStatus.class)
                         .stream()
+                        .filter(status -> !status.getModule().matches("\\A[system-](.*)"))
```
aaime: I'm looking at two different installations and don't think I have a system module anywhere... can you make an example of one?
Also this change does not match the description of the commit nor the ticket, which are only talking about the order (which I agree with and find useful).

sikeoka: There are currently two system modules (one for environment variables and one for system properties):
https://github.com/geoserver/geoserver/blob/master/src/platform/src/main/java/org/geoserver/platform/SystemEnvironmentStatus.java
https://github.com/geoserver/geoserver/blob/master/src/platform/src/main/java/org/geoserver/platform/SystemPropertyStatus.java

I generally try to perform filtering as early as possible but there aren't very many modules so this doesn't really matter for performance and I could remove it if you want.  Calling the ModuleStatusImpl copy constructor on the system modules will build a string containing all of the key/value pairs which is unnecessary if they're just being filtered a few lines down.

aaime: Never mind, I did not notice that the filtering was present already, just done in the code section removed below.

PR #3349, opened 2019-02-01
[GEOS-9119] Add missing parameter to gwcBlobStore bean configuration

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-9119


aaime:
Works for me but.. how does GeoServer even start without the parameter in question? I only see one constructor there....

tbarsballe:
I'm not sure myself. My best guess is GeoServer is pulling in the bean from the [GWC version | https://github.com/GeoWebCache/geowebcache/blob/master/geowebcache/web/src/main/webapp/WEB-INF/geowebcache-core-context.xml#L94-L99]of that file somehow. Probably worth investigating, although debugging through Spring context instantiation is always a pain.

tbarsballe:
Okay, I actually debugged into it, and it turns out Spring is using the GeoServer context file and just auto-wiring the missing parameter via spring magic. I imagine this is a bit dependant on spring version and other config, which also explains why it doesn't work on all configurations.

aaime:
Bizzarre :-) 
Anyways, still best to be clear in the app-context.

PR #3350, opened 2019-02-01
[GEOS-9116] Update remaining instances of 2.15-SNAPSHOT to 2.16-SNAPSHOT

tbarsballe:
See https://osgeo-org.atlassian.net/browse/GEOS-9116


aaime:
Thanks!

PR #3354, opened 2019-02-05
GEOS-9096 jdbcconfig/hazelcast thread safety fixes

NielsCharlier:


src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
```diff
@@ -1477,17 +1472,36 @@ public boolean canSort(Class<? extends CatalogInfo> type, String propertyName) {
...
     }
 
-    void clear(Info info) {
+    void clearCache(Info info) {
         if (info instanceof ServiceInfo) {
             // need to figure out how to remove only the relevant cache
             // entries for the service info, like with InfoIdenties below,
             // that will be able to handle new service types.
             disposeServiceCache();
-        }
```
sikeoka: Removing this curly brace is breaking the Java compilation?

NielsCharlier: yeah there appear to be some automatic merging errors. I am fixing some unit tests as well and will repush

NielsCharlier: fixed now

sikeoka:
Are you able to successfully import a new catalog with these changes?  It seems like there are cases where PostModify events can be sent without a corresponding Modify event which will cause the ConfigDatabase.releaseWriteLock(String) method to throw a java.lang.IllegalMonitorStateException when it tries to release a lock that it isn't currently holding.

NielsCharlier:
@sikeoka I have tested it and it imports successfully, however it does indeed log those exceptions as errors (the exception are caught by the event dispatcher and do not stop the continuation). 

Personally, I think that the PostModify events shouldn't be sent without Modify events, because there are other event handlers that assume this flow.

But what I could do is be even more lenient about it. Do a check before releasing the lock or capture the exception and log at level warning or just info. What do you reckon?

sikeoka:
Catching the exception and logging something at info sounds good to me and maybe log the full stack trace at something lower than info.

I'm also curious about what happens if something goes wrong and Modify event is processed but no corresponding PostModify event is ever processed.  The basic flow seems to be:

fireModify
db.save
firePostModify

and with the database (and possibly Hazelcast) in between, there doesn't seem to be anything to protect against a lock being held indefinitely if something doesn't work as expected?

NielsCharlier:
@sikeoka Good point. It doesn't seem to normally happen (we already have this in production), but I do have some suggestions for improvements to protect against this.

I see two possible causes this could happen
1) an exception thrown in db.save. Could in theory happen if for example your database is temporarily inaccessible. Looking at the code, this is actually problematic for more than one reason. For example, the proxy is still committed so the catalog info object so it will still appear changed until the next reload from database. Also, there are only runtime exceptions which may not be caught at all and may crash whatever is calling catalog.save. So there seems to be some inherent assumption that this will not happen.

The only way to deal with this I see is to catch all exceptions thrown by db.save, log them as errors and simply continue.

2) a glitch in hazelcast communication. We have logged hazelcast messages and it seems perfectly reliable. However, in case this would exceptionally still occur, there is not much that can be done to detect this. All I think we can do is make sure that locks that are not released cannot cause any further issues.

I suggest to change the lock.lock() call in acquireLock to a tryLock with a timer; assuming that if the timer expires that we are dealing with an abandoned lock. In that case the lock will be ignored and then released with the next post-mod event.

Does this seem good to you? Do you have any other suggestions?

sikeoka:
That seems reasonable enough to me.

src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
```diff
@@ -1040,7 +1006,33 @@ private void updateQueryableProperties(
                 valueLoader = new ConfigLoader(id);
             }
 
-            info = cache.get(id, valueLoader);
+            Semaphore lock = locks.get(id);
+            if (lock == null) {
+                lock = new Semaphore(1);
+                locks.put(id, lock);
+            }
```
sikeoka: It probably wouldn't matter much but this code block (and below in acquireWriteLock) can be performed atomically using: java.util.concurrent.ConcurrentHashMap.computeIfAbsent(K, Function<? super K, ? extends V>)

Semaphore lock = locks.computeIfAbsent(id, x -> new Semaphore(1));

NielsCharlier: That is very elegant, thanks for the tip because I use that pattern a lot and wasn't aware yet of this :)

NielsCharlier: I changed it.

sikeoka: You changed it down in acquireWriteLock but not up here.

NielsCharlier: right, you are sharp :) fixed

sikeoka:
Everything looks okay to me.  Are you able to merge this and backport to 2.15.x?

NielsCharlier:
Done and thanks!

PR #3362, opened 2019-02-09
[GEOS-8762] Fix Monitor asynchronous post processing does not play well with security

fernandor777:
This PR:
- Fix Authentication propagation between Monitor filter and its underlying PostProcessTask thread.
- For testing purposes adds an (optional-nullable) audit functional consumer to Monitor filter for having feedback of RequestData and Authentication variables from threadpool's background thread scope.

https://osgeo-org.atlassian.net/browse/GEOS-8762

### DIFF
--- a/src/extension/monitor/core/src/main/java/org/geoserver/monitor/MonitorFilter.java
+++ b/src/extension/monitor/core/src/main/java/org/geoserver/monitor/MonitorFilter.java
@@ -247,7 +247,7 @@ public class MonitorFilter implements GeoServerFilter {
      * Audit consumer function to be executed on the underlying PostProcessTask thread. Will receive
      * {@link RequestData} and {@link Authentication} from thread execution.
      */
-    public void setExecutionAudit(BiConsumer<RequestData, Authentication> executionAudit) {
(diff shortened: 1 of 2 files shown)


src/extension/monitor/core/src/main/java/org/geoserver/monitor/MonitorFilter.java
```diff
@@ -231,26 +243,40 @@ String getHttpReferer(HttpServletRequest req) {
         }
     }
 
+    /**
+     * Audit consumer function to be executed on the underlying PostProcessTask thread. Will receive
+     * {@link RequestData} and {@link Authentication} from thread execution.
+     */
+    public void setExecutionAudit(BiConsumer<RequestData, Authentication> executionAudit) {
```
aaime: Since it's a testing facility I'd make the methods around it at least package private. 
Another possible approach would be to manually register a RequestPostProcessor that does the same work using GeoServerExtensionsHelper.singleton

fernandor777: Fixed, thanks.

PR #3367, opened 2019-02-15
[GEOS-9082] Update community script plugins for Java 11 compatibility

tbarsballe:
All script-* modules now compile, except for script-js, which depends on gt-jdbc-spatialite (which is currently disabled pending Java 11 migration).
script-core, script-web, script-bsh, script-py and script-groovy all pass tests. 
 script-rb does not - they appear to require new geoscript releases.

Given this, I've only re-enabled script-py and script-groovy in the release configuration, and have de-activated script-js and script-rb by default.

I've manually tested script-py on a GeoServer running under Java 11, and was able to run a python script via WPS without error.
I have not manually tested groovy.

---

This should get backported to 2.15.x, but not until after the 2.15.0 release, given how soon it is.


tbarsballe:
Looks like script-groovy doesn't build under java 11 due to issues with Groovy itself:

```
Compilation failure: Compilation failure: 
[ERROR] /home/travis/build/geoserver/geoserver/src/community/script/groovy/src/main/java/org/geoserver/script/groovy/GroovyPlugin.java:[1,1] 1. ERROR in /home/travis/build/geoserver/geoserver/src/community/script/groovy/src/main/java/org/geoserver/script/groovy/GroovyPlugin.java (at line 1)
[ERROR] 	/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
[ERROR] 	^
[ERROR] The type java.lang.Object cannot be resolved. It is indirectly referenced from required .class files
```

Looks like JDK11 support requires Groovy 3, which will require updates in geoscript-groovy. 
I'll disable script-groovy for now.

aaime:
Pity, would have been nice to see continued support for geoscript.
Even with the groovy part disabled, the python one would survive yes?

tbarsballe:
Yep, python one is still working.
The groovy one might be functional again in the future - it ultimately needs some updates over in https://github.com/geoscript/geoscript-groovy, (Groovy version update + GeoTools 21 API updates) and that project is at least somewhat active. If it does get updated, then getting the the groovy plugin working again should be pretty straightforward)

tbarsballe:
Backported to 2.15.x

PR #3373, opened 2019-02-19
Workspace admin was unable to view service admin pages inside Workspace Edit Page [GEOS-9108]

imranrajjad:
[GEOS-9108] Workspace admin was unable to view WMS/WFS/WMTS/WCS admin pages inside Workspace Edit Page


jodygarnett:
I was not aware we wanted a workspace admin to have access to these pages? Many of these settings control use of system resources and may not be appropriate to a workspace admin?

aaime:
A workspace admin should be able to configure per workspace services, but not the global ones, or not?

jodygarnett:
Quickly:
- editing the root layer definition: yes
- resource consumption limits: not so much

imranrajjad:
so the fields related to system configuration should not be visible to Workspace admin? is that the way forward?

aaime:
@jodygarnett there is no way to distinguish those two sets of fields automatically though, one would have to do so manually, service by service, and either have two tabs, or just two WebMarkupContainer.
Wondering what can be done using the REST API though... like if the REST API allows to edit them, then closing down the UI is no use.

jodygarnett:
That is a really good point about the REST API @aaime 

I think the two tabs is the easiest way forward (rather than field by field). It may even be a usability win - have a second "settings" tab that has all the integration settings (for format management, resource use, etc...).

Actually is it useful to change those kind of low level settings workspace by workspace?

imranrajjad:
so I am guessing options starting from **Authority URLs for the root WMS Layer** should not be visible to workspace admin?

aaime:
Meh, there is no "start and end", options should be checked one by one, service by service, to see if they are better suited for a global or a local admin. The service limits are one case where the workspace admin should be able to restrict, but not to expand, the service limits... but again, it depends on installation, I don't believe it's so clear. A global admin might decide to have a WS with larger limits and set is so, and then comes the local one, and then, what do you do?

aaime:
Linked from ticket, closing for now.

jodygarnett:
Is it worth setting up a proposal, there is a strong story here.

PR #3379, opened 2019-03-01
GEOS-7990 Layer sorting index

PascalLike:
This is a new feature to allow the definition of a layer order in getCapabilities as discussed here [GEOS-7990](https://osgeo-org.atlassian.net/browse/GEOS-7990)

This is the second attempt after https://github.com/geoserver/geoserver/pull/2125 . I adapted this code to the recent master and added the suggestions from @aaime , I hope it's better now.

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/SecureCatalogImpl.java
+++ b/src/main/src/main/java/org/geoserver/security/SecureCatalogImpl.java
@@ -1479,34 +1479,11 @@ public class SecureCatalogImpl extends AbstractDecorator<Catalog> implements Cat
 
     @Override
     public <T extends CatalogInfo> CloseableIterator<T> list(
-            Class<T> of, Filter filter, Integer offset, Integer count, SortBy sortBy) {
-        Filter securityFilter = securityFilter(of, filter);
-
-        CloseableIterator<T> filtered;
-        filtered = delegate.list(of, securityFilter, offset, count, sortBy);
-
-        // create secured decorators on-demand. Assume this method is used only for listing, not
-        // for accessing a single resource by name/id, thus use hide policy for mixed mode
-        final Function<T, T> securityWrapper = securityWrapper(of, MixedModeBehavior.HIDE);
-        final CloseableIterator<T> filteredWrapped;
-        filteredWrapped = CloseableIteratorAdapter.transform(filtered, securityWrapper);
-
-        // wrap the iterator in a notNull filter to ensure any filtered
-        // layers (result is null) don't get passed on from the securityWrapper
-        // Function. When the AccessLevel is HIDDEN and a layer gets filtered
-        // out via a CatalogFilter - for example, this can happen with a
-        // LocalWorkspaceCatalogFilter and a virtual service request
-        return CloseableIteratorAdapter.filter(
-                filteredWrapped, com.google.common.base.Predicates.<T>notNull());
-    }
-
-    public <T extends CatalogInfo> CloseableIterator<T> list(
             Class<T> of, Filter filter, Integer offset, Integer count, SortBy... sortBy) {
         Filter securityFilter = securityFilter(of, filter);
 
         CloseableIterator<T> filtered;
-        // HACK here, go straigth to the facade of the delegate to get a method supporting sortby[]
-        filtered = delegate.getFacade().list(of, securityFilter, offset, count, sortBy);
+        filtered = delegate.list(of, securityFilter, offset, count, sortBy);
 
         // create secured decorators on-demand. Assume this method is used only for listing, not
         // for accessing a single resource by name/id, thus use hide policy for mixed mode
--- a/src/main/src/main/java/org/vfny/geoserver/global/CoverageInfoIndexComparator.java
+++ b/src/main/src/main/java/org/vfny/geoserver/global/CoverageInfoIndexComparator.java
@@ -1,0 +1,28 @@
+/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2013 OpenPlans
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.vfny.geoserver.global;
+
+import java.util.Comparator;
+
+import org.apache.commons.lang.builder.CompareToBuilder;
+import org.geoserver.catalog.CoverageInfo;
+
+/**
+ * A comparator used to sort {@link CoverageInfo} objects by label 
+ * (useful for capabilities output, for example)
+ * @author Antonio Cerciello - Geocat
+ */
+public class CoverageInfoIndexComparator implements Comparator {
+
+    public int compare(Object o1, Object o2) {
+        CoverageInfo c1 = (CoverageInfo) o1;
+        CoverageInfo c2 = (CoverageInfo) o2;
+        
+        // this will take care of null values as well
+        return new CompareToBuilder().append(c1.getSortIndex(), c2.getSortIndex()).append(c1.getTitle(), c2.getTitle()).toComparison();
+    }
+
+}
--- a/src/main/src/main/java/org/vfny/geoserver/global/CoverageInfoIndexComparator.java
+++ b/src/main/src/main/java/org/vfny/geoserver/global/CoverageInfoIndexComparator.java
@@ -1,18 +1,17 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
- * (c) 2001 - 2013 OpenPlans
+/* (c) 2019 Open Source Geospatial Foundation - all rights reserved
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
  */
 package org.vfny.geoserver.global;
 
 import java.util.Comparator;
-
 import org.apache.commons.lang.builder.CompareToBuilder;
 import org.geoserver.catalog.CoverageInfo;
 
 /**
- * A comparator used to sort {@link CoverageInfo} objects by label 
- * (useful for capabilities output, for example)
+ * A comparator used to sort {@link CoverageInfo} objects by label (useful for capabilities output,
+ * for example)
+ *
  * @author Antonio Cerciello - Geocat
  */
 public class CoverageInfoIndexComparator implements Comparator {
@@ -20,9 +19,11 @@ public class CoverageInfoIndexComparator implements Comparator {
     public int compare(Object o1, Object o2) {
         CoverageInfo c1 = (CoverageInfo) o1;
         CoverageInfo c2 = (CoverageInfo) o2;
-        
-        // this will take care of null values as well
-        return new CompareToBuilder().append(c1.getSortIndex(), c2.getSortIndex()).append(c1.getTitle(), c2.getTitle()).toComparison();
-    }
 
+        // this will take care of null values as well
+        return new CompareToBuilder()
+                .append(c1.getSortIndex(), c2.getSortIndex())
+                .append(c1.getTitle(), c2.getTitle())
+                .toComparison();
+    }
 }
--- a/src/main/src/test/java/org/vfny/geoserver/global/IndexComparatorsTest.java
+++ b/src/main/src/test/java/org/vfny/geoserver/global/IndexComparatorsTest.java
@@ -1,0 +1,92 @@
+/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2013 OpenPlans
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.vfny.geoserver.global;
+
+import java.io.IOException;
+import java.util.Set;
+import java.util.TreeSet;
+
+import org.geoserver.catalog.FeatureTypeInfo;
+import org.geoserver.catalog.ResourceInfo;
+import org.geoserver.catalog.impl.CoverageInfoImpl;
+import org.geoserver.catalog.impl.CoverageStoreInfoImpl;
+import org.geoserver.catalog.impl.FeatureTypeInfoImpl;
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * Tests covering the former functionality of GeoServerDataDirectory.
+ * 
+ * Much of this functionality depends on the availability of GeoServerResourceLoader
+ * in the application context as the bean "resourceLoader".
+ * 
+ * @author Antonio Cerciello - Geocat
+ */
+public class IndexComparatorsTest {
+    
+    @Test
+    public void TestFeatureTypeSort() throws IOException {
+        
+        Set<FeatureTypeInfo> unordered = new TreeSet<FeatureTypeInfo>(new FeatureTypeInfoIndexComparator());
+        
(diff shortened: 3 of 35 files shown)


aaime:
Hey @PascalLike while waiting for a human review I'd suggest you to rebase to the latest master and then run the build with the "-Dqa" extra maven parameter to get the opinion of PMD/Spotbugs/Checkstyle/Errorprone. You already have some in your checkout, but the spotbugs checks have been upgraded a notch since you branched off from master.

PascalLike:
Hi @aaime , I did the rebase and the output of the _mvn clean install_ with **"-Dqa"** parameter is still BUILD SUCCESS. Thanks for the suggestion

src/main/src/main/java/org/vfny/geoserver/global/CoverageInfoIndexComparator.java
```diff
@@ -0,0 +1,30 @@
+/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
```
aaime: New file, so there is no OpenPlans copyright, should only be "(c) 2019 Open Source Geospatial Foundation - all rights reserved"

aaime: Also, why not have a generic ResourceInfoComparator instead? It would cover also WMS/WMTS cascaded layers and whatever else we might add in the future.

src/main/src/test/java/org/vfny/geoserver/global/IndexComparatorsTest.java
```diff
@@ -0,0 +1,88 @@
...
+ * application context as the bean "resourceLoader".
+ *
+ * @author Antonio Cerciello - Geocat
+ */
+public class IndexComparatorsTest {
+
+    @Test
+    public void TestFeatureTypeSort() throws IOException {
+
+        Set<FeatureTypeInfo> unordered =
```
aaime: Confusing naming, a "TreeSet" is ordered by definition. At a first glance I thought you were adding elements in a list and was wondering where the call to Collections.sort was :-D

src/wcs2_0/src/main/java/org/geoserver/wcs2_0/response/WCS20GetCapabilitiesTransformer.java
```diff
@@ -572,7 +582,7 @@ private void handleContents() {
 
             @SuppressWarnings("unchecked")
             final Set<CoverageInfo> coverages =
-                    new TreeSet<CoverageInfo>(new CoverageInfoLabelComparator());
+                    new TreeSet<CoverageInfo>(new CoverageInfoIndexComparator());
```
aaime: This is not backwards compatible, unless the layer indexes are populated the current output order should not be changed, I have seen existing applications depending on it (as you can see, the previous order was not happening out of a accident, it was forced by a comparator). The compatibility can be re-established by having a delegating or composite comparator, first compare by index, and if the indexes are same, use the other criteria. 
This method may be used to build the chained comparator: https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/ComparatorUtils.html#chainedComparator-java.util.Collection-
(did not check if available in the version of commons we have in the classpath)

src/wcs2_0/src/main/java/org/geoserver/wcs2_0/response/WCS20GetCapabilitiesTransformer.java
```diff
@@ -673,9 +683,9 @@ private void handleBoundingBox(BoundingBox boundingBox) throws Exception {
         }
 
         private void handleLanguages() {
-            //            start("ows:Languages");
-            //            // TODO
-            //            end("ows:Languages");
+            // start("ows:Languages");
```
aaime: Lots of comments have been reformatted, but this is not the work of google-format, it happened in a IDE I'm guessing. I could have closed an eye for a few cases, but they are everywhere... can you please back out all of these changes?

src/web/core/src/main/java/org/geoserver/web/data/layergroup/LayerGroupEditPage$LayerGroupTab.html
```diff
@@ -26,6 +26,10 @@
          <input wicket:id="generateBounds" type="submit" wicket:message="value:generateBounds"/>
          <input wicket:id="generateBoundsFromCRS" type="submit" wicket:message="value:generateBoundsFromSRS"/>
        </li>
+       <li>
+         <label for="sortIndex"><wicket:message key="sortIndex">Index for sorting layers on getcapabilities</wicket:message></label>
```
aaime: "on getcapabilities" should probably be "in GetCapabilities". @bradh I need the help of a native english speaker here, what say you? :-)

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.html
```diff
@@ -102,7 +102,18 @@ <h3><wicket:message key="title">Basic Settings</wicket:message></h3>
             </li>
           </ul>
         </fieldset>
+      </li>      
+      <li>
+        <fieldset>
+            <legend><span><wicket:message key="sortIndex">Index for sorting layers on getcapabilities</wicket:message></span></legend>
```
aaime: Same as above.

src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
```diff
@@ -703,8 +705,8 @@ private void handleLayers() {
             // handle root bounding box
             List<LayerGroupInfo> layerGroups;
             List<LayerInfo> layers;
-            SortBy lgOrder = asc("name");
-            SortBy order = asc("name");
+            SortBy lgOrder = asc("sortIndex");
```
aaime: Not backwards compatible, the current sorting order is the desired default and must not be modified. Please see comment above. This case is complicated though, the catalog machinery does not support sorting on multiple SortBy conditions, it must be extended... :-(

src/wms/src/main/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformer.java
```diff
@@ -751,7 +751,7 @@ private void handleLayers() {
             // namespace filter
             addNameSpaceFilterIfNeed(filter, "resource.namespace.prefix");
             // order by name ASC
-            SortBy order = asc("name");
+            SortBy order = asc("sortIndex");
```
aaime: Same as above

src/wms/src/main/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformer.java
```diff
@@ -861,7 +861,7 @@ private void handleLayerTree(
                     data,
                     new Comparator<LayerInfo>() {
                         public int compare(LayerInfo o1, LayerInfo o2) {
-                            return o1.getName().compareTo(o2.getName());
+                            return o1.getSortIndex().compareTo(o2.getSortIndex());
```
aaime: Same as above

src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/LayerWorkspaceTest.java
```diff
@@ -47,10 +46,12 @@ public void testLayerOrderGetCapabilities() throws Exception {
         Document doc = getAsDOM("/wms?service=WMS&request=getCapabilities&version=1.1.1", true);
         List<String> originalList = layerNameList(doc);
         assertFalse(originalList.isEmpty());
-        List<String> names =
-                originalList.stream().map(x -> removeLayerPrefix(x)).collect(Collectors.toList());
-        List<String> orderedNames = names.stream().sorted().collect(Collectors.toList());
-        assertTrue(orderedNames.equals(names));
+        int prevIndex = 0;
```
aaime: Same as above, without setting any index the previous test checks the desired behavior and should not be modified.

PascalLike:
Hi @aaime 

> As implemented the current implementation cannot be merged. Major pain points:
> 
> * Breaks  backwards compatibility of sorting by name, which is the default desired behavior (on upgrade that should not change, unless someone sets up layer order indexes)

I'll try to manage backwards compatibility. To manage this double ordering in WMS is a bit tricky indeed, but let's try. Thanks for the tips.

> * Lots of comments have been reformatted for no good reason (looks like a mass reformat by IDE, probably a "format on save" setting)

I can fix this, luckly I have a separate commit done after **mvn fmt:format** (the first time I submitted the PR I got an error from Travis (in the logs) suggesting to run this command to fix it)

> * There are UI changes but no associated documentation change, those should be added too, probably around these two areas
>   
>   * https://docs.geoserver.org/latest/en/user/data/webadmin/layers.html#edit-layer-data
>   * https://docs.geoserver.org/latest/en/user/data/webadmin/layergroups.html#edit-a-layer-group

Ok


Thank you very much for the suggestions

src/main/src/test/java/org/vfny/geoserver/global/IndexComparatorsTest.java
PascalLike: I feel very guilty about this... I don't know what I was thinking! :smile:

aaime:
The existing code is already fully formatted with "mvn fmt:format", which also runs at every "mvn install", the comments reformats must have have happened inside your IDE I'm afrai.d

aaime:
I started GeoServer forgetting to switch away from the branch and noticed this:

Selezione_612

That's a bit ugly, wouldn't it be better to have an empty field instead? How about having the index be null by default in the config, and update the comparators accordingly?

PascalLike:
> I started GeoServer forgetting to switch away from the branch and noticed this:
> 
> Selezione_612
> 
> That's a bit ugly, wouldn't it be better to have an empty field instead? How about having the index be null by default in the config, and update the comparators accordingly?

The idea was to have the layers without index added to the end.
Integer max that you mentioned before is a good suggestion, even if a bit weird for the end-user.
On the other hand, null (and then the empty field) is better to show, but less natural to order, mostly I was worried about SortBy behaviour. I can give it a try.

aaime:
Btw, if you need directions/suggestions just ask ok? :-)

aaime:
Checkstyle spotted dangling parameters in javadocs:

```
[INFO] There are 2 errors reported by Checkstyle 6.18 with /home/travis/build/geoserver/geoserver/src/../build/qa/checkstyle.xml ruleset.
[ERROR] src/main/java/org/geoserver/catalog/Catalog.java:[1563,8] (javadoc) JavadocMethod: Unused @param tag for 'sortBy1'.
[ERROR] src/main/java/org/geoserver/catalog/Catalog.java:[1564,8] (javadoc) JavadocMethod: Unused @param tag for 'sortBy1'.
```

src/main/src/main/java/org/geoserver/security/SecureCatalogImpl.java
```diff
@@ -1524,6 +1524,24 @@ public void setDefaultDataStore(WorkspaceInfo workspace, DataStoreInfo defaultSt
                 filteredWrapped, com.google.common.base.Predicates.<T>notNull());
     }
 
+    @Override
+    public <T extends CatalogInfo> CloseableIterator<T> list(
+            Class<T> of,
+            Filter filter,
+            Integer offset,
+            Integer count,
+            SortBy sortBy1,
```
aaime: This is clumsy, the time we need a three way sort for whatever reason are we going to have a new public method for it? The underlying API allows for a open ended list of sort values to be used, it's just easier to change the existing list method to take a "SortBy..." instead.
If you're out of time on this one I'll be happy to take a stab at it (instead of losing the PR again like last time).

PascalLike: I'm still doing the changes that you requested, it was not really ready for review, but useful anyway, thanks. I'll tell you when is ready again for your review, to not waste your time. 
I've done the replacement with variable arguments `SortBy...` but there is still some part not ready for the behaviour of null (and the resulting SortBy[] {null} ), I need to investigate better. I guess this was the reason there was `SortBy` instead of `SortBy...`

PascalLike:
Hi @aaime , hopefully now the PR is worth some attention. I sorted out the issues you found, except the ugly default value for sort index in the form (now is Integer.MAX). Since it's a presentation issue, I was thinking to solve it directly in the form, I guess with Wicket, that I don't know much.
I would like to do something like "if sortIndex == Integer.MAX then show an empty field". If you have a suggestion or a better idea please tell me.

aaime:
Hi @PascalLike. I believe it's all good besides the index. I still have an issue in general with the default value, I don't think it's a UI only issue, that would show up also in the REST API for example.
I checked around, the catalog comparators for sorting are null safe, as the ComparatorBuilder is also null safe, so is there any actual good reason why null cannot be the default instead?

PascalLike:
Thank you for the suggestions @aaime 
I prefer the IModel solution you suggested. Now I'm trying to understand why the value is not returned when I try to edit the layer (returns to the default value, even if seems saved correctly). Probably `new IndexModel(((ResourceInfo) model.getObject()).getSortIndex())` is not the right way. I need to check better.

jodygarnett:
Reading through this is it worth writing a proposals capturing intended design?

src/web/core/src/main/java/org/geoserver/web/data/layergroup/LayerGroupEditPage$LayerGroupTab.html
jodygarnett: How about:

> GetCapabilities order:

Order implies sorting?

aaime:
Compile failures now:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project gs-wms: Compilation failure: Compilation failure: 
[ERROR] /home/travis/build/geoserver/geoserver/src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java:[675,34] variable layerGroups is already defined in method handleLayers()
[ERROR] /home/travis/build/geoserver/geoserver/src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java:[676,29] variable layers is already defined in method handleLayers()
```

aaime:
Legit build failure in the QA checks, possibly a missing copyright header:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.0.0:check (default) on project gs-web-core: You have 1 Checkstyle violation. -> [Help 1]
```

@jodygarnett what are we doing with this one?

jodygarnett:
@pascallike should we close this?

PascalLike:
@jodygarnett, our customer is not interested in this functionality anymore, but I would be delighted to follow up if there is any interest in such a feature, and also to not waste the work done until now. But I wouldn't want to add unrequired complexity if it's really of no use.
@aaime, I'll check the copyright headers, thanks, I was confused by the two checks, that switched status at some point and I was thinking to some malfunction

PR #3382, opened 2019-03-04
[GEOS-8486] create new user through REST API

imranrajjad:
https://osgeo-org.atlassian.net/browse/GEOS-8486

added XStreamPersister  dedicated to aliasing and white listing JaxbUser class for XML and JSON marshalling/unmarshalling

### DIFF
--- a/src/restconfig/src/main/java/org/geoserver/rest/security/RestConfigXStreamPersister.java
+++ b/src/restconfig/src/main/java/org/geoserver/rest/security/RestConfigXStreamPersister.java
@@ -1,0 +1,31 @@
+/* (c) 2019 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.rest.security;
+
+import java.util.logging.Logger;
+import org.geoserver.config.util.XStreamPersister;
+import org.geoserver.config.util.XStreamPersisterInitializer;
+import org.geoserver.rest.security.xml.JaxbUser;
+import org.geotools.util.logging.Logging;
+
+/**
+ * XStreamPersisterInitializer implementation for gs-rest-config
+ *
+ * @author ImranR
+ */
+public class RestConfigXStreamPersister implements XStreamPersisterInitializer {
+
+    /** logging instance */
+    static Logger LOGGER = Logging.getLogger(RestConfigXStreamPersister.class);
+
+    @Override
+    public void init(XStreamPersister persister) {
+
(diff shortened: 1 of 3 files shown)


src/restconfig/src/main/java/org/geoserver/rest/security/RestConfigXStreamPersister.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.util.logging.Logger;
+import org.geoserver.config.util.XStreamPersister;
+import org.geoserver.config.util.XStreamPersisterInitializer;
+import org.geoserver.rest.security.xml.JaxbUser;
+import org.geotools.util.logging.Logging;
+
+/**
+ * @author ImranR
+ * @since 2.15
```
aaime: GeoServer only "API" points are considered the interfaces for extension points, no real need to add "@since" markers

src/restconfig/src/main/resources/applicationContext.xml
```diff
@@ -1,20 +1,23 @@
...
-        http://www.springframework.org/schema/context
-        http://www.springframework.org/schema/context/spring-context.xsd">
-
-  <!-- <mvc:annotation-driven/> -->
-  <context:component-scan base-package="org.geoserver.catalog.rest"/>
-  <context:component-scan base-package="org.geoserver.security.rest"/>
-  <context:component-scan base-package="org.geoserver.catalog.rest"/>
-  <context:component-scan base-package="org.geoserver.service.rest"/>
-
+<?xml version="1.0" encoding="UTF-8"?>
```
aaime: Looks like the file newlines have been changed, check your git or IDE configuration and avoid it.

src/restconfig/src/test/java/org/geoserver/rest/security/XStreamParserTest.java
```diff
@@ -0,0 +1,37 @@
...
+import com.thoughtworks.xstream.XStream;
+import com.thoughtworks.xstream.io.json.JettisonMappedXmlDriver;
+import org.geoserver.rest.security.xml.JaxbUser;
+import org.junit.Test;
+
+public class XStreamParserTest {
+
+    // https://osgeo-org.atlassian.net/browse/GEOS-8486
+    @Test
+    public void jsonReadWriteTest() {
```
aaime: Please add an integration test too, for both the XML and JSON case, the reason why we are dealing with this regression is the lack of such test.

aaime:
Looks good to me, @imranrajjad can you backport?

PR #3387, opened 2019-03-07
GEOS-9137: Add WMS configuration options for densification on reproje…

mbarto:
…ction and disabling wrapping dateline heuristic


mbarto:
@aaime pushed the docs, and support of format_options as a bonus

aaime:
Looks good to me, thanks!

PR #3389, opened 2019-03-07
[GEOS-6977] Fixed WCS 1.0.0 GetCoverage KVP ignoring elevation and custom dimension parameters.

sikeoka:
See https://osgeo-org.atlassian.net/browse/GEOS-6977

Incorrect parsing of elevation ranges that did not allow decimal numbers was also fixed and some duplicate/unnecessary elevation parsing that had the same incorrect elevation range parsing was removed.

### DIFF
--- a/src/wcs1_0/src/main/java/applicationContext.xml
+++ b/src/wcs1_0/src/main/java/applicationContext.xml
@@ -94,6 +94,11 @@
     <constructor-arg ref="geoServer"/>
   </bean>
 
+  <bean id="wcs100ElevationKvpParser" class="org.geoserver.wcs.kvp.ElevationKvpParser">
+    <constructor-arg value="ELEVATION"/>
+    <constructor-arg ref="geoServer"/>
+  </bean>
+
 
   <!-- kvp request readers -->
 
--- a/src/wcs1_0/src/main/java/org/geoserver/wcs/kvp/Wcs10GetCoverageRequestReader.java
+++ b/src/wcs1_0/src/main/java/org/geoserver/wcs/kvp/Wcs10GetCoverageRequestReader.java
@@ -45,6 +45,7 @@ import org.geotools.metadata.i18n.Errors;
 import org.geotools.referencing.CRS;
 import org.geotools.referencing.crs.DefaultGeographicCRS;
 import org.geotools.util.DateRange;
+import org.geotools.util.NumberRange;
 import org.opengis.referencing.FactoryException;
 import org.opengis.referencing.NoSuchAuthorityCodeException;
 import org.opengis.referencing.crs.CoordinateReferenceSystem;
@@ -396,6 +397,7 @@ public class Wcs10GetCoverageRequestReader extends EMFKvpRequestReader {
      * @param rangeSubset
      * @param axis
      */
+    @SuppressWarnings("unchecked")
     private void checkTypeAxisRange(
             final RangeSubsetType rangeSubset, Object axis, String axisName) {
         if (axis instanceof String) {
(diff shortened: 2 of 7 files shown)


sikeoka:
I'm not sure why AppVeyor has failed three times even though Travis seems fine.

src/wcs1_0/src/main/java/applicationContext.xml
```diff
@@ -94,6 +94,11 @@
     <constructor-arg ref="geoServer"/>
   </bean>
 
+  <bean id="wcs100ElevationKvpParser" class="org.geoserver.wcs.kvp.ElevationKvpParser">
```
aaime: Won't this conflict with the "elevation" KVP parser in WMS? Best to limit it to the WCS service, and eventually to the version 1.0 version too.

src/wcs1_0/src/main/java/org/geoserver/wcs/kvp/Wcs10GetCoverageRequestReader.java
```diff
@@ -389,13 +394,35 @@ private RangeSubsetType parseRangeSubset(Map kvp, String coverageName) {
...
+        CoverageInfo coverage = this.catalog.getCoverageByName(coverageName);
+        if (coverage != null) {
+            for (Entry<String, ?> entry : coverage.getMetadata().entrySet()) {
+                String name = entry.getKey();
+                if (name.startsWith(ResourceInfo.CUSTOM_DIMENSION_PREFIX)
+                        && entry.getValue() instanceof DimensionInfo
+                        && ((DimensionInfo) entry.getValue()).isEnabled()) {
+                    name = name.substring(ResourceInfo.CUSTOM_DIMENSION_PREFIX.length());
+                    Object value = kvp.get(name);
+                    // ignore anything that got converted by a KVP parser so that they
```
aaime: Best to use the rawKVP then, it contains the unparsed values.

sikeoka: It is intentional that kvp is used here instead of rawKVP.  This assumes that anything that got parsed from a String is a keyword and should be ignored as a custom dimension so that it isn't being handled both as a keyword and a custom dimension.

aaime: Hmm... wondering about the WMS convention of prefixing with DIM_ to avoid conflicts (since we are in "vendor behavior" territory anyways). For WCS 2.0 there is no issue as the name is not used a key, did you check what WCS 1.1 is doing, if anything at all?

src/wcs1_0/src/main/java/applicationContext.xml
sikeoka: I wasn't sure what to do about that but the conflict already exists with the WCS and WMS TimeKvpParser.  I'm wondering if that should be a separate PR that limits the WCS and WMS TimeKvpParser and ElevationKvpParser.

src/wcs1_0/src/main/java/org/geoserver/wcs/kvp/Wcs10GetCoverageRequestReader.java
sikeoka: As far as I can tell, WCS 1.1 only handles the time dimension.  It does not have elevation or custom dimension support implemented even for XML requests.

Unlike WMS, the DIM_ prefix isn't in the WCS 1.0 specification but I could definitely add it if it would help.

aaime: Are custom dimensions part of the WCS 1.0 specification at all?

sikeoka: Yes, it is part of the specification although GeoServer's WCS 1.0 DescribeCoverage doesn't actually show the custom dimensions.
> If the range set of the selected coverage consists of compound values, GetCoverage requests may include constraints defined on the parameter(s) of the compound range set. Such constraints are expressed as “PARAMETER=[value]”, where  
a) The variable string PARAMETER matches the name of an AxisDescription element defined on that range set, and  
b) [value] is one of the acceptable values defined in the corresponding AxisDescription element.

aaime: Oh ok, then let's not change the current code.

src/wcs1_0/src/main/java/applicationContext.xml
aaime: It the PR is done promptly I see no problem, if you're not sure about doing it, I'd avoid introducing yet another potential conflict.

sikeoka: I just went ahead and fixed the WCS ElevationKvpParser.

aaime:
Looks good to me

PR #3392, opened 2019-03-08
[GEOS-9142] Added missing metadata to WCS 2.0.1 GetCapabilities and DescribeCoverage.

sikeoka:
See https://osgeo-org.atlassian.net/browse/GEOS-9142


sikeoka:
Will create a new PR later.

PR #3393, opened 2019-03-08
Add gs-fgdb community module

tbarsballe:
Initial import from https://github.com/boundlessgeo/geoserver-exts/tree/master/fgdb (module donated to GeoServer by Boundless)

Note: The first commit is just a straight import from geoserver-exts, with FGDBDataStoreFactory removed/inlined where necessary (Using JniOGRDataStoreFactory in its place)

While it compiles, no additional validation has been done.


aaime:
Thank you Torben! Just a small nitpick, if the module is a importer plugin, would it make sense to call it importer-fgdb instead?

tbarsballe:
I didn't get the rename done prior to Boundless switching over to Planet Federal, but now that Planet Federal's got a CLA signed again, I've applied the suggested rename. I agree that importer-fgdb makes more sense as a name.

src/community/release/ext-importer-fgdb.xml
```diff
@@ -0,0 +1,19 @@
+<assembly>
+  <id>importer-fgdb</id>
```
aaime: I have vague memories that a plugin module needs a name ending with -plugin or something bad (can't remember what) might happen.  At the very least all other community modules I've checked have an assembly whose name ends with -plugin.

src/community/importer-fgdb/src/main/resources/git.properties
```diff
@@ -0,0 +1,15 @@
+#Generated by Git-Commit-Id-Plugin
```
aaime: Is this file really meant to be committed?

aaime:
Added a couple more small comments, besides that everything seems to be ready for merge.

src/community/importer-fgdb/src/main/resources/git.properties
tbarsballe: Nope, it isn't, leftover from the repo it was copied from. I've removed it.

src/community/release/ext-importer-fgdb.xml
tbarsballe: At a guess, it may affect the name of the zipfile created by the assembly process. I've updated it.

tbarsballe:
Fixed based on feedback. I've also fixed a couple methods that got removed by the depreciation removal. If the CI is happy with it, I'll merge.

PR #3395, opened 2019-03-11
[GEOS-9142] Added missing metadata to WCS 2.0.1 GetCapabilities and DescribeCoverage.

sikeoka:
See https://osgeo-org.atlassian.net/browse/GEOS-9142

I am not familiar with the EO module but the sample DescribeEOCoverageSet responses in the module actually include the xmlns:ows namespace even though GeoServer wasn't writing it to its responses.


aaime:
Btw, this is not a fix but an improvement, please hold backporting on 2.15.x until 2.15.1 release is out (it's due Monday, might slip a few days).

sikeoka:
@aaime , did you mean until after the 2.14.3 or 2.15.1 release?

aaime:
The 2.14.3 one (yet to be completed)

PR #3396, opened 2019-03-13
Quote highest-basedir property to fix build in a path with spaces

bencaradocdavies:
Same fix as in https://github.com/geotools/geotools/pull/2317


bencaradocdavies:
This change quotes too mcuh.

PR #3404, opened 2019-03-18
[GEOS-9154] Env variables not always expanded when classifying layer

aaime:


src/extension/sldService/src/main/java/org/geoserver/sldservice/rest/ImageReader.java
```diff
@@ -134,15 +142,26 @@ public ImageReader invoke() throws IOException, TransformException, FactoryExcep
...
                     CoverageUtils.mergeParameter(
                             parameterDescriptors,
                             readParameters,
-                            readGeometry,
-                            AbstractGridFormat.READ_GRIDGEOMETRY2D.getName().getCode());
+                            readFilter,
+                            ImageMosaicFormat.FILTER.getName().getCode());
         }
 
+        // if (!readGeometry.equals(originalGeometry)) {
```
dromagnoli: is this intentionally left commented to explain the use case or is it an oversight instead?

aaime: Oversight, will fix

PR #3406, opened 2019-03-20
[GEOS-9155] Fixed GetFeatureInfo templates not being able to retrieve metadata for raster layers.

sikeoka:
See https://osgeo-org.atlassian.net/browse/GEOS-9155

### DIFF
--- a/src/main/src/main/java/org/geoserver/template/FeatureWrapper.java
+++ b/src/main/src/main/java/org/geoserver/template/FeatureWrapper.java
@@ -116,16 +116,10 @@ public class FeatureWrapper extends BeansWrapper {
     }
 
     private Catalog getCatalog() {
-        if (gsCatalog != null) return gsCatalog;
-
         try {
-            return (gsCatalog = (Catalog) GeoServerExtensions.bean("catalog2"));
+            return (gsCatalog = (Catalog) GeoServerExtensions.bean("catalog"));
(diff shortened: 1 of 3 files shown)


sikeoka:
This is a case where a trivial bug is quite difficult to test without additional changes.  org.geoserver.wms.featureinfo.HTMLFeatureInfoOutputFormat has a static reference to an org.geoserver.template.FeatureWrapper instance which also has a static reference to the GeoServer catalog so the FeatureWrapper's catalog will be different than the unit test's catalog for everything after the first test class.  It also doesn't help that FeatureWrapper has what appears to be the only reference anywhere in GeoServer to a bean named "catalog2".

src/main/src/main/java/org/geoserver/template/FeatureWrapper.java
```diff
@@ -116,16 +116,10 @@ public FeatureWrapper(TemplateFeatureCollectionFactory templateFeatureCollection
     }
 
     private Catalog getCatalog() {
-        if (gsCatalog != null) return gsCatalog;
```
aaime: Bizarre indeed... dunno, it's stuff that happened 11 and 8 years ago respectively, I'm guessing once upon a time there use to be a catalog2: https://github.com/geoserver/geoserver-history/blame/master/src/main/src/main/java/org/geoserver/template/FeatureWrapper.java#L129

PR #3407, opened 2019-03-21
[GEOS-9158] Simplify JSON encoding of 'xml attribute' values

aaime:


src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/GeoJsonOutputFormatWfsTest.java
```diff
@@ -79,6 +79,7 @@ public void testGetGeoJsonResponseWfs20() throws Exception {
                 getAsJSON(
                         "wfs?request=GetFeature&version=2.0"
                                 + "&typenames=st_gml32:Station_gml32&outputFormat=application/json");
+        print(response);
```
nmco: A miss?

aaime: Forgot it yes

PR #3408, opened 2019-03-21
[GEOS-9159] Add missing commons-compress dep for excel module.

bradh:


aaime:
Looks good to me, thanks!

PR #3410, opened 2019-03-21
[GEOS-9161] Complex attributes with date simple content are not encoed properly by the ComplexJSONBuilder

aaime:
Comes with a fix for "[GEOS-9160] JSON encoding of 1-d linear geometries fails" as I could not get the date test running otherwise.

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/json/ComplexGeoJsonWriter.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/json/ComplexGeoJsonWriter.java
@@ -7,6 +7,7 @@ package org.geoserver.wfs.json;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Date;
 import java.util.HashMap;
 import java.util.Iterator;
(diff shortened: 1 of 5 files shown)


src/wfs/src/main/java/org/geoserver/wfs/json/ComplexGeoJsonWriter.java
```diff
@@ -416,7 +417,10 @@ private Object getSimpleContent(ComplexAttribute property) {
             return null;
         }
         Object value = simpleContent.getValue();
-        if (value instanceof Number || value instanceof String || value instanceof Character) {
+        if (value instanceof Number
```
nmco: Wondering if this check is rely needed or if we could just return whatever is the simple content value? But I agree that if it is not clear then let's keep it, feels safer.

aaime: Indeed I did not touch cause I thought the existing if was there for a reason (But I don't know which one, ha!)

PR #3411, opened 2019-03-22
[GEOS-9163] Added 'https://' protocol support for default Data Store Edit Panel and URLModel class

fernandor777:
Added support for https:// on default datastore edit panel, and added tests for URLModel class use cases.

https://osgeo-org.atlassian.net/browse/GEOS-9163


aaime:
Looks good to me, merging. Feel free to backport as needed :-D

PR #3412, opened 2019-03-22
Add schema.cache.dir system property usage to user manual

fernandor777:
Adds a warning about schema.cache.dir system property requirement for remote URL mapping file to user manual.


aaime:
I think it's ok, but I'd like to hear @nmco second opinion on it?

PR #3415, opened 2019-03-25
[GEOS-9165] App-schema indexes integration test for fix solr query row limit

fernandor777:
This PR adds integration testing for: App-schema indexes makes query without correct maxFeatures number on "rows" parameter to Apache Solr.

https://osgeo-org.atlassian.net/browse/GEOS-9165

Requires:
https://github.com/geotools/geotools/pull/2336


nmco:
Looks good, thank you @fernandor777.

Suggestion for the commit message:
```
[GEOS-9165] App-schema indexes makes query without correct maxFeatures number on "rows" parameter to Apache Solr
```

The correct issue is:
https://osgeo-org.atlassian.net/browse/GEOS-9165

PR #3428, opened 2019-03-29
[GEOS-9145] GetLegendGraphic in JPEG will fail if libjpeg-turbo is installed

aaime:
Got rid of a legacy class that was used only to write the JPEG legend and used ImgeWorker instead. This implies TurboJPEG is not going to be used for legends, even if installed. I don't see that as a problem, legends are not called as often as GetMap (for GetMap there is a dedicated GetMap output format and a filter that removes the normal JPEG one, if someone wants to really use TurboJPEG for writing legends, that would be the way).


aaime:
Failed on mapml formatting issue, which was already fixed on master, moving on.

PR #3429, opened 2019-03-29
[GEOS-8982] GetCapabilities on a single layer fails if a style is duplicated

fernandor777:
This PR adds extra LegendInfo validation rules on WMS Capabilities_1_3_0_Translator class, fixing the bug described on:
https://osgeo-org.atlassian.net/browse/GEOS-8982

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
+++ b/src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
@@ -13,6 +13,7 @@ import static org.geoserver.ows.util.ResponseUtils.appendQueryString;
 import static org.geoserver.ows.util.ResponseUtils.buildSchemaURL;
 import static org.geoserver.ows.util.ResponseUtils.buildURL;
 import static org.geoserver.ows.util.ResponseUtils.params;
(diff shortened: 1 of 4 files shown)


src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
```diff
@@ -1363,18 +1363,19 @@ protected void handleAttribution(PublishedInfo layer) {
...
-         * @param style The styel for the layer.
-         * @param sampleStyle The styel to use for sample sizing.
+         * @param style The style for the layer.
+         * @param sampleStyle The style to use for sample sizing.
          */
         protected void handleLegendURL(
                 LayerInfo layer, LegendInfo legend, StyleInfo style, StyleInfo sampleStyle) {
-            if (legend != null) {
+            // if legend is valid, use it
+            if (validateLegendInfo(legend)) {
```
aaime: Looks good to me. However, the very same fix also needs to be done in the WMS 1.1 caps transformer, can you add that one too and the associated test?

fernandor777: Added WMS 1.1.1 fix and test as well, thanks!

PR #3431, opened 2019-04-01
openlayers call geoserver when zoom is not integer the Map layers and text are blurred

page404:
openlayers call geoserver when zoom is not integer the Map layers and text are blurred.This is not what I want.


bradh:
This doesn't look like a fix (PR). If you are trying to log an issue, please use the JIRA after a full review of the procedures.

aaime:
And even before going to Jira, would be best to talk on the user list, see here:
http://geoserver.org/comm/

PR #3435, opened 2019-04-11
Backporting SLD service env variable support to 2.15.x

aaime:


src/extension/sldService/src/main/java/org/geoserver/sldservice/rest/RestEnvVariableCallback.java
```diff
@@ -0,0 +1,43 @@
...
+
+/** Parses and sets the environment variables, clears them at the end of the request */
+public class RestEnvVariableCallback extends DispatcherCallbackAdapter {
+
+    static final FormatOptionsKvpParser PARSER = new FormatOptionsKvpParser("env");
+
+    /**
+     * Parses and sets the environment variables from their "var1:value1;v2:value;..." syntax, that
+     * a {@link org.springframework.stereotype.Controller} retrieved from the request some way
+     * (ideally via a "env" KVP parameter, but we don't want to be prescrictive about it, in REST
```
bradh: prescriptive

PR #3436, opened 2019-04-11
Second round of complex JSON encoding

aaime:


src/wfs/src/main/java/org/geoserver/wfs/json/ComplexGeoJsonWriter.java
```diff
@@ -28,10 +35,41 @@
...
+                                                    .getSimpleName()
+                                                    .equals("AppSchemaDataAccessFactory"))) {
+                LOGGER.log(
+                        Level.FINE,
+                        "Could not find NonFeatureTypeProxy yet App-schema is around, probably the class changed name, package or does not exist anymore",
+                        e);
+            }
+            NON_FEATURE_TYPE_PROXY = null;
+        }
+    }
```
nmco: Nice!

src/wfs/src/main/java/org/geoserver/wfs/json/ComplexGeoJsonWriter.java
```diff
@@ -207,10 +264,15 @@ private void encodeChainedFeatures(String attributeName, List<Feature> chainedFe
         jsonWriter.key(attributeName);
         jsonWriter.array();
         for (Feature feature : chainedFeatures) {
-            // encode each chained feature
-            jsonWriter.object();
-            encodeProperties(null, feature.getType(), feature.getProperties());
-            jsonWriter.endObject();
+            // if it's GeoJSON compatible, encode as a full blown GeoJSON feature (must have a
+            // default geometry)
+            if (feature.getType().getGeometryDescriptor() != null) {
```
nmco: Question, should we also not do the ``isFullFeature`` test here? Maybe not, since only real features may appear as a list.

src/wfs/src/main/java/org/geoserver/wfs/json/ComplexGeoJsonWriter.java
```diff
@@ -360,6 +438,58 @@ private void encodeProperty(Property property) {
...
+     * there is no substitution group or inheritance, there are attribute groups sometimes found in
+     * these constructs, but not mandatory and not always present at the schema level.
+     *
+     * <p>This code works by recognizing the common alternation nomenclature, that is:
+     *
+     * <ul>
+     *   <li>The attribute type is called ${name}PropertyType
+     *   <li>It contains a single element inside, which is in turn another complex attribute itself
+     *   <li>The contained element type is called ${name}Type or the property is called ${name}
+     * </ul>
```
nmco: Well this look like the only safe path we can use when looking only at the XML \ XSD. In the future we may also want to add some information in the property user data that says that it was feature chained, hence an high probability of being a GML property type.

src/wfs/src/main/java/org/geoserver/wfs/json/ComplexGeoJsonWriter.java
aaime: Good idea, does not hurt to make an extra check.

src/wfs/src/main/java/org/geoserver/wfs/json/ComplexGeoJsonWriter.java
aaime: Works for me

PR #3442, opened 2019-04-18
GEOS-9182 retrieve layer type from mapping

NielsCharlier:


NielsCharlier:
closed until module split

PR #3443, opened 2019-04-18
GEOS-9184 style image upload: NPE if new style

NielsCharlier:


aaime:
Looks good to me, merging

PR #3444, opened 2019-04-23
[GEOS-9186] Allow logging filter to also dump request headers

aaime:


aaime:
Tried, works:

image

src/main/src/main/java/org/geoserver/filters/LoggingFilter.java
```diff
@@ -94,6 +105,7 @@ public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain)
     public void init(FilterConfig filterConfig) {
         enabled = getConfigBool("enabled", filterConfig);
         logBodies = getConfigBool("log-request-bodies", filterConfig);
+        logHeaders = getConfigBool("log-request-headers", filterConfig);
```
jodygarnett: Okay this is easier to manage next to "log-request-bodies" setting.

I initially wondered why this was not being handled as a log4j configuration option.

bradh:
@aaime Do you still need this branch?

aaime:
Nope

PR #3450, opened 2019-04-27
Speed up data_dir load when there are lots of DataStoreInfos

groldan:
ConfigurationPasswordEncryptionHelper.decode(String)
is called by XstreamPersister -> StoreInfoConverter
for each DataStoreInfo read from the data directory at
start up time.

This patch makes it use a ConcurrentMap internally
instead of a HashMap, and removes the sychronization
on its CACHE member variable. Additionally, and most
importantly, caches the encryptable store parameter
names by StoreInfo.getType() too, avoiding the fallback
to full DataAccessFactory parameter lookup, which
implies a rather expensive call to getCatalog() \
.getResourcePool().getDataStoreFactory((DataStoreInfo) info)

As a result, loading time when there are lots of
DataStoreInfos is reduced to about 1/3 of the time
it takes without this patch.

### DIFF
--- a/src/main/src/main/java/org/geoserver/config/GeoServerLoader.java
+++ b/src/main/src/main/java/org/geoserver/config/GeoServerLoader.java
@@ -5,6 +5,7 @@
  */
 package org.geoserver.config;
 
+import com.google.common.base.Stopwatch;
 import java.io.*;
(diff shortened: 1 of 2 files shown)


aaime:
@groldan you should still have merge powers, or not?

PR #3451, opened 2019-04-28
Fix gwc tile layer load performance regression at startup

groldan:
DefaultTileLayerCatalog listens to Resouce events under
the data directory's gwc-layers folder resource in order to
notify its local instance listeners of tile layer configuration
changes even when on a shared directory by several service
instances.

To do so, it was using both a "folder" listener on gwc-layers,
and one listener per tile layer info resource (i.e. gwc-layers/*.xml).

This resulted in a severe startup hit when there are several
tile layers.

This patch uses a single listener on the gwc-cache folder resource
to handle all the scenarios, and hence removes the need to set up
and cache one resource listener per tile layer.

### DIFF
--- a/src/gwc/src/main/java/org/geoserver/gwc/layer/DefaultTileLayerCatalog.java
+++ b/src/gwc/src/main/java/org/geoserver/gwc/layer/DefaultTileLayerCatalog.java
@@ -9,6 +9,7 @@ import static com.google.common.base.Throwables.propagate;
 import static com.google.common.base.Throwables.propagateIfInstanceOf;
 
 import com.google.common.base.Preconditions;
+import com.google.common.base.Stopwatch;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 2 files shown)


aaime:
Yep, this one really got me mad but the original author cannot be relied to fix on his mess and the reviewer did not catch the problem..... Thank you for fixing it!

groldan:
oh, cool then, glad to be back in the game :P

groldan:
ok, merging then, thanks for the review @NielsCharlier

groldan:
Indeed, I didn't need to add any extra test case cause it was well covered already so I based the refactoring on those

aaime:
@groldan are you going to backport this one? I believe 2.15.x is affected, not sure about 2.14.x

groldan:
@aaime  sorry I couldn't come back to this before. Created a ticket for the backport: https://osgeo-org.atlassian.net/browse/GEOS-9219

PR #3455, opened 2019-04-29
[GEOS-9153] WCS 2.0 scaling policies do not account for scaling factor already applied during read (due to subsampling and overview)

aaime:
No new tests because it fixes a build failure that would occur merging https://github.com/geotools/geotools/pull/2354
Basically the GeoTools change triggers a but in GeoServer which then fails the build in the WCS 2.0 module, unless this PR gets applied.

### DIFF
--- a/src/wcs2_0/src/main/java/org/geoserver/wcs2_0/ScalingPolicy.java
+++ b/src/wcs2_0/src/main/java/org/geoserver/wcs2_0/ScalingPolicy.java
@@ -102,7 +102,7 @@ enum ScalingPolicy {
             }
 
             // return coverage unchanged if we don't scale
-            if (scaleFactors[0] == 1 && scaleFactors[1] == 0) {
+            if (scaleFactors[0] == 1 && scaleFactors[1] == 1) {
                 // NO SCALING do we need interpolation?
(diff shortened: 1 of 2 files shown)


src/wcs2_0/src/main/java/org/geoserver/wcs2_0/ScalingPolicy.java
```diff
@@ -82,19 +82,27 @@ public GridCoverage2D scale(
...
+            if (scaleFactors[1] <= 0) {
+                throw new WCS20Exception(
+                        "Invalid scale factor",
+                        WCS20Exception.WCS20ExceptionCode.InvalidScaleFactor,
+                        String.valueOf(scaleFactors[0]));
             }
 
             // return coverage unchanged if we don't scale
-            if (scaleFactor == 1) {
+            if (scaleFactors[0] == 1 && scaleFactors[1] == 0) {
```
dromagnoli: sorry for the silly question: is "scaleFactors[1] == 0" correct?
was wondering if that should be == 1

aaime: You are indeed right, thanks for catching this!

PR #3457, opened 2019-04-29
[GEOS-8624] WFS3: ?resultType=hits support

fernandor777:
This PR adds support for ?resultType=hits parameter in WFS3 module.

https://osgeo-org.atlassian.net/browse/GEOS-8624

### DIFF
--- a/src/community/wfs3/src/main/java/org/geoserver/wfs3/kvp/ResultTypeKvpParser.java
+++ b/src/community/wfs3/src/main/java/org/geoserver/wfs3/kvp/ResultTypeKvpParser.java
@@ -1,0 +1,28 @@
+/* (c) 2019 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.wfs3.kvp;
+
+import net.opengis.wfs20.ResultTypeType;
+import org.geoserver.ows.KvpParser;
+import org.geotools.util.Version;
+
+/**
+ * ResultType parser for WFS 3.0.
+ *
+ * @author Fernando Mino - Geosolutions
+ */
+public class ResultTypeKvpParser extends KvpParser {
+
+    public ResultTypeKvpParser() {
+        super("resultType", ResultTypeType.class);
+        setService("WFS");
+        setVersion(new Version("3.0.0"));
+    }
+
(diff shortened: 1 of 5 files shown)


src/community/wfs3/src/main/java/org/geoserver/wfs3/kvp/ResultTypeKvpParser.java
```diff
@@ -0,0 +1,28 @@
...
+import net.opengis.wfs20.ResultTypeType;
+import org.geoserver.ows.KvpParser;
+import org.geotools.util.Version;
+
+/**
+ * ResultType parser for WFS 3.0.
+ *
+ * @author Fernando Mino - Geosolutions
+ */
+public class ResultTypeKvpParser extends KvpParser {
```
aaime: I believe the gs-wfs ResultTypeKvpParser should already be covering this, is there a need to roll a custom WFS 3 one?

src/wfs/src/main/java/org/geoserver/wfs/response/v2_0/HitsOutputFormat.java
```diff
@@ -34,8 +36,36 @@ public boolean canHandle(Operation operation) {
         return request != null && request.getResultType() == ResultTypeType.HITS;
     }
 
+    public String getMimeType(Object value, Operation operation) throws ServiceException {
```
aaime: Output formats are normally format specific (with a few exceptions, like XSLT or OGR based format generators, where the format is decided at runtime). Can you roll a superclass and have two small subclasses handling the details of GML vs JSON encoding instead?

aaime:
Legit compile failure due to API change:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project gs-nsg-wfs-profile: Compilation failure
[ERROR] /home/travis/build/geoserver/geoserver/src/community/nsg-profiles/nsg-wfs-profile/src/main/java/org/geoserver/nsg/pagination/random/IndexOutputFormat.java:[83,5] method does not override or implement a method from a supertype
```

aaime:
@fernandor777 what should we do about this one?

fernandor777:
Since this PR is throwing compile failure due to API change, I'll close it until check deeply how to fix it correctly.

PR #3458, opened 2019-04-29
[GEOS-9133] hyperspectral image configuration support in "bin" package

fernandor777:
This PR adds hyperspectral image configuration support in "bin" package due to form too big.

This is a draft since an hyperspectral image is required for testing purposes.

https://osgeo-org.atlassian.net/browse/GEOS-9133

### DIFF
--- a/src/release/jetty/etc/jetty.xml
+++ b/src/release/jetty/etc/jetty.xml
@@ -94,7 +94,19 @@
       -->
     </New>
 
-
+    <!-- =========================================================== -->
+    <!-- Set maxFormContentSize and maxFormKeys Request configuration-->
+    <!-- for handling hyperspectral images and large post.           -->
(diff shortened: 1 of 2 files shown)


aaime:
Generally speaking, it looks good to me, a hyperspectral image for tests can be found in the geoserver own sources at geoserver/src/main/src/test/java/org/geoserver/data/test/hyper.tiff

aaime:
@fernandor777 did you check the patch with the hyperspecral file I've indicated?

fernandor777:
Yes, tests with hyper.tiff layer save now works with new default Jetty parameters on both running from Start.java from IDE and Executing Jetty on binary package.

nmco:
Is this ready to be merged @fernandor777 and @aaime ?

PR #3466, opened 2019-05-03
Add sld_body for WMS 2.14.x

JePedrosa:
Hello, I'm trying to load a WMS layer with a sld body that I created dynamically but I always get some length error, I'm trying to pass it by post method but I do not condigo to do it, I'm trying to load it in leaflet. Next I put the code:

            url = url_geoserver + "/wms/reflect?sld_body=" + sld_body;

            window[nombre_tabla] = L.WMS.source(url, {
                format: 'image/gif',
                version: '1.3.0',
                maxZoom: 30,
                transparent: true,
                cql_filter: valor_filtrado,
                zIndex: zIndex_elemento,
                env: env,
                test: val,
                tiled: false,
            });
            window[nombre_tabla].getLayer(nombre_capa).addTo(map);

Thank you


aaime:
Please don't use fake PRs to ask questions, subscribe and actively participate in the user mailing list instead: http://geoserver.org/comm/
This is also likely suitable for the OpenLayers user list.

PR #3467, opened 2019-05-03
Web-resource community module: fix test webpage security

NielsCharlier:


NielsCharlier:
self-merging because 1) change is extremely trivial; 2) community module I am looking after

PR #3469, opened 2019-05-03
GEOS-9200: adding support for setting access type in S3 Blob Store UI

mbarto:


mbarto:
I tried to write a test, but since they were all failing I reverted (thx for the other PR, I will have a look and also add a test for my PR).
I will also add some documentation.

PR #3470, opened 2019-05-06
New community module: WFS FlatGeobuf output format

bjornharrtell:
See https://github.com/bjornharrtell/flatgeobuf for more info about the format.

TODO:
* Figure out what to do with multiple feature collections

Missing upstream:
* ~~Handle MultiLineString and MultiPolygon in Java impl.~~
* ~~Handle attributes in Java impl.~~
* ~~OpenLayers reader POC~~
* ~~Publish flatgeobuf on Maven Central~~
* ~~Publish flatgeobuf on npm~~

### DIFF
--- a/src/community/flatgeobuf/README.md
+++ b/src/community/flatgeobuf/README.md
@@ -1,0 +1,14 @@
+# WFS 3 prototype
+
+This module contains a prototype based on the current draft specification of WFS 3.0, developed as part of the 
+WFS 3.0 hackaton first, and then further developed to match Draft 1 spec and conformance tests, see:
+
+* https://github.com/opengeospatial/WFS_FES
+* https://github.com/opengeospatial/wfs3hackathon/blob/master/Implementations.md
+
+Implementation wise:
+* The module basically acts as an internal proxy around WFS 2.0, using a  servlet filter to adapt protocols. The long term approach would likely be to have a new MVCDispatcher that allows usage of Spring annotations instead (TBD).
+ 
+This implementation contains the following prototype WFS3 extensions:
+* Tiles extension, returning MapBOX/JSON/TopoJSON tiles. It does not cache tiles and will likely be removed when WFS3 is pushed to supported land, but served as a base to study a possible WMTS 2.0 API and WFS 3 tile extension in Testbed 14. Some bits can likely be re-used once a final version comes out.
+* Styles extension, with the ability to get/put/delete styles (must be secured using service security). The API is not really compatible with GeoServer style management and should also be removed, but was used to provide feedback to OGC in the vector tiles pilot, which will likely be used for WMTS 2.0 (and some bits can likely be re-used once a final version comes out).
--- a/src/community/flatgeobuf/README.md
+++ b/src/community/flatgeobuf/README.md
@@ -1,14 +1,5 @@
-# WFS 3 prototype
+# FlatGeobuf WFS output format prototype
 
-This module contains a prototype based on the current draft specification of WFS 3.0, developed as part of the 
-WFS 3.0 hackaton first, and then further developed to match Draft 1 spec and conformance tests, see:
+This module contains a prototype FlatGeobuf WFS output format.
 
-* https://github.com/opengeospatial/WFS_FES
-* https://github.com/opengeospatial/wfs3hackathon/blob/master/Implementations.md
-
-Implementation wise:
-* The module basically acts as an internal proxy around WFS 2.0, using a  servlet filter to adapt protocols. The long term approach would likely be to have a new MVCDispatcher that allows usage of Spring annotations instead (TBD).
- 
-This implementation contains the following prototype WFS3 extensions:
-* Tiles extension, returning MapBOX/JSON/TopoJSON tiles. It does not cache tiles and will likely be removed when WFS3 is pushed to supported land, but served as a base to study a possible WMTS 2.0 API and WFS 3 tile extension in Testbed 14. Some bits can likely be re-used once a final version comes out.
-* Styles extension, with the ability to get/put/delete styles (must be secured using service security). The API is not really compatible with GeoServer style management and should also be removed, but was used to provide feedback to OGC in the vector tiles pilot, which will likely be used for WMTS 2.0 (and some bits can likely be re-used once a final version comes out).
+* https://github.com/bjornharrtell/flatgeobuf
--- a/src/community/flatgeobuf/src/main/java/org/geoserver/wfs/flatgeobuf/FlatGeobufOutputFormat.java
+++ b/src/community/flatgeobuf/src/main/java/org/geoserver/wfs/flatgeobuf/FlatGeobufOutputFormat.java
@@ -5,13 +5,10 @@
  */
 package org.geoserver.wfs.flatgeobuf;
 
-import org.wololo.flatgeobuf.geotools.FeatureCollectionConversions;
 import java.io.IOException;
 import java.io.OutputStream;
-import java.math.BigInteger;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import net.sf.json.JSONException;
 import org.geoserver.config.GeoServer;
 import org.geoserver.platform.Operation;
 import org.geoserver.platform.ServiceException;
@@ -21,6 +18,7 @@ import org.geoserver.wfs.request.FeatureCollectionResponse;
 import org.geotools.data.simple.SimpleFeatureCollection;
 import org.geotools.feature.FeatureCollection;
 import org.opengis.feature.simple.SimpleFeatureType;
+import org.wololo.flatgeobuf.geotools.FeatureCollectionConversions;
 
 /**
  * A GetFeatureInfo response handler specialized in producing FlatGeobuf data for a GetFeatureInfo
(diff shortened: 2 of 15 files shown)


bjornharrtell:
OpenLayers reader POC now done at https://github.com/bjornharrtell/ol3/tree/flatgeobuf.

aaime:
This should really be converted to a community module. I can be persuaded to help you move around stuff, add poms and the like, but the code seems incomplete right now.

bjornharrtell:
Fully agreed @aaime. I've gotten into unexpected toubles at the client side due to https://github.com/google/flatbuffers/issues/5325 that's I'm trying to solve before doing more work here because without a reasonable consumer example this has not that much value.

bjornharrtell:
Resolved client side issues and now able to consume via OpenLayers with the example at https://github.com/bjornharrtell/ol3/blob/flatgeobuf/examples/vector-flatgeobuf.js.

aaime:
Thiking out loud, a easy to use demo could be put toghether by adding the customized OL in a flageobuf module and adding a custom preview link for it, using the extension point being added here:
https://github.com/geoserver/geoserver/pull/3521
@prushforth is working on adding something similar for the mapml module

bjornharrtell:
@aaime, would definitely like that to be part of the community module. I want to get started on that but haven't had the energy and time recently.

bjornharrtell:
This is now in a feature complete state. Still a POC until I can get decided on a stable flatgeobuf spec version. I've looked into creating a custom preview page but it seems quite complex to me.

aaime:
Any news on this one? Starting to see some flatgeobuf examples on twitter.

bjornharrtell:
@aaime: yep it's quite high on my list now that I've gotten flatgeobuf into GDAL. :)

bjornharrtell:
Finally got this updated to the current spec and it looks promising - consuming WFS with outputformat application/flatgeobuf via OpenLayers POC is performing well.

However - it will only be in effect if modding the core as done in https://github.com/geoserver/geoserver/pull/3470/commits/eb4a606035e5c7725e6e4fddec1a1afbcc0c5c87 (and CORS added for OpenLayers).

Theoretically QGIS master can consume this as WFS too, I will try to get around to play with that soon for some real performance comparisons with GML and GeoJSON output.

bjornharrtell:
Ready for review @aaime. Any idea about how to use this without the mentioned core mods?

src/community/flatgeobuf/src/main/java/org/geoserver/wfs/flatgeobuf/FlatGeobufOutputFormat.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /** capabilities output format string. */
+    public String getCapabilitiesElementName() {
+        return "FlatGeobuf";
+    }
+
+    /** Returns the mime type */
+    public String getMimeType(Object value, Operation operation) throws ServiceException {
+        return "application/flatgeobuf";
```
aaime: Should it be "application/x-flatgeobuf"? Or is it already registered at IANA? Just thinking out loud here, not a problem to keep it as is.

src/community/flatgeobuf/src/main/java/org/geoserver/wfs/flatgeobuf/FlatGeobufOutputFormat.java
```diff
@@ -0,0 +1,85 @@
...
+                (SimpleFeatureCollection) featureCollection.getFeature().get(0);
+        FeatureCollectionConversions.serialize(fc, featureCount, output);
+    }
+
+    /**
+     * Is WFS configured to return feature and collection bounds?
+     *
+     * @return
+     */
+    protected boolean isFeatureBounding() {
```
aaime: What is this for?

src/community/flatgeobuf/README.md
```diff
@@ -0,0 +1,5 @@
+# FlatGeobuf WFS output format prototype
```
aaime: README works. If you want more people to find and try out this output format, you can add a half a page doc page and enough to package it as a zip for nightly builds. It's going to be somewhat annoying to do, but extends the potential set of testers from people doing code to people that can unzip archives :-D
Anyways, by no means a requirement, and can be done later.

aaime:
About not changing core, you really don't need much in core. Here is a PR that you can use as a template:
https://github.com/geoserver/geoserver/pull/3873

Mandatory improvements before the PR is suitable for merge:

-  [New profile in web/app]( https://github.com/geoserver/geoserver/pull/3873/files#diff-c7520907c06605253d7d5fc34211b9c2R1749) to make the module available to developers (mandatory)
- [Inclusion in the community build](https://github.com/geoserver/geoserver/pull/3873/files#diff-a86330a2df4b2631f56f8aac3aa6f5b7R282) and [profile to add it by itself in the build](https://github.com/geoserver/geoserver/pull/3873/files#diff-a86330a2df4b2631f56f8aac3aa6f5b7R650)

Optional ones that make it better, but I'm more than happy to merge without them:

- [Documentation page](https://github.com/geoserver/geoserver/pull/3873/files#diff-9fe9e7767e54936ed83427fde1661d8bR1) to help users try it out and [link from the community page summary](https://github.com/geoserver/geoserver/pull/3873/files#diff-326d68dde1203c1a0e3ecabd101ddb84R57)
-  [Assembly descriptor](https://github.com/geoserver/geoserver/pull/3873/files#diff-a13d27ce1e9c30d0ec3b3167f5baf1ceR1) to make a zip available for users of nightly builds, [reference to the descriptor](https://github.com/geoserver/geoserver/pull/3873/files#diff-a86330a2df4b2631f56f8aac3aa6f5b7R102) and [dependency](https://github.com/geoserver/geoserver/pull/3873/files#diff-5722047120a971de700f246992614df5R383) in the ``community/release`` module to have the jars end up where the assembly descriptor can find them

aaime:
Ah, another bit that you may find of interest: https://github.com/geotools/geotools/pull/2689

src/community/flatgeobuf/src/main/java/org/geoserver/wfs/flatgeobuf/FlatGeobufOutputFormat.java
bjornharrtell: Need to involved OCG (or another similar org.) to get application/flatgeobuf, so yes perhaps it should be application/x-flatgeobuf for now.

bjornharrtell:
@aaime got around to add the minimal docs and descriptors in https://github.com/geoserver/geoserver/pull/3470/commits/3e82c480613af8bc401936b49ec6ad8fc7728f5c so perhaps this is ready to merge. :)

src/community/release/ext-flatgeobuf.xml
```diff
@@ -0,0 +1,16 @@
...
+    <formats>
+        <format>zip</format>
+    </formats>
+    <includeBaseDirectory>false</includeBaseDirectory>
+    <fileSets>
+        <fileSet>
+            <directory>release/target/dependency</directory>
+            <outputDirectory></outputDirectory>
+            <includes>
+                <include>gs-flatgeobuf*.jar</include>
```
aaime: You have to list all jars that are needed fro the flatgeobuf output format to work, that are not found in the basic GeoServer installation. At a minimum, I'd expect that the flatgeobuf java library you maintain is a requirement, maybe more?

bjornharrtell: I see, I've added the dependencies but unsure if it's correct to add things that geotools/geoserver already depend on (i.e jts and perhaps json)

aaime: Probably not... if you want to be sure, you can do the following:
* cd src/community
* mvn clean install -DskipTests -PcommunityRelease -T4 (in my case the best number for T is 3/4 the number of cores available)
* mvn assembly:attached -nsu

You'll find the zip for your module in target/release. Try to stick it into a GeoServer nightly build and see if it starts and works as expected. https://build.geoserver.org/geoserver/master/

Rinse and repeat until it does :-D

bjornharrtell: Will do :)

bjornharrtell: Yep that was a good exercise, found and fixed issues and successfully tested locally.

aaime:
Selezione_999(058)  :smile:

bjornharrtell:
Added doc to index and rebased on master

bjornharrtell:
Hmm code style error but it doesn't tell me what is wrong and I fail to understand how to do the check locally :s

```
[INFO] Processed 1 files (1 non-complying).
[ERROR] Non complying file: /home/travis/build/geoserver/geoserver/src/community/flatgeobuf/src/main/java/org/geoserver/wfs/flatgeobuf/FlatGeobufOutputFormat.java
[INFO] FAILURE build of project org.geoserver.community:gs-flatgeobuf
```

bjornharrtell:
Hmm mabye `mvn fmt:format` does the trick.. :)

aaime:
It just really tells you the file. Exactly, I see you figured it out.
Generally speaking, "mvn install" does run the formatting, and one is supposed to run a build on the modified code, and its dependencies, when doing a PR.

bjornharrtell:
@aaime this looks green now, AppVeyor build failure seem unrelated.

aaime:
Fire! Time to do some noise about it then. The module will show up in the docs and in the nightly builds in a few hours. Any client app that can be used to play with it?

bjornharrtell:
GDAL master and QGIS nightly for Windows has support but I wouldn't expect it to work out of the box via WFS but now I'm motivated to look into that and make the needed changes.

Consuming in OpenLayers is rather straight forward via flatgeobuf ref implementation but example code only lives in my experimental branch at https://github.com/bjornharrtell/ol3/blob/flatgeobuf/examples/vector-flatgeobuf.js

Then there is this naive d3 example: https://observablehq.com/@bjornharrtell/streaming-flatgeobuf

bjornharrtell:
Tracking work to consume this in GDAL at https://github.com/OSGeo/gdal/pull/2135. Should eventually enable this to be used via QGIS.

PR #3477, opened 2019-05-07
GEOS-9198: environment properties support in authkey mapper parameters

mbarto:


src/community/authkey/src/main/java/org/geoserver/security/AuthenticationKeyFilterConfig.java
```diff
@@ -75,4 +81,39 @@ public void setUserGroupServiceName(String userGroupServiceName) {
     public void setMapperParameters(Map<String, String> mapperParameters) {
         this.mapperParameters = mapperParameters;
     }
+
+    public void setAllowEnvParametrization(boolean allowEnvParametrization) {
```
aaime: What is going to set this property? I thought env parametrization was controlled by a single system variable, "-DALLOW_ENV_PARAMETRIZATION=true", see https://docs.geoserver.org/stable/en/user/community/backuprestore/configtemplate.html

Code wise it seems everything is referring to GeoServerEnvironment.ALLOW_ENV_PARAMETRIZATION.

src/community/authkey/src/test/java/org/geoserver/security/AuthKeyAuthenticationTest.java
```diff
@@ -142,6 +143,43 @@ public void testMapperParameters() throws Exception {
         assertEquals("value2", fakeMapper.getMapperParameter("param2"));
     }
 
+    @Test
+    public void testMapperParametersFromEnv() throws Exception {
+        String authKeyUrlParam = "myAuthKeyParams";
+        String filterName = "testAuthKeyParams2";
+
+        AuthenticationKeyFilterConfig config = new AuthenticationKeyFilterConfig();
+        config.setAllowEnvParametrization(true);
```
aaime: Following up with the above, the system variable can be set in a @Before and reset in a @After

src/community/authkey/src/main/java/org/geoserver/security/AuthenticationKeyFilterConfig.java
mbarto: used by the tests

src/community/authkey/src/test/java/org/geoserver/security/AuthKeyAuthenticationTest.java
mbarto: in theory, in practice it didn't work, I suppose because ALLOW_ENV_PARAMETRIZATION is statically initialized, do you know of another way to make it work?

src/community/authkey/src/main/java/org/geoserver/security/AuthenticationKeyFilterConfig.java
```diff
@@ -75,4 +81,39 @@ public void setUserGroupServiceName(String userGroupServiceName) {
...
+                        target.getMapperParameters().put(key, (String) value);
+                    }
+                }
+            }
+        }
+
+        return target;
+    }
+
+    protected boolean isEnvParametrizationAllowed() {
```
aaime: Ah, I missed this bit.

src/community/authkey/src/test/java/org/geoserver/security/AuthKeyAuthenticationTest.java
aaime: You're right, the system variable is static final... gah, not test friendly at all!
The variable should be made non final I believe, or you can try to force the test to run with a different classloader, see:

https://stackoverflow.com/questions/42102/using-different-classloaders-for-different-junit-tests

Also, I believe setting the variable should be done in a @BeforeClass to make that case work. @afabiani  ?

mbarto: @aaime @afabiani made a new commit, seems to be working this way

PR #3479, opened 2019-05-08
[GEOS-9205] Failure to encoding GeoJSON when returning a complex feature out of the single feature WFS3 endpoint

aaime:


src/community/wfs3/src/test/java/org/geoserver/wfs3/ComplexFeatureTest.java
```diff
@@ -0,0 +1,73 @@
+/*
```
nmco: Wrong project :yum:

src/community/wfs3/pom.xml
```diff
@@ -145,6 +145,43 @@
       <artifactId>jsoup</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.geotools</groupId>
+      <artifactId>gt-app-schema</artifactId>
+      <version>${gt.version}</version>
+    </dependency>
```
nmco: So WFS 3 depends on App-Schema now? Just making sure this is not a forgotten dependency.

nmco:
Just some small comments, otherwise looks good.

src/community/wfs3/pom.xml
aaime: Needed to run a request based on app-schema... ah, but you're right, could have been a test dependency. Fixing.

src/community/wfs3/src/test/java/org/geoserver/wfs3/ComplexFeatureTest.java
aaime: Right, intellij automatic headers hit again.

aaime:
Updated

PR #3480, opened 2019-05-08
Fix for LDAP modules testing with Apache Directory embedded server.

fernandor777:
This PR Refactors and fixes the LDAP Junit testing setup for embedded Apache Directory server and optimizes testing times by reusing server startups and ldiff setups using its testing annotations framework and upgrading versions.

Now LDAP integration testing is active again.

See:
https://directory.apache.org/apacheds/advanced-ug/7-embedding-apacheds.html

PR #3487, opened 2019-05-10
[GEOS-9199]  Hierarchical LDAP Groups Support

fernandor777:
Currently Geoserver LDAP module supports direct user to group role binding, but does not link the user with any upper related group containing (with member attribute) its direct parent group.  

This PR adds hierarchical LDAP groups support enhancing the existing classes.

Requires:
https://github.com/geoserver/geoserver/pull/3480

https://osgeo-org.atlassian.net/browse/GEOS-9199


src/security/ldap/src/main/java/org/geoserver/security/ldap/BindingLdapAuthoritiesPopulator.java
```diff
@@ -146,6 +160,8 @@ public BindingLdapAuthoritiesPopulator(ContextSource contextSource, String group
         if (password != null) {
             // authenticate and execute role extraction in the authenticated
             // context
+            Supplier<DirContext> ctxSupplier =
```
mbarto: Is this needed? Why don't you use the authenticate ctx from ldapTemplate.authenticate anymore?

src/security/ldap/src/main/java/org/geoserver/security/ldap/LDAPSecurityProvider.java
```diff
@@ -103,6 +104,13 @@ public GeoServerAuthenticationProvider createAuthenticationProvider(
                 authPopulator =
                         new BindingLdapAuthoritiesPopulator(
                                 ldapContext, ldapConfig.getGroupSearchBase());
+                // set hierarchical configurations
+                BindingLdapAuthoritiesPopulator bindPopulator =
```
mbarto: This is not used anywhere, am I wrong?

fernandor777: It is only used as re-typed reference for setting some attributes:
```
BindingLdapAuthoritiesPopulator bindPopulator =
                        (BindingLdapAuthoritiesPopulator) authPopulator;
bindPopulator.setUseNestedParentGroups(ldapConfig.isUseNestedParentGroups());
(...)
```

mbarto: sorry, correct

src/security/ldap/src/main/java/org/geoserver/security/ldap/BindingLdapAuthoritiesPopulator.java
fernandor777: Indeed, using authenticate is the right way.  This was used because reuse Context objects is not allowed after doing a search.  Fixing it, thanks.

fernandor777: Refactored to use a functional consumer that wraps authentication process and propagate to its called methods.

mbarto: ok, for the future I would prefer if you don't force push, so that I can see in a new commit the changes

fernandor777: Ok, makes sense.

src/security/ldap/src/main/java/org/geoserver/security/ldap/LDAPUserGroupService.java
```diff
@@ -327,9 +406,68 @@ public void executeWithContext(
...
+                    @Override
+                    public void executeWithContext(
+                            DirContext ctx, LdapEntryIdentification ldapEntryIdentification) {
+                        final String dn =
+                                LDAPUtils.getLdapTemplateInContext(ctx, template)
+                                        .searchForSingleEntry(
+                                                groupSearchBase,
+                                                groupNameFilter,
+                                                new String[] {groupName})
+                                        .getDn()
```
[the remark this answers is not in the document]
fernandor777: I agree.  Previous versions of lookupDN method returns relative DN:

https://github.com/geoserver/geoserver/blob/445625cdad44e79c48682b0c9852e1e2586c018e/src/security/ldap/src/main/java/org/geoserver/security/ldap/LDAPBaseSecurityService.java#L236

I'll check this change doesn't break former code/tests.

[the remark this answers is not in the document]
fernandor777: Changing lookupDn method to return full dn breaks some former tests:
```
Failed tests:   testGetRolesForUserUsingUserFilter(org.geoserver.security.ldap.LDAPRoleServiceTest$LDAPRoleServiceLdiff2Test)
  testGetRolesForUserAuthenticatedUsingUserFilter(org.geoserver.security.ldap.LDAPRoleServiceTest$LDAPRoleServiceLdiff2Test)
  testHierarchicalGroupsForUser(org.geoserver.security.ldap.LDAPUserGroupServiceTest): expected:<2> but was:<0>
  testGroupsForUser(org.geoserver.security.ldap.LDAPUserGroupServiceTest): expected:<1> but was:<0>
```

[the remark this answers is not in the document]
mbarto: Please try to understand if the tests are correct, or just expect relative dn as a consequence of the code returning it.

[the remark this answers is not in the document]
fernandor777: Analyzing lookupDn method is used in:
```
	org.geoserver.security.ldap.LDAPUserGroupService.getGroupsForUser(GeoServerUser)
	org.geoserver.security.ldap.LDAPRoleService.getRolesForUser(String)
```	
One time LDAPUserGroupService and one time on LDAPRoleService, but on heavily reused methods.
That's why these services are configured in different way than authorization, for example requiring adding domain parts to the groupSearch pattern:
```
member={1},dc=example,dc=org
```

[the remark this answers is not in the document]
mbarto: Ok, understood, let's leave it like it is then

src/security/ldap/src/main/java/org/geoserver/security/ldap/BindingLdapAuthoritiesPopulator.java
```diff
@@ -186,33 +216,123 @@ public void executeWithContext(
...
+            String dn = roleNameDn.getRight();
+            if (convertToUpperCase) {
+                role = role.toUpperCase();
+            }
+            boolean addedSuccesfuly =
+                    authorities.add(new SimpleGrantedAuthority(rolePrefix + role));
+            // search nested Ldap groups,
+            // only if role was added successfully for avoiding circular references
+            // if maxGroupSearchLevel == -1 -> no depth limit
+            if (maxGroupSearchLevel == -1 || (depth > 0 && addedSuccesfuly))
```
nmco: So if there is no depth limit, we don't care about circular dependencies?

simboss:
@fernandor777 please, address the feedback so that we can move forward and backport.

src/security/ldap/src/main/java/org/geoserver/security/ldap/BindingLdapAuthoritiesPopulator.java
fernandor777: Right, fixed. Thanks.

PR #3488, opened 2019-05-10
Encode the code attribute in WMS service exception reports.

sikeoka:
This pull request is to XML encode the service exception code before writing it into the WMS service exception report.


aaime:
Did not check, is the content type of the exception set? Looks otherwise good.

sikeoka:
Yes, it is setting a content type.

aaime:
Good thanks!

PR #3492, opened 2019-05-13
[GEOS-9207] REST exception handler and controllers do not always set the response content type

aaime:


src/extension/geofence-server/src/main/java/org/geoserver/geofence/server/rest/RulesRestController.java
```diff
@@ -206,7 +208,9 @@ public JaxbRuleList count(
             adminService.setDetails(id, rule.getLayerDetails().toLayerDetails(null));
         }
 
-        return new ResponseEntity<Long>(id, HttpStatus.CREATED);
+        HttpHeaders headers = new HttpHeaders();
+        headers.setContentType(MediaType.TEXT_PLAIN);
```
sikeoka: These headers aren't used.

src/rest/src/main/java/org/geoserver/rest/RestInterceptor.java
```diff
@@ -23,4 +24,14 @@ public boolean preHandle(
...
+
+    @Override
+    public void postHandle(
+            HttpServletRequest request,
+            HttpServletResponse response,
+            Object handler,
+            ModelAndView modelAndView)
+            throws Exception {
+        super.postHandle(request, response, handler, modelAndView);
+    }
```
sikeoka: This doesn't seem to be doing anything useful?

src/restconfig/src/test/java/org/geoserver/rest/catalog/StyleControllerTest.java
```diff
@@ -416,7 +424,7 @@ public void testPostToWorkspace() throws Exception {
         String xml = "<style>" + "<name>foo</name>" + "<filename>foo.sld</filename>" + "</style>";
         MockHttpServletResponse response =
                 postAsServletResponse(RestBaseController.ROOT_PATH + "/workspaces/gs/styles", xml);
-        assertEquals(201, response.getStatus());
```
sikeoka: Is removing the assert here necessary?

src/extension/geofence-server/src/main/java/org/geoserver/geofence/server/rest/RulesRestController.java
aaime: Right, the change in "produces" does the trick, not this bit, I'll remove.

src/rest/src/main/java/org/geoserver/rest/RestInterceptor.java
aaime: True, leftover from a failed approach.

src/restconfig/src/test/java/org/geoserver/rest/catalog/StyleControllerTest.java
aaime: No, copy/paste error while I was adding the assertions around.

aaime:
Feedback from @sikeoka review applied.

PR #3501, opened 2019-05-15
GEOS-9216: Vertical gap with APH

dromagnoli:


simboss:
do not forget to backport this one.

src/wms/src/main/java/org/geoserver/wms/map/RenderedImageMapOutputFormat.java
```diff
@@ -1507,7 +1507,7 @@ private RenderedImage applyBackgroundTransparency(
         if (roiCandidate instanceof ROI) {
             ROI imageROI = (ROI) roiCandidate;
             try {
-                roi = imageROI.intersect(new ROIGeometry(mapRasterArea));
+                roi = new ROIGeometry(mapRasterArea).intersect(imageROI);
```
mbarto: Can you explain why inverting the intersection you get a "better" result? They are not using the same intersection alghoritm I suppose...

dromagnoli: Right. The intersection goes through a slightly different code path
Moreover, with this change the ROIGeometry will also check if the intersecting ROI is a geometry and use a geometry intersection in that case.

PR #3502, opened 2019-05-15
[GEOS-9218] Parameter Extractor plugin prevents the Monitor plugin to log requests

fernandor777:
If the Parameter Extraction plugin and the Monitor Plugin are installed, every request that is rewritten by the Parameter Extraction plugin is not logged in the Monitoring Plugin audit files.

This PR fix RequestWrapper class and its path extract methods, so Monitor module can read its paths properly.

See:
https://osgeo-org.atlassian.net/browse/GEOS-9218

### DIFF
--- a/src/community/params-extractor/src/main/java/org/geoserver/params/extractor/RequestWrapper.java
+++ b/src/community/params-extractor/src/main/java/org/geoserver/params/extractor/RequestWrapper.java
@@ -8,6 +8,7 @@ import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import javax.servlet.http.HttpServletRequest;
@@ -18,8 +19,6 @@ public final class RequestWrapper extends HttpServletRequestWrapper {
     private final UrlTransform urlTransform;
     private final Map originalParameters;
 
-    private final Pattern pathsPattern;
-
     private final String pathInfo;
     private final String servletPath;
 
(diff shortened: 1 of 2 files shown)


src/community/params-extractor/src/main/java/org/geoserver/params/extractor/RequestWrapper.java
```diff
@@ -91,15 +93,15 @@ public String getParameter(String name) {
     }
 
     private String extractPathInfo(String requestUri) {
-        Matcher matcher = pathInfoPattern.matcher(requestUri);
+        Matcher matcher = pathsPattern.matcher(requestUri);
```
aaime: Basically these methods are now repeating the same thing, wouldn't it make more sense to build the matcher just one in the calling method, set the variables accordingly, and remove these two methods?

src/community/params-extractor/src/main/java/org/geoserver/params/extractor/RequestWrapper.java
```diff
@@ -30,8 +29,11 @@ public RequestWrapper(UrlTransform urlTransform, HttpServletRequest request) {
         super(request);
         this.urlTransform = urlTransform;
         originalParameters = request.getParameterMap();
-        pathInfoPattern = Pattern.compile("^" + request.getContextPath() + "([^/]+?).*$");
-        servletPathPattern = Pattern.compile("^" + request.getContextPath() + "[^/]+?/([^/]+?).*$");
+        pathsPattern =
+                Pattern.compile(
+                        "^"
+                                + request.getContextPath()
+                                + "(/[^\\?^/]+)(/[^\\?]*[^/^\\?])?/?(\\??.*)?$");
```
aaime: Whoa this regex is evil... I put it into a tool that explains regexes, and still not sure why it's working:

image

The third capturing group seems to be non necessary (nothing using it?), and it would seem like the regex allows for multiple question marks.

Tool here: http://xenon.stanford.edu/~xusch/regexp/

fernandor777: Thanks for the awesome tool.  Regular expression improved with non capturing groups and added boorkmark char support (#).

src/community/params-extractor/src/main/java/org/geoserver/params/extractor/RequestWrapper.java
fernandor777: Indeed, moved to root caller method, also maintaining pattern object as class field not needed anymore.

fernandor777:
I made a regex simplification and added null-safe checks, thanks for the review.

aaime:
Looks good to me

simboss:
can you merge andrea?

aaime:
The GWC failure is the usual random test failing, merging.

PR #3506, opened 2019-05-15
[GEOS-9199] Hierarchical LDAP Groups Support - 2.15.x backport

fernandor777:
Currently Geoserver LDAP module supports direct user to group role binding, but does not link the user with any upper related group containing (with member attribute) its direct parent group.

This PR adds hierarchical LDAP groups support enhancing the existing classes.

2.15.x backport.

https://osgeo-org.atlassian.net/browse/GEOS-9199


aaime:
Was there a backport request on list? I don't remember...

aaime:
It would be backportable now, if only there had been a request on the devel list to discuss the backport of it....

PR #3517, opened 2019-05-17
Add extensibility of common preview formats via beans

prushforth:
## Description

Allow format modules to add a bean to a layer preview to the "Common Formats" list of links on the Layer Preview page.  Built specifically for the MapML community module.

## Checklist

For all pull requests:

-> change is not visible to end-users, except those who have a module installed that uses the extension point
-> no JIRA ticket, if needed could create one
-> a test failed due to my local re-ordering of links, so I updated the test to be aware of the new wicket component and it continues to test order of the 'standard' format links.
-> uncertain if this warrants a documentation update. Let me know.

### DIFF
--- a/src/web/demo/src/main/java/org/geoserver/web/demo/CommonFormatLink.java
+++ b/src/web/demo/src/main/java/org/geoserver/web/demo/CommonFormatLink.java
@@ -1,0 +1,59 @@
+/* (c) 2019 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.web.demo;
+
+import org.apache.wicket.markup.html.link.ExternalLink;
+import org.geoserver.web.ToolLinkExternalInfo;
+
+/**
+ * Extension point for MapPreviewPage. Subclasses will implement getFormatLink which returns an
+ * ExternalLink to a layer preview in the subclass format represented by the subclass e.g.
+ * GMLFormatLink.
+ *
+ * @author prushforth
+ */
+public abstract class CommonFormatLink extends ToolLinkExternalInfo
+        implements Comparable<CommonFormatLink> {
+
+    private final String componentId = "theLink";
+
+    private int order = 1000;
+
+    public CommonFormatLink() {
+        super();
+    }
+
+    /**
+     * Returns an ExternalLink object that is used to link to the layer preview.
+     *
+     * @param layer the PreviewLayer object for which the preview link is returned
+     * @return
+     */
+    public abstract ExternalLink getFormatLink(PreviewLayer layer);
+
+    @Override
+    public void setComponentClass(Class<ExternalLink> componentClass) {}
+
+    public String getComponentId() {
+        return componentId;
+    }
+
+    /**
+     * @param order orders the list of common formats, by default new formats will be added at the
+     *     of the list.
+     */
+    public void setOrder(int order) {
+        this.order = order;
+    }
+
+    public int getOrder() {
+        return order;
+    }
+
+    @Override
+    public int compareTo(CommonFormatLink other) {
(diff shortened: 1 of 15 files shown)


aaime:
The PR checklist is new and still being worked on, the PRs opened in these weeks are a guinea pigs for it, sorry :-D. 

In case of a new extension point, it definitely needs a ticket, the question normally is whether or not it needs a full [GeoServer Improvement Proposal](https://github.com/geoserver/geoserver/wiki/Proposals).
Since it's small I'm not too concerned about it, but let's hear from @jodygarnett too.

In the meantime I'll have a review of the changes. And oh, builds are all red too, with tests failing, so definitely something is going on.

src/web/demo/src/main/java/org/geoserver/web/demo/CommonFormatLink.java
```diff
@@ -0,0 +1,40 @@
+/* (c) 2019 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.web.demo;
+
+import org.apache.wicket.markup.html.link.ExternalLink;
+import org.geoserver.web.ToolLinkExternalInfo;
+
+public abstract class CommonFormatLink extends ToolLinkExternalInfo
```
aaime: Add a javadoc for new public classes please, especially since this is an extension point that plugins can implement.

src/web/demo/src/main/java/org/geoserver/web/demo/CommonFormatLink.java
```diff
@@ -0,0 +1,40 @@
...
+        implements Comparable<CommonFormatLink> {
+
+    private final String componentId = "theLink";
+    private int order = 1000;
+
+    public CommonFormatLink() {
+        super();
+    }
+
+    public abstract ExternalLink getFormatLink(
```
aaime: Abstract methods also need documentation, plugin implementors need to implement them.

I'm worried about the "wmsVisible, wfsVisible" bits, what if we want to extend the preview page later, we'd have to break this method signature. Maybe best to expose the layer contained in PreviewLayer instead, and have the extension point compute the visibility on its own instead?  The MapPreviewPage.hasServiceSupport method could be moved into the WMS facade class for reuse in both places.

aaime:
Build failures happen in the app-schema-tests extension:

```
Failed tests:   testGml311Links(org.geoserver.test.web.Gml311LinksTest)
  testGml32Links(org.geoserver.test.web.Gml32LinksTest)
```

src/web/demo/src/main/java/org/geoserver/web/demo/CommonFormatLink.java
prushforth: OK

src/web/demo/src/main/java/org/geoserver/web/demo/CommonFormatLink.java
prushforth: I'll have a go at that

prushforth:
Oops I missed that test, as I only ran the tests for the demo module. I'll see if that can be fixes without messing about too much.

prushforth:
I will fix this up, probably next week. Thanks for your review

src/web/demo/src/main/java/org/geoserver/web/demo/CommonFormatLink.java
prushforth: What is the WMS facade class?

prushforth: I moved hasServiceSupport into PreviewLayer, if that is what you meant.  Seems to work ok, and those parameters are no longer required.  I'll push it and you can tell me what you think.

PR #3521, opened 2019-05-20
[GEOS-9222] Permit extensibility of Common Formats from Layer Preview page

prushforth:
## Description

Allow extension modules to add to the set of "Common Formats" links in the Layer Preview page on a per-layer basis.

Note that one (important) file MapPreviewPage.html shows itself to be 100% different from that in master.  I verified my settings for core autocrlf=true to be correct.  I looked at the history of the file and it appears that the previous check-in of the file also showed a 100% difference.   So this check-in may fix that problem from now on, so long as devs have the correct autocrlf setting locally (per the [recommendations](https://docs.geoserver.org/latest/en/developer/source.html#git-client-configuration)).

## Checklist

For all pull requests:

### DIFF
--- a/src/web/demo/src/main/java/org/geoserver/web/demo/MapPreviewPage.html
+++ b/src/web/demo/src/main/java/org/geoserver/web/demo/MapPreviewPage.html
@@ -14,7 +14,9 @@
   
   <!-- The fragment for the common links -->
   <wicket:fragment wicket:id="commonLinks">
-  <div style="white-space: nowrap;"><a target="_blank" wicket:id="ol"></a>&nbsp;<a target="_blank" wicket:id="kml"></a>&nbsp;<a target="_blank" wicket:id="gml"></a></div>
+    <span style="white-space: nowrap;" wicket:id="commonFormat">
+      <a target="_blank" href="#" wicket:id="theLink">theTitle</a>&nbsp;
+    </span>
   </wicket:fragment>
   
   <!--  The custom drop down with optgroups -->
--- a/src/web/demo/src/test/java/org/geoserver/web/demo/MapPreviewPageTest.java
+++ b/src/web/demo/src/test/java/org/geoserver/web/demo/MapPreviewPageTest.java
@@ -212,8 +212,14 @@ public class MapPreviewPageTest extends GeoServerWicketTestSupport {
                     path = c.getPageRelativePath();
 
                     // check visible links
-                    ExternalLink olLink = (ExternalLink) c.get("itemProperties:3:component:ol");
-                    ExternalLink gmlLink = (ExternalLink) c.get("itemProperties:3:component:gml");
+                    ExternalLink olLink =
+                            (ExternalLink)
+                                    c.get("itemProperties:3:component:commonFormat:0")
+                                            .getDefaultModelObject();
+                    ExternalLink gmlLink =
+                            (ExternalLink)
+                                    c.get("itemProperties:3:component:commonFormat:1")
+                                            .getDefaultModelObject();
 
                     assertEquals(
                             "http://localhost:80/context/cite/wms?service=WMS&amp;version=1.1.0&amp;request=GetMap&amp;layers=cite%3ALakes%20%2B%20a%20plus&amp;bbox=-180.0%2C-90.0%2C180.0%2C90.0&amp;width=768&amp;height=384&amp;srs=EPSG%3A4326&amp;format=application/openlayers",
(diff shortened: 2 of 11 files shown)


src/web/demo/src/test/java/org/geoserver/web/demo/MapPreviewPageTest.java
```diff
@@ -246,20 +252,4 @@ public void testNameURLEncoding() {
             catalog.save(ft);
         }
     }
-
-    /** Test for layer group service support check */
-    @Test
-    public void testHasServiceSupport() throws Exception {
```
aaime: GeoServer tests are slow to start due to partial application startup, moving them around in their own test class slows the build for everybody. What was the reason for the move?

src/web/demo/src/main/java/org/geoserver/web/demo/MapPreviewPage.html
```diff
@@ -1,34 +1,36 @@
...
-      </optgroup>
-      <optgroup label="WFS" wicket:id="wfs">
-           <option wicket:id="wfsFormats"></option>
-       </optgroup>
-    </select>
-  </wicket:fragment>
-</wicket:extend>
-</body>
-</html>
+<html xmlns:wicket="http://wicket.apache.org/">
```
aaime: Full reformat again, please avoid.

aaime: Btw, the reason why we don't want full reformats is simple, GeoServer has a long history and we use "git blame" a lot to find where some code originated while looking at code for bug fixes and new feature development purposes, a change like this is like a wall in the history. Sure, tooling allows to blame the revision before a particular change, but it's needlessly annoying.

aaime:
Btw @prushforth as you probably have noticed I have time to look at non work related PRs only in the weekends (outside working hours), and not all weekends (sometimes I have family keeping me fully busy, sometimes I just want to develop something on my own instead of just reviewing other people's code). If we had more people doing this we'd be faster, but I'm afraid you're stuck with me and the little time I can dedicate to this activity.

src/web/demo/src/main/java/org/geoserver/web/demo/MapPreviewPage.html
prushforth: Hi Andrea,

I believe that the full reformat here is caused by a previous full reformat on this file in the [previous commit](https://github.com/geoserver/geoserver/commit/1d6ddea79ba5c5705dbe10100fda2616622ac4fa#diff-ea942e9262fed375f477a322e498b5c8).  Wouldn't it be better to fix this problem now instead of everyone who changes that file having the same problem?  If you mix line ending styles within the one file, you can avoid the full reformat, but then the next person would have to figure that situation out again, which would waste people's time like this file is currently wasting both our time...

prushforth:
> Btw @prushforth as you probably have noticed I have time to look at non work related PRs only in the weekends (outside working hours), and not all weekends (sometimes I have family keeping me fully busy, sometimes I just want to develop something on my own instead of reviewing other people's code). If we had more people doing this we'd be faster, but I'm afraid you're stuck with me and the little time I can dedicate to this activity.

I totally appreciate the great comments you have given me, and I apologize if it seems like I'm wasting your time.  I'm trying to contribute positively, and not with hacks, and I am a bit green on the collaboration side of things so I do appreciate all the efforts you've made.

src/web/demo/src/test/java/org/geoserver/web/demo/MapPreviewPageTest.java
prushforth: You suggested to move the hasServiceSupport method into the WMS facade class (which I did not know what that was).  However, I found that moving the method into the PreviewLayer class (and making it public) had the effect of making it available in the MapPreviewPage.java where it came from and was still needed, as well as in the CommonFormatBean and its subclasses, by virtue of it being a parameter of the abstract getFormatLink(PreviewLayer layer) method, where it is also needed.  That move eliminated the wmsVisible and wfsVisible parameters from the getFormatLink(...) call, and now the implementations of CommonFormatBean can calculate their own visibility as needed with no / minimal overhead.

So I created a test class for the PreviewLayer class, (which did not have any tests), in order to move the test method.  Given that PreviewLayer doesn't have any other tests, I would be willing to write some, if you think it's a good idea, and if you'll give me some pointers on what might be useful.  Maybe I used the wrong base class?  Could a different base class reduce the overhead of setup in this case?

aaime: Hum... if you move to have a more unit test in the other class, please still leave the integration test here, GeoServer has no testers and no form of functional testing, so integration tests like this one are the only proof that something is actually working. Unit testing can also be added of course, but no development is complete until there are integration tests making sure there are no regressions.

That said, hasServiceSupport needs a fully started GeoServer, I don't see an obvious way to mock around it (though there are likely long/complicated ways to do so).

GeoServer is a very old application (over 15 years), working with it calls for practicality rather than perfectionism.

prushforth: ok, I'll leave it where it was.

aaime:
Well, all seems in order, merging

prushforth:
Thank you!

PR #3523, opened 2019-05-22
[GEOS-9218] Parameter Extractor plugin prevents the Monitor plugin to log requests (2.14.x backport)

fernandor777:
Parameter Extractor plugin prevents the Monitor plugin to log requests.

## Description

If the Parameter Extraction plugin and the Monitor Plugin are installed, every request that is rewritten by the Parameter Extraction plugin is not logged in the Monitoring Plugin audit files.

This PR fixes RequestWrapper class and its path extract methods, so Monitor module can read its paths properly.

JIRA issue:
https://osgeo-org.atlassian.net/browse/GEOS-9218

### DIFF
--- a/src/community/params-extractor/pom.xml
+++ b/src/community/params-extractor/pom.xml
@@ -28,6 +28,11 @@
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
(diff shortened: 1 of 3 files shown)


fernandor777:
Conflict resolution in progress (due to previous PR merged and reverted).

fernandor777:
Conflict solved, backport PR ready.

PR #3529, opened 2019-05-25
WFS3: add tests for YAML.

bradh:
Re-uses JSON tests where appropriate.


aaime:
Doh, nice way to run the tests, did not think about it. Merging.

PR #3544, opened 2019-06-03
Add MapML preview to Preview Layers 'Common Formats' list

prushforth:
## Description

Add Preview Layers visibility to the MapML Community Module, using the Preview Layers 'Common Format' extensibility mechanism from GEOS-9222. @cmhodgson please review; squashing all the commits will be appropriate on merge, I believe. @jodygarnett I have added license headers to all the client files per our discussion, and I have pared down the set of client files to the bare minimum, I think.  I sent the last pull request of a feature branch which failed the travis and appveyor build checks.  I think this one should work, as I've integrated the commits on top of upstream master.

## Checklist

For all pull requests:

A few unit tests included.  Needs more tests.  Interactively tested so far.


src/community/mapml/src/main/java/org/geoserver/mapml/MapMLConfigurationPanel.java
```diff
@@ -1,4 +1,4 @@
-/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2019 Open Source Geospatial Foundation - all rights reserved
```
jodygarnett: thanks!

src/community/mapml/src/main/java/org/geoserver/mapml/MapMLController.java
```diff
@@ -82,9 +219,44 @@ public Mapml mapML(
             @RequestParam("format") Optional<String> format)
             throws NoSuchAuthorityCodeException, TransformException, FactoryException, IOException {
         LayerInfo layerInfo = geoServer.getCatalog().getLayerByName(layer);
-        if (layerInfo == null) {
-            // TODO better error handling
-            throw new RuntimeException("Invalid layer name: " + layer);
+        // TODO Is DefaultGeographicCRS.WGS84 compatible with Web Mercator / EPSG:3857 ?
+        // If I understand the remarks on www.epsg-registry.org on this account,
+        // the answer is yes, but EPSG:3857 sucks and don't use it. sigh.
```
jodygarnett: Like to remove TODO's if we can but this is just a community module.

Is there any way resolve your unhappiness with DefaultGeographicCRS.WGS84?

src/community/mapml/pom.xml
```diff
@@ -36,6 +41,11 @@
       <groupId>org.geoserver</groupId>
       <artifactId>gs-wms</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.locationtech.proj4j</groupId>
+      <artifactId>proj4j</artifactId>
+      <version>1.0.0</version>
```
jodygarnett: Two things here:

- we keep version numbers in the root pom.xml in a dependency management section (if not we get madness)
- geotools includes its own referencing implementation so proj4j should not be necessary (why is it being used?)

jodygarnett:
@cmhodgson since this is a community module it is kind of yours to manage until such time as you want to submit it as an extension. Can I ask that you provide your name and contact details to the root pom.xml (so potential contributors know who to contact for review).

This is quite exciting functionality and I would love for it to appear in a future release of GeoServer.

src/community/mapml/pom.xml
prushforth: proj4 is used because I had existing code that used it.  I will have a look to see about refactoring the dependency to be geotools instead of proj4j.  Also I'll remove the ${project.version} reference.

jodygarnett: Okay, most anything you need to do with projections can be accomplished using the [CRS utility class](http://docs.geotools.org/latest/userguide/library/referencing/crs.html) and [JTS](https://docs.geotools.org/latest/userguide/library/main/jts.html) utility class.

src/community/mapml/src/main/java/org/geoserver/mapml/tcrs/TiledCRS.java
```diff
@@ -0,0 +1,284 @@
...
+                }
+            }
+        }
+        // the centre of the extent in decimal tiles... not truncated
+        Point centre = pb.getCentre().divideBy(TILESIZE);
+        Collections.sort(tiles, new TileComparator(centre));
+        return tiles;
+    }
+
+    public List<TileCoordinates> getTilesForExtent(LatLngBounds extent, int zoom, long start) {
```
jodygarnett: If it helps geotools has a WMTS client extension that has very similar logic to what you are accomplishing here.

prushforth: I'll have a look, thanks!

src/community/mapml/src/main/java/org/geoserver/mapml/MapMLController.java
prushforth: I'm not unhappy with DefaultGeographicCRS.WGS84, however the epsg registry says 3857 uses "spherical development" of WGS84, which I find to be a bit opaque in the sense that I am wondering if legitimately I can use a geographic bounding box based on DefaultGeographicCRS.WGS84 with the subsequent code: bbox.expandToInclude(li.getResource().getLatLonBoundingBox()) wherein I'm getting lat-lon bounding boxes and adding them to the default one, but maybe those bounding boxes are incompatible, so that's my main concern.   The TODO should really read "Learn Geotools".

prushforth:
I believe the build failures are caused by some encoding incompatibility between GWC and gs-rest modules:  

Failed tests:   testPutBadLayerEndpoint(org.geoserver.gwc.RESTIntegrationTest): expected:<[T]here is a mismatch b...> but was:<[]here is a mismatch b...>
  testDeleteNonExistentLayer(org.geoserver.gwc.RESTIntegrationTest): expected:<[Unknown layer: ]badLayerName> but was:<[]badLayerName>

I will see if I can fix them.

src/community/mapml/pom.xml
prushforth: > we keep version numbers in the root pom.xml in a dependency management section (if not we get madness)

I found that gs-web-demo is not in the dependency management section, but there is a <gs.version> in the <properties> section, so is it ok if I just use ${gs.version} instead of ${project.version} ? I didn't want to mess with the GeoServer pom.xml :-).

jodygarnett: because gs-web-demo is being built as part of this project it does not need to be defined as part of the dependency management section.

All the other sections use  ${project.version}:

```xml
   <dependency>
      <groupId>org.geoserver</groupId>
      <artifactId>gs-main</artifactId>
    </dependency>
```

No need to define a custom property since maven already provides one for the current build.

jodygarnett:
> I will see if I can fix them.

You are welcome to join us on gitter chat as we try and fix this and other failures.

prushforth:
> > I will see if I can fix them.
> 
> You are welcome to join us on gitter chat as we try and fix this and other failures.

Thanks I will try to tune  in- gitter sometimes doesn't work from my office.

src/community/mapml/src/main/java/org/geoserver/mapml/MapMLController.java
jodygarnett: To answer your implied question you can use JTS transform method to transform bounds into the same reference system before using expand to include. 

If you want to get extra fancy you can ask it to densify the edges, transform, and then make a new bounds (for bounding boxes near the poles with a curve to them).

See [JTS.transform](https://docs.geotools.org/latest/userguide/library/main/jts.html#transform) in the docs.

PR #3545, opened 2019-06-03
Fix GWC RESTIntegration test to work after recent Spring MVC fix

tbarsballe:
After some recent changes on master (I'm not sure exactly which one, but I'm guessing the depreciation removal PR), the GWC RESTIntegrationTest is failing.

Looking into the failures, it appears they are only affecting tests that were modified to work around some changes introduced by the Spring MVC REST update a couple years ago (see: https://github.com/geoserver/geoserver/commit/a354c9dd5177c0a739efcfc40bece4ece5b7afd2 ).

After restoring the tests to their original state (or something close to it) everything seems to be passing again.

### DIFF
--- a/src/gwc/src/test/java/org/geoserver/gwc/RESTIntegrationTest.java
+++ b/src/gwc/src/test/java/org/geoserver/gwc/RESTIntegrationTest.java
@@ -213,7 +213,7 @@ public class RESTIntegrationTest extends GeoServerSystemTestSupport {
         assertEquals(
                 expected,
                 response.getContentAsString()
-                        .substring(response.getContentAsString().indexOf(":") + 2));
+                        .substring(response.getContentAsString().indexOf(":") + 1));
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


jodygarnett:
So the test changes appear okay, but the result is not passing on Travis for AppVeyor

tbarsballe:
Closed & reopened to give the CI instances a fresh build -  Travis is happy with it this time around, appveyor is still complaining, although looking at the log it seems to be just stopping partway though the build with no error - I wonder if it is running into a time limit?

jodygarnett:
So appveyor is still failing, seems to the same failure addressed by #3552 

```
[00:10:04] testCopy(org.geoserver.rest.resources.ResourceControllerTest)  Time elapsed: 32 sec  <<< FAILURE!
[00:10:04] org.junit.ComparisonFailure: expected:<This is my test. [� �]> but was:<This is my test. [? ?]>
[00:10:04] 	at org.junit.Assert.assertEquals(Assert.java:115)
[00:10:04] 	at org.junit.Assert.assertEquals(Assert.java:144)
[00:10:04] 	at org.geoserver.rest.resources.ResourceControllerTest.testCopy(ResourceControllerTest.java:443)
```

PR #3557, opened 2019-06-04
GEOS-9235: Customizing OWS Exception Handler content-type

dromagnoli:


src/ows/src/test/java/org/geoserver/ows/OWS10ServiceExceptionHandlerTest.java
```diff
@@ -187,4 +218,17 @@ public void testHandleServiceExceptionNullMessages() throws Exception {
         // used to contain an extra " null" at the end
         assertEquals("hello service exception NullPointerException", exceptionText);
     }
+
+    @Test
+    public void exceptionType() throws Exception {
```
mbarto: Do you think a similar test is needed also for OWS 1.1?

dromagnoli: @mbarto Added a test for that too.

PR #3578, opened 2019-06-14
GEOS-9169 ResourcePool datastore cache cluster-safe

NielsCharlier:
This one is not really possible to test because you need a clustered set-up with hazelcast.

However, it is a minimal and trivial change, that does not break any tests and can be easily understood to improve cluster-safety.


aaime:
I'm guessing this is useful in combination with the hazelcast clustering, but I have no experience with it. Does the module propagate catalog change events to other modules? If so the change looks legit.

NielsCharlier:
@aaime that is exactly how it works

NielsCharlier:
Is it okay to merge this?

aaime:
Merged and added labels for automatic backport.

PR #3580, opened 2019-06-14
GEOS-9185 style image preview jdbcstore

NielsCharlier:
Style image preview breaks with jdbcstore. Using in memory stream rather than file reference to resolve.

### DIFF
--- a/src/web/wms/src/test/java/org/geoserver/wms/web/data/StyleEditPageTest.java
+++ b/src/web/wms/src/test/java/org/geoserver/wms/web/data/StyleEditPageTest.java
@@ -27,7 +27,6 @@ import org.apache.wicket.feedback.FeedbackMessage;
 import org.apache.wicket.markup.html.form.DropDownChoice;
 import org.apache.wicket.markup.html.form.TextArea;
 import org.apache.wicket.markup.html.form.TextField;
-import org.apache.wicket.request.resource.ResourceReference;
 import org.apache.wicket.util.tester.FormTester;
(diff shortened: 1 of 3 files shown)


src/web/wms/src/test/java/org/geoserver/wms/web/data/StyleEditPageTest.java
```diff
@@ -232,14 +231,6 @@ public void testInsertImage() throws Exception {
...
         tester.executeAjaxEvent("dialog:dialog:content:form:userPanel:image", "change");
         tester.assertVisible("dialog:dialog:content:form:userPanel:display");
-        assertTrue(
-                ((ResourceReference)
-                                ((org.apache.wicket.markup.html.image.Image)
-                                                tester.getComponentFromLastRenderedPage(
-                                                        "dialog:dialog:content:form:userPanel:display"))
-                                        .getDefaultModelObject())
-                        .getName()
-                        .equals("somepicture.png"));
```
tbarsballe: I assume this was removed because wicket is a pain and no longer exposes the image name when using the new stream-based implementation?
Have you verified manually that the image preview still shows up?

While I would prefer a similar check here in place of this one, I understand that it might not be feasible to get that working with wicket, so won't insist upon it.

NielsCharlier:
@tbarsballe  thanks for review. yeah the resource stream doesn't have any name really, so that cannot be tested any longer, however the test still checks the image display is visible and afterwards checks the file name in the inserted block of javacript

PR #3596, opened 2019-06-21
Change staging directory from release to distribution

jodygarnett:
This prevents gitignore from conflicting with src/release folder.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,8 +18,8 @@ target
 cucumber-report
 .mvn
 
-# Release process artifacts
-release/
+# staging release artifacts
+distribution/
 
 # NSG CITE testst data directory
 data/citensg-1.0/gwc/cite*
--- a/build/build_release.sh
+++ b/build/build_release.sh
@@ -97,7 +97,7 @@ fi
 # move to root of source tree
 pushd .. > /dev/null
 
-dist=`pwd`/release/$tag
+dist=`pwd`/distribution/$tag
 mkdir -p $dist/plugins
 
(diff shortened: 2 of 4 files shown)


bradh:
Ah. Missed docs update. Looks like there are two places the `release/<TAG>` gets mentioned. First one is at https://github.com/geoserver/geoserver/blob/bb7a1a12ee3400901d8ade4b1c5946639a240d46/doc/en/developer/source/release-guide/index.rst#build-the-release and the second is a few paragraphs down.

jodygarnett:
> Ah. Missed docs update. Looks like there are two places the `release/<TAG>` gets mentioned. 

Remember this takes place on a worker, there needs to be a separate step (as with nightlies) to sync releases over to the build server. It is not strictly needed since we are uploading to source forge ....

bradh:
> I am not sure we have any plans to make that work, new build server collects the artifacts on the job page.
> 
> Remember this takes place on a worker, there needs to be a separate step (as with nightlies) to sync releases over to the build server. It is not strictly needed since we are uploading to source forge ....

Valid. Probably just needs docs deletion.

tbarsballe:
> I am not sure we have any plans to make that work, new build server collects the artifacts on the job page.

The original plan was to make that work eventually, but I still haven't been able to figure out how to get worker nodes to ssh to the master nodes (and that is a bit of an inversion of the expected secure process, so I'm not entirely sure we even want workers to be able to do that)

Either way, looks good to me. I'll verify everything actually works when I do the release next week.

PR #3610, opened 2019-06-24
WIP: add a crumbs path to wfs3 html representation

pvgenuchten:
This PR is a work in progress, not to be merged. It introduces a crumbspath in the html representation of wfs3 (ogc api features) responses. And adds some best practices used by search engines, such as a title header and a canonical url.

@aaime already responded on gitter to this investigation with:

> btw, the wfs3 module is abandonware, I'll keep it there for a while because there is code that can be useful, but the action moved to ogcapi. At the same time, ogcapi is still quite green and needs improvement/cleanup in the dispatcher before others can jump in an implement other services

Goal of the PR to show how minimal changes can improve the usability of the html representation. A first next task would be to move the developments to the OAPI branch and evaluate if building of the crumbspath makes sense as part of the templates, this can probably better be managed in the java code.

A consideration from this effort is that each change currently needs a clean install. To consider is to move the templates outside the compiled package (into the datafolder), so users can easily override it to their needs at runtime.

image

### DIFF
--- a/src/community/wfs3/src/main/resources/org/geoserver/wfs3/response/collection.ftl
+++ b/src/community/wfs3/src/main/resources/org/geoserver/wfs3/response/collection.ftl
@@ -1,3 +1,6 @@
+<#global pagetitle=model.name>
+<#global pagepath="/collections/"+model.name>
+<#global pagecrumbs="<a href='"+serviceLink("wfs3")+"'>Home</a><a href='"+serviceLink("wfs3")+"/collections'>Collections</a><b>"+model.name+"</b>">
 <#include "common-header.ftl">
        <h2>${model.name}</h2>
        <#assign collection=model>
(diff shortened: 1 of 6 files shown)


aaime:
@pvgenuchten if you want to work from the IDE just use Start.java to run GeoServer and force the IDE to build the files, you should be able to work without restarting and without "clean install"

About the templates, I've tried to setup things so that templates can be overridden from the data dir, but never actually tried it out (never had the need, constantly short on time). See if placing the templates in ``$GEOSERVER_DATA_DIR/workspaces`` allows you to override the default ones contained in the classpath.

aaime:
@pvgenuchten so what should we do with this PR? Are you interested in moving it over to the ogcapi module?

pvgenuchten:
If you like the approach, let me move iT to oapi module. I’ll check about using the ide

aaime:
I do like the approach, not in love with the tall green bar. IMHO would be nicer if the green line was like today, and the breadcrums just above it, same blue color as they are now. Just my 2c ;-)

aaime:
Closing due to lack of activity (but still like the idea, would love to see it ported over to the api module)

PR #3630, opened 2019-07-02
[GEOS-9276] WMS configuration UI, watermark file input allows to validate a remote HTTP URL

fernandor777:
## Description

In WMS configuration page (main or workspace) the watermark file input allows to validate a remote http url making a request on it to check if exists.  This PR makes Watermark input detects introduced text is not a file url and don't make any tcp/ip connection to reach the resource.

Issue:
https://osgeo-org.atlassian.net/browse/GEOS-9276

## Checklist

For all pull requests:

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/wicket/FileExistsValidator.java
+++ b/src/web/core/src/main/java/org/geoserver/web/wicket/FileExistsValidator.java
@@ -55,6 +55,7 @@ public class FileExistsValidator implements IValidator<String> {
         try {
             URI uri = new URI(uriSpec);
             if (uri.getScheme() != null && !"file".equals(uri.getScheme())) {
+                // if delegate is null, remote URL is allowed
                 if (delegate != null) {
                     delegate.validate(validatable);
                     InputStream is = null;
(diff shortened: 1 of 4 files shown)


aaime:
Euh, part of the fix is right, not allow remote if a remote was not meant to be used, but allowing for a remote URL in watermark was actually intended functionality. We should provide a whitelist of sorts instead. Going to discuss it a bit.

aaime:
Damn, merged the wrong PR. Will revert, @fernandor777 let's talk when you are back online.

fernandor777:
Currently watermark URL converts (and adds) "file:/" prefix when this prefix is not found in the inserted string value on UI, that's why I guessed it's local file path only.  So since remote URLs are allowed this should be an unknown bug.

So I'll need to fix the 'file:/' only prefix forced bug and limit the remote URLs to a whitelist.  About this whitelist, should we use any existing list in geoserver?  If not, should we create a central domains/pattern whitelist configuration or a dedicated one (declaring it's only for watermark fetching)?.

Whitelist should have UI configuration or another simpler approach(property file, ...)?

PR #3644, opened 2019-07-05
[GEOS-9228] Renaming a layer doesn't update Data Security rules

imranrajjad:
https://osgeo-org.atlassian.net/browse/GEOS-9228

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/SecuredResourceNameChangeListener.java
+++ b/src/main/src/main/java/org/geoserver/security/SecuredResourceNameChangeListener.java
@@ -1,0 +1,170 @@
+/* (c) 2019 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.security;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import org.geoserver.catalog.Catalog;
+import org.geoserver.catalog.CatalogException;
+import org.geoserver.catalog.CatalogInfo;
+import org.geoserver.catalog.ResourceInfo;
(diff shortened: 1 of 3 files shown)


aaime:
@tbarsballe are you interested in reviewing this one? It's for that ticket you opened about security rules not being updated on rename.

src/main/src/main/java/org/geoserver/security/SecuredResourceNameChangeListener.java
```diff
@@ -0,0 +1,170 @@
...
+        } else if (event.getSource() instanceof ResourceInfo) {
+            // if a layer has been renamed
+            String oldResourceName =
+                    String.valueOf(
+                            event.getOldValues().get(event.getPropertyNames().indexOf("name")));
+            String newResourceName =
+                    String.valueOf(
+                            event.getNewValues().get(event.getPropertyNames().indexOf("name")));
+            // dont do anything if name has not changed
+            if (oldResourceName.equalsIgnoreCase(newResourceName)) return;
```
tbarsballe: The logic up to here is the same as that for workspace - you could consolidate and move this part outside of the if statements (ultimately not important, and up to personal preference, but that approach would be a bit more concise)

src/main/src/main/java/org/geoserver/security/SecuredResourceNameChangeListener.java
```diff
@@ -0,0 +1,170 @@
...
+                List<DataAccessRule> rulesToModifyList =
+                        getDataAccessRule(resourceInfo, oldResourceName);
+                for (DataAccessRule rule : rulesToModifyList) rule.setLayer(newResourceName);
+
+                try {
+                    dao.storeRules();
+                } catch (IOException e) {
+                    LOGGER.log(Level.SEVERE, e.getMessage(), e);
+                }
+            }
```
tbarsballe: The issue also affects layer groups (both global and local), and they aren't handled here - please add them.

PR #3650, opened 2019-07-08
[GEOS-9280] Uploading of SLD from GeoNode fails

robward-scisys:
## Description

Fixes a problem where the wrong REST endpoint is called when POSTing a style. Needed for backwards capability with GeoNode 2.10rc5

## Checklist

For all pull requests:


sikeoka:
This regression was caused by fixing a reflected cross-site scripting vulnerability that affected practically every REST endpoint: https://github.com/geoserver/geoserver/pull/3492

Doesn't this PR re-introduce the vulnerability?

robward-scisys:
Steve,

I’m not sure, this area of code has not been touched for a long time.
Looking at the code before I made the change I’m not sure how the correct method was ever called.  Something recently, within the last couple of weeks, has tightened up the checking and subsequently existing functionality no longer works.

Can you point me at any unit tests for this cross-site scripting?

________________________________________________________________

Robert Ward
Technical Architect – Enterprise Solutions & Defence Division
SCISYS UK Limited
T: +44 (0)1249 466575
E: robert.ward@scisys.co.uk<mailto:robert.ward@scisys.co.uk>
www.scisys.co.uk<https://eur01.safelinks.protection.outlook.com/?url=http%3A%2F%2Fwww.scisys.co.uk%2F&data=02%7C01%7CCarol.Harris%40gov.wales%7Ca4cfe203deb3406f6e7e08d6e834eee1%7Ca2cc36c592804ae78887d06dab89216b%7C0%7C0%7C636951711382394479&sdata=g7DlluMtbC28PIHfu8rmCil9mbER9zz%2F%2BYAuGGLpKu8%3D&reserved=0>

sikeoka:
GeoServer doesn't seem to have any restrictions on valid style names so I think that a safer fix would be to remove the produces value and @ResponseStatus annotation from that stylePost method and have it return a ResponseEntity instead of a String like the stylePost method below it does.  The produces value was added in the PR that I linked to: 
https://github.com/geoserver/geoserver/pull/3492/files#diff-3a9a3be06eeccf4ada24672ffeaab4d7L109

afabiani:
@sikeoka @robward-scisys the problem Robert is trying to fix is basically a wrong endpoint hit by the REST Controller when trying to create a new style.
The regression is easily reproducible through a curl request similar to the following one:

`curl -X POST -H 'Content-type: application/xml' -H 'Authorization: Basic YWRtaW46Z2Vvc2VydmVy' -H 'Accept: application/xml' -d '<style><name>ne_110m_land</name><filename>ne_110m_land.sld</filename></style>' "http://localhost:8080/geoserver/rest/workspaces/geonode/styles?name=ne_110m_land"`

Long story short the call above instead of being served correctly by this method

https://github.com/geoserver/geoserver/blob/07397c353c00b41046d53fcbe21e44fb454754ef/src/restconfig/src/main/java/org/geoserver/rest/catalog/StyleController.java#L113

is now served by this one instead

https://github.com/geoserver/geoserver/blob/07397c353c00b41046d53fcbe21e44fb454754ef/src/restconfig/src/main/java/org/geoserver/rest/catalog/StyleController.java#L161

I'm not fully sure the approach followed by Robert is the correct one. Probably the best way would be to somehow establish the order of the method mappings. The weird thing is why this behavior changed in GeoServer. The definition of the post mappings seems to be correct. Nevertheless, it does not work anymore. For sure this is a regression since now it is not possible anymore to create new styles by using the rest endpoints.

I'll try to spend some time too in order to better understand which could be the correct approach.

Any hint on which change could be the cause of this new error?

robward-scisys:
For the record the python package that is calling the REST end point that the GeoNode project uses is gsconfig.

The style is created in the `create_style()` [here](https://github.com/boundlessgeo/gsconfig/blob/master/src/geoserver/catalog.py)

The code that is failing is in lines 1063-1072

```
            headers = {
                "Content-type": "application/xml",
                "Accept": "application/xml"
            }
            xml = "<style><name>{0}</name><filename>{0}.sld</filename></style>".format(name)
            style = Style(self, name, workspace, style_format)

            resp = self.http_request(style.create_href, method='post', data=xml, headers=headers)
            if resp.status_code not in (200, 201, 202):
                FailedRequestError('Failed to create style {} : {}, {}'.format(name, resp.status_code, resp.text))
```

afabiani:
> The produces value was added in the PR that I linked to:
https://github.com/geoserver/geoserver/pull/3492/files#diff-3a9a3be06eeccf4ada24672ffeaab4d7L109

This is probably the cause. Not sure why the "produces" value was not added to the other too

https://github.com/geoserver/geoserver/blob/07397c353c00b41046d53fcbe21e44fb454754ef/src/restconfig/src/main/java/org/geoserver/rest/catalog/StyleController.java#L204

aaime:
The GeoNode API seems to be using a mistake in the previous implementations that has now been corrected. The code in question is not generating XML at all, it's returning a simple plain text string, thus the content type in "produces" is the correct one, "application/xml" makes no sense for it?

afabiani:
@aaime are not GeoNode API, those are the "gsconfig" ones.

I'll make a try by patchnig them and see what happens with

`"Accept": "tesxt/plain"`

instead of 

`"Accept": "application/xml"`

however, I still guess the "produces" value should be set on the second postMapping too.

sikeoka:
The produces value was added to cause GeoServer to set the text/plain Content-Type on the response but it also had an unintended side effect of causing the method to only handle requests that do not have the Accept header or have text/plain in the Accept header.  Other REST API methods that had a produces value added to it likely have the same regression.  My previous comment provides a way to explicitly set the text/plain Content-Type response header but without any restrictions on the Accept request header.

aaime:
@sikeoka yep, I did not realize that would have been the case before. On one side, this is the correct behavior, the method cannot produce anything else but "text/plain", having a client ask for xml and get back text is a mistake (the correct answer is "I cannot produce xml, sorry"). 
On the other side, it's an unintended change in behavior, so probably best to setup this method like the others and preserve backwards compatibility.

afabiani:
@robward-scisys @sikeoka @aaime I confirm that using the correct "Accept" header fixes the issue.

Try as an instance the following one (tested with GeoServer 2.15.2):

```curl -X POST -H 'Content-type: application/xml' -H 'Accept: text/plain' -H 'Authorization: Basic YWRtaW46Z2Vvc2VydmVy' -H 'Accept: application/xml' -d '<style><name>ne_110m_land</name><filename>ne_110m_land.sld</filename></style>' "http://localhost:8080/geoserver/rest/workspaces/geonode/styles?name=ne_110m_land"```

afabiani:
We confirmed that this is not an issue with GeoServer. We can close this PR.

Currently, the issue was fixed on GeoNode by updating the rest api client library.

aaime:
@afabiani following your lead and closing.

PR #3660, opened 2019-07-15
[GEOS-9073] Encoding issues with the importer plugin

imranrajjad:
Fixes as per https://osgeo-org.atlassian.net/browse/GEOS-9073
As per new changes data json object can be passed data charset encoding

```
{
   "import": {
      "targetWorkspace": {
         "workspace": {
            "name": "topp"
         }
      },
      "data": {
        "type": "file",
        "file": "<path to shape>",
	**"charsetEncoding":"UTF-8"**
      }
   }
}
```


simboss:
side note, we need to backport this fix up to 2.14 as it impacts GeoNode as well.

mostafafallah:
as I see the problem not fixed yet completely. 
first step json :
with Post Method :
{
  "import": {
    "targetStore": {
      "dataStore": {
        "name": "postGIS"
      }
    },
    "targetWorkspace": {
      "workspace": {
        "name": "webgis"
      }
    }
  }
}

Second Step :
I am posting the returned ImportId of step 1 into this url and sending to sever as post
http://localhost:8070/geoserver/rest/imports/{importId}/tasks
with this request I am sending shape file too

and then do next steps 

everything is ok but the result is not unicode

Where I have to do set charset exactly ?

PR #3671, opened 2019-07-22
Update build scripts with SKIP COMMUNITY option

jodygarnett:
Based on email discussion.

### DIFF
--- a/build/build_release.sh
+++ b/build/build_release.sh
@@ -20,6 +20,7 @@ function usage() {
   echo
   echo "Environment variables:"
   echo " SKIP_BUILD : Skips main release build"
+  echo " SKIP_COMMUNITY : Skips community release build"
   echo " SKIP_TAG : Skips tag on release branch"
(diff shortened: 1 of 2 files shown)


build/build_release.sh
```diff
@@ -279,9 +280,21 @@ if [ -z $SKIP_BUILD ]; then
...
+   set +e
+   mvn clean install -P communityRelease -DskipTests $MAVEN_FLAGS || true
+   set -e
+   popd > /dev/null
+else
+   echo "Skipping mvn clean install -P communityRelease -DskipTests"
 fi
 
+
 mvn assembly:attached $MAVEN_FLAGS
```
jodygarnett: @tbarsballe this line was hanging out on its own, unclear if it wants to use communityRelease profile

tbarsballe: That looks like its supposed to be top-level, so I expect not, since `mvn assembly:attached -PcommunityRelease` doesn't actually seem to do anything different than `mvn assembly:attached` if you're not in the community dir

PR #3672, opened 2019-07-23
Show empty option in dropdowns (null workspace in style)

NielsCharlier:
see mailing list discussion


src/web/core/src/main/java/org/geoserver/web/css/geoserver.css
```diff
@@ -1341,6 +1341,8 @@ li.select2-results__option {
     /*Altered two below to make word wrap work */
     word-wrap: break-word !important;
     white-space: normal;
+    /* show empty option properly */
+    min-height:18px;
```
jodygarnett: nice fix!

PR #3674, opened 2019-07-25
expand axis order discussion with examples

jodygarnett:
This is currently an addition to the docs expanding the WFS axis order discussion with examples, highlighting combinations that produce unexpected results.


jodygarnett:
The above failures:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.22.2:test (default-test) on project gs-wms: There are test failures.
```

Should be unrelated to this documentation change.

doc/en/user/source/services/wfs/axis_order.rst
```diff
@@ -5,38 +5,56 @@ Axis ordering
 
 WFS 1.0.0 servers return geographic coordinates in longitude/latitude (x/y) order, the most common way to distribute data. For example, many shapefiles adopt this order by default. 
 
-However, the traditional axis order for geographic and cartographic systems is the opposite—latitude/longitude (y/x)—and the later WFS specifications respect this. The default axis ordering support is:
+However, the traditional axis order for geographic and cartographic systems is the opposite—latitude/longitude (y/x) — and the later WFS specifications respect this. The default axis ordering support is:
```
bradh: I think something like:

However, the traditional axis order for geographic and cartographic systems is the opposite: latitude/longitude (y/x),  and the later WFS specifications respect this

jodygarnett: Thanks I will resolve it to have a similar structure as the proceeding sentence:

> However, the traditional axis order for geographic and cartographic systems in latitude/longitude (y/x) order, and the later WFS specifications respect this.

doc/en/user/source/services/wfs/axis_order.rst
```diff
@@ -49,7 +67,33 @@ The default data directory includes the following dataset illustrating this chal
...
+  
+   The :file:`PRJ` did not provide an axis-order and GeoSever has filled in an assumption. This describing the data in `x/y` order which matches our data and we could use it unmodified.
+   
+   .. figure:: img/native_srs.png
+      
+      Native SRS for topp:states
+
+#. Locate :guilabel:`Declared SRS` and click on :guilabel:`EPSG:WGS 84...` link to see the definition used to publish this content.
    
+   This is the internal definition of ``EPSG:4326`` as shown in the SRSList above, which also describes the data in `x/y` order matching our data. This definition provides slightly more readable names along with additional  ``AUTHORY`` information that may be helpful to client applications.
```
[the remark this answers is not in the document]
jodygarnett: fixed

jodygarnett:
Updated, thanks for the feedback @bradh, rebased in the hopes of getting a passing build.

jodygarnett:
Build still fails on both AppVeyor and Travis on the same three tests...

doc/en/user/source/services/wfs/axis_order.rst
```diff
@@ -0,0 +1,676 @@
+.. _wfs_basics_axis:
+
+Axis ordering
+=============
+
+WFS 1.0.0 servers return geographic coordinates in longitude/latitude (x/y) order, the most common way to distribute data. For example, many shapefiles adopt this order by default.
```
aaime: Simply talking about longitude and latitude is limiting and gives false impressions. There is a number of projected CRSs that are north/east oriented in their EPSG definition, those are affected by the axis order rules too. Would be better to talk about "respecting EPSG axis order" vs "forcing east/north" order. It is also to be noted that there are CRSs, like the polar stereographic based ones, where talking about east and north makes no sense, having a pole in the middle of the CRS axis.

jodygarnett:
Andrea fixed build upstream at geotools level (failed build resulted in conflicting copies of jai-ext being pulled in).

PR #3684, opened 2019-07-26
mv manual to new name

ianturton:
Try to force manual document to be called geoserver-2.x.t-user-manual.pdf

### DIFF
--- a/build/build_release.sh
+++ b/build/build_release.sh
@@ -364,13 +364,13 @@ unlink readme
 
 popd > /dev/null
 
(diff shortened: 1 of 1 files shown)


tbarsballe:
Just as an FYI, the build_release.sh script only ever runs off of the master branch - when you are releasing 2.14, it still uses the master version of all the release scripts.
I (should) have fixed the name issue on master with this commit: https://github.com/geoserver/geoserver/commit/858bbef7fff9fd3a973a30036ac69ef5f8367e83

PR #3689, opened 2019-07-31
[GEOS-9296] Fix race condition in audit logger initialization

sikeoka:
This pull request synchronizes the audit logger initialization to fix a race condition that could cause the audit log files to be written to the wrong directory. https://osgeo-org.atlassian.net/browse/GEOS-9296

## Checklist

For all pull requests:

- [X] Commits changing the UI, existing user workflows, or adding new functionality, need to include documentation updates

### DIFF
--- a/src/extension/monitor/core/src/main/java/org/geoserver/monitor/auditlog/AuditLogger.java
+++ b/src/extension/monitor/core/src/main/java/org/geoserver/monitor/auditlog/AuditLogger.java
@@ -62,7 +62,7 @@ public class AuditLogger implements RequestDataListener, ApplicationListener<App
 
     MonitorConfig config;
 
-    RequestDumper dumper;
+    volatile RequestDumper dumper;
 
     int rollLimit;
 
@@ -80,8 +80,8 @@ public class AuditLogger implements RequestDataListener, ApplicationListener<App
         templateConfig.setTemplateLoader(new AuditTemplateLoader(loader));
     }
 
-    void initDumper() throws IOException {
-        if (getProperty("enabled", Boolean.class, false)) {
+    synchronized void initDumper() throws IOException {
+        if (this.dumper == null && getProperty("enabled", Boolean.class, false)) {
             // prepare the config
             rollLimit = getProperty("roll_limit", Integer.class, DEFAULT_ROLLING_LIMIT);
             path = System.getProperty("GEOSERVER_AUDIT_PATH");
--- a/src/extension/monitor/core/src/test/java/org/geoserver/monitor/auditlog/AuditLoggerTest.java
+++ b/src/extension/monitor/core/src/test/java/org/geoserver/monitor/auditlog/AuditLoggerTest.java
@@ -1,0 +1,103 @@
+/* (c) 2019 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.monitor.auditlog;
+
+import static org.hamcrest.Matchers.arrayWithSize;
+import static org.hamcrest.Matchers.emptyArray;
+import static org.hamcrest.Matchers.greaterThan;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import org.geoserver.monitor.MonitorConfig;
+import org.geoserver.monitor.RequestData;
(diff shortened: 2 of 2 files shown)


sikeoka:
@aaime, any chance you could take a look at this PR please?  Thank you.

src/extension/monitor/core/src/main/java/org/geoserver/monitor/auditlog/AuditLogger.java
```diff
@@ -80,8 +80,8 @@ public AuditLogger(MonitorConfig config, GeoServerResourceLoader loader) throws
         templateConfig.setTemplateLoader(new AuditTemplateLoader(loader));
     }
 
-    void initDumper() throws IOException {
-        if (getProperty("enabled", Boolean.class, false)) {
+    synchronized void initDumper() throws IOException {
+        if (this.dumper == null && getProperty("enabled", Boolean.class, false)) {
```
aaime: Doh, this makes for a working double checked locking, but one that is less obvious to spot... maybe best to do the usual pattern in the caller method, "requestPostProcessed" instead?, .e.g:

```
   if (dumper == null) {
      synchronized(this) {
          if (dumper == null) initiDumper();
      }
```

src/extension/monitor/core/src/test/java/org/geoserver/monitor/auditlog/AuditLoggerTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    private File bad;
+    private File good;
+    private GeoServerResourceLoader loader;
+    private MonitorConfig config;
+    private GenericApplicationContext context;
+
+    @Before
+    public void setUp() throws Exception {
+        this.bad = this.directory.newFolder("bad");
```
aaime: I don't understand this one... even if two threads end up performing the initialization, won't they both stop when they [find the system variable](https://github.com/geoserver/geoserver/blob/a38bc435cb9889f60a84fb67c815f5ad6c218035/src/extension/monitor/core/src/main/java/org/geoserver/monitor/auditlog/AuditLogger.java#L86) and end up ignoring the "bad" config value? The config field is used only if the path retrieved from the system property is null or empty, I don't see code resetting it to null or empty.

I would expect the race condition to have two dumpers created against the same directory and end up overwriting the first few request worth of logs, likely causing the log files to be invalid.

src/extension/monitor/core/src/main/java/org/geoserver/monitor/auditlog/AuditLogger.java
sikeoka: There are two places in requestPostProcessed where initDumper is called.  It seems a little messy to have to do that synchronized code twice.

src/extension/monitor/core/src/test/java/org/geoserver/monitor/auditlog/AuditLoggerTest.java
aaime: No feedback on this comment?

sikeoka: There are three possible bad effects of this race condition that I've seen.

1. What you mentioned where multiple RequestDumper threads will be created that try to write to the same log file.  Only one thread will succeed so a few requests will not be logged and there may be a few extra RequestDumper threads that will never exit.  This is not dependent on the system property and is self-correcting other than a minor leak.
2. Multiple RequestDumper threads will be created that try to write to different log files.  A few audit log files may be created with only a single request in them and there may be a few extra RequestDumper threads that will never exit.  This is not dependent on the system property and is self-correcting other than a minor leak.
3. The most severe bug is dependent on the system property and will not automatically correct itself.  The only viable option to fix this with a clustered GeoServer environment using a shared data directory is to restart any bugged GeoServer nodes until they are in a good state.  initDumper() sets the AuditLogger path to the system property, creates the resource from the path, and then sets the path to the property from the config file.  When multiple threads call initDumper() at the same time, it is possible for one thread to set the path to the config file value between the other thread setting it to the system property and creating the resource.

src/extension/monitor/core/src/main/java/org/geoserver/monitor/auditlog/AuditLogger.java
sikeoka: Do you have any additional thoughts about this @aaime?

src/extension/monitor/core/src/test/java/org/geoserver/monitor/auditlog/AuditLoggerTest.java
sikeoka: My unit test checks for all of these possible error conditions.

aaime:
I still don't understand the race condition, but I guess I won't get enough time and peace to look at it, I have no working hours  to look at it, and during the weekend the house is noisy. Going to trust you on this one and merging.

PR #3673, opened 2019-07-23
Mapfish Print - Error when loading fonts, with configDir in windows machines

tixvieira:
Upgrade of mapfish-print lib version to support font configuration on windows machines.

Reference:
https://github.com/mapfish/mapfish-print/pull/998

Already release on central maven:
http://central.maven.org/maven2/org/mapfish/print/print-lib/2.1.5/

Are you ok with this change?
Can you merge it please?

Thanks

PR #3691, opened 2019-08-01
Mapfish Print - Error when loading fonts, with configDir in windows machines

tixvieira:
Merge from: https://github.com/geoserver/geoserver/pull/3673

Upgrade of mapfish-print lib version to support font configuration on windows machines.

Reference:
mapfish/mapfish-print#998

Already release on central maven:
http://central.maven.org/maven2/org/mapfish/print/print-lib/2.1.5/


mbarto:
Looks good to me, merging. Thanks!

PR #3695, opened 2019-08-06
GEOS-9293 Style rest PUT issues

NielsCharlier:

### DIFF
--- a/src/main/src/main/java/org/geoserver/config/GeoServerResourcePersister.java
+++ b/src/main/src/main/java/org/geoserver/config/GeoServerResourcePersister.java
@@ -12,19 +12,17 @@ import java.net.URISyntaxException;
 import java.util.Arrays;
 import java.util.List;
 import java.util.logging.Logger;
-import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.io.IOUtils;
+import org.geoserver.catalog.Catalog;
 import org.geoserver.catalog.CatalogException;
-import org.geoserver.catalog.SLDHandler;
-import org.geoserver.catalog.StyleHandler;
 import org.geoserver.catalog.StyleInfo;
-import org.geoserver.catalog.Styles;
 import org.geoserver.catalog.WorkspaceInfo;
 import org.geoserver.catalog.event.CatalogAddEvent;
 import org.geoserver.catalog.event.CatalogListener;
 import org.geoserver.catalog.event.CatalogModifyEvent;
 import org.geoserver.catalog.event.CatalogPostModifyEvent;
 import org.geoserver.catalog.event.CatalogRemoveEvent;
+import org.geoserver.catalog.impl.ResolvingProxy;
 import org.geoserver.platform.GeoServerResourceLoader;
 import org.geoserver.platform.resource.Resource;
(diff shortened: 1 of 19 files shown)


src/main/src/main/java/org/geoserver/config/GeoServerResourcePersister.java
```diff
@@ -143,40 +108,6 @@ public void handleRemoveEvent(CatalogRemoveEvent event) {
...
-     * Determine unique name of the form <code>newName.extension</code>. newName will have a number
-     * appended as required to produce a unique resource name.
-     *
-     * @param resource Resource being renamed
-     * @param newName proposed name to use as a template
-     * @param extension extension
-     * @return New UNDEFINED resource suitable for use with rename
-     * @throws IOException If unique resource cannot be produced
-     */
-    private Resource uniqueResource(Resource resource, String newName, String extension)
```
tbarsballe: I don't quite agree with this removal - even if it isn't currently used anywhere other then style, it may be better to keep it in ResourcePersister and make it public rather than move the logic to CatalogImpl - if there is the need to rename some other type of resource in the future, this has a better chance of keeping all the logic in one place.

src/restconfig/src/main/java/org/geoserver/rest/catalog/StyleController.java
```diff
@@ -470,16 +470,6 @@ public void stylePut(
...
-        // ensure no workspace change
-        if (info.getWorkspace() != null) {
-            if (!info.getWorkspace().equals(original.getWorkspace())) {
-                throw new RestException(
-                        "Can't change the workspace of a style, instead "
-                                + "DELETE from existing workspace and POST to new workspace",
-                        HttpStatus.FORBIDDEN);
-            }
-        }
-
```
tbarsballe: Looks good.

NielsCharlier:
@tbarsballe thanks for review. The custom block of code in StyleEditPage you refer to, does not handle renames. It just writes out the raw style code from the form into the file. (Purposely the old location, see: https://github.com/geoserver/geoserver/blob/master/src/web/wms/src/main/java/org/geoserver/wms/web/data/StyleEditPage.java#L116)
The ResourcePersister was handling renames/workspace change in all cases, the extensive unit test relied on it as well. So it should be sufficient that this test still passes. I don't think I can remove any code in that block that writes out the raw style.

src/main/src/main/java/org/geoserver/config/GeoServerResourcePersister.java
NielsCharlier: ResourcePersister has a specific function of handling catalog events, it is not a class to expose  utility methods to other classes. I would suggest moving it to the "Resources" class then.

NielsCharlier: I have added a commit to do so

tbarsballe: Great, that works for me

tbarsballe:
Okay, in that case, your changes seem sufficient. Looks good to me.

NielsCharlier:
Fantastic! thanks again

PR #3696, opened 2019-08-06
GEOS-9290 jdbcconfig: update catalog info references

NielsCharlier:

### DIFF
--- a/src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
+++ b/src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
@@ -41,6 +41,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
@@ -58,15 +59,25 @@ import javax.sql.DataSource;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.collections.Predicate;
 import org.apache.wicket.util.string.Strings;
+import org.geoserver.catalog.Catalog;
 import org.geoserver.catalog.CatalogInfo;
+import org.geoserver.catalog.CatalogVisitor;
+import org.geoserver.catalog.CoverageInfo;
+import org.geoserver.catalog.CoverageStoreInfo;
+import org.geoserver.catalog.DataStoreInfo;
+import org.geoserver.catalog.FeatureTypeInfo;
 import org.geoserver.catalog.Info;
 import org.geoserver.catalog.LayerGroupInfo;
 import org.geoserver.catalog.LayerInfo;
 import org.geoserver.catalog.MetadataMap;
+import org.geoserver.catalog.NamespaceInfo;
 import org.geoserver.catalog.Predicates;
 import org.geoserver.catalog.PublishedInfo;
 import org.geoserver.catalog.ResourceInfo;
+import org.geoserver.catalog.StoreInfo;
 import org.geoserver.catalog.StyleInfo;
+import org.geoserver.catalog.WMSLayerInfo;
+import org.geoserver.catalog.WMSStoreInfo;
 import org.geoserver.catalog.WorkspaceInfo;
 import org.geoserver.catalog.event.CatalogAddEvent;
 import org.geoserver.catalog.event.CatalogListener;
@@ -1131,6 +1142,9 @@ public class ConfigDatabase {
         // objects)
         resolveTransient(real);
 
+        // if this came from the cache, force update references
+        real.accept(new CatalogReferenceUpdater());
+
         return real;
     }
 
@@ -1468,7 +1482,7 @@ public class ConfigDatabase {
         return !propertyTypes.isEmpty();
     }
 
-    void clearCache(Info info) {
+    public void clearCache(Info info) {
         if (info instanceof ServiceInfo) {
             // need to figure out how to remove only the relevant cache
             // entries for the service info, like with InfoIdenties below,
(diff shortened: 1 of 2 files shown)


src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
```diff
@@ -1678,4 +1692,110 @@ public void handleSettingsAdded(SettingsInfo settings) {
...
+        public void visit(CoverageInfo coverage) {
+            if (coverage.getNamespace() != null) {
+                coverage.setNamespace(
+                        getById(coverage.getNamespace().getId(), NamespaceInfo.class));
+            }
+            coverage.setStore(getById(coverage.getStore().getId(), StoreInfo.class));
+        }
+
+        @Override
+        public void visit(LayerInfo layer) {
```
sikeoka: Does the Resource also need to be updated?

src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
```diff
@@ -1678,4 +1692,110 @@ public void handleSettingsAdded(SettingsInfo settings) {
...
+
+        @Override
+        public void visit(StyleInfo style) {
+            if (style.getWorkspace() != null) {
+                style.setWorkspace(getById(style.getWorkspace().getId(), WorkspaceInfo.class));
+            }
+        }
+
+        @Override
+        public void visit(LayerGroupInfo layerGroup) {
```
sikeoka: Do the RootLayer, RootLayerStyle and Styles also need to be updated?

src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
```diff
@@ -1678,4 +1692,110 @@ public void handleSettingsAdded(SettingsInfo settings) {
...
+
+        @Override
+        public void visit(WMSLayerInfo wmsLayer) {
+            if (wmsLayer.getNamespace() != null) {
+                wmsLayer.setNamespace(
+                        getById(wmsLayer.getNamespace().getId(), NamespaceInfo.class));
+            }
+            wmsLayer.setStore(getById(wmsLayer.getStore().getId(), StoreInfo.class));
+        }
+    }
```
sikeoka: It probably wouldn't hurt to override the default implementations of visit(WMTSStoreInfo store) and visit(WMTSLayerInfo layer).

src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
NielsCharlier: no, there is a separate mechanism for that, the layer gets invalidated if the resource changes

src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
NielsCharlier: indeed

src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
NielsCharlier: why, they don't seem to have any references?

src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
NielsCharlier: commit added for this

src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
sikeoka: Those two methods were added using Java 8 default methods.

src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
sikeoka: LayerGroupInfo styles are a List (not a Set like LayerInfo) and the order of the styles does matter.

src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
NielsCharlier: I understand, but I see no reason to override them?

src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
NielsCharlier: Yes, indeed. Fixed this.

src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
sikeoka: Why would you need to do something for WMSStoreInfo and WMSLayerInfo but not WMTSStoreInfo and WMTSLayerInfo?

NielsCharlier: sorry, you are right, it does ahve references! I changed it

sikeoka:
Any chance you could make the PR a single commit instead of four commits?  Other than that, it looks good to me.

NielsCharlier:
Thanks @sikeoka. you can squash merge with github :) Just fixed a style error and waiting for new build

PR #3710, opened 2019-08-15
GEOS-8948 CSW-ISO: support Feature Catalogues

NielsCharlier:
only the changes in csw-core and csw-api need review


src/community/csw-iso/src/main/java/org/geoserver/csw/records/iso/FeatureCatalogueDescriptor.java
```diff
@@ -0,0 +1,143 @@
...
+
+    public static FeatureType FEATURECATALOGUE_TYPE;
+    public static AttributeDescriptor FEATURECATALOGUE_DESCRIPTOR;
+
+    static {
+        FeatureTypeFactory typeFactory = new FeatureTypeFactoryImpl();
+
+        EmfComplexFeatureReader reader = EmfComplexFeatureReader.newInstance();
+
+        SchemaIndex index = null;
```
nmco: I would add a log here, usually querying external URLs are a source of issues.

src/community/csw-iso/src/main/java/org/geoserver/csw/records/iso/FeatureCatalogueDescriptor.java
```diff
@@ -0,0 +1,143 @@
...
+    // singleton
+
+    private FeatureCatalogueDescriptor() {}
+
+    private static FeatureCatalogueDescriptor INSTANCE;
+
+    public static FeatureCatalogueDescriptor getInstance() {
+        if (INSTANCE == null) {
+            INSTANCE = GeoServerExtensions.bean(FeatureCatalogueDescriptor.class);
+            if (INSTANCE == null) {
```
nmco: This is a race condition, multiple instances may end-up being initiated, not sure fi this is a problem or not.

src/community/csw-iso/src/main/java/org/geoserver/csw/records/iso/FeatureCatalogueDescriptor.java
```diff
@@ -0,0 +1,143 @@
...
+
+    // singleton
+
+    private FeatureCatalogueDescriptor() {}
+
+    private static FeatureCatalogueDescriptor INSTANCE;
+
+    public static FeatureCatalogueDescriptor getInstance() {
+        if (INSTANCE == null) {
+            INSTANCE = GeoServerExtensions.bean(FeatureCatalogueDescriptor.class);
```
nmco: This mix between spring and singletons instantiated manually may end in situations where the bean is not return because it doesn't exists but because it was not instantiated yet.

src/community/csw-iso/src/main/java/org/geoserver/csw/response/MetaDataTransformer.java
```diff
@@ -74,6 +76,20 @@ private void encodeProperty(Feature f, Property p) {
                         MetaDataDescriptor.NAMESPACES.getPrefix(p.getName().getNamespaceURI());
 
                 AttributesImpl atts = new AttributesImpl();
+                if (p.isNillable()) {
+                    Property prop =
+                            ((ComplexAttribute) p)
+                                    .getProperty(ComplexFeatureConstants.SIMPLE_CONTENT);
+                    boolean nil =
+                            prop == null || prop.getValue() == null || prop.getValue().equals("");
```
nmco: You could use here the same ``Strings.isNullOrEmpty`` you used bellow.

src/community/csw-iso/src/main/resources/net/opengis/schemas/iso/19139/20070417/gmx/extendedTypes.xsd
```diff
@@ -6,7 +6,7 @@
 	</xs:annotation>
 	<!-- ================================== Imports ================================== -->
 	<xs:import namespace="http://www.isotc211.org/2005/gco" schemaLocation="http://schemas.opengis.net/iso/19139/20070417/gco/gco.xsd"/>
-	<xs:import namespace="http://www.w3.org/1999/xlink" schemaLocation="http://www.w3.org/1999/xlink.xsd"/>
+	<!-- <xs:import namespace="http://www.w3.org/1999/xlink" schemaLocation="http://www.w3.org/1999/xlink.xsd"/>   -->
```
nmco: If it is not needed indeed I would remove.

src/community/csw-iso/src/main/resources/schemas/iso/19139/20070417/gmx/extendedTypes.xsd
```diff
@@ -6,7 +6,7 @@
 	</xs:annotation>
 	<!-- ================================== Imports ================================== -->
 	<xs:import namespace="http://www.isotc211.org/2005/gco" schemaLocation="http://schemas.opengis.net/iso/19139/20070417/gco/gco.xsd"/>
-	<xs:import namespace="http://www.w3.org/1999/xlink" schemaLocation="http://www.w3.org/1999/xlink.xsd"/>
+	<!--  <xs:import namespace="http://www.w3.org/1999/xlink" schemaLocation="http://www.w3.org/1999/xlink.xsd"/>  -->
```
nmco: Same as above.

src/community/csw-iso/src/test/java/org/geoserver/csw/store/internal/iso/FeatureCatalogueTest.java
```diff
@@ -0,0 +1,156 @@
+package org.geoserver.csw.store.internal.iso;
```
nmco: Copyright header is missing.

src/extension/csw/api/src/main/java/org/geoserver/csw/records/GenericRecordBuilder.java
```diff
@@ -55,33 +56,25 @@
...
 
-        @Override
-        public ComplexTreeLeaf clone() {
-            ComplexTreeLeaf leaf = new ComplexTreeLeaf();
-            leaf.value.putAll(value);
-            leaf.userData = userData;
-            leaf.descriptor = descriptor;
-            return leaf;
+        public boolean cleanUp() {
+            // TODO Auto-generated method stub
```
nmco: ?

src/community/csw-iso/src/main/java/org/geoserver/csw/records/iso/FeatureCatalogueDescriptor.java
```diff
@@ -0,0 +1,143 @@
...
+import org.opengis.feature.type.Name;
+import org.opengis.filter.Filter;
+import org.opengis.filter.expression.PropertyName;
+import org.xml.sax.helpers.NamespaceSupport;
+
+/**
+ * Describes the ISO FeatureCatalogue records and provides some handy constants to help building
+ * features representing FC_FeatureCatalogue.
+ *
+ * @author Niels Charlier
```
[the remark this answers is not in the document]
NielsCharlier: fair enough, force of habit

src/community/csw-iso/src/main/java/org/geoserver/csw/records/iso/FeatureCatalogueDescriptor.java
```diff
@@ -0,0 +1,143 @@
...
+
+    static {
+        FeatureTypeFactory typeFactory = new FeatureTypeFactoryImpl();
+
+        EmfComplexFeatureReader reader = EmfComplexFeatureReader.newInstance();
+
+        SchemaIndex index = null;
+        try {
+            index =
+                    reader.parse(
```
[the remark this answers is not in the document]
NielsCharlier: It actually reads it locally! Using resolvers

src/community/csw-iso/src/main/java/org/geoserver/csw/records/iso/FeatureCatalogueDescriptor.java
NielsCharlier: as said above, it reads it actually finds it locally using resolver

src/community/csw-iso/src/main/java/org/geoserver/csw/records/iso/FeatureCatalogueDescriptor.java
NielsCharlier: the descriptors all work like this, if there are two instances that is actually no problem because the class only provides information

src/community/csw-iso/src/main/java/org/geoserver/csw/response/MetaDataTransformer.java
NielsCharlier: no bcause getValue() is not necessarily a String... so we have to check if it is null (of any type) or equal to the empty string

src/extension/csw/api/src/main/java/org/geoserver/csw/records/GenericRecordBuilder.java
NielsCharlier: todo needs to be removed there indeed and better comment added

NielsCharlier:
thanks for the review @nmco

PR #3711, opened 2019-08-16
ModificationProxyCloner deep copy fix

NielsCharlier:
The ModificationProxyCloner deep copies only one level, but it should deep copy infinitely.

(This is relevant for the upcoming metadata module, which puts deeper level metadata in resources)

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/impl/ModificationProxyCloner.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/ModificationProxyCloner.java
@@ -63,8 +63,14 @@ class ModificationProxyCloner {
      * copying the catalog, and re-attaching to it, in there)
      *
      * @param source
+     * @throws IllegalAccessException
+     * @throws InstantiationException
+     * @throws InvocationTargetException
+     * @throws NoSuchMethodException
      */
-    static <T> T clone(T source) {
+    static <T> T clone(T source)
+            throws InstantiationException, IllegalAccessException, NoSuchMethodException,
+                    InvocationTargetException {
         // null?
         if (source == null) {
             return null;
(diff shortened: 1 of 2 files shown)


aaime:
"infinitely" is a bit scary, but the changes seem to only make it deep copy lists and map. I can't remember, was there self-loop prevention? (whatever data structure that eventually ends up pointing to itself).

NielsCharlier:
hmm, there is no self-loop prevention indeed. but this is only adding support for collection data-structures that contain other collection datastructures. Self-referencing datastructures like CRS won't be affected. If someone has a collection that contains itself, you would indeed get an infinite loop, but I don't think that is a thing, not even sure if the XML serialiser could handle that.

aaime:
XStream can handle it, I was using it in a project that had a complex object graph in 2004, with self loops, it's going to generate references to objects that have been serialized already, see: https://x-stream.github.io/graphs.html

NielsCharlier:
Apparently there are other parts of geoserver that cannot handle it though, because I tried to store a list with a self-reference in the metadata of a catalog resource and got:
17 Aug 15:43:38 WARN [catalog.impl] - Catalog listener threw exception handling event.
java.lang.StackOverflowError
	at java.util.AbstractCollection.toString(AbstractCollection.java:454)
	at java.lang.String.valueOf(String.java:2994)

NielsCharlier:
The toString of abstractcollection can't handle it ?

aaime:
Well we don't do tostring in normal operations (only in some debugging scenarios with high logging level), but we clone objects a lot. Make sure that cloning does not cause geoserver to fail.

aaime:
Wondering if there is some easy to implement safety that can be taken, like a recurse counter of some kind? Or maybe just add a try/catch around those and log the potential stack overflow exception with a meaningful message?

NielsCharlier:
I think catching a stack overflow is rather weird, then I would prefer counting the recursion level. 
However, in my opinion there isn't really any risk. The only way this change could cause an infinite loop if there is a loop using java collections only (if there is a class that references itself, is clone method should take this into account or it would already fail); but if you do have such a thing other things will break in any case.

aaime:
Fair enough, let's see down the road if/when it breaks then. Merge at will.

NielsCharlier:
thanks for the review @aaime

PR #3713, opened 2019-08-16
GEOS-9237 remove default ports from base url

NielsCharlier:


src/ows/src/main/java/org/geoserver/ows/util/ResponseUtils.java
```diff
@@ -410,12 +410,12 @@ public static String buildSchemaURL(String baseURL, String path) {
      */
     public static String baseURL(HttpServletRequest req) {
         StringBuffer sb = new StringBuffer(req.getScheme());
-        sb.append("://")
-                .append(req.getServerName())
-                .append(":")
-                .append(req.getServerPort())
-                .append(req.getContextPath())
-                .append("/");
+        sb.append("://").append(req.getServerName());
```
aaime: Looks good. Having a test covering the https case to would be nice.

NielsCharlier: Added a test and fixed app-schema tests

aaime: Thanks, the rest looks good, let's see what Travis/Appveyor have to say.

PR #3714, opened 2019-08-19
Fix the output format of dds plugin

jinghm318:
Fix the output format of dds plugin.


src/community/dds/src/main/java/org/geoserver/wms/worldwind/BilMapResponse.java
```diff
@@ -1,352 +1,372 @@
...
-    public MapProducerCapabilities getCapabilities(String outputFormat) {
-        // FIXME become more capable
-        return new MapProducerCapabilities(false, false, false, false, null);
-    }
-
-    static {
-        RecodeRaster.register(JAI.getDefaultInstance());
-    }
-}
+/* (c) 2014-2015 Open Source Geospatial Foundation - all rights reserved
```
aaime: This is a full reformat, it is not admissible (actual changes are hidden) and won't pass the Travis QA tests. Please build the module with Maven, it will automatically format it with the expected conventions.

jinghm318: > This is a full reformat, it is not admissible (actual changes are hidden) and won't pass the Travis QA tests. Please build the module with Maven, it will automatically format it with the expected conventions.

I have fixed the format.

bradh: This is still showing as every line changed.

jinghm318: > This is still showing as every line changed.

It's strange that I have built the extension with Maven.You see it passed the Travis test.I just add some lines.
I'm not familiar with java and I will try to rebuild the entire project instead of building the dds extension.

bradh:
Is this the change you're really trying to make:

```
--- a/src/community/dds/src/main/java/org/geoserver/wms/worldwind/BilMapResponse.java
+++ b/src/community/dds/src/main/java/org/geoserver/wms/worldwind/BilMapResponse.java
@@ -29,6 +29,7 @@ import javax.media.jai.JAI;
 import javax.media.jai.TiledImage;
 import org.geoserver.catalog.MetadataMap;
 import org.geoserver.data.util.CoverageUtils;
+import org.geoserver.platform.Operation;
 import org.geoserver.platform.ServiceException;
 import org.geoserver.wms.GetMapRequest;
 import org.geoserver.wms.MapLayerInfo;
@@ -89,6 +90,19 @@ public final class BilMapResponse extends RenderedImageMapResponse {
         super(OUTPUT_FORMATS, wms);
     }
 
+    @Override
+        GetMapRequest request = (GetMapRequest) operation.getParameters()[0];
+        if (request.getFormat().contains("16")) {
+            return OUTPUT_FORMATS[3];
+        } else if (request.getFormat().contains("32")) {
+            return OUTPUT_FORMATS[4];
+        } else {
+            return MIME_TYPE;
+        }
+    }
+
     @Override
     public void formatImageOutputStream(
             RenderedImage image, OutputStream outStream, WMSMapContent mapContent)
```

?

jinghm318:
@bradh Yes！

PR #3725, opened 2019-08-28
[GEOS-9311] Switching from System Status to Modules tab gives an error

nmco:

## Checklist

For all pull requests:


nmco:
Associated issue:
https://osgeo-org.atlassian.net/browse/GEOS-9311

This PRs fixes the above issues and adds a test case, it also adds the system status tab as a core tab directly (which it is now) instead of using the status page tabs extension points.

Wicket manages tab content, when we click on one of them,  by switching the current tab panel with the new clicked tab panel. System status tab contains a panel that is continuously refreshed, it involves sending a request to the server (Ajax) and get the new values back. The issue described in the issue above happen when we click on the system status tab, triggering an Ajax call, and then switching to another tab. Two things happen here:  wicket just switched the panels, so the system status panel is not available anymore, hence when the Ajax request comes back it doesn't find the system status panel.

To fix this a panel was added inside the tab panel and the refreshing behavior added to it.

nmco:
This is a bug and shoudl be backported to 2.15.x, since in that version system status is an extension and not core, the changes that add directly the system status tab in status page shoudl be, obviously, ignored.

PR #3729, opened 2019-09-01
[GEOS-9312] Cascaded WMS - Options to select and force Image Format and Styles in requests made to remote WMS server

imranrajjad:
-no unit test tess (TODO)
-for demo and initial review
-covers issue 1 and 3

## Checklist

For all pull requests:

- [] Confirm you have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/master/CONTRIBUTING.md) 
- [x ] Make sure the first PR targets the master branch, eventual backports will be managed later. This can be ignored if the PR is fixing an issue that only happens in a specific branch, but not in newer ones.

- [ x] There is a ticket in Jira describing the issue/improvement/feature (a notable exemptions is, changes not visible to end users)
- [ x] PR for bug fixes and small new features are presented as a single commit


imranrajjad:
JIRA ticket https://osgeo-org.atlassian.net/projects/GEOS/issues/GEOS-9312

imranrajjad:
It wont compile unless this PR is merged

https://osgeo-org.atlassian.net/projects/GEOT/issues/GEOT-6360

nmco:
I manually tested this, and it seems to be working as at the exception of the UI which puzzles me a bit: 

image

It is not clear to me why we have two times ``WMS Settings``, if the new options are only related with the remote WMS, then I would suggest something like ``Remote WMS Settings``.

Regarding the styles, why not having similar to the usual styles selection:
image

It would allow to select the default style and the available styles based on the styles supported by the remote WMS server. Right now we can basically just define the default style and it will be up to the user to find out the other ones, if they are available they shoudl show up on the local capabilities document.

Regarding the format selection, no sure what those option are for :yum:, IMHO we should have something similar to the available styles selection. On the left column the formats supported by the remote WMS server, on the right (after the user selection), the formats from the remote server we actually want to use, where the order defines the priority.

For example, let's say the remote WMS server supports PNG, PNG8, JPEG and GIF, but we only trust PNG and JPEG. Then we would select them as the available formats for usage, putting PNG first since it should be our fallback.

An user performs a request requesting JPEG, since that's an available format (locally) we use it to query the remote server.  

An user performs a request  requesting  GeoTIFF, which not available (neither locally or remotely), so we use our preferred format to query the remote WMS server which is PNG. 

An user performs a request  requesting  PNG8, which is not available  (locally), so we use our preferred format to query the remote WMS server which is PNG.

imranrajjad:
@nmco thanks for the detailed feedback

small note on the current behavior.

Lets say remote server supports PNG and JPEG and user has **unchecked** the Force Remote Format.

If local request is in JPEG and its supported in remote server, the remote server will get JPEG request, user does not need to do anything on UI.

If user wants to enforce a remote format, irrespective of the locally requested format, then user can check the **Force Remote Format** and make the remote server respond in JPEG even if the local request is on PNG, since its a light format

nmco:
What I propose does exactly that, but in a more user friendly way, and gives the possibility to have formats that indeed you want to delegate, or did I miss something?

imranrajjad:
no you did not miss anything, only difference that currently the whole list of remotely available styles is scanned against the locally requested image format (if checkbox is unchecked), we can limit the scan list by taking user preferences as per your suggestion

nmco:
One note on the above, IMHO by default we shoudl support everything the remote server supports. The user should only need to touch the UI if he wants to limit something, for example I don't want this style or format to be used.

imranrajjad:
got it, more like an ignore list

src/main/src/main/java/org/geoserver/catalog/CascadedLayerInfo.java
```diff
@@ -0,0 +1,43 @@
...
+ * application directory.
+ */
+package org.geoserver.catalog;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import org.geotools.styling.Style;
+
+/** @author ImranR */
```
nmco: Useless java doc, we have git blame. Javadoc  describing what this interfaces does and on its methods needs to be added.

src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java
```diff
@@ -1541,10 +1542,11 @@ public LayerInfo buildLayer(CoverageInfo coverage) throws IOException {
      */
     public LayerInfo buildLayer(WMSLayerInfo wms) throws IOException {
         LayerInfo layer = buildLayer((ResourceInfo) wms);
-
         layer.setDefaultStyle(getDefaultStyle(wms));
+        // geos-9312
```
nmco: Useless comment, tickets go way, I would rather add a comment describing what we are doing if you feel it necessary.

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
```diff
@@ -0,0 +1,254 @@
...
+        super.setEnabled(delegate.enabled());
+        super.setType(delegate.getType());
+    }
+
+    @Override
+    public void reset() {
+        // select all formats for use
+        selectedRemoteStyles.addAll(getRemoteStyles());
+        // set first style is selected
+        if (!selectedRemoteStyles.isEmpty()) forcedRemoteStyle = selectedRemoteStyles.get(0);
```
nmco: Is true that WMS capabilities documents don't allow the definition of a default style, so picking up the first one is the good approach:

> If the server advertises several styles for a layer, and the client sends a request for the default style, the choice of
which style to use as default is at the discretion of the server. The ordering of styles in the service metadata does
not indicate which is the default.

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
```diff
@@ -0,0 +1,254 @@
...
+            // read from cap doc
+            return layerInfo
+                    .getWMSLayer(null)
+                    .getStyles()
+                    .stream()
+                    .map(s -> s.getName())
+                    .collect(Collectors.toList());
+
+        } catch (Exception e) {
+            LOGGER.log(
```
nmco: Two comments here:
1. use the LOGGER native capabilities to log exception, the ``e.getMessage()`` if I'm not mistaken will only print the exception message not the stack trace.
2. I was expecting an exception to be throw here, what the reasoning for a simple log? IMHO if we cannot obtain the list of remote styles, then something is wrong.

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
```diff
@@ -0,0 +1,254 @@
...
+                    .getCapabilities()
+                    .getRequest()
+                    .getGetMap()
+                    .getFormats()
+                    .stream()
+                    .filter(CascadedLayerInfoImpl::isImage)
+                    . // only image formats
+                    collect(Collectors.toList());
+        } catch (IOException e) {
+            LOGGER.log(
```
nmco: Same comments as above.

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
```diff
@@ -0,0 +1,254 @@
...
+        return this.prefferedFormat;
+    }
+
+    @Override
+    public void setPrefferedFormat(String prefferedFormat) {
+        this.prefferedFormat = prefferedFormat;
+    }
+
+    @Override
+    public boolean isSelectedRemoteFormat(String format) {
```
nmco: The propose of this method is not clear, is it testing if the selected format is a valid remote format? In that case I would suggest to rename the method to something like _isFormatValid(String format)_.

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
```diff
@@ -0,0 +1,254 @@
...
+        if (prefferedFormat.equalsIgnoreCase(format)) return true;
+        else return selectedRemoteFormats.contains(format);
+    }
+
+    public static boolean isImage(String format) {
+        return format.startsWith("image") && !format.contains("xml") && !format.contains("svg");
```
nmco: At a first look this check looks quite fragile, but I may be wrong.

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    @Override
+    public void setPrefferedFormat(String prefferedFormat) {
+        this.prefferedFormat = prefferedFormat;
+    }
+
+    @Override
+    public boolean isSelectedRemoteFormat(String format) {
+        if (prefferedFormat.equalsIgnoreCase(format)) return true;
```
nmco: So at this stage we are 100% sure that preferred format is valid right?

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
nmco: Actually the real default is to send _""_, and that's what we shoudl be doing.

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
```diff
@@ -0,0 +1,254 @@
...
+        try {
+            return layerInfo
+                    .getWMSLayer(null)
+                    .getStyles()
+                    .stream()
+                    .map(CascadedLayerInfoImpl::getStyleInfo)
+                    .collect(Collectors.toSet());
+
+        } catch (IOException e) {
+            LOGGER.severe(e.getMessage());
```
nmco: Same comment as above for the others exceptions handling.

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
```diff
@@ -0,0 +1,254 @@
...
+        }
+        // on error default to super
+        return super.getStyles();
```
nmco: Why?

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
```diff
@@ -0,0 +1,254 @@
...
+        try {
+            return layerInfo
+                    .getWMSLayer(null)
+                    .getStyles()
+                    .stream()
+                    .filter(s -> s.getName().equalsIgnoreCase(name))
+                    .map(s -> style)
+                    .findFirst();
+        } catch (IOException e) {
+            LOGGER.severe(e.getMessage());
```
nmco: Same as above, maybe I'm missing something, but if we expect a behavior and it fails then we shoudl fail no? The remote style doesn't exists, how do we recover from that?

src/main/src/main/java/org/geoserver/catalog/impl/StyleInfoImpl.java
```diff
@@ -97,6 +97,9 @@ public void setFilename(String filename) {
     }
 
     public Style getStyle() throws IOException {
+        // for cap doc return remote name
+        if (metadata != null)
+            if (metadata.containsKey("isRemote")) return CascadedLayerInfoImpl.getStyleInfo(this);
         return catalog.getResourcePool().getStyle(this);
```
nmco: The two lines of code above are a unreadable for me, and the comment doesn't explain much.

src/web/wms/src/main/java/applicationContext.xml
```diff
@@ -59,5 +59,6 @@
      <property name="icon" value="palette.png"/>
      <property name="order" value="30"/>
      <property name="authorizer" ref="workspaceAdminAuthorizer"/>
-  </bean>  
+  </bean>
```
nmco: ?

src/web/wms/src/main/java/org/geoserver/wms/web/publish/WMSLayerConfig.html
```diff
@@ -49,6 +49,27 @@ <h3><wicket:message key="defaultTitle">WMS Settings</wicket:message></h3>
...
           </li>
         </ul>
+        <ul id="remotestyles" wicket:id="remotestyles">
+        <h3><wicket:message key="remoteDefaultTitle">Cascaded WMS Settings</wicket:message></h3>
+          <li>
+            <label for="remoteStylesDropDown"><wicket:message key="remoteStylesLabel">Select Default Remote Style</wicket:message></label>
+            <select id="remoteStylesDropDown" wicket:id="remoteStylesDropDown" class="text"/>
+          </li>
+          <li>
+            <label for="extraRemoteStyles">Select Additional Remote Styles</label>
```
nmco: The internationalization is missing, i.e. ``<wicket:message key="``.

src/web/wms/src/main/java/org/geoserver/wms/web/publish/WMSLayerConfig.html
```diff
@@ -49,6 +49,27 @@ <h3><wicket:message key="defaultTitle">WMS Settings</wicket:message></h3>
...
+            <span id="extraRemoteStyles" wicket:id="extraRemoteStyles"></span>
+          </li>
+        </ul>
+        <ul id="remoteformats" wicket:id="remoteformats">
+          <li>
+            <label for="remoteFormatsPalette">Select Default Remote Format</label>
```
nmco: The internationalization is missing, i.e. ``<wicket:message key="``.

src/web/wms/src/main/java/org/geoserver/wms/web/publish/WMSLayerConfig.html
```diff
@@ -49,6 +49,27 @@ <h3><wicket:message key="defaultTitle">WMS Settings</wicket:message></h3>
...
+            <select id="remoteFormatsDropDown" wicket:id="remoteFormatsDropDown" class="text"/>
+          </li>
+          <li>
+            <label for="remoteFormatsPalette">Select Remotely supported formats</label>
```
nmco: The internationalization is missing, i.e. ``<wicket:message key="``.

src/web/wms/src/main/java/org/geoserver/wms/web/publish/WMSLayerConfig.java
```diff
@@ -1,161 +1,237 @@
...
-            return new StringResourceModel(interpolationMethod.name(), parent).getString();
-        }
-
-        @Override
-        public String getIdValue(WMSInterpolation object, int index) {
-            return object.name();
-        }
-    }
-}
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
```
nmco: What happen here? Is impossible for me to understand what new or old.

src/web/wms/src/test/java/org/geoserver/wms/web/data/publish/WMSLayerConfigTest.java
```diff
@@ -167,4 +176,59 @@ public Component buildComponent(String id) {
...
+        cb.setStore(store);
+        store.setCapabilitiesURL(caps);
+        WMSLayerInfo wmsLayer = cb.buildWMSLayer("roads");
+        wmsLayer.setName("roads");
+        getCatalog().add(wmsLayer);
+        CascadedLayerInfo gsLayer = (CascadedLayerInfo) cb.buildLayer(wmsLayer);
+        gsLayer.reset();
+        getCatalog().add(gsLayer);
+
+        // final LayerInfo layer = getCatalog().getLayerByName("roads");
```
nmco: Left over commented code.

src/web/wms/src/test/java/org/geoserver/wms/web/data/publish/WMSLayerConfigTest.java
```diff
@@ -167,4 +176,59 @@ public Component buildComponent(String id) {
 
         tester.assertModelValue("form:panel:defaultInterpolationMethod", WMSInterpolation.Bicubic);
     }
+
+    @Test
+    public void testWMSCascadeSettings() throws Exception {
+        MockHttpClient wms11Client = new MockHttpClient();
+        URL wms11BaseURL = new URL(TestHttpClientProvider.MOCKSERVER + "/wms11");
+        URL capsDocument =
+                WMSLayerConfigTest.class.getResource("/org/geoserver/wms/web/data/caps111.xml");
```
nmco: If that file, which is added in this PR, is only used for this class, then it shoudl be added under `org.geoserver.wms.web.data.publish` and `this.getClass().getResource("caps111.xml")` used to retrieve it.

src/wms/src/main/java/org/geoserver/wms/GetMap.java
```diff
@@ -669,7 +669,24 @@ WebMap executeInternal(
...
+                    WMSLayer Layer = null;
+                    // GEOS-9312
+                    if (mapLayerInfo.getLayerInfo() instanceof CascadedLayerInfo) {
+                        CascadedLayerInfo cascadedLayerInfo =
+                                (CascadedLayerInfo) mapLayerInfo.getLayerInfo();
+
+                        String style = request.getStyles().get(i).getName();
+                        String imageFormat = request.getFormat();
+                        // if passed style does not exist in remote, default to forced
+                        if (!cascadedLayerInfo.isSelectedRemoteStyles(style))
```
nmco: If the passed style doesn't exists it shoudl fail, I mean right now if send a WMS requets to GeoServer with some invalid style I expect the request to fail, otherwise how do I know that requested style doesn't actually exists?

src/wms/src/main/java/org/geoserver/wms/GetMap.java
```diff
@@ -669,7 +669,24 @@ WebMap executeInternal(
...
+                            style = cascadedLayerInfo.getForcedRemoteStyle();
+
+                        // if the format is not selected then fall back to preffered
+                        if (!cascadedLayerInfo.isSelectedRemoteFormat(imageFormat))
```
nmco: This doesn't make much sense to me, if the client requests a PNG and it is not supported by the remote, then it should query the remote using whatever is the default format, but return a PNG to the client.

Another thing, that behavior should be handle by the WMS remote store, not by GeoServer, although I'm not sure at 100% this is possible.

src/wms/src/main/java/org/geoserver/wms/GetMap.java
```diff
@@ -669,7 +669,24 @@ WebMap executeInternal(
...
+                            imageFormat = cascadedLayerInfo.getPrefferedFormat();
+
+                        Layer = new WMSLayer(wms, gt2Layer, style, imageFormat);
+                    } else Layer = new WMSLayer(wms, gt2Layer);
```
nmco: Not sure what's the propose of this, shoudl not WMS layers always have a preferred image format and style?

src/wms/src/main/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformer.java
```diff
@@ -1033,6 +1034,14 @@ protected void handleLayer(final LayerInfo layer, boolean isRoot) throws IOExcep
                 handleRootCrsList(srsList);
             } else {
                 element("SRS", srs);
+                // overridingServiceSRS
```
nmco: What's going on here?

nmco:
Picture of the new UI options:
image

I would suggest the following alternatives for the highlighted labels:

1. Default Remote Style
1. Additional Remote Styles
1. Default Remote Format
1. Additional Remote Formats

I think they are more consistent with the GeoServer UI.

nmco:
Found two possible issues:
- the styles are not outputted in WMS 1.3.0 capabilities document, they are in WMS 1.1.0.
- performing a WMS request using an invalid style or remote style that is not selected as usable, doesn't throw an exception, but instead a map renderized using the default style.

nmco:
Another issue, when migrating from an existing data directory with a remote layer configuration, the layer configuration page doesn't allow us to edit the styles or image formats, is like the new functionalities are disabled for already existing remote layers, which doesn't make much sense:
image

IMHO, the existing layers shoudl be migrated to the new version (this a common behavior in GeoServer for migrations), the existing remote layers shoudl have the new options available.

src/main/src/main/java/org/geoserver/catalog/CascadedLayerInfo.java
imranrajjad: will remove it

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
imranrajjad: understood

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
imranrajjad: understood

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
imranrajjad: we only want image formats to show up in the lists (image/png, image/jpeg), no?

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
imranrajjad: yes don't look further if the default format matches before checking the complete list

src/main/src/main/java/org/geoserver/catalog/impl/StyleInfoImpl.java
imranrajjad: Allow me to explain

This is to make Cascaded layer advertise remote styles in GetCapability request. Since we have implemented the GetStyles and GetDefaultStyle in CascadedLayer, the code is trying to look for the Style (which is a remote style) in the catalog, but we never saved the Remote styles in Catalog (because they exist on another server and we don't want to them show up as local resources). So here we are looking for a flag in metadata map of Style to indicate that don't look in the catalog for this Style, just take it as it is.

src/web/wms/src/main/java/applicationContext.xml
imranrajjad: that's eclipse doing its thing, I will try to revert it

src/web/wms/src/main/java/org/geoserver/wms/web/publish/WMSLayerConfig.html
imranrajjad: understood will add that

src/wms/src/main/java/org/geoserver/wms/GetMap.java
imranrajjad: understood, if it goes not exist, we throw an exception

src/wms/src/main/java/org/geoserver/wms/GetMap.java
imranrajjad: this is to retrain backward compatbility, but it will not be required after migration is in place

src/wms/src/main/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformer.java
imranrajjad: advertising additionally supported SRS in WMS capability doc, as per our dicussion

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
imranrajjad: understood

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
imranrajjad: keeping native functionality intact, I am open to better altenate

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
imranrajjad: understood

src/main/src/main/java/org/geoserver/catalog/impl/StyleInfoImpl.java
imranrajjad: I will fix the comments

src/web/wms/src/test/java/org/geoserver/wms/web/data/publish/WMSLayerConfigTest.java
imranrajjad: understood

src/wms/src/main/java/org/geoserver/wms/GetMap.java
imranrajjad: the list comes from Capability document read during creation of store, even if its possible to re-query remote WMS service that will generate a lot of remote requests, No?

I will throw exception if the format is unknown

imranrajjad:
> Another issue, when migrating from an existing data directory with a remote layer configuration, the layer configuration page doesn't allow us to edit the styles or image formats, is like the new functionalities are disabled for already existing remote layers, which doesn't make much sense:
> 
> IMHO, the existing layers shoudl be migrated to the new version (this a common behavior in GeoServer for migrations), the existing remote layers shoudl have the new options available.

Understood, what if user wants to go back? or User has fired same data directory with an old Geoserver?

The Cascaded Layer gets persisted under a different XMl tag inside layer.xml. This will require to throw away the CascadedLayerInfo class and encapsulate all the functionality inside an extension of Metadata map object so that it can be instantiated for old layers and saved inside layer metadata. I am open to suggestion 🥇

imranrajjad:
> Found two possible issues:
> 
> the styles are not outputted in WMS 1.3.0 capabilities document, they are in WMS 1.1.0.
> performing a WMS request using an invalid style or remote style that is not selected as usable, doesn't throw an exception, but instead a map renderized using the default style.

understood, will be fixed

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
nmco: You could have an empty option, meaning use the default, would that make sense?

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
nmco: kk

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
nmco: kk

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
nmco: what do you mean by native functionality? In term of the user workflow I configure or I'm configuring  a remote layer, if I cannot obtain the remote styles then it shoudl crash no? and tell the user looks something bad happen when trying to reach the remote server and get the available styles.

src/main/src/main/java/org/geoserver/catalog/impl/StyleInfoImpl.java
nmco: kk.

src/web/wms/src/main/java/applicationContext.xml
nmco: kk

src/web/wms/src/main/java/org/geoserver/wms/web/publish/WMSLayerConfig.html
nmco: kk

src/web/wms/src/main/java/org/geoserver/wms/web/publish/WMSLayerConfig.html
nmco: kk

src/web/wms/src/main/java/org/geoserver/wms/web/publish/WMSLayerConfig.html
nmco: kk

src/web/wms/src/test/java/org/geoserver/wms/web/data/publish/WMSLayerConfigTest.java
nmco: kk

src/web/wms/src/test/java/org/geoserver/wms/web/data/publish/WMSLayerConfigTest.java
nmco: kk

src/wms/src/main/java/org/geoserver/wms/GetMap.java
nmco: kk

src/wms/src/main/java/org/geoserver/wms/GetMap.java
nmco: What do you mean by this: "I will throw exception if the format is unknown"?

`the list comes from Capability document read during creation of store, even if its possible to re-query remote WMS service that will generate a lot of remote requests, No?`

No you shoudl not query it each time.

src/wms/src/main/java/org/geoserver/wms/GetMap.java
nmco: kk

src/wms/src/main/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformer.java
nmco: No idea what you are talking about, This PR is about formats and styles.

nmco:
Regarding the migrations, I was wrong about this (thx @aaime for the heads up):

> IMHO, the existing layers shoudl be migrated to the new version (this a common behavior in GeoServer for migrations), the existing remote layers shoudl have the new options available.

**Migrations shoudl not happen automatically**, but when the admin opens the UI to edit an already existing WMS remote layer it should see and be able to edit the new configuration options. When the admin clicks save, then we migrate whatever needs to be migrated to accommodate the new functionalities.

> The Cascaded Layer gets persisted under a different XMl tag inside layer.xml. This will require to throw away the CascadedLayerInfo class and encapsulate all the functionality inside an extension of Metadata map object so that it can be instantiated for old layers and saved inside layer metadata. I am open to suggestion 1st_place_medal

No sure what you are talking about here, but we don't supports going backwards, for example, a GS data directory coming from a GS 2.15.x shoudl work whit a GS 2.16.x but the opposite is not true.

src/wms/src/main/java/org/geoserver/wms/GetMap.java
imranrajjad: understood

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
imranrajjad: understood and agreed 👍

src/wms/src/main/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformer.java
imranrajjad: ok this will be reverted

src/main/src/main/java/org/geoserver/catalog/impl/CascadedLayerInfoImpl.java
imranrajjad: agreed

src/web/wms/src/main/java/org/geoserver/wms/web/publish/WMSLayerConfig.java
imranrajjad: this is eclipse , I will fix it

src/wms/src/main/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformer.java
imranrajjad: ok will remove it

nmco:
Legit Travis compilation failure:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:testCompile (default-testCompile) on project gs-wms: Compilation failure: Compilation failure: 
[ERROR] /home/travis/build/geoserver/geoserver/src/wms/src/test/java/org/geoserver/wms/wms_1_3/WMSCascadeTest.java:[254,5] illegal start of expression
[ERROR] /home/travis/build/geoserver/geoserver/src/wms/src/test/java/org/geoserver/wms/wms_1_3/WMSCascadeTest.java:[254,12] illegal start of expression
[ERROR] /home/travis/build/geoserver/geoserver/src/wms/src/test/java/org/geoserver/wms/wms_1_3/WMSCascadeTest.java:[254,44] ';' expected
[ERROR] /home/travis/build/geoserver/geoserver/src/wms/src/test/java/org/geoserver/wms/wms_1_3/WMSCascadeTest.java:[254,54] not a statement
[ERROR] /home/travis/build/geoserver/geoserver/src/wms/src/test/java/org/geoserver/wms/wms_1_3/WMSCascadeTest.java:[254,63] ';' expected
[ERROR] /home/travis/build/geoserver/geoserver/src/wms/src/test/java/org/geoserver/wms/wms_1_3/WMSCascadeTest.java:[284,48] ';' expected
[ERROR] /home/travis/build/geoserver/geoserver/src/wms/src/test/java/org/geoserver/wms/wms_1_3/WMSCascadeTest.java:[284,58] not a statement
[ERROR] /home/travis/build/geoserver/geoserver/src/wms/src/test/java/org/geoserver/wms/wms_1_3/WMSCascadeTest.java:[284,67] ';' expected
[ERROR] /home/travis/build/geoserver/geoserver/src/wms/src/test/java/org/geoserver/wms/wms_1_3/WMSCascadeTest.java:[315,2] reached end of file while parsing
[ERROR] /home/travis/build/geoserver/geoserver/src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/WMSCascadeTest.java:[136,7] illegal start of expression
[ERROR] /home/travis/build/geoserver/geoserver/src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/WMSCascadeTest.java:[136,14] illegal start of expression
[ERROR] /home/travis/build/geoserver/geoserver/src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/WMSCascadeTest.java:[136,46] ';' expected
[ERROR] /home/travis/build/geoserver/geoserver/src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/WMSCascadeTest.java:[136,56] not a statement
[ERROR] /home/travis/build/geoserver/geoserver/src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/WMSCascadeTest.java:[136,65] ';' expected
[ERROR] /home/travis/build/geoserver/geoserver/src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/WMSCascadeTest.java:[166,48] ';' expected
[ERROR] /home/travis/build/geoserver/geoserver/src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/WMSCascadeTest.java:[166,58] not a statement
[ERROR] /home/travis/build/geoserver/geoserver/src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/WMSCascadeTest.java:[166,67] ';' expected
[ERROR] /home/travis/build/geoserver/geoserver/src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/WMSCascadeTest.java:[197,2] reached end of file while parsing
```

nmco:
Some is not right with the default style, now we can select empty to mean whatever is the default on the remote, but behind the scenes looks like the first one picked up stills being consider the default, and when querying the remote server if we use that style "" is used, which is wrong.

nmco:
The default style, empty should be selected by default:

image

nmco:
Wrong label used for the formats:
image

nmco:
FYK @imranrajjad no error is throw if a wrong style is send.

imranrajjad:
I will address these change when moving implementation into WMSLayerInfo from CascadedLayerInfo

imranrajjad:
https://osgeo-org.atlassian.net/projects/GEOT/issues/GEOT-6360 needs to be backported to 22.x Geotools

nmco:
Looks like a legit build failure:
```
[ERROR] Tests run: 10, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 2.844 s <<< FAILURE! - in org.geoserver.wms.wms_1_1_1.WMSCascadeTest
[ERROR] testCascadeGetLegendRequestJSON[0 APH enabled: true](org.geoserver.wms.wms_1_1_1.WMSCascadeTest)  Time elapsed: 0.063 s  <<< ERROR!
net.sf.json.JSONException: Invalid JSON String
	at org.geoserver.wms.wms_1_1_1.WMSCascadeTest.testCascadeGetLegendRequestJSON(WMSCascadeTest.java:180)

[ERROR] testCascadeGetLegendRequestJSON[1 APH enabled: false](org.geoserver.wms.wms_1_1_1.WMSCascadeTest)  Time elapsed: 0.009 s  <<< ERROR!
net.sf.json.JSONException: Invalid JSON String
	at org.geoserver.wms.wms_1_1_1.WMSCascadeTest.testCascadeGetLegendRequestJSON(WMSCascadeTest.java:180)
```

@imranrajjad the person how submit the pull request shoudl push forward to make sure the pull request is merged, this includes:

- communicating properly, saying things like: I have fixed this and that, I have addressed the review it shoudl be good now
- check that the automated builds are not failing

imranrajjad:
@nmco issues addressed, this is ready to be merged

PR #3739, opened 2019-09-07
[GEOS-9324] Add a parameter to GetLegendGraphic request using LEGEND_OPTIONS that allows to hide rules that are not matching any feature.

dersteppenwolf:
Add _hideEmptyRules_ parameter to LEGEND_OPTIONS in order to hide  rules that are not matching any feature.

Current legend including rules without features:
current

New legend using the  _hideEmptyRules_ parameter  removing rules without features:
expected


src/wms/src/main/java/org/geoserver/wms/legendgraphic/FeatureCountProcessor.java
```diff
@@ -231,9 +241,9 @@ public FeatureCountProcessor(GetLegendGraphicRequest request) {
             TargetLabelUpdater duplicatingVisitor = new TargetLabelUpdater(label);
```
ianturton: Do you have to clone the rule, or would a reference be ok? Probably doesn't matter but for complex rules this could take a while

doc/en/user/source/services/wms/get_legend_graphic/index.rst
```diff
@@ -94,6 +94,7 @@ Here is a description of the various parameters that can be used in ``LEGEND_OPT
     - **rows** enables **multirow** layout when layout is **horizontal**. The value is the the maximum rows number of legend. The columns used are equal to the next greater integer of <total of icons>/<number of rows>.
     - **grouplayout** Orientation of groups of layer, possible values are **horizontal** and **vertical** (default if not specified).
     - **countMatched** When set to true, adds at the end of each label the number of features matching that rule in the current map. Requires extra parameters, see details in the :ref:`dedicated section <content-dependent>`.
+    - **hideEmptyRules** When set to true and **countMatched** is also true, hides rules that are not matching any feature.
```
aaime: I believe it would be prefereable to have them independent. Something like: 

```suggestion
    + **hideEmptyRules** When set to true hides rules that are not matching any feature. 
```

If one wants both it's easy to add both parameters, the common behavior found in other software is to just hide without count (unfortunately last round of founding was instead about counting without hiding).

src/wms/src/main/java/org/geoserver/wms/legendgraphic/FeatureCountProcessor.java
aaime: I doubt it would make a significant difference, to get the count we had to do a data access and process all the rules (that seems at least an order of magnitude more expensive, or more).

dersteppenwolf:
Hi @aaime , 
The new commit implements your suggestion:
 _"When set to true hides rules that are not matching any feature"_

Best.

aaime:
Everything looks good, besides the docs build is failing with this warning:

``
     [exec] /home/aaime/devel/git-gs/doc/en/user/source/services/wms/get_legend_graphic/index.rst:97: WARNING: Bullet list ends without a blank line; unexpected unindent.
``

The bullet is a + instead of a - in the text ;-)

aaime:
Looks good, thanks! Merging.

dersteppenwolf:
Great!

PR #3743, opened 2019-09-09
GEOS-9138 csw: mapping file error reporting

NielsCharlier:


NielsCharlier:
thanks for the review @nmco

PR #3745, opened 2019-09-09
GEOS-9326 Hz-cluster: fix service events

NielsCharlier:
Hz-cluster: unlike for catalog items, for services only post-modify events are sent, not other events such as modify, add, delete
As a consequence the service config web pages are not cluster-safe


sikeoka:
Can you create a JIRA ticket for this?

NielsCharlier:
ok done

NielsCharlier:
thanks @sikeoka

sikeoka:
@NielsCharlier any change you could backport this?  Since this is a bug fix, it would be nice to have this at least on 2.16.x so someone doesn't have to wait until 2.17 for this to be fixed.

NielsCharlier:
@sikeoka https://github.com/geoserver/geoserver/pull/3786

PR #3794, opened 2019-09-27
[GEOS-9190]: Unable to create feature type through REST if two stores on different workspaces have the same name

groldan:
Fix a bug by which it's not possible to create a FeatureType
through the REST API on a store if another store with the same
name exists on a different Workspace (race condition: may succeed
if the FeatureTypeInfo request body deserialization picks up
the correct StoreInfo by chance)

## Checklist

For all pull requests:

### DIFF
--- a/src/restconfig/src/test/java/org/geoserver/rest/catalog/FeatureTypeControllerTest.java
+++ b/src/restconfig/src/test/java/org/geoserver/rest/catalog/FeatureTypeControllerTest.java
@@ -69,7 +69,35 @@ public class FeatureTypeControllerTest extends CatalogRESTTestSupport {
                 dom.getElementsByTagName("featureType").getLength());
     }
 
+    @Test // GEOS-9190
+    public void testCreateFeatureTypeSameStoreNameDifferentWorkspace() throws Exception {
+        final boolean configureFeatureType = false;
+        final String ws1 = "gs";
+        final String ws2 = "sf";
+        // create two stores named "pds" on different workspaces
+        addPropertyDataStore(ws1, configureFeatureType);
+        addPropertyDataStore(ws2, configureFeatureType);
+
+        String xml = "<featureType><name>pdsa</name><store>pds</store></featureType>";
+
+        MockHttpServletResponse response;
+        String ws1FetureTypesPath =
+                BASEPATH + "/workspaces/" + ws1 + "/datastores/pds/featuretypes";
+        String ws2FeatureTypesPath =
+                BASEPATH + "/workspaces/" + ws2 + "/datastores/pds/featuretypes";
+
+        response = postAsServletResponse(ws1FetureTypesPath, xml, "text/xml");
+        assertEquals(201, response.getStatus());
+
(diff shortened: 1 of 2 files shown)


src/restconfig/src/test/java/org/geoserver/rest/catalog/FeatureTypeControllerTest.java
```diff
@@ -69,7 +69,35 @@ public void testGetAllByWorkspace() throws Exception {
...
+    @Test // GEOS-9190
+    public void testCreateFeatureTypeSameStoreNameDifferentWorkspace() throws Exception {
+        final boolean configureFeatureType = false;
+        final String ws1 = "gs";
+        final String ws2 = "sf";
+        // create two stores named "pds" on different workspaces
+        addPropertyDataStore(ws1, configureFeatureType);
+        addPropertyDataStore(ws2, configureFeatureType);
+
+        String xml = "<featureType><name>pdsa</name><store>pds</store></featureType>";
```
aaime: I was wondering how this was happening, the deserializer is trying to get at the URI template variables and pick the workspaceName, is that part not working?
https://github.com/geoserver/geoserver/blob/fbcb98ae85fe81807ebc0463c5e42034c61c08e4/src/restconfig/src/main/java/org/geoserver/rest/catalog/FeatureTypeController.java#L511

groldan: When POSTing to `{ws}/datastores/{ds}/featuretypes` there's no `featureTypeName` URI template variable at all. See immediately bellow the lines you mentioned:
```java
if (workspace == null || datastore == null || featuretype == null) {
    return null;
}
```
The configured persister seems ok to lookup existing FT's but can't resolve the correct wfs/ds for a new one, so that method ultimately ends up at `AbstractReflectionConverter.doUnmarshal(...)` retrieves the `store` field for the FT as `value = unmarshallField(context, result, type, field);`. `context` is the XStream `UnmarshallingContext` object and lacks the context information and hence the logic to retrieve the necessary store from the correct workspace.

The call chaing reaches `XStreamPersister$RerefenceConverter.unmarshal() -> ResolvingProxy.resolve(Catalog catalog, T ref)`, with `ref` being a `StoreInfo` proxy whose name is the required store name (e.g. `pds`), and [here](https://github.com/geoserver/geoserver/blob/290813cbc463e68459b7eb7d8e6b031b0e192657/src/main/src/main/java/org/geoserver/catalog/impl/ResolvingProxy.java#L92)'s where the magic (doesn't) happen: `resolved = (T) catalog.getStoreByName(ref, StoreInfo.class);` ends up using the default workspace.

All that said, I think my fix for this specific issue should be good enough, since touching anything else would imply digging deeper into fundamental geoserver API issues such as the validity of deferring calls to the default workspace.

The issue here is, if I POST a FT to a given ws/store, that one should be used. The FT body being incomplete is assumed and supposed to be fulfilled by the API (e.g. in this case I don't need to provide a full FeatureTypeInfo). Either that or the API is strengthened to expect and require full objects as request bodies, which would break the API altogether.

PR #3804, opened 2019-10-02
[GEOS-9355] implemented support to use params-extractor filter as a standard servlet filter, to intercept the security filters chain

mbarto:
Allows the params-extract filter to run before the security filters chain, so that parameters used by those filters can be succesfully extracted and used (for example by the authkey module)

## Checklist

For all pull requests:

### DIFF
--- a/src/community/params-extractor/src/main/java/org/geoserver/params/extractor/Filter.java
+++ b/src/community/params-extractor/src/main/java/org/geoserver/params/extractor/Filter.java
@@ -78,22 +78,21 @@ public final class Filter implements GeoServerFilter, ExtensionPriority {
             HttpServletRequest httpServletRequest = (HttpServletRequest) request;
             if (httpServletRequest.getRequestURI().contains("web/wicket")
                     || httpServletRequest.getRequestURI().contains("geoserver/web")) {
-                chain.doFilter(request, response);
-                return;
-            }
-            UrlTransform urlTransform =
-                    new UrlTransform(
-                            httpServletRequest.getRequestURI(),
-                            httpServletRequest.getParameterMap());
-            String originalRequest = urlTransform.toString();
-            rules.forEach(rule -> rule.apply(urlTransform));
-            if (urlTransform.haveChanged()) {
-                Utils.info(
-                        LOGGER,
-                        "Request '%s' transformed to '%s'.",
-                        originalRequest,
-                        urlTransform.toString());
-                request = new RequestWrapper(urlTransform, httpServletRequest);
+            } else {
+                UrlTransform urlTransform =
+                        new UrlTransform(
+                                httpServletRequest.getRequestURI(),
+                                httpServletRequest.getParameterMap());
+                String originalRequest = urlTransform.toString();
+                rules.forEach(rule -> rule.apply(urlTransform));
+                if (urlTransform.haveChanged()) {
+                    Utils.info(
+                            LOGGER,
+                            "Request '%s' transformed to '%s'.",
+                            originalRequest,
+                            urlTransform.toString());
+                    request = new RequestWrapper(urlTransform, httpServletRequest);
+                }
             }
         }
         chain.doFilter(request, response);
--- a/src/community/params-extractor/src/main/java/org/geoserver/params/extractor/Filter.java
+++ b/src/community/params-extractor/src/main/java/org/geoserver/params/extractor/Filter.java
@@ -7,11 +7,16 @@ package org.geoserver.params.extractor;
 import java.io.IOException;
 import java.util.List;
 import java.util.logging.Logger;
-import javax.servlet.*;
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
 import org.geoserver.config.GeoServerDataDirectory;
 import org.geoserver.filters.GeoServerFilter;
 import org.geoserver.platform.ExtensionPriority;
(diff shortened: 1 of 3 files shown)


src/community/params-extractor/src/main/java/org/geoserver/params/extractor/Filter.java
```diff
@@ -33,34 +47,52 @@ public int getPriority() {
     }
 
     @Override
-    public void init(FilterConfig filterConfig) throws ServletException {}
+    public void init(FilterConfig filterConfig) throws ServletException {
```
aaime: So this method does not get called when the bean is just picked up from the spring context, but only when using it as a servlet filter, right? I'd clarify this.

src/community/params-extractor/src/main/java/org/geoserver/params/extractor/Filter.java
```diff
@@ -7,21 +7,35 @@
...
 import org.geoserver.platform.resource.Resource;
 import org.geotools.util.logging.Logging;
 
 public final class Filter implements GeoServerFilter, ExtensionPriority {
 
     private static final Logger LOGGER = Logging.getLogger(Filter.class);
 
+    static boolean USE_AS_SERVLET_FILTER = false;
+
+    private boolean servletInstance = false;
```
aaime: Two flags... so the static one is used to know that there is an instance used as a servlet filter, and the other one to know which one is the servlet filter, to disable the other? I'd add comments to clarify it, was not obvious to me at a first read.

src/community/params-extractor/src/main/java/org/geoserver/params/extractor/Filter.java
```diff
@@ -33,34 +47,52 @@ public int getPriority() {
...
+            rules.forEach(rule -> rule.apply(urlTransform));
+            if (urlTransform.haveChanged()) {
+                Utils.info(
+                        LOGGER,
+                        "Request '%s' transformed to '%s'.",
+                        originalRequest,
+                        urlTransform.toString());
+                chain.doFilter(new RequestWrapper(urlTransform, httpServletRequest), response);
+            } else {
+                chain.doFilter(request, response);
```
aaime: Code is getting a bit convoluted with 3 places that call chain.doFilter... could it be flipped so that there is a single one calling onto doFilter, and the one that is active, which just replaces request with the wrapper?

src/community/params-extractor/src/test/java/org/geoserver/params/extractor/FilterTest.java
```diff
@@ -0,0 +1,86 @@
...
+import org.geoserver.config.GeoServerDataDirectory;
+import org.junit.Before;
+import org.junit.Test;
+
+public class FilterTest extends TestSupport {
+    FilterConfig filterConfig;
+
+    @Before
+    public void setUp() {
+        filterConfig =
```
aaime: Can probably just use MockFilterConfig from Spring?

src/community/params-extractor/src/main/java/org/geoserver/params/extractor/Filter.java
```diff
@@ -33,34 +47,52 @@ public int getPriority() {
     }
 
     @Override
-    public void init(FilterConfig filterConfig) throws ServletException {}
+    public void init(FilterConfig filterConfig) throws ServletException {
+        GeoServerDataDirectory dataDirectory =
+                GeoServerExtensions.bean(GeoServerDataDirectory.class);
+        if (dataDirectory != null) {
+            USE_AS_SERVLET_FILTER = true;
```
aaime: Does it make sense to move this to the no argument constructor, so that we have the management of the two flags in the same place?

src/community/params-extractor/src/main/java/org/geoserver/params/extractor/Filter.java
```diff
@@ -33,34 +47,52 @@ public int getPriority() {
...
+            Resource resource = dataDirectory.get(RulesDao.getRulesPath());
```
aaime: This code duplicates the constructor receiving a data dir, can it be shared in a method that both call?

src/community/params-extractor/src/main/java/org/geoserver/params/extractor/Filter.java
mbarto: yes, should I write a comment on the method?

aaime: Yes please :-)

src/community/params-extractor/src/main/java/org/geoserver/params/extractor/Filter.java
```diff
@@ -7,58 +7,100 @@
...
-                    originalRequest,
-                    urlTransform.toString());
-            chain.doFilter(new RequestWrapper(urlTransform, httpServletRequest), response);
-        } else {
-            chain.doFilter(request, response);
+        if (isEnabled()) {
+            HttpServletRequest httpServletRequest = (HttpServletRequest) request;
+            if (httpServletRequest.getRequestURI().contains("web/wicket")
+                    || httpServletRequest.getRequestURI().contains("geoserver/web")) {
+                chain.doFilter(request, response);
```
aaime: Can we have just one call to doFilter? This if could have an else.

mbarto: done

src/community/params-extractor/src/main/java/org/geoserver/params/extractor/Filter.java
mbarto: done

aaime:
Looks good to me, let's see if the build checks like it too

PR #3839, opened 2019-10-19
update git config options

jodygarnett:
Signed-off-by: Jody Garnett <jody.garnett@gmail.com>

Updating developers guide to address reoccurring file unicode issue.

## Checklist

For all pull requests:


doc/en/developer/source/source.rst
```diff
@@ -166,30 +169,112 @@ Each branch has the following structure::
...
+
++-------------------------+---------------+------------------+
+| Representation          | Example       | Operating System |
++=========================+===============+==================+
+| Precomposed form        | ``Ü``         | Linux, Windows   |
++-------------------------+---------------+------------------+
+| Decomposed form         | ``U`` + ``¨`` | macOS            |
++-------------------------+---------------+------------------+
+
+Care must be taken by macOS configure the git client appropriately to committing files in decomposed form.
```
mprins: Care must be taken by macOS users to configure the git client appropriately for committing files in decomposed form.

doc/en/developer/source/source.rst
```diff
@@ -12,28 +12,128 @@ To clone the repository::
...
+
+.. code-block:: bash
+
+   $ git config --global --get-regexp core.*
+
+::
+
+   core.autocrlf input
+   core.safecrlf true
+   core.ignorecase true
```
aaime: If thise setting does not make sense for case sensitive file systems please mark it as a separate setting in a note.

jodygarnett: I will provide a second example then.

jodygarnett: I have split this into two examples, and after research/reflection am now asking Mac and Windows users to set this value to `true` (so even though their file system does not care - we are asking the gitclient to care so that files work correctly on Linux).

I also provided an example of how to fix any mistakes using `git mv`.

jodygarnett:
@tbarsballe as a committer working on `macOS` can I ask you to test this once it is merged.

jodygarnett:
I would like to backport this change as it "fixes" file path committed in decomposed (two character) representation.

PR #3852, opened 2019-10-21
[GEOS-9377] Geoserver startup gets blocked with unresponsive WMSStore

imranrajjad:
-moved remote calls for styles out of getters
-remote calls for styles will only be made when configuring/reloading Cascaded WMS Layer from UI

## Checklist

For all pull requests:

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/WMSLayerInfo.java
+++ b/src/main/src/main/java/org/geoserver/catalog/WMSLayerInfo.java
@@ -54,4 +54,10 @@ public interface WMSLayerInfo extends ResourceInfo {
     public Set<StyleInfo> getStyles();
 
     public StyleInfo getDefaultStyle();
+
+    public void reloadRemoteStyles();
+
+    public List<StyleInfo> getAllAvailableRemoteStyles();
+
+    public void setAllAvailableRemoteStyles(List<StyleInfo> allAvailableRemoteStyles);
 }
--- a/src/main/src/main/java/org/geoserver/catalog/WMSLayerInfo.java
+++ b/src/main/src/main/java/org/geoserver/catalog/WMSLayerInfo.java
@@ -47,17 +47,13 @@ public interface WMSLayerInfo extends ResourceInfo {
 
     void reset();
 
-    public String getPrefferedFormat();
+    public String getPreferredFormat();
 
-    public void setPrefferedFormat(String prefferedFormat);
+    public void setPreferredFormat(String prefferedFormat);
 
     public Set<StyleInfo> getStyles();
 
     public StyleInfo getDefaultStyle();
 
-    public void reloadRemoteStyles();
-
     public List<StyleInfo> getAllAvailableRemoteStyles();
-
-    public void setAllAvailableRemoteStyles(List<StyleInfo> allAvailableRemoteStyles);
 }
--- a/src/main/src/main/java/org/geoserver/catalog/impl/WMSLayerInfoImpl.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/WMSLayerInfoImpl.java
@@ -35,6 +35,8 @@ public class WMSLayerInfoImpl extends ResourceInfoImpl implements WMSLayerInfo {
 
     private List<String> selectedRemoteStyles = new ArrayList<String>();
 
+    private List<StyleInfo> allAvailableRemoteStyles = new ArrayList<StyleInfo>();
+
     protected WMSLayerInfoImpl() {}
 
     public WMSLayerInfoImpl(Catalog catalog) {
@@ -58,6 +60,7 @@ public class WMSLayerInfoImpl extends ResourceInfoImpl implements WMSLayerInfo {
     public void reset() {
         selectedRemoteStyles.clear();
         selectedRemoteFormats.clear();
+        reloadRemoteStyles();
         // select all formats for use
         selectedRemoteStyles.addAll(remoteStyles());
         // set empty to take whatever is on remote server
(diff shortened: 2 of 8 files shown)


imranrajjad:
@aaime @nmco review please

nmco:
` -remote calls for styles will only be made when configuring/reloading Cascaded WMS Layer from UI`
This is not clear to me, can you elaborate a bit more? what will be the workflow:
- when GeoServer starts
- when we query the layer
- when reload the remote store

... and finally what does this PR actually chnages on the current behavior.

imranrajjad:
> -remote calls for styles will only be made when configuring/reloading Cascaded WMS Layer from UI
> This is not clear to me, can you elaborate a bit more? what will be the workflow:
> 
> when GeoServer starts
> when we query the layer
> when reload the remote store
> 
> ... and finally what does this PR actually chnages on the current behavior.

**Current Behavior**

- Remote styles calls are inside getStyles() method, which as @aaime pointed out should not have active code and they should only act as access handlers
- When Geoserver starts up and XML is de-persister, while generically invoking getters / setters the remote call is sent (when Geoserver has not even loaded the Global settings), and hence the delay/blocker on startup

**Changes in PR**

- The remote call to fetch styles have been moved to a separate utility method, which is only invoked through the UI, not on Startup or Store reload.
- All available Styles are persisted inside WMSLayerInfoImpl and no remote calls are required at startup.

Hope that clarifies

src/main/src/main/java/org/geoserver/catalog/WMSLayerInfo.java
```diff
@@ -54,4 +54,10 @@
     public Set<StyleInfo> getStyles();
 
     public StyleInfo getDefaultStyle();
+
+    public void reloadRemoteStyles();
+
+    public List<StyleInfo> getAllAvailableRemoteStyles();
+
+    public void setAllAvailableRemoteStyles(List<StyleInfo> allAvailableRemoteStyles);
```
aaime: In the GeoServer configuration there should not be setters for collections, all collections should be "live collections", that is, you only have a getter, the collection returned is never null, and you can modify it directly.

I see that other recent changes ignored it, there are setters for remote styles, also, it's "preferred", not "preffered" (typo in method name).

src/main/src/main/java/org/geoserver/catalog/impl/WMSLayerInfoImpl.java
```diff
@@ -125,7 +126,7 @@ public String getPrefferedFormat() {
 
     @Override
     public void setPrefferedFormat(String prefferedFormat) {
-        this.prefferedFormat = prefferedFormat;
+        this.prefferedFormat = (prefferedFormat == null) ? "image/png" : prefferedFormat;
```
aaime: "preferred" here too.

src/main/src/main/java/org/geoserver/catalog/impl/WMSLayerInfoImpl.java
```diff
@@ -255,4 +256,16 @@ public void setSelectedRemoteFormats(List<String> selectedRemoteFormats) {
     public void setSelectedRemoteStyles(List<String> selectedRemoteStyles) {
         this.selectedRemoteStyles = selectedRemoteStyles;
     }
+
+    public void reloadRemoteStyles() {
```
aaime: Does not really need to be part of the interface, the UI can also do "getAllAvailableRemoteStyles().clear()"
and then "getAllAvailableRemoveStyles().addAll(getRemoveStyleInfo())" (using live collections).

I'd also call it just "remoteStyles" without further qualifications.

aaime:
Btw, checked, the GeoServer startup is not blocked any longer.

src/main/src/main/java/org/geoserver/catalog/impl/WMSLayerInfoImpl.java
imranrajjad: there is a high chance, this spelling change in methods will break the XML persist/de-persist, as in might cause data directory issues.

anyway will correct the names

aaime: You are indeed right, the old xml entry will be ignored. However, this is a recent change that landed on master only, right? So we are not breaking any backwards compat with released software.

src/main/src/main/java/org/geoserver/catalog/WMSLayerInfo.java
imranrajjad: Current implementation persists the available list of StyleInfo inside the WMSLayerInfo, because the StyleInfo carries remote URL of style itself on host server, which is used in cascading legend requests and generating capability document.

But your point makes sense, I will remove the live collection outside the bean class and use it only when needed, e.g UI, generating capability document etc

imranrajjad: will remove the getter

src/main/src/main/java/org/geoserver/catalog/impl/WMSLayerInfoImpl.java
imranrajjad: will make the change

imranrajjad:
@aaime thank you for your review, I have applied the requested changes, please let me know if any more changes or explanation is required?

src/main/src/main/java/org/geoserver/catalog/impl/WMSLayerInfoImpl.java
imranrajjad: done

aaime:
Seems ok, merging, thanks!

PR #3860, opened 2019-10-26
[GEOS-9191] FileSystemWatcher: 100% CPU usage while idle running on a large data directory

groldan:
Running against a rather large data directory leads `FileSystemWatcher`'s polling thread to be fixed at  100% usage on one CPU core (which one varies every a handful of seconds).

The bellow [data_dir_100k.tgz](https://drive.google.com/open?id=13nhQz_0QGm459muwQ2Qgcre_5HaXvshM) archive contains a data directory with +86K layers and has been used to conduct the testing for this patch.

This performance issue with `FileSystemWatcher` reveals itself since `DefaultTileLayerCatalog` started listening to changes to the `<data_dir>/gwc-layers` directory (which contains tile layer config files, not tile images) in order to update tile layer configs when they're updated externally on a cluster deployment (added at   https://github.com/geoserver/geoserver/commit/966a8af90fc6b68725a78b09829c3cbc7360242e#diff-54440e68426497a15eea28040661977aR93.

Profiling and the CPU bound nature of the issue showed the culprit has nothing to do with listing the directory contents, which takes just some milliseconds, but in the way `FileSystemWatcher` computed the changed, created, and removed files inside it, by calling `List.removeAll(Collection)` several times on lists of `java.io.File` objects.

This patch uses `java.util.Set` semantics instead and improves the logic to not make copies of collections, with the following results:

Baseline: [master@e23559d4](e23559d4a366e45a0bcf44811393cc718b13ef1f)
OS: GNU/Linux Ubuntu 19.04 (kernel 5.0.0-29-generic)
JVM: OpenJDK 64-Bit Server VM (Zulu 8.40.0.25-CA-linux64) (build 25.222-b10, mixed mode)
Processor: AMD Threadripper 1950X (16 cores, 32 threads)
Datadir filesystem: 
JVM args: -Xms4G -Xmx4G
```
                  Baseline       This PR
Poll runtime    ~7.5-8.5 mins    ~200ms
%CPU             100%             ~20-25%
Heap old gen     ~0               ~0
Heap eden space ~50MB        ~50MB
```
JVM args: -Xms1G -Xmx1G
```text
                  Baseline       This PR
Poll runtime     ~8-10 mins       ~200ms
%CPU             100%             ~20-25%
Heap old gen     ~0               ~0
Heap eden space ~50MB        ~50MB
```
Note on Eden space utilization: 
Each poll run consistently increases the amount of Eden space usage by about 50MB. That means that, with 4G max heap size, it takes about 20-22 runs to reach the point where the GC will free that space (at around 1.1GB).
There's nothing we can do to reduce the amount of eden space heap usage, since almost 100% of it happens at `File.listFiles()` inside the native call to `java.io.Filesystem.list():String[]` call. Using NIO's `DirectoryStream` from `Files#newDirectoryStream(Path)` instead of `File.listFiles()` results in no runtime improvement, and increases the temporary memory peak considerably, although there might be an inflexion point where it's more beneficial than loading the whole list of `File` objects, haven't reached it. Also, `File.listFiles()` is lighter weight than `File.list():String[]` due to some internal implementation details.

---

Additionally, `FileSystemWatcher` related code's readability has been improved by removing lots of unnecessary null checks on objects that can't be null and splitting logic into smaller units.
`FileSystemResourceTheoryTest` test suite got  some improvements, and dcd53c3b713971004ca4c9c772d01592b66bcd31 fixes a potential issue with `FileSystemResourceStore.fileSystemWatcher`'s lazy initialization.

---

## Checklist

For all pull requests:

- [N/A] Commits changing the UI, existing user workflows, or adding new functionality, need to include documentation updates

### DIFF
--- a/src/platform/pom.xml
+++ b/src/platform/pom.xml
@@ -98,5 +98,10 @@
     <artifactId>guava</artifactId>
     <scope>test</scope>
   </dependency>
+  <dependency>
+    <groupId>org.awaitility</groupId>
+    <artifactId>awaitility</artifactId>
+    <scope>test</scope>
+  </dependency>
  </dependencies>
 </project>
--- a/src/platform/src/main/java/org/geoserver/platform/resource/FileSystemWatcher.java
+++ b/src/platform/src/main/java/org/geoserver/platform/resource/FileSystemWatcher.java
@@ -5,19 +5,33 @@
  */
 package org.geoserver.platform.resource;
 
+import static java.util.concurrent.TimeUnit.MICROSECONDS;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+
 import java.io.File;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.EnumMap;
+import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 import org.geoserver.platform.resource.ResourceNotification.Kind;
+import org.geotools.util.logging.Logging;
 import org.springframework.beans.factory.DisposableBean;
 import org.springframework.scheduling.concurrent.CustomizableThreadFactory;
 
@@ -35,41 +49,45 @@ import org.springframework.scheduling.concurrent.CustomizableThreadFactory;
  */
 public class FileSystemWatcher implements ResourceNotificationDispatcher, DisposableBean {
 
-    interface FileExtractor {
-        public File getFile(String path);
-    }
+    private static final Logger LOGGER = Logging.getLogger(FileSystemWatcher.class);
 
     /** Change to file system */
     static class Delta {
+        /** Watched directory where changes occurred */
         final File context;
 
         final Kind kind;
 
-        final List<File> created;
+        /** Paths relative to the context directory that have been created */
+        final List<String> created;
 
-        final List<File> removed;
+        /** Paths relative to the context directory that have been removed */
+        final List<String> removed;
 
-        final List<File> modified;
+        /** Paths relative to the context directory that have been modified */
+        final List<String> modified;
 
         public Delta(File context, Kind kind) {
             this.context = context;
             this.kind = kind;
-            this.created = null;
-            this.removed = null;
-            this.modified = null;
+            this.created = this.removed = this.modified = Collections.emptyList();
         }
 
         public Delta(
                 File context,
                 Kind kind,
-                List<File> created,
-                List<File> removed,
-                List<File> modified) {
+                List<String> created,
+                List<String> removed,
+                List<String> modified) {
             this.context = context;
             this.kind = Kind.ENTRY_MODIFY;
-            this.created = created != null ? created : (List<File>) Collections.EMPTY_LIST;
-            this.removed = removed != null ? removed : (List<File>) Collections.EMPTY_LIST;
-            this.modified = modified != null ? modified : (List<File>) Collections.EMPTY_LIST;
+            this.created = created == null ? Collections.emptyList() : created;
+            this.removed = removed == null ? Collections.emptyList() : removed;
+            this.modified = modified == null ? Collections.emptyList() : modified;
+        }
+
+        public int size() {
+            return created.size() + removed.size() + modified.size();
         }
 
         @Override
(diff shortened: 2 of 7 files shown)


groldan:
A small correction: problem exposed at commit 5c11618a447ec2dd34539c13c8db8190acd69a47 instead of 966a8af90fc6b68725a78b09829c3cbc7360242e (i.e. when switched to adding a single listener to the gwc-layers directory instead of one listener per file inside it)

aaime:
Probably better if the original author of FileSystemWatcher has a look. @jodygarnett ?

src/platform/src/main/java/org/geoserver/platform/resource/FileSystemWatcher.java
```diff
@@ -35,41 +49,41 @@
...
+    private static final Logger LOGGER = Logging.getLogger(FileSystemWatcher.class);
 
     /** Change to file system */
     static class Delta {
         final File context;
 
         final Kind kind;
 
-        final List<File> created;
+        final List<String> created;
```
jodygarnett: What is the reason to move from File to Strings (assume the strings are paths).

src/platform/src/test/java/org/geoserver/platform/resource/FileSystemResourceTheoryTest.java
```diff
@@ -100,39 +111,50 @@ public void fileEvents() throws Exception {
...
-        assertNull("No events expected", n);
+        ResourceNotification notification =
+                new ResourceNotification(".", Kind.ENTRY_CREATE, 1_000_000L);
+        CompletableFuture.runAsync(
+                () -> {
+                    // try {
+                    // TimeUnit.MILLISECONDS.sleep(50);
+                    // } catch (InterruptedException e) {
+                    // throw new RuntimeException(e);
+                    // }
```
jodygarnett: remove commented out code

src/platform/src/test/java/org/geoserver/platform/resource/FileSystemResourceTheoryTest.java
```diff
@@ -181,87 +204,27 @@ public void directoryEvents() throws Exception {
...
-        }
-
-        @Override
-        public String toString() {
-            return "Await [event=" + event + "]";
+        public ResourceNotification await(int timeout, TimeUnit unit) {
+            return Awaitility.await()
+                    .pollInterval(5, TimeUnit.MILLISECONDS)
+                    .atMost(timeout, unit)
+                    .untilAtomic(this.reference, IsNull.notNullValue());
```
jodygarnett: I am happy to see *Await* class replaced

jodygarnett:
I like this patch/improvement - I think it can be accepted (although I would prefer watching individual tile sets rather than the whole gwc-layers directory).

src/platform/pom.xml
```diff
@@ -98,5 +98,10 @@
     <artifactId>guava</artifactId>
     <scope>test</scope>
   </dependency>
+  <dependency>
+    <groupId>org.awaitility</groupId>
+    <artifactId>awaitility</artifactId>
```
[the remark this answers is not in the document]
groldan: yeah it's great to assert on async processes. Seems pretty mature and alive: https://github.com/awaitility/awaitility
being a test dependency I hope there's no objection.

src/platform/src/main/java/org/geoserver/platform/resource/FileSystemWatcher.java
groldan: Because it more closely represents its intent (context directory + relative resource paths), which is exactly what the end-user of this data expects (`ResourceNotification`). Otherwise `ResourceNotification` needs to convert the created/removed/modified files back to relative paths: https://github.com/geoserver/geoserver/pull/3860/files#diff-4edea39a9ddbb6e029dc5a4a6974b7bfL133
Which is also quite some processing and temporary objects we can avoid as we have that information handy at the time of creating the `Delta`: https://github.com/geoserver/geoserver/pull/3860/files#diff-00129441b1b220c30473e61ec5339cc2R245

groldan:
> I like this patch/improvement - I think it can be accepted (although I would prefer watching individual tile sets rather than the whole gwc-layers directory).

Having one listener per file in `gwc-layers` was even worse, which was amended at commit 5c11618a447

So to be clear, this patch aims at fixing the current huge performance penalty of listening for changes to a directory.

Given the gwc-layers listener was added to make cluster nodes aware of changes to gwc layer configurations, I think that
a- this should be disabled by default and enabled through a system property or something like that
b- another mechanism should be used to notify nodes on a cluster of gwc config changes. If using the JMS clustering extension, then JMS events. Otherwise, this could still be enabled or a proper mechanism be implemented for whatever clustering orchestration solution is used.

> Had a quick look, my understanding is this is still a work in progress
It's supposed to be ready actually.

src/platform/src/test/java/org/geoserver/platform/resource/FileSystemResourceTheoryTest.java
groldan: done

groldan:
@jodygarnett ok, I've fixed the javadoc formatting issue that was making the java 11 build fail.
What about backporting to 2.16.x? and 2.15.x?
I've some 2.15.x production instances that would benefit if the fix makes it into next week's `2.16.1` release.

jodygarnett:
Add the backport tags, I consider this a bug fix.

PR #3871, opened 2019-10-30
[GEOS-9388] App-Schema WFS GetFeature includes namespaces for all workspaces on isolated workspace

fernandor777:
App-Schema WFS GetFeature GML 3.1 and 3.2 reponses include namespaces from all workspaces on isolated workspace (virtual service) request.  This PR fixes this issue.

Requires:
https://github.com/geotools/geotools/pull/2674

JIRA issue:
https://osgeo-org.atlassian.net/browse/GEOS-9388

## Checklist

For all pull requests:

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/xml/FeatureTypeSchemaBuilder.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/xml/FeatureTypeSchemaBuilder.java
@@ -9,6 +9,7 @@ import static org.geoserver.ows.util.ResponseUtils.buildURL;
 import static org.geoserver.ows.util.ResponseUtils.params;
 
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
@@ -17,6 +18,8 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.logging.Level;
@@ -52,9 +55,11 @@ import org.geoserver.platform.resource.Resource.Type;
 import org.geoserver.wfs.GMLInfo;
 import org.geoserver.wfs.WFSInfo;
 import org.geotools.feature.NameImpl;
+import org.geotools.feature.type.Types;
 import org.geotools.gml2.GMLConfiguration;
 import org.geotools.gml3.v3_2.GML;
 import org.geotools.wfs.v2_0.WFS;
+import org.geotools.xlink.XLINK;
 import org.geotools.xs.XS;
 import org.geotools.xsd.Configuration;
 import org.geotools.xsd.Schemas;
@@ -197,7 +202,7 @@ public abstract class FeatureTypeSchemaBuilder {
 
             if (!simple) {
                 // complex features may belong to different workspaces
-                addAllNamespacesFromCatalog(schema);
+                addRequiredNamespaces(featureTypeInfos, schema);
             }
 
             // would result in some xsd:include or xsd:import if schema location is specified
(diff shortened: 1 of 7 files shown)


fernandor777:
This is currently a work in progress.

src/wfs/src/main/java/org/geoserver/wfs/xml/FeatureTypeSchemaBuilder.java
```diff
@@ -365,20 +369,63 @@ public final XSDSchema buildSchemaInternal(
...
+                final Map<String, String> schemaNamespacesMap =
+                        schema.getQNamePrefixToNamespaceMap();
+                final List<NamespaceInfo> catalogNamespaces = catalog.getNamespaces();
+                for (FeatureTypeInfo featureTypeInfo : featureTypeInfos) {
+                    Object mapObject =
+                            featureTypeInfo
+                                    .getFeatureType()
+                                    .getUserData()
+                                    .get(Types.DECLARED_NAMESPACES_MAP);
+                    if (!(mapObject instanceof Map)) continue;
```
aaime: Just to make sure... it a simple feature is in an isolated workspace, and it won't have the map cause it's a simple one, encoding is still going to produce a valid document because simple types have no extra namespace dependencies?

Thinking out loud, wouldn't it make sense to use only the known dependencies regardless of whether the workspace is isolated or not?

fernandor777: This method is only used when it is a complex feature case, so simple features namespaces handling is not touched.  The only-on-insolated-workspace condition enforce is for backward compatibility.

src/wfs/src/main/java/org/geoserver/wfs/xml/FeatureTypeSchemaBuilder.java
```diff
@@ -365,20 +369,63 @@ public final XSDSchema buildSchemaInternal(
...
+                                        final String prefix =
+                                                nsFromCatalog
+                                                        .map(NamespaceInfo::getPrefix)
+                                                        .orElse(entry.getKey());
+                                        schemaNamespacesMap.put(prefix, uri);
+                                    });
+                }
+            } else {
+                // No isolated workspace, so add secondary namespaces from the full catalog
+                for (NamespaceInfo nameSpaceinfo : catalog.getNamespaces()) {
```
nmco: This is wrong, the ``DECLARED_NAMESPACES_MAP`` shoudl already contain all necessary namespaces no? So why adding all the namespaces?

IMHO the behavior now should be, is a simple feature? business as usual. Is a complex feature? then we get the necessary namespaces from the complex type and we add them, makes sense?

nmco:
@fernandor777 this is not a isolated workspaces problem only, this is a problem for all WFS requests that need to output namespaces and target a virtual service.

What happens is that until now it was impossible to know beforehand which namespaces were need to encode a complex feature, simple features only need a namespace and we knew it, hence all the available namespaces (workspaces namespaces) available in GeoServer were used. This was an ugly solution but it worked.

Now, thanks to this improvement in GeoTools:
https://github.com/geotools/geotools/pull/2674

... when the GML encoder needs to encode a complex feature it can obtain the necessary namespaces from the feature type user data, including de-facto only the ones that are needed, and this shoudl be the behavior, for non local workspaces requests, for non isolated workspaces requests and for global requests.

src/wfs/src/main/java/org/geoserver/wfs/xml/FeatureTypeSchemaBuilder.java
fernandor777: Indeed, makes sense.  Applied a change for now include all complex features use cases.

fernandor777:
@nmco Indeed, makes sense. Applied a change for now include all complex features use cases.

fernandor777:
Appveyor Security UI Core Module failure seems unrelated.

PR #3888, opened 2019-11-08
[GEOS-9395] restconfig module for WMTS service configuration

groldan:
Core module (`restconfig-wmts`) to enable configuring basic WMTS service metadata (global and workspace-local) through the REST API just like it's done for WMS, WFS, and WCS.

## Checklist

For all pull requests:

### DIFF
--- a/doc/en/api/1.0.0/owsservices.yaml
+++ b/doc/en/api/1.0.0/owsservices.yaml
@@ -980,6 +980,233 @@ paths:
         200:
           description: OK
 
+  /services/wmts/settings:
+    get:
+      operationId: getWMTSSettings
+      description: Retrieves Web Map Tile Service settings globally for the server.
+      produces:
+        - application/xml
+        - application/json
+        - text/html
+      responses:
+        200:
+          description: OK
+          schema:
+            $ref: "#/definitions/WMTSInfo"
+          examples:
+            application/xml: |
+              <wmts>
+                <enabled>true</enabled>
+                <name>WMTS</name>
+                <title>Web Map Tile Service</title>
+                <maintainer>http://geoserver.org/comm</maintainer>
+                <abstrct>This server implements the WMTS specification 1.0</abstrct>
+                <accessConstraints>NONE</accessConstraints>
+                <fees>NONE</fees>
+                <versions>
+                  <org.geotools.util.Version>
+                    <version>1.0.0</version>
+                  </org.geotools.util.Version>
+                </versions>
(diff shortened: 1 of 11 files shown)


doc/en/user/source/rest/api/services.rst
```diff
@@ -215,4 +215,74 @@ Controls Web Map Service settings for a given workspace.
      -
      -
 
+``/services/wmts/settings[.<format>]``
```
jodygarnett: I am not sure this page needs to be maintained, idea was to replace it with the Open API documentation. We just have not removed these as we are not 100% sure we captured everything.

doc/en/api/1.0.0/owsservices.yaml
```diff
@@ -980,6 +980,233 @@ paths:
         200:
           description: OK
 
+  /services/wmts/settings:
```
jodygarnett: Q: Long term should this file be broken up into multiple individual files?

src/restconfig-wmts/src/test/java/org/geoserver/rest/service/LocalWMTSSettingsControllerTest.java
```diff
@@ -0,0 +1,222 @@
...
+                        .getStatus());
+        boolean thrown = false;
+        try {
+            JSON json =
+                    getAsJSON(
+                            RestBaseController.ROOT_PATH
+                                    + "/services/wmts/workspaces/sf/settings.json");
+        } catch (JSONException e) {
+            thrown = true;
+        }
```
jodygarnett: For a shorter test consider:
```
@Test(expected = JSONException.class)
```

groldan: right, mind, it's a copy of the other module... maybe we could do some general cleanup but not as part of this pr? do you see this as a blocker?

doc/en/user/source/rest/api/services.rst
groldan: well in any case it's now just as useless as the other ones

doc/en/api/1.0.0/owsservices.yaml
groldan: Longer term we should provide OpenAPI 3 yaml files, and use tags to split the api into sub-apis, so it wouldn't matter on which files each path is. But would still probably keep these ones on the same file

PR #3894, opened 2019-11-12
[GEOS-9399] SQL View with NULL SRID issue

imranrajjad:
-Replace all bound calculations with re-projecting native bbox to traget srs
-only check for native SRS if resource has more than one SRS else stay
legacy

https://osgeo-org.atlassian.net/projects/GEOS/issues/GEOS-9399

## Checklist

For all pull requests:


imranrajjad:
@nmco refactoring changes applied and native SRS selection code has been completely isolated from legacy code

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
```diff
@@ -105,13 +93,14 @@ public BasicResourceConfig(String id, IModel<ResourceInfo> model) {
         refForm.add(latLonPanel);
         refForm.add(computeLatLonBoundsLink(refForm, nativeBBox, latLonPanel));
 
-        // check other native srs for Cascaded WFS or WMS resource
-        List<String> otherSRS = getOtherSRS(model.getObject());
-
-        // native srs , declared srs, and srs handling dropdown
+        // ALERT - CORE CHANGE
```
nmco: ?

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
```diff
@@ -105,13 +93,14 @@ public BasicResourceConfig(String id, IModel<ResourceInfo> model) {
...
+        // for resources coming from WMS and WFS Datastore
+        // then check if it has more than one SRS, then add FIND SRS button in Native SRS
+        // which will allow user to set other SRS as Native SRS
+        List<String> otherSRS = getOtherSRSFromWFSOrWMS(model.getObject());
         CRSPanel nativeCRS;
-
         if (otherSRS.isEmpty()) {
+            // legacy behavior
```
nmco: Why calling this legacy? the way I see it there is the behavior for normal layers, and the behavior for cascaded WMS and WFS stores no?

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
```diff
@@ -105,13 +93,14 @@ public BasicResourceConfig(String id, IModel<ResourceInfo> model) {
...
-
-        // native srs , declared srs, and srs handling dropdown
+        // ALERT - CORE CHANGE
+        // for resources coming from WMS and WFS Datastore
+        // then check if it has more than one SRS, then add FIND SRS button in Native SRS
+        // which will allow user to set other SRS as Native SRS
+        List<String> otherSRS = getOtherSRSFromWFSOrWMS(model.getObject());
         CRSPanel nativeCRS;
-
         if (otherSRS.isEmpty()) {
```
nmco: IMHO this should be a check validating if this is a remote cascaded WFS or WMS store, the inside if that's the case you may try to retrieve multiple SRS, makes sense?

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
```diff
@@ -457,4 +326,130 @@ public void validate(final Form<?> form) {
...
+        try {
+            if (resourceInfo instanceof FeatureTypeInfo) {
+                FeatureTypeInfo ft = (FeatureTypeInfo) resourceInfo;
+                return "EPSG:"
+                        + CRS.lookupEpsgCode(
+                                ft.getFeatureSource(null, null).getInfo().getCRS(), false);
+            } else {
+                CatalogBuilder cb = new CatalogBuilder(GeoServerApplication.get().getCatalog());
+                cb.setStore(resourceInfo.getStore());
+                ReferencedEnvelope bounds = cb.getNativeBounds(resourceInfo);
```
nmco: What an weird way of get the native bounds, and then using it to retrieve the CRS, is there not a more straightforward way?

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
imranrajjad: will remove

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
```diff
@@ -457,4 +326,130 @@ public void validate(final Form<?> form) {
...
+    private List<String> getOtherSRSFromWFSOrWMS(ResourceInfo resourceInfo) {
+        // first check if its WFS-NG
+        List<String> otherSRS = DataStoreUtils.getOtherSRSFromWfsNg(resourceInfo);
+        // second check if its wms
+        if (otherSRS.isEmpty()) otherSRS = DataStoreUtils.getOtherSRSFromWMSStore(resourceInfo);
+
+        return otherSRS;
+    }
+
+    private CRSPanel getSelectableNativeCRSPanel(
```
nmco: I would add a comment describing what's this about, new panel for cascaded stores?

nmco:
The cleans a bit the ``BasicResourceConfig`` by isolating the behavior of cascaded stuff. That say, looks like a test is missing.

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
imranrajjad: the method `getOtherSRSFromWFSOrWMS `only works on WFS and WMS Stores, for any other type of resource the list will come **empty**. I will add this to the comments

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
imranrajjad: will change

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
imranrajjad: will add

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
imranrajjad: that is a good point
the function `getActualNativeSRSCode` purpose is to find out if currently selected native SRS is the actual native SRS or user has modified it? That is why I used the CatalogBuilder to get ReferencedEnvelope and then use its Native SRS, I would be glad to modify this, if their exists a simpler way.

imranrajjad:
@nmco please see the test https://github.com/geoserver/geoserver/blob/498aefcacd0ac54c8b160020caed0f50fff8d985/src/web/core/src/test/java/org/geoserver/web/data/resource/ResourceConfigurationPageTest.java#L319

It simulates Native SRS selection from actual to different one.

nmco:
Yes but since the test didn't fail i either broken or not testing the NULL SRID situation no? We need a test that reproduces the situation the PR aims to fix, which means that before your fix it shoudl fail and after your fix it should pass. I don't think we need an online test for this, just simulate the situation where the native SRS is NULL.

imranrajjad:
@nmco added unit test that asserts that BasicResourceConfig does not complain about feature without SRID.

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
```diff
@@ -457,4 +327,133 @@ public void validate(final Form<?> form) {
             }
         }
     }
+
+    public boolean addOtherSRS(ResourceInfo resourceInfo) {
```
aaime: I can see this is used to perform checks in other parts of the code... What if I configure a cascaded layer via REST, will this metadata info entry be populated too?

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
```diff
@@ -457,4 +327,133 @@ public void validate(final Form<?> form) {
...
+    }
+
+    private String getActualNativeSRSCode(ResourceInfo resourceInfo) {
+
+        try {
+            if (resourceInfo instanceof FeatureTypeInfo) {
+                FeatureTypeInfo ft = (FeatureTypeInfo) resourceInfo;
+                return "EPSG:"
+                        + CRS.lookupEpsgCode(
+                                ft.getFeatureSource(null, null).getInfo().getCRS(), false);
```
aaime: Bizzarre, no guarantee info is properly implemented, it's an API that sees little usage. Better to get the feature type and get the CRS from there instead.

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
aaime: There is nothing straightforward, each store has its own native fast way. Coverage readers expose the native SRS as a dedicated method in the reader class, cascaded sources have a different approach. One option would be to add a new method in the catalog builder that does the native SRS computation without computing the BBOX, mimicking what getNativeBounds does, but without triggering a full bounds calculation, which as we know, can be really expensive. Then you can do without the if here.

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
```diff
@@ -89,11 +80,8 @@ public BasicResourceConfig(String id, IModel<ResourceInfo> model) {
         final EnvelopePanel nativeBBox = new EnvelopePanel("nativeBoundingBox", nativeBBoxModel);
         nativeBBox.setOutputMarkupId(true);
         refForm.add(nativeBBox);
-        boolean showComputeNativeBoundsLink =
-                model.getObject()
-                        .getSRS()
-                        .equalsIgnoreCase(getActualNativeSRSCode(model.getObject()));
-        refForm.add(computeNativeBoundsLink(refForm, nativeBBox, showComputeNativeBoundsLink));
+        AjaxSubmitLink nativeBoundsLink = computeNativeBoundsLink(refForm, nativeBBox);
```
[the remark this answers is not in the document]
aaime: Maybe add a test for this, if it's not there already.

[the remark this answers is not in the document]
imranrajjad: https://github.com/geoserver/geoserver/blob/498aefcacd0ac54c8b160020caed0f50fff8d985/src/web/core/src/test/java/org/geoserver/web/data/resource/ResourceConfigurationPageTest.java#L319

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
imranrajjad: I will verify that

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
imranrajjad: @aaime understood will do

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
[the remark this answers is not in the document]
imranrajjad: I will update an existing test to verify Find link is invisible

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
imranrajjad: @aaime to configure via REST, this will not be populated if its not provided inside XML, What should be best behavior if the XML tags or not found? shall the server attempt to populate them on its own if not found? because this is not just OTHER_SRS metadata map entry, its also other lists e.g available, selected, preferred image formats and remote styles. Shalll I break this into another PR for simplicity and only dead with Native CRS identification in this PR?

aaime: All parameters that need to be present for the server to work properly should be in the main information set, not in the metadata map (it was a mistake to put them there if there is no automatic fallback), as the metadata map is for "optional extras". If not set, the server should be able to do something sane on the fly, or auto-configure the values if missing. Whether this belongs in a separate PR or not, depends on the amount of rework required.

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
aaime: Yes

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
imranrajjad: @aaime indeed the amount of re-work is indeed enough to qualify for separate PR.

Secondly about the metadata map is used because the functionality works across to two different types of resources (WMS and WFS-NG Feature type), hence the metadata map was used. Even if WMSLayerInfo gets this as a direct value, what will be done for WFS-NG? because its FeatureTypeInfo?

imranrajjad:
@aaime PR is ready for merge,

for REST interface related, I will create a new PR with proper context and requirements ASAP.

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
aaime: You do two separate code paths, like for getting the native CRS.

PR #3917, opened 2019-11-25
Revert "[GEOS-9385] GWC initialization thrashes CPU and filesystem at startup "

aaime:
Reverts geoserver/geoserver#3866 as it breaks startup with the release data directory


aaime:
Never mind, local issue, even if it's a real problem (a miconfig should not prevent startup) it was not due to #3866

groldan:
> Never mind, local issue, even if it's a real problem (a miconfig should not prevent startup) it was not due to #3866

Oh I think I've came across the same issue, had a couple files under `gwc-layers` whose `LayerInfo`'s didn't exist anymore (not due to a bug but to manual data directory mangling). Thought of creating a separate issue (given it also fails without this patch) but looks like I forgot. There you go: https://osgeo-org.atlassian.net/browse/GEOS-9411

PR #3925, opened 2019-12-06
Run mvn jetty:run in test scope

bjornharrtell:
This allows use of the cross-origin filter in web.xml. Without it you will get the error `java.lang.ClassNotFoundException: org.eclipse.jetty.servlets.CrossOriginFilter`.

### DIFF
--- a/src/web/app/pom.xml
+++ b/src/web/app/pom.xml
@@ -327,6 +327,7 @@
           -->
           <!-- Make sure it's not looking for annotations everywhere -->
           <contextXml>${project.build.testOutputDirectory}/jetty-context.xml</contextXml>
+          <useTestScope>true</useTestScope>
         </configuration>
       </plugin>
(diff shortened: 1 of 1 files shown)


src/web/app/pom.xml
```diff
@@ -327,6 +327,7 @@
           -->
           <!-- Make sure it's not looking for annotations everywhere -->
           <contextXml>${project.build.testOutputDirectory}/jetty-context.xml</contextXml>
+          <useTestScope>true</useTestScope>
```
aaime: I never use ``jetty:run``, anyone else does? Otherwise I'll trust you on this one :-D
Personally I always start GeoServer from the IDE, using Start.java in the web/app module.

bjornharrtell: I have IDE issues. :)

PR #3926, opened 2019-12-10
JSON-LD callback compliant with OGCAPI and some fixes

taba90:

## Checklist

For all pull requests:

- [X] Commits changing the UI, existing user workflows, or adding new functionality, need to include documentation updates


src/community/json-ld/src/main/java/org/geoserver/jsonld/JsonLdGenerator.java
```diff
@@ -117,7 +117,7 @@ public void writeValueNode(String entryName, JsonNode valueNode) throws IOExcept
      * @throws IOException
      */
     public void writeResult(Object result) throws IOException {
-        if (result instanceof String || result instanceof Number) {
+        if (result instanceof String || result instanceof Number || result instanceof Boolean) {
```
nmco: Bellow we have:
```
} else {
    writeNull();
}
```
I would recommend:
```
} else if (result == null) {
    writeNull();
} else {
    throw new RuntimeException("We can't handle the provided object o type '" + result.getClass() + "'."):   
}
```
... this will make the method more resilient, we write NULL when the object is indeed NULL, if we we don't handle the object then we throw an exception.

src/community/json-ld/src/main/java/org/geoserver/jsonld/builders/AbstractJsonBuilder.java
```diff
@@ -12,17 +12,14 @@
...
     public AbstractJsonBuilder(String key) {
-        this.isFeaturesField = key != null && key.equalsIgnoreCase("features");
         this.key = key;
     }
 
     public AbstractJsonBuilder() {}
 
     protected void writeKey(JsonLdGenerator writer) throws IOException {
-        if (key != null && !key.equals("") && !isFeaturesField) writer.writeFieldName(key);
+        if (key != null && !key.equals("")) writer.writeFieldName(key);
```
nmco: I would suggest:
```
/**
     * Will write the key if it is neither empty or NULL.
     */
    protected void writeKey(JsonLdGenerator writer) throws IOException {
        if (key != null && !key.trim().isEmpty()) {
            writer.writeFieldName(key);
        } else {
            // should we throw an exception at least log something?
        }
    }
```

If the key is NULL or empty, is a normal situation? should we not throw an exception or at least log something?

src/community/json-ld/src/main/java/org/geoserver/jsonld/builders/impl/DynamicValueBuilder.java
```diff
@@ -141,4 +144,19 @@ private String workXpathFunction(String expression) {
         }
         return expression;
     }
+
+    private boolean canWriteValue(Object result) {
```
nmco: I would add a comment describing briefly what the method does and what the invoker should be aware of.

nmco: Suggestion:
```
/**
     * A value can only be wrote if it is non NULL and not an empty list. This method supports
     * complex features attributes, this method will be invoked recursively on the attribute value.
     */
    private boolean canWriteValue(Object result) {
        if (result instanceof Attribute) {
            // this is an attribute from a complex feature let's test its value
            return canWriteValue(((Attribute) result).getValue());
        } else if (result instanceof List) {
            // the value is a list, let's check it is not empty
            return ((List) result).size() != 0;
        }
        // let's check we don't have a NULL value
        return result != null;
    }
```

src/community/json-ld/src/main/java/org/geoserver/jsonld/JsonLdGenerator.java
taba90: fixed

src/community/json-ld/src/main/java/org/geoserver/jsonld/builders/impl/DynamicValueBuilder.java
taba90: fixed thanks

src/community/json-ld/src/main/java/org/geoserver/jsonld/builders/AbstractJsonBuilder.java
taba90: fixed

src/community/json-ld/src/main/java/org/geoserver/jsonld/request/JsonLdTemplateCallback.java
```diff
@@ -40,33 +41,28 @@ public JsonLdTemplateCallback(GeoServer gs, JsonLdConfiguration configuration) {
...
-        if (getFeature != null) {
-            Map rawKvp = request.getRawKvp();
-            boolean jump =
-                    request.getHttpRequest().getMethod().equalsIgnoreCase("GET")
-                            && (rawKvp.get("CQL_FILTER") == null
-                                    || !rawKvp.get("CQL_FILTER").toString().contains("."));
-            if (!jump) {
-                for (int i = 0; i < getFeature.getQueries().size(); i++) {
-                    Query q = getFeature.getQueries().get(i);
+        if (!request.getHttpRequest().getServletPath().equals("/ogc")
```
aaime: Servlet path tests are fragile:

- The service could be a workspace specific one, not 100% sure to do the various wrappers, but at this point you might see "workspace/ogc" as the servlet path
- The OGC services might end up being moved to another position before upgrade (not sure)
- Other plugins in the future might be using json-ld (e.g., dunno, GetFeatureInfo, a WPS process) but not based on this plugin

It seems safer to me, to use ``"WFS".equalsIgnoreCase(request.getService())`` to test for WFS, along with the output format type.

src/community/json-ld/src/main/java/org/geoserver/jsonld/request/JsonLdTemplateCallBackOGC.java
```diff
@@ -0,0 +1,90 @@
...
+    private JsonLdConfiguration configuration;
+
+    public JsonLdTemplateCallBackOGC(GeoServer gs, JsonLdConfiguration configuration) {
+        this.catalog = gs.getCatalog();
+        this.configuration = configuration;
+    }
+
+    @Override
+    public Operation operationDispatched(Request request, Operation operation) {
+        boolean isJsonLd =
```
aaime: Weak test, content negotiation in OGC API services can also be done via "accept" headers.

src/community/json-ld/src/main/java/org/geoserver/jsonld/request/JsonLdTemplateCallBackOGC.java
```diff
@@ -0,0 +1,90 @@
...
+                request.getKvp().get("f") != null
+                                && request.getKvp().get("f").equals(JSONLDGetFeatureResponse.MIME)
+                        ? true
+                        : false;
+        if (request.getHttpRequest().getServletPath().equals("/ogc") && isJsonLd) {
```
aaime: Better to check if the service is "Features" instead, there are multiple services at "/ogc" and one of the others could legitimately use json-ld too, but not based on this plugin (theoretically two plugins could be doing json-ld for features too, but that would be a straight conflict, not much we can do about it).

src/community/json-ld/src/main/java/org/geoserver/jsonld/request/JsonLdTemplateCallBackOGC.java
```diff
@@ -0,0 +1,90 @@
...
+            try {
+                String strFilter = (String) request.getKvp().get("FILTER");
```
aaime: The current spec uses filter and filter-lang, check if the lang is cql too.

src/community/json-ld/src/main/java/org/geoserver/jsonld/request/JsonLdTemplateCallBackOGC.java
```diff
@@ -0,0 +1,90 @@
...
+                    RootBuilder root = configuration.getTemplate(typeInfo);
+                    if (root != null) {
+                        JsonLdPathVisitor visitor =
+                                new JsonLdPathVisitor(typeInfo.getFeatureType());
+                        /* Todo find a better way to replace json-ld path with corresponding template attribute*/
+                        // Get filter from string in order to make it accept the visitor
+                        Filter f = (Filter) XCQL.toFilter(strFilter).accept(visitor, root);
+                        // Taking back a string from Function cause
+                        // OGC API get a string cql filter from query string
+                        String newFilter = ECQL.toCQL(f).replaceAll("\"", "").replaceAll("/", ".");
```
aaime: Hmm... unsure about these replacement, some names have to be quoted, say for example an attribute is called "between", just like one of the CQL keywords..

src/community/json-ld/src/main/java/org/geoserver/jsonld/request/JsonLdTemplateCallBackOGC.java
```diff
@@ -0,0 +1,90 @@
...
+                        newFilter = ExpressionsUtils.workXpathAttributeSyntax(newFilter);
```
aaime: Again a method called "work"?

src/community/json-ld/src/main/java/org/geoserver/jsonld/request/JsonLdTemplateCallback.java
taba90: fixed

src/community/json-ld/src/main/java/org/geoserver/jsonld/request/JsonLdTemplateCallBackOGC.java
taba90: fixed

src/community/json-ld/src/main/java/org/geoserver/jsonld/request/JsonLdTemplateCallBackOGC.java
taba90: fixed

src/community/json-ld/src/main/java/org/geoserver/jsonld/request/JsonLdTemplateCallBackOGC.java
taba90: fixed

src/community/json-ld/src/main/java/org/geoserver/jsonld/request/JsonLdTemplateCallBackOGC.java
taba90: now it's checking if after "\"" there is a space char assuming thus that an eventual xpath ended. Do you think this would be enough?

src/community/json-ld/src/main/java/org/geoserver/jsonld/request/JsonLdTemplateCallBackOGC.java
taba90: fixed

taba90:
Thanks for your reviews @nmco and @aaime , your feedbacks have been applied

nmco:
Looks good to me, what do you think @Andrea?

src/community/json-ld/src/main/java/org/geoserver/jsonld/request/JsonLdTemplateCallBackOGC.java
```diff
@@ -0,0 +1,99 @@
...
+                    "Unknown collection " + collectionId,
+                    ServiceException.INVALID_PARAMETER_VALUE,
+                    "collectionId");
+        }
+        return featureType;
+    }
+
+    private boolean isJsonLdMIME(Request request) {
+        String accept = request.getHttpRequest().getHeader(HttpHeaders.ACCEPT);
+        if (accept != null && accept.contains(JSONLDGetFeatureResponse.MIME)) return true;
```
aaime: Upside down, "f" override accepts, so it should be tested first.

src/community/json-ld/src/main/java/org/geoserver/jsonld/request/JsonLdTemplateCallBackOGC.java
```diff
@@ -0,0 +1,99 @@
...
+        this.configuration = configuration;
+    }
+
+    @Override
+    public Operation operationDispatched(Request request, Operation operation) {
+        if ("FEATURES".equalsIgnoreCase(request.getService()) && isJsonLdMIME(request)) {
+            try {
+                String filter = (String) request.getKvp().get("FILTER");
+                replaceJsonLdPathWithFilter(filter, operation);
+                String filterLang = (String) request.getKvp().get("FILTER-LANG");
```
aaime: This is not how it works. A OGC request uses something like "FILTER=<cqlFilterHere>&FILTER-LANG=CQL-TEXT", but it could also support something like "FILTER=xmlFilterHere&FILTER-LANG=OGC-FILTER-1.0". You should pick the contents of "filter" but only if filter-lang equals to "cql-text" (make it case insensitive just to be sure).

taba90: fixed

src/community/json-ld/src/main/java/org/geoserver/jsonld/request/JsonLdTemplateCallBackOGC.java
taba90: fixed

aaime:
Looks good, thanks, merging.

taba90:
JIRA ticket https://osgeo-org.atlassian.net/browse/GEOS-9455

PR #3932, opened 2019-12-13
[GEOS-9431] Integrated GWC does not kick in when making request not qualified by workspace

imranrajjad:
Integrated WMS requests not qualified for any workspace get ignored by Integrated GWC, since it expects layers to be always prefixed with workspace. The bug has been fixed and GWC will fall back to catalog if it is unable to find a layer in TileLayerDispatcher on first attempt.

JIRA:https://osgeo-org.atlassian.net/browse/GEOS-9431

## Checklist

For all pull requests:

### DIFF
--- a/src/gwc/src/main/java/org/geoserver/gwc/GWC.java
+++ b/src/gwc/src/main/java/org/geoserver/gwc/GWC.java
@@ -750,7 +750,7 @@ public class GWC implements DisposableBean, InitializingBean, ApplicationContext
     public final ConveyorTile dispatch(
             final GetMapRequest request, StringBuilder requestMistmatchTarget) {
 
-        String layerName = request.getRawKvp().get("LAYERS");
+        final String layerName = request.getRawKvp().get("LAYERS");
         /*
          * This is a quick way of checking if the request was for a single layer. We can't really
          * use request.getLayers() because in the event that a layerGroup was requested, the request
@@ -761,19 +761,18 @@ public class GWC implements DisposableBean, InitializingBean, ApplicationContext
             return null;
         }
 
-        if (!tld.layerExists(layerName)) {
-            // GEOS-9431-check in catalog also if not found TileLayerDispatcher
-            LayerInfo layerInfo = catalog.getLayerByName(layerName);
-            if (layerInfo != null) layerName = layerInfo.prefixedName();
-            else {
-                requestMistmatchTarget.append("not a tile layer");
-                return null;
-            }
+        // GEOS-9431 acquire prefixed name if not prefixed already
+        final String getPrefixedName =
+                (!layerName.contains(":")) ? getPrefixedName(layerName) : layerName;
+
+        if (!tld.layerExists(getPrefixedName)) {
+            requestMistmatchTarget.append("not a tile layer");
+            return null;
         }
 
         final TileLayer tileLayer;
         try {
-            tileLayer = this.tld.getTileLayer(layerName);
+            tileLayer = this.tld.getTileLayer(getPrefixedName);
         } catch (GeoWebCacheException e) {
             throw new RuntimeException(e);
         }
@@ -818,6 +817,14 @@ public class GWC implements DisposableBean, InitializingBean, ApplicationContext
             log.log(Level.INFO, "Error dispatching tile request to GeoServer", e);
         }
         return tileResp;
+    }
+
+    private String getPrefixedName(String layerName) {
+        PublishedInfo info = catalog.getLayerByName(layerName);
+        if (info == null) info = catalog.getLayerGroupByName(layerName);
+        if (info != null) return info.prefixedName();
+        if (log.isLoggable(Level.INFO)) log.info("Unable to find a prefix for : " + layerName);
+        return layerName;
     }
 
     ConveyorTile prepareRequest(
--- a/src/gwc/src/main/java/org/geoserver/gwc/GWC.java
+++ b/src/gwc/src/main/java/org/geoserver/gwc/GWC.java
@@ -750,7 +750,7 @@ public class GWC implements DisposableBean, InitializingBean, ApplicationContext
     public final ConveyorTile dispatch(
             final GetMapRequest request, StringBuilder requestMistmatchTarget) {
 
-        final String layerName = request.getRawKvp().get("LAYERS");
+        String layerName = request.getRawKvp().get("LAYERS");
         /*
          * This is a quick way of checking if the request was for a single layer. We can't really
          * use request.getLayers() because in the event that a layerGroup was requested, the request
@@ -762,8 +762,13 @@ public class GWC implements DisposableBean, InitializingBean, ApplicationContext
         }
 
         if (!tld.layerExists(layerName)) {
-            requestMistmatchTarget.append("not a tile layer");
-            return null;
+            // GEOS-9431-check in catalog also if not found TileLayerDispatcher
+            LayerInfo layerInfo = catalog.getLayerByName(layerName);
+            if (layerInfo != null) layerName = layerInfo.prefixedName();
+            else {
+                requestMistmatchTarget.append("not a tile layer");
+                return null;
+            }
         }
 
         final TileLayer tileLayer;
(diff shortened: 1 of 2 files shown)


src/gwc/src/main/java/org/geoserver/gwc/GWC.java
```diff
@@ -762,8 +762,13 @@ public final ConveyorTile dispatch(
         }
 
         if (!tld.layerExists(layerName)) {
-            requestMistmatchTarget.append("not a tile layer");
-            return null;
+            // GEOS-9431-check in catalog also if not found TileLayerDispatcher
+            LayerInfo layerInfo = catalog.getLayerByName(layerName);
```
aaime: This should indeed be using the LocalWorkspaceCatalog

src/gwc/src/test/java/org/geoserver/gwc/GWCIntegrationTest.java
```diff
@@ -2086,4 +2087,30 @@ private void createLayerGroup(String layerGroupName, QName... layersNames) throw
...
+     * Test verifying that GWC integration will work when layer belongs to default workspace and
+     * layer name is passed without workspace being part of layer name or OWS request itself
+     */
+    @Test
+    public void testDirectDefaultWorkspaceWMSIntegration() throws Exception {
+        final GWC gwc = GWC.get();
+        gwc.getConfig().setDirectWMSIntegrationEnabled(true);
+        final String layerName = BASIC_POLYGONS.getLocalPart();
+
+        String request =
```
aaime: This is not testing what the ticket reports, which is, a request on a local workspace service, aka virtual services, e.g.:

```cite/wms?service=WMS&request=GetMap&version=1.1.1&format=image/png&...```

See also: https://docs.geoserver.org/latest/en/user/configuration/virtual-services.html

File: src/gwc/src/test/java/org/geoserver/gwc/GWCIntegrationTest.java [hunk 3]
```diff
@@ -159,6 +159,7 @@ public void cleanupDispatcherRequest() {
 
     private void prepareDataDirectory(SystemTestData testData) throws Exception {
         Catalog catalog = getCatalog();
+        catalog.setDefaultWorkspace(getCatalog().getWorkspaceByName(BASIC_POLYGONS.getPrefix()));
```
aaime: No need to make this default in a test actually using per workspace services.

src/gwc/src/main/java/org/geoserver/gwc/GWC.java
```diff
@@ -762,8 +762,13 @@ public final ConveyorTile dispatch(
...
+            if (layerInfo != null) layerName = layerInfo.prefixedName();
```
nmco: I'm assuming that the local workspace catalog wrapper is used here? In practice this means that:
- if not a virtual service request, the default workspace will be used
- if in the context of a virtual service request the local workspace will be used

src/gwc/src/main/java/org/geoserver/gwc/GWC.java
nmco: This needs to vouch for layer groups.

src/gwc/src/main/java/org/geoserver/gwc/GWC.java
nmco: This breaks the workflow of GWC, the check should be done by ``!tld.layerExists(prefixedName)``.

src/gwc/src/test/java/org/geoserver/gwc/GWCIntegrationTest.java
```diff
@@ -2086,4 +2087,30 @@ private void createLayerGroup(String layerGroupName, QName... layersNames) throw
         catalogBuilder.calculateLayerGroupBounds(layerGroup);
         getCatalog().add(layerGroup);
     }
+
+    /**
+     * Test verifying that GWC integration will work when layer belongs to default workspace and
+     * layer name is passed without workspace being part of layer name or OWS request itself
```
nmco: Missing dot.

src/gwc/src/test/java/org/geoserver/gwc/GWCIntegrationTest.java
```diff
@@ -2086,4 +2087,30 @@ private void createLayerGroup(String layerGroupName, QName... layersNames) throw
...
+     */
+    @Test
+    public void testDirectDefaultWorkspaceWMSIntegration() throws Exception {
```
nmco: I would also add a test for when a virtual service is used and the provided layer name is not qualified, maybe this one already exists. I would also add the same tests for layer groups.

nmco:
Sorry @aaime, looks liek we reviewed at the same time :yum:.

src/gwc/src/main/java/org/geoserver/gwc/GWC.java
imranrajjad: understood,will do, make a separate method to looks for the name as layer or layergroup

src/gwc/src/test/java/org/geoserver/gwc/GWCIntegrationTest.java
imranrajjad: will add

src/gwc/src/main/java/org/geoserver/gwc/GWC.java
imranrajjad: understood, will do

src/gwc/src/main/java/org/geoserver/gwc/GWC.java
imranrajjad: will LocalWorkspaceCatalog

src/gwc/src/test/java/org/geoserver/gwc/GWCIntegrationTest.java
imranrajjad: I will update the test to cover this scenario

src/gwc/src/test/java/org/geoserver/gwc/GWCIntegrationTest.java
imranrajjad: understood, will do

src/gwc/src/main/java/org/geoserver/gwc/GWC.java
imranrajjad: yes  that's correct

src/gwc/src/main/java/org/geoserver/gwc/GWC.java
aaime: As in, the catalog reference you get from Spring should already contain LocalWorkspaceCatalog as one of the wrappers, don't think there is a need to be explicit.

imranrajjad:
@nmco @aaime thank you very much for your review and feedback. I have addressed the requested changes as suggested. Please let me know if more changes are required

src/gwc/src/main/java/org/geoserver/gwc/GWC.java
```diff
@@ -820,6 +819,14 @@ public final ConveyorTile dispatch(
         return tileResp;
     }
 
+    private String getPrefixedName(String layerName) {
+        PublishedInfo info = catalog.getLayerByName(layerName);
+        if (info == null) info = catalog.getLayerGroupByName(layerName);
+        if (info != null) return info.prefixedName();
+        if (log.isLoggable(Level.SEVERE)) log.severe("Unable to find and prefix " + layerName);
```
aaime: Severe? That would appear even in production logging, it's too much, consider this would be a client error (if the client uses the names reported in the capabilities documents, there should be no issue). Level should be INFO at most, if not FINE.

Also the mssage, "Unable to find and prefix " --> "Unable to fid a prefix for: "

imranrajjad: understood, will change

imranrajjad: this is changed

aaime:
Looks good to me, waiting for @nmco to also have a look

nmco:
Looks good to me too, merging! @imranrajjad  since this a bug fix it shoudl be backported to 2.16.x and 2.15.x.

PR #3952, opened 2020-01-03
GEOS-9285 Escape roles in backend before using in frontend to fix XSS

1Jesper1:
GEOS-9285 Escape roles in backend before using in frontend to fix XSS
It would be better to prevent the XSS in the frontend, if someone has an idea to do this, feel free to change this. For now I escape the roles in the backend.

## Checklist

For all pull requests:

### DIFF
--- a/src/extension/wps/web-wps/src/main/java/org/geoserver/wps/web/ProcessSelectionPage.java
+++ b/src/extension/wps/web-wps/src/main/java/org/geoserver/wps/web/ProcessSelectionPage.java
@@ -35,7 +35,6 @@ import org.geoserver.wps.process.GeoServerProcessors;
 import org.geoserver.wps.web.FilteredProcessesProvider.FilteredProcess;
 import org.geotools.process.ProcessFactory;
 import org.opengis.feature.type.Name;
-import org.springframework.web.util.HtmlUtils;
 
 /**
  * A page listing all WPS process for specific group, allowing enable/disable single process and
@@ -70,7 +69,7 @@ public class ProcessSelectionPage extends AbstractSecurityPage {
         GeoServerRoleService roleService = getSecurityManager().getActiveRoleService();
         try {
             for (GeoServerRole r : roleService.getRoles()) {
-                availableRoles.add(HtmlUtils.htmlEscape(r.getAuthority()));
+                availableRoles.add(r.getAuthority());
             }
         } catch (IOException e1) {
             LOGGER.log(Level.FINER, e1.getMessage(), e1);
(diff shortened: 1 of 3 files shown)


1Jesper1:
Why do the builds fail?

mprins:
not sure if related but: https://travis-ci.org/geoserver/geoserver/jobs/632476374#L6495-L6502

aaime:
Might be one of the random test failures, re-kicked the builds.

1Jesper1:
Thanks @aaime but the AppVeyor build still fails

1Jesper1:
@aaime Could you rekick the AppVeyor build please?

aaime:
I have checked a bit the docs, it seems that Wicket should be doing the escaping automatically for us... in any case, did you try calling setEscapeModelStrings on the TextArea field?

aaime:
Also, this is the administration UI, one needs full admin rights to get there, it's not something open to the public. And roles are manually entered by the same administrator. What's the actual XSS risk?

1Jesper1:
@aaime Yes, I tried that, but it didn't seem to work.

aaime:
Final note, the PR lacks tests, as such, it cannot be merged. I would also recommend checking that data gets properly un-escaped on the save side, confirming that it does not end up saving HTML escape codes in the security subsystem configuration.

Btw, forget about the Windows faiulres, we are, we are looking into getting a different Windows build server to see if it's more reliable. If that also fails, Windows tests will have to be dropped, we don't have anyone working on Windows anymore that would be willing to take on this (if you want to step up for it, or know someone that would, it would be great news indeed!).

1Jesper1:
@aaime XSS possibility to let another admin execute an action. I have changed the place of the html escaping to the RolesAutoCompletedBehavior file. I don't know how to create a test for this case, I need to mock the RolesAutoCompletedBehavior class and then invoke the GetChoices protected method.

aaime:
Was this tested interactively? I checked out the branch, added a role called "a&b", and here is what shows up in the UI after completion:

image

Not what was intended, I presume?

1Jesper1:
@aaime This output already happened before my changes. I found the XSS.

aaime:
Definitely not, without your change the auto-complete results in:

image

which is the expected result (if one names a role "a&b" that's the output I expect to see in the UI).

If you have an actual attack vector, could you share it with geoserver-security@lists.osgeo.org? Sorry , never received formal training regarding security and know close to nothing on client side development either, so would not know where to start looking at this myself.

aaime:
Closing due to inactivity, tried to get some traction  the geoserver-security list for a repeatable test case, in order to prepare a fix that does not break role editing.

PR #3954, opened 2020-01-07
Fix tools page (GEOS-9348)

jodygarnett:
The tools page lists all instances of ToolLinkExternalInfo, some recent work to make the Layer Preview page extendable reused this as a base class without testing the tools page.

This patch double checks that an instance has a description before including it in the tools page, this has the effect of filtering out all the CommonFormatLink instances.

## Checklist

For all pull requests:

- [x] Will merge as a single commit for ease of backporting

### DIFF
--- a/doc/en/developer/source/quickstart/checkout.txt
+++ b/doc/en/developer/source/quickstart/checkout.txt
@@ -8,8 +8,8 @@ Check out the source code from the git repository.::
 To list the available branches.::
 
   % git branch
-     2.14.x
      2.15.x
+     2.16.x
    * master
 
 Choose ``master`` for the latest development.::
@@ -18,7 +18,7 @@ Choose ``master`` for the latest development.::
 
 Or chose a stable branch for versions less likely to change often::
 
-  % git checkout 2.14.x
+  % git checkout 2.16.x
 
 In this example we will pretend that your source code is in a directory
 called ``geoserver``, but a more descriptive name is recommended.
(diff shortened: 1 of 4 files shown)


jodygarnett:
It is worth squashing this commit so backport is easier.

aaime:
Please squash following the suggested form (for non comunity modules, it's a requirement):
"Commit message must be in the form "[GEOS-XYZW] Title of the Jira ticket" (export to XML in Jira generates the message in this exact form)" 
I saw your comment on the GT PR template, please raise the issue on the devel list to change the current practice if you have trouble aligning with it, explain why it could be troublesome for others too.

doc/en/developer/source/quickstart/checkout.txt
```diff
@@ -8,8 +8,8 @@ Check out the source code from the git repository.::
 To list the available branches.::
 
   % git branch
-     2.14.x
      2.15.x
+     2.16.x
```
aaime: The PR looks good, no issues with it. This change is good too, but do we want to really refer to actual branches? It would require some documentation pass every time we cut a new stable... can we refer to "abstract" branches instead?
Anyways, just an idea, this change does not make things worse, actually it just makes the doc up to date for the next couple of months.

jodygarnett:
Waiting for tests, this should be ready to merge and backport

doc/en/developer/source/quickstart/checkout.txt
jodygarnett: I hate thinking when starting with a new codebase, so seeing real version numbers helps me :)

PR #3970, opened 2020-01-16
[backport 2.15.x] re-add select folders with precomposeunicode=true

jodygarnett:
This is a follow up to "macOS users important gitconfig changes" email thread where folders with unicode paths need:
- to be removed using core.precomposeunicode=false
- re-added using core.precomposeunicode=true

This change had occurred on master and 2.16.x but not applied to the 2.15.x branch (which remains active).

This is not a functional change and is only visible to macOS users, it is a quality of life improvement preventing the requirement to use "reset --hard" when switching to the 2.15.x branch.


jodygarnett:
Note this change, by its nature, cannot be squashed.

jodygarnett:
rebased (one cherry-pick at a time) to take advantage of change to travis build

jodygarnett:
It is indeed just renames

PR #3971, opened 2020-01-16
[GEOS-9458] Fix WFS 2.0 join queries with unqualified typenames

groldan:
Improve org.geoserver.wfs.WFSWorkspaceQualifier to resolve
typenames on the local workspace when a WFS request with
multiple elements in the 'typeNames' (not 'typeName') parameter
are specified.

checkOriginallyUnqualified() was only looking at TYPENAME (WFS 1.0),
and not TYPENAMES (2.0).

Since by the time TypeNameKvpParser resolves the requested type names,
the LocalWorkspace is not set, it may either resolve to an unqualified
QName, or one on any other workspace than the requested one (due to the
lenient resolution in TypeNameKvpParser.parseToken() calling
Catalog.getFeatureTypeByName(String), which in turns returns a feature
type in the default workspace or any other workspace if none is found).
This lead to either trying to use the wrong FeatureType or not finding
it at all.

Also improve the readability of WFSWorkspaceQualifier.qualifyRequest(
 WorkspaceInfo, PublishedInfo, Service, Request)

## Checklist

> Reviewing is a process done by project maintainers, mostly on a volunteer basis. We try to keep the overhead as small as possible and appreciate if you help us to do so by completing the following items. Feel free to ask in a comment if you have troubles with any of them.

For all pull requests:

- [x] Confirm you have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/master/CONTRIBUTING.md) 
- [x] You have sent a Contribution Licence Agreement (CLA) as necessary (not required for small changes, e.g., fixing typos in documentation)
- [x] Make sure the first PR targets the master branch, eventual backports will be managed later. This can be ignored if the PR is fixing an issue that only happens in a specific branch, but not in newer ones.

The following are required only for core and extension modules (they are welcomed, but not required, for community modules):
- [x] There is a ticket in Jira describing the issue/improvement/feature (a notable exemptions is, changes not visible to end users)
- [x] PR for bug fixes and small new features are presented as a single commit
- [x] Commit message must be in the form "[GEOS-XYZW] Title of the Jira ticket" (export to XML in Jira generates the message in this exact form)
- [x] New unit tests have been added covering the changes
- [ ] This PR passes all existing unit tests (test results will be reported by travis-ci after opening this PR)
- [ ] This PR passes the [QA checks](https://docs.geoserver.org/latest/en/developer/qa-guide/index.html) (QA checks results will be reported by travis-ci after opening this PR)
[...]

PR #3972, opened 2020-01-16
[backport-2.16.x] [GEOS-9458] Fix WFS 2.0 join queries with unqualified typenames

groldan:
Backport of #3971

Improve org.geoserver.wfs.WFSWorkspaceQualifier to resolve
typenames on the local workspace when a WFS request with
multiple elements in the 'typeNames' (not 'typeName') parameter
are specified.

checkOriginallyUnqualified() was only looking at TYPENAME (WFS 1.0),
and not TYPENAMES (2.0).

Since by the time TypeNameKvpParser resolves the requested type names,
the LocalWorkspace is not set, it may either resolve to an unqualified
QName, or one on any other workspace than the requested one (due to the
lenient resolution in TypeNameKvpParser.parseToken() calling
Catalog.getFeatureTypeByName(String), which in turns returns a feature
type in the default workspace or any other workspace if none is found).
This lead to either trying to use the wrong FeatureType or not finding
it at all.

Also improve the readability of WFSWorkspaceQualifier.qualifyRequest(
 WorkspaceInfo, PublishedInfo, Service, Request)

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/WFSWorkspaceQualifier.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/WFSWorkspaceQualifier.java
@@ -5,10 +5,13 @@
  */
 package org.geoserver.wfs;
 
-import java.util.ArrayList;
+import com.google.common.base.Strings;
+import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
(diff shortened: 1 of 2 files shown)


aaime:
2.15.x could use a backport of this fix too

groldan:
> 2.15.x could use a backport of this fix too
Got it: #4012

PR #4031, opened 2020-02-06
[GEOS-9487] GSIP 187 - New extension point in ResourcePool for retyping feature types and features

imranrajjad:
- added interface and code pointers in gs-main
- added unit tests

GSIP : https://github.com/geoserver/geoserver/wiki/GSIP-187
JIRA : https://osgeo-org.atlassian.net/browse/GEOS-9487

## Checklist

For all pull requests:

- [x] There is a ticket in Jira [GEOS-9487](https://osgeo-org.atlassian.net/browse/GEOS-9487)
- [ ] Committs changing the REST API, or any configuration object, should check it the REST API docs (Swagger YAML files and classic documentation) need to be updated.


src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -1382,15 +1389,22 @@ else if (ppolicy == ProjectionPolicy.NONE && nativeCRS != null) {
             }
 
             // return a normal
```
nmco: This comment needs to be updated, a comments explaining a bit was going on would be welcome :smile_cat:.

aaime: Agreed, and would add a note about the order of operations, why first create the geoserver feature source and then run the wrappers after it, instead of the opposite. I believe it has to be this way to match the feature type building process?
However picture this case, a wrapper is building a geometry type that was not there. GeoServer feature source applies the projection handling, so doing it before, will make the code lose it. _Maybe_ it's better to apply the callbacks before any sort of CRS handling is done, both here and computing the feature type, since the CRS handling is one of the user visible/user configurable things. Let's talk about it at least?

nmco: Good catch @aaime (and thank you for the offline explanation :wink:). So @imranrajjad, like me you may have not noticed that this code:
```java
            // return a normal
            return GeoServerFeatureLocking.create(
                    fs,
                    new GeoServerFeatureSource.Settings(
                            schema,
                            info.filter(),
                            resultCRS,
                            info.getProjectionPolicy().getCode(),
                            getTolerance(info),
                            info.getMetadata()));
```

Is actually wrapping the original feature source ``fs`` with a wrapper that will provide extra GeoServer functionalities, for example, the ability to define a layer with a CQL filter or the ability of forcing a layer for publishing different from the native one. 

Hence this extension wrapping  should happen before this is done, and we shoudl actually wrap the original feature source, the variable ``fs``.

The integration tests should also be extended to tests this situations:
- A layer whit a CQL filter associated, querying the layer using WMS or WFs should filter out the features not matching the layer associated CQL
- A layer with the declared CRS different form the native one, querying the layer iwht either WFS or WMS should return featrues expressed in the declared SRS not in the native one.

imranrajjad: @nmco its done

imranrajjad:
@aime @nmco 

 thank you for your review, I applied your suggestions, let me know if something else is needed.

nmco:
Thx for the update @imranrajjad, please see my comments.

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -1381,6 +1388,13 @@ else if (ppolicy == ProjectionPolicy.NONE && nativeCRS != null) {
                 schema = typeBuilder.buildFeatureType();
             }
 
+            // applying wrappers using
+            // implementations of org.geoserver.catalog.RetypeFeatureTypeCallback
```
nmco: Please shorten this comment, if the class is moved or renamed this will be come just noise.

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -1381,6 +1388,13 @@ else if (ppolicy == ProjectionPolicy.NONE && nativeCRS != null) {
...
+            for (RetypeFeatureTypeCallback callback :
+                    GeoServerExtensions.extensions(RetypeFeatureTypeCallback.class)) {
+                fs = (SimpleFeatureSource) callback.wrapFeatureSource(info, fs);
```
nmco: You should not assume this will always by a ``SimpleFeatureSource`, check this method returning type.

src/main/src/test/java/org/geoserver/catalog/MockRetypeFeatureTypeCallback.java
```diff
@@ -0,0 +1,71 @@
...
+import org.opengis.feature.type.AttributeDescriptor;
+import org.opengis.feature.type.FeatureType;
+import org.opengis.feature.type.PropertyDescriptor;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+
+/**
+ * This is a mock implementation
+ *
+ * @author ImranR
+ */
```
nmco: Noisy JavaDoc, better no doc that noise, you can configure your IDE to not create this templates. Since there is not final dot in your sentence I'm not even sure you finished it ...

src/main/src/test/java/org/geoserver/catalog/MockRetypeFeatureTypeCallback.java
```diff
@@ -0,0 +1,71 @@
...
+    public static final String RETYPED = "RETYPED";
+    public static final String RETYPED_GEOM_COLUMN = "GENERATED_POINT";
+
+    public static final String LONG_FIELD = "lon";
+    public static final String LAT_FIELD = "lat";
+    public static final int EPSG_CODE = 4326;
+
+    @Override
+    public FeatureType retypeFeatureType(FeatureTypeInfo featureTypeInfo, FeatureType src) {
+        // only work for the test file : test/resources/org/geoserver/catalog/longlat.properties
```
nmco: ?

src/main/src/test/java/org/geoserver/catalog/MockRetypeFeatureTypeCallback.java
```diff
@@ -0,0 +1,71 @@
...
+            builder.add(RETYPED_GEOM_COLUMN, Point.class);
+            for (PropertyDescriptor ad : src.getDescriptors()) {
+                builder.add((AttributeDescriptor) ad);
+            }
+            FeatureType newType = builder.buildFeatureType();
+            newType.getUserData().put(RETYPED, true);
+            return newType;
+
+        } catch (Exception e) {
+            e.printStackTrace();
```
nmco: I don't think you want to swallow this exception, if that's the case please log it.

src/main/src/test/java/org/geoserver/catalog/MockRetypeFeatureTypeCallback.java
```diff
@@ -0,0 +1,71 @@
...
+        }
+        return src;
+    }
+
+    @Override
+    public FeatureSource wrapFeatureSource(
+            FeatureTypeInfo featureTypeInfo, FeatureSource featureSource) {
+        // only work for the test file : test/resources/org/geoserver/catalog/longlat.properties
```
nmco: ?

src/main/src/test/java/org/geoserver/catalog/MockRetypeFeatureTypeCallback.java
```diff
@@ -0,0 +1,71 @@
...
+import org.opengis.feature.type.FeatureType;
+import org.opengis.feature.type.PropertyDescriptor;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+
+/**
+ * This is a mock implementation
+ *
+ * @author ImranR
+ */
+public class MockRetypeFeatureTypeCallback implements RetypeFeatureTypeCallback {
```
nmco: The JavaDoc of this class shoulder really explain what's going on: it will generate a geometry attribute on the fly? If that's the case it could also use a more focused named than ``MockRetypeFeatureTypeCallback``.

src/main/src/test/java/org/geoserver/catalog/RetypeFeatureTypeCallbackTest.java
```diff
@@ -0,0 +1,98 @@
...
+import org.geotools.referencing.CRS;
+import org.junit.Test;
+import org.locationtech.jts.geom.Point;
+import org.opengis.feature.Feature;
+import org.opengis.feature.type.FeatureType;
+import org.opengis.referencing.crs.CoordinateReferenceSystem;
+import org.vfny.geoserver.global.GeoServerFeatureSource;
+
+/**
+ * This test makes use of mock RetypeFeatureTypeCallback implementation provided in
```
nmco: ?

src/main/src/test/java/org/geoserver/catalog/RetypeFeatureTypeCallbackTest.java
```diff
@@ -0,0 +1,98 @@
...
+ *
+ * @author ImranR
```
nmco: ?

src/main/src/test/java/org/geoserver/catalog/RetypeFeatureTypeCallbackTest.java
```diff
@@ -0,0 +1,98 @@
...
+                    LONG_LAT_NO_GEOM_ON_THE_FLY_LAYER,
+                    MockData.DEFAULT_PREFIX);
+
+    @Override
+    protected void onSetUp(SystemTestData testData) throws Exception {
+        super.onSetUp(testData);
+        setupBasicLayer(testData);
+    }
+
+    private void setupBasicLayer(SystemTestData testData) throws IOException {
```
nmco: This is the basic layer for a specific test, the method name is too generic.

src/main/src/test/java/org/geoserver/catalog/RetypeFeatureTypeCallbackTest.java
```diff
@@ -0,0 +1,98 @@
...
+                LONG_LAT_NO_GEOM_ON_THE_FLY_QNAME,
+                props,
+                LONG_LAT_NO_GEOM_ON_THE_FLY_LAYER + ".properties",
+                getClass(),
+                getCatalog());
+    }
+
+    @Test
+    public void testResourcePoolOperations() throws Exception {
+        ResourcePool pool = ResourcePool.create(getCatalog());
```
nmco: This tests performs at least three different tests:
- that a geometry attribute will be generated on the fly, i.e. testing the extension points
- asserts that GeoServer re-projection still work, I would provide some context explained why we are explicitly tests this
- asserts that layer CQL filters are still working, same comment as for previous point

Please implemented different tests properly documented, this will make the tests easier to debug, understand and resilient.

src/main/src/test/java/org/geoserver/catalog/retype/MockRetypedFeatureCollection.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.NoSuchElementException;
+import org.geoserver.catalog.FeatureTypeInfo;
+import org.geotools.data.simple.SimpleFeatureCollection;
+import org.geotools.data.simple.SimpleFeatureIterator;
+import org.geotools.feature.collection.DecoratingSimpleFeatureCollection;
+import org.opengis.feature.simple.SimpleFeature;
+import org.opengis.feature.simple.SimpleFeatureType;
+
+/*
+ * A Mock implementation of DecoratingSimpleFeatureCollection to handle features with geometries generated on the fly
```
nmco: ?

src/main/src/test/java/org/geoserver/catalog/retype/MockRetypedFeatureConverter.java
```diff
@@ -0,0 +1,142 @@
...
+import org.opengis.feature.simple.SimpleFeature;
+import org.opengis.feature.simple.SimpleFeatureType;
+import org.opengis.feature.type.AttributeDescriptor;
+
+/*
+ * A Mock Converter Helper class to demonstrate how to
+ * Retype Feature by generating Geometries on the fly
+ * Defining Geomtry attribute on the fly
+ * Query handling which ensure that dynamic field is not delegated to Data Store
+ *
```
nmco: Looks like this is formatted like a poem, not even sure how this passed the auto formatting validation.

src/main/src/test/java/org/geoserver/catalog/retype/MockRetypedFeatureConverter.java
```diff
@@ -0,0 +1,142 @@
...
+ * **/
+public class MockRetypedFeatureConverter {
```
nmco: Why this generic ``Mock```names? This doesn't mock anything and its not generic no? This implements a concrete behavior to validate an extension point no? Why not put this class inside the test a private static inner classes?

src/main/src/test/java/org/geoserver/catalog/retype/MockRetypedFeatureConverter.java
```diff
@@ -0,0 +1,142 @@
...
+                Point point = geometryFactory.createPoint(new Coordinate(x, y));
+                point.setSRID(MockRetypeFeatureTypeCallback.EPSG_CODE);
+
+                featureBuilder.add(point);
+                for (Property prop : simpleFeature.getProperties()) {
+                    featureBuilder.set(prop.getName(), prop.getValue());
+                }
+                simpleFeature = featureBuilder.buildFeature(simpleFeature.getID());
+            } catch (Exception e) {
+
```
nmco: ?

src/main/src/test/java/org/geoserver/catalog/retype/MockRetypedFeatureConverter.java
```diff
@@ -0,0 +1,142 @@
...
+            }
+        }
+        return simpleFeature;
+    }
+
+    private String getAsString(SimpleFeature simpleFeature, String name) {
```
nmco: ?

src/main/src/test/java/org/geoserver/catalog/retype/MockRetypedFeatureConverter.java
```diff
@@ -0,0 +1,142 @@
...
+                    existingProperties.add(MockRetypeFeatureTypeCallback.LONG_FIELD);
+                // make sure latitude field is present
+                if (!existingProperties.contains(MockRetypeFeatureTypeCallback.LAT_FIELD))
+                    existingProperties.add(MockRetypeFeatureTypeCallback.LAT_FIELD);
+
+                properties = new ArrayList<>(existingProperties);
+            }
+
+        } catch (Exception e) {
+            e.printStackTrace();
```
nmco: ?

src/main/src/test/java/org/geoserver/catalog/retype/MockRetypedSource.java
```diff
@@ -0,0 +1,75 @@
...
+import org.geotools.data.Query;
+import org.geotools.data.simple.SimpleFeatureCollection;
+import org.geotools.data.simple.SimpleFeatureSource;
+import org.opengis.feature.simple.SimpleFeatureType;
+import org.opengis.filter.Filter;
+
+/*
+ * A Mock Retyped Feature Source demonstrating how to handle dynamic dynamic Geometry generations
+ * */
+public class MockRetypedSource extends DecoratingSimpleFeatureSource {
```
nmco: Same comments above.

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
imranrajjad: will do

src/main/src/test/java/org/geoserver/catalog/MockRetypeFeatureTypeCallback.java
imranrajjad: will change

src/main/src/test/java/org/geoserver/catalog/MockRetypeFeatureTypeCallback.java
imranrajjad: this implementation will only work for vector layer using this property file as a data source.

src/main/src/test/java/org/geoserver/catalog/MockRetypeFeatureTypeCallback.java
imranrajjad: will handle it

src/main/src/test/java/org/geoserver/catalog/MockRetypeFeatureTypeCallback.java
imranrajjad: this implementation will only work for vector layer using this property file as a data source.

src/main/src/test/java/org/geoserver/catalog/RetypeFeatureTypeCallbackTest.java
imranrajjad: will fix

src/main/src/test/java/org/geoserver/catalog/RetypeFeatureTypeCallbackTest.java
imranrajjad: will remove

src/main/src/test/java/org/geoserver/catalog/RetypeFeatureTypeCallbackTest.java
imranrajjad: will change

src/main/src/test/java/org/geoserver/catalog/RetypeFeatureTypeCallbackTest.java
imranrajjad: added

src/main/src/test/java/org/geoserver/catalog/retype/MockRetypedFeatureCollection.java
imranrajjad: removed

src/main/src/test/java/org/geoserver/catalog/retype/MockRetypedFeatureConverter.java
imranrajjad: will remove

src/main/src/test/java/org/geoserver/catalog/retype/MockRetypedFeatureConverter.java
imranrajjad: will change the name, its being used in two classes as a helper class hence cannot be made private inner static

src/main/src/test/java/org/geoserver/catalog/retype/MockRetypedFeatureConverter.java
imranrajjad: will handle

src/main/src/test/java/org/geoserver/catalog/retype/MockRetypedSource.java
imranrajjad: will remove

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
imranrajjad: will add a check

imranrajjad:
@nmco 
Un-related failure on Java 11
`[INFO] GeoWebCache (GWC) Module ........................... FAILURE [02:56 min]`

imranrajjad:
@nmco requested changes applied

- Moved test implementations to Unit test as inner class
- Broke the unit test into three unit tests as instructed

src/main/src/test/java/org/geoserver/catalog/RetypeFeatureTypeCallbackTest.java
```diff
@@ -0,0 +1,467 @@
...
+                        existingProperties.add(TestRetypeFeatureTypeCallback.LONG_FIELD);
+                    // make sure latitude field is present
+                    if (!existingProperties.contains(TestRetypeFeatureTypeCallback.LAT_FIELD))
+                        existingProperties.add(TestRetypeFeatureTypeCallback.LAT_FIELD);
+
+                    properties = new ArrayList<>(existingProperties);
+                }
+
+            } catch (Exception e) {
+                e.printStackTrace();
```
nmco: ?

src/main/src/test/java/org/geoserver/catalog/RetypeFeatureTypeCallbackTest.java
```diff
@@ -0,0 +1,467 @@
...
+        }
+
+        @Override
+        public SimpleFeatureType getSchema() {
+
+            SimpleFeatureType src = super.getSchema();
+            try {
+                return converter.defineGeometryAttributeFor(featureTypeInfo, src);
+            } catch (Exception e) {
+                e.printStackTrace();
```
nmco: ?

nmco:
@imranrajjad those ``e.printStackTrace();`` need to be replace whit more meaningful exception handling. 

I think that the main test case for ``RetypeFeatureTypeCallbackTest.java`` is too complex, but better a complex test than no test.

imranrajjad:
@nmco done, lets wait for builds to finish

indeed the tests are complex, I believe they will serve as a good example for anyone to use this extension point in future.

imranrajjad:
@nmco all green, squash please 👍

nmco:
Thx @imranrajjad, I have asked on the ML if this one can still go in the 2.17-RC, let's wait for feedback there before merging.

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -1111,7 +1111,13 @@ private FeatureType buildFeatureType(
                 }
             }
             ft = tb.buildFeatureType();
+            // extension point for retyping the feature type
+            for (RetypeFeatureTypeCallback callback :
+                    GeoServerExtensions.extensions(RetypeFeatureTypeCallback.class)) {
+                ft = callback.retypeFeatureType(info, ft);
```
jodygarnett: protect yourself with a try catch here so you can clearly blame any implementation that causes problems

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -1381,6 +1387,13 @@ else if (ppolicy == ProjectionPolicy.NONE && nativeCRS != null) {
                 schema = typeBuilder.buildFeatureType();
             }
 
+            // applying wrappers using implementations of RetypeFeatureTypeCallback
+            for (RetypeFeatureTypeCallback callback :
+                    GeoServerExtensions.extensions(RetypeFeatureTypeCallback.class)) {
+                if (SimpleFeatureSource.class.isAssignableFrom(fs.getClass()))
```
jodygarnett: Same advise on try / catch around to catch and report errors when calling any extension

nmco:
Good feedback @jodygarnett, I think that's something we shoudl definitely improve (the try and catch). And yes as per the proposal we are targeting 2.16.4 also. I would suggest this, make a PR to master fixing the try catch, then backport to 2.16.x, what do you think @jodygarnett and @imranrajjad?

imranrajjad:
@nmco its already been merged, shall I do a PR and then we do the backport?

PR #4042, opened 2020-02-10
GEOS-9457: Restore the usage instruction of Windows Installer (only add ref to dev manual)

ronaldhoek:
Below the list of available binaries, I added a note which:

- states GeoServer not longer provides the Windows Installers
- tells the user can create the Windows Installer themselves when needed

This way we hope to get fewer questions about the 'missing' Windows Installer in the download section.

## Checklist

### DIFF
--- a/doc/en/user/source/installation/index.rst
+++ b/doc/en/user/source/installation/index.rst
@@ -18,4 +18,4 @@ There are many ways to install GeoServer on your system. This section will discu
    war
    upgrade
 
-.. note:: At this time we no longer provide a Windows Installer, because the benefics do not outstand the costs (it's basically a wrapper around the win_binary). However if you realy need the Windows installer, you can create one following `the instruction on this site <../../developer/win-installer.html>`.
+.. note:: At this time, for several reasons, we no longer provide the Windows Installer. However if you realy need the Windows installer, you can create one following `the instruction on this site <../../developer/win-installer.html>`.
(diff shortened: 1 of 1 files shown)


doc/en/user/source/installation/index.rst
```diff
@@ -17,3 +17,5 @@ There are many ways to install GeoServer on your system. This section will discu
    linux
    war
    upgrade
+
+.. note:: At this time we no longer provide a Windows Installer, because the benefics do not outstand the costs (it's basically a wrapper around the win_binary). However if you realy need the Windows installer, you can create one following `the instruction on this site <../../developer/win-installer.html>`.
```
aaime: The wording is misleading, it's not the cost benefit per se, we need:
- A well secured Windows server managed by a trusted organization that can host the OSGeo certificate used to sign installers. This alone would reinstate the presence of the installer in releases.
- Someone to redo the installer from scratch, using Tomcat  instead of a reduced Jetty, to get a production worthy installer, instead of a demo oriented one. Not needed, but useful for those that actually want to run production on Windows.

ronaldhoek: I was looking for a 'simple' an 'small' note, without a complete motivation.
Therefore I now replaced the 'specific' motivation by an 'open' motivation.

Or would you rather have this motivation written out? 
If so, it would be best to have it outside the 'note' section?

aaime: "At this time we no longer provide a Windows Installer, due to lack of a secure Windows machine where the installer can be built and signed" would be better.

ronaldhoek: Final update using your suggestion @aaime

aaime:
Looks good thanks!

ronaldhoek:
> Looks good thanks!

Your welcome ;)

PR #4047, opened 2020-02-12
[GEOS-9498] Add Backward compatibility on Layergroups with legacy Cascaded WMS Layers

imranrajjad:
JIRA: https://osgeo-org.atlassian.net/browse/GEOS-9498
- fixes in LayerGroupHelper to resolve correct remote style
for cascaded wms layer memeber layer of group
- added new unit tests

## Checklist

For all pull requests:

- [ ] Committs changing the REST API, or any configuration object, should check it the REST API docs (Swagger YAML files and classic documentation) need to be updated.

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/LayerGroupHelper.java
+++ b/src/main/src/main/java/org/geoserver/catalog/LayerGroupHelper.java
@@ -14,6 +14,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.geoserver.catalog.LayerGroupInfo.Mode;
 import org.geoserver.catalog.impl.StyleInfoImpl;
+import org.geoserver.catalog.impl.WMSLayerInfoImpl;
 import org.geoserver.platform.GeoServerExtensions;
 import org.geoserver.platform.ServiceException;
 import org.geotools.geometry.jts.ReferencedEnvelope;
@@ -241,7 +242,18 @@ public class LayerGroupHelper {
                     PublishedInfo p = group.getLayers().get(i);
                     StyleInfo s = group.getStyles().get(i);
                     if (p instanceof LayerInfo) {
-                        styles.add(group.getStyles().get(i));
+                        StyleInfo styleInfo = group.getStyles().get(i);
+                        if (((LayerInfo) p).getResource() instanceof WMSLayerInfo) {
+                            // legacy < 2.16.2 support
+                            // old wms cascaded layers are by-default configured to use local
+                            // `raster` style
+                            // this utility method call ensures that layer is set to use default on
+                            // remote
+                            WMSLayerInfo wmsLayerInfo =
+                                    (WMSLayerInfo) ((LayerInfo) p).getResource();
+                            styleInfo = processWMSLayer(wmsLayerInfo, styleInfo);
+                        }
+                        styles.add(styleInfo);
                     } else if (p instanceof LayerGroupInfo) {
                         allStylesForRendering((LayerGroupInfo) p, styles, false);
                     } else if (p == null && s != null) {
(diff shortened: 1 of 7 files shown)


imranrajjad:
@nmco for your review and merge

src/main/src/main/java/org/geoserver/catalog/LayerGroupHelper.java
```diff
@@ -241,7 +242,18 @@ private static void allStylesForRendering(
                     PublishedInfo p = group.getLayers().get(i);
                     StyleInfo s = group.getStyles().get(i);
                     if (p instanceof LayerInfo) {
-                        styles.add(group.getStyles().get(i));
```
nmco: Why doesn't the ``group.getStyles()`` already contains the correct style?  When a layer is added the layer group a style is selected right? at that moment it would make  sense to select the right style for remote layers too.

src/main/src/main/java/org/geoserver/catalog/LayerGroupHelper.java
```diff
@@ -531,6 +543,29 @@ private static void expandStyleGroup(
...
     }
 
+    /**
+     * Ensures that WMSLayer Info inside layergroup associates with correct remote style
+     *
+     * @return returns a compatible StyleInfo representing a remote style if required
+     */
+    private static StyleInfo processWMSLayer(WMSLayerInfo wmsLayerInfo, StyleInfo styleInfo) {
+
+        if (styleInfo == null) styleInfo = wmsLayerInfo.getDefaultStyle();
```
nmco: I find this method hard to read:
- Why is named process layer? it doesn't seem to process anything specific, it just retrieves the correct style for a remote layer no? so why not name it something like ``getRemoteWmsLayerStyle``.
- What's the provided ``styleInfo`` object?

src/main/src/main/java/org/geoserver/catalog/LayerGroupHelper.java
```diff
@@ -531,6 +543,29 @@ private static void expandStyleGroup(
...
+        else if (!wmsLayerInfo.findRemoteStyleByName(styleInfo.getName()).isPresent()) {
+            if (LOGGER.isLoggable(Level.FINE)) {
```
nmco: If the select style is not a valid style for this layer then an exception should be throw. Otherwise this may hide this fact and everything will look fine until it doesn't.

src/main/src/main/java/org/geoserver/catalog/LayerGroupHelper.java
imranrajjad: understood will change name.

The provided style is whatever style was configured against the layer during layergroup configuration. In case of cascaded wms layer, its the default raster

src/main/src/main/java/org/geoserver/catalog/LayerGroupHelper.java
```diff
@@ -531,6 +543,29 @@ private static void expandStyleGroup(
...
+                LOGGER.fine(
+                        styleInfo.getName()
+                                + " style is not a known remote style for WMS Layer "
+                                + wmsLayerInfo
+                                + ","
+                                + " Re-configure the Resourcer");
```
aaime: Resourcer?

imranrajjad: 🤦‍♂

src/main/src/main/java/org/geoserver/catalog/LayerGroupHelper.java
imranrajjad: because the style group does not know about the layer default

imranrajjad:
@aaime un-related failure on Java 11


```
[ERROR] directoryCreateEventWithContents(org.geoserver.platform.resource.FileSystemResourceTheoryTest)  Time elapsed: 0.555 s  <<< ERROR!
685
org.awaitility.core.ConditionTimeoutException: Lambda expression in org.awaitility.core.ConditionFactory that uses java.util.concurrent.atomic.AtomicReference: expected not null but was null within 500 milliseconds.
686
	at org.geoserver.platform.resource.FileSystemResourceTheoryTest.directoryCreateEventWithContents(FileSystemResourceTheoryTest.java:263)
```

Shall I rebase?

aaime:
I've re-kicked the build

imranrajjad:
@aaime 

Java 11 failing on un-related module 
`[INFO] GeoWebCache (GWC) Module ........................... FAILURE [02:43 min]`

PR #4052, opened 2020-02-13
Metadata: plugin configuration and tutorial update

NielsCharlier:


aaime:
@NielsCharlier I see only changes in community modules and associated documentation?
If so, it does not need review, you can just merge it (you could have even committed directly, though I agree going through a PR is better).

PR #4098, opened 2020-03-02
[GSIP-184] Promote MBStyle to extension

jodygarnett:
The Vector Tile extension for GeoServer is a popular way to generate tiles for use in MapBox.GL or OpenLayers.

The MBStyle extension is a great addition to this workflow, enabling the same styling for both client side and server side.

See also:

* https://github.com/geoserver/geoserver/wiki/GSIP-184
* https://osgeo-org.atlassian.net/browse/GEOS-9523

## Checklist

For all pull requests:

- [x] Commit message must be in the form "[GEOS-XYZW] Title of the Jira ticket" (using GSIP here)


src/extension/mbstyle/src/test/java/org/geoserver/community/mbstyle/web/MBStyleEditPageTest.java
```diff
@@ -81,20 +82,14 @@ public void testMbstyleChange() throws Exception {
...
-                                .symbolizers()
-                                .get(0);
-        ExternalGraphic eg = (ExternalGraphic) ps.getGraphic().graphicalSymbols().get(0);
-        assertEquals(
-                eg.getURI(),
-                "http://localhost:8080/geoserver/styles/mbsprites#icon=${strURLEncode('circle')}&size=${strURLEncode(Interpolate(POP_MAX,0,0.7,40000000,3.7,'numeric'))}");
+        Style style = getCatalog().getStyleByName("mbstyle").getStyle();
+        Symbolizer sym = style.featureTypeStyles().get(0).rules().get(0).symbolizers().get(0);
+
+        assertTrue(sym instanceof PointSymbolizer);
```
jodygarnett: @tbarsballe there is a significant regression found by this test, to be handled in the geotools code.

It looks like references to a sprite sheet are not being picked or have changed. Still I would like to track this as a separate issue rather than hold up the pull request (which just covers geoserver integration).

aaime:
Did you try an assembly? Does the zip contain the jars you'd expect?
I think you need to make ``src/release/pom.xml`` depend on the mbstyle module for that to happen, but I don't see that file modified (basically what you removed from the ``src/community/release/pom.xml`` should go into ``src/release/pom.xml``).

jodygarnett:
> Did you try an assembly? Does the zip contain the jars you'd expect?

Let me test that, right now I am making the jira tickets.

> I think you need to make `src/release/pom.xml` depend on the mbstyle module for that to happen, but I don't see that file modified (basically what you removed from the `src/community/release/pom.xml` should go into `src/release/pom.xml`).

I did a search and replace in the the codebase, but may of missed something :)

src/extension/mbstyle/src/test/java/org/geoserver/community/mbstyle/web/MBStyleEditPageTest.java
jodygarnett: Reporting issue here: https://osgeo-org.atlassian.net/browse/GEOT-6522

tbarsballe: The test itself is intended to just make sure geoserver properly updates the cached SLD for non-SLD styles in the resource pool when a style is saved using the UI; it just happens to use a style that contains an icon.
The icon lookup is to verify the conversion worked.

As for the spritesheet reference - for named icons matching the well-known icon names in GeoServer (e.g. 'circle'), mbstyle should be bypassing the spritesheet entirely. I can't recall where exactly that special case was implemented - I'll try and track it down. So, the issue isn't quite that its not finding the spritesheet, since the spritesheet listed in the style is just a dummy value and doesn't exist 

As an aside, commenting out a portion of the test and setting  the test to `@Ignore` is overkill - just use Ignore and preserve the test content.

jodygarnett: Andrea points out this change was intensional (https://osgeo-org.atlassian.net/browse/GEOT-6443) so I will update this test case to resolve.

tbarsballe: Right, that makes sense. You should just be able to grab a spritesheet + sprite name from one of the GeoTools MBStyle tests - the appearance of the style doesn't matter here, just that it gets properly translated into an SLD symbol

doc/en/user/source/styling/mbstyle/installing.rst
```diff
@@ -3,8 +3,14 @@
 Installing the GeoServer MBStyle extension
 ==========================================
 
- #. Download the extension from the `nightly GeoServer community module builds <https://build.geoserver.org/geoserver/master/community-latest/>`_.
+The MBStyle extension is listed on the GeoServer download page.
```
tbarsballe: An actual link to the download page might be nice here, since there was a link before.

jodygarnett: Trying to think what to do for that, either link to the geoserver release page, or straight to source forge.

tbarsballe: I'd say just the GeoServer release page would be simplest

jodygarnett: Actually that ended up being hard:

- We could hard code it (and then have a maintenance trouble updating each page)
- There was no way to use substitutions (since those cannot be used in generating links)
- Ended up making three external link definitions (artifacts, extension plugins, community plugins) ... and then making those dynamic based on "snapshot" (build server) or "release" (source forge)

Please have a look.

tbarsballe: You went way fancier then I was expecting, I was just thinking linking http://geoserver.org , or http://geoserver.org/release/{version} 
I was just looking for something of comparable utility to the prior link.

Your solution does seem reasonable, assuming it is stable - i.e. doesn't break next time the we update versions. It looks stable enough, but sometimes the release scripts miss a version # somewhere (or a step gets inadvertently skipped).

jodygarnett:
Thanks everyone for the review, I believe I have accounted for all the feedback.

doc/en/user/source/styling/mbstyle/installing.rst
jodygarnett: Yeah as above there was no way to add "version" into a URL, so no kill like overkill.

PR #4100, opened 2020-03-03
[GSIP-185] web-resource extension

jodygarnett:
GeoServer Resource API provides programatic management of configuration files and icons. The REST API makes this functionality available for automation, we would like to make web-resource module an extension to offer this capability to web administration application.

See also:

* [GSIP-185](https://github.com/geoserver/geoserver/wiki/GSIP-185)
* [GEOS-9524](https://osgeo-org.atlassian.net/browse/GEOS-9524)

## Checklist

For all pull requests:

- [x] There is a ticket in Jira: [GEOS-9524](https://osgeo-org.atlassian.net/browse/GEOS-9524)


doc/en/user/source/configuration/tools/resource/browser.rst
```diff
@@ -0,0 +1,68 @@
+.. _web_resource:
+
+Resource Browser
+================
+
+The :guilabel:`Resource Browser` provides a tree showing configuration folders, along with actions to edit and mange resources.
+
+.. figure:: img/browser.png
+   
+   Catalog Bulk Load Tool
```
aaime: Bulk load tool?

doc/en/user/source/extensions/controlflow/index.rst
```diff
@@ -165,22 +165,6 @@ A complete example
...
-  # don't allow a single user to perform more than 6 requests in parallel
-  # (6 being the Firefox default concurrency level at the time of writing)
-  user=6
-  # don't allow the execution of more than 16 tile requests in parallel
-  # (assuming a server with 4 cores, GWC empirical tests show that throughput
-  # peaks up at 4 x number of cores. Adjust as appropriate to your system)
-  ows.gwc=16
-
-
+.. literalinclude:: controlflow.properties
```
aaime: Why the unrelated changes?

doc/en/user/source/configuration/tools/resource/browser.rst
jodygarnett: it was undocumented

doc/en/user/source/extensions/controlflow/index.rst
jodygarnett: So I could reuse the code example

aaime:
Ah one thing that came to mind, is the module accessible only to the full administrator? Or also to the workspace ones? If the latter, are the workspace admins locked down into their own workspace folders?

jodygarnett:
> Ah one thing that came to mind, is the module accessible only to the full administrator? Or also to the workspace ones? If the latter, are the workspace admins locked down into their own workspace folders?

Just tested, only available to admin. Workspace admin does not see any tools.

image

aaime:
NIce, then only the current (problematic, to me) position of the docs is left. If you don't agree let's maybe discuss on list, not sure everyone participates to PR feedbacks (I mean they should, but in practice there is no much activity).

jodygarnett:
> NIce, then only the current (problematic, to me) position of the docs is left. If you don't agree let's maybe discuss on list, not sure everyone participates to PR feedbacks (I mean they should, but in practice there is no much activity).

I agree, this is just not the place to fix, would prefer a doc wide cleanup.

PR #4104, opened 2020-03-04
Fix nightly build

jodygarnett:
Fixing nightly build after some relative path license references did not work out.


jodygarnett:
I tested locally using from the `src` folder:
```
mvn clean install -T 2C -Prelease -DskipTests
mvn assembly:attached -Prelease -T 2C
```

I think this agrees with [build_release.sh|https://github.com/geoserver/geoserver/blob/master/build/build_release.sh#L29]

PR #4111, opened 2020-03-05
[GEOS-9526] review extension license information

jodygarnett:
Working on a couple of glitches:

* [x] maven assembly file inclusion sensitive to location of `mvn -f src/pom.xml`
* [x] upgrade to maven assembly plugin 3.2.0 to use of `mvn assembly:single`
* [x] review extension licenses
* [x] Avoid multiple files called LICENSE.txt by changing to GPL.md and LGPL.md as needed
* [x] use markdown licenses where available
* [x] distinguish between license NOTICE.md (describing how licenses are used) and the licenses themselves

See [GEOS-9526](https://osgeo-org.atlassian.net/browse/GEOS-9526) for detailed discussion.

## Checklist

For all pull requests:

- [x] Confirm you have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/master/CONTRIBUTING.md) 
- [x] You have sent a Contribution Licence Agreement (CLA) as necessary (not required for small changes, e.g., fixing typos in documentation)
- [x] Make sure the first PR targets the master branch, eventual backports will be managed later. This can be ignored if the PR is fixing an issue that only happens in a specific branch, but not in newer ones.

The following are required only for core and extension modules (they are welcomed, but not required, for community modules):
- [x] There is a ticket in Jira [GEOS-9526](https://osgeo-org.atlassian.net/browse/GEOS-9526)
- [x] PR for bug fixes and small new features are presented as a single commit
- [x] Commit message must be in the form "[GEOS-XYZW] Title of the Jira ticket" (export to XML in Jira generates the message in this exact form)
- [x] The pull request contains changes related to a single objective. If multiple focuses cannot be avoided, each one is in its own commit and has a separate ticket describing it.
- [ ] This PR passes all existing unit tests (test results will be reported by travis-ci after opening this PR)
- [x] This PR passes the [QA checks](https://docs.geoserver.org/latest/en/developer/qa-guide/index.html) (QA checks results will be reported by travis-ci after opening this PR)

Submitting the PR does not require you to check all items, but by the time it gets merged, they should be either satisfied or inapplicable.

PR #4115, opened 2020-03-07
[GEOS-9527] include readme in ogr-wps extension

jodygarnett:
Follow up to https://github.com/geoserver/geoserver/pull/4111


jodygarnett:
Using this to trigger a build and confirm builds are now fixed

PR #4147, opened 2020-03-27
[GEOS-9539]: avoid fixing correct urls in URLKvpParser

mbarto:
URLKvpParser fixes every url, also non broken ones, introducing a wrong encoding in some cases.
This fix tries to use the url right away, and fix it only if it is not valid.

## Checklist

For all pull requests:

### DIFF
--- a/src/ows/src/test/java/org/geoserver/ows/kvp/URLKvpParserTest.java
+++ b/src/ows/src/test/java/org/geoserver/ows/kvp/URLKvpParserTest.java
@@ -1,0 +1,37 @@
+/* (c) 2020 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.ows.kvp;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.fail;
+
+import java.net.URL;
+import org.junit.Before;
+import org.junit.Test;
+
+public class URLKvpParserTest {
+    URLKvpParser parser;
+
+    @Before
+    public void setUp() {
+        parser = new URLKvpParser("url");
+    }
+
+    @Test
+    public void testValidUrl() {
+        try {
+            String validUrl =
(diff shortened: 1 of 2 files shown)


src/ows/src/test/java/org/geoserver/ows/kvp/URLKvpParserTest.java
```diff
@@ -0,0 +1,37 @@
...
+public class URLKvpParserTest {
+    URLKvpParser parser;
+
+    @Before
+    public void setUp() {
+        parser = new URLKvpParser("url");
+    }
+
+    @Test
+    public void testValidUrl() {
```
bradh: Is there already a test for invalid URL?

mbarto: @bradh the code was not tested before, and it's not even clear to me why the fixUrl was introduced, because it was breaking a valid url, that now is instead working properly.
To give you some context: this parser is only used at the moment to parse WMS SLD parameter.
When parsing a url with a complex querystring, used to produce a dynamic style with the sldService extension, the fixed url was indeed broken, while with the new code path it works perfectly.
The doubt was if we wanted to remove fixURL completely, but not understanding its real purpose we decided to use it as a fallback.
Long story short: I couldn't find a test case for an invalid url, because I couldn't imagine the purpose of the original code.

aaime: My guess is that lots of existing tests in GeoServer are not properly url-encoding the urls, so if you skip the "fixurl" part they will start failing. But it's just a guess.

aaime: @jdeolive I know it was from many many years ago, but do you remember by any chance why the "on the fly url fixing" was introduced in the URLKvpParser, to start with? Seems like it came in with the first implementation of KVP parsers.

aaime: Actually tried it out, removing the fixurl part, letting it throw an exception, and the build still works. Tried to put together a URL that would throw a fixable MalformedURLException, but could not... maybe the behavior of Java own URL class changed over time?
In any case, I'd keep the code as is for safety, since we don't know exactly what it fixes, and move on.

PR #4157, opened 2020-04-03
GSIP-188: Add extension point for ResourcePool to convert coverage in…

joshfix:
https://github.com/geoserver/geoserver/wiki/GSIP-188
https://osgeo-org.atlassian.net/browse/GEOS-9563
​
ImageInputStreamSpi implementations advertise an inputClass. When utilities like GeoTools' ImageIOExt.getImageInputStreamSPI() are invoked, they try to match the provided input object class to the advertised inputClass of the ImageInputStreamSpi. When creating a GridCoverageReader in the ResourcePool, GeoServer only supplies the GridCoverageReaders with URLs in the form of Strings, with one exception. There is a single method to check if the input String represents a File URL, and if so, it converts the input String to a File.​

This input object is eventually used by ImageIO to determine which ImageInputStream to use by comparing the input class to the advertised inputClass as previously mentioned. The power of being able to advertise an input class is severely crippled by GeoServer’s inability to create readers with alternative input classes. Creating new reader implementations becomes difficult because there are already specific ImageInputStreamSpis to support String, URL, File, InputStream, etc. Ambiguity is introduced when more than one reader advertises the same inputClass. For example, S3ImageInputStreamImplSpi and StringImageInputStreamSpi both advertise the input class as String. This results in the ImageIO createImageInputStream method selecting the first SPI it comes across that supports String.​

This proposal seeks to implement a new extension point that will allow the ResourcePool to supply configurable input objects that extend beyond String/File. This enhancement ensures the specific, desired ImageInputStreamSpi implementation is able to be identified and selected.​

## Checklist

For all pull requests:

- [n/a] Commits changing the UI, existing user workflows, or adding new functionality, need to include documentation updates (screenshots, text)
- [n/a] Commits changing the REST API, or any configuration object, should check if the REST API docs (Swagger YAML files and classic documentation) need to be updated.

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/CoverageReaderFileConverter.java
+++ b/src/main/src/main/java/org/geoserver/catalog/CoverageReaderFileConverter.java
@@ -1,0 +1,80 @@
+/* (c) 2014 - 2020 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2013 OpenPlans
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.catalog;
+
+import java.io.File;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Optional;
+import javax.annotation.Nullable;
+import org.geoserver.platform.resource.Files;
+import org.geoserver.platform.resource.Resource;
+import org.geoserver.platform.resource.Resources;
+import org.geotools.util.factory.Hints;
+import org.opengis.coverage.grid.GridCoverageReader;
+
+/**
+ * Attempts to convert the source input object for a {@link GridCoverageReader} to {@link File}.
+ *
+ * @author joshfix Created on 2/25/20
+ */
+public class CoverageReaderFileConverter implements CoverageReaderInputObjectConverter<File> {
+
+    private final Catalog catalog;
+
+    public CoverageReaderFileConverter(Catalog catalog) {
+        this.catalog = catalog;
+    }
+
+    /**
+     * Performs the conversion of the input object to a file object. If this converter is not able
+     * to convert the input to a File, an empty {@link Optional} will be returned.
+     *
+     * @param input The input object.
+     * @param coverageInfo The grid coverage metadata, may be <code>null</code>.
+     * @param hints Hints to use when loading the coverage, may be <code>null</code>.
+     * @return
+     */
+    @Override
+    public Optional<File> convert(
+            Object input, @Nullable CoverageInfo coverageInfo, @Nullable Hints hints) {
+        if (!(input instanceof String)) {
+            return Optional.empty();
+        }
+        String urlString = (String) input;
+        return canConvert(urlString) ? convertFile(urlString) : Optional.empty();
+    }
+
+    /**
+     * Checks to see if the input string is a file URI.
+     *
+     * @param input The input string.
+     * @return Value representing whether or not this converter is able to convert the provided
+     *     input to File.
+     */
+    protected boolean canConvert(String input) {
+        // Check to see if our "url" points to a file or not, otherwise we use the string itself for
+        // reading
+        try {
+            URI uri = new URI(input);
+            return uri.getScheme() != null && "file".equalsIgnoreCase(uri.getScheme());
+        } catch (URISyntaxException e) {
+            return false;
+        }
+    }
+
+    /**
+     * Performs the conversion to File.
+     *
+     * @param input The input string.
+     * @return The Optional object containing the File
+     */
+    protected Optional<File> convertFile(String input) {
+        Resource resource = Files.asResource(catalog.getResourceLoader().getBaseDirectory());
+        final File readerFile = Resources.find(Resources.fromURL(resource, input), true);
+        return (readerFile != null) ? Optional.of(readerFile) : Optional.empty();
+    }
+}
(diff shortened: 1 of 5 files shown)


src/main/src/main/java/org/geoserver/catalog/CoverageReaderFileConverter.java
```diff
@@ -0,0 +1,80 @@
+/* (c) 2014 - 2020 Open Source Geospatial Foundation - all rights reserved
```
aaime, line 1: New files just need  ``(c) 2020 Open Source Geospatial Foundation - all rights reserved``
There is no need to add the rest.

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -1,4 +1,4 @@
-/* (c) 2014 - 2017 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2014 - 2020 Open Source Geospatial Foundation - all rights reserved
```
aaime, line 1: No requirement to update the dates anymore, we have determined it has no legal value.

src/main/src/main/java/org/geoserver/catalog/CoverageReaderFileConverter.java
joshfix: Do you want the unnecessary openplans lines removed?

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
joshfix: Should I revert the dates?

src/main/src/main/java/org/geoserver/catalog/CoverageReaderFileConverter.java
aaime: @jodygarnett do we have to remove them?

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
aaime: @jodygarnett do we have to revert the dates?

src/main/src/main/java/org/geoserver/catalog/CoverageReaderFileConverter.java
jodygarnett: For new files, yes because they had nothing to do with OpenPlans (it would just make things more complicated to have them there).

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
jodygarnett: No value to update the dates (no harm in it either) --> no need to revert.

aside: We do ask that a date be added when a new file is created.

aaime:
Various failures... at least some could be random, I've restarted the Travis portion, which is the one tha must pass (the Win/OSX are aspirational at this point...).

However, noted that the QA one had a legit failure:
```
[ERROR] Non complying file: /home/travis/build/geoserver/geoserver/src/main/src/test/java/org/geoserver/catalog/ResourcePoolTest.java
```

aaime:
The Travis build failure looked suspicious so I've run the build locally, as a double check.
It indeed fails, some change in this PR may be causing it:

```
Test set: org.geoserver.importer.mosaic.ImporterMosaicTest
Tests run: 8, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 3.077 s <<< FAILURE! - in org.geoserver.importer.mosaic.ImporterMosaicTest
testPopulateEmptyMosaic(org.geoserver.importer.mosaic.ImporterMosaicTest)  Time elapsed: 0.149 s  <<< ERROR!
java.io.IOException: 
Failed to create reader from ./emptyMosaic and hints Hints:
  EXECUTOR_SERVICE = java.util.concurrent.ThreadPoolExecutor@213e10c9[Running, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 0]
  REPOSITORY       = org.geoserver.catalog.CatalogRepository@3cfc0190                                                                                   
System defaults:
  FORCE_LONGITUDE_FIRST_AXIS_ORDER = true  
  FORCE_AXIS_ORDER_HONORING        = http  
  COMPARISON_TOLERANCE             = 1.0E-9
```

src/main/src/main/java/org/geoserver/catalog/CoverageReaderFileConverter.java
```diff
@@ -0,0 +1,79 @@
...
+     * @param input The input string.
+     * @return Value representing whether or not this converter is able to convert the provided
+     *     input to File.
+     */
+    protected boolean canConvert(String input) {
+        // Check to see if our "url" points to a file or not, otherwise we use the string itself for
+        // reading
+        try {
+            URI uri = new URI(input);
+            return uri.getScheme() != null && "file".equalsIgnoreCase(uri.getScheme());
```
aaime, line 62: There is indeed a behavioral change, the previous code was trying to lookup the files using the resource loader even if the string representation had no scheme, this one does not. The following change fixes the specific build failure.

```suggestion
            return uri.getScheme() == null || "file".equalsIgnoreCase(uri.getScheme());
```

I've verified this change makes for a successful build (with all extensions included, -Prelease):

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  06:12 min (Wall Clock)
[INFO] Finished at: 2020-05-06T15:36:28+02:00
[INFO] ------------------------------------------------------------------------
```

aaime: @joshfix are you still working on this one?

aaime:
The Java 11 failure looks like a maven issue unrelated to this change. Merging, thanks @joshfix

PR #4186, opened 2020-04-13
review build_release script and assembly definitions

jodygarnett:
This PR checks that everything is able to be released after recent repository changes:

* [x] revise extension content based on RC review
* [x] break community and communityRelease out into profiles
* [x] removes gs-geoscript-py from default communityRelease profile

## Checklist

For all pull requests:

### DIFF
--- a/src/community/pom.xml
+++ b/src/community/pom.xml
@@ -36,9 +36,6 @@
       <configuration>
         <descriptors>
           <descriptor>release/ext-arcsde.xml</descriptor>
-          <!-- <descriptor>release/ext-groovy.xml</descriptor> -->
-          <!-- <descriptor>release/ext-javascript.xml</descriptor> -->
-          <!-- <descriptor>release/ext-spatialite.xml</descriptor> -->
           <descriptor>release/ext-jdbcconfig.xml</descriptor>
           <descriptor>release/ext-dds.xml</descriptor>
           <descriptor>release/ext-wms-eo.xml</descriptor>
@@ -49,8 +46,6 @@
           <descriptor>release/ext-pgraster.xml</descriptor>
           <descriptor>release/ext-dyndimension.xml</descriptor>
           <descriptor>release/ext-flatgeobuf.xml</descriptor>
-          <!-- If you are re-enabling this, remember to re-add the documentation -->
-          <!-- <descriptor>release/ext-rest-ext.xml</descriptor> -->
           <descriptor>release/ext-kmlppio.xml</descriptor>
           <descriptor>release/ext-gpx.xml</descriptor>
           <descriptor>release/ext-wps-download.xml</descriptor>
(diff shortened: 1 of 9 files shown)


src/community/pom.xml
```diff
@@ -73,7 +66,6 @@
           <descriptor>release/ext-jdbcstore.xml</descriptor>
```
tbarsballe, line 66: You should remove ext-python.xml just above as well

src/community/pom.xml
```diff
@@ -321,6 +320,21 @@
...
+      <build>
+       <plugins>
+         <plugin>
+          <artifactId>maven-assembly-plugin</artifactId>
+          <version>3.2.0</version>
+          <configuration>
+            <descriptors>
+              <descriptor>release/ext-groovy.xml</descriptor>
+              <descriptor>release/ext-javascript.xml</descriptor>
+              <descriptor>release/ext-python.xml</descriptor>
```
tbarsballe, line 332: I don't think this is going to work how you want it to without significant revisions to src/community/release/pom.xml, since it is used for assembly and depends upon everything in communityRelease at the top level

src/community/release/pom.xml
```diff
@@ -440,4 +403,74 @@
...
+     <id>rest-upload</id>
+     <dependencies>
+       <dependency>
+         <groupId>org.geoserver.community</groupId>
+         <artifactId>gs-rest-upload</artifactId>
+         <version>${project.version}</version>
+       </dependency>
+     </dependencies>
+   </profile>
+ </profiles>
```
tbarsballe, line 475: This looks to be close to what is neccessary for individual assembly, but does not appear to be the same number of modules that you added release descriptors to in the community pom; spatialite at least appears to be missing (so assembly would fail)

src/community/release/pom.xml
```diff
@@ -440,4 +403,74 @@
...
+       </dependency>
+       <dependency>
+         <groupId>org.geoserver.script</groupId>
+         <artifactId>gs-script-js</artifactId>
+         <version>${project.version}</version>
+       </dependency>
+       <dependency>
+         <groupId>org.geoserver.script</groupId>
+         <artifactId>gs-script-groovy</artifactId>
+         <version>${project.version}</version>
```
tbarsballe, line 441: As per the commented out lines above, script-js and script-groovy don't work at all, so this doesn't help for building script-py very much.

src/community/pom.xml
jodygarnett: I saw the profile names used in release/pom.xml and used the same names here.

src/community/release/pom.xml
jodygarnett: No it does not, but it does place them all into the same profile.

src/community/release/pom.xml
jodygarnett: As you noted the codebase was in consistent previously.

gs-spatialite was not mentioned anywhere here before, I can add it back as a profile, if only as a starting point for anyone wishing to restore it.

src/community/pom.xml
jodygarnett: Done, moved it down to the appropriate profile.

PR #4190, opened 2020-04-16
[GEOS-9333] choose image dialog for legend graphic

NielsCharlier:
In the style edit page, the "choose image" dialog that exists in the SLD editor could be used in the legend panel as well, so that the user can either select or upload an image

## Checklist

For all pull requests:


aaime:
Please do not remove the checklist, use it and fill instead. 
Part of the checklist includes updating the documentation for user visible changes, this is one of them.

I've tried to use the image selection dialog in the past and found it to be ... useless... this is because we normally keep images in a sub-directory, and the image dialog seems to only list images found in the styles directory itself. I guess that limitation still holds in this PR?

NielsCharlier:
Yes. The main purpose of the dialog is actually not the selection but to allow the user to upload a new image using the GUI. Especially useful when using jdbc resource store or if the user doesn't have immediate access to the file structure of the server.

The dialog does indeed assume storage of the images in the same folder as the style. It would be a good addition to allow the selection of a subfolder, or perhaps a possibility to configure it to always use a specific subdirectory instead of the same directory. The last option is perhaps better, because not only is it less work to implement ; it would not burden the user with more clicking steps and choices.

I'm sorry about the checklist, I will not remove it in future PR's. I will add a documentation update.

NielsCharlier:
The orange bar is a horizontal scroll bar on my system.
The svg should indeed be removed in this case as an option

NielsCharlier:
fixed extensions

aaime:
Looks good to me, merging.

PR #4212, opened 2020-04-27
[GEOT-6551] SLD 1.1 doesn't support namespaces in OGC PropertyName

NielsCharlier:
accompanies https://github.com/geotools/geotools/pull/2908
checks won't work -> can be redone when geotools issue is merged

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/xml/filter/v1_1/PropertyNameTypeBinding.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/xml/filter/v1_1/PropertyNameTypeBinding.java
@@ -12,7 +12,6 @@ import org.geotools.gml3.GML;
 import org.geotools.gml3.bindings.GML3EncodingUtils;
 import org.geotools.xsd.ElementInstance;
 import org.geotools.xsd.Node;
-import org.opengis.filter.FilterFactory;
 import org.opengis.filter.FilterFactory2;
 import org.opengis.filter.expression.PropertyName;
 import org.xml.sax.helpers.NamespaceSupport;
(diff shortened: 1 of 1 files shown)


src/wfs/src/main/java/org/geoserver/wfs/xml/filter/v1_1/PropertyNameTypeBinding.java
```diff
@@ -31,8 +31,8 @@
     NamespaceSupport namespaceSupport;
 
     public PropertyNameTypeBinding(
-            FilterFactory filterFactory, NamespaceSupport namespaceSupport, Catalog catalog) {
-        super(filterFactory);
+            FilterFactory2 filterFactory, NamespaceSupport namespaceSupport, Catalog catalog) {
+        super(filterFactory, namespaceSupport);
         this.namespaceSupport = namespaceSupport;
```
aaime, line 36: The base class has the namespace support, use it instead of duplicating it. I've added a suggestion to make that field protected in the GeoTools PR.

NielsCharlier:
@jodygarnett This is just to comply with the new geotools API. There is nothing new being implemented here that needs to be tested...

src/wfs/src/main/java/org/geoserver/wfs/xml/filter/v1_1/PropertyNameTypeBinding.java
NielsCharlier: Done

src/wfs/src/main/java/org/geoserver/wfs/xml/filter/v1_1/PropertyNameTypeBinding.java
```diff
@@ -27,12 +26,9 @@
...
 
-    /** parser namespace mappings */
-    NamespaceSupport namespaceSupport;
-
     public PropertyNameTypeBinding(
-            FilterFactory filterFactory, NamespaceSupport namespaceSupport, Catalog catalog) {
-        super(filterFactory);
+            FilterFactory2 filterFactory, NamespaceSupport namespaceSupport, Catalog catalog) {
+        super(filterFactory, namespaceSupport);
         this.namespaceSupport = namespaceSupport;
```
aaime, line 32: Is this just me, or this assignement became useless?

```
        this.namespaceSupport = namespaceSupport;
```

NielsCharlier:
To clarify, this class already had propertyname support. I have added it to its superclass in geotools as well, which required a change to how this class calls its super. This change does not fix a bug or adds a feature, that is done by the change in geotools.

src/wfs/src/main/java/org/geoserver/wfs/xml/filter/v1_1/PropertyNameTypeBinding.java
NielsCharlier: yes, that is true.

aaime:
Merging to avoid compile issues after changes in GT2

PR #4215, opened 2020-04-28
[GEOS-9196] decode special characters with relative paths

NielsCharlier:
https://osgeo-org.atlassian.net/browse/GEOS-9196

## Checklist

For all pull requests:

### DIFF
--- a/src/platform/src/main/java/org/geoserver/platform/resource/Resources.java
+++ b/src/platform/src/main/java/org/geoserver/platform/resource/Resources.java
@@ -12,9 +12,11 @@ import java.io.InputStream;
 import java.io.ObjectInputStream;
 import java.io.OutputStream;
 import java.io.Serializable;
+import java.io.UnsupportedEncodingException;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLConnection;
+import java.net.URLDecoder;
 import java.net.URLStreamHandler;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -592,6 +594,12 @@ public class Resources {
         // pgraster://user:pass@server:port or similar custom store URLs.
         if (url.startsWith("file:")) {
             url = url.substring(5); // remove 'file:' prefix
+
+            // revert encoded special characters and spaces
+            try {
+                url = URLDecoder.decode(url, "UTF-8");
+            } catch (UnsupportedEncodingException e) {
+            }
 
             File f = new File(url);
 
--- a/src/platform/src/test/java/org/geoserver/platform/GeoServerResourceLoaderTest.java
+++ b/src/platform/src/test/java/org/geoserver/platform/GeoServerResourceLoaderTest.java
@@ -9,11 +9,15 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
+import com.google.common.io.Files;
 import java.io.File;
 import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.charset.Charset;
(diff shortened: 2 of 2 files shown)


aaime:
Looks good to me... @jodygarnett could you have a second look? Resource is kind of your area of expertise.

src/platform/src/test/java/org/geoserver/platform/GeoServerResourceLoaderTest.java
```diff
@@ -162,4 +166,22 @@ public void testSetBaseDirectory() throws IOException {
...
+
+    @Test
+    public void fromRelativeURLTest() throws IOException {
+        GeoServerResourceLoader loader = new GeoServerResourceLoader();
+        tempFolder.create();
+        File tempDir = tempFolder.getRoot();
+        loader.setBaseDirectory(tempDir);
+
+        Resource res = loader.fromURL("file:relative/with/special+characters%23");
+        assertEquals("relative/with/special characters#", res.path());
```
jodygarnett, line 178: I always get frustrated with special characters.

@NielsCharlier can you add a test for the unicode precomposed characters which are known to cause problems on macOS.

> For example, the character "ü" can be encoded as the single code point U+00FC "ü" or as the base character U+0075 "u" followed by the non-spacing character U+0308 "¨"

src/platform/src/main/java/org/geoserver/platform/resource/Resources.java
```diff
@@ -593,6 +595,12 @@ public static Resource fromURL(Resource baseDirectory, String url) {
         if (url.startsWith("file:")) {
             url = url.substring(5); // remove 'file:' prefix
 
+            // revert encoded special characters and spaces
+            try {
+                url = URLDecoder.decode(url, "UTF-8");
+            } catch (UnsupportedEncodingException e) {
```
jodygarnett, line 601: Why is this exception being eaten? nom nom nom ...

jodygarnett: Q: This is detecting for relative paths, but misses the absolute path thing above?
Q: Does URLs.urlToFile need similar logic?

NielsCharlier: The exception is eaten, because it cannot occur.
See https://docs.oracle.com/javase/8/docs/api/java/nio/charset/Charset.html
"Every implementation of the Java platform is required to support the following standard charsets. "

As explained on the jira ticket, absolute paths already support special characters, and only relative ones fail. This is exactly because URLS.urlTOFile already has the logic, and it is being called for absolute paths.

src/platform/src/test/java/org/geoserver/platform/GeoServerResourceLoaderTest.java
NielsCharlier: DOne.

aaime:
Looks like one of @jodygarnett questions is still outstanding

NielsCharlier:
Which one? He asked the same questions twice, and they are answered above...

aaime:
Ah ok, saw it now... yeah, twice the same question. Then I think we just need @jodygarnett to complete the review.

src/platform/src/test/java/org/geoserver/platform/GeoServerResourceLoaderTest.java
jodygarnett: Thanks, now to check if the macOS build passes :)

jodygarnett:
Thanks Niels, rebased onto master, backport if needed.

NielsCharlier:
Thanks @jodygarnett

PR #4232, opened 2020-05-10
[GEOS-9611] Upgrade from Apache Common Collections v3 to v4

aaime:
Follows up with https://github.com/geotools/geotools/pull/2796

The changes involve only supported modules for the moment. The old and new libraries can coexist, but have to check if the war still contains the old one, otherwise the community modules packaging instructions would have to be updated.

## Checklist

For all pull requests:


soc:
`MultiValueMap` → `MultiValuedMap` :-)

aaime:
Checked, gs-main still depends to commons-collections 3.x via json-lib, so the jar stays in the release package. Which means, the community modules do not need to be fixed right now.

soc:
@aaime I did a migration of `net.sf.json-lib` to `org.json` a while ago in an unrelated project, maybe that's an option here, too?

aaime:
Afraid not, it's not GPL compatible (just found out): https://www.json.org/license.html and https://github.com/stleary/JSON-java/blob/master/LICENSE

In particular, due to this restriction "The Software shall be used for Good, not Evil."
Well meaning for sure, but impractical to apply.

Checked out dependencies, GWC REST module actually uses it, it should be removed.
I've opened a ticket about it: https://github.com/GeoWebCache/geowebcache/issues/850

PR #4243, opened 2020-05-15
Clarify suggested Java versions

aaime:
Provides a clarification in answer to https://osgeo-org.atlassian.net/browse/GEOS-9619

## Checklist

For all pull requests:

### DIFF
--- a/doc/en/user/source/production/container.rst
+++ b/doc/en/user/source/production/container.rst
@@ -5,6 +5,11 @@ Container Considerations
 
 Java web containers such as `Tomcat <http://tomcat.apache.org>`_ or `Jetty <http://www.mortbay.org/jetty/>`_ ship with configurations that allow for fast startup, but don't always deliver the best performance.
 
+Use a supported Java version
+----------------------------
+
+GeoServer is built and tested with Java 8, `supported up to 2026 <https://adoptopenjdk.net/support.html#roadmap>`_, and Java 11, `supported up to 2024 <https://adoptopenjdk.net/support.html#roadmap>`_. More recent releases of Java are supported for only six months and contain experimental features, they are not tested with GeoServer, thus not recommended for production: the application might not start or misbehave, and the recommended JVM flags might no longer be applicable.
+
 Optimize your JVM
 -----------------
 


tomsaul:
Thanks for that clarification.  I suspect it would also be beneficial to modify the docs that current say 'Additionally you may experiment with -XX:+UseStringDeduplicationJVM to ask G1 to better manage common text strings in memory.' to something like 'Additionally for Java versions before release 12 (when it was deprecated), you may experiment with -XX:+UseStringDeduplicationJVM to ask G1 to better manage common text strings in memory.'

I have not been able to find exactly which version it was removed in, or if this absence is distribution specific.

mprins:
@tomsaul any X options are non-standard so YMMV, as Oracle states:

> Options that begin with -X are non-standard (not guaranteed to be supported on all VM implementations), and are subject to change without notice in subsequent releases of the JDK.
> Options that are specified with -XX are not stable and are subject to change without notice.

With the current Java runtime landscape scattering a situation where each major cloud provider has their own JVM (with specific options and tuning for their environment) and then some suggesting to use "X" options may not be a good idea.

doc/en/user/source/production/container.rst
```diff
@@ -5,6 +5,11 @@ Container Considerations
 
 Java web containers such as `Tomcat <http://tomcat.apache.org>`_ or `Jetty <http://www.mortbay.org/jetty/>`_ ship with configurations that allow for fast startup, but don't always deliver the best performance.
 
+Use a supported Java version
+----------------------------
+
+GeoServer is built and tested with Java 8, `supported up to 2026 <https://adoptopenjdk.net/support.html#roadmap>`_, and Java 11, `supported up to 2024 <https://adoptopenjdk.net/support.html#roadmap>`_. More recent releases of Java are supported for only six months and contain experimental features, they are not tested with GeoServer, thus not recommended for production: the application might not start or misbehave, and the recommended JVM flags might no longer be applicable.
+
```
jodygarnett, line 12: We already have a place for this in our user manual (literally the previous page) and I would like to cross link there, rather than duplicate suggestions.

Introducing a heading "Running on Java 8" there may be more clear?

aaime: @jodygarnett works for me. Can you amend this PR accordingly? 
The PR was triggered by the -XX:+UseStringDeduplicationJVM option not working anymore in recent versions of GeoServer. What's your take on it? Just remove it?

jodygarnett: I was going to ask you to just close this pull request, the previous page [Java Considerat ions](https://docs.geoserver.org/latest/en/user/production/java.html) talks about using a supported java version very clearly.

We also have lots of warnings in the [installation](https://docs.geoserver.org/latest/en/user/installation/index.html#installation) pages on use of Java 8 or Java 11.

image

aaime: Roger that, closing.

PR #4256, opened 2020-05-19
[GEOT-6575] Removing short name functionality and clarifying alias functionality

travisb-fe:
Fixes [GEOT-6575](https://osgeo-org.atlassian.net/browse/GEOT-6575).

Short names are an outdated elasticsearch concept, deprecated in 2.0: https://www.elastic.co/guide/en/elasticsearch/reference/2.1/breaking_20_mapping_changes.html#_fields_cannot_be_referenced_by_short_name

Thus, I have removed the functionality. The display name was updated, now only reflecting the name. In addition, I've updated the documentation to be more clear about the use of aliases - they must be defined in your elasticsearch index in order for the plugin to properly use the alias. It will not generate one for you.

Coincides with GeoTools PR: https://github.com/geotools/geotools/pull/2938

## Checklist

For all pull requests:


aaime:
I know in GeoServer this is just a community module, but it has history outside of it, and thus, supposedly, users. Is the migration transparent to those already having an existing installation?
Is a migration guide or a warning needed in the docs?

travisb-fe:
I think that is a fair point - I will update the documentation to indicate it will no longer be compatible and how to migrate existing data going forward. I hope it shouldn't affect too many users, as elasticsearch 1.x is **very** old at this point (2.0 released in 2015), but aliases effectively accomplish the same thing as short names but more directly, so that's an option.

sjudeng:
@travisb-fe Thanks for the updates and including docs. I thought the short name functionality in the plugin was just to enable the simplified attribute names on the GeoTools/GeoServer filtering side but internally the full name is always used in Elasticsearch queries. Can you confirm it attempts to use short names in Elasticsearch queries when configured?

travisb-fe:
That didn't seem to be the case when I use it. If I use the short names and plug them into a CQL filter, it just sends the request to elastic using that short name (it doesn't refer back to the full name from Elasticsearch). I assumed that was on purpose to emulate the short names feature from Elasticsearch. 

If I examine the request that goes by here: https://github.com/geotools/geotools/blob/master/modules/unsupported/elasticsearch/src/main/java/org/geotools/data/elasticsearch/RestElasticClient.java#L258

This `requestString` reveals a requestBody for the search query using the short name, not the full name. This makes sense given the results I get - always `null` for the short names, because elastic doesn't have it, but works fine with the full name.

So if it's supposed to use the full name internally, we can do that, but then I have a different bug to fix. I'll look into this a bit more and circle back.

sjudeng:
Thanks for running that down. If that's the case then it seems like it is a bug as it would fail with more recent Elasticsearch versions that no longer support queries by short names, right? And then I'd definitely agree it's a good idea to remove it.

travisb-fe:
Yeah. I think there's a way to accomplish what you mentioned probably via the AttributeDescriptor, but I couldn't quickly figure it out. I would be in favor of just removing this for now.

We could add the short name functionality you described in a future ticket perhaps, but I would also wonder if there is another name we could call it to not confuse it with the Elasticsearch concept (which I got confused for).

doc/en/user/source/community/elasticsearch/index.rst
```diff
@@ -170,6 +168,8 @@ The initial layer configuration panel for an Elasticsearch layer will include an
 
 To return to the field table after it has been closed, click the "Configure Elasticsearch fields" button below the "Feature Type Details" panel on the layer configuration page.
 
+The ``Short Names`` feature has been removed as it is not compatible with Elasticsearch 2.0 and beyond. Previous fields which used the short names will be reverted to the full name, but you can still use aliasing to accomplish the same effect.
```
sjudeng, line 171: Consider moving this note down to the [faq](https://github.com/geoserver/geoserver/blob/8678a31902381bf13ec7bb3808a4d1f4985301e8/doc/en/user/source/community/elasticsearch/index.rst#id1) section where there are other caveats and upgrade notes. Maybe preface with something like "when updating from ElasticGeo 2.16.0 note that ...".

travisb-fe: Sounds good. Done.

sjudeng:
@aaime Generally having a migration guide for going from ElasticGeo to GeoTools/GeoServer is a good idea. I can try and put that together at some point unless someone else does it first. I think this update is good without it though as @travisb-fe found that using the short names feature in the plugin with modern Elasticsearch versions currently results in no feature results and I think that behavior will be the same after these updates.

aaime:
All right, let's merge then.

PR #4257, opened 2020-05-21
[GEOS-9627] vector tiles - buffer around tile is too small

davidblasby:
This fixes an issue with oversampling (default is 16*) and the "extra" buffer around the tile to account for "wide" styles.  Before, it wasn't taking into account the oversampling, which meant the buffer was too small.

## Checklist

For all pull requests:

### DIFF
--- a/src/extension/vectortiles/src/main/java/org/geoserver/wms/vector/VectorTileMapOutputFormat.java
+++ b/src/extension/vectortiles/src/main/java/org/geoserver/wms/vector/VectorTileMapOutputFormat.java
@@ -114,13 +114,15 @@ public class VectorTileMapOutputFormat extends AbstractMapOutputFormat {
                                     paintArea,
                                     VectorMapRenderUtils.getMapScale(mapContent, renderingArea),
                                     (FeatureType) featureSource.getSchema()));
-            // buffer is in pixels (style pixels), need to convert to paint area pixels
-            buffer *=
-                    Math.max(
-                            Math.max(
-                                    this.tileBuilderFactory.getOversampleX(),
-                                    this.tileBuilderFactory.getOversampleY()),
-                            1); // if 0 (i.e. test case), don't expand
+            if (this.tileBuilderFactory.shouldOversampleScale()) {
+                // buffer is in pixels (style pixels), need to convert to paint area pixels
+                buffer *=
+                        Math.max(
+                                Math.max(
+                                        this.tileBuilderFactory.getOversampleX(),
+                                        this.tileBuilderFactory.getOversampleY()),
+                                1); // if 0 (i.e. test case), don't expand
+            }
             Pipeline pipeline =
                     getPipeline(mapContent, renderingArea, paintArea, sourceCrs, buffer);
 
--- a/src/extension/vectortiles/src/main/java/org/geoserver/wms/vector/VectorTileMapOutputFormat.java
+++ b/src/extension/vectortiles/src/main/java/org/geoserver/wms/vector/VectorTileMapOutputFormat.java
@@ -114,6 +114,13 @@ public class VectorTileMapOutputFormat extends AbstractMapOutputFormat {
                                     paintArea,
                                     VectorMapRenderUtils.getMapScale(mapContent, renderingArea),
                                     (FeatureType) featureSource.getSchema()));
+            // buffer is in pixels (style pixels), need to convert to paint area pixels
+            buffer *=
+                    Math.max(
+                            Math.max(
+                                    this.tileBuilderFactory.getOversampleX(),
+                                    this.tileBuilderFactory.getOversampleY()),
+                            1); // if 0 (i.e. test case), don't expand
             Pipeline pipeline =
                     getPipeline(mapContent, renderingArea, paintArea, sourceCrs, buffer);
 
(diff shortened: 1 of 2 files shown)


src/extension/vectortiles/src/main/java/org/geoserver/wms/vector/VectorTileMapOutputFormat.java
```diff
@@ -114,6 +114,13 @@ public WebMap produceMap(final WMSMapContent mapContent) throws ServiceException
                                     paintArea,
                                     VectorMapRenderUtils.getMapScale(mapContent, renderingArea),
                                     (FeatureType) featureSource.getSchema()));
+            // buffer is in pixels (style pixels), need to convert to paint area pixels
+            buffer *=
+                    Math.max(
+                            Math.max(
+                                    this.tileBuilderFactory.getOversampleX(),
+                                    this.tileBuilderFactory.getOversampleY()),
```
jodygarnett, line 122: Was getOverSampleX() an getOverSampleY() already part of the API?

There is some support for automatically determining the buffer from the style, or making use of the user suplied value (if the style is "too dynamic").

jodygarnett:
I do not know enough about this plugin to review efficiently.

src/extension/vectortiles/src/test/java/org/geoserver/wms/vector/VectorTileMapOutputFormatTest.java
```diff
@@ -201,6 +203,34 @@ public void testNoRulesByScale() throws Exception {
...
+        Rectangle renderingArea = new Rectangle(256, 256);
+
+        WMSMapContent mapContent = createMapContent(mapBounds, renderingArea, 32, pointLayer);
+
+        MapBoxTileBuilderFactory mbbf = new MapBoxTileBuilderFactory();
+        VectorTileMapOutputFormat vtof = new VectorTileMapOutputFormat(mbbf);
+
+        VectorTileMapOutputFormat vtof_spy = Mockito.spy(vtof);
+
+        // here's the test - the buffer (from style) is 32 bytes
```
aaime, line 219:
```suggestion
        // here's the test - the buffer (from style) is 32 pixels
```

src/extension/vectortiles/src/main/java/org/geoserver/wms/vector/VectorTileMapOutputFormat.java
aaime: There is actually a layer setting for the buffer, used when the style is too dynamic. This it's part of the layer metadata map, see usage here:
https://github.com/geoserver/geoserver/blob/b76bba7178d3505cd45e318e659c3b32faee4b5e/src/wms/src/main/java/org/geoserver/wms/GetMap.java#L808

aaime: Also, looking at other code, should this be guarded? E.g., quoting other bits using the oversamplign factors:

```
if (this.tileBuilderFactory.shouldOversampleScale()) {
if (this.tileBuilderFactory.shouldOversampleScale()) {
            paintArea =
                    new Rectangle(
                            this.tileBuilderFactory.getOversampleX() * mapWidth,
                            this.tileBuilderFactory.getOversampleY() * mapHeight);
        }
}
```

I'm also not familiar with this code, see the output format class has a oversampling factor, but the tile builder has two separate ones, not sure how related those are. "git blame" shows they have been coded by two different people, so it may well be they are not properly connected.

davidblasby: I've put the guard in, and fixed the byte/pixels typo.

I think the in-the-query-buffer is already being handled;
https://github.com/geoserver/geoserver/blob/master/src/extension/vectortiles/src/main/java/org/geoserver/wms/vector/PipelineBuilder.java#L298

(NOTE: this might want to be multiplied by the oversample factor, but I'm not sure how people are using it).

The oversample factor is hard coded to 16, which is pretty high.  Perhaps we should drop this to 2 or 4?

jodygarnett: Okay so oversampleX&Y are in part to support the hi-def displays, so a 512x512 "pixel" tile contains 4096x4096 detail (to support those 400 dpi phones do doubt).

So combining these two things: the layer buffer calculated from the style (or supplied by user override) x oversample should be correct.

Only remaining "huh" question is why there are different oversample X and Y...

aaime: I think the oversample factor was decided to support highres displays? I believe @travislbrundage did it, why not ask him directly? :-D

davidblasby:
I think the oversampling is a bit of a mess - however, it appears to be working (now).

Perhaps we can make a ticket to review it in the future.

I think this PR is mergeable...

aaime:
I also don't know enough about the module to have a firm opinion... @DBlasby you're listed as the maintainer of the module anyways, your call.

jodygarnett:
@davidblasby I noted you cannot merge this as we cleaned up commit access last year.

I will restore your commit access as module maintainer of the vector tile plugin.

aaime:
Checked with @jodygarnett , I'm merging this one for the time being.

aaime:
Squash-merged to get the commit with the right message (citing the ticket).

RoeGross:
[Vector Tiles](https://sourceforge.net/projects/geoserver/files/GeoServer/2.20.4/extensions/geoserver-2.20.4-vectortiles-plugin.zip)


EPSG_4326_15

{"type":"FeatureCollection","totalFeatures":"unknown","features":[]}

PR #4268, opened 2020-05-29
WMS-Capabilities-LayerGroup : Adding a default Style in capabilities …

julienperrot:
Hello,

It's a small add to WMS capabilities : 
As the GetLegendGraphic works for the layerGroups and some tools need the "Style" block, it add a standard block.

For all pull requests:

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
+++ b/src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java
@@ -12,6 +12,11 @@ import static org.geoserver.ows.util.ResponseUtils.appendQueryString;
 import static org.geoserver.ows.util.ResponseUtils.buildSchemaURL;
 import static org.geoserver.ows.util.ResponseUtils.buildURL;
 import static org.geoserver.ows.util.ResponseUtils.params;
+import static org.geoserver.wms.capabilities.CapabilityUtil.LAYER_GROUP_STYLE_ABSTRACT_PREFIX;
+import static org.geoserver.wms.capabilities.CapabilityUtil.LAYER_GROUP_STYLE_ABSTRACT_SUFFIX;
+import static org.geoserver.wms.capabilities.CapabilityUtil.LAYER_GROUP_STYLE_NAME;
+import static org.geoserver.wms.capabilities.CapabilityUtil.LAYER_GROUP_STYLE_TITLE_PREFIX;
+import static org.geoserver.wms.capabilities.CapabilityUtil.LAYER_GROUP_STYLE_TITLE_SUFFIX;
 import static org.geoserver.wms.capabilities.CapabilityUtil.validateLegendInfo;
 
 import com.google.common.collect.Iterables;
@@ -1056,8 +1061,11 @@ public class Capabilities_1_3_0_Transformer extends TransformerBase {
                 qatts.addAttribute(
                         "", "cascaded", "cascaded", "", String.valueOf(cascadedHopCount));
             }
+
+            String layerName = layer.prefixedName();
+
             start("Layer", qatts);
-            element("Name", layer.prefixedName());
+            element("Name", layerName);
             // REVISIT: this is bad, layer should have title and anbstract by itself
             element("Title", layer.getResource().getTitle());
             element("Abstract", layer.getResource().getAbstract());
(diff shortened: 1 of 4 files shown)


aaime:
The checklist is not optional, [I've clarified it in the template now](https://github.com/geoserver/geoserver/blob/master/.github/PULL_REQUEST_TEMPLATE.md).

I did not check the changes, but the PR lacks tests, so it does not qualify for merge. I'm afraid changing 3 files is beyond the small change. The template does not specify is too much, makes only an example, but the [contribution guidelines linked from the checklist](https://github.com/geoserver/geoserver/blob/master/CONTRIBUTING.md) are more specific:

> For a small change to a single source file a project committer can review and apply the change on your behalf. This is a quick workaround allowing us to correct spelling mistakes in the documentation, clarify a javadoc, or accept a very small fix.

julienperrot:
Hello,

I try to send the CLA but I received a "Recipient address rejected" from info@osgeo.org :-(
Do you have a solution ?

I'm currently writing some unit tests asscicated to the changes

Regards

jodygarnett:
> Recipient address rejected" from [info@osgeo.org](mailto:info@osgeo.org)

That is not good, can you try secretary@osgeo.org (and make a note about info@osgeo.org not working).

julienperrot:
Hello,

I went through the list of points and checked off as I went along.

The document of XXXX has been correctly sent and received by the OSGeo (confirmation by email on June 5th).

The two points that remain unchecked concern the associated JIRA ticket, which to my knowledge there is none.

I'll let you let me know what you get back!

Thank you

Sincerely

aaime:
I confirm reception of the CLA.

aaime:
The tools in question are broken btw, the XML schema for WMS capabilities states:

```
<element ref="wms:Style" minOccurs="0" maxOccurs="unbounded"/>
```

So it's an optional element. However, wms:name is not required to have chars in it, seems it could be empty:

```
<element name="Name" type="string">
<annotation>
<documentation> The Name is typically for machine-to-machine communication. </documentation>
</annotation>
</element>
```

So it seems it would not hurt... unless there are other types of broken tools out there requiring non empty content in case the element is present.

Personally I would not mind too much merging this one, but I'd vote to revert it if common/popular tools get in trouble with these changes (e.g., has anyone tried QGIS with this patch?).
@ianturton @jodygarnett what do you think?

aaime:
And oh, I've created the missing Jira ticket too: https://osgeo-org.atlassian.net/browse/GEOS-9659

julienperrot:
> So it's an optional element. However, wms:name is not required to have chars in it, seems it could be empty:

Hello,

I had first added a "default" name (like "default"), but I was afraid that some tools would not use the LegendURL block correctly, but would rebuild a URL from this "Name".
In this case, you have to leave the style parameter "empty" (or absent) for the URL to be meaningful.

So including an empty "Name" tag seemed to me to be the simplest and most backward compatible solution.

Thanks for the JIRA ticket!

Julien

aaime:
Merged and marked as resolved for 2.18-RC

ahuarte47:
> 
> I had first added a "default" name (like "default"), but I was afraid that some tools would not use the LegendURL block correctly, but would rebuild a URL from this "Name".
> In this case, you have to leave the style parameter "empty" (or absent) for the URL to be meaningful.
> 
> So including an empty "Name" tag seemed to me to be the simplest and most backward compatible solution.
> 
Hi @julienperrot, the INSPIRE validator (https://inspire.ec.europa.eu/validator/) throws errors for LayerGroups because the presence of [empty style names](https://github.com/geoserver/geoserver/blob/master/src/wms/src/main/java/org/geoserver/wms/capabilities/Capabilities_1_3_0_Transformer.java#L1213). What do you think about defining as default style-name the style-name of the first child layer? Can this be better?

aaime:
It would not make much sense. Does the INSPIRE validator mandate the usage of a non empty style? If not, I would just make the presence of the empty style vs no style at all configurable in the WMS panel.

ahuarte47:
The INSPIRE validator throws errors if the LayerGroup does not any style too, this PR adds one style, but the validator still fails because the style has no name.

aaime:
Ah, this is a problem, because we have no reserved style name. I believe GeoServer should also throw exceptions is anything but an empty style used in GetMap requests using layer groups, because it really does not make sense to have a style for groups right now (it's baked in the style configuration). This probably leaves the option of using an unlikely style name... "InspireBizzarreStyleNameForGroups"! (partially joking, the name should be more serious....).

pdorio:
This new feature creates several problems. I opened an issue on JIRA. Could someone explain me how to solve my problem?

https://osgeo-org.atlassian.net/browse/GEOS-9804

PR #4269, opened 2020-05-29
[GEOS-9635] Workspace settings: Enabling services does not enable them internally

oldschoolBavarianReferenceSystem:
Details can be found in the [bug description.](https://osgeo-org.atlassian.net/browse/GEOS-9635)

## Checklist

For all pull requests:

*Small enough to be not relevant I suppose.*
*If the PR is OK, I can open another one for the 2.17.x branch.*

*Sorry, no*
*N/A*
*N/A*

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/data/workspace/WorkspaceEditPage.java
+++ b/src/web/core/src/main/java/org/geoserver/web/data/workspace/WorkspaceEditPage.java
@@ -536,6 +536,7 @@ public class WorkspaceEditPage extends GeoServerSecuredPage {
             ServiceInfo global = gs.getService(serviceClass);
             OwsUtils.copy(global, newService, serviceClass);
             newService.setWorkspace(wsModel.getObject());
+            newService.setEnabled(true);
 
             // hack, but need id to be null so its considered unattached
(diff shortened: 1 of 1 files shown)


aaime:
Not sure I understand the issue... the UI should still allow you to change the configuration of the local workspace service and enable it (just follow the service name, it's a link). The checkboxes in the workspace page do not allow enable a workspace specific service, that is always available anyways, they just allow it to have a configuration that's separate form the global one.

oldschoolBavarianReferenceSystem:
Dear aaime,

you are absolutely right, this PR/Bug is nonesense! I'm closing it.

We really thought the checkbox before the service name in the workspace settings is for enabling/disabling that service within the workspace. It is *NOT*. Instead (as you said) it **triggers a side-effect**: When checked, a copy of the global configuration will be created after hitting "Save". When unchecked, that workspace-local copy will be dropped after hitting "Save". One tiny part of that configuration is then a switch to enable/disable the service.

I wonder if the UX and/or documentation of that part could be improved. I didn't expect that checkboxes have a side-effect. The help-text is pretty clear about what's happening, now that I'm convinced about how it works. Just: I didn't expect that to happen and first that text sounded a bit crazy/misleading to me.

PR #4270, opened 2020-05-29
[GEOT-6577] Allow plugin to handle multiple valid date formats from elastic

travisb-fe:
Updates the UI and documentation to reflect changes to the elasticsearch community module.

Coincides with GeoTools PR: https://github.com/geotools/geotools/pull/2953

## Checklist

For all pull requests:

### DIFF
--- a/src/community/elasticsearch/src/main/java/org/geoserver/elasticsearch/ElasticConfigurationPage.java
+++ b/src/community/elasticsearch/src/main/java/org/geoserver/elasticsearch/ElasticConfigurationPage.java
@@ -305,7 +305,16 @@ abstract class ElasticConfigurationPage extends Panel {
                                 return new Fragment(id, "empty", ElasticConfigurationPage.this);
                             }
                         } else if (property == ElasticAttributeProvider.DATE_FORMAT) {
-                            List<String> validFormats = att.getValidDateFormats();
+                            List<String> validFormats = null;
+                            if (att.getValidDateFormats() == null) {
+                                if (att.getDateFormat() != null) {
+                                    validFormats = new ArrayList<String>();
+                                    validFormats.add(att.getDateFormat());
+                                    att.setValidDateFormats(validFormats);
+                                }
+                            } else {
+                                validFormats = att.getValidDateFormats();
+                            }
                             if (validFormats != null) {
                                 Fragment f =
                                         new Fragment(id, "label", ElasticConfigurationPage.this);
--- a/src/community/elasticsearch/src/main/java/org/geoserver/elasticsearch/ElasticConfigurationPage.java
+++ b/src/community/elasticsearch/src/main/java/org/geoserver/elasticsearch/ElasticConfigurationPage.java
@@ -305,10 +305,19 @@ abstract class ElasticConfigurationPage extends Panel {
                                 return new Fragment(id, "empty", ElasticConfigurationPage.this);
                             }
                         } else if (property == ElasticAttributeProvider.DATE_FORMAT) {
-                            if (att.getDateFormat() != null) {
+                            List<String> validFormats = att.getValidDateFormats();
+                            if (validFormats != null) {
                                 Fragment f =
                                         new Fragment(id, "label", ElasticConfigurationPage.this);
-                                f.add(new Label("label", String.valueOf(att.getDateFormat())));
+                                String format = "";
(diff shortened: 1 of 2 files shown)


src/community/elasticsearch/src/main/java/org/geoserver/elasticsearch/ElasticConfigurationPage.java
```diff
@@ -326,10 +326,19 @@ protected Component getComponentForProperty(
...
                         } else if (property == ElasticAttributeProvider.DATE_FORMAT) {
-                            if (att.getDateFormat() != null) {
+                            List<String> validFormats = att.getValidDateFormats();
+                            if (validFormats != null) {
                                 Fragment f =
                                         new Fragment(id, "label", ElasticConfigurationPage.this);
-                                f.add(new Label("label", String.valueOf(att.getDateFormat())));
+                                String format = "";
+                                for (int i = 0; i < validFormats.size(); i++) {
+                                    if (i != valid_formats.size()) {
```
travisb-fe, line 335: Just noticed an error here, wrong variable name. Also working on backwards compatibility, so will have to update these things.

src/community/elasticsearch/src/main/java/org/geoserver/elasticsearch/ElasticConfigurationPage.java
```diff
@@ -305,10 +305,28 @@ protected Component getComponentForProperty(
...
+                            List<String> validFormats = null;
+                            if (att.getValidDateFormats() == null) {
+                                if (att.getDateFormat() != null) {
+                                    validFormats = new ArrayList<String>();
+                                    validFormats.add(att.getDateFormat());
+                                    att.setValidDateFormats(validFormats);
+                                }
+                            } else {
+                                validFormats = att.getValidDateFormats();
+                            }
```
travisb-fe, line 317: On the mailing list I discussed about using a Strategy pattern to interpret the date behaviour. However, due to how small the code sections are that use the date format, I found this way of handling it to be pretty straightforward, but I'm not sure it's the correct approach.

It does work that will be unnecessary for all future DataStores. Essentially this just "corrects" existing DataStores that were using the old date format. It could even be removed in a later release. Otherwise, I am not sure how old DataStores would ever get updated to use the new date format behaviour.

The Strategy option discussed would also work and be a permanent solution, but it wouldn't update old DataStores as far as I can tell. The old DataStores with the old date behaviour is technically incorrect and can fail to interpret Dates correctly. We could leave them alone and instruct users to recreate the DataStores with the new way of handling date formats (which would be handled via the Strategy, allowing both to exist simultaneously).

I am not sure which approach is better here, but hopefully I made the intent clear. If this way is not recommended, and using the Strategy is more proper, I'll update to that.

travisb-fe:
Finally pushed some changes to make it backwards compatible. I made a comment on the code in question since I am unsure what I've done makes sense.

travisb-fe:
Hi, just bumping this because the related GeoTools PR got merged (thank you @aaime )... but the GeoServer UI needs updated to handle it.

Back when I created this, running tests all passed locally, and I don't particularly understand the build errors being seen on this PR. To me, they look unrelated.

aaime:
@travisb-fe they are indeed likely unrelated. So we'll wait for you to upgrade the UI, right?

travisb-fe:
Yeah - that is what I am attempting to do here, although I don't know if it's particularly the "correct" way to do it, it seems to work when I've tested it.

aaime:
@travisb-fe what should we do about this PR? It's been over 6 months since the last update.

travisb-fe:
I can rebase off latest master to update it. This was working for me at the time, but I don't know if I've missed something.

aaime:
From previous conversation I though you still had to do some work on the UI, e.g.:

"@travisb-fe they are indeed likely unrelated. So we'll wait for you to upgrade the UI, right?"
"Yeah - that is what I am attempting to do here, although I don't know if it's particularly the "correct" way to do it, it seems to work when I've tested it."

But then I did not see a new commit in the PR, so thought you had to still push something that was working locally for you.
Probably just a misunderstanding.

travisb-fe:
Yeah, you are right, I miscommunicated there. My mistake.

I have updated and pushed to this branch again, I'll double check the CI passes, but otherwise I think it is fine to merge.

aaime:
Thanks, merged! :-D

PR #4279, opened 2020-06-02
Startup.bat: detect java from "Program Files"

jratike80:
See mailing list topic https://sourceforge.net/p/geoserver/mailman/message/37027845/.

Fix works for me with system java located at:
   C:\Program Files\AdoptOpenJDK\jdk-11.0.5.10-hotspot\bin\java.exe

Should be safe to backport into 2.17.1 release.


aaime:
Hmm... duh, wanted to change the title using a "squash and merge" approach, but something went wrong... oh well, will do it right in the backports. The ticket is here, for release notes: https://osgeo-org.atlassian.net/browse/GEOS-9641

PR #4284, opened 2020-06-04
[GEOS-9599](2.16.x bkpt) VerticalCRS reprojection (Vertical Grid Shift) on WPS Ras…

dromagnoli:
…ter Download


dromagnoli:
the Details of travis builds says passing although the main page says pending check :)
Merging it

PR #4291, opened 2020-06-08
[GEOS-9594] CSW: performance issue: every page loads entire catalog

NielsCharlier:
This change changes performances, and does not resolve a bug or add new functionality. That is why there is no new unit test. However, the existing unit tests are there to prove that this new implementation preserves functionality.
The change also simplifies the code.

## Checklist

For all pull requests:

### DIFF
--- a/src/extension/csw/core/src/main/java/org/geoserver/csw/store/internal/CatalogStoreFeatureCollection.java
+++ b/src/extension/csw/core/src/main/java/org/geoserver/csw/store/internal/CatalogStoreFeatureCollection.java
@@ -11,15 +11,21 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import org.geoserver.catalog.Catalog;
+import org.geoserver.catalog.PublishedInfo;
+import org.geoserver.catalog.PublishedType;
 import org.geoserver.csw.feature.AbstractFeatureCollection;
 import org.geoserver.csw.feature.MemoryFeatureCollection;
 import org.geoserver.csw.records.CSWRecordDescriptor;
 import org.geoserver.csw.records.RecordDescriptor;
 import org.geotools.data.store.FilteringFeatureCollection;
+import org.geotools.factory.CommonFactoryFinder;
 import org.geotools.feature.FeatureCollection;
+import org.geotools.filter.visitor.DuplicatingFilterVisitor;
 import org.opengis.feature.Feature;
 import org.opengis.feature.type.FeatureType;
 import org.opengis.filter.Filter;
+import org.opengis.filter.FilterFactory2;
+import org.opengis.filter.expression.PropertyName;
 import org.opengis.filter.sort.SortBy;
 
 /**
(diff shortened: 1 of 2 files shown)


aaime:
Are existing tests are covering this? Like, having enough records to trigger a paging, and going through it? (can't remember by heart)

NielsCharlier:
There definitely are. Also the combination of layers and layer groups etc

src/extension/csw/core/src/test/java/org/geoserver/csw/store/internal/LayerGroupTest.java
```diff
@@ -55,7 +55,7 @@ public void testRecords() throws Exception {
                 "csw?service=CSW&version=2.0.2&request=GetRecords&typeNames=csw:Record"
                         + "&resultType=results&elementSetName=full&maxRecords=100";
         Document d = getAsDOM(request);
-        // print(d);
+        print(d);
```
aaime, line 58: Please comment again.

src/extension/csw/core/src/main/java/org/geoserver/csw/store/internal/CatalogStoreFeatureCollection.java
```diff
@@ -90,4 +105,55 @@ protected void closeIterator(Iterator<Feature> close) {}
...
     }
+
+    private Filter catalogFilter() {
+
+        // ignore catalog info's without id
+        Filter result =
+                ff.and(filter, ff.not(ff.isNull(mapping.getIdentifierElement().getContent())));
+
+        // build filter compatible with layergroups and resources
+        result =
```
aaime, line 116: Can you split this code a bit, so that it does not end up being nested so much? Creating a separate inner class would probably do the trick

src/extension/csw/core/src/main/java/org/geoserver/csw/store/internal/CatalogStoreFeatureCollection.java
```diff
@@ -90,4 +105,55 @@ protected void closeIterator(Iterator<Feature> close) {}
...
+                                                                                        .getNamespaceContext());
+                                                            }
+                                                        },
+                                                        null))));
+        return result;
+    }
+
+    @Override
+    public int size() {
+        return Math.min(
```
aaime, line 153: Same as above, split code a bit.

src/extension/csw/core/src/main/java/org/geoserver/csw/store/internal/CatalogStoreFeatureIterator.java
```diff
@@ -94,107 +79,46 @@ public CatalogStoreFeatureIterator(
...
         this.filter = filter;
         catalogFacade = catalog.getFacade();
         this.mapping = mapping;
 
-        Filter advertised = ff.equals(ff.property("advertised"), ff.literal(true));
-
-        layerIt =
-                catalogFacade.list(
-                        ResourceInfo.class, ff.and(filter, advertised), null, null, sortOrder);
+        it = catalogFacade.list(PublishedInfo.class, filter, offset, count, sortOrder);
```
aaime, line 87: I know it was already using the CatalogFacade but, this seems like a grave issue... dodging the catalog means no security is applied, anyone can see every layer regardless of what the security subsystem says? 
The catalog offers a similar list method, it's the one that should be used instead (and I believe that would also account for "advertised", along with workspace specific services and the like).

NielsCharlier: That is a good point. However, it seems, strangely enough, that the Catalog list method only allows a single SortOrder, while the facade method allows multiple. CSW requires multiple sorts. Fixing this security leak would therefore require a catalog API change. ouch :/

aaime: Hmm.. never mind, it seems that what you get there is a SecuredCatalogFacade, which is applying security. Never mind this one, but please check the other two feedbacks.

NielsCharlier: Ok. I see the securecatalog does have a list with multi-sortby method to support this. It should ideally be added to the catalog interface as well.

NielsCharlier:
@aaime The changes were made

aaime:
Thanks, merged!

PR #4319, opened 2020-06-19
[GEOS-9655] WMTS layer configuration page doesn't show all the native srs in getCapabilties document

taba90:
Jira ticket https://osgeo-org.atlassian.net/browse/GEOS-9655

This pull request makes possible to select from the find button in the layer configuration page, the other native srs when configuring a WMTS layer.
## Checklist

For all pull requests:

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
+++ b/src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
@@ -91,7 +91,7 @@ public class BasicResourceConfig extends ResourceConfigurationPanel {
         // for resources coming from WMS and WFS Datastore
         // then check if it has more than one SRS, then add FIND SRS button in Native SRS
         // which will allow user to set other SRS as Native SRS
-        List<String> otherSRS = getOtherSRSFromWFSOrWMSOrWMTS(model.getObject());
+        List<String> otherSRS = getOtherSRS(model.getObject());
         CRSPanel nativeCRS;
         if (otherSRS.isEmpty()) {
             // normal behavior for resoureces not belonging to WFS and WMS Store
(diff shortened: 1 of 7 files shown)


src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
```diff
@@ -360,11 +356,16 @@ private String getActualNativeSRSCode(ResourceInfo resourceInfo) {
         return "";
     }
 
-    private List<String> getOtherSRSFromWFSOrWMS(ResourceInfo resourceInfo) {
+    private List<String> getOtherSRSFromWFSOrWMSOrWMTS(ResourceInfo resourceInfo) {
```
aaime, line 359:
```suggestion
    /**
     * Returns a list of alternative SRS for resources that support multiple ones, e.g., WMS, WFS, WMTS cascaded layers
     */
    private List<String> getOtherSRS(ResourceInfo resourceInfo) {
```

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
```diff
@@ -360,11 +356,16 @@ private String getActualNativeSRSCode(ResourceInfo resourceInfo) {
...
         // first check if its WFS-NG
         List<String> otherSRS = DataStoreUtils.getOtherSRSFromWfsNg(resourceInfo);
```
aaime, line 361: I'd go and check if it's a FeatureTypeInfo here.

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
taba90: fixed

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
taba90: fixed

taba90:
thanks @aaime your feedbacks have been applied

PR #4324, opened 2020-06-22
[GEOS-9658] SecuredFeatureSource incorrectly logs when requesting Com…

taba90:
…plexFeatures with a sld style

Jira ticket https://osgeo-org.atlassian.net/browse/GEOS-9658
This pr fix a problem in SecuredFeatureSource with a logging message wrongly pointing to a security issue with datastore when requesting ComplexFeatures with a style associated to them.

## Checklist

For all pull requests:

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/decorators/SecuredFeatureSource.java
+++ b/src/main/src/main/java/org/geoserver/security/decorators/SecuredFeatureSource.java
@@ -89,7 +89,11 @@ public class SecuredFeatureSource<T extends FeatureType, F extends Feature>
                     ReTypingFeatureCollection retyped = new ReTypingFeatureCollection(sfc, target);
                     return (FeatureCollection) SecuredObjects.secure(retyped, policy);
                 } else {
-                    if (!readQuery.equals(Query.ALL)) {
+                    List<PropertyName> readProps = readQuery.getProperties();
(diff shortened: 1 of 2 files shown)


src/main/src/main/java/org/geoserver/security/decorators/SecuredFeatureSource.java
```diff
@@ -89,12 +89,14 @@ protected SecuredFeatureSource(FeatureSource<T, F> delegate, WrapperPolicy polic
                     ReTypingFeatureCollection retyped = new ReTypingFeatureCollection(sfc, target);
                     return (FeatureCollection) SecuredObjects.secure(retyped, policy);
                 } else {
-                    // complex feature store eh? No way to fix it at least warn the admin
-                    LOGGER.log(
-                            Level.SEVERE,
-                            "Complex store returned more properties than allowed "
-                                    + "by security (because they are required by the schema). "
-                                    + "Either the security setup is broken or you have a security breach");
+                    if (!readQuery.equals(Query.ALL)) {
```
aaime, line 92: Ah ok, so this hides the log in case the security subsystem specifies no restrictions right?
However, the check seems to be too broad, what if the security subsystem was returning a filter to reduce the features, while keeping all the attributes? Better to check the readQuery for the property list, and see if all properties were allowed instead.

PR #4341, opened 2020-06-23
[GEOS-9657] Allow static method calls in GetFeatureInfo templates

awaterme:
PR enable the usage of static methods in GetFeatureInfo HTML templates as requested by [GEOS-9657](https://osgeo-org.atlassian.net/browse/GEOS-9657).

## Checklist

For all pull requests:

- [X] Commits changing the UI, existing user workflows, or adding new functionality, need to include documentation updates (screenshots, text)
- [X] Commits changing the REST API, or any configuration object, should check if the REST API docs (Swagger YAML files and classic documentation) need to be updated.

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/featureinfo/FreeMarkerTemplateManager.java
+++ b/src/wms/src/main/java/org/geoserver/wms/featureinfo/FreeMarkerTemplateManager.java
@@ -4,13 +4,26 @@
  */
 package org.geoserver.wms.featureinfo;
 
-import freemarker.template.*;
-import java.io.*;
+import static org.geoserver.wms.featureinfo.FreemarkerStaticsAccessRule.fromPattern;
+
+import freemarker.template.Configuration;
+import freemarker.template.SimpleHash;
+import freemarker.template.Template;
+import freemarker.template.TemplateException;
+import freemarker.template.TemplateHashModel;
+import freemarker.template.TemplateModel;
+import freemarker.template.TemplateModelException;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.OutputStreamWriter;
 import java.nio.charset.Charset;
 import java.util.List;
 import net.opengis.wfs.FeatureCollectionType;
+import org.apache.log4j.Logger;
 import org.geoserver.catalog.ResourceInfo;
 import org.geoserver.ows.Dispatcher;
+import org.geoserver.platform.GeoServerExtensions;
 import org.geoserver.platform.GeoServerResourceLoader;
 import org.geoserver.platform.ServiceException;
 import org.geoserver.template.DirectTemplateFeatureCollectionFactory;
@@ -19,6 +32,7 @@ import org.geoserver.template.GeoServerTemplateLoader;
 import org.geoserver.template.TemplateUtils;
 import org.geoserver.wms.GetFeatureInfoRequest;
 import org.geoserver.wms.WMS;
+import org.geoserver.wms.featureinfo.FreemarkerStaticsAccessRule.RuleItem;
 import org.geotools.feature.FeatureCollection;
 import org.opengis.feature.simple.SimpleFeatureType;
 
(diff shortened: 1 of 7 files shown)


src/wms/src/main/java/org/geoserver/wms/featureinfo/FreeMarkerTemplateManager.java
```diff
@@ -66,11 +85,43 @@ public TemplateModel wrap(Object object) throws TemplateModelException {
                                     "geoJSON",
                                     getStaticModel(
                                             "org.geoserver.wms.featureinfo.GeoJSONTemplateManager"));
+                            addConfiguredStatics(map);
```
aaime, line 88: This call is made every time a wrapping happens, meaning:

- The indications provided in system variables/env vars/context variables are processed over and over
- All the warning message are repeated for each request

Since the configuration is static by nature (should not change at runtime), all this should be processed once only in a static initializer that builds a map with the necessary configuration. This map is then going to be merged into the SimpleHash every time.

I can see how that would make testing harder, I guess you can add a protected static method "reset(String rulesSpec)" that changes the contents of that map, which the test will use.

src/wms/src/main/java/org/geoserver/wms/featureinfo/FreemarkerStaticsAccessRule.java
```diff
@@ -0,0 +1,135 @@
...
+            return alias;
+        }
+
+        public boolean isNumberedAlias() {
+            return numberedAlias;
+        }
+    }
+
+    /** Instance signals unrestricted access */
+    private static final FreemarkerStaticsAccessRule UNRESTRICTED =
```
aaime, line 55: Having UNRESTRICTED and DISABLED be both using the default constructor is confusing. I see how it's working, checking an equality in Unrestricted, but it's quite odd. Please change it to have a constructor taking a boolean, the unrestricted flag, instead (so unrestricted would be ``new FreemarkerStaticsAccessRule(true)`` and disabled would be ``new FreemarkerStaticsAccessRule(false)``.

src/wms/src/main/java/org/geoserver/wms/featureinfo/FreeMarkerTemplateManager.java
awaterme: Yes, you are right, I will adjust it.

src/wms/src/main/java/org/geoserver/wms/featureinfo/FreemarkerStaticsAccessRule.java
awaterme: Yes, you are right, I will adjust it.

awaterme:
Hi Andrea,
thank you very much for reviewing!
I have adjusted the pull request as you suggested.
With best regards,
Andreas

aaime:
The Travis build failed, in particular, in the QA checks:

```
[ERROR] Failed to execute goal com.coveo:fmt-maven-plugin:2.4.0:check (default) on project gs-wms: Found 3 non-complying files, failing build -> [Help 1]
```

It means changes were made to the code, but a local maven build was not run after them (that also reformats the files to match the coding conventions). The checklist above also links to [how to run the QA checks locally](https://docs.geoserver.org/latest/en/developer/qa-guide/index.html), I'd do it at least on the gs-wms module (running on the entire codebase takes quite a bit of time).... you never know, after formatting is fixed the other tests might be finding something else.

aaime:
The other changes look good, once it passes QA checks it's good to merge.

awaterme:
Hi André,

seems I missed the maven checks on the second commit. Sorry! It's fixed now.

Many thanks once again,
Andreas

PS: 
Most of the 1200+ tests of the WMS module fail on my machine, when running maven. 
Eclipse does not even compile gs-main; complains about missing classes from indryia-jar, in tests and in src (indirectly referenced). On my machine the build is incredibly slow, Eclipse rebuilds the workspace again and again. I can not imagine how you work on a daily basis on such a big project. However I use a virtual machine, which is certainly part of the problem.

aaime:
Thanks, squash-merged onto the master branch (it's a single commit there).

About the dev environment, I have a Ryzen 1700x with 32GB RAM, and a SSD drive. The full GeoServer with extensions (over 100 modules) builds in around 7 minutes, on JDK 8, tests included. With full test pass of course. The command line I use is:

``mvn clean install -T6 -Prelease``

The Travis builds for example run on very limited hardware (2 vcpus I believe?) and take longer, on this PR it took 26 minutes to build on JDK8 (-T2 there of course) and 36 minutes on JDK 11 (yep, it's that much slower to build, believe the compiler is to blame). Still, all green, all tests passed, at least on Linux (Mac and Windows are somewhat more finicky, and we don't have devs actively trying to keep the build going on those platforms).

PR #4343, opened 2020-06-24
[MapML Community Module] Use global proxyBaseUrl setting for base URL…

prushforth:
… when set

In the MapML Community Module, I was not taking into account the Global Settings Proxy Base URL value, which the admin can use to set a base URL value to use that doesn't assume geoserver's context and port number i.e. how the outside world sees the service.  So, this value is conditionally used when set to a non-blank value.   

## Checklist

For all pull requests:

### DIFF
--- a/src/community/mapml/src/main/java/org/geoserver/mapml/MapMLController.java
+++ b/src/community/mapml/src/main/java/org/geoserver/mapml/MapMLController.java
@@ -50,6 +50,7 @@ import org.geoserver.mapml.xml.ProjType;
 import org.geoserver.mapml.xml.RelType;
 import org.geoserver.mapml.xml.Select;
 import org.geoserver.mapml.xml.UnitType;
+import org.geoserver.ows.URLMangler;
 import org.geoserver.ows.util.ResponseUtils;
 import org.geoserver.wms.WMS;
 import org.geotools.geometry.jts.ReferencedEnvelope;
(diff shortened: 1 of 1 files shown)


src/community/mapml/src/main/java/org/geoserver/mapml/MapMLController.java
```diff
@@ -137,7 +137,8 @@ public String Html(
         final String workspace;
         String styleName = style.orElse("");
         String imageFormat = format.orElse("image/png");
-        String baseUrl = ResponseUtils.baseURL(request);
+        String proxyBaseUrl = geoServer.getSettings().getProxyBaseUrl();
```
aaime, line 140: Please don't build URLs by hand, use ResponseUtils.buildURL, that will take care of applying the proxy base correctly (the configured base url is just one way it can be specified), as well as any other plugin that can modify the backlinks (e.g., parameter extractor, authkey and the like):

https://github.com/geoserver/geoserver/blob/835bd43c1e70bd3434537d40d4452fc8e6b607fe/src/ows/src/main/java/org/geoserver/ows/util/ResponseUtils.java#L353

prushforth: Hopefully the changes made address what you mean; the URL used is entirely built by ResponseUtils.buildUrl now.

src/community/mapml/src/main/java/org/geoserver/mapml/MapMLController.java
```diff
@@ -137,7 +138,13 @@ public String Html(
...
+        String baseUrl =
+                ResponseUtils.buildURL(
+                        proxyBaseUrl != null ? proxyBaseUrl : ResponseUtils.baseURL(request),
```
aaime, line 144: Should still just use the ResponseUtils.baseURL(request), the proxy base URL should not be extracted from the config and used in this role, will work in some simple cases, but don't think it will do in general (e..g, when the parameter extractor plugin is in use). That said, just a suggestion, this is a community plugin, the only requirement is not to break the compile phase of the build (tests are not run, that's a privilege for supported modules).

prushforth:
I will investigate the issue that gave rise to this request, which has since been closed.  I need to know exactly what the problem was before fixing it without understanding exactly what they encountered.  Thanks for your review and comments.  I may get back to you.

PR #4346, opened 2020-06-26
Added the Error Prone Compiler steps to IntelliJ Quickstart

vava23:
GeoServer uses Error Prone tool which is not compatible with Java 8 and causes a compile error. However, Java 8 is the preferrable version according to documentation.

The issue can be easily fixed by installing the Error Prone Compiler plugin. I propose updating the developer documentation to reflect this step.

## Checklist

For all pull requests:


aaime:
ErrorProne should not be active by default, it's part of the maven build only if "qa" system variable is set (e.g., "-Dqa" in the build).
Not sure why IntelliJ ends up using it on some machines, looks like an IntelliJ bug to me.

I don't have it enabled in my IntelliJ and would not recommend doing so, the GeoServer code base is massive, the last thing one needs is to slow down the compiler IMHO. Best to finish the work, and do the QA checks in a build (which include errorprone, but also PMD and Spotbugs) as the last step before committing code.

vava23:
>ErrorProne should not be active by default, it's part of the maven build only if "qa" system variable is set (e.g., "-Dqa" in the build).

I see. That's what I got right after importing the project:
изображение

>I don't have it enabled in my IntelliJ and would not recommend doing so, the GeoServer code base is massive, the last thing one needs is to slow down the compiler IMHO.

Fair enough. Sorry for the straightforward suggestion. That did the trick for me, but I've got not so much experience in Java so far. Maybe there are some more reasonable ways to fix the issue. Seems that switching to _errorprone8_ profile doesn't help, but I should try it once again.

aaime:
Maybe the text can be amended to state that one either:

- Disables the errorprone profile (anyone has a clue why some people get it on by default?)
- Adds the necessary plugins to IDEA, with a warning that the builds will be slower.

With those changes I'd be happy to merge the docs changes.

vava23:
Done that. Also moved the text under **Run GeoServer from Intellij** paragraph, as it seems to be just one of possible issues.

aaime:
Looks good thanks, merging.

PR #4353, opened 2020-06-30
[GEOS-9649] NullPointerException for geoJSON output format with Compl…

taba90:
…exFeatures if some feature misses geometry value

JIRA ticket https://osgeo-org.atlassian.net/browse/GEOS-9649

This pr fixes a NPE when requesting ComplexFeatures in a geoJSON format and the following condition are met:

- a geometry attribute is null for some of the features;
- the geometry attribute has minOccurs = 0 or encodeIfEmpty is set to false.

## Checklist

For all pull requests:

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/json/ComplexGeoJsonWriter.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/json/ComplexGeoJsonWriter.java
@@ -176,7 +176,7 @@ class ComplexGeoJsonWriter {
             }
             // store the attribute name and geometry value of the current feature
             geometryAttribute = feature.getProperty(geometryType.getName());
-            geometry = (Geometry) geometryAttribute.getValue();
+            geometry = geometryAttribute != null ? (Geometry) geometryAttribute.getValue() : null;
         } else {
             // this feature seems to not have a geometry, write the default axis order
             jsonWriter.setAxisOrder(CRS.AxisOrder.EAST_NORTH);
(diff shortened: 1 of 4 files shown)


src/wfs/src/main/java/org/geoserver/wfs/json/ComplexGeoJsonWriter.java
```diff
@@ -176,7 +176,7 @@ private Property encodeGeometry(Feature feature) {
             }
             // store the attribute name and geometry value of the current feature
             geometryAttribute = feature.getProperty(geometryType.getName());
-            geometry = (Geometry) geometryAttribute.getValue();
+            geometry = geometryAttribute != null ? (Geometry) geometryAttribute.getValue() : null;
```
nmco, line 179: I see, the geometry can be optional in the schema and we were always expecting something. That say, as per the [GeoJSON specification](https://tools.ietf.org/html/rfc7946), if the feature is unallocated it should be represented as a JSON NULL value, citing the specification relevant part:

> A Feature object has a member with the name "geometry".  The value
> of the geometry member SHALL be either a Geometry object as
> defined above or, in the case that the Feature is unlocated, a
> JSON null value.

That say, is worth mentioning, the a JOSN value is represented like [this](https://www.json.org/json-en.html):

```json
{
   "key":null
}
```

Can you double check @Marco in your test that not onyl the value of the geometry attribute is ``NULL`` but also that the key is indeed present?

taba90: fixed

taba90:
thanks for your review @nmco feedback applied

nmco:
Looks good thx @taba90, this can be bacported right away to 2.17.x and 2.16.x.

PR #4373, opened 2020-07-02
[GEOS-9555] WMS remote SLD fetching lacks a timeout, can block the GetMap request indefinitely

fernandor777:
This PR includes new timeout configurations for remote SLD fetching.

JIRA issue:
https://osgeo-org.atlassian.net/browse/GEOS-9555

## Checklist

For all pull requests:


src/wms/src/main/java/org/geoserver/wms/map/GetMapKvpRequestReader.java
```diff
@@ -764,6 +796,35 @@ private InputStream getHttpInputStream(URL styleUrl) throws IOException {
         }
     }
 
+    /** Executes the HTTP request with the max request time settings. */
+    private CloseableHttpResponse executeRequest(HttpCacheContext cacheContext, HttpGet httpget)
+            throws IOException, ClientProtocolException {
+        int hardTimeout = 60000;
+        String maxTimeStr =
+                GeoServerExtensions.getProperty(GETMAPKVPREQUESTREADER_MAXREQUEST_TIME_VARIABLE);
+        if (StringUtils.isNotEmpty(maxTimeStr)) {
```
nmco, line 805: This code is almost a copy paste of the code above, but in one you check for a ``NULL`` value and here you checked for a ``NULL`` and emptiness. I would create an helper method and avoid the duplication.

src/wms/src/main/java/org/geoserver/wms/map/GetMapKvpRequestReader.java
```diff
@@ -164,6 +179,23 @@ public void handleServiceChange(
                 (CacheConfiguration) wms.getRemoteResourcesCacheConfiguration().clone());
     }
 
+    private int getTimeoutMillis() {
+        String timeoutStr =
+                GeoServerExtensions.getProperty(GETMAPKVPREQUESTREADER_TIMEOUT_VARIABLE);
+        try {
+            if (timeoutStr != null) {
```
nmco, line 186: Worth checking for emptiness like you do bellow?

src/wms/src/main/java/org/geoserver/wms/map/GetMapKvpRequestReader.java
```diff
@@ -164,6 +179,23 @@ public void handleServiceChange(
...
+                int timeout = Integer.parseInt(timeoutStr);
+                return timeout;
+            }
+        } catch (NumberFormatException ex) {
+            LOGGER.log(
+                    Level.FINE,
+                    "Number format exception trying to parse the connection timeout global variable",
```
nmco, line 193: Basically here you are swallowing the exception and use the default value:

1. This should be warning, not a debug log message. 
2. You log message should be improved, it should log the value provided by the user and log the default value that will be used.
3, Your log message should have a better context, _onnection timeout_ is a bit too generic.

src/wms/src/main/java/org/geoserver/wms/map/GetMapKvpRequestReader.java
```diff
@@ -764,6 +796,35 @@ private InputStream getHttpInputStream(URL styleUrl) throws IOException {
...
+    private CloseableHttpResponse executeRequest(HttpCacheContext cacheContext, HttpGet httpget)
+            throws IOException, ClientProtocolException {
+        int hardTimeout = 60000;
+        String maxTimeStr =
+                GeoServerExtensions.getProperty(GETMAPKVPREQUESTREADER_MAXREQUEST_TIME_VARIABLE);
+        if (StringUtils.isNotEmpty(maxTimeStr)) {
+            try {
+                hardTimeout = Integer.parseInt(maxTimeStr);
+            } catch (NumberFormatException ex) {
+                LOGGER.log(
```
nmco, line 809: Same comments as  above.

src/wms/src/main/java/org/geoserver/wms/map/GetMapKvpRequestReader.java
```diff
@@ -764,6 +796,35 @@ private InputStream getHttpInputStream(URL styleUrl) throws IOException {
...
+                        Level.FINE,
+                        "Bad number format on mas request time variable, proceeding to use the default.",
+                        ex);
+            }
+        }
+        TimerTask task =
```
nmco, line 815: You should has some comment explaining what's going on here, I guess you use the time to implement the time-out?

nmco: Could [this](https://stackoverflow.com/questions/3000214/java-http-client-request-with-defined-timeout) be a better option?

src/wms/src/test/java/org/geoserver/wms/map/GetMapKvpRequestReaderTest.java
```diff
@@ -1111,4 +1126,121 @@ public void testTransparencyValueInInspireGetMapRequest() throws Exception {
...
+                    public void handle(com.sun.net.httpserver.HttpExchange t) throws IOException {
+                        try {
+                            t.sendResponseHeaders(200, 5000000000l);
+                            TimeUnit.SECONDS.sleep(4);
+                            OutputStream outputStream = t.getResponseBody();
+                            outputStream.write("This is a bad style".getBytes());
+                            outputStream.flush();
+                            outputStream.close();
+                        } catch (InterruptedException e) {
+                            e.printStackTrace();
```
nmco, line 1240: Left over?

src/wms/src/test/java/org/geoserver/wms/map/GetMapKvpRequestReaderTest.java
```diff
@@ -1111,4 +1126,121 @@ public void testTransparencyValueInInspireGetMapRequest() throws Exception {
...
+            GetMapRequest request = (GetMapRequest) reader.createRequest();
+            Instant startInstant = Instant.now();
+            try {
+                reader.setLaxStyleMatchAllowed(false);
+                request =
+                        (GetMapRequest)
+                                reader.read(request, parseKvp(kvp), caseInsensitiveKvp(kvp));
+                fail("The style looked up, 'ThisStyleDoesNotExists', should not have been found");
+            } catch (ServiceException e) {
+                // System.out.println(e);
```
nmco, line 1161: Left over?

doc/en/user/source/services/wms/global.rst
```diff
@@ -25,3 +25,13 @@ ENABLE_JSONP
 
 Can be ``true`` or ``false`` (defaults to: ``false``).
 When ``true`` the JSONP (text/javascript) output format is enabled.
+
+GETMAPKVPREQUESTREADER_TIMEOUT
+-------------
+
+Sets the max connection timeout time in milliseconds, to connect or await new socket packets in the KVP request reader.  The default is 30000.
+
+GETMAPKVPREQUESTREADER_MAX_REQUEST_TIME
```
nmco, line 34: I would try to improve the name, for example ``GET_MAP_KVP_REQUEST_READER_TIMEOUT_MS``, anyway, the name seems to imply that is a time-out specific WMS GetMAP requests encoded as KVP, but no the description? What about WMS requests encoded as POST (GeoServer supports this although this is not standard).

doc/en/user/source/services/wms/global.rst
```diff
@@ -25,3 +25,13 @@ ENABLE_JSONP
 
 Can be ``true`` or ``false`` (defaults to: ``false``).
 When ``true`` the JSONP (text/javascript) output format is enabled.
+
+GETMAPKVPREQUESTREADER_TIMEOUT
```
nmco, line 29: Same as bellow.

nmco: Would it be worth to allow this configuration \ limits to be set on GeoServer UI? On the global configuration page or WMS service  since this looks specific?

src/wms/src/main/java/org/geoserver/wms/map/GetMapKvpRequestReader.java
aaime: Yeah, the timeout should be defined at the http client level.

fernandor777: - The provided link is about the Connection Timeout.  This timer bound execution is for limiting the max whole-request time, for which there is not an available setting on HttpClient API. This approach follows a recommended strategy for aborting the request with a hard time limit from some technical articles about HttpClient usage.
- Connection and Socket data packet timeout configuration is provided on the ```GETMAPKVPREQUESTREADER_TIMEOUT``` global setting and sets the connection and socket timeout in the HttpClient API since it's available.

src/wms/src/main/java/org/geoserver/wms/map/GetMapKvpRequestReader.java
fernandor777: Makes sense, adding it.

src/wms/src/main/java/org/geoserver/wms/map/GetMapKvpRequestReader.java
fernandor777: Replacing this with a common task method

src/wms/src/main/java/org/geoserver/wms/map/GetMapKvpRequestReader.java
fernandor777: Changed with better logging on a common task method

src/wms/src/main/java/org/geoserver/wms/map/GetMapKvpRequestReader.java
fernandor777: Done

src/wms/src/test/java/org/geoserver/wms/map/GetMapKvpRequestReaderTest.java
fernandor777: Yes, thanks, removing

src/wms/src/test/java/org/geoserver/wms/map/GetMapKvpRequestReaderTest.java
fernandor777: Yes, removing

src/wms/src/test/java/org/geoserver/wms/map/GetMapKvpRequestReaderTest.java
```diff
@@ -9,20 +9,31 @@
 import static org.hamcrest.CoreMatchers.not;
 import static org.junit.Assert.assertThat;
 
+import com.sun.net.httpserver.HttpHandler;
+import com.sun.net.httpserver.HttpServer;
```
aaime, line 13: Hmmm... ``com.sun``? This is not a public API... however, it does seem to work even in JDK 11, so I guess it won't be a problem short term.
As a possible alternative, see wiremock with random ports: http://wiremock.org/docs/getting-started/
But I would not get crazy about it, it's just a test after all, probably good as is.

fernandor777:
Thanks for your review, I added the new settings as WMS service configurations on GeoServer UI as recommended.

aaime:
There is a valid PMD violation here, probably a leftover method:

```
[INFO] --- maven-pmd-plugin:3.11.0:check (default) @ gs-wms ---
[INFO] PMD Failure: org.geoserver.wms.map.GetMapKvpRequestReader:800 Rule:UnusedPrivateMethod Priority:3 Avoid unused private methods such as 'getIntGlobalVariable(String,int)'..
[INFO] PMD Failure: org.geoserver.wms.map.GetMapKvpRequestReader:800 Rule:UnusedPrivateMethod Priority:3 Avoid unused private methods such as 'getIntGlobalVariable(String,int)'..
```

fernandor777:
Indeed, QA issue solved, thanks!

aaime:
@jodygarnett and this one too :-D

PR #4415, opened 2020-07-14
GEOS-9425 fix the bug: sync the style inside existing workspaces, the slave will cannot see the sld file.

swikecc:
During my debug, I found loader (an object of GeoServerResourceLoader) will return the relative path when the style inside existing workspaces, and will start with workspaces. This will result in styleAbsolutePath.indexOf("workspaces")==0, but the code on line 5 will ignore this situation. So that, the result is when sync the style inside existing workspaces, the sld file will be synced to the root directory of styles. Eventually it will cause that the slave cannot see the sld file.


aaime:
Normally we ask for a test to be added, but this is a community module, so various QA elements do not apply: merging.
Please next time do not remove the checklist from the PR, it's part of the template and contributors are supposed to follow it.

aaime:
Ah, feel free to create backport PRs if you want the fix to show up in the next 2.17.x and 2.16.x releases.

swikecc:
ok，thanks

PR #4418, opened 2020-07-15
Fix invalid cast to LayerInfoImpl

colecloudtostreet:
When I try to upload layers to Geonode, I get a 500 response with this stack trace https://pastebin.com/yiLbJFTp

It's caused by a failed request from Geonode to Geoserver which gives us this error:

org.geoserver.security.decorators.SecuredLayerInfo cannot be cast to org.geoserver.catalog.impl.LayerInfoImpl

The full stack trace from the Geoserver log is in here https://pastebin.com/mq5BWFaf

I think this is because I reconfigured Geonode to use HTTPS. Maybe there's a deeper issue here but, just by looking at the hierarchy of superclasses/interfaces, it seems like no cast is needed here, we should just treat the object.

I'm having trouble building a WAR that contains all the plugins needed in my production environment, so I can't test this in the environment where I'm running into the issue. Would someone else be able to test passing a SecuredLayerInfo as `orig` here? Thanks

https://osgeo-org.atlassian.net/browse/GEO-283

## Checklist

For all pull requests:

### DIFF
--- a/src/extension/importer/rest/src/main/java/org/geoserver/importer/rest/ImportTaskController.java
+++ b/src/extension/importer/rest/src/main/java/org/geoserver/importer/rest/ImportTaskController.java
@@ -519,12 +519,12 @@ public class ImportTaskController extends ImportBaseController {
         l.setResource(resource);
         // @hack workaround OWSUtils bug - trying to copy null collections
         // why these are null in the first place is a different question
-        LayerInfoImpl impl = (LayerInfoImpl) orig.getLayer();
-        if (impl.getAuthorityURLs() == null) {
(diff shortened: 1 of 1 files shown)


aaime:
See the checklist, you need to write a test that would fail without the change in question. Do you know what GeoNode request has been made, can it be reproduced on a stand-alone GeoServer?

aaime:
Been a month since my last message, any feedback on it?

colecloudtostreet:
Hey, thanks for following up. I don't have time to look into this now but I'll get back to you when we try to enable HTTPS again.

PR #4444, opened 2020-07-31
[GEOS-9710] WFS-templating module should support env parametrization also for portions of xpath [GEOS-9714] Wfs-templating plugin silently fails if the bacwkard mapping path provided in filter is not valid

taba90:
This pr allows 
1. to use variable substitution also for portion of an xpath inside the xpath function in a template. eg. xpath(strConcat('anXpathPortion',env('xpatPortion2', 'xpathPortion2'));

2. fixes a silent failing of the wfs-templating module in resolving wrong json path against the template.

JIRA tickets
 https://osgeo-org.atlassian.net/browse/GEOS-9710
https://osgeo-org.atlassian.net/browse/GEOS-9714

## Checklist

For all pull requests:


src/community/wfs-templating/src/main/java/org/geoserver/wfstemplating/expressions/XpathFunction.java
```diff
@@ -0,0 +1,67 @@
...
+
+import org.geotools.filter.AttributeExpressionImpl;
+import org.geotools.filter.FunctionExpressionImpl;
+import org.geotools.filter.capability.FunctionNameImpl;
+import org.opengis.filter.capability.FunctionName;
+import org.opengis.filter.expression.ExpressionVisitor;
+import org.opengis.filter.expression.PropertyName;
+import org.xml.sax.helpers.NamespaceSupport;
+
+public class XpathFunction extends FunctionExpressionImpl implements PropertyName {
```
aaime, line 13: A class that's both a Function and a PropertyName is, I believe, something we haven't seen before. Seems pretty similar to "FilterFunction_property" down in Geotools, which also allows to dynamically reference a property, and which has received special treatment to work with the various visitors correctly.
Now, this is just an output format, so there is no data access and the many DuplicatingFilterVisitor subclasses being involved into it. If it actually adds something (namespaces for example?) then I would clearly document it as something specific to this module, that should not be used anywhere else, and point people to "FilterFunction_property" for general usage. As another option, the existing "FilterFunction_property" function could be expanded to support what's missing here.

taba90: @aaime thanks for your point I've updated the documentation pointing the FilterFunction_porperty

taba90:
Failings seems unralated to this pull request: build is succesfull locally and the changes that it brings are all contained in a community module, without any pom.xml modification.

PR #4449, opened 2020-08-07
Remove dependency on a wicket class from jdbcconfig.internal.ConfigDatabase

groldan:
Remove usage of wicket's `Strings` utility class from jdbcconfig's
internal code, so it the module can/could be used even if web-core is
excluded as a dependency.

## Checklist

For all pull requests:

### DIFF
--- a/src/community/jdbcconfig/pom.xml
+++ b/src/community/jdbcconfig/pom.xml
@@ -54,11 +54,6 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
     </dependency>
-    <dependency>
-      <!-- blazing fast LZW compression/decompression -->
-      <groupId>com.ning</groupId>
-      <artifactId>compress-lzf</artifactId>
-    </dependency>
 
     <!-- Testing -->
     <dependency>
--- a/src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
+++ b/src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
@@ -54,11 +54,11 @@ import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import javax.sql.DataSource;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.collections.Predicate;
-import org.apache.wicket.util.string.Strings;
 import org.geoserver.catalog.Catalog;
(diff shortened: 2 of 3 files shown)


src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/internal/ConfigDatabase.java
```diff
@@ -677,8 +677,8 @@ private Info lookUpRelatedObject(
         String[] steps = localPropertyName.split("\\.");
         // Step back through ancestor property references If starting at a.b.c.d, then look at
         // a.b.c, then a.b, then a
-        for (int i = steps.length - 1; i >= 0; i--) {
-            String backPropName = Strings.join(".", Arrays.copyOfRange(steps, 0, i));
+        for (int len = steps.length; len > 0; len--) {
```
sikeoka, line 680: Is it correct to remove the " - 1" here?  If localPropertyName is "a.b.c.d", then the backPropName values should be "a.b.c", "a.b" and "a".

groldan: Hey, as far as I can see, it's doing exactly the same than before:
Instead of
`String backPropName = Strings.join(".", Arrays.copyOfRange(steps, 0, i));`
it does:
`String backPropName = Arrays.stream(steps).limit(len).collect(Collectors.joining("."));`
The key being `Stream.limit()`, which for the left-most step must be `1`, not `0`.

sikeoka: Running the old code on "a.b.c.d" would get [ "a.b.c", "a.b", "a", "" ] while your code gets [ "a.b.c.d", "a.b.c", "a.b", "a" ].  The empty string in the old code is useless and shouldn't be there.  Adding back the "- 1" to your changes gets the desired result of [ "a.b.c", "a.b", "a" ].

src/community/jdbcconfig/pom.xml
```diff
@@ -54,11 +54,6 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
     </dependency>
-    <dependency>
-      <!-- blazing fast LZW compression/decompression -->
-      <groupId>com.ning</groupId>
-      <artifactId>compress-lzf</artifactId>
-    </dependency>
```
sikeoka, line 61: This isn't really relevant to this PR but if you are going to remove the dependency in the pom, you should also remove it from the release assembly:
https://github.com/geoserver/geoserver/blob/master/src/community/release/ext-jdbcconfig.xml#L13

groldan: good catch. Amended the commits to have both changes.

PR #4451, opened 2020-08-14
[GEOS-9716] Add simple output formats support for complex features

fernandor777:
This PR adds Simple features output formats support for App-Schema complex features, by adding a configuration (disabled by default) on WFS service and Layer configuration.

This functionality will allow GeoServer to convert complex features to Simple feature for output formats that only work with simple features like:
- CSV
- KML
- SHAPE-ZIP

Ticket:
https://osgeo-org.atlassian.net/browse/GEOS-9716

## Checklist

For all pull requests:


src/main/src/main/java/org/geoserver/catalog/impl/ResourceInfoImpl.java
```diff
@@ -77,6 +77,8 @@
 
     protected List<String> disabledServices = new ArrayList<>();
 
+    protected Boolean activateComplexToSimpleOutput = false;
```
aaime, line 80: This is a mouthful. I'd go with "simpleConversionEnabled"

src/main/src/main/java/org/geoserver/catalog/impl/ResourceInfoImpl.java
```diff
@@ -383,6 +385,16 @@ public void setDisabledServices(List<String> disabledServices) {
         this.disabledServices = disabledServices;
     }
 
+    @Override
+    public Boolean getActivateComplexToSimpleOutput() {
```
aaime, line 389: It's a getter for a boolean, should use "is" instead of "get" as the prefix.

src/wfs/src/main/java/org/geoserver/wfs/response/ShapeZipOutputFormat.java
```diff
@@ -83,7 +83,7 @@
  * @author ported to gs 1.6.x by Saul Farber, MassGIS, saul.farber@state.ma.us
  */
 public class ShapeZipOutputFormat extends WFSGetFeatureOutputFormat
-        implements ApplicationContextAware {
+        implements ApplicationContextAware, ComplexFeatureAwareFormat {
```
aaime, line 86: The format is not really "aware" on its own here... it's more like "ComplexFeatureUnware" really, and needs other code to help it with conversion from simple to complex.

Maybe you should be marking GML3OutputFormat and GeoJSONOutputFormat as ComplexFeatureAware instead, and the XSLT and OGR output formats, with a default implementation of supportsComplexFeatures that returns true all the time.

src/extension/app-schema/app-schema-web/pom.xml
```diff
@@ -0,0 +1,198 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.geoserver.extension</groupId>
+        <artifactId>gs-app-schema</artifactId>
+        <version>2.18-SNAPSHOT</version>
+    </parent>
+
+    <groupId>org.geoserver.extension</groupId>
+    <artifactId>gs-app-schema-web</artifactId>
```
aaime, line 19: This name is configuring. Normally modules with "web" inside have UI elements and can be removed if the Wicket GUI is tossed away. This one instead seems to contain the complex to simple conversion machinery... should have been named ``app-schema-core`` instead.

src/extension/app-schema/app-schema-web/src/main/java/org/geoserver/complex/ComplexToSimpleFeatureCollection.java
```diff
@@ -0,0 +1,216 @@
...
+        /**
+         * Transform the original complex feature into a simple feature, using the convention and
+         * rules. Returns the resulting simple feature.
+         */
+        private SimpleFeature convert(Feature feature) {
+            SimpleFeatureBuilder builder = new SimpleFeatureBuilder(featureType);
+            for (PropertyDescriptor descriptor : featureType.getDescriptors()) {
+                String simpleName = descriptor.getName().getLocalPart();
+                Object attributeValue = getComplexAttributeValue(simpleName, feature);
+                builder.set(descriptor.getName(), attributeValue);
```
aaime, line 183: So here the full, qualified name is used, however, in getComplexAttributeValue, the lookup is made based on the simple name. Confusing?

src/extension/app-schema/app-schema-web/src/main/java/org/geoserver/complex/ComplexToSimpleOutputCommons.java
```diff
@@ -0,0 +1,28 @@
...
+ * application directory.
+ */
+package org.geoserver.complex;
+
+import java.util.List;
+import javax.xml.namespace.QName;
+import org.geoserver.ows.Request;
+
+/** Common constants and functions for complex to simple features conversion. */
+public class ComplexToSimpleOutputCommons {
```
aaime, line 12: Doesn't need to be public, does it?

src/extension/app-schema/app-schema-web/src/main/java/org/geoserver/complex/ComplexToSimpleFeatureCollection.java
```diff
@@ -0,0 +1,216 @@
...
+import org.opengis.util.ProgressListener;
+import org.xml.sax.helpers.NamespaceSupport;
+
+/**
+ * Feature collection Wrapper. Converts complex features to simple features on the fly based on the
+ * Layer info rules.
+ */
+public class ComplexToSimpleFeatureCollection implements SimpleFeatureCollection {
+
+    private static final Logger LOG = Logging.getLogger(ComplexToSimpleFeatureCollection.class);
```
aaime, line 45: In GeoServer loggers are (almost) uniformly called ``LOGGER``

src/extension/app-schema/app-schema-web/src/main/java/org/geoserver/complex/ComplexToSimpleFeatureCollection.java
```diff
@@ -0,0 +1,216 @@
...
+            this.delegate = delegate;
+        }
+
+        public boolean hasNext() {
+            return delegate.hasNext();
+        }
+
+        public SimpleFeature next() throws NoSuchElementException {
+            SimpleFeature simpleFeature = convert(delegate.next());
+            LOG.finer(() -> "Converted simple feature: " + simpleFeature);
```
aaime, line 170: Careful here, it's true that the string concatenation does not happen unless needed, but the creation of the lambda ([which is, itself, also an object](https://www.tutorialspoint.com/are-lambda-expressions-objects-in-java#:~:text=Yes%2C%20any%20lambda%20expression%20is,it%20like%20any%20other%20object.), instantiated and carried around) is not avoided. For something happening for all returned features, it should be guarded.

src/extension/app-schema/app-schema-web/src/main/java/org/geoserver/complex/ComplexToSimpleOutputHandler.java
```diff
@@ -0,0 +1,94 @@
...
+    /**
+     * Builds the {@link NamespaceSupport} instance from the catalog available namespaces.
+     *
+     * @param catalog the GeoServer catalog
+     * @return the resulting instance
+     */
+    private NamespaceSupport buildNamespaceSupport(Catalog catalog) {
+        List<NamespaceInfo> namespaces = catalog.getNamespaces();
+        NamespaceSupport nsSupport = new NamespaceSupport();
+        for (NamespaceInfo namespaceInfo : namespaces) {
```
aaime, line 89: Just thinking out loud here... is this enough? Does one have a workspace for all secondary namespaces when setting up an app-schema mapping?

src/extension/app-schema/app-schema-web/src/main/java/org/geoserver/complex/FeatureTypeConverter.java
```diff
@@ -0,0 +1,187 @@
...
+import org.opengis.feature.type.FeatureType;
+import org.opengis.feature.type.GeometryDescriptor;
+import org.opengis.feature.type.Name;
+import org.opengis.feature.type.PropertyDescriptor;
+import org.xml.sax.helpers.NamespaceSupport;
+
+/**
+ * Converts a complex feature type to a simple one based on the provided transformation rules map.
+ */
+public class FeatureTypeConverter {
```
aaime, line 31: Guess also this class could have been package private. In general, I'd suggest to limit visibility to the minimum actually needed, makes easier to change classes/methods later, no backwards compatibility to preserve.

src/extension/app-schema/app-schema-web/src/main/java/org/geoserver/complex/OutputFormatsSupport.java
```diff
@@ -0,0 +1,42 @@
...
+import org.geoserver.wfs.response.ComplexFeatureAwareFormat;
+
+/**
+ * Provides high level function to check if the provided format is supported by complex to simple
+ * features conversion.
+ */
+public class OutputFormatsSupport {
+
+    /** Checks if the provided format is supported by complex to simple features conversion. */
+    public static boolean isSupported(String format, Object value, Operation operation) {
```
aaime, line 22: Single method used only in the dispatcher callback, why roll a separate class for it?

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/SimpleOutputTest.java
```diff
@@ -0,0 +1,276 @@
...
+import org.geotools.data.FeatureSource;
+import org.geotools.data.complex.AppSchemaDataAccess;
+import org.geotools.feature.NameImpl;
+import org.geotools.jdbc.JDBCFeatureStore;
+import org.junit.Test;
+import org.opengis.feature.type.Name;
+import org.w3c.dom.Document;
+
+/** Set of tests for simple output formats on complex features requests. */
+public class SimpleOutputTest extends StationsAppSchemaTestSupport {
```
aaime, line 32: Good! I would also check that GML3 encoding was unaffected (e..g, did not lose namespaces and the like).

doc/en/user/source/data/webadmin/layers.rst
```diff
@@ -297,6 +297,7 @@ Sets the WFS specific publishing parameters.
 
 * **Per-Request Feature Limit**—Sets the maximum number of features for a layer a WFS GetFeature operation should generate (regardless of the actual number of query hits)
 * **Maximum number of decimals**—Sets the maximum number of decimals in GML output.
+* **Activate complex to simple features conversion** - Enables the conversion of complex features to simple features, using only SF-0 (simple) attributes for compatible output formats like CSV, KML, SHAPE-ZIP.
```
aaime, line 300: One might wonder why this is even needed. I'd at least suggest to explain that it would come with some loss of information. Or it could be made even simpler, and just happen if the target output format is not able to handle complex features natively, the current alternative is a hard exception... what would be the harm?

fernandor777:
Thanks for your review @aaime , the requested changes makes sense and were pushed to the PR.

fernandor777:
Handled some build issues with the new module name.  All tests green on local.  
There is a problem with jar dependencies on Travis build, I'll try to restart the build.

src/main/src/main/java/org/geoserver/catalog/ResourceInfo.java
```diff
@@ -374,4 +374,16 @@
 
     /** Set the list of disabled services names for this resource */
     void setDisabledServices(List<String> disabledServices);
+
+    /**
+     * Flag that determines if complex features will be converted to simple feature for compatible
+     * output formats.
+     */
+    Boolean getSimpleConversionEnabled();
```
aaime, line 382: As per my previous comment, this is a boolean properly, should be using "is" instead of "get":

```suggestion
    Boolean isSimpleConversionEnabled();
```

doc/en/user/source/data/webadmin/layers.rst
```diff
@@ -297,6 +297,7 @@ Sets the WFS specific publishing parameters.
 
 * **Per-Request Feature Limit**—Sets the maximum number of features for a layer a WFS GetFeature operation should generate (regardless of the actual number of query hits)
 * **Maximum number of decimals**—Sets the maximum number of decimals in GML output.
+* **Activate complex to simple features conversion** - If the target output format does not handle complex features nativelty, this option enables the conversion of complex features to simple features, using only SF-0 (simple) attributes.  This means that nested features and multiple-value attributes will be omitted from the final result.
```
aaime, line 300: Typo (and while I was at it, expanded explanation:

```suggestion
* **Activate complex to simple features conversion** - If the target output format does not handle complex features natively, this option enables the conversion of complex features to simple features, using only SF-0 (simple) attributes.  This means that nested features and multiple-value attributes will be omitted from the final result, instead of throwing errors while generating the output. Output formats capable of handling complex features are not affected.
```

aaime (changes requested):

src/main/src/main/java/org/geoserver/catalog/ResourceInfo.java
fernandor777: Understood, fixed.

doc/en/user/source/data/webadmin/layers.rst
fernandor777: Fixed and added the expanded explanation, thanks.

aaime:
Merged, thanks

aaime:
@jodygarnett we could use this commit in the release

PR #4462, opened 2020-08-24
[GEOS-9719] GeoJSONPPIO should use global number of decimals setting

jnh5y:
Addresses the rounding issue for the GeoJSON WPS PPIO identified here:
https://gis.stackexchange.com/questions/272009/geoserver-increase-the-precision-of-the-json-return-of-wps-call

## Checklist

For all pull requests:


jnh5y:
@ianturton Thanks for responding with enough of a suggestion on the GIS Exchange to help me and my colleague find and fix this issue quickly!  (We weren't the reporters; just happy to knock out a small thing like this!)

jnh5y:
Ok, the unit tests I added rely on the global settings being constant during the test.  That means that the test cases can fail if they run concurrently.  Is there an example of sorting this out in other unit tests?

aaime:
We don't really run same module tests concurrently. Even with a parallel build, each module gets its own private JVM and runs tests sequentially inside of it.

jnh5y:
@aaime thanks.  Running the new test in IntelliJ worked fine.  Building from the command line with -T4 has been throwing errors.  

I'll try building without the -T locally.

aaime:
In general, or for this particular PR? I normally run with -T6 without troubles (on a machine with 8 physical cores, a Ryzen 1700x)

jnh5y:
It is happening on this PR with my new test.  I wonder if I'm missing something obvious in setting/resetting the global settings?

jodygarnett:
> It is happening on this PR with my new test. I wonder if I'm missing something obvious in setting/resetting the global settings?

Testing the output functions is what is important, rather than your ability to manage settings. Can you refactor so this functionality can be tested by a simple unit test?

src/extension/wps/wps-core/src/test/java/org/geoserver/wps/ppio/GeoJSONPPIOTest.java
```diff
@@ -0,0 +1,163 @@
...
+        int dec = global.getSettings().getNumDecimals();
+        global.getSettings().setNumDecimals(4);
+        getGeoServer().save(global);
+
+        ByteArrayOutputStream os2 = new ByteArrayOutputStream(1024);
+        new GeoJSONPPIO.Geometries(gs).encode(point, os2);
+        String output2 = os2.toString();
+        assertEquals(output2, "{\"type\":\"Point\",\"coordinates\":[1.1235,2]}");
+
+        global.getSettings().setNumDecimals(dec);
```
jodygarnett, line 81: I would usually handle this in a try/finally block for extra safety...

jnh5y: Even in a unit test?

jnh5y:
> > It is happening on this PR with my new test. I wonder if I'm missing something obvious in setting/resetting the global settings?
> 
> Testing the output functions is what is important, rather than your ability to manage settings. Can you refactor so this functionality can be tested by a simple unit test?

I didn't see an obvious way to do that.  In some sense, my change is about the 'system' working together.  My second commit addresses the situation in the unit test framework, so I think this is good to go!

src/extension/wps/wps-core/src/test/java/org/geoserver/wps/ppio/GeoJSONPPIOTest.java
jodygarnett: yeah, this was based on testing changes coordinate reference system configuration (so any failure produced so many other failures it was hard to track down.)

Keep in mind if you run into multi-threaded testing problems in the future.

jodygarnett:
@jnh5y can we squash this into a single commit, your second commit does not have the [GEOS-9719] in the commit message.

jnh5y:
@jodygarnett thanks for the review and approval!

PR #4479, opened 2020-09-09
[GEOS-9733]: Importer stop scanning the whole directory when looking …

dromagnoli:
…for supplemental files

## Checklist

For all pull requests:

- [] Commits changing the REST API, or any configuration object, should check if the REST API docs (Swagger YAML files and classic documentation) need to be updated.


src/extension/importer/core/src/main/java/org/geoserver/importer/DefaultSupplementalFileExtensionsProvider.java
```diff
@@ -0,0 +1,46 @@
...
+        implements SupplementalFileExtensionsProvider {
+    private Set<String> acceptedInputExtensions;
+    private Set<String> supplementalExtensions;
+    private Set<String> upperCaseSupplementalExtensions;
+
+    public DefaultSupplementalFileExtensionsProvider(
+            Set<String> acceptedInputExtensions, Set<String> supplementalExtensions) {
+        this.acceptedInputExtensions = Collections.unmodifiableSet(acceptedInputExtensions);
+        this.supplementalExtensions = Collections.unmodifiableSet(supplementalExtensions);
+        HashSet<String> upperCase = new HashSet<>();
```
aaime, line 23: Better practice, declare interface, create instance.
```suggestion
        Set<String> upperCase = new HashSet<>();
```

src/extension/importer/core/src/main/java/org/geoserver/importer/SpatialFile.java
```diff
@@ -99,36 +101,42 @@ public String apply(@Nullable StyleHandler input) {
 
         // getBaseName only gets the LAST extension so beware for .shp.aux.xml stuff
         final String baseName = getBaseName(file.getName());
+        final String baseExtension = getExtension(file.getName());
+        final String basePath = file.getParent();
+
+        // look for style files
```
aaime, line 107: I would add a comment explaining why it's done this way, on a folder with few files this approach is actually more work. However, it's not that much more work, while on a directory with lots of files, this is more efficient.

src/extension/importer/core/src/main/java/org/geoserver/importer/SpatialFile.java
```diff
@@ -99,36 +101,42 @@ public String apply(@Nullable StyleHandler input) {
...
-
-                String ext = f.getName().substring(baseName.length());
-                // once the basename is stripped, extension(s) should be present
-                if (ext.charAt(0) == '.') {
-                    if (".prj".equalsIgnoreCase(ext)) {
-                        prjFile = f;
-                    } else if (styleFile == null && styleExtensions.contains(ext.substring(1))) {
-                        // TODO: deal with multiple style files? for now we just grab the first
-                        styleFile = f;
+        // Looking for supplemental files
```
aaime, line 118: This actually changes the behavior, some files that were caught before will not be in the list anymore. However, in fact getSuppFiles is not used for anything active in the import, it's just reflected out in the generated JSON when using the REST API, but not used to drive actual import, or to move files, or anything like that. So in the end, pretty invisible to the user, and not particularly useful (one may even just stop collecting them...).

aaime (approved):

src/extension/importer/core/src/main/java/org/geoserver/importer/DefaultSupplementalFileExtensionsProvider.java
dromagnoli: Fixed

src/extension/importer/core/src/main/java/org/geoserver/importer/SpatialFile.java
dromagnoli: Done

src/extension/importer/core/src/main/java/org/geoserver/importer/SpatialFile.java
dromagnoli: Right

PR #4496, opened 2020-09-17
Update GH Windows workflow

mprins:
Add some options to close TCP connection for Maven and update workflow items while here.

Windows builds have been failing for a while due to connection resets.
Apparently this is caused by Azure hanging up the connection after 4mins. without proper notification as described here: https://github.com/actions/virtual-environments/issues/1499#issuecomment-689467080

## Checklist

For all pull requests:

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -8,20 +8,23 @@ jobs:
     runs-on: [windows-latest]
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
+      with:
+        # 500 commits, set to 0 to get all
+        fetch-depth: 500
     - name: Set up JDK 1.8
       uses: actions/setup-java@v1
       with:
         java-version: 1.8
     - name: Maven repository caching
-      uses: actions/cache@v1.1.0
+      uses: actions/cache@v2
       with:
         path: ~/.m2/repository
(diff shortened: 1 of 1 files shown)


aaime:
Interesting find! Wondering if this would affect the osx builds too... are they running on Azure too? Hmm... nope, it seems not, this arcticle points at macstadium: https://docs.github.com/en/actions/reference/specifications-for-github-hosted-runners

mprins:
Yeah, but the Mac builds have been going strong, I'm not sure this relates to OSgeo-repo-failed builds, bid Windows has been having things like this: https://github.com/geoserver/geoserver/pull/4492#issuecomment-694191023 where downloading from Maven central fails

mprins:
Have restarted the windows build a few times, but I think once your have a populated cache for a branch it doesn't matter anymore anyway. In any case does not make things worse and gives us timestamped logging of the Maven build.

aaime:
@mprins the cache does not contain any SNAPSHOT dependency from GeoTools or GeoWebCache though, those are deleted before caching, and thus downloaded every time. So is this good to go, or do you have to work on it more?

mprins:
Nope, I'm done. Lets see how it goes. 

This is actually to fix problems with Maven central like https://github.com/geoserver/geoserver/pull/4504/checks?check_run_id=1159734676 and https://github.com/geoserver/geoserver/pull/4510/checks?check_run_id=1180646081

not snapshotted dependencies from the osgeo repo (though those may be affected too)

PR #4503, opened 2020-09-23
Changes representing three questions about the monitoring plugin.

jnh5y:
While trying out the monitor extension, I had some questions about the API/implementation.  This goal of this draft is to discuss any potential changes.

## Checklist

For all pull requests:

### DIFF
--- a/src/extension/monitor/core/src/main/java/org/geoserver/monitor/Monitor.java
+++ b/src/extension/monitor/core/src/main/java/org/geoserver/monitor/Monitor.java
@@ -73,7 +73,7 @@ public class Monitor implements ApplicationListener<ApplicationEvent> {
 
         // notify listeners
         for (RequestDataListener listener : listeners) {
-            listener.requestUpdated(req);
+            listener.requestStarted(req);
         }
         // have the DAO persist/propagate the change
         if (config.getMode() != Mode.HISTORY) {
--- a/src/extension/monitor/core/src/main/java/org/geoserver/monitor/MonitorFilter.java
+++ b/src/extension/monitor/core/src/main/java/org/geoserver/monitor/MonitorFilter.java
@@ -184,7 +184,7 @@ public class MonitorFilter implements GeoServerFilter {
             data.setResourcesProcessingTime(renderingTimeLayers);
             if (data.getEndTime() == null) data.setEndTime(new Date());
         }
-        monitor.update();
+        // monitor.update();
         data = monitor.current();
 
         monitor.complete();
--- a/src/extension/monitor/core/src/main/java/org/geoserver/monitor/ows/MonitorCallback.java
+++ b/src/extension/monitor/core/src/main/java/org/geoserver/monitor/ows/MonitorCallback.java
@@ -96,7 +96,7 @@ public class MonitorCallback implements DispatcherCallback {
             }
         }
 
-        monitor.update();
+        // monitor.update();
 
         return operation;
     }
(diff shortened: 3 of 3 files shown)


src/extension/monitor/core/src/main/java/org/geoserver/monitor/Monitor.java
```diff
@@ -73,7 +73,7 @@ public RequestData start() {
 
         // notify listeners
         for (RequestDataListener listener : listeners) {
-            listener.requestUpdated(req);
+            listener.requestStarted(req);
```
jnh5y, line 76: The 'requestStarted' method doesn't seem to be used anywhere else.  I'm recommending this change since the data provided this early in the request seems somewhat limited. 

This change would allow only those who want to see data at the very beginning of the MonitorFilter to get access to it.

src/extension/monitor/core/src/main/java/org/geoserver/monitor/MonitorFilter.java
```diff
@@ -184,7 +184,7 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
             data.setResourcesProcessingTime(renderingTimeLayers);
             if (data.getEndTime() == null) data.setEndTime(new Date());
         }
-        monitor.update();
+        // monitor.update();
```
jnh5y, line 187: I'm making a small assumption that any implementation which implements 'requestUpdated' would also implement 'requestedCompleted'.  

If that assumption is reasonable, then it is undesirable to call update() here since complete() is called two lines later.

src/extension/monitor/core/src/main/java/org/geoserver/monitor/ows/MonitorCallback.java
```diff
@@ -96,7 +96,7 @@ public Operation operationDispatched(Request request, Operation operation) {
             }
         }
 
-        monitor.update();
+        // monitor.update();
```
jnh5y, line 99: For the calls to monitor.update() in the MonitorCallback, I see them calls in MonitorFilter which call with basically the same information.  

I'm wondering if it makes sense to think of MonitorCallback as just setting data and rely on MonitorFilter for calls which would send data to the Monitor implementations?

src/extension/monitor/core/src/main/java/org/geoserver/monitor/Monitor.java
aaime: The RequestDataListener interface provides basic lifecycle hooks, similar to DispatcherCallback. Some are more informative than others, the reason why they are there is to allow following a request cycle. 
I don't understand why you'd want to break an existing extension point (not necessarily used only inside GeoServer), is there a significant overhead in this call? Please elaborate.

src/extension/monitor/core/src/main/java/org/geoserver/monitor/MonitorFilter.java
aaime: A lifecycle callback is meant to be generic, does not require one to implement specific methods, if you would be absurd to require so.

aaime:
Is there any more discussion to be had here?

src/extension/monitor/core/src/main/java/org/geoserver/monitor/Monitor.java
jnh5y: The overhead isn't necessarily significant (since that'd depend on the implementation).

I am asking about this since I was surprised by the number of calls to requestUpdated.  After I put the PR, I realized that as an implementer, I can ask some quick checks to skip over some of the calls which provide less information than I'd care to report/record.

jnh5y:
@aaime I don't think so.  Thanks for answering my questions!  I'll close this PR.

PR #4524, opened 2020-10-02
Docs: Clarifications to WMS cascading

jratike80:
- Inform users that WMS versions 1.1.1 and 1.3.0 are supported for cascading.
- Update the usage of URL field according to current behavior and add example URLs.


aaime:
Looks good, thanks, merged!

PR #4542, opened 2020-10-21
web.xml validation with web-app_3_1.xsd

jodygarnett:
Based on some research inspired by recent https://github.com/geoserver/geoserver/commit/08e0888e36840240915355ee6a0e24a5200ad665 I had a chance to look into updating our `web.xml` file.
 
## Checklist

For all pull requests:

- [ ] PR for bug fixes and small new features are presented as a single commit --> please use squash!

### DIFF
--- a/src/web/app/src/main/webapp/WEB-INF/web.xml
+++ b/src/web/app/src/main/webapp/WEB-INF/web.xml
@@ -1,6 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE web-app PUBLIC "-//Sun Microsystems, Inc.//DTD Web Application 2.3//EN" "http://java.sun.com/dtd/web-app_2_3.dtd">
-<web-app>
+<web-app xmlns="http://xmlns.jcp.org/xml/ns/javaee"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee
+                             http://xmlns.jcp.org/xml/ns/javaee/web-app_3_1.xsd"
+  version="3.1"
+  metadata-complete="true">
     <display-name>GeoServer</display-name>
   
       <context-param>
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -647,7 +647,7 @@
    <dependency>
     <groupId>xmlunit</groupId>
     <artifactId>xmlunit</artifactId>
-    <version>1.3</version>
+    <version>1.6</version>
    </dependency>
    <dependency>
     <groupId>org.easymock</groupId>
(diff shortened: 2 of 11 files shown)


jodygarnett:
Upgrading XMLUnit seems to be a trouble

jodygarnett:
The above validates! But is 15 seconds slow :( Even if it forced me to copy all the XSD and DTDs locally before validating web.xml

I had a look at upgrading to XMLUnit 2.x series but there were regressions.

src/web/app/src/main/webapp/WEB-INF/web.xml
```diff
@@ -1,6 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE web-app PUBLIC "-//Sun Microsystems, Inc.//DTD Web Application 2.3//EN" "http://java.sun.com/dtd/web-app_2_3.dtd">
-<web-app>
+<web-app xmlns="http://xmlns.jcp.org/xml/ns/javaee"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee
+                             http://xmlns.jcp.org/xml/ns/javaee/web-app_3_1.xsd"
+  version="3.1"
```
aaime, line 6: Did you check what Tomcat version is needed to support this? Does it need to be documented?
Seems it would work with Tomcat 8, but not anymore with Tomcat 7.
@randomorder you have some pulse on the deploys, is Tomcat 7 still common, or used at all?

mprins: Tomcat 7 is still supported by the Apache crew receiving regular patch releases, it has an announced EOL of March 31 2021: https://tomcat.apache.org/tomcat-70-eol.html dropping it now may be a tad early.

We still see some tomcat 7 deployments at client sites, though few.

randomorder: Hi @aaime  @mprins

We've been using Tomcat 8 for all new installations for a while now and upgraded tomcat 7 to 8 for all geoserver upgrades done for our clients because we had some issues with updating / upgrading GeoServer on tomcat 7 based deployments (bytecode processing libraries if I remember correctly).

Summarizing, it is not a big deal in my opinion because:

For fresh installations
Tomcat 9 is default from Ubuntu 20.04 LTS repos
Tomcat 8 and 9 from Ubuntu 18.04 LTS repos
Tomcat 8 and 9 on CentOS 7 (haven't looked at CentOS 8 but is safe to assume that it will be 8+)

For GeoServer updates / upgrades it is not unlikely that they'll have to update tomcat 7 anyways due to the bytecode processing library bug

aaime:
@jodygarnett what to do about this one? In order to get merged it would need a bit more work.

PR #4546, opened 2020-10-23
[GEOS-9773] Allow internal GWC to tile WMS requests that don't contain TILED=TRUE in the request parameters

ianturton:
If we have enabled the direct integration in geoserver we have to explicitly add TILED=True to requests sent from the client to have GWC kick in.
This makes it very difficult to introduce caching for clients which we do not control or legacy ones.

We would rather have an additional checkback in the "Caching Defaults" page right below Enable direct integration with GeoServer WMS that:
Has title "Explicitly require TILED Parameter"
By default is checked
If unchecked, in case clients do not send the TILED parameter they behave like it is sent and set to True.
This allows clients to explicitly unset it for certain requests by setting it to False.

## Checklist

For all pull requests:


ianturton:
Looks like the Mac Fail is unrelated

doc/en/user/source/geowebcache/webadmin/defaults.rst
```diff
@@ -31,6 +31,11 @@ With direct integration, tile caching is enabled for all standard WMS requests t
 
 This setting is disabled by default. When enabling this option, it is a good idea to also turn on :ref:`gwc_webadmin_diskquotas` as well, to prevent unbounded growth of the stored tiles.
 
+Explicitly require TILED Parameter
```
aaime, line 34: Nice to see docs! Maybe an update to the screenshot above would help too.

doc/en/user/source/geowebcache/webadmin/defaults.rst
```diff
@@ -31,6 +31,11 @@ With direct integration, tile caching is enabled for all standard WMS requests t
 
 This setting is disabled by default. When enabling this option, it is a good idea to also turn on :ref:`gwc_webadmin_diskquotas` as well, to prevent unbounded growth of the stored tiles.
 
+Explicitly require TILED Parameter
+``````````````````````````````````
+When this parameter is checked direct integration requires that the ``tiled=true`` parameter be set in all requests that will be cached. If this parameter is unchecked all incoming requests will be considered for caching, the request must still conform to all required parameters.
```
aaime, line 36:
```suggestion
When this parameter is checked direct WMS integration requires that the ``tiled=true`` parameter be set in all requests that will be cached. If this parameter is unchecked all incoming requests will be considered for caching, the request must still conform to all required parameters.

```

src/gwc/src/main/java/org/geoserver/gwc/config/GWCConfig.java
```diff
@@ -28,6 +28,8 @@
 
     private boolean directWMSIntegrationEnabled;
 
+    private boolean requireTiledParameter = true;
```
aaime, line 31: Wondering, is this object stored using XStream? I have vague memories that XStream dodges the normal construction mechanism and ends up ignoring field initializers too. I would give it a look.
Believe that normally we use Boolean object wrapper, and in the getter, if null, return the desired default value... at least, that's what we do for CatalogInfo objects :-D

src/gwc/src/test/java/org/geoserver/gwc/wms/CachingExtendedCapabilitiesProviderTest.java
```diff
@@ -122,6 +122,24 @@ public void testTileSets() throws Exception {
         assertXpathExists(tileSetPath + "[1]/Format", dom);
         assertXpathExists(tileSetPath + "[1]/Layers", dom);
         assertXpathExists(tileSetPath + "[1]/Styles", dom);
+
+        // Test RequireTiledParameter==false
+        GWC.get().getConfig().setDirectWMSIntegrationEnabled(true);
+        GWC.get().getConfig().setRequireTiledParameter(false);
+        dom = dom(get("wms?request=getCapabilities&version=1.1.1"), false);
+        print(dom);
```
aaime, line 130: Leftover print?

src/gwc/src/test/java/org/geoserver/gwc/wms/CachingExtendedCapabilitiesProviderTest.java
```diff
@@ -122,6 +122,24 @@ public void testTileSets() throws Exception {
         assertXpathExists(tileSetPath + "[1]/Format", dom);
         assertXpathExists(tileSetPath + "[1]/Layers", dom);
         assertXpathExists(tileSetPath + "[1]/Styles", dom);
+
+        // Test RequireTiledParameter==false
```
aaime, line 126: Confused, if the new setting affecting capabilities results at all?

src/web/gwc/src/main/resources/GeoServerApplication.properties
```diff
@@ -7,6 +7,8 @@ GWC.diskQuotaLoadFailed = Loading quota store failed, the disk quota subsystem i
 GWCSettingsPage.description                = Configure the global settings for the embedded GeoWebCache
 GWCSettingsPage.enableWMSIntegration       = Enable direct integration with GeoServer WMS
 GWCSettingsPage.enableWMSIntegration.title = Allows the regular WMS to serve cached content. When this option is enabled, use tiled=true for GetCapabilities in the GeoServer WMS to include the WMS-C VendorSpecificCapabilities
+GWCSettingsPage.requireTiledParameter      = Explicitly require TILED Parameter
+GWCSettingsPage.requireTiledParameter.title= When set to true, explicitly require TILED Parameter, otherwise accept all requests for tiling
```
aaime, line 11:
```suggestion
GWCSettingsPage.requireTiledParameter.title= When set to true, explicitly require TILED Parameter, otherwise consider all requests for tile caching
```

doc/en/user/source/geowebcache/webadmin/defaults.rst
ianturton: I have updated that image

src/gwc/src/test/java/org/geoserver/gwc/wms/CachingExtendedCapabilitiesProviderTest.java
ianturton: Direct Integration makes changes by adding in TileSets if tiled=true is sent, so I matched that. I suspect that no one actually uses that feature as it would need hand coding on the client side, but it seemed safest to match the existing behaviour

ianturton (commented):
So that's how to force it to be true on startup, it had been driving me nuts!

src/gwc/src/test/java/org/geoserver/gwc/wms/CachingExtendedCapabilitiesProviderTest.java
aaime: Oh, I see... hum, that grows the capablities document quite a bit though. I also believe that nobody is really using that feature, the idea here is to use the cache just because clients happen to align to the grid. @simboss what do you think?

aaime:
Manually checked it, appears to be working fine. Merging!

PR #4552, opened 2020-10-28
Fixing GEOMETRY_COLUMNS create table code snippit

Illutax:
The types "LINE", "COLLECTION", "MULTILINE", "GEOMETRY" are not supported.

Supported Simple types are:

-Point
-LineString
-CircularString
-CompoundCurve
-Polygon
-CurvePolygon

Collection types are:
-MultiPoint
-MultiLineString
-MultiPolygon
-GeometryCollection

Source: https://docs.microsoft.com/en-us/sql/relational-databases/spatial/spatial-data-types-overview?view=sql-server-ver15

## Checklist

For all pull requests:

### DIFF
--- a/doc/en/user/source/data/database/sqlserver.rst
+++ b/doc/en/user/source/data/database/sqlserver.rst
@@ -96,7 +96,7 @@ The table has the following structure (the table name is flexible, just specify 
 	   SRID INTEGER NOT NULL,
 	   TYPE VARCHAR(30) NOT NULL,
 	   UNIQUE(F_TABLE_SCHEMA, F_TABLE_NAME, F_GEOMETRY_COLUMN),
-	   CHECK(TYPE IN ('POINT', 'LINESTRING', 'CIRCULARSTRING', 'POLYGON', 'CURVEPOLYGON', 'MULTIPOINT', 'MULTILINESTRING', 'MULTIPOLYGON', 'GEOMETRYCOLLECTION') ));
+	   CHECK(TYPE IN ('POINT', 'LINESTRING', 'POLYGON', 'MULTIPOINT', 'MULTILINESTRING', 'MULTIPOLYGON', 'GEOMETRYCOLLECTION') ));
 
 When the table is present the store first searches it for information about each geometry column
 to be classified, and falls back on data inspection only if the table does not contain any information.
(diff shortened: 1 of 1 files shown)


mprins (changes requested):
I don't think this is correct; the values in this TYPE column are AFAIK the names of JTS geometry classes, not native database types.

See:  https://github.com/geotools/geotools/blob/5d5f62f4f3f409f506bd941ac877744b6dfe83cc/modules/library/main/src/main/java/org/geotools/geometry/jts/Geometries.java#L84-L110


/cc @aaime

Illutax:
> I don't think this is correct; the values in this TYPE column are AFAIK the names of JTS geometry classes, not native database types.
> 
> See: https://github.com/geotools/geotools/blob/5d5f62f4f3f409f506bd941ac877744b6dfe83cc/modules/library/main/src/main/java/org/geotools/geometry/jts/Geometries.java#L84-L110
> 
> /cc @aaime

I've encountered an problem when trying to add a layer containing a jts.MultiLineString. The GeoServer logged an ContractViolation in doing so.

aaime:
@mprins the geometry type lookup is a SQLServer specific issue, the code doing the lookup is here:
https://github.com/geotools/geotools/blob/029dc5c8ccc503de3cfd7f4d1c939d8056c260c4/modules/plugin/jdbc/jdbc-sqlserver/src/main/java/org/geotools/data/sqlserver/SQLServerDialect.java#L327

Once it has found the geometry type name as a string, it checks it against the TYPE_TO_CLASS_MAP defined here:
https://github.com/geotools/geotools/blob/029dc5c8ccc503de3cfd7f4d1c939d8056c260c4/modules/plugin/jdbc/jdbc-sqlserver/src/main/java/org/geotools/data/sqlserver/SQLServerDialect.java#L98

So, it seems it's incorrect to include curved geometry types in the list, they would not be recognized anyways. ``CIRCULARSTRING`` and ``CURVEPOLYGON`` should be removed, while ``MULTILINESTRING`` and ``GEOMETRYCOLLECTION`` are indeed missing. I see ``TYPE_TO_CLASS_MAP``  also "measured" variants, but I see no actual use for them (nothing special is done in the code for the "M"), adding them to the table definition list would just cause confusion I believe.

Illutax:
CLA was send.

aaime:
Thanks for the CLA. The PR is blocked on my feedback about entries, see https://github.com/geoserver/geoserver/pull/4552#issuecomment-720408433

mprins (approved):

PR #4553, opened 2020-10-30
[GEOS-9756] Layer configured with missing style throws NPE

ianturton:
remove broken styles from layers on load
## Checklist

For all pull requests:

### DIFF
--- a/src/main/src/main/java/org/geoserver/config/GeoServerLoader.java
+++ b/src/main/src/main/java/org/geoserver/config/GeoServerLoader.java
@@ -213,18 +213,25 @@ public abstract class GeoServerLoader {
 
                 LOGGER.info("Loaded layer '" + l.getName() + "'");
                 boolean badStyles = false;
+                if (null == l.getDefaultStyle().getStyle()) {
+                    LOGGER.log(
+                            Level.SEVERE,
+                            "Layer '" + l.getName() + "' references a missing style");
+                    badStyles = true;
+                    l.setDefaultStyle(catalog.getStyleByName("generic"));
+                }
                 for (StyleInfo style : l.getStyles()) {
                     if (null == style) {
                         LOGGER.log(
                                 Level.SEVERE,
                                 "Layer '" + l.getName() + "' references a missing style");
+                        badStyles = true;
                     }
-                    badStyles = true;
                 }
                 if (badStyles) {
                     l.getStyles().removeAll(Collections.singleton(null));
                 }
-                if (null == l.getDefaultStyle()) {
+                if (null == l.getDefaultStyle().getStyle()) {
                     l.setDefaultStyle(catalog.getStyleByName("generic"));
                 }
             } catch (Exception e) {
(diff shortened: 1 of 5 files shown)


ianturton:
I couldn't manage to work out how to test this, but I'm open to suggestions of examples that I should look at for tests

src/main/src/main/java/org/geoserver/config/GeoServerLoader.java
```diff
@@ -211,15 +211,18 @@ public void accept(LayerContents lc) {
...
                 LOGGER.info("Loaded layer '" + l.getName() + "'");
-
+                List<StyleInfo> badStyles = new ArrayList<>();
                 for (StyleInfo style : l.getStyles()) {
                     if (null == style) {
                         LOGGER.log(
                                 Level.SEVERE,
                                 "Layer '" + l.getName() + "' references a missing style");
                     }
+                    badStyles.add(style);
```
aaime, line 221: So this is collecting a list of nulls, which are all equals (actually all same) ... a boolean to track presence would do the same I guess. But I'm guessing it would not be same when removing them, removeAll should go an remove each one, otherwise if there are more than one, not all would be removed? :-D

ianturton: You're right they are all Null - I should have thought of that! As for what removeAll does with many Nulls? https://stackoverflow.com/questions/60417963/does-arraylist-removeallcollection-cares-about-duplicates seems to indicate removing a single Null should work - I'll simplfy

aaime:
To test this, let the GeoServer test run its course, get the location of the data dir, programmatically alter some layer files so that they are pointing to invalid style identifiers (maybe both default and one fake alternate style) and then force a reload. That should do the trick.

Looking at the PR, think I see one issue: the default style is stored in its own separate field, if null it should probably be set to a default style (CatalogBuilder.getDefaultStyle could be used for the purpose I believe).

aaime:
Waiting for a test.

ianturton:
Still looking at default styles - as you suspected it's tricky

aaime:
Test failure:

```
[ERROR] Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.639 s <<< FAILURE! - in org.geoserver.catalog.impl.CatalogProxiesTest
[ERROR] testDanglingReferenceOnModificationProxy(org.geoserver.catalog.impl.CatalogProxiesTest)  Time elapsed: 0.634 s  <<< FAILURE!
java.lang.AssertionError: expected null, but was:<StyleInfoImpl[generic]>
	at org.geoserver.catalog.impl.CatalogProxiesTest.testDanglingReferenceOnModificationProxy(CatalogProxiesTest.java:50)
```

I believe it's just testing for something that is not happening anymore, due to this change?

ianturton:
That is a very odd thing to test for and I've deleted it as it should never happen anymore

PR #4574, opened 2020-11-17
[2.18.x backport] [GEOS-9793] Update PostgreSQL (to 42.2.18) and Oracle (to 19.8.0.0) JDBC drivers

mprins:
Both of these have had some patch releases, so they should be updated.
```
com.oracle.database.jdbc:ojdbc8 ................. 19.7.0.0 -> 19.8.0.0
org.postgresql:postgresql ......................... 42.2.14 -> 42.2.18
```

backports #4573 to 2.18.x

resolves [GEOS-9793](https://osgeo-org.atlassian.net/browse/GEOS-9793)

see also https://github.com/geotools/geotools/pull/3208 (geotools 24.x PR)

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -1994,9 +1994,9 @@
   <jackson2.version>2.10.1</jackson2.version>
   <compress-lzf.version>1.0.3</compress-lzf.version>
   <marlin.version>0.9.3</marlin.version>
-  <postgresql.jdbc.version>42.2.14</postgresql.jdbc.version>
+  <postgresql.jdbc.version>42.2.18</postgresql.jdbc.version>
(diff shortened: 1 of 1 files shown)


aaime:
Not thrilled to see driver upgrades the day of the release, but I see GeoTools has been upgraded, having the two project mis-aligned could be worse than upgrading....

mprins:
I understand, I don't think it's a problem to leave this 'till the end of the week/after the release, the GeoServer build will just override these versions. (FWIW I've been using both of these versions without any problems)

I'm not sure why GeoServer doesn't rely on the transitive dependency provided by GeoTools (like it does for MS SQL), but that's for another time.

PR #4602, opened 2020-12-14
[GEOS-9828]: REST support ImageMosaic Harvesting COG datasets

dromagnoli:
Adding support for COG harvest on ImageMosaic using REST

## Checklist

For all pull requests:


src/community/cog/src/test/java/org/geoserver/rest/catalog/CogRemoteHarvestTest.java
```diff
@@ -0,0 +1,118 @@
...
+import org.custommonkey.xmlunit.XMLAssert;
+import org.custommonkey.xmlunit.XMLUnit;
+import org.geoserver.data.test.SystemTestData;
+import org.geoserver.rest.RestBaseController;
+import org.junit.Before;
+import org.junit.Test;
+import org.springframework.mock.web.MockHttpServletResponse;
+import org.w3c.dom.Document;
+
+public class CogRemoteHarvestTest extends CatalogRESTTestSupport {
```
aaime, line 25: This test talks to a remote resource that might not be available (e.g., machine not connected to internet), or the image in question might be eventually removed, or moved. Is it possible to make it an online test, so that it runs only when a  given profile is used, or at least check if the remote file is there?
Could be done with an Assume I suppose, and code similar to this to verify the remote file is at least reachable: https://github.com/geoserver/geoserver/blob/master/src/web/core/src/test/java/org/geoserver/web/GeoServerWicketOnlineTestSupport.java#L19
I was also considering calling this "CogRemoteHarvestOnlineTest", but unlike GeoTools, GeoServer does not have online test skipping, AFAIK.

src/restconfig/src/main/java/org/geoserver/rest/catalog/CoverageStoreFileController.java
```diff
@@ -453,4 +459,27 @@ protected Resource findPrimaryFile(Resource directory, String format) {
...
     }
+
+    /** Return the remote URL provided in the request. */
+    protected URL handleRemoteUrl(HttpServletRequest request) {
+
+        try {
+            // get the URL to be harvested
+            final String stringURL;
+            try (InputStream inStream = request.getInputStream()) {
+                stringURL = IOUtils.getStringFromStream(inStream);
```
aaime, line 470: A bit dangerous, there is no indication of the charset that might have been used.
Better to get the reader from the request instead of the stream:

```
   test = IOUtils.toString(request.getReader()); 
```

Using commons-io IOUtils.

src/restconfig/src/main/java/org/geoserver/rest/catalog/CoverageStoreFileController.java
```diff
@@ -453,4 +459,27 @@ protected Resource findPrimaryFile(Resource directory, String format) {
...
+        } catch (Throwable t) {
+            if (t instanceof RestException) {
+                throw (RestException) t;
+            } else {
+                throw new RestException(
+                        "Error while retrieving the remote URL:",
+                        HttpStatus.INTERNAL_SERVER_ERROR,
+                        t);
+            }
+        }
```
aaime, line 483:
```suggestion
        } catch (RestException t) {
           throw t;
        } catch (Throwable t) {
                throw new RestException(
                        "Error while retrieving the remote URL:",
                        HttpStatus.INTERNAL_SERVER_ERROR,
                        t);
        }
```

aaime (changes requested):

src/restconfig/src/main/java/org/geoserver/rest/catalog/CoverageStoreFileController.java
dromagnoli: Done

dromagnoli:
@aaime all feedbacks have been addressed

PR #4606, opened 2020-12-21
[GEOS-9834] GetLegendGraphic results in java.lang.NegativeArraySizeException or OutOfMemory

spike83:
According to my bug report [GEOS-9834](https://osgeo-org.atlassian.net/browse/GEOS-9834) I opened this pr to fix the scaling of the image.

## Checklist

For all pull requests:

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/legendgraphic/LegendGraphicBuilder.java
+++ b/src/wms/src/main/java/org/geoserver/wms/legendgraphic/LegendGraphicBuilder.java
@@ -78,6 +78,7 @@ public abstract class LegendGraphicBuilder {
      */
     protected final double MINIMUM_SYMBOL_SIZE = 3.0;
 
+    double dpiScaleFactor;
     protected int w;
     protected int h;
     boolean forceLabelsOn = false;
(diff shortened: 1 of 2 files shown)


src/wms/src/main/java/org/geoserver/wms/legendgraphic/LegendGraphicBuilder.java
```diff
@@ -153,8 +153,8 @@ protected Style resizeForDPI(GetLegendGraphicRequest request, Style gt2Style) {
         double standardDpi = RendererUtilities.getDpi(Collections.emptyMap());
         if (dpi != standardDpi) {
             double scaleFactor = dpi / standardDpi;
-            w = ((int) Math.round(w * scaleFactor));
-            h = ((int) Math.round(h * scaleFactor));
+            w = ((int) Math.round(request.getWidth() * scaleFactor));
+            h = ((int) Math.round(request.getHeight() * scaleFactor));
```
aaime, line 157: I am not familiar with this bit of code, but after review, the fix seems legit.
However, I'm not convinced it's the best way to solve this. The method is called over and over, once for each layer, the reason why we're having this problem it's that the code does two separate things:
* Resize the style based on the DPI, its intended purpose from the method name.
* As a side effect, also change the w and h fields... and if done over and over, well, that can cause very bad results indeed.

I would propose to move the side effect to another method that is called outside of the loop instead, e..g, ``resizeBoundsForDPI``. The call would be before the "for" loops in ``BufferedImageLegendGraphicBuilder`` and ``JSONLegendGraphicBuilder``. @ianturton you seem to be the original author of this method (as far as git goes at least), what do you think?

ianturton: I was just looking at that and thinking it looked familiar! 

Looks as if whatever I was working on only ever called that section of code once. So I would pull out the resize as @aaime suggests.

mprins:
I restarted the build as it locked up on downloading something from the central maven repo

aaime:
Looks good to me, merging! 
@spike83 if you want the fix also on the stable and maintenance branch, you'll have to make backports PRs.

PR #4627, opened 2020-12-23
[GEOS-9830] Allow the possibility to define allowedArea in a SRID different from 4326 in geofence [GEOS-9831] Possibility to use SpatialFilter in Geofence also to clip features

taba90:
This pull request
- adds srid handling on allowed area (geofence) https://osgeo-org.atlassian.net/browse/GEOS-9830
- adds the possibility to define clip spatial filter (geofence) https://osgeo-org.atlassian.net/browse/GEOS-9831
- updates geofence version https://osgeo-org.atlassian.net/browse/GEOS-9875

blocked by geotools: https://osgeo-org.atlassian.net/browse/GEOT-6769

## Checklist

For all pull requests:


src/main/src/main/java/org/geoserver/security/decorators/SecuredFeatureSource.java
```diff
@@ -104,12 +105,37 @@ protected SecuredFeatureSource(FeatureSource<T, F> delegate, WrapperPolicy polic
...
+    }
+
+    private FeatureCollection decoratesForClipping(
+            VectorAccessLimits limits, FeatureCollection collection) {
+        if (!(collection instanceof SimpleFeatureCollection))
+            throw new RuntimeException("Cannot clip on Complex Features");
+        Geometry clipFilter = limits.getClipVectorFilter();
+        Geometry intersectFilter = limits.getIntersectVectorFilter();
+        if (clipFilter != null) {
+            if (intersectFilter != null)
```
aaime, line 129: As noted in the GeoTools PR, this approach is inefficient. Better to check if there is a intersectFilter while working in ``getReadQuery`` and "and" an intersection filter there, so that it send down to the data source for efficient processing.

src/main/src/main/java/org/geoserver/security/VectorAccessLimits.java
```diff
@@ -166,6 +183,22 @@ private void writeProperties(List<PropertyName> attributes, ObjectOutputStream o
...
 
+    public Geometry getClipVectorFilter() {
+        return clipVectorFilter;
+    }
+
+    public void setClipVectorFilter(Geometry clipVectorFilter) {
+        this.clipVectorFilter = clipVectorFilter;
+    }
+
+    public Geometry getIntersectVectorFilter() {
```
aaime, line 194: Actually, since there is read filter, wondering if GeoFence cannot just build the full read filter by itself, rather than adding this field. This would also simplify SecuredFeatureSource.

src/extension/geofence/src/main/java/org/geoserver/geofence/GeofenceAccessManager.java
```diff
@@ -580,8 +661,12 @@ AccessLimits buildResourceAccessLimits(
         AccessLimits accessLimits = null;
         if (info instanceof FeatureTypeInfo) {
             // merge the area among the filters
-            if (reprojArea != null) {
-                Filter areaFilter = FF.intersects(FF.property(""), FF.literal(reprojArea));
+            if (intersectsArea != null) {
+                Filter areaFilter = FF.intersects(FF.property(""), FF.literal(intersectsArea));
```
aaime, line 665: Ok so it seems to be already here. Why does it need to be added as a separate field too?

taba90: thanks for your point @aaime : 
the scenario that the usage of ClipIntersectionFeatureCollection and the new field try to handle is one where we have two spatial filter one per type and is needed to put them in "OR". 
This happens when a user is inside two groups one with an intersects filter and the other with the clip. Geofence will return an accessInfo with both filter types. Simply delegating to the database the intersection and the applying the clip with the ClipFeatureCollection would return only the clipped geometries. 
In this case the intersects filter is still delegated to the database but the intersection is performed again to check wheter or not return feature according to the two spatial filters.

aaime: So in case you have both, you need to clip the geometry, but only if it does not also overlap with the intersect filter, in which case you return it fully?
That's bizzarre, but even if we wanted to have this functionality, it should be buried inside the GeoFence module, something like this does not belong in GeoTools.

aaime: Actually scratch that, it woudl have to be in gs-main... do we really need that behavior though?
If really needed, then the odd feature collection wrapper should be a package private class somewhere in gs-main. I mean, it's even hard to explain, we don't want to expose it.

taba90: Odd but... if a user belongs to two groups one with an intersects filter and the other with the clip filter and the two overlaps,  if the intersects doesn't win over the clip then we would have cases where the user is meant to have access to a geometry that he cannot actually sees.

src/main/src/main/java/org/geoserver/security/decorators/ClipIntersectsFeatureIterator.java
```diff
@@ -0,0 +1,184 @@
...
+                next.setAttribute(name, ((Geometry) next.getAttribute(name)).union(intersected));
+        }
+    }
+
+    private Map<Name, Geometry> getIntersectingGeometries(
+            Map<Name, Geometry> geometryAttributes, SimpleFeatureType type) {
+        Map<Name, Geometry> intersectedGeometries = new HashMap<>();
+        for (Name name : geometryAttributes.keySet()) {
+            Geometry geom = geometryAttributes.get(name);
+            if (geom instanceof GeometryCollection) {
```
aaime, line 126: This bit seems weird too. A multi-geometry is still cosidered "one", while here it's taken apart in smaller bits?
Say the geometry is Italy and all its big and small islands, if I undrestand correctly, only the bits of that large multipolygon intersecting the geometry filter would be returned? That's now how spatial databases work, it's surprising behavior to me.

aaime:
I've merged the GeoTools PR, will kick this one once the new jars are ready.

src/main/src/main/java/org/geoserver/security/decorators/ClipIntersectsFeatureIterator.java
taba90: fixed

taba90:
thanks for your review @aaime feedback on GeometryCollection intersects has been applied and the builds are now green

PR #4640, opened 2021-01-06
update conformance statements

bradh:
The OGC API specs are pretty fluid, but it looks (from https://github.com/opengeospatial/ogcapi-tiles/issues/20 and https://github.com/opengeospatial/ogcapi-tiles/pull/24 amongst other places) like the conformance classes have `conf` rather than `req`. This PR updates the implementation to refer to common correctly.

I also had to add a bit of pixel tolerance to one test.

## Checklist

For all pull requests:

- [X] This PR passes all existing unit tests (test results will be reported by Continuous Integration after opening this PR)
- [X] This PR passes the [QA checks](https://docs.geoserver.org/latest/en/developer/qa-guide/index.html) (QA checks results will be reported by Continuous Integration after opening this PR)

### DIFF
--- a/src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/api/ConformanceClass.java
+++ b/src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/api/ConformanceClass.java
@@ -1,0 +1,13 @@
+/* (c) 2021 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.api;
+
+/** Conformance statements for the conformance page. */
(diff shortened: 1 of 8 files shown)


aaime:
The published standard is still using "conf", I would not change the code until an actual release using "ref" is created:
http://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_response_3

bradh:
> The (one and only so far) published standard is still using "conf", I would not change the code until an actual release using "ref" is created:
> http://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_response_3

I think this is what I have changed it to. Was `req` now `conf`.

aaime:
LOL you're quite right. Merging.

PR #4643, opened 2021-01-09
[GEOS-6903] Unwanted dependency on HSQLDB in gs-web-wps

bradh:
Cleanup of old code (ticket picked at random).

See https://osgeo-org.atlassian.net/browse/GEOS-6903

Note that there is still a dependency on hsqldb in wps-core (because of https://github.com/geoserver/geoserver/blob/master/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/validator/MultiplicityValidator.java#L7). Happy to take hints on fixing that as a subsequent ticket - this only deals with web-wps.

## Checklist

For all pull requests:

- [N/A] New unit tests have been added covering the changes

### DIFF
--- a/src/extension/wps/web-wps/src/main/java/org/geoserver/wps/web/WPSExecuteTransformer.java
+++ b/src/extension/wps/web-wps/src/main/java/org/geoserver/wps/web/WPSExecuteTransformer.java
@@ -5,11 +5,14 @@
  */
 package org.geoserver.wps.web;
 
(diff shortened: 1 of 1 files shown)


mprins (approved):

aaime:
Well duh, that validator definitely has the wrong import :-D

bradh:
> Well duh, that validator definitely has the wrong import :-D

So just change it to `util.Collection`?

PR #4673, opened 2021-01-26
ogcapi: fix deprecated usage, unchecked warnings, minor typos

bradh:
Minor code cleanups:
 - Fixes title issue in queryables for Tiles (also added unit test coverage for this)
 - Updates some deprecated uses of `Stream.stream(Collection ...)`, and one use of `URLEncoder.encode(String)` (updated existing case too)
 - Switches a few raw types to parameterised version.
 - Fixes some documentation typos.
 - Removes a couple of loggers on abstract base classes (which didn't need them directly) and adds one lower level logger
 - Adds some final markers for methods called from constructors, and for a few private member variables.
 - Renames a couple of variables that shadowed member variables.
 - Removes an unneeded unboxing.
 - Updates the license header on tiles VectorLayerMetadata to be GS GPL (was GeoTools LGPL)

## Checklist

For all pull requests:

- [N/A] There is a ticket in Jira describing the issue/improvement/feature (a notable exemptions is, changes not visible to end users)
- [N/A] Commit message must be in the form "[GEOS-XYZW] Title of the Jira ticket" (export to XML in Jira generates the message in this exact form)
- [X] This PR passes all existing unit tests (test results will be reported by Continuous Integration after opening this PR)
- [X] This PR passes the [QA checks](https://docs.geoserver.org/latest/en/developer/qa-guide/index.html) (QA checks results will be reported by Continuous Integration after opening this PR)

### DIFF
--- a/src/community/ogcapi/ogcapi-tiles/src/main/java/org/geoserver/ogcapi/tiles/TiledCollectionDocument.java
+++ b/src/community/ogcapi/ogcapi-tiles/src/main/java/org/geoserver/ogcapi/tiles/TiledCollectionDocument.java
@@ -8,7 +8,6 @@ import static org.geoserver.ogcapi.tiles.TilesService.isStyleGroup;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
-import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.LinkedHashSet;
@@ -43,7 +42,7 @@ import org.springframework.http.HttpStatus;
 
 /** Description of a single collection, that will be serialized to JSON/XML/HTML */
 @JsonPropertyOrder({"id", "title", "description", "extent", "links", "styles"})
-public class TiledCollectionDocument extends AbstractCollectionDocument {
+public class TiledCollectionDocument extends AbstractCollectionDocument<TileLayer> {
     static final Logger LOGGER = Logging.getLogger(TiledCollectionDocument.class);
     public static final String DEFAULT_STYLE_NAME = "_";
     WMS wms;
@@ -61,7 +60,7 @@ public class TiledCollectionDocument extends AbstractCollectionDocument {
      *     TiledCollectionsDocument}, otherwise it's full and assumed to be the main response
      */
     public TiledCollectionDocument(WMS wms, TileLayer tileLayer, boolean summary)
-            throws FactoryException, TransformException, IOException {
+            throws FactoryException, TransformException {
         super(tileLayer);
         // basic info
         this.layer = tileLayer;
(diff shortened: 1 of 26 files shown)


src/community/ogcapi/ogcapi-tiles/src/main/java/org/geoserver/ogcapi/tiles/TiledCollectionDocument.java
```diff
@@ -43,7 +42,7 @@
 
 /** Description of a single collection, that will be serialized to JSON/XML/HTML */
 @JsonPropertyOrder({"id", "title", "description", "extent", "links", "styles"})
-public class TiledCollectionDocument extends AbstractCollectionDocument {
+public class TiledCollectionDocument extends AbstractCollectionDocument<TileLayer> {
```
aaime, line 45: Does it need to be public?

src/community/ogcapi/ogcapi-tiles/src/main/java/org/geoserver/ogcapi/tiles/VectorLayerMetadata.java
```diff
@@ -22,7 +10,7 @@
 
 /** Maps to the JSON structure describing a layer of vector tiles */
 @JsonInclude(JsonInclude.Include.NON_NULL)
-class VectorLayerMetadata {
+public class VectorLayerMetadata {
```
aaime, line 13: Does it need to be public?

aaime:
Looks pretty good overall, just wondering about a couple of visibility changes

src/community/ogcapi/ogcapi-tiles/src/main/java/org/geoserver/ogcapi/tiles/TiledCollectionDocument.java
bradh: Is this comment possibly against the wrong line?

src/community/ogcapi/ogcapi-tiles/src/main/java/org/geoserver/ogcapi/tiles/VectorLayerMetadata.java
bradh: Probably not, in that it worked before. I likely changed it in response to some IDE complaint.

It doesn't seem like something that should be private though, and maybe it might be useful to expose it for additional modules. 

Can revert it (and other visibility changes to `public`) if you'd prefer.

aaime: I try to keep classes package private unless I have a need for them to be public. Makes long term mainteinance easier to do, less exposed surface that downstream users might be depending onto.
That said, this is a community module, so anyone depending on it does at their own peril.
No need to revert.

src/community/ogcapi/ogcapi-tiles/src/main/java/org/geoserver/ogcapi/tiles/TiledCollectionDocument.java
aaime: Yep. Was wondering in general about changes of visibilty between package private and public. See comment below. No action required.

src/community/ogcapi/ogcapi-tiles/src/main/java/org/geoserver/ogcapi/tiles/VectorLayerMetadata.java
bradh: As it turns out, this is exported through public API e.g. https://github.com/geoserver/geoserver/blob/master/src/community/ogcapi/ogcapi-tiles/src/main/java/org/geoserver/ogcapi/tiles/TileJSON.java#L53-L57

Knew there was some reason for this...

PR #4705, opened 2021-02-05
use Charset definition in URL encoding

bradh:
This PR changes uses of URLDecode and URLEncode that previously used String values to identify the encoding. That results in needing to handle an UnsupportedEncodingException that can never happen (because UTF-8 and ASCII are always supported). The cleaner way is to use a StandardCharsets value, which is more obvious and avoids the exception.

## Checklist

For all pull requests:

- [X] Make sure the first PR targets the master branch, eventual backports will be managed later. This can be ignored if the PR is fixing an issue that only happens in a specific branch, but not in newer ones. [No backports on this]

- [N/A] There is a ticket in Jira describing the issue/improvement/feature (a notable exemptions is, changes not visible to end users)
- [N/A] Commit message must be in the form "[GEOS-XYZW] Title of the Jira ticket" (export to XML in Jira generates the message in this exact form)
- [N/A] New unit tests have been added covering the changes

### DIFF
--- a/src/community/geogig/src/main/java/org/geogig/geoserver/config/PostgresConfigBean.java
+++ b/src/community/geogig/src/main/java/org/geogig/geoserver/config/PostgresConfigBean.java
@@ -6,7 +6,6 @@ package org.geogig.geoserver.config;
 
 import com.google.common.base.Preconditions;
 import java.io.Serializable;
-import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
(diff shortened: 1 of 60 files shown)


aaime:
Nice cleanup :-D

aaime:
I guess it's an application of this IntelliJ cleanup?

image

bradh:
> I guess it's an application of this IntelliJ cleanup?
> 
> image

I use NetBeans, so it was a manual cleanup.

Unfortunately its not available in Java 8 (10+ only), which is why the CI is red. Should have been more careful, but at least the process stuff saved me the revert.

aaime:
Oh I see, StandardCharsets is there since Java 7, but URLDecode learned to use Charset instead of String arguments only in Java 10. There are still lots of places where the constants can be used, althought with smaller reduction on the try/catch blocks, e.g.:

image

I'll try to give it a crack with IntelliJ once I'm done with the other pending cleanups.

bradh:
> Oh I see, StandardCharsets is there since Java 7, but URLDecode learned to use Charset instead of String arguments only in Java 10. There are still lots of places where the constants can be used, althought with smaller reduction on the try/catch blocks, 
The concept is valid, just failed on specific execution in this case. 

> I'll try to give it a crack with IntelliJ once I'm done with the other pending cleanups.
Ok.

PR #4720, opened 2021-02-12
Geos 9903 - Fix issue with http headers missing 

roarbra:
Fixes a problem with WMTS cascade when using http headers. These headers must be set in constructor of WebMapTilServer. These have been added to Geotools in GEOT-6808.

Need to remove constructors with additional null.

## Checklist

For all pull requests:


aaime:
Uh oh, I've merged the GeoTools equivalent sure that I would find a GeoServer corresponding PR, but it's gone.
I cannot have a build broken for @jodygarnett to cut the RC Monday. Can you prepare a new one quickly? Otherwise I'll have to revert the GeoTools changes.

roarbra:
@aaime thanks for looking at this. I'm sorry for the inconveniences. I had to close this PR, because there were so much strange commits. The important commit is [this](https://github.com/geoserver/geoserver/pull/4720/commits/14a5ccd80fabc555e07ed70eb9e2e6f993c55357)

But that one do also have some problems. The file _src/web/app/src/test/java/org/geoserver/web/Start.java_ 
shouldn't have been a part of it.

I can take a look at it later today or early tomorrow.

aaime:
A fix during the weekeend sounds good, thank you!

PR #4722, opened 2021-02-12
[GEOS-9901] SLDService invalid ColorMap ramp when raster has single value - [GEOS-9902] SldService invalid ColorMap type interval, with negative values

taba90:
Jira tickets:
https://osgeo-org.atlassian.net/browse/GEOS-9901
https://osgeo-org.atlassian.net/browse/GEOS-9902

this pr fixes a bug occuring with classifications on single value raster producing invalid ColorMaps
and a bug occuring for closed interval with a negative value as the last break.

There is already a method handling the singleValue input, but worked only in the quantile case.The pr replace it with an ad hoc method for producing a consistent color map entry for each colorMap type, namely ramp (continuous) and intervals (open and closed). 
The pr makes also sure that every classification method returns always 1 or two breaks when the input is a single value raster.

## Checklist

For all pull requests:


src/extension/sldService/src/test/java/org/geoserver/sldservice/rest/ClassifierTest.java
```diff
@@ -1823,8 +1823,8 @@ public void testClassifyRasterSingleFloat() throws Exception {
         assertEquals(2, entries.length);
         ColorMapEntry cm0 = cm.getColorMapEntry(0);
         assertThat(cm0.getQuantity().evaluate(null, Float.class), Matchers.lessThanOrEqualTo(10f));
-        assertEquals("#FF071C", cm0.getColor().evaluate(null, String.class));
-        assertEquals(1, cm0.getOpacity().evaluate(null, Double.class), 0);
+        assertEquals("#000000", cm0.getColor().evaluate(null, String.class));
```
mbarto, line 1837: It would be useful to describe what this test is testing. You changed the test because the previous one was testing the actual code  behaviour, not the expected behaviour.
What I would do is explaining better what we are testing here, either using comments or refactoring the tests to make the intent more evident. It would be useful to do this for changed tests and new ones.

src/extension/sldService/src/test/java/org/geoserver/sldservice/rest/ClassifierTest.java
```diff
@@ -2805,4 +2816,23 @@ public void testClassifyRasterSingleByteClosed() throws Exception {
...
+                        + "&colors=0xFF071C,0xFFA92E&ramp=custom";
+        Document dom = getAsDOM(restPath, 200);
+        print(dom);
+        RasterSymbolizer rs = getRasterSymbolizer(dom);
+        ColorMap cm = rs.getColorMap();
+        ColorMapEntry[] entries = cm.getColorMapEntries();
+        // the max value of the raster is -50. The last entry should have the next
+        // value after it since we are requesting closed intervals
+        ColorMapEntry entry = entries[7];
+        assertEquals(-49.99999999999999, entry.getQuantity().evaluate(null, Double.class), 0d);
```
mbarto, line 2836: Maybe this test depends too much on the internal implementation that is offsetting your value by a very small amount, I would relax this test a little bit (value between -49.5 and -50 ?)

mbarto (approved):
LGTM, just a couple of comments to improve the tests a little bit

src/extension/sldService/src/test/java/org/geoserver/sldservice/rest/ClassifierTest.java
taba90: fixed

src/extension/sldService/src/test/java/org/geoserver/sldservice/rest/ClassifierTest.java
taba90: fixed

taba90:
thanks for your review @mbarto  your feedbacks have been applied

PR #4731, opened 2021-02-15
[GEOS-9909] Geofence spatialFilterType and CRS in allowed area not included in restOutput

taba90:
Jira ticket https://osgeo-org.atlassian.net/browse/GEOS-9909

## Checklist

For all pull requests:

### DIFF
--- a/src/extension/geofence-server/src/test/java/org/geoserver/geofence/rest/RulesRestControllerTest.java
+++ b/src/extension/geofence-server/src/test/java/org/geoserver/geofence/rest/RulesRestControllerTest.java
@@ -32,6 +32,8 @@ import org.geoserver.rest.RestBaseController;
 import org.geotools.gml3.bindings.GML3MockData;
 import org.junit.Before;
 import org.junit.Test;
+import org.locationtech.jts.geom.Geometry;
+import org.locationtech.jts.geom.MultiPolygon;
 import org.locationtech.jts.io.ParseException;
 import org.locationtech.jts.io.WKTReader;
 import org.springframework.http.HttpStatus;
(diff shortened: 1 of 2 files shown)


src/extension/geofence-server/src/test/java/org/geoserver/geofence/rest/RulesRestControllerTest.java
```diff
@@ -576,7 +580,7 @@ public void testRestControllerPaths() throws Exception {
                 assertEquals("DE_USNG_UTM18", layerDetails.getString("defaultStyle"));
                 assertEquals("Northings >= 100", layerDetails.getString("cqlFilterRead"));
                 assertEquals(
-                        "MULTIPOLYGON (((-180 -90, -180 90, 180 90, 180 -90, -180 -90)))",
+                        "SRID=4326;MULTIPOLYGON (((-180 -90, -180 90, 180 90, 180 -90, -180 -90)))",
```
aaime, line 583: Interesting... this is making the API non backwards compatible though, or am I misunderstanding?

etj: Confirmed, it is not backward compatible.
The GeoFence part of the SRID user story did break compatibility, so that GeoFence had to change version from 3.4 to 3.5

etj: See also https://github.com/geoserver/geofence/wiki/Compatibility-matrix-(GS-2.17-2.19,-GF-3.4.x-3.5.x)

aaime: Hum... I guess it's late, but just thinking out loud, wondering if it was really necessary to do... the SRS could have been a separate field, with a default value making it work like before?

etj: It may be Good that the whole field become unparsable between different versions: if new data arrives to old authorization logic, the old logic would deal with the geometry as it was 4326, possibly giving access to unauthorized areas, and probably this could be even worse than a runtime error.

aaime: @etj that's not how GeoServer handle backwards compatibility. GeoServer upgrades must be transparent if at all possible, an old client should keep on working un-modified against a newer version of GeoServer.

If someone tries to use a newer client against an old version of the server, that's their problem, we're not in control of clients.

That said, in this case I'm guessing damage is done already. Please keep this in mind and avoid backwards incompatibilities, if at all  possible, in the future.

PR #4748, opened 2021-02-17
[GEOS-9887][GSIP-195]: Migrate WPS-Download to extension

dromagnoli:
Migrate WPS-Download from community modules to extensions.

* https://github.com/geoserver/geoserver/wiki/GSIP-195

## Checklist

For all pull requests:

- [x] There is a ticket in Jira https://osgeo-org.atlassian.net/browse/GEOS-9887


doc/en/user/source/extensions/wps-download/index.rst
```diff
@@ -1,22 +1,30 @@
...
+Installing the WPS Download extension
+-------------------------------------
 
-#. Download the WPS download module from the `nightly GeoServer community module builds <https://build.geoserver.org/geoserver/master/community-latest/>`_.
+The WPS Download extension is listed among the other extension downloads on the GeoServer download page.
 
-   .. warning:: Make sure to match the version of the extension to the version of the GeoServer instance.
+The installation process is similar to other GeoServer extensions:
+
+#. Download the :download_extension:`wps-download`
```
jodygarnett, line 19: Thanks for using download directive!

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/GridGeometryProvider.java
```diff
@@ -505,10 +513,8 @@ private GridGeometry2D computeGridGeometry2D(
         // requested gridGeometry such that the resulting scale doesn't perfectly match the
         // requested resolution
         if (Math.abs(scaleX - resolution[0]) > 1E-6 || (Math.abs(scaleY - resolution[1]) > 1E-6)) {
-            if (crsRequestHandler != null
-                    && crsRequestHandler.getReferenceFeatureForAlignment() != null) {
-                SimpleFeature referenceFeature =
-                        crsRequestHandler.getReferenceFeatureForAlignment();
+            SimpleFeature referenceFeature = crsRequestHandler.getReferenceFeatureForAlignment();
```
aaime, line 516: crsRequestHandler is not checked for null anymore, is it guaranteed to be non null here?

aaime (changes requested):
Running a grep I've found a couple of issues:

```
> git grep -B 1 gs-wps-download
extension/wps-download/pom.xml- <groupId>org.geoserver.extension</groupId>
extension/wps-download/pom.xml: <artifactId>gs-wps-download</artifactId>
--
release/ext-wps-download.xml-      <includes>
release/ext-wps-download.xml:        <include>gs-wps-download*.jar</include>
--
release/ext-wps.xml-        <exclude>gs-wps-cluster-hazelcast*</exclude>
release/ext-wps.xml:        <exclude>gs-wps-download*</exclude>    (1)
--
release/pom.xml-     <groupId>org.geoserver.extension</groupId>
release/pom.xml:     <artifactId>gs-wps-download</artifactId>
--
web/app/pom.xml-          <groupId>org.geoserver.community</groupId> (2)
web/app/pom.xml:          <artifactId>gs-wps-download</artifactId>
```

(1): the WPS main zip package must not include wps-download, it should be in its own separate zip (not sure about hazelcast, it also has its own zip, should not be there either....)
(2): the web/app pom still refers to wps-download using the community groupId

I've also checked the contents of the zip, seems fine.

dromagnoli:
> release/ext-wps.xml-        <exclude>gs-wps-cluster-hazelcast*</exclude>
> release/ext-wps.xml:        <exclude>gs-wps-download*</exclude>    (1)
> --
> release/pom.xml-     <groupId>org.geoserver.extension</groupId>
> release/pom.xml:     <artifactId>gs-wps-download</artifactId>
> --
> web/app/pom.xml-          <groupId>org.geoserver.community</groupId> (2)
> web/app/pom.xml:          <artifactId>gs-wps-download</artifactId>
> ```
> 
> (1): the WPS main zip package must not include wps-download, it should be in its own separate zip (not sure about hazelcast, it also has its own zip, should not be there either....)

Note that they are actually on an <exclude> flag indeed.
I had to put it gs-wps-download into an exclude section because I tested the release zip and found it within the wps-plugin.zip.
So I have used that <exclude> which is actually what are you suggesting, right?


> (2): the web/app pom still refers to wps-download using the community groupId

DAMN!
I did it the first time I have prepared the PR... Then git mixed the 2 commits so I did it again, this time probably forgetting it... fixing. Thanks for the check

src/community/wps-download/src/main/java/org/geoserver/wps/gs/download/GridGeometryProvider.java
aaime: > > (1): the WPS main zip package must not include wps-download, it should be in its own separate zip (not sure about hazelcast, it also has its own zip, should not be there either....)
> > ```
> 
> Note that they are actually on an flag indeed.
> I had to put it gs-wps-download into an exclude section because I tested the release zip and found it within the wps-plugin.zip.
> So I have used that which is actually what are you suggesting, right?

Ah yes, did not see they were in exclude. Good.

PR #4752, opened 2021-02-18
Prefix template lookup to prevent conflict between ogcapi services

jodygarnett:
Issue found while writing docs.

## Checklist

For all pull requests:

### DIFF
--- a/src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/FreemarkerTemplateSupport.java
+++ b/src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/FreemarkerTemplateSupport.java
@@ -27,12 +27,11 @@ import org.springframework.stereotype.Component;
  * Support class that locates the templates based on the current response and eventual {@link
  * LocalWorkspace}.
  *
- * <p>
- *     Located in workspace using service landingPage prefix, or obtained from jar:
- *     <ul>
- *         <li>ogc/features/landingPage.ftl</li>
- *     </ul>
- * </p>
+ * <p>Located in workspace using service landingPage prefix, or obtained from jar:
+ *
+ * <ul>
+ *   <li>ogc/features/landingPage.ftl
+ * </ul>
  */
 @Component
 public class FreemarkerTemplateSupport {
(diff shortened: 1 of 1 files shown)


src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/FreemarkerTemplateSupport.java
```diff
@@ -45,27 +52,37 @@ public FreemarkerTemplateSupport(GeoServerResourceLoader loader) {
...
-                            if (source != null) {
-                                return new ClassTemplateSource(path, source);
-                            }
-                        }
-
-                        return source;
-                    }
-                };
+    GeoServerTemplateLoader templateLoader =
+        new GeoServerTemplateLoader(clazz, resoureLoader) {
```
bradh, line 60: Typo: `resourceLoader`

bradh (commented):
The formatting looks off, but that could be Github presentation.

src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/FreemarkerTemplateSupport.java
```diff
@@ -45,27 +52,37 @@ public FreemarkerTemplateSupport(GeoServerResourceLoader loader) {
...
+          @Override
+          public Object findTemplateSource(String path) throws IOException {
+            Object source = null;
+
+            APIService service = clazz.getAnnotation(APIService.class);
+            if (service != null) {
+              source = super.findTemplateSource(service.landingPage() + "/" + path);
```
aaime, line 67: Ok so it seems this would make it at "ogcapi/features" right?
I don't see a test, given it's a community module, it's not require either, but... did you try it out manually and verified it can be customized? 
Some docs would help user finding this possibilty... however the module has none so far.

aaime: Ah, found the docs in the other PR.  Assuming you also tried it out, juding from the docs.

aaime (commented):
Looks good to me in general, @bradh feedback needs to be addressed yet.

src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/FreemarkerTemplateSupport.java
jodygarnett: Yes it is working fine, and you found the docs.

jodygarnett: Good:

* This allows for `templates/ogc/features/landingPage.ftl` to be customized.

Bad: 

* It does not fix the issue of someone does add `templates/landingPage.ftl` all the services will find it and will break :(

aaime: Agreed

PR #4779, opened 2021-02-23
[GEOS-9927] Add StyleComponentInfo in AbstractStylePage

weskamm:
This adds a new entry point for extensions to be instantiated in the AbstractStylePage and thus get access to the style edit pages. For more information, see https://osgeo-org.atlassian.net/browse/GEOS-9927

## Checklist

For all pull requests:

### DIFF
--- a/src/web/wms/src/main/java/org/geoserver/wms/web/data/AbstractStylePage.java
+++ b/src/web/wms/src/main/java/org/geoserver/wms/web/data/AbstractStylePage.java
@@ -613,6 +613,23 @@ public abstract class AbstractStylePage extends GeoServerSecuredPage {
                     }
                 };
         add(cancelLink);
+
+        // add additional style components that may be used e.g. by extensions
+        List<StyleComponentInfo> compInfo =
+                getGeoServerApplication().getBeansOfType(StyleComponentInfo.class);
+        for (StyleComponentInfo comp : compInfo) {
+            try {
+                Class<?> component = comp.getComponentClass();
+                Component c =
+                        (Component)
+                                component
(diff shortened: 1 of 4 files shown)


aaime:
@weskamm if you want to run the QA locally, the [QA guide](https://docs.geoserver.org/latest/en/developer/qa-guide/index.html) shows you how to do that. Running them on the whole source code takes a lot of time, but if you cd into the one module modified, then it's pretty quick (just one module being checked) and gives you a better turnaround between changes and checks.

weskamm:
Thanks for the hint, i think i finally figured out the issues

aaime:
The change looks legit, but there is no test. We don't normally merge PRs without one.

Thinking out loud, at test could be done that uses a test component and verifies it plugged in correctly. During tests it's possible to inject a testing bit of Spring app context, overriding a method like in this example (taken from a WCS test):

```
    protected void setUpSpring(List<String> springContextLocations) {
        super.setUpSpring(springContextLocations);
        springContextLocations.add("classpath:/org/geoserver/wcs/ResourceAccessManagerContext.xml");
    }
```

You can find a full example of an extensible page being tested here:
https://github.com/geoserver/geoserver/blob/7afff47045ed55f4d17a838245bc091ddcffeff0/src/web/core/src/test/java/org/geoserver/web/GeoServerHomePageTest.java

weskamm:
I tried hard adding a test, but i did not find an easy way to mockup a bean for testing without changing lots of code. I will have a look at your recommendation

weskamm:
Thanks for pointing out the direction, i have added a test now :tada:

aaime:
Looking good, thanks! Merging.

PR #4780, opened 2021-02-23
[GEOS-9928] Make GeoServer runs its startup loading process as root

fernandor777:
GeoServer is not Root/Admin during startup process, making it fails to load secured workspaces settings (WMS, WFS, ...).  Current workaround is to reload catalog from Admin user UI.  This PR introduces a fix to this issue, making GeoServer uses the root user and admin role during its startup loading process.

JIRA issue:
https://osgeo-org.atlassian.net/browse/GEOS-9928

## Checklist

For all pull requests:

### DIFF
--- a/src/main/src/main/java/org/geoserver/config/GeoServerLoaderListener.java
+++ b/src/main/src/main/java/org/geoserver/config/GeoServerLoaderListener.java
@@ -1,0 +1,43 @@
+/* (c) 2021 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.config;
+
+import org.geoserver.catalog.Catalog;
+import org.geoserver.config.util.XStreamPersister;
+
+/**
+ * Listener interface for {@linkplain DefaultGeoServerLoader} load catalog and geoServer startup
+ * process.
+ */
+public interface GeoServerLoaderListener {
+
+    public static final GeoServerLoaderListener EMPTY_LISTENER = new EmptyGeoServerLoaderListener();
+
+    /**
+     * Listener method executed after GeoServer catalog loading procedure.
+     *
+     * @param catalog GeoServer catalog instance
+     * @param xp {@linkplain XStreamPersister} instance
+     */
+    void loadCatalog(Catalog catalog, XStreamPersister xp);
+
+    /**
+     * Listener method executed after GeoServer setup loading procedure.
+     *
+     * @param geoServer GeoServer instance
+     * @param xp {@linkplain XStreamPersister} instance
+     */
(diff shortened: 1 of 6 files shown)


src/main/src/main/java/org/geoserver/config/GeoServerLoaderListener.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Listener interface for {@linkplain DefaultGeoServerLoader} load catalog and geoServer startup
+ * process.
+ */
+public interface GeoServerLoaderListener {
+
+    public static final GeoServerLoaderListener EMPTY_LISTENER = new EmptyGeoServerLoaderListener();
+
+    void loadCatalog(Catalog catalog, XStreamPersister xp);
```
aaime, line 18: Javadoc missing. Is the even happening before or after loading?

src/main/src/main/java/org/geoserver/config/GeoServerLoaderListener.java
```diff
@@ -0,0 +1,35 @@
...
+
+    void loadGeoServer(GeoServer geoServer, XStreamPersister xp);
```
aaime, line 20: Same as above.

src/main/src/main/java/org/geoserver/config/GeoServerLoaderListener.java
```diff
@@ -0,0 +1,35 @@
...
+
+    default boolean isEmpty() {
```
aaime, line 22: Not sure this is useful. The empty listener does nothing anyways? The check has to call this, which actually does _some_ work, making things slower than just calling the empty method.

aaime (commented):
Overall looks good, see inline comments for minor issues.

src/main/src/main/java/org/geoserver/config/GeoServerLoaderListener.java
fernandor777: Indeed, added Javadoc describing it is running after loading.

src/main/src/main/java/org/geoserver/config/GeoServerLoaderListener.java
fernandor777: Indeed, added Javadoc.

src/main/src/main/java/org/geoserver/config/GeoServerLoaderListener.java
fernandor777: Makes sense, removed the isEmpty method and using the empty instance methods instead checking this method now.

fernandor777:
Thanks for your review @aaime, I fixed the PR based on your comments.

aaime (approved):
Looks good, thanks!

PR #4782, opened 2021-02-23
[GEOS-9331] Upgrade Freemarker to 2.3.31

aaime:
Upgrades the freemarker library to the latest version. The planted a "bomb" in the 2.3.21 version that was difficult to figure out, a change in behavior that was meant to be an improvement, but turned out to be very damaging in GeoServer. Thankfully, once figured out the nature of the issue, it was not difficult to work around it.

## Checklist

For all pull requests:

- [x] ~~New unit tests have been added covering the changes~~ As a version upgrade, only checked that existing tests do pass, and updated them according to changed behavior (small changes in json escaping)

### DIFF
--- a/src/main/src/main/java/org/geoserver/template/TemplateUtils.java
+++ b/src/main/src/main/java/org/geoserver/template/TemplateUtils.java
@@ -5,9 +5,12 @@
 
 package org.geoserver.template;
 
+import static freemarker.template.Configuration.DEFAULT_INCOMPATIBLE_IMPROVEMENTS;
+
 import freemarker.core.TemplateClassResolver;
 import freemarker.template.Configuration;
 import freemarker.template.TemplateException;
+import freemarker.template.Version;
 import freemarker.template.utility.ClassUtil;
 import java.util.Arrays;
 import java.util.Collection;
@@ -18,6 +21,9 @@ import java.util.Collection;
  * @author Kevin Smith, Boundless
  */
 public class TemplateUtils {
+    /** Reference feature version for Freemarker templates */
+    public static Version FM_VERSION = Configuration.VERSION_2_3_0;
+
     /** Classes that should not be resolved in Freemarker templates */
     private static final Collection<String> ILLEGAL_FREEMARKER_CLASSES =
             Arrays.asList(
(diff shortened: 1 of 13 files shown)


dromagnoli (approved):

bradh (commented):
Should we set a legacy "fixed" version, as described in https://freemarker.apache.org/docs/pgui_config_settings.html ? I can't see that in the current code, so it could be a follow-on activity.

aaime:
PR updated to resolve deprecated API usage. Also comes with fixing the reference version to the one that was already implied (basically replaced deprecated API calls by in-lining them, which reveals the defaults).

aaime:
Also centralized the reference version number in a shared constant.

src/main/src/main/java/org/geoserver/template/TemplateUtils.java
```diff
@@ -18,6 +21,9 @@
  * @author Kevin Smith, Boundless
  */
 public class TemplateUtils {
+    /** Reference feature version for Freemarker templates */
+    public static Version GS_VERSION = Configuration.VERSION_2_3_0;
```
bradh, line 25: Its in the noise, but GS_VERSION seems a bit strange. FM_VERSION or FREEMARKER_CONFIG_VERSION?

aaime: FM_VERSION it is, updated.

bradh:
We have a bunch of stuff like https://github.com/geoserver/geoserver/blob/master/src/community/ogcapi/ogcapi-features/src/main/resources/org/geoserver/ogcapi/features/landingPage.ftl#L41

Is this change backwards-compatible (enough) that we don't need to deal with that yet?

aaime:
@bradh those comments are reminders to improve the templates once we upgrade, so they are actually waiting for it, no backwards compatibility concern there (it's about doing stuff better once we have the newer version).

The upgrade will likely be backwards incompatible to some, the amount of changed piled up by Freemarker in these "bugfix releases" is nothing short of mind boggling, have a look, if you're more patient than me you might even read them all:
* https://freemarker.apache.org/docs/versions_2_3_19.html
* https://freemarker.apache.org/docs/versions_2_3_20.html
* https://freemarker.apache.org/docs/versions_2_3_21.html
* https://freemarker.apache.org/docs/versions_2_3_22.html
* https://freemarker.apache.org/docs/versions_2_3_23.html
* https://freemarker.apache.org/docs/versions_2_3_24.html
* https://freemarker.apache.org/docs/versions_2_3_25.html
* https://freemarker.apache.org/docs/versions_2_3_26.html
* https://freemarker.apache.org/docs/versions_2_3_27.html
* https://freemarker.apache.org/docs/versions_2_3_28.html
* https://freemarker.apache.org/docs/versions_2_3_29.html
* https://freemarker.apache.org/docs/versions_2_3_30.html
* https://freemarker.apache.org/docs/versions_2_3_31.html

That said, Freemarker has piled up a number of important upgrades, including fixing some CVEs. I'm proposing this PR only for master only, and we'll add a warning in the release notes. For 2.18.x we'll have to find another way.

PR #4663, opened 2021-01-18
[GEOS-9883] Add pom.xml formatting + formatting QA check

mprins:
Adds an XML format check + xml fomatter using [au.com.acegi:xml-format-maven-plugin](https://acegi.github.io/xml-format-maven-plugin/index.html) for all the pom files and Spring context files

formats:
- [x] `pom.xml` files
- [x] `applicationContext.xml` and `applicationSecurityContext.xml` files 
- [ ] `web.xml`
- [ ] ...

resolves [GEOS-9883](https://osgeo-org.atlassian.net/browse/GEOS-9883)

## Checklist

> Reviewing is a process done by project maintainers, mostly on a volunteer basis. We try to keep the overhead as small as possible and appreciate if you help us to do so by completing the following items. Feel free to ask in a comment if you have troubles with any of them.

**Submitting the PR does not require you to check all items, but by the time it gets merged, they should be either satisfied or not applicable.**

For all pull requests:

- [x] Confirm you have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/master/CONTRIBUTING.md) 
- [x] You have sent a Contribution Licence Agreement (CLA) as necessary (not required for small changes, e.g., fixing typos in documentation)
- [x] Make sure the first PR targets the master branch, eventual backports will be managed later. This can be ignored if the PR is fixing an issue that only happens in a specific branch, but not in newer ones.

The following are required only for core and extension modules (they are welcomed, but not required, for community modules):
- [x] There is a ticket in Jira describing the issue/improvement/feature (a notable exemptions is, changes not visible to end users)
- [ ] PR for bug fixes and small new features are presented as a single commit
- [ ] Commit message must be in the form "[GEOS-XYZW] Title of the Jira ticket" (export to XML in Jira generates the message in this exact form)
- [x] The pull request contains changes related to a single objective. If multiple focuses cannot be avoided, each one is in its own commit and has a separate ticket describing it.
- [x] New unit tests have been added covering the changes
- [x] This PR passes all existing unit tests (test results will be reported by Continuous Integration after opening this PR)
- [x] This PR passes the [QA checks](https://docs.geoserver.org/latest/en/developer/qa-guide/index.html) (QA checks results will be reported by Continuous Integration after opening this PR)
- [ ] Commits changing the UI, existing user workflows, or adding new functionality, need to include documentation updates (screenshots, text)
- [ ] Commits changing the REST API, or any configuration object, should check if the REST API docs (Swagger YAML files and classic documentation) need to be updated.

PR #4800, opened 2021-02-27
Format app context files with the Eclipse xml formatter

aaime:
Same goal as https://github.com/geoserver/geoserver/pull/4663, but trying out a different tool.

## Checklist

For all pull requests:

### DIFF
--- a/build/qa/xml-format.properties
+++ b/build/qa/xml-format.properties
@@ -1,0 +1,7 @@
+maxLineLength=80
+splitMultiAttrs=false
+tabInsteadOfSpaces=false
+tabWidth=4
+wellFormedValidation=WARN
+wrapLongLines=true
+spaceBeforeEmptyCloseTag=false
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -1764,7 +1764,35 @@
        </execution>
      </executions>
    </plugin>
-   
+
+      <plugin>
+        <groupId>net.revelc.code.formatter</groupId>
+        <artifactId>formatter-maven-plugin</artifactId>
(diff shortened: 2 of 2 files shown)


src/security/pom.xml
```diff
@@ -1,59 +1,61 @@
...
+        <artifactId>geoserver</artifactId>
+        <version>2.19-SNAPSHOT</version>
+    </parent>
 
-  <groupId>org.geoserver</groupId>
-  <artifactId>gs-security</artifactId>
-  <packaging>pom</packaging>
-  <name>GeoServer Security Modules</name>
+    <groupId>org.geoserver</groupId>
+    <artifactId>s-security</artifactId>
```
bradh, line 21: Is this from the tooling or manual edits? Certainly looks a bit strange, and appears to be (part of) what is breaking the build.

aaime: Hum... wondering if the plugin is safe to parallel builds

aaime: Indeed I'm getting other failures as well. Reported here:
https://github.com/revelc/formatter-maven-plugin/issues

aaime:
Reduced the scope to just formatting the application context files. Unfortunately, it's still not working in a parallel build. Leaving this as a reference for the revelc people to look at, if they so desire.

aaime:
For reference, this command, getting most of the modules in the build, almost systematically fails on a AMD Ryzen 1700x, 8 core:

```
mvn formatter:format -nsu -T6 -Prelease,communityRelease
```

The build above involves over 200 modules. The failures involve different modules at each run.

PR #4811, opened 2021-03-03
[GEOS-9949] System dependency affecting tests in core platform module (Paths.java)

prushforth:
The first component of an absolute path on windows is a drive letter, which if it doesn't get a "\" appended to it gets treated oddly in some circumstances only, apparently.  Affects tests FileWrapperResourceTheoryTest, ResourcesTest in core platform module.

## Checklist

For all pull requests:

### DIFF
--- a/src/platform/src/main/java/org/geoserver/platform/resource/Paths.java
+++ b/src/platform/src/main/java/org/geoserver/platform/resource/Paths.java
@@ -423,7 +423,12 @@ public class Paths {
      */
     public static File toFile(File base, String path) {
         for (String item : Paths.names(path)) {
-            base = new File(base, item);
+            base =
+                    new File(
+                            base,
+                            base == null && File.separator.equalsIgnoreCase("\\")
+                                    ? item + "\\"
+                                    : item);
         }
         return base;
     }
--- a/src/platform/src/test/java/org/geoserver/platform/resource/PathsTest.java
+++ b/src/platform/src/test/java/org/geoserver/platform/resource/PathsTest.java
@@ -12,6 +12,7 @@ import static org.geoserver.platform.resource.Paths.path;
 import static org.geoserver.platform.resource.Paths.sidecar;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import java.io.File;
@@ -32,6 +33,8 @@ public class PathsTest {
     final String SUBFOLDER = "directory/folder";
 
     final String FILE3 = "directory/folder/file3.txt";
+
+    final boolean WIN = System.getProperty("os.name").startsWith("Windows");
 
     @Test
     public void pathTest() {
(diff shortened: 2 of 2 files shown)


src/platform/src/main/java/org/geoserver/platform/resource/Paths.java
```diff
@@ -423,7 +423,12 @@ public static String convert(String path, String filename) {
      */
     public static File toFile(File base, String path) {
         for (String item : Paths.names(path)) {
-            base = new File(base, item);
+            base =
+                    new File(
+                            base,
+                            base == null && File.separator.equalsIgnoreCase("\\")
+                                    ? item + "\\"
+                                    : item);
```
jodygarnett, line 431: Do you have a test case to show what you are  does? You can make a windows only test case and it will be picked up by the GitHub actions.

We are explicitly using the file constructors here rather than ``/`` or ``\\`` to avoid system dependencies so I am a little startled by your suggested change.

But I see your point, the javadoc is clear that "in the case of absolute path base should be null ... 

- Is the path you are thinking of using the form ``C:\\path\filename.txt`` ?
- Network shares should be of the form ``\\\\machine\\dirve\path\filename.ext``

Consider checking File.listRoots() - to check first item, seems complicated.

jodygarnett (changes requested):
A test case is required, also not sure I understand what kind of absolute path you are having difficulty working with (please include some examples in the test case).

src/platform/src/main/java/org/geoserver/platform/resource/Paths.java
prushforth: ok that's a good suggestion; I realized that this did not show up ususally and I wondered why it didn't when the system is tested under windows github action, and I definitely didn't consider the network share situation.  I'll try again.  Thanks!

prushforth: Note that there is some post-Paths.toFile work going on [here ](https://github.com/geoserver/geoserver/blob/770dc6f7023bc2ab32597cfc7a3a9cc35ff3b608/src/platform/src/test/java/org/geoserver/platform/resource/FileWrapperResourceTheoryTest.java#L32-L33) that is addressing on linux the problem that is encountered on Windows, namely an absolute path appearing as relative after toFile(null, path).  I pushed a test that fails on linux and unfortunately I can't deal with it until tonight.

prushforth:
Closing for now.  Will set up a couple of options for discussion.  As well, I'll try to install Windows subsystem for Linux and see if I can actually run the Linux tests too.

src/platform/src/main/java/org/geoserver/platform/resource/Paths.java
prushforth: @jodygarnett  So I set up two branches that solve the problem (on Windows, Linux TBD).  The first branch is the same approach taken here, with hopefully the linux test failure removed, [here](https://github.com/geoserver/geoserver/compare/main...Maps4HTML:GEOS-9949-A).

The second approach leaves the Paths.toFile method unchanged, and applies some compensating gyrations to where the failure occurs (in FileWrapperResourceTheoryTest.getResource method) the branch is [here](https://github.com/geoserver/geoserver/compare/main...Maps4HTML:GEOS-9949-B).

Neither approach is great, even assuming they work on both systems.  I'm starting to think all this is to work around a bug in java...File that doesn't allow it to correctly resolve paths in the constructor. 

Please let me know what you think.  No rush, I realize you have a ton on your plate right now. Thanks!

PR #4897, opened 2021-03-23
Fix for issue with GSR FeatureEncoder converting number based attributes to strings

jsmithmap:
Fix for issue with GSR FeatureEncoder converting number based attribute values to strings. This was preventing Attribute based SLD styles using tags like <ogc:PropertyIsLessThan> from working in some clients as all numbers were strings in the GeoServices REST JSON output.

### DIFF
--- a/src/community/gsr/src/main/java/org/geoserver/gsr/translate/feature/FeatureEncoder.java
+++ b/src/community/gsr/src/main/java/org/geoserver/gsr/translate/feature/FeatureEncoder.java
@@ -59,6 +59,10 @@ public class FeatureEncoder {
                     value = ((java.util.Date) prop.getValue()).getTime();
                 } else if (prop.getValue() instanceof java.lang.Boolean) {
                     value = ((Boolean) prop.getValue()) ? Integer.valueOf(1) : Integer.valueOf(0);
+                } else if (prop.getValue() instanceof java.lang.Integer) {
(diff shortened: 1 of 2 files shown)


aaime:
From the QA build:

"Error: 3,419 [ERROR] Failed to execute goal com.coveo:fmt-maven-plugin:2.4.0:check (default) on project gs-gsr: Found 1 non-complying files, failing build -> [Help 1]"

You can fix the formatting by building the module from maven, on the command line.

jsmithmap:
Thanks Andrea!  It is now passing all of the checks.  If everything looks good with this change I would like to backport it to the 2.17.x branch and ideally 2.18.x and 2.19.x as well.  I assume I just need to cherry-pick the change and then submit pull requests for each of those branches.  Is that correct?

aaime:
Hi @jsmithmap, all build checks are green, so I'm merging the changes.

About backport, yes, a cherry pick should do the trick. The 2.18.x and 2.19.x branches are alive and nightly builds are currently being produced, while the 2.17.x branch is now dead and new nightly builds are no longer being produced (the last one is from March 1st): https://build.geoserver.org/geoserver/
You can still backport there too, but you'll have to create the jars and eventually zip of the module yourself.

aaime:
I've squash merged the commit, please use commit id e05a263bebf5d4fc981e7aedc52c5aada783e2bb for cherry-picks.

PR #4917, opened 2021-03-28
Upgrade flatgeobuf dependency

bjornharrtell:


bradh:
Restarted the Mac OS build - failure looks unrelated.

bradh:
Do you need an upgrade over at https://github.com/geotools/geotools/blob/main/modules/unsupported/flatgeobuf/pom.xml#L30 for consistency? Otherwise both will likely end up in your GeoServer at some stage.

bjornharrtell:
@bradh yes probably and done in https://github.com/geotools/geotools/pull/3425.

I wonder, can a community geoserver module depend on a community geotools module? Then this module could (and should?) depend on geotools only. On a related note, I'm considering migration all geotools related things from flatgeobuf into geotools.

bradh:
@bjornharrtell As long as it doesn't break the build, there aren't a lot of constraints on geoserver community modules.  Depending on a community module in geotools is entirely reasonable to me.

bjornharrtell:
I'm well under way to move needed logic into the geotools community module so I'm closing this as that is the better way forward.

PR #4921, opened 2021-04-01
[GEOS-10004] Add ability to animate env variables in download-WPS

weskamm:
[GEOS-10004](https://osgeo-org.atlassian.net/browse/GEOS-10004)

This PR adds the ability to create animations via the download-WPS extension without specifying the TIME parameter but instead using an "animateParam" and "animateValues" parameter to animated via SLD variable substitution.

## Checklist

For all pull requests:


weskamm:
Not sure why the build is failing, maybe a random error which disappears when that job is restarted?

doc/en/user/source/extensions/wps-download/mapAnimationDownload.rst
```diff
@@ -225,3 +227,85 @@ Decoration Layout
...
+            <wps:BoundingBoxData crs="EPSG:4326">
+              <ows:LowerCorner>-180 -90</ows:LowerCorner>
+              <ows:UpperCorner>180 90</ows:UpperCorner>
+            </wps:BoundingBoxData>
+          </wps:Data>
+        </wps:Input>
+        <wps:Input>
+          <ows:Identifier>animateParam</ows:Identifier>
+          <wps:Data>
+            <wps:LiteralData>env</wps:LiteralData>
```
aaime, line 267: Seems like any parameter can be used, like in the animator. Which means, the documentation should state so, rather than referring to "SLD variable substitution" as the only possibility (besides time).

src/extension/wps-download/src/main/java/org/geoserver/wps/gs/download/DownloadAnimationProcess.java
```diff
@@ -159,8 +174,18 @@ public RawData execute(
                 new AWTSequenceEncoder(NIOUtils.writableChannel(output.file()), frameRate);
 
         DownloadServiceConfiguration configuration = confiGenerator.getConfiguration();
-        TimeParser timeParser = new TimeParser(configuration.getMaxAnimationFrames());
-        Collection parsedTimes = timeParser.parse(time);
+        int totalTimes;
```
aaime, line 177: Does not represent a count of times anymore. 
```suggestion
        int totalFrames;
```

src/extension/wps-download/src/main/java/org/geoserver/wps/gs/download/DownloadAnimationProcess.java
```diff
@@ -194,16 +218,24 @@ public RawData execute(
...
-                String mapTime = toWmsTimeSpecification(parsedTime);
+            for (int i = 0; i < totalTimes; i++) {
+                String mapTime = null;
+                String animateValue = null;
+                if (parsedTimes != null) {
+                    // turn parsed time into a specification and generate a "WMS" like request based
+                    // on it
+                    mapTime = toWmsTimeSpecification(parsedTimes.toArray()[i]);
+                } else {
+                    animateValue = animateValues.split(",")[i];
```
aaime, line 229: Splitting over and over and over? Please prepare a list of parsed values while parsing (code above), something like:

```
List<Object> frameValues = new ArrayList<>();
if (time) {
  parse the times one by one
} else {
  split the animation params by comma
}
```

src/extension/wps-download/src/main/java/org/geoserver/wps/gs/download/DownloadMapProcess.java
```diff
@@ -301,6 +305,9 @@ RenderedImage buildImage(
         if (time != null) {
             template.put("time", time);
         }
+        if (animateParam != null && animateValue != null) {
```
aaime, line 308: See what' happening here? You're basically doing the same as time, but with generic parameters.
I'd suggest to get rid of the time parameter altoghether, and fill the param name as "time" in the caller (that is, use the generic parameter path at this point). The method interface is broken anyways, might as well keep it clean.

src/wms/src/main/java/org/geoserver/wms/RenderingVariables.java
```diff
@@ -37,6 +38,12 @@ public static void setupEnvironmentVariables(WMSMapContent mapContent) {
         EnvFunction.setLocalValue(WMS_SRS, mapContent.getRequest().getSRS());
         EnvFunction.setLocalValue(WMS_WIDTH, mapContent.getMapWidth());
         EnvFunction.setLocalValue(WMS_HEIGHT, mapContent.getMapHeight());
+        // add custom env variables
+        Map<String, Object> env = mapContent.getRequest().getEnv();
```
aaime, line 42: Can work, but where are env variables handled now? Doesn't this change make the existing handling reduntant?
Could you elaborate on why this change was necessary to start with?

src/extension/wps-download/src/main/java/org/geoserver/wps/gs/download/DownloadAnimationProcess.java
```diff
@@ -107,12 +108,26 @@ public RawData execute(
...
                         name = "time",
-                        min = 1,
+                        min = 0,
                         description =
                                 "Map time specification (a range with "
                                         + "periodicity or a list of time values)"
                     )
                     String time,
+            @DescribeParameter(
+                        name = "animateParam",
```
aaime, line 118: So these two are equivalents of "aParam", "aValues" of the old animator, right?

doc/en/user/source/extensions/wps-download/mapAnimationDownload.rst
```diff
@@ -13,6 +13,8 @@ The map and animation downloads work off a set of common parameters:
  * ``bbox`` : a geo-referenced bounding box, controlling both output area and desired projection
  * ``decoration`` : the name of a decoration layout to be added on top of the map
  * ``time`` : a WMS ``time`` specification used to drive the selection of times across the layers in the map, and to control the frame generation in the animation
+ * ``animateParam`` : the name of a SLD variable that should be used for animation. Cannot be used in conjunction with the ``time`` parameter
+ * ``animateValues`` : the comma separated values used for the ``animateParam`` variable
```
aaime, line 17:
```suggestion
 * ``animateValues`` : the comma separated list of values used for the ``animateParam`` variable
```

aaime (changes requested):

doc/en/user/source/extensions/wps-download/mapAnimationDownload.rst
weskamm: Yes this is because i feel unsure if we shall support other parameters besides env. i haven't tested the behaviour with different layers and at least the bbox parameter would be conflict with the current process arguments. I think i will reduce this to only supporting env variables

aaime: I think it's good to have some liberty, there are other variables that could be used for animations, like:
- elevation
- custom dimensions
- view params
I'd suggest to just document which parameter names might be problematic.

src/wms/src/main/java/org/geoserver/wms/RenderingVariables.java
weskamm: This change was needed to get (geoserver)-side requests with env parameters working at all.
Before this change, server side generated and triggered requests lost their possibly given env parameters.
That may be the reason why the old animator endpoint implementation did not work with env parameters as mentioned by some users on the old blog post about that feature. 

I did not find any other server side code that produces GetMap Requests together with env-variables so this may be a long forgotten bug. Handling of client side requests seems to run through other parts of code which correctly handle env parameters.

I can link the matching parts of code if you need, just give me a sign

src/extension/wps-download/src/main/java/org/geoserver/wps/gs/download/DownloadAnimationProcess.java
weskamm: see discussion on top, still need to decide if it shall handle all possible params or just env params

doc/en/user/source/extensions/wps-download/mapAnimationDownload.rst
weskamm: i will have a look at supporting all parameters, but this may take some time

src/wms/src/main/java/org/geoserver/wms/RenderingVariables.java
aaime: We use environment injection in multiple places, it's definitely still working. Since it can be used in places other than WMS, it' s handled by a DispatcherCallback:

https://github.com/geoserver/geoserver/blob/6e9e25c0c7cdda9ada9f33f8255130d3afc76801/src/main/src/main/java/org/geoserver/ows/EnviromentInjectionCallback.java#L20

This allows injection across all protocols. However, during the animation bulding, the callback is not invoked, as the frame production does not go through the dispatcher to start with (guess the animator had a similar setup).

This code change is redundant in the WMS code path, could you move it in the animator itself instead?

doc/en/user/source/extensions/wps-download/mapAnimationDownload.rst
aaime: It's just a matter of amending the documentation, mind, the code seems to be generic enough as it is.

src/wms/src/main/java/org/geoserver/wms/RenderingVariables.java
weskamm: Could you give me a hint on how usage of the dispatcher would look like in this case?

weskamm: ping @aaime : I can see no obvious way how to do this, any suggestions?

aaime: "This code change is redundant in the WMS code path, could you move it in the animator itself instead?"
I meant, do not change the WMS code path, during normal WMS operations env vars are already working properly, you're addressing a situation that is specific to how the animator ends up calling the WMS classes.

Instead, do those calls inside the download process module, probably inside DownloadMapProcess, somewhere before calling GetMap.run. The issue is specific to gs-wps-download, does not happen in a normal WMS GetMap, so the solution should not mess with the WMS code.

weskamm:
Sorry but i am currently missing time to finish this PR. I hope to continue on this but will close for now

weskamm:
All right, i finally found time to finish this one.
It seems like i cannot reopen this PR, maybe someone else can do so?
Here is my updated fork: https://github.com/weskamm/geoserver/tree/animatevariables

@aaime please review once again, i addressed the issues you mentioned and removed the changed in the renderingVaraibles class

aaime:
"the animatevariables branch was force pushed or recreated" says Github, I cannot re-open either.
Believe you can open a new pull request starting from your branch though.

weskamm:
i opened a new one here https://github.com/geoserver/geoserver/pull/5271

PR #4951, opened 2021-04-08
Swedish translation improvements

bjornharrtell:
The swedish translation are using some strange words. Especially jarring is the translation of Data Store or Data Source to "Förråd" which should be "Datakälla".

I've opted to also use english terms for some technical things that are more appropriate IMO (as a native swede).


mprins (approved):
windows build failure appears to be unrelated as these files don't touch the GWC module and those errors are `java.io.FileNotFoundException`

PR #4976, opened 2021-04-15
Note about issue with cascaded WFS and GeoJSON

JakobMiksch:
references to this issue: https://osgeo-org.atlassian.net/browse/GEOS-10021#
Adds information about an unwanted behavior of the cascaded WFS. 

## Checklist

For all pull requests:

- [ ] Confirm you have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/main/CONTRIBUTING.md) 

### DIFF
--- a/doc/en/user/source/data/cascaded/wfs.rst
+++ b/doc/en/user/source/data/cascaded/wfs.rst
@@ -29,7 +29,7 @@ To connect to an external WFS, it is necessary to load it as a new datastore.  T
    * - :guilabel:`Enabled`
      - Enables the store.  If disabled, no data from the external WFS will be served.
    * - :guilabel:`GET_CAPABILITIES_URL`
-     - URL to access the capabilities document of the remote WFS. Providing `?version=1.0.0` or `?version=1.1.0` is required for beeing able to publish GeoJSON format. See `respective Issue <https://osgeo-org.atlassian.net/browse/GEOS-10021#>`_.
+     - URL to access the capabilities document of the remote WFS. Providing `&version=1.0.0` or `&version=1.1.0` is required for beeing able to publish GeoJSON format. See `respective Issue <https://osgeo-org.atlassian.net/browse/GEOS-10021#>`_.
    * - :guilabel:`PROTOCOL`
      - When checked, connects with POST, otherwise uses GET.
    * - :guilabel:`USERNAME`
--- a/doc/en/user/source/data/cascaded/wfs.rst
+++ b/doc/en/user/source/data/cascaded/wfs.rst
@@ -29,7 +29,7 @@ To connect to an external WFS, it is necessary to load it as a new datastore.  T
    * - :guilabel:`Enabled`
      - Enables the store.  If disabled, no data from the external WFS will be served.
    * - :guilabel:`GET_CAPABILITIES_URL`
-     - URL to access the capabilities document of the remote WFS.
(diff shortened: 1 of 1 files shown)


doc/en/user/source/data/cascaded/wfs.rst
```diff
@@ -29,7 +29,7 @@ To connect to an external WFS, it is necessary to load it as a new datastore.  T
    * - :guilabel:`Enabled`
      - Enables the store.  If disabled, no data from the external WFS will be served.
    * - :guilabel:`GET_CAPABILITIES_URL`
-     - URL to access the capabilities document of the remote WFS.
+     - URL to access the capabilities document of the remote WFS. Providing `&version=1.0.0` or `&version=1.1.0` is required for beeing able to publish GeoJSON format. See `respective Issue <https://osgeo-org.atlassian.net/browse/GEOS-10021#>`_.
```
bradh, line 32: being is spelt incorrectly.

```suggestion
     - URL to access the capabilities document of the remote WFS. Providing `&version=1.0.0` or `&version=1.1.0` is required to able to publish GeoJSON format. See `respective Issue <https://osgeo-org.atlassian.net/browse/GEOS-10021#>`_.

```

bradh (changes requested):
One typo.

I am dubious about this change at all, until its actually confirmed on JIRA.

mprins (changes requested):
I don't see any value in linking to a Jira ticket that describes a problem a user has and not a solution. 
Documentation should provide a description on what should be done. From the problem description it is not even clear if the upstream WFS produces a valid capabilities document.

Also this sets up documentation to go stale.

aaime:
Agree... Jira has 866 open tickets marked as "bug". Many are not confirmed, but still, I'm guessing a good percentage are legit, with varying degrees of severity and likeliness to stumble into them.

We cannot set a precedent to alter the documentation for every issue found, it would be un-manageable. I'd make an exception for cases where the developers have decided, for good reason, not to fix a given issue, then maybe the explanation would fit as documentation (open to debate).

This particular issue, if confirmed legit, just needs developer time/funding to be fixed. Belongs in Jira.

That said... wondering if we should link JIra from the docs sidebar, as it still is a good source of information... Something like this:
image

ianturton:
On Fri, 16 Apr 2021 at 09:08, Andrea Aime ***@***.***> wrote:

> Agree... Jira has 866 open tickets marked as "bug". Many are not
> confirmed, but still, I'm guessing a good percentage are legit, with
> varying degrees of severity and likeliness to stumble into them.
>
> We cannot set a precedent to alter the documentation for every issue
> found, it would be un-manageable. I'd make an exception for cases where the
> developers have decided, for good reason, not to fix a given issue, then
> maybe the explanation would fit as documentation (open to debate).
>
> This particular issue, if confirmed legit, just needs developer
> time/funding to be fixed. Belongs in Jira.
>

I've already said it is likely a bug/enhancement request but it is in an
unsupported GeoTools module and I for one don't feel like fixing it in my
free time. There is no really compelling reason to set the output format to
anything other than GML anyway.

That said... wondering if we should link JIra from the docs sidebar, as it
> still is a good source of information... Something like this:
> [image: image]
> <https://user-images.githubusercontent.com/325433/114993003-a4fb8a80-9e9b-11eb-9332-1fc2cf51f65b.png>
>

That is probably a good idea.

Ian

aaime:
Superceded by https://github.com/geoserver/geoserver/pull/4984

PR #5008, opened 2021-04-27
[GEOS-9844] Add standardDeviation classification for raster in SLDService module

taba90:
[GEOS-9844](https://osgeo-org.atlassian.net/browse/GEOS-9844)

This pr:

- adds standardDeviationClassification for raster data in the SLD Service

- consequently updates the capabilities endpoint

## Checklist

For all pull requests:

### DIFF
--- a/src/extension/sldService/src/main/java/org/geoserver/sldservice/utils/classifier/RasterSymbolizerBuilder.java
+++ b/src/extension/sldService/src/main/java/org/geoserver/sldservice/utils/classifier/RasterSymbolizerBuilder.java
@@ -281,6 +281,55 @@ public class RasterSymbolizerBuilder {
         return getColorMapFromBreaks(breaks, open, continuous, percentages);
     }
 
+    /**
+     * Builds a {@link ColorMap} based on equal intervals between the min and max value found in the
+     * raster
+     *
+     * @param image The source image
+     * @param intervals Number of resulting intervals
+     * @param continuous If the resulting ColorMap should be of type interval (discrete) or of type
+     */
+    public ColorMap standardDeviationClassification(
+            RenderedImage image, int intervals, boolean open, boolean continuous) {
+        ImageWorker iw = getImageWorker(image);
+        Statistics[][] stats = getStatistics(iw);
+        double mean = (double) stats[0][0].getResult();
+        double stddev = (double) stats[0][1].getResult();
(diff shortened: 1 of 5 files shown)


src/extension/sldService/src/main/java/org/geoserver/sldservice/utils/classifier/RasterSymbolizerBuilder.java
```diff
@@ -281,6 +281,55 @@ public ColorMap jenksClassification(
         return getColorMapFromBreaks(breaks, open, continuous, percentages);
     }
 
+    /**
+     * Builds a {@link ColorMap} based on equal intervals between the min and max value found in the
```
aaime, line 285: This seems to produce a set of intervals that are starting at the min value, and go up by "stddev".
If that's the case, then it's wrong. A standard deviation classification must be centered on the mean, and have breaks that start from it, and go up and down using the stddev as the driver for interval.
The GeoTools function used for [vector data is here](A standard deviation classification must be centered on the mean, and have breaks that start from it, and go up and down using the stddev as the interval), please check it out.

taba90: thanks for your review @aaime. The classification looks to me to be centered on the mean as it works mostly as the geotools one for vecrtors, check [here ](https://github.com/geoserver/geoserver/pull/5008/files/2b36870170243591220226b44554fdc4ea7f6941#diff-2f44a67dc3fdd17d03f4e5238058dc3eec211fc4f0e6e9763d22f0853feef951R318).
The thing that is different and that is the one probably generating confusion, is the way the first and last interval are generated: the geotools function return null for those value implicitly pointing to an openess of the class to - and + Infinity. In this case the method checks the min value is smaller then the second interval - stdev and the reverse is done for the last interval. If this is conceptually wrong I might change it to have - the max double for the first interval and the max double for the last one.

aaime: That would be better indeed. In general, the break points should not interact with the actual min and max of the dataset, they are purely defined by average and std deviation. A version of the code that does not raise eyebrowses, is one that is not using min and max at all.

PR #5014, opened 2021-04-30
[GEOS-9759]: Set Response Cache Headers in LayerGroups

zvrablik:
[GEOS-9759](https://osgeo-org.atlassian.net/browse/GEOS-9759) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

WMS/WMTS LayerGroups may configure HTTP max-age for all layers in the layer group. This configuration override the individual layer HTTP max-age settins. GWC uses minimal value of layerGroup  HTTP max-age and all layers in the layer group.

## Checklist

For all pull requests:


src/gwc/src/main/java/org/geoserver/gwc/layer/GeoServerTileLayer.java
```diff
@@ -1204,6 +1204,13 @@ public int getExpireClients(int zoomLevel) {
     /** Returns the max age of a layer group by looking for the minimum max age of its components */
     private int getGroupMaxAge(LayerGroupInfo lg) {
         int maxAge = Integer.MAX_VALUE;
+        if (isCachingEnabled(lg.getMetadata())) {
```
aaime, line 1207: So this sets an initial value for the group. However, it does not exit the method, meaning also allows for searching for a stricter value in nested layers and groups. Is this intended? I thought the group caching headers were supposed to override the single layer ones, instead as implemented, they provide just a starting pont.

src/web/core/src/main/java/applicationContext.xml
```diff
@@ -293,6 +293,13 @@
     <property name="componentClass" value="org.geoserver.web.publish.RootLayerConfig" />
   </bean>
 
+  <!-- LayerGroup configuration panels -->
+  <bean id="httpLayerGroupConfigPanel" class="org.geoserver.web.publish.LayerGroupConfigurationPanelInfo">
```
aaime, line 297: Would be good to have a test checking that the layer group caching panel shows up in the UI, and that submitting works (e.g. modify LayerGroupEditPageTest or LayerGroupPageTest)

src/wms/src/main/java/org/geoserver/wms/MapLayerInfo.java
```diff
@@ -79,16 +79,31 @@
     /** The extra constraints that can be set when an external SLD is used */
     private FeatureTypeConstraint[] layerFeatureConstraints;
 
+    private MetadataMap overrideMetadata;
```
aaime, line 82: This name is going to cause confusion, what you have here is not a generic override. See for example this snipped from line 370:

```java
    public static String getRegionateAttribute(FeatureTypeInfo layerInfo) {
        return layerInfo.getMetadata().get("kml.regionateAttribute", String.class);
    }
```

Here the metadata map is accessed directly. Should we use the "override"? No, we don't really want to, but the name suggests otherwise, the code has been given a metadata map override, why ignore it here.

Better to call it for what it is, a ``layerGroupMetadata``, and allow code to use it on a case by case basis.

src/wms/src/main/java/org/geoserver/wms/MapLayerInfo.java
```diff
@@ -230,6 +246,8 @@ public String getSRS() {
      * @return true if we should, false if we should omit the header
      */
     public boolean isCachingEnabled() {
+        boolean cachingEnabled = this.isCachingEnabled(this.overrideMetadata);
```
aaime, line 249: Again, seems confusing, see below.

src/wms/src/main/java/org/geoserver/wms/MapLayerInfo.java
```diff
@@ -240,12 +258,16 @@ public boolean isCachingEnabled() {
...
         }
         ResourceInfo resource = layerInfo.getResource();
-        MetadataMap metadata = resource.getMetadata();
-        Boolean cachingEnabled = null;
-        if (metadata != null) {
-            cachingEnabled = metadata.get(ResourceInfo.CACHING_ENABLED, Boolean.class);
-        }
-        return cachingEnabled == null ? false : cachingEnabled.booleanValue();
+
+        cachingEnabled = cachingEnabled || this.isCachingEnabled(resource.getMetadata());
```
aaime, line 262: If there is a group metadata, and it's behaving as an override, then only the group metadata is to be considered.
Otherwise, only the layer metadata should be considerd. Or-ing them would seem like they can both be considered, but it's not really the case, right?

src/wms/src/main/java/org/geoserver/wms/MapLayerInfo.java
```diff
@@ -257,12 +279,29 @@ public boolean isCachingEnabled() {
...
+        // maxAge may be deffined in override metadata, but caching
+        // may be disabled
+        if (this.isCachingEnabled(this.overrideMetadata)) {
+            maxAge = getCacheMaxAge(this.overrideMetadata);
         }
-        ResourceInfo resource = layerInfo.getResource();
-        Integer val = resource.getMetadata().get(ResourceInfo.CACHE_AGE_MAX, Integer.class);
-        return val == null ? 0 : val;
+
+        if (maxAge == 0) {
```
aaime, line 290: So if caching is enabled in the group, but max age is set to zero, then we consult the single layer settings?
Then it's not an override, but something in between. 
I'm not 100% sure, but is this consistent with the behavior in the tile layer case? Does not seem like it is.

doc/en/user/source/data/webadmin/layergroups.rst
```diff
@@ -158,3 +158,12 @@ Finally, it is possible to override the service settings and force a **Yes** to
 .. figure:: img/data_layers_root_in_capabilities.png
   
    Layer groups root layer in capabilities options
+
+HTTP Settings
+^^^^^^^^^^^^^
+
+Cache parameters that apply to the HTTP response from client requests.
```
aaime, line 165: Best to clearly specificy what is going to happen if both the group and the layers/grous nested in this group have caching headers setup. I believe the group headers should take precedence, since the request is asking for the group directly, but open to discussion.

aaime (changes requested):

src/gwc/src/main/java/org/geoserver/gwc/layer/GeoServerTileLayer.java
zvrablik: There was implemented approach pick smallest max age, I thought it will be better to follow same and pick the group max-age only if it is really shortest.
I am going to change it to pick either layer group http if exists or iterate to find min max age of all layers from group.

src/wms/src/main/java/org/geoserver/wms/MapLayerInfo.java
zvrablik: ok

src/web/core/src/main/java/applicationContext.xml
zvrablik: There were tests already in parent class. [Here](https://github.com/geoserver/geoserver/blob/4323399e06370758d8dec3b9022f01726b3390bb/src/web/core/src/test/java/org/geoserver/web/publish/HTTPLayerConfigTest.java#L68-L84)

Is this enough?

aaime: Is the parent class testing the integration of the panel with layer groups? I want to see the application context declaration is correctly picked up, and working, against a layer group.

src/wms/src/main/java/org/geoserver/wms/MapLayerInfo.java
zvrablik: I think in this case it is correct.
Caching is enabled in case the layerGroup caching is enabled or the layer info caching is enabled.
Could I leave it as it is?

src/wms/src/main/java/org/geoserver/wms/MapLayerInfo.java
zvrablik: Good point. Thanks. I changed it to return value from layer group metadata if enabled and do not continue to layer max age.

doc/en/user/source/data/webadmin/layergroups.rst
zvrablik: That was the discussion we had last week. Both could happen. Considering the other comments I think it should be using the layer group HTTP timeout and if not defined it should use layer HTTP timeout, with the exception of remote layers as it already was in the code.

aaime: A private discussion between us is no good to users. Please document it :-D

src/wms/src/main/java/org/geoserver/wms/MapLayerInfo.java
aaime: Yes it is technically correct, that's why I used confusing instead of incorrect. Put yourself in the context of someone jumping in on this code without much context, what can they think it does? I would like to see code that clearly reflects the documented behavior. Believe it's easier to understand if once checked if the group header is set, then that's used, and we exit the method.

doc/en/user/source/data/webadmin/layergroups.rst
```diff
@@ -158,3 +158,12 @@ Finally, it is possible to override the service settings and force a **Yes** to
...
+
+* **Response Cache Headers**— If selected, GeoServer will not request the same tile twice within the time specified in :guilabel:`Cache Time`. One hour measured in seconds (3600), is the default value for :guilabel:`Cache Time`. Layer group **Response Cache Headers** configuration replace **Response Cache Headers** configured in layers defined in layer group.
```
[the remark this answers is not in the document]
zvrablik: @aaime Thanks. I have updated the documentation.

aaime:
Backport as you see fit, mind it's a new feature so it either needs a 30 days wait or a vote for an early backport.

PR #5113, opened 2021-06-16
[GEOS-10116] authkey: Resolve REST role service config from environment

rcoup:
[GEOS-10116](https://osgeo-org.atlassian.net/browse/GEOS-10116)

[GEOS-9198](https://osgeo-org.atlassian.net/browse/GEOS-9198) / #3477 added  support for resolving env parameters in the authkey module configuration, but not for the REST role service. 

This extends support for resolving parameters in the following `GeoServerRESTRoleServiceConfig` properties:

* `baseUrl`
* `authApiKey`

## Checklist

For all pull requests:

- [X] Confirm you have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/main/CONTRIBUTING.md) 
- [X] Make sure the first PR targets the `main` branch, eventual backports will be managed later. This can be ignored if the PR is fixing an issue that only happens in a specific branch, but not in newer ones.

- [x] ~Commits changing the UI, existing user workflows, or adding new functionality, need to include documentation updates (screenshots, text)~
- [x] ~Commits changing the REST API, or any configuration object, should check if the REST API docs (Swagger YAML files and classic documentation) need to be updated.~

### DIFF
--- a/src/extension/authkey/src/main/java/org/geoserver/security/GeoServerRestRoleService.java
+++ b/src/extension/authkey/src/main/java/org/geoserver/security/GeoServerRestRoleService.java
@@ -30,6 +30,8 @@ import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import net.minidev.json.JSONArray;
+import org.geoserver.platform.GeoServerEnvironment;
+import org.geoserver.platform.GeoServerExtensions;
 import org.geoserver.security.config.SecurityNamedServiceConfig;
 import org.geoserver.security.event.RoleLoadedListener;
 import org.geoserver.security.impl.AbstractGeoServerSecurityService;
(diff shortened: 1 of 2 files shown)


afabiani (changes requested):
I really don't think this will be sufficient to achieve what stated into the ticket.

You will need to allow the GeoServerRESTRoleService class to resolve the values too, otherwise it will keep continue using the parametrized ones.

Moreover, why only two values have been parametrized?

rcoup:
> Moreover, why only two values have been parametrized?

AFAIK the main point of environment parameterisation is the properties that are dependent on the deployment? (that seemed to be the case for some other examples I saw when investigating). For my case, the URLs point to different services in different environments, but the other configuration of the module (paths, role names, etc) is the same. Encouraging `$` in environment variables seems a little fraught, but I can certainly add it... Note that the cache variables aren't able to be set via the UI atm.

> I really don't think this will be sufficient to achieve what stated into the ticket. You will need to allow the GeoServerRESTRoleService class to resolve the values too, otherwise it will keep continue using the parametrized ones.

Has worked in my testing... other uses of GeoServerEnvironment wrt Config classes don't seem to re-resolve it?

rcoup:
Ah. Maybe you mean if the user updates the URL in the web UI to a new parameterised value? I haven't explicitly checked that, will do that now.

rcoup:
@afabiani I've rearranged it if you could please take another look? 😃

afabiani (approved):

PR #5124, opened 2021-06-25
Revise template section into distinct example with a note about include

jodygarnett:
Some clarifications from gitter discussion

## Checklist

For all pull requests:


doc/en/user/source/community/ogc-api/features/index.rst
```diff
@@ -201,11 +217,15 @@ To override an OGC API Features template:
            </#list>
       <#include "common-footer.ftl">
 
+#. Many templates are constructed using ``#include``, for example :file:`collection.ftl` above uses ``<#include "common-header.ftl">`` located next to :file:`collections.ftl`.
+
+   Presently each family of templates manages its own :file:`common-header.ftl` (as shown in the difference between :file:`ogc/features` service templates, and getfeature templates above).
+
 #. A restart is required, as templates are cached.
```
aaime, line 224: I'm not sure this is actually the case, I've been changing templates a lot without having to restart during development. Wondering if it's somehow context specific.

Everything other change looks ok.

jodygarnett: It would be great if the template was not cached, I did not figure out what was happening.

aaime: Templates are cached through all GeoServer (the classic Freemarker ones for WMS have always been) and are listening to file changes. I cannot reproduce the issue you're trying to report in the docs. Can you confirm it's actually happening on your end? How does it happen?

aaime: @jodygarnett mind, above I was not asking about a diagnosis, but about steps to reproduce.

PR #5128, opened 2021-07-03
Update imagepyramid.rst: remove JPEG compression

syldor:
Following this tutorial leads to an error as imagepyramid does not seem to handle JPEG compression, per here: https://gis.stackexchange.com/questions/213257/problem-with-rendering-jpeg-compressed-image-caused-by-java-lang-classnotfoun/213384#213384

```
03 Jul 03:13:44 WARN [io.imageio] - Unable to compute source area for URL file:/home/syldor/bmpyramid/0/kornelimuenster_Ortho_projection_2_01_01.tif
javax.imageio.IIOException: Unsupported Image Type
	at java.desktop/com.sun.imageio.plugins.jpeg.JPEGImageReader.readInternal(JPEGImageReader.java:1139)
	at java.desktop/com.sun.imageio.plugins.jpeg.JPEGImageReader.read(JPEGImageReader.java:1110)
	at it.geosolutions.imageioimpl.plugins.tiff.TIFFJPEGDecompressor.decodeRaw(TIFFJPEGDecompressor.java:282)
	at it.geosolutions.imageio.plugins.tiff.TIFFDecompressor.decode(TIFFDecompressor.java:2637)
	at it.geosolutions.imageioimpl.plugins.tiff.TIFFImageReader.decodeTile(TIFFImageReader.java:1764)
```

### DIFF
--- a/doc/en/user/source/tutorials/imagepyramid/imagepyramid.rst
+++ b/doc/en/user/source/tutorials/imagepyramid/imagepyramid.rst
@@ -21,7 +21,7 @@ In order to build the pyramid we'll use the `gdal_retile.py <http://www.gdal.org
 The following commands will build a pyramid on disk::
 
    mkdir bmpyramid
-   gdal_retile.py -v -r bilinear -levels 4 -ps 2048 2048 -co "TILED=YES" -co "COMPRESS=JPEG" -targetDir bmpyramid bmreduced.tiff
+   gdal_retile.py -v -r bilinear -levels 4 -ps 2048 2048 -co "TILED=YES" -targetDir bmpyramid bmreduced.tiff
    
 The `gdal_retile.py  <http://www.gdal.org/gdal_retile.html>`_ user guide provides a detailed explanation for all the possible parameters, here is a description of the ones used in the command line above:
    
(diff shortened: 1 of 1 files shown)


bradh:
I think this is possibly OK, but the assumption is bad. In particularly, JPEG does work. However JPEG+Alpha won't work, and whether that turns up in the retiled geotiff depends on the input data. I assume you aren't trying it with the blue marble that the tutorial specifies (because the source is gone) and you are using some other data instead.

aaime:
Yep, the RGB+Alpha thing that GDAL does is weird, they set the color model as CMYK but then interpret it as RGBA. Seems like a vendor specific thing to me. Looking for funding in order to implement support for it, as weird as it looks, it's useful.

syldor:
@bradh you're right I'm using with another dataset, thank you for your explanation.

@aaime who is vendor here, GDAL? Or Images providers?

bradh:
@syldor vendor specific means "something outside the core spec". Do you want to update this PR?

syldor:
@bradh I would be ok to update the PR but I'm not too sure what to write, so maybe I close it down.

PR #5129, opened 2021-07-05
[GEOS-10133] Connecting to WMS Service via Http Proxy

roarbra:
[GEOS-10133](https://osgeo-org.atlassian.net/browse/GEOS-10133)

Fixes the bug referenced in the JIRA. Didn't see immediately how to write a unit test for this specific problem though. It would involve a running proxy server and wms server, or some mocking to fully test the functionality. The introduced test makes the assumption that we will get an UnknownHostException for the missing proxy server.

## Checklist

For all pull requests:

- [X] Confirm you have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/main/CONTRIBUTING.md) 
- [X] Make sure the first PR targets the `main` branch, eventual backports will be managed later. This can be ignored if the PR is fixing an issue that only happens in a specific branch, but not in newer ones.

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
+++ b/src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
@@ -96,7 +96,6 @@ import org.geotools.gml2.GML;
 import org.geotools.http.HTTPClient;
 import org.geotools.http.HTTPClientFinder;
 import org.geotools.http.HTTPConnectionPooling;
-import org.geotools.http.SimpleHttpClient;
 import org.geotools.measure.Measure;
 import org.geotools.ows.wms.Layer;
 import org.geotools.ows.wms.WMSCapabilities;
@@ -1941,9 +1940,7 @@ public class ResourcePool {
                 mtClient.setMaxConnections(maxConnections);
             }
         } else {
-            client =
-                    HTTPClientFinder.createClient(
-                            new Hints(Hints.HTTP_CLIENT, SimpleHttpClient.class));
+            client = HTTPClientFinder.createClient();
         }
         String username = info.getUsername();
         String password = info.getPassword();
--- a/src/main/src/test/java/org/geoserver/catalog/ResourcePoolTest.java
+++ b/src/main/src/test/java/org/geoserver/catalog/ResourcePoolTest.java
@@ -37,6 +37,7 @@ import java.io.Serializable;
 import java.lang.reflect.Field;
 import java.net.URI;
 import java.net.URL;
+import java.net.UnknownHostException;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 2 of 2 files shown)


src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -1941,9 +1940,7 @@ private HTTPClient getHTTPClient(HTTPStoreInfo info) {
                 mtClient.setMaxConnections(maxConnections);
             }
         } else {
-            client =
-                    HTTPClientFinder.createClient(
-                            new Hints(Hints.HTTP_CLIENT, SimpleHttpClient.class));
+            client = HTTPClientFinder.createClient();
```
aaime, line 1943: From a mere user point of view, the two calls seem equivalent. But I guess they are not? What is happening here?

roarbra:
When specifyng a specific client, you will get that irrespective of any
other settings. Without the http_client hint you’re open for any client,
and will get the one that suits your settings. Such as a proxy setting.

Best regards,

Roar Brænden

man. 12. jul. 2021 kl. 15:18 skrev Andrea Aime ***@***.***>:

aaime:
But then it means the store could get a multithreaded client even if not requested during configuration? That's a no-go then, we want to keep the choice (depending on the environment, one might work and the other not).

aaime:
@roarbra is there another way to solve this issue, while ensuring the client is the one chosen by the user while setting up WMS/WFS cascading?

roarbra:
@aaime As long as the SimpleHttpClient doesn't support proxy servers, and we want to respect the user's decision to not use a multithreaded client, we must inform the user about the problem, and let him choose to use multithreaded client.

How do we get a nice clean error message with red borders within Geoserver?

aaime:
I guess my lack of familiary with the classes is starting to show but... the reported talked about using the JVM settings:
"-Dhttp.proxySet=true -Dhttp.proxyHost=our.proxy.de -Dhttp.proxyPort=8080 -Dhttps.proxyHost=our.proxy.de -Dhttps.proxyPort=8080"
And the SimpleHttpClient was uses URL class, which in turn uses the above settings [according to the Java own docs](https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html)... it's something we used in the past too, in fact, at one point we had to fix the multithreaded client because it was ignoring those same variables. 

The reporter also points at a _regression_ between 2.18.x and 2.19.x, so the new client code must be broken in some way.

roarbra:
Oh, then we should just add HTTPProxy as an implementation of SimpleHttpClient.

roarbra:
This PR should be closed. There is a new issue [GEOT-6959](https://osgeo-org.atlassian.net/jira/software/c/projects/GEOT/issues/GEOT-6959?filter=reportedbyme), that should solve the problem.

PR #5161, opened 2021-07-20
[GEOS-10032] Group Layer in Catalag Mode Hide not in capabilities when unauthenticated

stefan-overkamp:
[GEOS-10032](https://osgeo-org.atlassian.net/browse/GEOS-10032)

This pull request solves the issue, that layergroups are not output to capabilities in catalog mode hide, when the user is not the admin user. The reason lies in a not applicable filter put on layergroups by DefaultResourceAccessManager.

## Checklist

For all pull requests:

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
+++ b/src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
@@ -550,7 +550,7 @@ public class DefaultResourceAccessManager implements ResourceAccessManager {
             String wsNameProperty;
             if (LayerGroupInfo.class.isAssignableFrom(clazz)) {
                 // resource.store.workspace.name is not applicable for layergroups
-                wsNameProperty = "workspace.name";           
+                wsNameProperty = "workspace.name";
             } else if (PublishedInfo.class.isAssignableFrom(clazz)) {
                 wsNameProperty = "resource.store.workspace.name";
             } else {
--- a/src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
+++ b/src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
@@ -548,7 +548,10 @@ public class DefaultResourceAccessManager implements ResourceAccessManager {
 
             // get the right ws property name
             String wsNameProperty;
-            if (PublishedInfo.class.isAssignableFrom(clazz)) {
+            if (LayerGroupInfo.class.isAssignableFrom(clazz)) {
+                // resource.store.workspace.name is not applicable for layergroups
+                wsNameProperty = "workspace.name";           
+            } else if (PublishedInfo.class.isAssignableFrom(clazz)) {
                 wsNameProperty = "resource.store.workspace.name";
             } else {
                 wsNameProperty = "store.workspace.name";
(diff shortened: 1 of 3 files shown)


aaime:
Hi @stefan-overkamp, thanks for the fix.

A test is mandatory before the PR can be considered for merge.
Maybe you can add one in [SecureCatalogImplTest](https://github.com/geoserver/geoserver/blob/main/src/security/security-tests/src/test/java/org/geoserver/security/impl/SecureCatalogImplTest.java)?

stefan-overkamp:
@aaime Sorry, but I think I have not the knowledge to write the test.
I looked into [SecureCatalogImplTest](https://github.com/geoserver/geoserver/blob/main/src/security/security-tests/src/test/java/org/geoserver/security/impl/SecureCatalogImplTest.java) especially into testWMSLayerGroupDenyWSAllow.

The layergroup nurc.wsContainerD is available by the workspace rule nurc.*.r=*

I have set a breakpoint in SecureCatalogImpl [line: 1475]	
CloseableIterator<T> filtered = delegate.list(of, securityFilter, offset, count, sortBy);
and did a debug run of SecureCatalogImplTest.

The assertNotNull(sc.getLayerGroupByName("nurc", "wsContainerD")); never steps to my breakpoint.
So no securityFilter is applied.

Then I made debug run of geoserver with the release datadir and added a group layer "test" in the nurc workspace.
Connecting with QGIS (to http://localhost:8080/geoserver/wms or http://localhost:8080/geoserver/nurc/wms, isolated)
stepped into the  breakpoint. The filter was (without my fix):
[Filter.INCLUDE AND [ true = InternalFunctionImpl() ] AND [ resource.store.workspace.name = nurc ] AND Filter.INCLUDE AND [ true = InternalFunctionImpl() ] AND Filter.INCLUDE]

with  "resource.store.workspace.name = nurc" no layergroup are in the capabilities document, only the contained layers.

Conclusion. The assertNotNull(sc.getLayerGroupByName("nurc", "wsContainerD")); 
test is not asserting that layergroups are in the capabilities.

Hints for a test:
I am asking for layergroups in an isolated workspace accessible by wsname.*.r=* as unauthenticated user.
- SecurityContextHolder.getContext().setAuthentication(null)
- the catalog mode is HIDE

Maybe someone else could have a look into SecureCatalogImpl 
If I could donate some money for up to 3 hours work, please inform me at overkamp@posteo.de

aaime:
The issue here is that you're looking at methods accessing a single layer groups, they are not affected by the filtering issue.
``ResourceAccessManager.getSecurityFilter`` is used only by the ``Catalog.list`` methods, which is there to make external catalog storage (e.g. ``JDBCConfig``) possible when dealing with many layers. Think a situation where there are 10k layers (not that uncommon). With the default catalog, they are all in memory, but there is a single copy of them, when a list gets retrieved, it's just duplicated pointers.
However, when working against an externally stored catalog, a request asking to get all the layers would retrieve them from the database, causing all of them to be instantiated in memory, once for every request. 10 concurrent requests would cause 100k layer objects to be created, and so on.

The ``Catalog.list`` method offers an escape route, providing an iterator, only the current layer object is created, while the others will be allocated later.

Long story short, in order to test your changes you have to call ``Catalog.list`` instead. There is a catch though... the test delegates to a mock catalog, which is not initialized to support that list call... oh well, I got close enough to the needed patch just to research the issue, that I'll push what I have on this PR.

aaime:
Looks everything is green. Are you happy with the explanation and test @stefan-overkamp ?

overkamp:
Great, many thanks for the explanation and the test. I'm happy.

aaime:
I've squash merged the PR into a single commit: https://github.com/geoserver/geoserver/commit/3fbc1cceee6be7150d9937e2dd69785bdf19aa17

That will make the fix available in 2.20.0 in September. If you want it into 2.19.x and 2.18.x backports of that commit to the respective branches are needed. You can do pull requests against those branches cherry-picking the above commit.
In case you care about 2.18.x, mind, it is about to end its life, the last release will be 2.18.5, in the second half of August.

PR #5182, opened 2021-07-27
Added non thread-safe warning

petersmythe:
# Checklist

### DIFF
--- a/doc/en/user/source/rest/index.rst
+++ b/doc/en/user/source/rest/index.rst
@@ -9,6 +9,7 @@ REST is an acronym for "`REpresentational State Transfer <http://en.wikipedia.or
 
 Operations on resources are implemented with the standard primitives of HTTP:  GET to read; and PUT, POST, and DELETE to write changes. Each resource is represented as a URL, such as ``http://GEOSERVER_HOME/rest/workspaces/topp``.
 
+.. warning:: While the OGC requests are thread-safe and can (actually must) be called in parallel, the REST write methods (PUT, POST, DELETE) are `not thread-safe <https://files.speakerdeck.com/presentations/2b500ab2712d44129a17f4b7e28c17df/slide_20.jpg>`.  Care must be taken to call these sequentially and synchronously, otherwise you risk corrupting your configuration files.
 
 API
 ---


doc/en/user/source/rest/index.rst
```diff
@@ -9,6 +9,7 @@ REST is an acronym for "`REpresentational State Transfer <http://en.wikipedia.or
 
 Operations on resources are implemented with the standard primitives of HTTP:  GET to read; and PUT, POST, and DELETE to write changes. Each resource is represented as a URL, such as ``http://GEOSERVER_HOME/rest/workspaces/topp``.
 
+.. warning:: While the OGC requests are thread-safe and can (actually must) be called in parallel, the REST write methods (PUT, POST, DELETE) are `not thread-safe <https://files.speakerdeck.com/presentations/2b500ab2712d44129a17f4b7e28c17df/slide_20.jpg>`.  Care must be taken to call these sequentially and synchronously, otherwise you risk corrupting your configuration files.
```
aaime, line 12: It's true that the catalog is not thread safe, but the REST API is protected by a global read/write lock, so no care should actually be needed, [write methods are getting serialized automatically](https://github.com/geoserver/geoserver/blob/6e9e25c0c7cdda9ada9f33f8255130d3afc76801/src/restconfig/src/main/java/org/geoserver/rest/catalog/RestConfigurationLockCallback.java#L25). Well, unless the admin was reckless and disabled the locks via the [undocumented system variable](https://github.com/geoserver/geoserver/blob/b76bba7178d3505cd45e318e659c3b32faee4b5e/src/main/src/main/java/org/geoserver/GeoServerConfigurationLock.java#L50).

aaime: By the way, the issue goes deeper than mere thread safety... a configuration operation can touch multiple configuration objects, in order to do it consistently we'd need a notion of transactions at the catalog level: would mean rolling new API to interact with the catalog, and some sort of transactional in memory storage for the default catalog implementation.
The current global lock is a "poor man's" serializable transaction protection level.

petersmythe: Hmm, thank you for that insight @aaime. I will have to investigate our corrupt config files problem further then.  (No, the unknown locks were not disabled).

Please pause on this PR for now.

aaime: Converted to draft so that it won't get accidentaly merged.

The lock is not an unbreakable armor.. just protects REST and GUI operations. A WPS process could get through it. Another possible cases for corrupted config files are OOMs happening while the file is being written, for example.

aaime:
No progress in a while, closing. More doc updates are welcomed! :-D (and also bug reports about concurrency issues in the catalog, best is reproducible)

PR #5187, opened 2021-07-29
[GEOS-10174] FeatureWrapper doesn't handle feature attributes max number greater then 1

taba90:
This pr fixes a bug in free marker FeatureWrapper that is not adding all the Attributes of same type present in a Feature to the free marker context.

  

# Checklist


src/community/ogcapi/ogcapi-features/src/test/java/org/geoserver/ogcapi/features/ComplexFeaturesTest.java
```diff
@@ -0,0 +1,26 @@
+package org.geoserver.ogcapi.features;
```
aaime, line 1: Missing copyright header

src/community/ogcapi/ogcapi-coverages/src/main/resources/applicationContext.xml
```diff
@@ -23,7 +23,7 @@
 
 
     <!-- Bridges to have OGC services show up as OWS services too -->
-    <bean id="featuresServiceFactory" class="org.geoserver.ogcapi.APIServiceFactoryBean">
+    <bean id="coveragesServiceFactory" class="org.geoserver.ogcapi.APIServiceFactoryBean">
```
aaime, line 26: Yup, my bad.

src/main/src/main/java/org/geoserver/template/FeatureWrapper.java
```diff
@@ -347,16 +347,24 @@ public Set entrySet() {
...
                 Name attName;
-                Map attributesMap;
                 for (PropertyDescriptor type : types) {
                     attName = type.getName();
-                    attributesMap = new AttributeMap(attName, feature);
-                    entrySet.add(
-                            new MapEntry<Object, Object>(attName.getLocalPart(), attributesMap));
+                    if (type.getMaxOccurs() > 1) {
+                        for (Property property : feature.getProperties(attName)) {
+                            addAttributeMap(attName, property);
```
aaime, line 354: So this will cause the entryset to have multiple entries with the same key, effectively making this a multimap? However the rest of the map machinery won't work properly, e.g., if one goes and asks for a key directly (it's possible to use an attribute name directly too in the templates, see usage of "feature.NAME" or "feature.THUMBNAIL" [here](https://docs.geoserver.org/latest/en/user/tutorials/GetFeatureInfo/html.html#custom-templates)).

I'm guessing this works in the templates at hand, because they end up iterating the entry set directly, which is going to return the list of entries produced here.

Thinking out loud, this might need a different treatment, like storing the multiple values in a list instead.

taba90: fixed

src/community/ogcapi/ogcapi-features/src/test/java/org/geoserver/ogcapi/features/ComplexFeaturesTest.java
taba90: fixed

taba90:
thanks for your review @aaime feedback have been applied.

PR #5189, opened 2021-07-30
[GEOS-10177] Allow Default Translation

ydeliorman:
[GEOS-10177](https://osgeo-org.atlassian.net/browse/GEOS-10177)

This pull request adds the possibility to have a default translation that is not associated with any language. 
  
A test is removed from [https://github.com/geoserver/geoserver/blob/5fc7333730a59c19d1319a1d568af562c740adaa/src/web/core/src/test/java/org/geoserver/web/data/resource/ResourceConfigurationPageTest.java#L89](url) because language is not a required field now.

## Checklist

For all pull requests:


src/web/core/src/main/java/org/geoserver/web/data/resource/InternationalStringPanel.java
```diff
@@ -301,21 +300,19 @@ private boolean i18nVisible() {
...
-                    String message =
-                            new StringResourceModel(
-                                            "InternationalStringPanel.duplicatedLocale",
-                                            InternationalStringPanel.this)
-                                    .getString();
-                    ValidationError error = new ValidationError();
-                    error.setMessage(message);
-                    iValidatable.error(error);
-                }
+            long count =
```
taba90, line 303: Here should not be a matter of simply remove the if condition, but the method should be validation also that there are no duplicated entries with null Locale. Once this is fixed I guess that the changes made in the related geotools pr made to GrowableInternationalString add method can be removed, leaving just the copy constructor fix.

doc/en/user/source/services/internationalization/index.rst
```diff
@@ -31,6 +31,10 @@ The response content language can be selected using the ``AcceptLanguages`` requ
 
 * The internationalized elements will be titles, abstracts and keywords.
 
+* If no language is defined after checking the i18n box, the text will be attached to a no Locale, which is the default language.
+
+* If no language code is specified, text value with default language will be returned.
```
taba90, line 36: I would remove these entries here and add the explanation in the `Default Language `section down here.
Should be made clear that: from UI it is possible to set a i18n content with null language and that that content will be used in case no default language was specified from Service/Global settings.
Moreover the part telling that GeoServer will try to use the server default locale should go since now the `GeoServerDefaultLocale `will instead return null.

doc/en/user/source/styling/sld/language.rst
```diff
@@ -52,4 +54,4 @@ Here is an example providing labels in multiple languages, integrating the ``lan
          </TextSymbolizer>
 
 
-If the Function has been defined in a style and the ``LANGUAGE`` parameter has not been specified in the request or has value ``LANGUAGE=*`` the default language will be used. See :ref:`internationalization` for details on ``Default Language``.
+If the Function has been defined in a style and the ``LANGUAGE`` parameter has not been specified in the request or has value ``LANGUAGE=*`` the default value will be returned. If there is no default value, default language will be returned. See :ref:`internationalization` for details on ``Default Language``.
```
taba90, line 57: I would put it differently telling that from the example it is possible to see that by putting an empty property name it is possible to match a value with a notExisting/* language parameter.

taba90 (changes requested):
Added some feedback on documentation and on the validation method for duplicated entries in the UI component.
There are builds failures that seem legit I guess will be succesfull when the geotools pr is merged.
```
Error: 8,103 [ERROR] Tests run: 16, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 5.569 s <<< FAILURE! - in org.geoserver.wms.web.data.WMSAdminPageTest
Error: 8,105 [ERROR] testEmptyInternationalContent(org.geoserver.wms.web.data.WMSAdminPageTest)  Time elapsed: 0.732 s  <<< FAILURE!
java.lang.AssertionError: 
expected no error message, but contains
   The Text Content for an international field is required
   The Text Content for an international field is required
	at org.geoserver.wms.web.data.WMSAdminPageTest.testEmptyInternationalContent(WMSAdminPageTest.java:310)
```

doc/en/user/source/services/internationalization/index.rst
ydeliorman: Done.

doc/en/user/source/styling/sld/language.rst
ydeliorman: Done

src/web/core/src/main/java/org/geoserver/web/data/resource/InternationalStringPanel.java
```diff
@@ -257,12 +272,25 @@ private GrowableInternationalString updateGrowableString() {
...
         List<GrowableStringModel.InternationalStringEntry> items = provider.getItems();
         for (GrowableStringModel.InternationalStringEntry entry : items) {
-            growableString.add(entry.getLocale(), entry.getText());
+            if (canAddElement(items, entry)) growableString.add(entry.getLocale(), entry.getText());
         }
         growableModel.setObject(growableString);
         return growableString;
     }
 
+    private boolean canAddElement(
```
aaime, line 281: Can you explain what is happening here?

src/wms/src/main/java/org/geoserver/wms/WMSInfoImpl.java
```diff
@@ -231,26 +230,26 @@ public void setGetMapMimeTypeCheckingEnabled(boolean getMapMimeTypeCheckingEnabl
...
     public String getRootLayerTitle() {
-        if (rootLayerTitle == null && internationalRootLayerTitle != null)
-            return internationalRootLayerTitle.toString(GeoServerDefaultLocale.get());
         return rootLayerTitle;
     }
 
     @Override
     public void setRootLayerTitle(String rootLayerTitle) {
-        this.rootLayerTitle = rootLayerTitle;
+        if (rootLayerTitle != null) {
```
aaime, line 238: This looks broken, one should be able to set a null title. Same for the abstract below.

src/wms/src/test/java/org/geoserver/wms/capabilities/GetCapabilitiesReponseTest.java
```diff
@@ -207,4 +207,29 @@ public void testAcceptLanguagesParameter() throws Exception {
                 "http://localhost:8080/geoserver/wms?request=GetLegendGraphic&format=image%2Fpng&width=20&height=20&layer=nature&Language=it",
                 "//Layer[Name='nature']/Style/LegendURL/OnlineResource/@xlink:href", dom);
     }
+
+    @Test
+    public void testNullLocale() throws Exception {
```
aaime, line 212: What is this testing? that the backlinks do not contain language if the locale has not been provided?

src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/GetMapIntegrationTest.java
```diff
@@ -2390,6 +2390,24 @@ public void testMultiLanguageStyle() throws Exception {
...
+        catalog.save(places);
+        String url =
+                "wms?LAYERS="
+                        + places.getName()
+                        + "&STYLES=multiLanguageStyle&FORMAT=image%2Fpng"
+                        + "&SERVICE=WMS&VERSION=1.1.1&REQUEST=GetMap&SRS=EPSG%3A4326&WIDTH=256"
+                        + "&HEIGHT=256&BBOX=0.0000,-0.0020,0.0035,0.0010";
+        BufferedImage image = getAsImage(url, "image/png");
+        URL urlPng = getClass().getResource("nolang_result.png");
+        ImageAssert.assertEquals(new File(urlPng.toURI()), image, 500);
```
aaime, line 2408: A threshold of 500 is a big one, for an image that does not have any label in it. The line rendering should only change slightly across platforms (thinner or thicker, antialiasing done in a different way). I'd go for a smaller  value, like 50?

aaime (changes requested):
See comments below.

doc/en/user/source/services/internationalization/index.rst
```diff
@@ -53,4 +53,16 @@ GeoServer allows defining a default language to be used when international conte
 
 .. figure:: img/DefaultLanguage.png
 
-GeoServer will first check for a service specific default locale. If not found, it will use the one in the global settings. If  the global default language has not been specified, GeoServer will use the server default locale.
+Alternatively from the i18n editor it is possible to set an i18n entry with empty language for each configurable field.
```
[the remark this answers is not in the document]
taba90: fixed

doc/en/user/source/styling/sld/language.rst
```diff
@@ -52,4 +54,4 @@ Here is an example providing labels in multiple languages, integrating the ``lan
          </TextSymbolizer>
 
 
-If the Function has been defined in a style and the ``LANGUAGE`` parameter has not been specified in the request or has value ``LANGUAGE=*`` the default language will be used. See :ref:`internationalization` for details on ``Default Language``.
+From the example it is possible to see that by putting an empty property name it is possible to match a value with a notExisting/* language parameter. If there is no default value, default language will be returned. See :ref:`internationalization` for details on ``Default Language``.
```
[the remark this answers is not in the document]
taba90: fixed

src/main/src/main/java/org/geoserver/catalog/impl/LayerGroupInfoImpl.java
```diff
@@ -402,35 +402,23 @@ private void checkMetadataNotNull() {
...
     }
 
     @Override
     public void setInternationalTitle(InternationalString internationalTitle) {
-        GrowableInternationalString growable;
-        if (internationalTitle == null) growable = new GrowableInternationalString(getTitle());
-        else growable = new GrowableInternationalString(internationalTitle);
-
-        this.internationalTitle = growable;
+        if (internationalTitle != null)
```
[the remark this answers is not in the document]
taba90: Fixed thanks for your point.

src/web/core/src/main/java/org/geoserver/web/data/resource/InternationalStringPanel.java
taba90: The support of null locale, forced the validation for double null locale entries to be delayed after the wicket validation phase, doing it before the submit (during wicket validation phase in the IValidator, it is not possible to tell when a Locale is null because a new entry was added or because it is meant to be null).
This in turn causes the fact that when the user plays with the UI a newly added field with null locale entry might be added to the GrowableString. Before the null locale support the null values where filter out before adding entires to the Growable instance, but now this might cause entries with null key to be duplicated and make the Growable throw an exception.
I've added a clarifying comment over the method.

src/wms/src/main/java/org/geoserver/wms/WMSInfoImpl.java
taba90: fixed.

src/wms/src/test/java/org/geoserver/wms/capabilities/GetCapabilitiesReponseTest.java
taba90: fixed.

src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/GetMapIntegrationTest.java
taba90: fixed.

taba90:
thanks for your review @aaime feedbacks applied.

aaime:
Squash merged on master

PR #5191, opened 2021-08-02
[GEOS-10181] Features-templatings avoid replacing OGCAPI HTML response if template do not exists

taba90:
This pr fixes a bug causing features templating to replaces always ogcapi-features HTML response, even without a defined template for the requested featuretype.

  

# Checklist


src/community/features-templating/features-templating-wfs/src/main/java/org/geoserver/featurestemplating/wfs/HTMLTemplateResponse.java
```diff
@@ -115,4 +119,35 @@ protected void afterEvaluation(TemplateOutputWriter writer, RootBuilder root, Fe
             throws IOException {
         super.afterEvaluation(writer, root, feature);
     }
+
+    @Override
+    protected boolean canHandleInternal(Operation operation) {
```
fernandor777, line 124: Please add some comments regarding what this method does on conditions and operations.  Even better if we can split some of them on private methods.

src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/MessageConverterResponseAdapter.java
```diff
@@ -106,8 +106,13 @@ protected Operation getOperation(T result, Request dr, MediaType mediaType) {
                 .stream()
                 .filter(
                         r ->
-                                getMediaTypeStream(r).anyMatch(mt -> mediaType.isCompatibleWith(mt))
-                                        || (r.canHandle(op) && r.getBinding().isInstance(result)))
+                                r.canHandle(op)
+                                        && (getMediaTypeStream(r)
+                                                        .anyMatch(
+                                                                mt ->
```
fernandor777, line 112: This stream chain became hard to read, please refactor the nested filter condition in a separated private method and add Javadoc comment about what it does.

fernandor777 (changes requested):
Looks good, some minor code improvements requests added.

src/community/features-templating/features-templating-wfs/src/main/java/org/geoserver/featurestemplating/wfs/HTMLTemplateResponse.java
taba90: fixed

src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/MessageConverterResponseAdapter.java
taba90: fixed

fernandor777 (approved):
Looks good, thanks for the improvements!

PR #5208, opened 2021-08-10
[GEOS-10185] LayerGroup legend internationalization styles returns multiple values

ydeliorman:
[GEOS-10185](https://osgeo-org.atlassian.net/browse/GEOS-10185)

This pull request fixed the layer titles returning multiple internationalized values.
  

# Checklist


src/main/src/main/java/org/geoserver/data/InternationalContentHelper.java
```diff
@@ -67,6 +67,20 @@ public InternationalContentHelper(
         verify();
     }
 
+    public InternationalContentHelper(String language) {
```
taba90, line 70: The LegendGraphicRequest gets the Locale set from the Kvp reader.
I guess it is better to stick to that one instead of adding a constructor to the InternationalContentHelper that is more bound the GetCapabilities i18n support.

src/wms/src/main/java/org/geoserver/wms/legendgraphic/BufferedImageLegendGraphicBuilder.java
```diff
@@ -126,8 +133,30 @@ public BufferedImage buildLegendGraphic(GetLegendGraphicRequest request)
         // a layer group is given)
         setup(request);
         List<RenderedImage> layersImages = new ArrayList<>();
+
+        Catalog catalog = (Catalog) GeoServerExtensions.bean("catalog");
```
taba90, line 137: instead of doing so you can simply retrieve the Locale already set in the [GetLegendGraphicRequest](https://github.com/geoserver/geoserver/blob/de184b2f9ed9a844b5291dd163e772132a07689b/src/wms/src/main/java/org/geoserver/wms/GetLegendGraphicRequest.java#L582) find the featureTypeInfo from catalog and apply the toString(Locale) on that one. 
Since the GrowableInternationalString toString method is returning always a result, I would also add a check that the requested locale is actually present among the ones hold by the internationalTitle and if not use the GeoServerDefaultLocale.

taba90 (changes requested):
Looking at the code I would create a new constructor having a Locale as a parameter in the InternationalContentHelper.
Also the Legend has a JSON output format too. I would then move the i18n logic into a separate method in the [abstract class ](https://github.com/geoserver/geoserver/blob/3f9a2a315516daeb719bfbba39b3957bba1f98f4/src/wms/src/main/java/org/geoserver/wms/legendgraphic/LegendGraphicBuilder.java#L62) and use it also in the for loop of the [JsonLegendGraphicBuilder](https://github.com/geoserver/geoserver/blob/3f9a2a315516daeb719bfbba39b3957bba1f98f4/src/wms/src/main/java/org/geoserver/wms/legendgraphic/JSONLegendGraphicBuilder.java#L268)

src/main/src/main/java/org/geoserver/data/InternationalContentHelper.java
ydeliorman: done

src/wms/src/main/java/org/geoserver/wms/legendgraphic/BufferedImageLegendGraphicBuilder.java
ydeliorman: done

ydeliorman:
Reviews are done.

taba90:
> Looking at the code I would create a new constructor having a Locale as a parameter in the InternationalContentHelper.
> Also the Legend has a JSON output format too. I would then move the i18n logic into a separate method in the [abstract class ](https://github.com/geoserver/geoserver/blob/3f9a2a315516daeb719bfbba39b3957bba1f98f4/src/wms/src/main/java/org/geoserver/wms/legendgraphic/LegendGraphicBuilder.java#L62) and use it also in the for loop of the [JsonLegendGraphicBuilder](https://github.com/geoserver/geoserver/blob/3f9a2a315516daeb719bfbba39b3957bba1f98f4/src/wms/src/main/java/org/geoserver/wms/legendgraphic/JSONLegendGraphicBuilder.java#L268)

The actually implemented solution turned out to be setting directly the localized label in the [MapLayerInfo](https://github.com/geoserver/geoserver/pull/5208/commits/3e71231e2119440844b811330793eaeb34cc9dc0#diff-418254f5c1aadb0c1b8901daf2e6965cd538cf9792e47fb7434175cff92b0598L290) when reading the KVP. This avoid the needs to retrieve a second time the FeatureTypeInfo from the catalog, while at the same time centralizing the localization of i18n content.

aaime (changes requested):
Looks good, added a comment. Also believe ``enimage.png`` and ``itimage.png`` are leftovers from previous testing attempts.

taba90:
thanks @aaime your feedback have been applied.

PR #5212, opened 2021-08-12
[GEOS-10190] i18n support for Contact Information

taba90:
[GEOS-10190](https://osgeo-org.atlassian.net/browse/GEOS-10190)

- This pr adds i18n support for Contact Inforomation in getCapabilities.

- Depends on this pr https://github.com/geoserver/geoserver/pull/5189

  

# Checklist


src/main/src/main/java/org/geoserver/config/impl/ContactInfoImpl.java
```diff
@@ -39,6 +43,34 @@
 
     String onlineResource;
 
+    InternationalString internationalAddress;
```
bradh, line 46: Do these really need to be package visible? Can they be `private`?

src/main/src/main/java/org/geoserver/config/impl/ContactInfoImpl.java
```diff
@@ -50,6 +82,8 @@ public void setId(String id) {
 
     @Override
     public String getAddress() {
+        if (address == null && internationalAddress != null)
```
bradh, line 85: I'm not sure I'm following the logic, but why do we need the `address == null` part of the test? Shouldn't we just prefer `internationalAddress` if its available?

[This is a general comment about the implementation, but the response will be same for all of these cases, so just putting it here]

src/web/core/src/main/java/org/geoserver/web/StringAndInternationalStringPanel.html
```diff
@@ -0,0 +1,24 @@
...
+        <li>
+            <input class="text" type="text" wicket:id="stringField" />
+        </li>
+        <li>
+            <div wicket:id="internationalField" />
+        </li>
+    </ul>
+</wicket:panel>
+</body>
+</html>
```
bradh, line 24: missing EOL.

bradh (commented):
A couple of very minor comments. Nothing significant noted.

bradh:
Test failures look relevant.

doc/en/user/source/services/internationalization/index.rst
```diff
@@ -16,10 +16,15 @@ Configuration
 GeoServer provides an i18n editor for the title and abstract of:
 
 * Layers configuration page.
+
```
aaime, line 19: Whitespace changes?

src/main/src/main/java/org/geoserver/config/impl/ContactInfoImpl.java
```diff
@@ -50,6 +82,8 @@ public void setId(String id) {
 
     @Override
     public String getAddress() {
+        if (internationalAddress != null)
```
aaime, line 85: It seems the logic is different from the FeatureTypeInfoImpl case, e.g.:

```java
        if (this.internationalTitle == null || this.internationalTitle.toString().equals("")) {
            return new GrowableInternationalString(title);
        }
        return this.internationalTitle;
```

I'd recommend a consistent approach unless there is a good reason to do otherwise. Since this bloc is repeated so much, it could also be moved to a utility class/method receiving the base string and the international string.

src/main/src/main/java/org/geoserver/config/impl/ContactInfoImpl.java
```diff
@@ -188,6 +248,191 @@ public void setOnlineResource(String onlineResource) {
...
+
+    @Override
+    public InternationalString getInternationalContactFacsimile() {
+        return internationalContactFacsimile;
+    }
+
+    @Override
+    public void setInternationalContactFacsimile(
+            InternationalString internationalContactFacsimile) {
+        if (internationalContactFacsimile != null)
```
aaime, line 271: The same could be done here, the block is repeated over an over.

aaime (changes requested):

aaime:
Thinking out loud... any documentation that needs to be updated?

src/main/src/main/java/org/geoserver/config/impl/ContactInfoImpl.java
taba90: fixed

src/main/src/main/java/org/geoserver/config/impl/ContactInfoImpl.java
taba90: in theory the two are mutually exclusive. However from UI the normal string field is set to null only when saving (not when pressing apply). The additional null check is meant for this case.

doc/en/user/source/services/internationalization/index.rst
taba90: fixed

src/main/src/main/java/org/geoserver/config/impl/ContactInfoImpl.java
taba90: fixed

src/main/src/main/java/org/geoserver/config/impl/ContactInfoImpl.java
taba90: fixed

taba90:
thanks for your reviews @bradh  and @aaime your feedbacks have been applied.

src/main/src/main/java/org/geoserver/util/InternationalStringUtils.java
```diff
@@ -0,0 +1,43 @@
...
+public class InternationalStringUtils {
+
+    /**
+     * Create a GrowableInternationalString from an InternationalString.
+     *
+     * @param internationalString the InternationalString from which create a
+     *     GrowableInternationalString.
+     * @return a GrowableInternationalString or null.
+     */
+    public static GrowableInternationalString newFromInternationalString(
```
aaime, line 20:
```suggestion
    public static GrowableInternationalString growable(
```

src/main/src/main/java/org/geoserver/util/InternationalStringUtils.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Return the string value or the default InternationalString value if the string one is null.
+     * By default is meant the value that matches the {@link GeoServerDefaultLocale}.
+     *
+     * @param string the string value.
+     * @param internationalString the internationalString instance.
+     * @return the string value or the internationalString default value.
+     */
+    public static String getStringOrFromInternational(
```
aaime, line 36:
```suggestion
    public static String getOrDefault(
```

aaime (changes requested):
Looks good, try to keep names shorter, we only have 100 lines where to put the code.

PR #5217, opened 2021-08-13
[GEOS-10193] Indirect imports will drop the target table if there is any failure during the import process

aaime:
[GEOS-10193](https://osgeo-org.atlassian.net/browse/GEOS-10193)

See ticket

# Checklist

### DIFF
--- a/src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
+++ b/src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
@@ -1504,11 +1504,13 @@ public class Importer implements DisposableBean, ApplicationListener {
                         e1);
             }
 
-            // attempt to drop the type that was created as well
-            try {
-                dropSchema(dataStoreDestination, featureTypeName);
-            } catch (Exception e1) {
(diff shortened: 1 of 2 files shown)


src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -1339,6 +1339,7 @@ void loadIntoDataStore(
             featureType = featureDataConverter.convertType(featureType, format, data, task);
             UpdateMode updateMode = task.getUpdateMode();
             final String uniquifiedFeatureTypeName;
+            boolean newSchema = false;
```
dromagnoli, line 1342: Isn't used anywhere?

dromagnoli (approved):
Sounds good to me.
maybe just an unused variable to be removed. please check

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
aaime: Right, PMD complained and I fixed it... but the change was left on my local disk :-D

PR #5222, opened 2021-08-16
Sphinx documentation theme fixes

rcoup:
1. remove JS so upstream sphinx copies are used. This fixes broken documentation search result links ([GEOS-9202](https://osgeo-org.atlassian.net/browse/GEOS-9202)).
2. update sphinx theme template to remove extraneous noise (html/display/behaviour is the same). Extend from the base template as much as possible rather than copying it.
3. change search box default text so it's clearer it's only searching the current docset. eg: "Search User Manual"
4. fix header links to match what is at [geoserver.org](https://docs.geoserver.org/)
5. fix breadcrumb root link to go back to the docset index

IMO it's worth backporting to stable & maintenance as well, since it makes the docs a lot more usable and fixes several broken links. I can prepare PRs if you want.

- [x] **Q:** Where's the root page for https://docs.geoserver.org ? The links in the page header are broken.

  

# Checklist


rcoup:
AFAICS CI failures are unrelated, but I don't have permssions to retry/rebuild.

aaime:
One of them is a failure that has been since fixed, could you rebase to main?

rcoup:
@aaime that worked 👍

aaime:
@jodygarnett @tbarsballe any chance you can have a look at this one? I don't have previous experience with the Sphinx setup.

tbarsballe:
This _looks_ good to me, although I'm less familiar with the template/structural portions of our documentation source.

@rcoup To answer your question about the location of the root page of the documentation, its not tracked in version control, the only place it lives is on the doc server. This is something we'd like to fix, but no one has gotten around to it yet. There is a little bit of a technical challenge there, since it is the only piece of our documentation that is entirely independent of the GeoServer version, so it would have to live outside of the usual documentation build structure.

I agree this is worth backporting, although probably hold off until this one is merged before taking the time to do that, incase others have more feedback

rcoup:
> There is a little bit of a technical challenge there, since it is the only piece of our documentation that is entirely independent of the GeoServer version, so it would have to live outside of the usual documentation build structure.

Dunno where the Jenkins build configs live... (in Jenkins? 🙈 ), but as a first step could Jenkins just push a static HTML file from eg: `doc/doc_index.html` to S3 when it's building `main`? It already has S3 credentials.

rcoup:
> This looks good to me, although I'm less familiar with the template/structural portions of our documentation source.

I'm not a Sphinx expert either but the search brokenness was frustrating me so I found a similar bug-report elsewhere and applied the same approach, after checking the JS was verbatim from a long-ago sphinx version. Other things were tweaks I picked up on while testing it, when I had it open and front-of-mind since this stuff doesn't get touched much 😃

tbarsballe:
> > There is a little bit of a technical challenge there, since it is the only piece of our documentation that is entirely independent of the GeoServer version, so it would have to live outside of the usual documentation build structure.
> 
> Dunno where the Jenkins build configs live... (in Jenkins? 🙈 ), but as a first step could Jenkins just push a static HTML file from eg: `doc/doc_index.html` to S3 when it's building docs for `main`? It already has S3 credentials.

The jenkins build configs do actually exist in a (private) repository, and that solution would work, but I'd prefer to have the landing page exist alongside the rest of the documentation in this (geoserver/geoserver) repository, and just only get built off of the master branch. It's not an especially difficult problem to solve, just a little time consuming to get everything hooked up properly (and its been rather low priority).

rcoup:
so... can we get this merged and address that separately? 😃  

At https://gist.github.com/rcoup/1e06bd54db2073450933787ad47dcb4c there's a patch for fixing the header links in the documentation root & a new version with the patch applied.

tbarsballe:
I was hoping that Jody could've looked over the more structural parts, but it seems like he's been busy. As such, I've gone ahead and merged. The other issue can be addressed separately.

I'll try and find some time this week to apply the index page fix, thanks for putting together the gist.

rcoup:
@tbarsballe it's expected that `latest` goes to the latest _release_? Or is it meant to be the nightly build?

https://docs.geoserver.org/main/en/developer/index.html ← has the nightly docs...

tbarsballe:
`latest` goes to the latest commit of the docs (so, main/nightly build). It is the same as `main`; i.e.:
https://docs.geoserver.org/main/en/developer/index.html
and
https://docs.geoserver.org/latest/en/developer/index.html
should be equivalent.

The latest *releases* are http://docs.geoserver.org/stable/en/developer/ (stable) and http://docs.geoserver.org/maintain/en/developer/ (maintenance)

PR #5225, opened 2021-08-17
[GEOS-10194] Improve importer LOGGING

aaime:
[GEOS-10194](https://osgeo-org.atlassian.net/browse/GEOS-10194)

See ticket for details.

# Checklist

### DIFF
--- a/src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
+++ b/src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
@@ -351,6 +351,7 @@ public class Importer implements DisposableBean, ApplicationListener {
         } else {
             contextStore.add(context);
         }
+        LOGGER.log(Level.FINE, "Created new context with id: {0}", context.getId());
         return context;
     }
 
@@ -384,16 +385,14 @@ public class Importer implements DisposableBean, ApplicationListener {
         if (!context.progress().isCanceled()) {
             contextStore.add(context);
         }
-        // JD: don't think we really need to maintain these, and they aren't persisted
-        // else {
-        //    context.setState(ImportContext.State.CANCELLED);
-        // }
+        LOGGER.log(Level.FINE, "Context created: {0}", context);
         return context;
     }
 
     public Long createContextAsync(
             final ImportData data, final WorkspaceInfo targetWorkspace, final StoreInfo targetStore)
             throws IOException {
+        LOGGER.log(Level.FINE, "Asynchronous context creation scheduled");
         return asynchronousJobs.submit(
                 new SecurityContextCopyingJob<ImportContext>() {
                     @Override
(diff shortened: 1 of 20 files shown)


src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -1075,6 +1110,10 @@ void doDirectImport(ImportTask task) throws IOException {
                     }
                 }
             }
+            LOGGER.log(
+                    Level.FINE,
+                    "Creating target store {0} for task {0}",
```
dromagnoli, line 1115: Second parameter should be {1} to get the task id

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -1629,14 +1668,26 @@ void addToCatalog(ImportTask task) throws IOException {
         // changed to deal with name clashes
         // resource.setNativeName(name);
         resource.setEnabled(true);
+        LOGGER.log(
+                Level.FINE,
+                "Creating target resource {0} for task {0}",
+                new Object[] {resource, task.getId()});
```
dromagnoli, line 1674: Second parameter should be {1} to get the task id

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -1629,14 +1668,26 @@ void addToCatalog(ImportTask task) throws IOException {
...
         catalog.add(resource);
 
         // add the layer (and style)
         if (layer.getDefaultStyle().getId() == null) {
+            LOGGER.log(
+                    Level.FINE,
+                    "Creating default style {0} for task {0}",
```
dromagnoli, line 1681: Second parameter should be {1} to get the task id

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -1629,14 +1668,26 @@ void addToCatalog(ImportTask task) throws IOException {
...
+                    new Object[] {layer.getDefaultStyle(), task.getId()});
             catalog.add(layer.getDefaultStyle());
         }
 
         layer.setEnabled(true);
+        LOGGER.log(
+                Level.FINE,
+                "Creating target layer  {0} for task {0}",
```
dromagnoli, line 1689: Second parameter should be {1} to get the task id
Also, I see 2 whitespaces between layer and {0}. Not a critical problem :) but since you need to fix the text using {1} I would also remove 1 whitespace

dromagnoli (approved):
I have seen a couple of wrong parameters indexes in the logs.
But once fixed them, we are ready to merge.

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
aaime: Good catch, fixed.

PR #5231, opened 2021-08-18
[GEOS-9937] Name of styles with colons are incorrect in REST API

shadab-geo:
[GEOS-9937](https://osgeo-org.atlassian.net/browse/GEOS-9937)

  

[GEOS-9937](https://osgeo-org.atlassian.net/browse/GEOS-9937)

In the current implementation, it looks that anything present before the colon in style name is considered as a workspace name. It proceeds even if the workspace with the given name is not present. This fix assigns workspace name only if one with the given name exists

# Checklist


dromagnoli:
Hi @shadab-geo, please, make sure to add a new test for that fix

nmco:
@dromagnoli I'm wondering if having a style name with colons makes sense:
- How can we actually create a style with colons? By editing manually the data directory?
- How will this play out with the workspace prefix business when querying the data?
- How will this play out with GWC?

More I think about this, more I think that we should discuss with the community if layers names, style names, etc ... shall be allowed to have a colon don't he name.

dromagnoli:
> @dromagnoli I'm wondering if having a style name with colons makes sense:
> 
I suspect older versions of GeoServer allow for that so the user reporting the bug was using these as a sort of prefix or label for categorization.

> * How can we actually create a style with colons? By editing manually the data directory?
Yep, with the current code, that's the only way to do that and replicate the issue.

> * How will this play out with the workspace prefix business when querying the data?
> * How will this play out with GWC?
> 
> More I think about this, more I think that we should discuss with the community if layers names, style names, etc ... shall be allowed to have a colon don't he name.

Yep. And/Or, if we want to drop this kind of "legacy-support" we should put a note in the upgrading pages from older versions.

nmco:
Makes sense @dromagnoli I think about older users, thanks for the explanation.

PR #5232, opened 2021-08-18
Updated expected location of dimension attributes

achamberlain:
Moved the TimeAttribute and ElevationAttribute properties from indexer.properties to the primary configuration file.

  

# Checklist

### DIFF
--- a/doc/en/user/source/data/raster/imagemosaic/configuration.rst
+++ b/doc/en/user/source/data/raster/imagemosaic/configuration.rst
@@ -70,6 +70,12 @@ The table below describes the various elements in this configuration file.
    * - LocationAttribute
      - Y
      - The name of the attribute path in the shapefile index. Default is ``location``.
+   * - TimeAttribute
+     - N
+     - Specifies the name of the time-variant attribute.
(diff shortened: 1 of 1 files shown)


dromagnoli:
Hi @achamberlain 
Note that when creating an imageMosaic from scratch, the file being used to create index, schema and properties is the indexer.
So, removing the timeAttribute and elevationAttribute from the indexer.properties will result into not having them in the mosaic properties.

So the 2 attributes need to be in the indexer.properties too.
We can update the doc by adding them in the mosaic's properties file but they should be kept in the indexer.properties too. Otherwise people creating mosaic from scratch with time and elevation will have hard times in figuring out why the dimensions didn't get recognized.

achamberlain:
Got it.  I hadn't appreciated they were correct in the `indexer.properties`, and merely needed _adding_ to the primary config file.   The description and nature of this change is now a bit misleading.  Best to close without merging, and open a new one?

dromagnoli:
that makes sense

aaime:
Closing as suggested, waiting for the updated PR

PR #5234, opened 2021-08-19
Bump org.owasp:dependency-check-maven to 6.2.2

mprins:
And update the config to suppress false positives in some of the community modules.

While here also remove a duplicate dependency declaration in `src/community/oseo/wb-oseo/pom.xml` that maven complains about.

  

# Checklist

### DIFF
--- a/src/community/oseo/web-oseo/pom.xml
+++ b/src/community/oseo/web-oseo/pom.xml
@@ -63,13 +63,6 @@
     </dependency>
     <dependency>
       <groupId>org.geoserver.community</groupId>
-      <artifactId>gs-oseo-core</artifactId>
-      <version>${gs.version}</version>
-      <classifier>tests</classifier>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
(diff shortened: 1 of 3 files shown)


src/community/oseo/web-oseo/pom.xml
```diff
@@ -61,13 +61,6 @@
       <classifier>tests</classifier>
       <scope>test</scope>
     </dependency>
-    <dependency>
```
bradh, line 64: This part looks unrelated and possibly undesired.

bradh (commented):
Generally looks OK except for the change in the web-oseo pom.

src/community/oseo/web-oseo/pom.xml
mprins: when you look ~10 lines up you can see the duplicate declaration: https://github.com/geoserver/geoserver/blob/0ef63e17696b7b4c278e78b3138968101c16d0f9/src/community/oseo/web-oseo/pom.xml#L43-L49

PR #5241, opened 2021-08-20
[GEOS-9937] Name of styles with colons are incorrect in REST API

shadabgada:
[GEOS-9937](https://osgeo-org.atlassian.net/browse/GEOS-9937)

  

In the current implementation, it looks that anything present before the colon in style name is considered as a workspace name. It proceeds even if the workspace with the given name is not present. This fix assigns workspace name only if one with the given name exists

# Checklist

### DIFF
--- a/src/restconfig/src/test/java/org/geoserver/rest/catalog/LayerControllerTest.java
+++ b/src/restconfig/src/test/java/org/geoserver/rest/catalog/LayerControllerTest.java
@@ -17,7 +17,14 @@ import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
 import java.io.IOException;
+import java.util.Arrays;
+import java.util.Objects;
 import net.sf.json.JSONObject;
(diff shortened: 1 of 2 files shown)


src/restconfig/src/test/java/org/geoserver/rest/catalog/LayerControllerTest.java
```diff
@@ -400,4 +405,43 @@ public void testPutDefaultWMSInterpolationMethod() throws Exception {
...
+
+        assertEquals(200, response.getStatus());
+
+        //test whether string before colon is not duplicated
+        assertFalse(response.getContentAsString().contains("demo:demo:RoadSegmentsDup"));
+    }
+
+    public void updateStyleNameInXmlFile() throws IOException {
+
+        BufferedReader reader = new BufferedReader(new FileReader(getDataDirectory().get("styles").dir().getPath() + "\\RoadSegments.xml"));
```
dromagnoli, line 432: a few suggestions on this method:

- Both BufferedReader and BufferedWriter implements AutoCloseable so you can use try-with-resources, so that you just open them in a try {} and they will get automatically closed at the end.
- You can have the writer writing the lines being read by the reader in the same while so you don't need a temporary stringBuilder

dromagnoli (changes requested):

dromagnoli:
Make sure to build the code with mvn to make sure all formatting get applied

PR #5243, opened 2021-08-23
[GEOS-10201] Geoserver fails to start on Windows 11 beta

mprins:
[GEOS-10201](https://osgeo-org.atlassian.net/browse/GEOS-10201)

Bump OSHI from 5.5.1 to 5.8.1 and also bump the transitive dependency JNA to 5.9.0, which should make it possible to run on Windows 11

I don't have access to Windows 11 (beta) so cannot test if this actually fixes the problem. (Windows 11 won't be availaible for a couple of months)

see also https://github.com/oshi/oshi/issues/1699

  

# Checklist

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -1332,7 +1332,7 @@
       <dependency>
         <groupId>com.github.oshi</groupId>
         <artifactId>oshi-core</artifactId>
-        <version>5.5.1</version>
+        <version>5.8.1</version>
       </dependency>
       <dependency>
         <groupId>org.postgresql</groupId>


dbwiddis:
FYI the issue is also associated with newer upcoming editions of Windows 10 available in insider builds.  Same fix applies.

mprins:
Thanks. The next major release cycle starts in ~2 weeks with 2.20-RC, and 2.20.0 ~2 weeks later so this should be in time I think, if there is a need it could be backported to 2.19.3 which is slated for 2nd half of October

PR #5248, opened 2021-08-25
[GEOS-10214] Java 17 compatibility

bjornharrtell:
[GEOS-10214](https://osgeo-org.atlassian.net/browse/GEOS-10214)

This PR aims to get GeoServer build and runtime compatible with Java 17.

Potential issues/upgrades needed (still undecided):

* [ ] Spring framework (see https://spring.io/blog/2021/03/16/spring-framework-5-3-5-available-now)
  * [ ] cglib (not officially compatible at this time but perhaps fixed somehow by spring framework)
  * [ ] servlet-api 3.1.0 -> 4.0.1

# Checklist


bjornharrtell:
Workaround for easymock issue was to use add-opens for the test run.

src/pom.xml
```diff
@@ -1581,7 +1581,7 @@
           <printSummary>true</printSummary>
           <testFailureIgnore>${allow.test.failure.ignore}</testFailureIgnore>
           <excludedGroups>${test.excludedGroups}</excludedGroups>
-          <argLine>@{argLine} -XX:+IgnoreUnrecognizedVMOptions --illegal-access=debug -Dfile.encoding=UTF-8</argLine>
+          <argLine>@{argLine} -XX:+IgnoreUnrecognizedVMOptions -Dfile.encoding=UTF-8 --add-opens=java.base/java.lang=ALL-UNNAMED</argLine>
```
aaime, line 1584: I'm guessing that without this one XStream would not work (among other things)?

bjornharrtell: @aaime jury is still out on that. I have no idea what xstream is used for, do you know what absolutely would break if it didn't workas user facing GeoServer functionality? Because I have run GeoServer on this branch and the java17 geotools branch locally and things seems to work to me (with no add opens).

aaime: Well, saving some configurations would break, but I don't know, off the top of my head, which one.

bjornharrtell: I tried doing a bunch of configuration changes and stuff and it seemed to work.

aaime:
Another interesting read, Spring's take on Java 17: https://spring.io/blog/2021/09/02/a-java-17-and-jakarta-ee-9-baseline-for-spring-framework-6

Long story short, things are setup so that there will be a full API break on Spring 6, which will require JDK 17 and Jakarta EE 9, while earlier versions will keep supporting Java 8. Java 11 is considered a dud:

> In case you’re wondering about JDK 11 as an LTS generation, please note that the commercial support timeframes for JDK 11 are shorter than for JDK 8, with JDK 11 LTS phasing out in late 2023 already. JDK 17 as the next LTS generation will provide support timeframes until 2026 at least. We consider JDK 8 to have a unique role in the ecosystem; in comparison, JDK 11 is a transitional release

aaime:
About the finalizer in ``FileSystemResourceStore$FileSystemResource$1``: if we manage to have the rest build, I won't shed tears on it and just remove it.

bjornharrtell:
@aaime yeah I read that one but hope that doesn't mean spring 5 cannot work on Java 17, it did not seem so when considering other information.

aaime:
I've pushed a commit with extra opens, they allow XStream to work properly in gs-restconfig (that was the cause of all test failures).

aaime:
Well well, getting close, 3 modules lef to fix for a first round. There is however a ton of opens to get this thing going, might want to see if we can reduce them a little bit...

aaime:
Managed to build a war and a bin. The bin does not start on Java 17 due to the Marlin renderer integration, once removed, it starts and seems to work fine (weirdly, even without opens).
I'm not sure how to make the Marlin bit conditional on the Java version, the scripts to check the Java version around assume some amount of extra utilities, e.g. this sone works:

java -version 2>&1 | head -1 | cut -d'"' -f2 | sed '/^1\./s///' | cut -d'.' -f1

But it assumes head, cut and sed are available... not sure they will be in stripped down docker machines, for example.
And then, Windows support should for the same check should be developed (for startup.bat).

Maybe the support for Java 17 should come with instructions on how to run the thing.
Also, haven't tried out on Tomcat yet.

aaime:
And we have a green build :-D

aaime:
Added docs too. I believe it's suitable for review.

bjornharrtell (commented):
Looks good to me! Thanks @aaime for bringing this home! :)

mprins (approved):

aaime:
All green, in it goes.

PR #5264, opened 2021-08-31
[GEOS-10221] Add online documentation root to version control

tbarsballe:
[GEOS-10221](https://osgeo-org.atlassian.net/browse/GEOS-10221)

Refer to https://osgeo-org.atlassian.net/browse/GEOS-10221

Thanks to @rcoup for [providing](https://gist.github.com/rcoup/1e06bd54db2073450933787ad47dcb4c) the updated index.html with fixed links.

# Checklist

### DIFF
--- a/doc/en/index.html
+++ b/doc/en/index.html
@@ -1,0 +1,481 @@
+
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
+  "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+<html xmlns="http://www.w3.org/1999/xhtml" lang="en-US" xml:lang="en-US">
+<head>
+  <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+  <title>GeoServer Documentation</title>
+  <link rel="stylesheet" href="_static/blueprint/screen.css" type="text/css" media="screen, projection" />
+  <link rel="stylesheet" href="_static/blueprint/print.css" type="text/css" media="print" /> 
+  <!--[if IE]>
+  <link rel="stylesheet" href="_static/blueprint/ie.css" type="text/css" media="screen, projection" />
+  <![endif]-->
+  <link rel="stylesheet" href="_static/default.css" type="text/css" />
+  <link href="_static/favicon.ico" rel="shortcut icon"/>
+  <script type="text/javascript" src="_static/jquery.js"></script>
+
+</head>
+<body class="index">
+  <div id="header" class="selfclear">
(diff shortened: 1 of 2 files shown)


rcoup:
ah, one thing to note here: the "Release Versions" section only has up to 2.13.2...

doc/en/index.html
```diff
@@ -0,0 +1,481 @@
...
+           You can also download the version-specific documentation on the appropriate
+           <a href="http://geoserver.org/download">Download</a> page.</p>
+
+        <table>
+
+          <tr>
+            <th>Release</th><th>User Manual</th><th>Developer Manual</th>
+          </tr>
+          <tr>
+            <td><strong><a href="http://geoserver.org/release/2.13.0/">2.13.2</a></strong></td>
```
mprins, line 107: this and some of the other urls below seem incorrect. eg. http://geoserver.org/release/2.13.2/ exists, so does http://geoserver.org/release/2.13.1/
```suggestion
            <td><strong><a href="http://geoserver.org/release/2.13.2/">2.13.2</a></strong></td>
```

doc/en/index.html
```diff
@@ -0,0 +1,481 @@
...
+            <td><a href="2.13.2/user/">User Manual</a></td>
+            <td><a href="2.13.2/developer/">Developer Manual</a></td>
+          </tr>
+          <tr>
+            <td><strong><a href="http://geoserver.org/release/2.13.0/">2.13.1</a></strong></td>
```
mprins, line 112: as above

```suggestion
            <td><strong><a href="http://geoserver.org/release/2.13.1/">2.13.1</a></strong></td>
```

mprins (changes requested):
some urls are incorrect

doc/en/index.html
rcoup: I left these as-is, sorry!

More importantly is probably missing all the 2.14-2.20 versioned URLs? 😬

tbarsballe: Not quite. All the 2.13.1, etc. links were generated manually. Since around 2.14, we've only been (automatically) generating `2.14.x` doc artifacts, so there's rather fewer new links required (6, instead of 20 or so). There's been a whole bunch of discussion in the PMC meetings and on the mailing lists on which of these archetypes to follow (docs for every point release vs. only major releases) and we've settled on the latter.
This'll also make updating it a good bit easier.

doc/en/index.html
tbarsballe: Fixed

tbarsballe:
I've fixed the two incorrect links, and added the missing ones up to the current stable release (2.19.x).

mprins (approved):

PR #5265, opened 2021-09-02
[GEOS-10222] WFS CSV OutputFormat delimiter

shadab-geo:
[GEOS-10222](https://osgeo-org.atlassian.net/browse/GEOS-10222)

  

# Checklist

### DIFF
--- a/doc/en/user/source/services/wfs/outputformats.rst
+++ b/doc/en/user/source/services/wfs/outputformats.rst
@@ -135,3 +135,15 @@ JSON output ``format_options``:
 JSON output ``system properties``:
 
 * ``json.maxDepth=<max_value>`` is used to determine the max number of allowed JSON nested objects on encoding phase.  By default the value is 100.
+
+
+CSV output
+----------------
+
+A Default CSV file uses a comma to separate values. Each line of the file is a data record. Each record consists of one or more fields, separated by commas. The separator can be changed using format_options as specified below.
+
+csv file output ``format_options``:
+
+* ``format_option=filename:<file>``: if a file name is provided, the name is used as the output file name. For example, ``format_options=filename:roads.csv``.
+* ``format_option=csvseparator:<csvseparator>`` (default is ```,``` ): if a separator is provided, it is used to separate values in output csv file. For example, ``format_options=csvseparator:-`` is used to get dash separated file.
+
--- a/doc/en/user/source/services/wfs/vendor.rst
+++ b/doc/en/user/source/services/wfs/vendor.rst
@@ -37,6 +37,7 @@ The supported format option is:
 * ``callback`` (default is ``parseResponse``)—Specifies the callback function name for the JSONP response format
 * ``id_policy`` (default is ``true``)- Specifies id generation for the JSON output format. To include feature id in output use an attribute name, or use ``format_options=id_policy:true`` for feature id generation. To avoid the use of feature id completely use ``format_options=id_policy:false``.
 * ``filename`` (default is :file:`features` or generated from feature type name)- provide a ``Content-Disposition`` header indicating the attachment filename (used as a suggestion by browsers saving content to disk using :command:`Save-As`). For example ``format_options=filename:content.txt``.
+* ``csvseparator`` (default is ```,``` )- Specifies a separator that can be used in output csv file
 
 Reprojection
 ------------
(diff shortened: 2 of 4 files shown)


mprins (changes requested):
LGTM, but since this is a user visible change documentation must be updated in https://github.com/geoserver/geoserver/blob/main/doc/en/user/source/services/wfs/outputformats.rst

doc/en/user/source/services/wfs/outputformats.rst
```diff
@@ -135,3 +135,15 @@ JSON output ``format_options``:
...
+
+
+CSV output
+----------------
+
+A Default CSV file uses a comma to separate values. Each line of the file is a data record. Each record consists of one or more fields, separated by commas. The separator can be changed using format_options as specified below.
+
+csv file output ``format_options``:
+
+* ``format_option=filename:<file>``: if a file name is provided, the name is used as the output file name. For example, ``format_options=filename:roads.zip``.
```
mprins, line 147: even though correct, I find it a bit counter-intuitive to have a CSV file with `.zip` extension
```suggestion
* ``format_option=filename:<file>``: if a file name is provided, the name is used as the output file name. For example, ``format_options=filename:roads.csv``.
```

mprins (approved):
thanks for the docs, one small nit-pick

doc/en/user/source/services/wfs/outputformats.rst
shadab-geo: Changes applied, Thank you :)

jratike80:
If I read the code right, the separator character is somehow escaped if it appears in a text string. How does it alter for example string `osm:highway` when `:` is used as a separator?

shadab-geo:
> If I read the code right, the separator character is somehow escaped if it appears in a text string. How does it alter for example string `osm:highway` when `:` is used as a separator?

It will enclose the string with double quotes and return it as `"osm:highway"`

jratike80:
All right. What if source data is like

```
osm:highway
osm_highway
```

will the output CSV with : as separator be like

```
attr:"osm:highway":next_attr
atrr:osm_highway:next_attr
```
I can see from topp:states
http://localhost:8080/geoserver/topp/ows?service=WFS&version=1.0.0&request=GetFeature&typeName=topp%3Astates3&maxFeatures=50&outputFormat=text%2Fcsv
that by default the geometry WKT is enclosed between "" but the string attributes are not:
`Missouri,29,W N Cen,MO,`

I wonder if it would be better to enclose always all the strings between double quotes
http://localhost:8080/geoserver/ows?service=wfs&version=2.0.0&request=DescribeFeatureType&typenames=topp:states
->
`"Missouri","29","W N Cen","MO",`

Pay attention to state_fips that is string in the schema but that has value 29. CSV readers may well interpret the attribute to be numeric.

Additional thinking is needed if user provides `"` as a separator (action: deny?), or strings contain " characters (action: escape \\" ?).

shadab-geo:
> All right. What if source data is like
> 
> ```
> osm:highway
> osm_highway
> ```
> 
> will the output CSV with : as separator be like
> 
> ```
> attr:"osm:highway":next_attr
> atrr:osm_highway:next_attr
> ```
That's right!

> I can see from topp:states
> http://localhost:8080/geoserver/topp/ows?service=WFS&version=1.0.0&request=GetFeature&typeName=topp%3Astates3&maxFeatures=50&outputFormat=text%2Fcsv
> that by default the geometry WKT is enclosed between "" but the string attributes are not:
> `Missouri,29,W N Cen,MO,`

This is expected behavior. The CSV separator, in this case, is a comma. The values of column 'the_geom' contain a comma in its value which is why it encloses it within the double-quotes.

> I wonder if it would be better to enclose always all the strings between double quotes
> http://localhost:8080/geoserver/ows?service=wfs&version=2.0.0&request=DescribeFeatureType&typenames=topp:states
> ->
> `"Missouri","29","W N Cen","MO",`

I am not sure about it. We need to check with other devs because this logic of enclosing string within double quote was already existing. (I have just added dynamic CSV separator in this PR)

> Pay attention to state_fips that is string in the schema but that has value 29. CSV readers may well interpret the attribute to be numeric.

consider a case with a negative number and dash as a separator, I guess it will crash the CSV format in that case.

> Additional thinking is needed if user provides `"` as a separator (action: deny?), or strings contain " characters (action: escape \" ?).

I agree we might want to deny adding `"` as a separator. 
If string contains `"`  for example `Mi"ssouri,29`
it will write as
` "Mi"ssouri",29`

dromagnoli:
@shadab-geo any other action needed on this ticket?

shadab-geo:
Adding fix to prevent double-quotes @dromagnoli

PR #5271, opened 2021-09-07
[GEOS-10004] Add ability to animate env variables in download-WPS

weskamm:
[GEOS-10004](https://osgeo-org.atlassian.net/browse/GEOS-10004)

This PR adds the ability to create animations via the download-WPS extension without specifying the TIME parameter but instead using an "animateParam" and "animateValues" parameter similar to the old behaviour.
  

# Checklist


doc/en/user/source/extensions/wps-download/mapAnimationDownload.rst
```diff
@@ -215,7 +217,87 @@ The ``formattedTimestamper`` decoration ensures the frame time is included in th
       </decoration>
     </layout>
 
+Sample variable-based DownloadAnimation request
++++++++++++++++++++++++++++++++++++++++++++++++
 
+If you want to animate a map based on variables, proceed as follows:
+
+* specify the ``animateParam`` variable. You can use this to animate SLD variables, as described in ``Variable substitution in SLD``
```
aaime, line 225: Dotted lists in English start with a capital letter and end with a full stop (unless the first word is a keyword that cannot be capitalized in any other way).

doc/en/user/source/extensions/wps-download/mapAnimationDownload.rst
```diff
@@ -215,7 +217,87 @@ The ``formattedTimestamper`` decoration ensures the frame time is included in th
...
+  or animate through a custom parameter
+* specify the ``animateValues`` parameter as a comma separated list of your values that will be used in the animation
```
aaime, line 227: We have parameters whose value is comma-separated, or can contain a comma. The documentation should specify a way to escape commas (e.g., see the [old animator docs for reference](https://docs.geoserver.org/stable/en/user/tutorials/animreflector.html#overview)).

src/extension/wps-download/src/main/java/org/geoserver/wps/gs/download/DownloadAnimationProcess.java
```diff
@@ -123,12 +127,26 @@ public DownloadAnimationProcess(
...
+                        min = 0,
                         description =
                                 "Map time specification (a range with "
                                         + "periodicity or a list of time values)"
                     )
                     String time,
+            @DescribeParameter(
+                        name = "animateParam",
+                        min = 0,
+                        max = 1,
```
aaime, line 139: No need to add max=1, it's the default value

src/extension/wps-download/src/main/java/org/geoserver/wps/gs/download/DownloadAnimationProcess.java
```diff
@@ -175,8 +193,19 @@ public DownloadAnimationProcess(
             AWTSequenceEncoder enc = new AWTSequenceEncoder(out, frameRate);
 
             DownloadServiceConfiguration configuration = confiGenerator.getConfiguration();
-            TimeParser timeParser = new TimeParser(configuration.getMaxAnimationFrames());
-            Collection parsedTimes = timeParser.parse(time);
+            int totalFrames;
+            Collection parsedTimes = null;
+            List<String> animateValuesList = new ArrayList<>();
+            if (!StringUtils.isEmpty(time)) {
```
aaime, line 199: The documentation states time and animateParam are incompatible, better to check if they are both specified and throw an exception.

src/extension/wps-download/src/main/java/org/geoserver/wps/gs/download/DownloadAnimationProcess.java
```diff
@@ -214,11 +242,17 @@ public DownloadAnimationProcess(
...
-                for (Object parsedTime : parsedTimes) {
-                    // turn parsed time into a specification, generates a "WMS" like request based
-                    // on it
-                    String mapTime = toWmsTimeSpecification(parsedTime);
-                    LOGGER.log(Level.FINE, "Building frame for time %s", mapTime);
+                for (int i = 0; i < totalFrames; i++) {
+                    String mapTime = null;
+                    String animateValue = null;
+                    if (parsedTimes != null) {
+                        // turn parsed time into a specification and generate a "WMS" like request
```
aaime, line 249: Reformat comment to fit in less lines? :-D

src/extension/wps-download/src/main/java/org/geoserver/wps/gs/download/DownloadAnimationProcess.java
```diff
@@ -175,8 +193,19 @@ public DownloadAnimationProcess(
...
-            Collection parsedTimes = timeParser.parse(time);
+            int totalFrames;
+            Collection parsedTimes = null;
+            List<String> animateValuesList = new ArrayList<>();
+            if (!StringUtils.isEmpty(time)) {
+                TimeParser timeParser = new TimeParser(configuration.getMaxAnimationFrames());
+                parsedTimes = timeParser.parse(time);
+                totalFrames = parsedTimes.size();
+            } else if (!StringUtils.isEmpty(animateParam) && !StringUtils.isEmpty(animateValues)) {
+                animateValuesList = Arrays.asList(animateValues.split(","));
```
aaime, line 204: Should give the possibilty to escape commas.
I _believe_ this code handled the escaping, but not 100% sure: https://github.com/geoserver/geoserver/blob/2.17.x/src/wms/src/main/java/org/geoserver/wms/animate/FrameCatalog.java#L51
If not that one, it may be in the vicinity (other classes of the same package, for example).

aaime (changes requested):

PR #5274, opened 2021-09-07
[GEOS-10221] Fix broken links in Release versions, add section on updating these links to release guide

tbarsballe:
[GEOS-10221](https://osgeo-org.atlassian.net/browse/GEOS-10221)

Follow-up to https://github.com/geoserver/geoserver/pull/5264

The documentation index page from that PR is now deploying automatically, and the updates can be seen on [docs.geoserver.org](https://docs.geoserver.org)

However, when adding the links for 2.14.x through 2.19.x to that PR, I neglected to test them - turns out the doc link structure changed between 2.13 and 2.14, corresponding with the doc server move. This PR fixes that error, and also adds a step to the release guide for adding a new link (with the correct structure) when a new RC branch is created.

See also https://osgeo-org.atlassian.net/browse/GEOS-10221

# Checklist

### DIFF
--- a/doc/en/index.html
+++ b/doc/en/index.html
@@ -105,33 +105,33 @@
           </tr>
           <tr>
             <td><strong><a href="http://geoserver.org/release/2.19.x/">2.19.x</a></strong></td>
-            <td><a href="2.19.x/user/">User Manual</a></td>
-            <td><a href="2.19.x/developer/">Developer Manual</a></td>
+            <td><a href="2.19.x/en/user/">User Manual</a></td>
+            <td><a href="2.19.x/en/developer/">Developer Manual</a></td>
           </tr>
           <tr>
             <td><strong><a href="http://geoserver.org/release/2.18.x/">2.18.x</a></strong></td>
-            <td><a href="2.18.x/user/">User Manual</a></td>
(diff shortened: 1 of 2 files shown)


doc/en/index.html
```diff
@@ -105,33 +105,33 @@ <h3>Release Versions</h3>
           </tr>
```
mprins, line 105: this might be for another PR, but I was wondering if the Chinese docs should be mentioned on this page? (as of `latest`) eg. https://docs.geoserver.org/latest/zhCN/user/

mprins (approved):

doc/en/index.html
tbarsballe: Probably best as its own issue/seperate PR, yeah. I think the "best" approach would be a drop-down in the doc header allowing you to switch languages directly from a version/page, but I doubt that's feasible.

Also the Chinese docs are only really present on main right now, and only (parts) of the user guide are translated, so we'd need to determine a way of linking only the docs that do exist without looking too inconsistent. Maybe after 2.20 is released and there's actually a released version with translated docs we can add a link there.

Back when we had (partial) Spanish and French docs, I don't think they were linked anywhere either, so we don't have any defined way of representing this.

PR #5277, opened 2021-09-09
Enhance the env-function documentation

weskamm:
This adds a hint for the usage of the env function in dynamic symbolizers (online resource in external graphic).
In that case one needs to put single quotes around the first argument to make things work.
  

# Checklist

### DIFF
--- a/doc/en/user/source/filter/function_reference.rst
+++ b/doc/en/user/source/filter/function_reference.rst
@@ -109,6 +109,25 @@ This function returns the value of environment variables
 defined in various contexts.
 WMS GetMap automatically defines somes variables :ref:`SLD rendering <sld_variable_substitution>`,
 while others can be provided using the ``env`` request parameter.
+Example usage in e.g. a dynamic symbolizer:
+
+``${env('size', 20)}``
+
+Example usage in a default Symbolizer:
+
+.. code-block:: xml
+
+  <PointSymbolizer uom="http://www.opengeospatial.org/se/units/metre">
+    <Graphic>
+    ...
+      <Size>
+        <ogc:Function name="env">
+          <ogc:Literal>size</ogc:Literal>
+          <ogc:Literal>20</ogc:Literal>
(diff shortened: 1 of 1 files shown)


doc/en/user/source/filter/function_reference.rst
```diff
@@ -108,7 +108,9 @@ Environment function
 This function returns the value of environment variables
 defined in various contexts.
 WMS GetMap automatically defines somes variables :ref:`SLD rendering <sld_variable_substitution>`,
-while others can be provided using the ``env`` request parameter.
+while others can be provided using the ``env`` request parameter. When using this function for
+dynamic symbolizers make sure to put the argument of this function in single quotes,
+e.g. ``${env('size', 20)}``.
```
aaime, line 113: It has to be put in single quotes everywhere. An unquoted string in CQL is an attribute name reference (no matter where, it's not just dynamic symbolizes). It's a also valid usage, it would mean the env var name to be looked after is contained in the attribute value.

weskamm: Yeah but that may not be clear to everyone.
Looking at the documentation here
https://docs.geoserver.org/stable/en/user/styling/sld/extensions/pointsymbols.html#dynamic-symbolizers
especially in the last code block at the bottom one could get the idea that function arguments may be given without quotes. Feel free to close this PR as it may just be me who got confused here :-)

aaime: At least, I'd reword it to avoid mentioning the dynamic symbolizer case (there is nothing specific to it). Maybe just add an example or two, that should be enough?

weskamm: I have added examples and reworded a bit. The dynamic symbolizer is still there as an example, as i dont know other ways of calling the env function besides the `classic` way

aaime: Looks good, thanks! Merged.

PR #5283, opened 2021-09-16
[GEOS-10232]SQL views are not saved if the layer is not saved too

shadab-geo:
[GEOS-10232](https://osgeo-org.atlassian.net/browse/GEOS-10232)

  

# Checklist

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/data/layer/SQLViewAbstractPage.html
+++ b/src/web/core/src/main/java/org/geoserver/web/data/layer/SQLViewAbstractPage.html
@@ -28,9 +28,12 @@
         </li>
     </ul>
 
-    <div class="button-group toolbar-sticky selfclear"><a href="#" class="form-button-save" wicket:id="save"><wicket:message
-        key="save">Save</wicket:message></a> <a href="#" class="form-button-cancel" wicket:id="cancel"><wicket:message
+    <div class="button-group toolbar-sticky selfclear">
+        <a href="#" class="form-button-save" wicket:id="save"><wicket:message key="save">Save</wicket:message></a>
+        <a href="#" class="form-button-save" wicket:id="ok"><wicket:message key="ok">OK</wicket:message></a>
+        <a href="#" class="form-button-cancel" wicket:id="cancel"><wicket:message
         key="cancel">Cancel</wicket:message></a></div>
+        <span wicket:id="notice">notice</span>
     </form>
     
 <wicket:fragment wicket:id="checkbox">
--- a/src/web/core/src/main/java/org/geoserver/web/data/layer/SQLViewAbstractPage.java
+++ b/src/web/core/src/main/java/org/geoserver/web/data/layer/SQLViewAbstractPage.java
@@ -18,6 +18,7 @@ import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
 import org.apache.wicket.Component;
 import org.apache.wicket.ajax.AjaxRequestTarget;
+import org.apache.wicket.markup.html.basic.Label;
 import org.apache.wicket.markup.html.form.CheckBox;
 import org.apache.wicket.markup.html.form.ChoiceRenderer;
 import org.apache.wicket.markup.html.form.DropDownChoice;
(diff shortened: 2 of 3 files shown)


src/web/core/src/main/java/org/geoserver/web/data/layer/SQLViewAbstractPage.html
```diff
@@ -28,9 +28,12 @@
         </li>
     </ul>
 
-    <div class="button-group toolbar-sticky selfclear"><a href="#" class="form-button-save" wicket:id="save"><wicket:message
-        key="save">Save</wicket:message></a> <a href="#" class="form-button-cancel" wicket:id="cancel"><wicket:message
+    <div class="button-group toolbar-sticky selfclear">
+        <a href="#" class="form-button-save" wicket:id="save"><wicket:message key="save">Save</wicket:message></a>
+        <a href="#" class="form-button-save" wicket:id="back"><wicket:message key="back">Back</wicket:message></a>
```
jodygarnett, line 33: I am not sure this is consistent with our user interface form button naming. We tried to make the user interface consistent with:
* page:Apply, Save, Cancel
* nested-page: OK, Cancel

See http://geoserver.org/announcements/2020/09/26/geoserver-2-18-0-released.html

src/web/core/src/main/java/org/geoserver/web/data/layer/SQLViewAbstractPage.java
```diff
@@ -305,7 +307,16 @@ protected Component getComponentForProperty(
...
                         onSave();
                     }
-                });
+                }.setVisible(newView));
+
+        form.add(
+                new SubmitLink("back") {
+                    @Override
+                    public void onSubmit() {
+                        onSave();
```
jodygarnett, line 316: I do not understand what this `back` button is doing? the implementation looks identical to `save` above?

src/web/core/src/main/resources/GeoServerApplication.properties
```diff
@@ -979,6 +979,10 @@ resetFilter = Reset
 
 save = Save
 
+back = Back
+
+notice = Make sure to save the layer configuration to get the SQLView changes applied
```
jodygarnett, line 984: The translation key `notice` is a bit general, usually we should have a naming convention like `back.description` (if that is what this is used for).

jodygarnett (changes requested):
I presently do not understand what this PR is doing with a `back` button.

A change to functionality will require a corresponding change to the documentation (updating screen snap and documenting the new button).

jodygarnett:
The SQLView is a sub-page with `OK` and `Cancel` buttons, it is not expected to be saved unless the layer is saved also.

It is just that screen is so complicated it was done as a sub-page (rather than making the layer editor more complicated). If you look at the WPS Security configuration you can see similar use of sub-pages.

I think this concept could be made more clear with "breadcrumbs" at the top of the page to indicate you are still working in the layer editor.

src/web/core/src/main/resources/GeoServerApplication.properties
jodygarnett: You could make the argument for an `apply` button here, which would apply the entire layer editor changes (including this sub page). This would be handy as it would also the view to be revised quickly for debugging.

src/web/core/src/main/java/org/geoserver/web/data/layer/SQLViewAbstractPage.java
aaime: It actually does not save anything, just updates the featuretype in memory, does not save it in the catalog.
Make sense as this is a sub-page of the layer page. So it's "back" to the layer page, you have to save there.

aaime:
> The SQLView is a sub-page with `OK` and `Cancel` buttons

Nope, it has a "save" name in there now. That's generating confusing, it's not saving anything.

aaime:
> I think this concept could be made more clear with "breadcrumbs" at the top of the page to indicate you are still working in the layer editor.

There is no notion of paths or breadcrumbs anywhere in GeoServer, but the OGC API pages.

dromagnoli:
Ok @shadab-geo , based on latest feedbacks an no other additional comments, I would say  let's go with the "ok" button instead of back.
I will also rename the generic "notice" properties in GeoServerApplication.properties with ok.notice since it's a special notice related to the "subsequent action needed after pressing the ok button"

shadab-geo:
> Ok @shadab-geo , based on latest feedbacks an no other additional comments, I would say let's go with the "ok" button instead of back.
> I will also rename the generic "notice" properties in GeoServerApplication.properties with ok.notice since it's a special notice related to the "subsequent action needed after pressing the ok button"

Thanks, @dromagnoli I have those applied changes

PR #5289, opened 2021-09-21
Revisiting OGC-API Tiles links based on current status of the spec

dromagnoli:
OGC-API Tiles get some updates on July, modifying a couple of URL resources.
Code has been updated to reflect the current state.

# Checklist

### DIFF
--- a/src/community/ogcapi/ogcapi-tiles/src/main/java/org/geoserver/ogcapi/tiles/TileJSONBuilder.java
+++ b/src/community/ogcapi/ogcapi-tiles/src/main/java/org/geoserver/ogcapi/tiles/TileJSONBuilder.java
@@ -42,6 +42,8 @@ import org.geowebcache.layer.meta.LayerMetaInformation;
 import org.geowebcache.layer.meta.TileJSON;
 import org.geowebcache.layer.meta.VectorLayerMetadata;
 import org.geowebcache.layer.meta.VectorLayerMetadata.GeometryType;
+import org.geowebcache.mime.MimeException;
+import org.geowebcache.mime.MimeType;
 import org.locationtech.jts.geom.Envelope;
 import org.locationtech.jts.geom.LineString;
 import org.locationtech.jts.geom.MultiLineString;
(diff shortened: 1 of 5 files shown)


src/community/ogcapi/ogcapi-tiles/src/main/java/org/geoserver/ogcapi/tiles/TileJSONBuilder.java
```diff
@@ -128,25 +128,33 @@ public TileJSON build() throws FactoryException, TransformException, IOException
         String tilesURL;
         APIRequestInfo requestInfo = APIRequestInfo.get();
         String baseURL = requestInfo.getBaseURL();
+        boolean isVector =
+                styleId == null
+                        && "application/vnd.mapbox-vector-tile".equalsIgnoreCase(tileFormat);
```
aaime, line 133: There are other formats to represent a vector tile (geojson, topojson).
You can use the the GWC class ``MimeType.createFromFormat(tileFormat)`` and then call ``isVector`` to find out if a given mime type is meant for vector tiles.

src/community/ogcapi/ogcapi-tiles/src/main/resources/org/geoserver/ogcapi/tiles/tileMatrixSets.ftl
```diff
@@ -1,5 +1,7 @@
 <#global pagetitle="Tile matrix sets">
-<#global pagecrumbs="<li class='breadcrumb-item'><a href='"+serviceLink("")+"'>Home</a></li><li class='breadcrumb-item'><a href='"+serviceLink("collections")+"'>Collections</a></li><li class='breadcrumb-item active'>"+pagetitle+"</li>">
+<#global pagecrumbs="<li class='breadcrumb-item'><a href='"+serviceLink("")+"'>Home</a></li>
+<!-- li class='breadcrumb-item'><a href='"+serviceLink("collections")+"'>Collections</a></li-->
```
aaime, line 3: Uh?

aaime (changes requested):
Based on the spec of the day... it looks good (it may well be correct for another month or two!). See comments for minor issues.

src/community/ogcapi/ogcapi-tiles/src/main/java/org/geoserver/ogcapi/tiles/TileJSONBuilder.java
dromagnoli: Right. fix on its way.

src/community/ogcapi/ogcapi-tiles/src/main/resources/org/geoserver/ogcapi/tiles/tileMatrixSets.ftl
dromagnoli: @aaime 
I'm Removing the commented part which I forgot to definitely delete.

Explaining my change:

The previous version was putting TileMatrixSets link under "Collections".
So something like:
Home / Collections / TileMatrixSets

I suspect it was a refuse from copy and paste from the author which initially committed that new ftl based on the Collections one.
TileMatrixSets should not be "children" of Collections ->
Home / TileMatrixSets

Does it make sense?

aaime: Yes it does, thank you!

PR #5294, opened 2021-09-23
make all KVP of a WFS GetFeature Request available as extendedProperties

securedimensions:
Plugin development may need to access vendor specific parameters submitted with a WFS GetFeature request.
For example, it might be required that a Bearer Access Token is submitted via the GET Query: "access_token=...".

The proposed code change is very simple, as it adds the not existing setter function setExtendedProperties() to populate all KVP into the existing slot for "extendedProperties". One can later use the existing getter function getExtendedProperties() from Request.java for fetching the "rawKVP":
  
There is an issue in the [GeoServer Jira](https://osgeo-org.atlassian.net/browse/GEOS-10244) describing the contribution.


aaime:
See my comment on the Jira ticket.

PR #4646, opened 2021-01-15
JDBCConfig load and concurrency improvements

aaime:
See:
* [JDBCConfig does not drop its caches on a GeoServer reset call](https://osgeo-org.atlassian.net/browse/GEOS-9863)
* [JDBCConfig is deadlock prone](https://osgeo-org.atlassian.net/browse/GEOS-9838)
* [JDBCConfig throws concurrent modification exceptions if hit by high concurrent load right after startup](https://osgeo-org.atlassian.net/browse/GEOS-9519)

## Checklist

> Reviewing is a process done by project maintainers, mostly on a volunteer basis. We try to keep the overhead as small as possible and appreciate if you help us to do so by completing the following items. Feel free to ask in a comment if you have troubles with any of them.

**Submitting the PR does not require you to check all items, but by the time it gets merged, they should be either satisfied or not applicable.**

For all pull requests:

- [x] Confirm you have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/master/CONTRIBUTING.md) 
- [x] You have sent a Contribution Licence Agreement (CLA) as necessary (not required for small changes, e.g., fixing typos in documentation)
- [x] Make sure the first PR targets the master branch, eventual backports will be managed later. This can be ignored if the PR is fixing an issue that only happens in a specific branch, but not in newer ones.

The following are required only for core and extension modules (they are welcomed, but not required, for community modules):
- [ ] There is a ticket in Jira describing the issue/improvement/feature (a notable exemptions is, changes not visible to end users)
- [x] PR for bug fixes and small new features are presented as a single commit
- [x] Commit message must be in the form "[GEOS-XYZW] Title of the Jira ticket" (export to XML in Jira generates the message in this exact form)
- [x] The pull request contains changes related to a single objective. If multiple focuses cannot be avoided, each one is in its own commit and has a separate ticket describing it.
- [x] New unit tests have been added covering the changes
- [ ] This PR passes all existing unit tests (test results will be reported by Continuous Integration after opening this PR)
- [ ] This PR passes the [QA checks](https://docs.geoserver.org/latest/en/developer/qa-guide/index.html) (QA checks results will be reported by Continuous Integration after opening this PR)
- [ ] Commits changing the UI, existing user workflows, or adding new functionality, need to include documentation updates (screenshots, text)
- [ ] Commits changing the REST API, or any configuration object, should check if the REST API docs (Swagger YAML files and classic documentation) need to be updated.

PR #5299, opened 2021-09-23
[GEOS-10246] Jdbc performance improvement from unnecessary transactions

NielsCharlier:
[GEOS-10246](https://osgeo-org.atlassian.net/browse/GEOS-10246)

  https://osgeo-org.atlassian.net/browse/GEOS-10246

# Checklist


sikeoka:
This PR seems okay to me however it is significantly modifying @aaime's PR (#4646) so he should probably review this one.

aaime:
That PR prevented a catalog deadlock that could be solved (by memory) only with a restart. In a deploy with Tomcat and a watchdog hitting GeoServer, enough requests would queue up before full start up, that the lock up was very frequent. At the time, I added a test that was reproducing the issue, if run enough times (it's a concurrent load type, won't fail every single time). If the test pass, I'm happy with the changes.

NielsCharlier:
Thanks, both. To be sure I ran all the tests again a few times, no issue there. The transactional behaviour is preserved for db access but avoided for cache access.

PR #5309, opened 2021-09-28
[GEOS-10257] Features templating UI preview fails on secured layers - [GEOS-10258] Features templating cache  might retain orphan templates

taba90:
# Checklist


src/community/features-templating/features-templating-web/src/main/java/org/geoserver/featurestemplating/web/TemplatePreviewPanel.java
```diff
@@ -381,10 +397,29 @@ private Catalog getCatalog() {
...
                 RequestConfig.custom().setConnectTimeout(60000).setSocketTimeout(60000).build();
-        return HttpClientBuilder.create().setDefaultRequestConfig(clientConfig).build();
+        CookieStore cookieStore = new BasicCookieStore();
+        RequestAttributes attributes = RequestContextHolder.getRequestAttributes();
+        if (attributes != null) {
+            BasicClientCookie cookie =
+                    new BasicClientCookie("JSESSIONID", attributes.getSessionId());
+            HttpServletRequest request = request();
+            cookie.setPath(request.getContextPath());
+            Calendar calendar =
```
aaime, line 410: Move the comment the line above, so that it does not get wrapped?

src/community/features-templating/features-templating-web/src/main/java/org/geoserver/featurestemplating/web/TemplatePreviewPanel.java
```diff
@@ -381,10 +397,29 @@ private Catalog getCatalog() {
...
+                    Calendar.getInstance(); // gets a calendar using the default time zone and
+            // locale.
+            int maxInactive = request.getSession().getMaxInactiveInterval();
```
aaime, line 413: Interesting... I don't know what you're doing here :-D
Can you explain?

src/community/features-templating/features-templating-rest/src/test/java/org/geoserver/featurestemplating/rest/TemplateRestControllerTest.java
```diff
@@ -420,5 +430,57 @@ private void newTemplateInfo(
         info.setFeatureType(featureType);
         TemplateInfoDAO.get().saveOrUpdate(info);
         TemplateFileManager.get().saveTemplateFile(info, rawTemplate);
+        return info;
+    }
+
+    @Test
+    public void testUpdatedCacheOnPut() throws Exception {
```
aaime, line 437: I would have expected this test to pass even before the changes... aren't the templates reloaded based on the last modification time? But I guess one has to wait up to a second to rely on that mechanism, while here we have an instant effect?

That said, I'm still confused... the title says "templating cache might retain orphan templates" and the description says "the template is moved from the feature type folder to the ws one" but this test only checks an update in place.

aaime (changes requested):

src/community/features-templating/features-templating-web/src/main/java/org/geoserver/featurestemplating/web/TemplatePreviewPanel.java
taba90: fixed

aaime (approved):

aaime:
Looks good, merged, thanks!

PR #5314, opened 2021-09-28
[GEOS-10259][GEOS-10260] Use class action-group to align actions horizontally across table

jodygarnett:
[GEOS-10259](https://osgeo-org.atlassian.net/browse/GEOS-10259)

[GEOS-10260](https://osgeo-org.atlassian.net/browse/GEOS-10260)

A small user interface change that makes me happy.

Documentation not updated as part of this change as the screen snaps are close enough not to confuse anyone.

1) workspaces

image

2) stores

image

3) layers, layer groups, styles etc...

image
image
image

4) security 

image
image
image
image
image

5) security panels (separate commit as title is introduced to provide visual separator)

These ones had heading breaks already:

image

User/Group page has expanding sections for each service, made use of panel title to provide a visual separation (now that action group does not provide a block of empty space.

image

Similar for role service (shown with two expanded roles services):

image

# Checklist

- [x] There is an issue https://osgeo-org.atlassian.net/browse/GEOS-10259

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/css/geoserver.css
+++ b/src/web/core/src/main/java/org/geoserver/web/css/geoserver.css
@@ -927,7 +927,8 @@ padding-top: 1em;
 }
 /* action group above tables */
 .action-group li {
-  display: inline;
+  display: inline-block;
+  margin-right: 10px;
 }
 
 /* internationalization */
--- a/src/web/core/src/main/java/org/geoserver/web/data/layer/LayerPage.html
+++ b/src/web/core/src/main/java/org/geoserver/web/data/layer/LayerPage.html
@@ -11,7 +11,7 @@
     </wicket:fragment>
     
     <wicket:fragment wicket:id="header">
-      <ul>
+      <ul class="action-group">
         <li><a class="add-link" href="#" wicket:id="addNew"><wicket:message key="addNew" /></a></li>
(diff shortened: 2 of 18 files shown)


aaime:
It's fine by me, but maybe something you want to run also on the devel list.
Lots of screenshots to re-do around the docs and external training material I suppose.

aaime:
Actually wait, maybe a tad too tightly packed? Let's discuss on list anyways:

image

MichelGabriel:
To make it less tight you can add some margin to the right side of the list elements in: `src/web/core/src/main/java/org/geoserver/web/css/geoserver.css`

around line 929 you can do something like:
```
.action-group li {
  display: inline;
  margin-right: 5px;
}
```

if that doesn't work you can try to add the margin to the `<a>`, like this:
```
.action-group li a {
  margin-right: 5px;
}
```

jodygarnett:
> Actually wait, maybe a tad too tightly packed? Let's discuss on list anyways:
> 
> image

Increased padding to 10px and it is less packed:
image

ianturton (approved):

jodygarnett:
Thanks for the review Ian, going to let this hang out for feedback. 

As this is a non functional change would it be okay to backport to 2.20.x branch for the anticipated release.

aaime:
Ok, it got a good amount of exposure and no negative feedback. Squash-merged for ease of backporting.

PR #5319, opened 2021-10-02
Remove setFinalStaticField and use new setter

bjornharrtell:
This removes the use of reflection to override a static final in geotools, which is forbidden since Java 12, and instead uses new public API introduced in geotools PR https://github.com/geotools/geotools/pull/3612.

### DIFF
--- a/src/extension/netcdf-out/src/test/java/org/geoserver/wcs2_0/WCSNetCDFTest.java
+++ b/src/extension/netcdf-out/src/test/java/org/geoserver/wcs2_0/WCSNetCDFTest.java
@@ -67,20 +67,6 @@ public class WCSNetCDFTest extends WCSNetCDFBaseTest {
         super.setUpTestData(testData);
     }
 
-    private void setFinalStaticField(String fieldName, boolean value)
-            throws NoSuchFieldException, SecurityException, IllegalArgumentException,
-                    IllegalAccessException {
-        // Playing with System.Properties and Static boolean fields can raises issues
-        // when running Junit tests via Maven, due to initialization orders.
-        // So let's change the fields via reflections for these tests
-        Field field = NetCDFCRSUtilities.class.getDeclaredField(fieldName);
-        field.setAccessible(true);
-        Field modifiersField = Field.class.getDeclaredField("modifiers");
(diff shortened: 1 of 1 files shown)


bjornharrtell:
@aaime - this would allow merge of https://github.com/geotools/geotools/pull/3612 with finishing GeoServer Java 17 compat effort.

aaime:
Finally had a few minutes to try out a combined build with the GeoTools Java 17. Unfortunately, it does not compile:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project gs-platform: Compilation failure
[ERROR] /home/aaime/devel/git-gs/src/platform/src/main/java/org/geoserver/platform/resource/FileSystemResourceStore.java:[240,39] unreported exception java.lang.Throwable; must be caught or declared to be thrown
[ERROR] 
```

I was surprised then checked the code in question vs the Java 17 API:

```java
                    @Override
                    @SuppressWarnings("deprecation") // finalize is deprecated in Java 9
                    protected void finalize() throws IOException {
                        if (!closed) {
                            String warn =
                                    "There is code leaving resource input streams open, locks around them might not be cleared! ";
                            if (!TRACE_ENABLED) {
                                warn +=
                                        "Add -D"
                                                + GS_LOCK_TRACE
                                                + "=true to your JVM options to get a full stack trace of the code that acquired the input stream";
                            }
                            LOGGER.warning(warn);

                            if (TRACE_ENABLED) {
                                LOGGER.log(
                                        Level.WARNING,
                                        "The unclosed input stream originated on this stack trace",
                                        tracer);
                            }
                        }
                        super.finalize();
                    }
                };
```

Here finalize throws a IOException. However in Java 17 ``FileInputStream`` has no finalize method, and ``java.lang.Object`` indeed [throws a Throwable](https://cr.openjdk.java.net/~iris/se/17/latestSpec//api/java.base/java/lang/Object.html#finalize()).

Aand... my 15 minutes are finished, family is back and cannot proceed further....

bjornharrtell:
@aaime thanks for giving it a try and dang it that I apparently didn't get it right 😆 I'll see what I can do about it hopefully in near future.

PR #5320, opened 2021-10-04
[GEOS-9950] LayerPreview use format string if translation not available

jodygarnett:
[GEOS-9950](https://osgeo-org.atlassian.net/browse/GEOS-9950)

This avoids filling up the logs if a localization is not available for a given output format.

Experimented with skipping output formats that did not have a localization, but that was not fair to third-party output formats being developed. 

# Checklist

- [x] There is an issue https://osgeo-org.atlassian.net/browse/GEOS-9950

### DIFF
--- a/src/web/demo/src/main/java/org/geoserver/web/demo/MapPreviewPage.java
+++ b/src/web/demo/src/main/java/org/geoserver/web/demo/MapPreviewPage.java
@@ -275,11 +275,18 @@ public class MapPreviewPage extends GeoServerBasePage {
         return f;
     }
 
+    /**
+     * Translate format (if translation available).
+     *
+     * @param prefix protocol
+     * @param format output format
+     * @return format translation (defaults to format if unavailable)
+     */
     private String translateFormat(String prefix, String format) {
         try {
             return getLocalizer().getString(prefix + format, this);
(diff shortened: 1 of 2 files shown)


src/web/demo/src/main/java/org/geoserver/web/demo/MapPreviewPage.java
```diff
@@ -218,7 +218,7 @@ private Component buildJSWMSSelect(
         RepeatingView wmsFormats = new RepeatingView("wmsFormats");
         for (int i = 0; i < wmsOutputFormats.size(); i++) {
             String wmsOutputFormat = wmsOutputFormats.get(i);
-            String label = translateFormat("format.wms.", wmsOutputFormat);
+            String label = translateFormat( "format.wms.", wmsOutputFormat );
```
jodygarnett, line 221: To experiment with skipping un translated output formats use identity check:

```
if( label == wmsOutputFormat) continue;  // Skip non localized output format
```

However this is unfair to community modules developing output formats,  as they are unable to update GeoServerApplication.properties

aaime:
Yup thanks!

jodygarnett:
Back ported to 2.20.x for the next release.

PR #5324, opened 2021-10-05
[GSIP-198] NSIS Windows Installer script

GeoSander:
This will update the current Windows installer scripts with a new one (also created with NSIS).
It includes some of the latest changes/suggestions by the user community.

The NSIS script is already in use by an Azure container, which is triggered by Jenkins. 
The resulting `.exe` file is pushed to SourceForge. New installers will be created following the proper naming convention `GeoServer-<version>-winsetup.exe`.


jodygarnett:
Thanks @GeoSander, github flagged you as a first-time contributor so I just approved running the build actions now.

This work is covered by GeoCat CLA with OSGeo.

GeoSander:
Not sure why some actions failed. I noticed this error in the build report:
```
Error:  Failed to execute goal on project gs-main: Could not resolve dependencies for project org.geoserver:gs-main:jar:2.21-SNAPSHOT: Failed to collect dependencies at org.geotools:gt-wfs-ng:jar:27-SNAPSHOT: Failed to read artifact descriptor for org.geotools:gt-wfs-ng:jar:27-SNAPSHOT: Could not transfer artifact org.geotools:gt-wfs-ng:pom:27-SNAPSHOT from/to osgeo-snapshots (https://repo.osgeo.org/repository/snapshot/): transfer failed for https://repo.osgeo.org/repository/snapshot/org/geotools/gt-wfs-ng/27-SNAPSHOT/gt-wfs-ng-27-SNAPSHOT.pom, status: 502 Bad Gateway
```

Seems that Maven has a hard time retrieving the GeoTools artifact?
Anyway, if this was temporary, I hope that someone can trigger the build action again - I am not entitled to do so.

aaime:
I've kicked the builds. Indeed they should not fail as there is nothing trying to build the installer.
Thinking out loud, the Linux builds are actually running an assembly to verify the changes in the PR are not breaking the release:
https://github.com/geoserver/geoserver/blob/main/.github/workflows/linux_jdk8.yml#L40

Wondering if the Windows build could be modified to check the Windows installer can also still be built:
https://github.com/geoserver/geoserver/blob/main/.github/workflows/windows.yml
Mind, just thinking out loud here!

aaime:
Pity. We're now in a situation where if a build on the jenkins server is broken, it pretty much stays that way. The real gateway is here, before merge. 
But it's good to know that at least we'll know the installer does not work before release day :-D

GeoSander:
Note that if Jenkins fails to build GeoServer, the post-build action is _not_ triggered, which means that _no_ Windows installer will be built and pushed to SourceForge.

jodygarnett:
Quickly:
- We can use pandoc locally to update to your rtf file (so you just have a static file to work with). Duplication (sigh)
- please replace the Tanuki with whatever is approprate for the new wrapper (did not look)
- Sorry for the duplicate infromation in LICENSE.txt and NOTICE.txt (duplication sigh), need to generate one from the other

Yes the installer should show the content from NOTICE.md, generating our a license.rtf file from NOTICE.md is the way to go.

PR #5375, opened 2021-10-30
[GEOS-10282] Follow up, add enforcer check

aaime:
[GEOS-10282](https://osgeo-org.atlassian.net/browse/GEOS-10282)

Follows up with GEOS-10282, adding a QA rule that should prevent improperly encoded files from landing in GeoServer (was not the issue, but we had a few).

# Checklist

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -2419,6 +2419,53 @@
         </plugins>
       </build>
     </profile>
+
+    <profile>
+      <id>enforcer</id>
+      <activation>
+        <property>
+          <name>qa</name>
+        </property>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <artifactId>maven-enforcer-plugin</artifactId>
+            <version>3.0.0</version>
+            <executions>
+              <execution>
+                <id>localization-encodings</id>
+                <goals>
+                  <goal>enforce</goal>
+                </goals>
+                <configuration>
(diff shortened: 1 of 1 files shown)


mprins (commented):
I haven't tried with `requireEncoding`but I'm sure you can have multiple invocations of the same rule in the `rules` in general (see eg. https://maven.apache.org/enforcer/enforcer-rules/requireProperty.html ) so I think you can have

```xml
...
  <execution>
    <id>localization-encodings</id>
    <goals>
      <goal>enforce</goal>
    </goals>
    <configuration>
      <rules>
        <requireEncoding>
          <encoding>UTF-8</encoding>
          <includes>src/main/resources/GeoServerApplication*.utf8.properties</includes>
        </requireEncoding>
        <requireEncoding>
          <encoding>ISO-8859-1</encoding>
          <acceptAsciiSubset>true</acceptAsciiSubset>
          <includes>src/main/resources/GeoServerApplication*.properties</includes>
          <excludes>src/main/resources/GeoServerApplication*.utf8.properties</excludes>
        </requireEncoding>
      </rules>
      <fastFail>false</fastFail>
    </configuration>
  </execution>
...
```

which would then only require one invocation of the `maven-enforcer-plugin` instead of two

aaime:
Good to know, PR updated.

PR #5402, opened 2021-11-16
[GEOS-10287] loadJSON function is added to OSEO-STAC

ydeliorman:
[GEOS-10287](https://osgeo-org.atlassian.net/browse/GEOS-10287)

loadJSON functionality is added to OSEO-STAC module which enables to load a JSON file in geoserver data directory.
  

# Checklist

### DIFF
--- a/src/community/oseo/oseo-stac/src/test/java/org/geoserver/ogcapi/stac/LoadJSONTest.java
+++ b/src/community/oseo/oseo-stac/src/test/java/org/geoserver/ogcapi/stac/LoadJSONTest.java
@@ -1,0 +1,76 @@
+package org.geoserver.ogcapi.stac;
+
+import static org.junit.Assert.assertEquals;
+
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import org.geoserver.config.GeoServerDataDirectory;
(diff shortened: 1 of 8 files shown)


src/community/oseo/oseo-stac/src/test/java/org/geoserver/ogcapi/stac/LoadJSONTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Test
+    public void testReadAndEvalJSON() throws Exception {
+        Document dom = getAsJSoup("ogc/stac/collections/SENTINEL2?f=html");
+
+        assertEquals("attribute1 => 1 attribute2 => 2", dom.select("h2").text());
+    }
+
+    @Test
+    public void testReadFileOutsideOfDataDir() throws Exception {
```
taba90, line 40: Seems to me that this method doesn't test what it claims to test. Could you clarify, fix it?

taba90 (commented):
Overall looks good to me.
I've only a point regarding the test for checking that file loading from outside the data dir is not permitted. See below.

ydeliorman:
review is done

taba90 (approved):
The pr looks fine. Tests have been added also to check that file outside the data dir are not loaded. Run the build locally and tests are all passing.

PR #5408, opened 2021-11-22
[GEOS-10312]: Adding support for COG mosaic on OSEO plugin

dromagnoli:
[GEOS-10312](https://osgeo-org.atlassian.net/browse/GEOS-10312)

  

# Checklist


doc/en/user/source/community/opensearch-eo/automation.rst
```diff
@@ -197,4 +197,27 @@ using the full name, in which case they will be all picked, or by using zero-bas
...
+   * - Attribute
+     - Description
+   * - cog
+     - Set it to true to specify the layer is made of COG datasets
+   * - cogUser
+     - (Optional) Credential to be set whenever basic HTTP authentication is needed to access the COG Datasets or an S3 Access KeyID is required
+   * - cogPassword
+     - (Optional) Password for the above user OR Secret Access Key for the above S3 KeyId.
+   * - cogRangeReader
+     - (Optional) Specify the desired RangeReader implementation. (default is HTTP based)
```
aaime, line 221: Examples of other possible values, or references of where to pick them?

src/community/oseo/oseo-rest/src/main/java/org/geoserver/opensearch/rest/CollectionLayerManager.java
```diff
@@ -310,6 +329,32 @@ private void configureSeparateBandsMosaic(
...
+        } else if (source instanceof SourceSPIProvider) {
+            return ((SourceSPIProvider) source).getReaderSpi();
+        }
+        return null;
+    }
+
+    private CogConfiguration getCogConfiguration(CollectionLayer layerConfiguration) {
+        CogConfiguration cogConfiguration = new CogConfiguration();
+        if (layerConfiguration != null) {
+            String rangeReader = layerConfiguration.getCogRangeReader();
```
aaime, line 349: Wondering if there is any sort of validation about the range reader chosen... like, what if there is a typo for example?
It would be nice to get an understandable error message.

aaime (changes requested):

doc/en/user/source/community/opensearch-eo/automation.rst
dromagnoli: Is right in the next line :)

See :ref:`cog_plugin_rangereader` from the COG plugin documentation, for the list of supported RangeReader implementations.

src/community/oseo/oseo-rest/src/main/java/org/geoserver/opensearch/rest/CollectionLayerManager.java
dromagnoli: Added some more code attempting a classforname

aaime (approved):

PR #5415, opened 2021-11-25
[GEOS-10316] Fix startup.sh to handle multiple java options

mwerlen:
[GEOS-10316](https://osgeo-org.atlassian.net/browse/GEOS-10316)

This PR aims to fix issue GEOS-10316 introduced by GEOS-10033.
startup.sh and shutdown.sh scripts were not working as expected when using multiples java options in JAVA_OPTS environment variable.
  

# Checklist

### DIFF
--- a/src/release/bin/shutdown.sh
+++ b/src/release/bin/shutdown.sh
@@ -65,4 +65,5 @@ if [ ! -r "${GEOSERVER_HOME}/start.jar" ]; then
 fi
 
 cd "${GEOSERVER_HOME}" || exit 1
-exec "${_RUNJAVA}" "${JAVA_OPTS:--DnoJavaOpts}" -DSTOP.PORT=8079 -DSTOP.KEY=geoserver -jar start.jar --stop
+IFS=$(printf '\n\t ')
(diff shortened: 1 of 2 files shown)


aaime:
Looks like it broke support for path in spaces, though? A test:

```
> ./bin/startup.sh 
GEOSERVER_HOME environment variable not found, using current 
directory.  If not set then running this script from other 
directories will not work in the future.
GEOSERVER DATA DIR is /home/aaime/devel/git-gs/src/target/release/bla bla bla/gs-2.21.x/data_dir
Error: Could not find or load main class s

```

mwerlen:
POSIX shell programming is a nightmare. I used a bash idiom `IFS=$'\n\t '` instead of a POSIX shell one `IFS=$(printf '\n\t ')`
It was sufficient to let my tests working.

If someone wants to play with them, they are here :  https://github.com/sogelink/geoserver/commit/7f368c7849acd6ec89c41a1dfa349bc49faa5bae

Could you test again ?

aaime:
Good, merging then! Thanks for the quick turnaround!

PR #5417, opened 2021-11-26
[GEOS-10318] Resolve attribute prefix name from GeoServer catalog on CSV output for complex features

fernandor777:
[GEOS-10318](https://osgeo-org.atlassian.net/browse/GEOS-10318)

When we activate complex to simple features automatic transformation, CSV output format for complex features doesn't resolve namespace URIs to prefixes on attributes names, making the attributes output use the entire URLs even having the namespace prefix registered on GeoServer catalog.

This PR introduce a resolving method to allow having attributes with prefixes.

JIRA ticket:
https://osgeo-org.atlassian.net/browse/GEOS-10318

# Checklist

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/response/CSVOutputFormat.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/response/CSVOutputFormat.java
@@ -24,7 +24,7 @@ import java.util.regex.Pattern;
 import org.apache.commons.lang3.StringUtils;
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.xsd.XSDElementDeclaration;
-import org.eclipse.xsd.impl.XSDElementDeclarationImpl;
+import org.geoserver.catalog.NamespaceInfo;
 import org.geoserver.config.GeoServer;
 import org.geoserver.feature.FlatteningFeatureCollection;
 import org.geoserver.platform.Operation;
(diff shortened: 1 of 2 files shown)


src/wfs/src/main/java/org/geoserver/wfs/response/CSVOutputFormat.java
```diff
@@ -384,4 +385,26 @@ public String getCapabilitiesElementName() {
...
+    String resolveNamespacePrefixName(String attributeName) {
+        if (StringUtils.isBlank(attributeName) || !attributeName.contains(":")) {
+            return attributeName;
+        }
+        int lastIndexOfSeparator = attributeName.lastIndexOf(":");
+        String namespaceUri = attributeName.substring(0, lastIndexOfSeparator);
+        NamespaceInfo namespace = this.gs.getCatalog().getNamespaceByURI(namespaceUri);
+        if (namespace != null) {
+            String localName =
+                    attributeName.substring(lastIndexOfSeparator + 1, attributeName.length());
```
aaime, line 405: Should not happen but... what if someone silly put a ":" at the end of the attribute name?

fernandor777: Indeed, good point.  We already allow ":" characters in the middle to support URIs definitions (excluding the last one taking it as namespace separator), so any attribute using it as final character should not be supported by this resolver method.  Applying an extra validation condition for this, thanks!

aaime:
Merged, thanks!

jodygarnett:
So this is available for 2.21-RC the, just updating the jira ticket.

PR #5418, opened 2021-11-29
[GEOS-10319] includeFlat a JSON database field to a features template

ydeliorman:
[GEOS-10319](https://osgeo-org.atlassian.net/browse/GEOS-10319)

includeFlat keyword now loads a JSON database field in freemarker templates.

For example, if we have the following JSON property:

```
myJSONProperty
{
   a: 10;
   b: 20;
}
```

and the following template:

```
flat:
{
   "properties": {
      c: 20;
      "$includeFlat": "${myJSONProperty};
   }
}
```

the result will be this:

```
"properties": {
      c: 20;
      a: 10;
      b: 20;
  }
```

# Checklist

### DIFF
--- a/doc/en/user/source/community/features-templating/directives.rst
+++ b/doc/en/user/source/community/features-templating/directives.rst
@@ -1070,7 +1070,17 @@ Regarding the :code:`$includeFlat` option is worth mentioning that in a JSON con
 * If a JSON object is included, then its properties are directly included in-place, which makes sense only within another object. 
 * If instead a JSON array is included, then its values are directly included in-place, which makes sense only within another array.
 
-The following JSON snippet shows the four possible syntax options for template inclusion:
+The following JSON snippet shows the five possible syntax options for template inclusion:
+
+${property}:
+
+.. code-block:: json
+   :linenos: 
+
+    {
+       "aProperty": "10", 
+       "bProperty": "20"
+    }
 
 .. code-block:: json
    :linenos: 
@@ -1081,7 +1091,8 @@ The following JSON snippet shows the four possible syntax options for template i
        "anArray" : [
           "$include{arrayElement.json}", 
           "$includeFlat{subArray.json}" 
-       ]
+       ],
+      "$includeFlat": "${property}"
     }
 
 Notes:
@@ -1094,6 +1105,7 @@ Notes:
    element, or the creation of a nested array.
 4) The ``subArray.json`` template (line 6) must be an array itself, the container array will be stripped and
    its values directly integrated inside ``anArray``.
+5) The ``${property}`` template (line 7) evaluates to a json, its properties will be passed to parent json without ``"$includeFlat"`` keyword.
 
 
 XML based templates (GML)
--- a/doc/en/user/source/community/features-templating/directives.rst
+++ b/doc/en/user/source/community/features-templating/directives.rst
@@ -1070,17 +1070,7 @@ Regarding the :code:`$includeFlat` option is worth mentioning that in a JSON con
 * If a JSON object is included, then its properties are directly included in-place, which makes sense only within another object. 
 * If instead a JSON array is included, then its values are directly included in-place, which makes sense only within another array.
 
-The following JSON snippet shows the five possible syntax options for template inclusion:
-
-${property}:
-
-.. code-block:: json
-   :linenos: 
-
-    {
-       "aProperty": "10", 
-       "bProperty": "20"
-    }
+The following JSON snippet shows the four possible syntax options for template inclusion:
 
 .. code-block:: json
    :linenos: 
@@ -1105,7 +1095,20 @@ Notes:
    element, or the creation of a nested array.
 4) The ``subArray.json`` template (line 6) must be an array itself, the container array will be stripped and
    its values directly integrated inside ``anArray``.
-5) The ``${property}`` template (line 7) evaluates to a json, its properties will be passed to parent json without ``"$includeFlat"`` keyword.
+
+In case of an includeFlat directive is specified and it's attribute value is a property interpolation directive, if the property name evaluates to a json it gets included flat in the final output.
+
+${property}:
+
+.. code-block:: json
+   :linenos: 
+
+    {
+       "aProperty": "10", 
+       "bProperty": "20"
+    }
+
+The ``${property}`` template (line 7) evaluates to a json, its properties will be passed to parent json without ``"$includeFlat"`` keyword.
 
 
 XML based templates (GML)
--- a/src/community/features-templating/features-templating-core/src/main/java/org/geoserver/featurestemplating/builders/impl/DynamicIncludeFlatBuilder.java
+++ b/src/community/features-templating/features-templating-core/src/main/java/org/geoserver/featurestemplating/builders/impl/DynamicIncludeFlatBuilder.java
@@ -3,13 +3,8 @@ package org.geoserver.featurestemplating.builders.impl;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Map;
-import org.geoserver.featurestemplating.builders.TemplateBuilder;
-import org.geoserver.featurestemplating.builders.TemplateBuilderMaker;
 import org.geoserver.featurestemplating.builders.visitors.TemplateVisitor;
-import org.geoserver.featurestemplating.readers.JSONTemplateReader;
-import org.geoserver.featurestemplating.readers.TemplateReaderConfiguration;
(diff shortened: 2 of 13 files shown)


bradh (commented):
I'm not sure that `includeFlat` is really a good description - I had no idea at all until I read the ticket at https://osgeo-org.atlassian.net/browse/GEOS-10319.

Also, it isn't clear what happens if the JSON being included has nested structure, like:

```
thingToInclude:
{
   a: 10;
   b: { c: 4, a: 8};
}
```

ydeliorman:
@bradh includeFlat is an [existing directive in the features templating](https://docs.geoserver.org/latest/en/user/community/features-templating/directives.html#json-based-templates-geojson-json-ld). The client asked to implement this functionality using the same directive name.

The result is going to be like below in your example:
```
"properties": {
      c: 20;
      a: 10;
      b: { c: 4, a: 8};
}
```

nmco:
Rephrasing what @ydeliorman wrote @bradh, we aimed at keeping some consistency and hence used the same name that's used in features templating.

doc/en/user/source/community/features-templating/directives.rst
```diff
@@ -1070,7 +1070,17 @@ Regarding the :code:`$includeFlat` option is worth mentioning that in a JSON con
...
+
+${property}:
+
+.. code-block:: json
+   :linenos: 
+
+    {
+       "aProperty": "10", 
+       "bProperty": "20"
+    }
```
taba90, line 1083: Looks a bit confusing. I would leave this pieces of documentation as it is with the four example that explains the four possible syntax of the include directive then I would a paragraph below explaining that in case of an includeFlat directive  is specified and its attribute value is a property interpolation directive, if the property name evaluates to a json it gets included flat in the final output.

src/community/features-templating/features-templating-core/src/main/java/org/geoserver/featurestemplating/builders/impl/DynamicIncludeFlatBuilder.java
```diff
@@ -0,0 +1,70 @@
...
+        TemplateBuilderMaker maker = configuration.getBuilderMaker();
+        maker.namespaces(configuration.getNamespaces());
+        writer.startObject(getKey(context), getEncodingHints());
+        jsonTemplateReader.getBuilderFromJson(getKey(context), node, this, maker);
+        for (TemplateBuilder child : getChildren()) child.evaluate(writer, context);
+        writer.endObject(getKey(context), encodingHints);
+    }
+
+    private boolean hasDynamic(JsonNode node) {
+        return node.toString().contains("${");
```
taba90, line 63: If my memory serves me well this methods as well as some pieces of the evaluate method are also present in the DynamicMergeBuilder. Maybe is a better idea to put the common pieces in the DynamicalueBuilder parent class  or in a separate one, avoiding code duplication.

src/community/features-templating/features-templating-core/src/main/java/org/geoserver/featurestemplating/readers/RecursiveJSONParser.java
```diff
@@ -127,6 +131,26 @@ private ObjectNode expandIncludesInObject(ObjectNode input) throws IOException {
             result = processMergeDirective(result);
         }
 
+        // iterate here again to put DYNAMIC_INCLUDE_FLAT_KEY as a parent node
```
taba90, line 134: This method has become too long. Is better to refact it a bit i.e. by putting this code into a separate method.

src/community/features-templating/features-templating-core/src/test/java/org/geoserver/featurestemplating/builders/impl/DynamicIncludeFlatBuilderTest.java
```diff
@@ -0,0 +1,126 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.opengis.feature.Feature;
+import org.opengis.feature.simple.SimpleFeature;
+import org.opengis.feature.simple.SimpleFeatureType;
+import org.xml.sax.helpers.NamespaceSupport;
+
+public class DynamicIncludeFlatBuilderTest extends DataTestCase {
+
+    private static final String BASE =
```
taba90, line 32: no merge here, it is an inclusion this should not be needed.

src/community/features-templating/features-templating-core/src/main/java/org/geoserver/featurestemplating/readers/RecursiveJSONParser.java
```diff
@@ -127,6 +131,26 @@ private ObjectNode expandIncludesInObject(ObjectNode input) throws IOException {
...
+        if (result.has(INCLUDE_FLAT_KEY)) {
+            ObjectNode dynamicIncludeFlatResult = mapper.getNodeFactory().objectNode();
+            Iterator<String> namesSecond = input.fieldNames();
+            while (namesSecond.hasNext()) {
+                String name = namesSecond.next();
+                JsonNode node = input.get(name);
+                if (isDynamicIncludeFlat(node)) {
+                    String key = DYNAMIC_INCLUDE_FLAT_KEY.concat(name);
```
taba90, line 142: doesn't have that much sense this here. It was done for the dynamic merge because in that case we could have different attribute names with the merge directive and by replacing all of them with a unique key in the same node we would have come to repeat attribute names. In this case there is no way to avoid this since we would always have the $includeFlat key.

src/community/features-templating/features-templating-core/src/test/java/org/geoserver/featurestemplating/builders/impl/DynamicIncludeFlatBuilderTest.java
```diff
@@ -0,0 +1,126 @@
...
+        fb.add(DYNAMIC_METADATA);
+        fb.add(STATIC_METADATA);
+        fb.add("dynamic value result");
+        jsonFieldSimpleFeature = fb.buildFeature("jsonFieldSimpleType.1");
+
+        store = new FileSystemResourceStore(new File("src/test/resources/jsonIncludes"));
+    }
+
+    @Test
+    public void testObtainDynamicIncludeFlatKeyword() throws IOException {
```
taba90, line 73: I would move this test into the JsonIncludesTest or the JsonTemplateReaderTest class.

taba90 (changes requested):
Overall looks good, but some things can be improved.

doc/en/user/source/community/features-templating/directives.rst
ydeliorman: done

src/community/features-templating/features-templating-core/src/main/java/org/geoserver/featurestemplating/builders/impl/DynamicIncludeFlatBuilder.java
ydeliorman: yes, thanks for that. put into DynamicValueBuilder

src/community/features-templating/features-templating-core/src/main/java/org/geoserver/featurestemplating/readers/RecursiveJSONParser.java
ydeliorman: removed as you suggested in your other comment since we do not have to identify 2 different nodes that as includeFlat keyword

src/community/features-templating/features-templating-core/src/test/java/org/geoserver/featurestemplating/builders/impl/DynamicIncludeFlatBuilderTest.java
ydeliorman: removed BASE string and other unused variable after deleting that.

src/community/features-templating/features-templating-core/src/main/java/org/geoserver/featurestemplating/readers/RecursiveJSONParser.java
ydeliorman: done

src/community/features-templating/features-templating-core/src/test/java/org/geoserver/featurestemplating/builders/impl/DynamicIncludeFlatBuilderTest.java
ydeliorman: moved into JsonIncludesTest

ydeliorman:
all feedbacks are applied @taba90.

taba90 (approved):

taba90:
looks good to me. I've ran also the test locally for both oseo and features templating and they are passing. Merging....

PR #5431, opened 2021-12-06
Add keycloak login button

leashjay:
# Keycloak Login Button Addition to Home Page  
Keycloak button added to home page alongside username-password login method. Allows keycloak filter to function alongside other security filter methods (rather than directing straight to KC splash panel) when filter is enabled.
  
# Checklist

- [x] (N/A - Feature) There is an issue in the [GeoServer Jira](https://osgeo-org.atlassian.net/browse/GEOS/summary) (except for changes that do not affect administrators or end users in any way).
- [x]  (N/A - Feature) Commit message(s) must be in the form ``[GEOS-XYZWV] Title of the Jira ticket``.


src/community/security/keycloak/src/main/java/org/geoserver/security/keycloak/GeoServerKeycloakFilter.java
```diff
@@ -291,11 +301,8 @@ protected AuthResults loadAuthn(HttpServletRequest request) {
 
     private static void logHttpRequest(Level level, HttpServletRequest request) {
         if (LOG.isLoggable(level)) {
-            LOG.log(level, "request.method   = " + request.getMethod());
```
aaime, line 294: Why was the logging removed here?

aaime (commented):
Seems reasonable to me, just added a question.
@afabiani can you also have a look? Do you have the ability to quickly test this change  interactively too?

src/community/security/keycloak/src/main/java/org/geoserver/security/keycloak/GeoServerKeycloakFilter.java
leashjay: Ah, I had put a reasonable amount it to resolve and issue with the redirects, I might have cleaned up this one by mistake. I will reinstated

afabiani (approved):

PR #5432, opened 2021-12-06
[GEOS-10324] adding arbitrary number of columns to ogcLink tables

ydeliorman:
[GEOS-10324](https://osgeo-org.atlassian.net/browse/GEOS-10324)

OgcLink tables in OSEO now support adding an arbitrary number of columns.

# Checklist


src/community/oseo/oseo-core/src/test/resources/postgis.sql
```diff
@@ -206,7 +206,10 @@ create table collection_ogclink (
   "method" varchar,
   "code" varchar,
   "type" varchar,
-  "href" varchar
+  "href" varchar,
```
aaime, line 209: Nope... the ``postgis.sql`` is linked from the documentation as the reference structure for the database. We cannot have testing bits in it.
That said, you can run ``alter table`` to add columns in ``collection_test_data.sql`` and ``product_test_data.sql``, those files are test only instead.

src/community/oseo/oseo-rest/src/main/java/org/geoserver/opensearch/rest/AbstractOpenSearchController.java
```diff
@@ -217,7 +223,35 @@ protected OgcLinks buildOgcLinksFromFeature(Feature feature, boolean notFoundOnE
         List<OgcLink> links = Collections.emptyList();
         Collection<Property> linkProperties =
                 feature.getProperties(OpenSearchAccess.OGC_LINKS_PROPERTY_NAME);
-        if (linkProperties != null) {
+
+        Map<String, String> unknownFields = new HashMap<>();
+        // taken from OgcLink properties. these properties are known(static) from OgcLink.
+        Set<String> knownFields =
```
aaime, line 229: This is a constant, please don't declare it in methods, but at the top of the file as a proper constant (static final, all uppercase name).

src/community/oseo/oseo-rest/src/main/java/org/geoserver/opensearch/rest/AbstractOpenSearchController.java
```diff
@@ -217,7 +223,35 @@ protected OgcLinks buildOgcLinksFromFeature(Feature feature, boolean notFoundOnE
...
+                    (SimpleFeatureImpl) ((ArrayList) linkProperties).get(0);
+
+            // iterate over feature to get unknown properties from OgcLink
+            List<AttributeType> types = simpleFeature.getType().getTypes();
+            for (int i = 0; i < types.size(); i++) {
+                String unknownColumnName = types.get(i).getName().getLocalPart();
+                if (!knownFields.contains(unknownColumnName)
+                        && simpleFeature.getAttributes().get(i) != null) {
+                    unknownFields.put(
+                            unknownColumnName, simpleFeature.getAttributes().get(i).toString());
```
aaime, line 251: Hum... string? No, the data type from the database should be preserved:
- When asking for the JSON representation of OGCLink, I expect numbers to be non-quoted, as well as booleans, and dates formatted as ISO dates.
- This code is not null safe, a property can legitimately be null.

Please change your test database to include columns with types int, float, boolean and date. Make sure the types are respected as the data flows across the layers of code, and in the REST API, as well as getting reported properly in templates (JSON templates is type aware too).

src/community/oseo/oseo-rest/src/main/java/org/geoserver/opensearch/rest/OgcLink.java
```diff
@@ -20,16 +25,25 @@
 
     String href;
 
+    Map<String, String> unknownFields = new HashMap<>();
```
aaime, line 28: I believe it should be ``Map<String, Object>``, see comments above about respecting types.

aaime (changes requested):
Looks like a pretty decent start, but does not respect the data types in the database. Please fix.

src/community/oseo/oseo-core/src/test/resources/postgis.sql
ydeliorman: restored to the original state. added add column queries to SQL files provided

src/community/oseo/oseo-rest/src/main/java/org/geoserver/opensearch/rest/AbstractOpenSearchController.java
ydeliorman: done

src/community/oseo/oseo-rest/src/main/java/org/geoserver/opensearch/rest/AbstractOpenSearchController.java
ydeliorman: Changed to the object type. Dates are written in ISO date format.

src/community/oseo/oseo-rest/src/main/java/org/geoserver/opensearch/rest/OgcLink.java
ydeliorman: done

ydeliorman:
review is done.

src/community/oseo/oseo-rest/src/main/java/org/geoserver/opensearch/rest/UnknownFieldsSerializer.java
```diff
@@ -0,0 +1,38 @@
+package org.geoserver.opensearch.rest;
```
aaime, line 1: Lacks standard copyright header

aaime (changes requested):

src/community/oseo/oseo-rest/src/main/java/org/geoserver/opensearch/rest/UnknownFieldsSerializer.java
ydeliorman: Added the same copyright of [EncodingHints](https://github.com/geoserver/geoserver/blob/26273424f084bbee64e0e7e672307b8fa8deb03c/src/community/features-templating/features-templating-core/src/main/java/org/geoserver/featurestemplating/builders/EncodingHints.java#L1-L4).

aaime (approved):

PR #5456, opened 2021-12-09
[GEOS-10328] Expire completed and stale importer contexts

aaime:
[GEOS-10328](https://osgeo-org.atlassian.net/browse/GEOS-10328)

See ticket

# Checklist


dromagnoli (approved):
LGTM

src/community/importer-jdbc/src/main/java/org/geoserver/importer/jdbc/JDBCImportStore.java
```diff
@@ -15,7 +15,7 @@
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.geoserver.importer.ImportContext;
-import org.geoserver.importer.ImportStore;
+import org.geoserver.importer.ImportStore;
```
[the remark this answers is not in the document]
aaime: The Google Java formatter is actually run on community modules too (it complained, I fixed it).

data/release/global.xml
```diff
@@ -15,7 +15,20 @@
     <onlineResource>http://geoserver.org</onlineResource>
     <verbose>false</verbose>
     <verboseExceptions>false</verboseExceptions>
+    <metadata>
+      <map>
+        <entry>
+          <string>CogSettings.Key</string>
```
jodygarnett, line 21: @aaime did you intend to include COG settings as part of this change?

aaime: I did not.

aaime: However looking at the diff, I don't see a change in globa.xml?
The current global.xml does not contain that metadata entry either:

https://github.com/geoserver/geoserver/blob/main/data/release/global.xml

Not sure why you're pointing at this particular PR?

aaime: Ah, if you look at the history, the change has been reverted in the same day. Again, what brought you to look at this particular PR?

PR #5461, opened 2021-12-10
[GEOS-10329] Resource configuration for cascaded WMTS will not update TileMatrixSet - [GEOS-10331] WMTSLayerInfo native CRS as base source for reprojecting WMTS layers

taba90:
[GEOS-10329](https://osgeo-org.atlassian.net/browse/GEOS-10329)

Jira tickets:
https://osgeo-org.atlassian.net/browse/GEOS-10329
https://osgeo-org.atlassian.net/browse/GEOS-10331
This pr:
- fixes bbox retrivial at configuration time for WMTS layers
- fix tests on srs selection from layer configuration page according to https://github.com/geotools/geotools/pull/3701
- add the necessary logic to pass down to the WMTSCoverageReader is geotools the nativeCRS as the source CRS for reprojection. 
  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java
+++ b/src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java
@@ -610,7 +610,8 @@ public class CatalogBuilder {
         } else if (rinfo instanceof WMTSLayerInfo) {
             // the logic to compute the native bounds is pretty convoluted,
             // let's rebuild the layer info
-            WMTSLayerInfo rebuilt = buildWMTSLayer(rinfo.getStore(), rinfo.getNativeName());
+            WMTSLayerInfo rebuilt =
+                    buildWMTSLayer(rinfo.getStore(), rinfo.getNativeName(), rinfo.getNativeCRS());
             bounds = rebuilt.getNativeBoundingBox();
         }
 
@@ -651,7 +652,8 @@ public class CatalogBuilder {
             nativeCRS = rebuilt.getNativeCRS();
 
         } else if (rinfo instanceof WMTSLayerInfo) {
-            WMTSLayerInfo rebuilt = buildWMTSLayer(rinfo.getStore(), rinfo.getNativeName());
+            WMTSLayerInfo rebuilt =
+                    buildWMTSLayer(rinfo.getStore(), rinfo.getNativeName(), rinfo.getNativeCRS());
             return rebuilt.getNativeCRS();
         }
         return nativeCRS;
@@ -799,7 +801,7 @@ public class CatalogBuilder {
         OwsUtils.resolveCollections(layer);
 
         // get a fully initialized version we can copy from
-        WMTSLayerInfo full = buildWMTSLayer(store, layer.getNativeName());
+        WMTSLayerInfo full = buildWMTSLayer(store, layer.getNativeName(), layer.getNativeCRS());
 
         // setup the srs if missing
         if (layer.getSRS() == null) {
@@ -1419,6 +1421,12 @@ public class CatalogBuilder {
     }
 
     WMTSLayerInfo buildWMTSLayer(StoreInfo store, String layerName) throws IOException {
+        return buildWMTSLayer(store, layerName, null);
+    }
+
+    WMTSLayerInfo buildWMTSLayer(
+            StoreInfo store, String layerName, CoordinateReferenceSystem nativeCRS)
+            throws IOException {
         if (store == null || !(store instanceof WMTSStoreInfo)) {
             throw new IllegalStateException("WMTS store not set.");
         }
@@ -1442,10 +1450,9 @@ public class CatalogBuilder {
         // TODO: handle axis order here ?
         // try to get the native SRS -> we use the bounding boxes, GeoServer will publish all of the
         // supported SRS in the root, if we use getSRS() we'll get them all
-        CoordinateReferenceSystem preferred = layer.getPreferredCRS();
-        if (preferred != null) {
-            wli.setSRS(CRS.toSRS(preferred));
-            wli.setNativeCRS(preferred);
+        if (nativeCRS != null) {
+            wli.setSRS(CRS.toSRS(nativeCRS));
+            wli.setNativeCRS(nativeCRS);
         } else {
 
             for (String srs : layer.getSrs()) {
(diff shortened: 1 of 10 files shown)


src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java
```diff
@@ -1442,10 +1450,10 @@ WMTSLayerInfo buildWMTSLayer(StoreInfo store, String layerName) throws IOExcepti
         // TODO: handle axis order here ?
         // try to get the native SRS -> we use the bounding boxes, GeoServer will publish all of the
         // supported SRS in the root, if we use getSRS() we'll get them all
-        CoordinateReferenceSystem preferred = layer.getPreferredCRS();
-        if (preferred != null) {
-            wli.setSRS(CRS.toSRS(preferred));
-            wli.setNativeCRS(preferred);
+        if (nativeCRS == null) nativeCRS = layer.getPreferredCRS();
```
aaime, line 1453: Confused, since we already have a preferred CRS, why go and use native? Are the two settings redundant?
Where is the native one coming from? (if it's the resource one, it's most of the time there, so the preferred one would be ignored, making it almost useless?)

src/web/core/src/test/java/org/geoserver/web/data/resource/ResourceConfigurationPageTest.java
```diff
@@ -634,9 +622,11 @@ public void testWMTSOtherCRSUrnFormat() throws IOException {
                         "publishedinfo:tabs:panel:theList:0:content:referencingForm:nativeSRS:popup:content:table:listContainer:items:3:itemProperties:0:component:link:label");
 
         // checks that they have been properly displayed with not urn format being cut
-        assertEquals("3857", epsgComponent1.getDefaultModel().getObject());
-        assertEquals("urn:ogc:def:crs:EPSG::3857", epsgComponent2.getDefaultModel().getObject());
-        assertEquals("urn:ogc:def:crs:EPSG::900913", epsgComponent3.getDefaultModel().getObject());
+        assertEquals("urn:ogc:def:crs:EPSG::3006", epsgComponent1.getDefaultModel().getObject());
```
aaime, line 625: Confused here too. Are these checks more in line with what's in the capabilities document? (and was the old caps document fully replaced, and why?)

src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java
taba90: The nativeCRS is picked up to allow computation of native bounds being coherent to the SRS selected by the user: when computing bounds the wmts is rebuilt. Without the information of the nativeCRS the method was pick up a random srs found from the underlying layer here https://github.com/geoserver/geoserver/blob/624645ccaec2ac32e28742a3740f073ea144bf76/src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java#L1459 and retrieving bbox down using that one https://github.com/geoserver/geoserver/blob/624645ccaec2ac32e28742a3740f073ea144bf76/src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java#L1498.
The preferred CRS is a concept coming from geotools and is unclear to what it should be used for, was a leftover in that CatalogBuilder code my fault, I've removed it.

src/web/core/src/test/java/org/geoserver/web/data/resource/ResourceConfigurationPageTest.java
taba90: GT was previously using this code to extract srs name https://github.com/geotools/geotools/blob/cffe95a3a62bd12cb341b10cc498f1b81ebfa7af/modules/extension/wms/src/main/java/org/geotools/ows/wms/Layer.java#L798
of a layer. This in turn caused the ui to show up among the native srs, thus the server supported ones, also srs definition actually not present in the wmts getCapabilities document.

In the previous test we had a capaibilites document with an EPSG:3857 supported srs only, and the code was showing also as natively supported also urn:ogc:def:crs:EPSG::3857 and urn:ogc:def:crs:EPSG::900913. Once the fix happens on geotools, the test for the proper display of urn:ogc:def:crs:EPSG:: definition could not use these same capabilities since contains just an EPSG definition.

The new capabilities document instead is declaring two srs one of which is using the urn:ogc:def:crs:EPSG:: notation, so that it was possibile to keep the already present tests to check the proper display of it.

aaime:
@taba90 I've tried to run the build locally to see if I could merge this one too, but got a failure:

```
Test set: org.geoserver.web.data.resource.ResourceConfigurationPageTest
Tests run: 19, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 3.893 s <<< FAILURE! - in org.geoserver.web.data.resource.ResourceConfigurationPageTest
testWMTSOtherCRSUrnFormat(org.geoserver.web.data.resource.ResourceConfigurationPageTest)  Time elapsed: 0.008 s  <<< ERROR!
java.lang.IllegalArgumentException: Store 'Another Mock WMTS Store' already exists in workspace 'gs'
        at org.geoserver.web.data.resource.ResourceConfigurationPageTest.configureWMTSStoreInfo(ResourceConfigurationPageTest.java:881)
        at org.geoserver.web.data.resource.ResourceConfigurationPageTest.testWMTSOtherCRSUrnFormat(ResourceConfigurationPageTest.java:590)
```

Rings any bell? Repeated it twice to make sure it was not a glitch.

taba90:
> @taba90 I've tried to run the build locally to see if I could merge this one too, but got a failure:
> 
> ```
> -------------------------------------------------------------------------------
> Test set: org.geoserver.web.data.resource.ResourceConfigurationPageTest
> -------------------------------------------------------------------------------
> Tests run: 19, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 3.893 s <<< FAILURE! - in org.geoserver.web.data.resource.ResourceConfigurationPageTest
> testWMTSOtherCRSUrnFormat(org.geoserver.web.data.resource.ResourceConfigurationPageTest)  Time elapsed: 0.008 s  <<< ERROR!
> java.lang.IllegalArgumentException: Store 'Another Mock WMTS Store' already exists in workspace 'gs'
>         at org.geoserver.web.data.resource.ResourceConfigurationPageTest.configureWMTSStoreInfo(ResourceConfigurationPageTest.java:881)
>         at org.geoserver.web.data.resource.ResourceConfigurationPageTest.testWMTSOtherCRSUrnFormat(ResourceConfigurationPageTest.java:590)
> ```
> 
> Rings any bell? Repeated it twice to make sure it was not a glitch.

My apologizes Andrea, but strangely they were successfull on my localmachine during the last run this morning. Anyway I fixed the tests setup and now builds are green.

PR #5517, opened 2021-12-13
[CI] optimize builds

mprins:
As discussed on the dev ML:

- [x] only builds the docs if something changed in the doc directory or the parent pom file
- [x] skip regular builds for changes in translated property files (except French), QA build is still run

# Checklist

### DIFF
--- a/.github/workflows/linux_jdk11.yml
+++ b/.github/workflows/linux_jdk11.yml
@@ -1,6 +1,10 @@
 name: Linux JDK 11 GitHub CI
 
-on: [pull_request]
+on:
+  pull_request:
+    paths-ignore:
+      - '**/src/main/resources/GeoServerApplication_*.properties'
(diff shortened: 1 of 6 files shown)


.github/workflows/linux_jdk11.yml
```diff
@@ -1,6 +1,9 @@
 name: Linux JDK 11 GitHub CI
 
-on: [pull_request]
+on:
+  pull_request:
+    paths-ignore:
+      - '**/src/main/resources/GeoServerApplication_*.properties'
```
aaime, line 6: Might cause builds to fail when the French localization is changed, as it's used in at least one test.

mprins: ahmm I was sure I took care of that...
must have pushed to the wrong repo... let me check

mprins: indeed pushed to different repo, should be good now

PR #5539, opened 2021-12-21
[GEOS-10340] Converting static typed RSS/ATOM XML bits to freemarker template

ydeliorman:
[GEOS-10340](https://osgeo-org.atlassian.net/browse/GEOS-10340)

Static typed RSS/ATOM bits in OSEO are converted to freemarker templates for collection and product.

# Checklist


src/community/oseo/oseo-service/src/main/java/org/geoserver/opensearch/eo/response/TemplatesProcessor.java
```diff
@@ -227,4 +518,299 @@ private Geometry toGeometry(Object argument) throws TemplateModelException {
...
+        return "href=\""
+                + builder.replaceAll("&", "&amp;")
+                + "\" rel=\""
+                + rel
+                + "\" type=\""
+                + type
+                + "\"";
+    }
+
+    private String encodeLink(String rel, String builder, String type, String title) {
```
aaime, line 690: All these bits of XML (this method and the ones similar to it) should be part of the freemarker template. The whole idea of a template is to have flexibility to customize the output, having hard-coded sequences of attributes does not help.
Please move them to the freemarker template. 

Escaping of & into &amp; should probably be done by Freemarker itself, putting it in XML mode, see: 
https://freemarker.apache.org/docs/dgui_misc_autoescaping.html

Also link building should not be hard-coded, but part of the template, so that the links can be modified.
One can have a support function to build a URL out of variables if needed (see the support available in the OGC modules) but links should be built, as much as possible, directly from the Feature objects, in the FreemarkerTemplate.

The whole point of this exercise is not to build the same output using a different approach, but allow full customization of the output, it does not have to have bits built in Java, as they cannot be customized.

Btw, did you check that the previous and current output,  using the default templates, are the same?

src/community/oseo/oseo-service/src/main/java/org/geoserver/opensearch/eo/response/TemplatesProcessor.java
```diff
@@ -69,6 +116,78 @@ public String processTemplate(String collection, String templateName, Feature fe
         return sw.toString();
     }
 
+    public String processTemplate(SearchResults searchResults)
```
aaime, line 119: This method code is made convoluted by the necessity to inspect the first feature to find out the collection name (and to some extent to see if it's a product or collection case, although that could have been identified looking at the request), and thus look up the right templates for it.

As a result, instead of having a linear header/content/footer sequence, we have the header handling mixed into the content loop. A simple way to avoid this is to use the PushBackFeatureIterator, which allows to read one feature, and push it back into the iterator... you can get the first, process the header, and then push it back, and process the contents, avoiding all the header logic mixup with the content logic.

src/community/oseo/oseo-service/src/main/resources/org/geoserver/opensearch/eo/response/generic-content-start.ftl
```diff
@@ -0,0 +1,16 @@
+<#assign a = model.attributes />
+<entry>
```
aaime, line 2: While the product and collection bits might be similar in the current output, the user might just want for them to be different once customized. I'd go for a simpler setup where you have _one_ content template for collection, and one for product, with a little duplication.  So instead of "start"-"product"-"end", just have "product" with everything inside, and the same for collection.

aaime (changes requested):
As written it does not seem flexible enough (not matching the objective to make the output fully customizable), see comments inline.

Also, please:
- Make sure that the old and new generated outputs are the same (the fact you did not have to change anything in the tests bodes well for it, but did you actually try to grab the two outputs and do a diff on them)?
- Eventually, we'll need documentation be updated as well, describing the new templates and their role.

src/community/oseo/oseo-service/src/main/java/org/geoserver/opensearch/eo/response/TemplatesProcessor.java
```diff
@@ -69,6 +116,78 @@ public String processTemplate(String collection, String templateName, Feature fe
...
+            throws IOException, TemplateException {
+        StringWriter sw = new StringWriter();
+        Template footer;
```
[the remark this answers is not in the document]
ydeliorman: done

src/community/oseo/oseo-service/src/main/java/org/geoserver/opensearch/eo/response/TemplatesProcessor.java
ydeliorman: Moved header logic to the start of the function.

src/community/oseo/oseo-service/src/main/java/org/geoserver/opensearch/eo/response/TemplatesProcessor.java
ydeliorman: I have changed java code to bits as much as possible. The remaining ones are using Java Optionals, DateTime and having complex logic that cannot be transformed to freemarker.

I have used dif checkers for the below tests.
SearchTest.testAllCollection
SearchTest.testAllSentinel2Products

src/community/oseo/oseo-service/src/main/resources/org/geoserver/opensearch/eo/response/generic-content-start.ftl
ydeliorman: Removed content-start and content-end files. There is only a single template for collection and one for product.

ydeliorman:
Changes are applied.

src/community/oseo/oseo-service/src/test/java/org/geoserver/opensearch/eo/GeoJSONSearchTest.java
```diff
@@ -96,7 +96,7 @@ private void checkSentinel2Collection(JSONArray features) {
         JSONObject sp = sample.getJSONObject("properties");
         assertEquals("The Sentinel-2 mission", sp.getString("title"));
         assertEquals(featureId, sp.getString("identifier"));
-        assertEquals("2015-07-01T08:20:21Z/2016-02-26T09:20:21Z", sp.getString("date"));
+        // assertEquals("2015-07-01T08:20:21Z/2016-02-26T09:20:21Z", sp.getString("date"));
```
aaime, line 99: Why this commented-out assertions? (there are others below too, and many more in MetadataTest).
Seems to imply a regression, output not being generated anymore as it used to be.

src/community/oseo/oseo-core/src/test/java/org/geoserver/opensearch/eo/store/GSFixtureUtilitiesDelegate.java
```diff
@@ -31,7 +31,7 @@ public static Properties loadProperties(File file) {
      * ".geoserver" in the user home directory.
      */
     public static File getFixtureDirectory() {
-        return new File(System.getProperty("user.home") + File.separator + ".geoserver");
+        return new File(System.getProperty("user.home") + File.separator + "geoserver");
```
aaime, line 34: No, it must be in ".geoserver" just like all other test fixtures, the pattern is "$home/.$projectName", so for example, we have a lot in ".geotools" too.

src/community/oseo/oseo-service/src/test/java/org/geoserver/opensearch/eo/SearchTest.java
```diff
@@ -976,46 +980,48 @@ public void testGenericHTMLTemplate() throws Exception {
         summaryHasLink(
                 dom,
                 oseo
-                        + "search?parentId=SENTINEL2&uid="
+                        + "search?parentId=SENTINEL2&amp;uid="
                         + uid
-                        + "&httpAccept=application%2Fatom%2Bxml");
+                        + "&amp;httpAccept=application%2Fatom%2Bxml");
```
aaime, line 985: No, this is a regression. The HTML part is inside a CDATA block, so the URLs inside the attribute do not need the XML escaping. Please fix (this and all the other similar changes).

src/community/oseo/oseo-stac/src/test/java/org/geoserver/ogcapi/stac/ItemsTest.java
```diff
@@ -143,7 +143,8 @@ public void testSentinelItemsHTML() throws Exception {
     }
 
     private void testSentinel2SampleHTML(Elements elements) {
-        assertTextContains(elements, "[data-tid='gbounds']", "-119,174, 33,333, -117,969, 34,338.");
+        // assertTextContains(elements, "[data-tid='gbounds']", "-119,174, 33,333, -117,969,
```
aaime, line 146: Why has it been commented out?

src/community/oseo/oseo-stac/src/test/java/org/geoserver/ogcapi/stac/SearchTest.java
```diff
@@ -428,7 +428,8 @@ public void testSentinelItemsHTML() throws Exception {
         Elements s2Body =
                 doc.select(
                         "div.card-header:has(a:contains(S2A_OPER_MSI_L1C_TL_MTI__20170308T220244_A008933_T11SLT_N02.04)) ~ div.card-body");
-        assertTextContains(s2Body, "[data-tid='gbounds']", "-119,174, 33,333, -117,969, 34,338.");
+        // assertTextContains(s2Body, "[data-tid='gbounds']", "-119,174, 33,333, -117,969,
```
aaime, line 431: Same here, why commented out?

aaime (changes requested):
Seeing several regressions compared to the current output, please fix.

src/community/oseo/oseo-service/src/test/java/org/geoserver/opensearch/eo/GeoJSONSearchTest.java
ydeliorman: Sorry, I forgot to open those up. I closed that at the beginning. Reverting back to old.

src/community/oseo/oseo-core/src/test/java/org/geoserver/opensearch/eo/store/GSFixtureUtilitiesDelegate.java
ydeliorman: fixed

src/community/oseo/oseo-stac/src/test/java/org/geoserver/ogcapi/stac/ItemsTest.java
ydeliorman: fixed

src/community/oseo/oseo-stac/src/test/java/org/geoserver/ogcapi/stac/SearchTest.java
ydeliorman: fixed

src/community/oseo/oseo-service/src/test/java/org/geoserver/opensearch/eo/SearchTest.java
ydeliorman: fixed.

PR #5542, opened 2021-12-21
[GEOS-10243] OSHI may generate high CPU load on Windows

shadab-geo:
[GEOS-10243](https://osgeo-org.atlassian.net/browse/GEOS-10243)

  

Added a checkbox in the system status panel to enable or disable statistics.

# Checklist


mprins (commented):
it might be a good idea to update OSHI as part of this PR (from 5.8.1 to 5.8.6) as several Windows /WMI bugs were fixed and better support for new Windows versions

https://github.com/geoserver/geoserver/blob/85338b87d001abe70b3365de69489efab026adb7/src/pom.xml#L1357

shadab-geo:
> it might be a good idea to update OSHI as part of this PR (from 5.8.1 to 5.8.6) as several Windows /WMI bugs were fixed and better support for new Windows versions
> 
> https://github.com/geoserver/geoserver/blob/85338b87d001abe70b3365de69489efab026adb7/src/pom.xml#L1357

Updated the OSHI version

doc/en/user/source/configuration/status.rst
```diff
@@ -222,6 +222,10 @@ The system information will be available in the GeoServer status page in the ``S
 
 If the ``System status`` tab is not present, it means that the plugin was not installed correctly. The ``System status`` tab content will be refreshed automatically every second.  
 
+Enable All Statistics
+^^^^^^^^^^^^^^^^^^^^^
+when enabled, It displays the information of the listed System Status Info properties. It is disabled by default. Disabling is useful if GeoServer is generating a high CPU load due to system status computations.
```
aaime, line 227:
```suggestion
When enabled, It displays the information of the listed System Status Info properties. It is disabled by default. Disabling is useful if GeoServer is generating a high CPU load due to system status computations.
```

src/main/src/main/java/org/geoserver/system/status/OSHISystemInfoCollector.java
```diff
@@ -104,433 +104,439 @@ public OSHISystemInfoCollector() {
...
-                        mv.setValue(os.getSystemUptime());
-                        si = Collections.singletonList(mv);
-                        break;
-                    }
-                case SYSTEM_AVERAGE_LOAD_1:
-                    {
-                        MetricValue mv = new MetricValue(info);
-                        double lv = pr.getSystemLoadAverage(1)[0];
-                        if (lv > 0) {
+        if (super.getStatisticsStatus()) {
```
aaime, line 107: If I read correclty, the is basically wraps the entire method body... if so, please invert the logic and exit right away, instead of nesting code. 
There is a lot of articles on the subject on the internet, try to search for "avoid deeply nested code". For a quick reference about this particular case, see "early return" [here](https://en.wikibooks.org/wiki/Computer_Programming/Coding_Style/Minimize_nesting#:~:text=Other%20than%20refactoring%20or%20avoiding,and%20cleanup%20are%20both%20visible)).

aaime: And oh, BTW, this code is NPE prone, getStatitsticsStatus() may return null. Please also add a test covering the case of the setting being missing, which will happen on upgrade (alternative, use a Boolean for the field, but return boolean from the getter, considering nulls too, e.g., ``return Boolean.TRUE.equals(statisticsEnabled))``

aaime (changes requested):

doc/en/user/source/configuration/status.rst
shadab-geo: updated!

src/main/src/main/java/org/geoserver/system/status/OSHISystemInfoCollector.java
shadab-geo: Applied the feedback!

aaime:
Builds are red now due to WMTS changes that hit both GT and GS. You'll need to rebase this PR for the builds to fully check your changes.

shadab-geo:
> Builds are red now due to WMTS changes that hit both GT and GS. You'll need to rebase this PR onto the current GeoServer main branch for the builds to fully check your changes.

Rebased to the current main branch, still 2 of 8 checks are failing.

shadab-geo:
> Builds are red now due to WMTS changes that hit both GT and GS. You'll need to rebase this PR onto the current GeoServer main branch for the builds to fully check your changes.

Thanks, @aaime,
Rebased the PR to the current main branch. Looks like one of the checks is failing

aaime (approved):

aaime:
@shadab-geo the OSX build failure is weird, might be due to stale jars. 
Trying to address that in https://github.com/geoserver/geoserver/pull/5545

aaime:
Merged, thanks! Backport as you see fit (if any).

PR #5543, opened 2021-12-21
[GEOS-10341]: Interpolation not being propagated to the reader when a renderingTransformation is involved

dromagnoli:
[GEOS-10341](https://osgeo-org.atlassian.net/browse/GEOS-10341)

[GEOS-10341]

  

# Checklist

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/map/DirectRasterRenderer.java
+++ b/src/wms/src/main/java/org/geoserver/wms/map/DirectRasterRenderer.java
@@ -29,6 +29,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 import javax.media.jai.ImageLayout;
 import javax.media.jai.Interpolation;
 import javax.media.jai.InterpolationNearest;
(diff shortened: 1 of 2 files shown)


src/wms/src/main/java/org/geoserver/wms/map/DirectRasterRenderer.java
```diff
@@ -1194,5 +1190,21 @@ protected GridCoverage2D readCoverage(
...
+                Interpolation interpolation, GeneralParameterValue[] readParams) {
+            if (interpolation != null) {
+                List<GeneralParameterValue> paramList = new ArrayList<>();
+                if (readParams != null) {
+                    paramList.addAll(Arrays.asList(readParams));
+                }
+                final Parameter<Interpolation> readInterpolation =
+                        (Parameter<Interpolation>) ImageMosaicFormat.INTERPOLATION.createValue();
+                readInterpolation.setValue(interpolation);
+                paramList.add(readInterpolation);
```
aaime, line 1204: What if an interpolation read parameter was already in the mix? I guess it would have to be replaced instead?

src/wms/src/test/java/org/geoserver/wms/map/RenderedImageMapOutputFormatExtendedTest.java
```diff
@@ -200,6 +203,97 @@ public void testRenderingTransformationChannelsSelectionFromCoverageView() throw
...
+        Interpolation interpolation = getScalingInterpolation(destImage);
+        assertNotNull(interpolation);
+        assertEquals(interpolation, Interpolation.getInstance(Interpolation.INTERP_BILINEAR));
+        dstImageMap.dispose();
+
+        map.dispose();
+    }
+
+    private Interpolation getScalingInterpolation(RenderedImage destImage) {
+        List<RenderedOp> opImage = new ArrayList<>();
```
aaime, line 261: This is a bit bizzarre... the test works only on the first found scale, so why not just return it directly?

aaime (changes requested):

src/wms/src/main/java/org/geoserver/wms/map/DirectRasterRenderer.java
dromagnoli: Feedback has been addressed.

src/wms/src/test/java/org/geoserver/wms/map/RenderedImageMapOutputFormatExtendedTest.java
dromagnoli: Right. Fixed

PR #5552, opened 2022-01-03
Dependency upgrade: xstream 1.4.19, jettison 1.4.1

groldan:
Current XStream version has several (29) [vulnerabilities](https://mvnrepository.com/artifact/com.thoughtworks.xstream/xstream/1.4.11.1), though **they won't affect** GeoServer, since it properly configures the security subsystem, an upgrade seems to be in place anyways.

This upgrades to the latest `1.4.19` version and Jettison `1.4.1` with a patch to preserve Jettison's legacy
behaviour when encoding single-element arrays.

# Checklist


groldan:
Supersedes #4572

aaime:
AFAIK the vulnerabilities occur only when the XStream security subsystem is not configured. The GeoServer code does initialize it (although I cannot guarantee), so it should not actually be vulnerable? Communication about potential vulnerabilities should be done in a responsible way, without spreading panic, and done only once fixes are released to the public, please.

groldan:
Build's failing on `CoverageControllerWCSTest.testPostAsJSON()`.

I wonder why `CoverageControllerWCSTest.testPostAsJSON()` expects a single-element-array:

```java
        JSONArray coverages = json.getJSONObject("coverages").getJSONArray("coverage");
        assertThat(coverages.size(), is(1));
```
While `CoverageControllerTest.testGetAsJSON()` expects an object:

```java
        JSONObject coverage = ((JSONObject) json).getJSONObject("coverage");
        assertNotNull(coverage);
```

aaime:
@groldan yes spreading panic, it should be obvious to anyone that has had exposure to customers, which go nuts over the off-chance possibility anything is wrong. The Log4J2 case should serve as an emblematic example, we got grilled over a case of "run a RCE with any HTTP request" when that was not even affecting GeoServer, and ended up fixing one that required pre-existing access to the server file system in order to be triggered. Why that? Because people panic before checking if they are even affected. Even just mentioning "vulnerability" makes some people trigger... and those same people typically don't check CVEs and don't do research (the latter kind normally checks if an issue is an actual one, and how it can be triggered, before taking out the big guns).

We have multiple sections about responsible disclosure:
- http://geoserver.org/comm/
- https://osgeo-org.atlassian.net/projects/GEOS/summary/

Please read them and be more careful in the future, even for issues that are already listed (and even more so for the ones that are not, or for which you have an attack vector), the more they are discussed in public, the more the nutjobs mentioned above will start panicking.

aaime:
> Build's failing on `CoverageControllerWCSTest.testPostAsJSON()`.
> 
> I wonder why `CoverageControllerWCSTest.testPostAsJSON()` expects a single-element-array:
> 
> ```java
>         JSONArray coverages = json.getJSONObject("coverages").getJSONArray("coverage");
>         assertThat(coverages.size(), is(1));
> ```
> 
> While `CoverageControllerTest.testGetAsJSON()` expects an object:
> 
> ```java
>         JSONObject coverage = ((JSONObject) json).getJSONObject("coverage");
>         assertNotNull(coverage);
> ```

I guessing it's a simple "because XStream used to generate outputs like this"... historically XStream was used for the configuration, the REST API came some time after tha, and that's when we also got a JSON encoding... I believe it was accepted as-is, and tests kept on being written using that "untouchable" behavior.
 
In #4572 I could not find a set of settings that made the output match the previous one. As weird as it is, and as much as I don't like the inconsistency, the test spotted a backwards incompatible change. We can have such change, it just requires discussion with the PSC and an indication to the users that the upgrade might break some REST automation. I'll be +1 on breaking the JSON encoding expectations for more correct output, btw.

groldan:
Hi, for the moment I fixed the test to expect an object, under the assumption that since it's already the expected behavior, seems more like a fix than anything else, like in we at least have a consistent output? Let me know what you think.

groldan:
> The write side is more problematic, will the change make the REST API stop accepting JSON it used to read?

Haven't changed anything related to writes (POST requests), relying on the existing set of tests.
The only thing I've changed is what that test expects after the POST was processed, and reads the coverage created by the POST request. For some strange reason it used to expect a single-element array, whereas everywhere else, a GET request with a single element returns an object.

Is that what you mean? just checking we're on the same page

aaime:
I believe we are not on the same page. The change in encoding happens because XStream + Jettison behavior changed, right?
I would expect them to be symmetric, if they encode something in a different way, they would also expect to read it back in that same different way. So if a REST automatic client is posting a JSON built the old way, I would expect to see an error, XStream + Jettison would not read it anymore. Does that make sense?

A deployed system is normally a complex beast, made of different components, built by different people (and sometimes different companies, on contract). Upgrading GeoServer (maybe to get a security fix?) should not require changing the other components of the deployment. If it does, then it should be clearly advertised in the docs (and discussed by the PSC).

groldan:
> The change in encoding happens because XStream + Jettison behavior changed, right?

I'm not sure if you realize what I'm trying to do is to preserve the old behavior?

working on that `XStreamCatalogListConverter` issue which seems to be the only problematic one now

aaime:
> > The change in encoding happens because XStream + Jettison behavior changed, right?
> 
> I'm not sure if you realize what I'm trying to do is to preserve the old behavior?

I am. But tests failing means there is currently a breakage in backward compatibility.

> 
> working on that `XStreamCatalogListConverter` issue which seems to be the only problematic one now

I was not aware it was limited to that converter, no previous message mentioned it. It's good news though, that's a converter under our control, so hopefully, it can be fixed.

src/rest/src/main/java/org/geoserver/rest/converters/XStreamCatalogListConverter.java
```diff
@@ -216,7 +217,13 @@ public void encodeCollectionLink(String link, HierarchicalStreamWriter writer) {
 
         @Override
         protected XStream createXStreamInstance() {
-            return new SecureXStream(new JettisonMappedXmlDriver());
+            // needed for Jettison 1.4.1
```
aaime, line 220: Ah ha!

aaime:
All green, well done! I can't do a review right now, will try to look at this in the next few days.

groldan:
Actually I'm trying to revert that test case to its original form and get exactly the same output as before out of `XStreamCatalogListConverter`. Added test cases to `CoverageControllerTest` for zero and single element lists, because those never returned unwrapped single-element results.

Problem is I either get
```
{"coverages":  {  "coverage": { "name": "tazdem"} } }
```
or 
```
{"coverages": [ {"coverage": {"name": "tazdem"} } ] }
```

But not the expected

```
{"coverages": {"coverage": [  { "name": "tazdem" } ] } }
```

groldan:
ok, managed to preserve `XStreamCatalogListConverter`'s behavior as of 75925d37f87102f19aedd43d8fbfef63b0e4cfbc (replaces previous commit that changed the failing test instead, and adds a specific test case for the single-element case)

src/restconfig/src/test/java/org/geoserver/rest/resources/ResourceControllerTest.java
```diff
@@ -310,6 +311,7 @@ public void testDirectoryXML() throws Exception {
     }
 
     @Test
+    @Ignore
```
mprins, line 314: Can you add a comment as to why this test should be ignored? 
```suggestion
    @Ignore("why is this ignored....")
```

mprins (commented):
I see `boolean useSerializeAsArray = false;` in several classes and I'm wondering if it would be a good idea to move that into a public final in eg. `XStreamPersister.java` (not sure how that would work out dependency-wise

src/restconfig/src/test/java/org/geoserver/rest/resources/ResourceControllerTest.java
groldan: I think it wasn't intentional, met me check

groldan: No, I missed that one. Was added by @aaime at 4bd66b2e222e464e03458892ca7f9737def3be00
I'll see what's going on

aaime: Yep it was part of the work in progress, that branch was not meant to be committed as-is

src/restconfig/src/test/java/org/geoserver/rest/catalog/DataStoreControllerTest.java
```diff
@@ -72,6 +72,7 @@ public void testGetAsJSON() throws Exception {
     @Test
     public void testGetAllAsJSON() throws Exception {
         JSON json = getAsJSON(ROOT_PATH + "/workspaces/sf/datastores.json");
+        print(json);
```
aaime, line 75: I'd get rid of all the extra "print" in this PR.

aaime (commented):
Looks pretty good in general! Agree with the two points raised by @mprins (ignored test, and sharing duplicated code if possible).

aaime:
Looks like there are legit test failures (at one point I believe the build was green though? maybe I'm confusing it with something else...)

groldan:
> Looks like there are legit test failures (at one point I believe the build was green though? maybe I'm confusing it with something else...)

It was green, but haven't seen that one that was `@Ignore`d. I'll get to it asap

aaime:
Needs a bump to 1.4.19: https://github.com/geoserver/geoserver/pull/5636

groldan:
just a heads up that I'm resuming work on this

aaime:
How are things going @groldan ?

groldan:
> How are things going @groldan ?

dancing with priorities @aaime, but I think I can finally get to this before the end of the week

groldan:
Closing as superseded by #5730 with a single commit and corrected branch name.

PR #5562, opened 2022-01-06
[GEOS-10350] Upgrade NetCDF to 4.6.19

sikeoka:
[GEOS-10350](https://osgeo-org.atlassian.net/browse/GEOS-10350)

Upgrade NetCDF to 4.6.19 to address security vulnerabilities in third-party libraries in Dependency Check scans.

Related to https://github.com/geotools/geotools/pull/3728

gs-netcdf extension already inherits its NETCDF dependencies from gt-netcdf and doesn't define its own NETCDF version so it didn't need any changes.
gs-netcdf-out extension had all of its NETCDF dependencies removed so that they are now inherited from gs-netcdf and it no longer needs the netcdf.version property.
gs-netcdf-ghrsst community module had all of its NETCDF dependencies removed so that they are now inherited from gs-netcdf-out but it still needs the netcdf.version property to put in a properties file.  gs-netcdf-ghrsst unit tests failed even before this upgrade and I am not attempting to fix that.

  

# Checklist

### DIFF
--- a/src/community/netcdf-ghrsst/pom.xml
+++ b/src/community/netcdf-ghrsst/pom.xml
@@ -17,7 +17,7 @@
   <name>NetCDF GHRSST encoder</name>
 
   <properties>
-    <netcdf.version>4.6.15</netcdf.version>
+    <netcdf.version>4.6.19</netcdf.version>
   </properties>
 
   <dependencies>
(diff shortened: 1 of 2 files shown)


dromagnoli (approved):
I have no memories about those slf4j exclusions in the poms.  Maybe some clashes in the eclipse project settings when using older NetCDF versions at the early stage of these modules.

Anyway, if you are confident with the changes I'm +1 on this.

PR #5564, opened 2022-01-07
[GEOS-10352] GetMap clip vendor does not always return fully transparent images outside of clip

aaime:
[GEOS-10352](https://osgeo-org.atlassian.net/browse/GEOS-10352)

Set ticket for details.
Depends on https://github.com/geotools/geotools/pull/3731

# Checklist


src/wms/src/main/java/org/geoserver/wms/map/DirectRasterRenderer.java
```diff
@@ -398,6 +402,7 @@ public RenderedImage render() throws FactoryException {
                 if (transparent && !image.getColorModel().hasAlpha()) {
                     image = addAlphaChannel(image);
                     worker.setImage(image);
+                    transparencyType = image.getColorModel().getTransferType();
```
dromagnoli, line 405: I'm wondering if that migth be an auto-complete issue.
Is actually getTransparencyType or is getTransferType correct, instead?

dromagnoli (commented):
See my comment about a doubt on a getter with probably a wrong name (maybe just a TYPO due to a autocomplete suggestion).

src/wms/src/main/java/org/geoserver/wms/map/DirectRasterRenderer.java
aaime: Auto-complete issue, well spotted.

dromagnoli: Cool. beside of that, +1 to merge

PR #5584, opened 2022-01-12
Revert "[GEOS-10350] Upgrade NetCDF to 4.6.19"

dromagnoli:
Reverts geoserver/geoserver#5562

The update of the NetCDF version makes the netcdf-out grib test fails on builds and jenkins PR.
Details incoming:

### DIFF
--- a/src/community/netcdf-ghrsst/pom.xml
+++ b/src/community/netcdf-ghrsst/pom.xml
@@ -17,7 +17,7 @@
   <name>NetCDF GHRSST encoder</name>
 
   <properties>
-    <netcdf.version>4.6.19</netcdf.version>
+    <netcdf.version>4.6.15</netcdf.version>
   </properties>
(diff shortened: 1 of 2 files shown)


sikeoka:
I tested this and confirmed that upgrading the NETCDF dependencies (specifically the grib dependency) would unfortunately break existing GRIB coverages that are affected by the GRIB table changes.  The GeoServer/GeoTools JIRA tickets can be closed since this is a potentially serious compatibility issue for what was supposed to be just a minor dependency upgrade to get rid of a Dependency Check issue on one of its transitive dependencies.
`jsoup-1.11.2.jar (pkg:maven/org.jsoup/jsoup@1.11.2, cpe:2.3:a:jsoup:jsoup:1.11.2:*:*:*:*:*:*:*) : CVE-2021-37714`

PR #5588, opened 2022-01-14
[GEOS-10351] [GSIP-206] Promote GeoPackage WFS and WMS output formats to an extension

jodygarnett:
[GEOS-10351](https://osgeo-org.atlassian.net/browse/GEOS-10351)

Create `geopkg-output-extension` using WFS and WMS output formats:

* This pull request benefits from https://github.com/geotools/geotools/pull/3740 to ensure features are saved in XY order
* GeoPkgGetMapOutputFormat has been improved to ensure that tiles are generated correctly regardless of request axis order (the same output is now generated for WMS 1.3 and WMS 1.0)
* `mbtiles` modle depends on `geopkg-output` for `AbstractTilesGetMapOutputFormat`
* `community/geopkg` module depends on `geopkg-output` for the classes `GeoPkg` and `GeoPkgDataStoreFactoryInitilizer`

Please see https://github.com/geoserver/geoserver/wiki/GSIP-206 for details.

This is presented as multiple-commits, please use squash-and-merge to reduce to a single commit once reviewed.

# Checklist

- [x] There is an issue https://osgeo-org.atlassian.net/browse/GEOS-10351


doc/en/user/source/extensions/geopkg-output/install.rst
```diff
@@ -0,0 +1,44 @@
...
+            <geopackage/>
+            <geopkg/>
+            <gpkg/>
+        </ResultFormat>
+    </GetFeature>
+
+
+.. note::
+
+    You can also verify installation by looking for ``GeoPKG Output Extension`` on the server's `Module Status Page <http://localhost:8080/geoserver/web/wicket/bookmarkable/org.geoserver.web.admin.StatusPage?5&filter=false>`__
```
jodygarnett, line 44: I am not sure if this link is always valid, due to URL randomization logic in Wicket. We may need to just say:

```
:guilabel:`ServerStatus` page, :guilabel:`Module Status` tab.
```

doc/en/user/source/extensions/geopkg-output/usage.rst
```diff
@@ -0,0 +1,96 @@
...
+WFS
+^^^
+
+
+Add ``&outputFormat=geopkg`` to your request. The result will be a GeoPackage (MIME type `application/geopackage+sqlite3`) containing the requested features.
+
+.. code-block::  
+
+    curl "http://localhost:8080/geoserver/wfs?service=wfs&version=2.0.0&request=GetFeature&typeNames=ws:layername&outputFormat=geopkg" \
+    -o wfs.gpkg
```
jodygarnett, line 17: Thanks for include a curl example

doc/en/user/source/extensions/geopkg-output/usage.rst
```diff
@@ -0,0 +1,96 @@
...
+     
+       default: find based on request SRS
+   * - flipy
+     - Do NOT set.
+
+       default: TRUE (required for GeoPackage - `The tile coordinate (0,0) always refers to the tile in the upper left corner of the tile matrix...`)
+
+.. note::
+
+    You can use `geopkg`, `geopackage`, or `gpkg` as the output format in the request.  Use `1.1.0`, or `1.3` as ``version=`` to specify which WMS version to use.
```
jodygarnett, line 94: Expect the above is `1.3.0`

src/community/geopkg/src/main/java/applicationContext.xml
```diff
@@ -7,19 +7,19 @@
...
       <constructor-arg ref="gwcFacade" />
     </bean>
-    
-    <bean id="geopkgGetFeatureOutputFormat" class="org.geoserver.geopkg.GeoPackageGetFeatureOutputFormat">
-      <constructor-arg ref="geoServer" />
-    </bean>
+
+<!--    <bean id="geopkgGetFeatureOutputFormat" class="org.geoserver.geopkg.GeoPackageGetFeatureOutputFormat">-->
+<!--      <constructor-arg ref="geoServer" />-->
+<!--    </bean>-->
```
jodygarnett, line 18: Please remove commented out code :)

src/community/geopkg/src/main/java/org/geoserver/geopkg/GeoPackageGetMapOutputFormatWPS.java
```diff
@@ -1,190 +1,34 @@
...
 import org.geoserver.platform.ServiceException;
-import org.geoserver.tiles.AbstractTilesGetMapOutputFormat;
-import org.geoserver.wms.GetMapRequest;
-import org.geoserver.wms.MapLayerInfo;
-import org.geoserver.wms.RasterCleaner;
-import org.geoserver.wms.WMS;
-import org.geoserver.wms.WMSMapContent;
-import org.geoserver.wms.WebMap;
-import org.geoserver.wms.WebMapService;
+import org.geoserver.wms.*;
```
jodygarnett, line 15: I thought a PMD check would catch this, but perhaps not run for community modules?

src/community/geopkg/src/test/java/org/geoserver/geopkg/wps/GeoHashCalculatorTest.java
```diff
@@ -18,6 +19,11 @@
     WKTReader wkt = new WKTReader();
     GeoHashCalculator calculator = GeoHashCalculator.DEFAULT;
 
+    @BeforeClass
+    public static void setup() {
+        System.setProperty("org.geotools.referencing.forceXY", "true");
```
jodygarnett, line 24: I would love to find out what code was changing forceXY to false? 

If this is required for the test case use CRS.reset method in both setup and teardown - similar to https://github.com/geotools/geotools/blob/213393c288d52f7e302df1dba6cc3a23f77fe356/modules/plugin/imagepyramid/src/test/java/org/geotools/gce/imagepyramid/AbstractPyramidTest.java#L34-L44

src/community/geopkg/src/test/java/org/geoserver/geopkg/wps/GeoHashCollectionTest.java
```diff
@@ -9,13 +9,19 @@
...
 import org.opengis.feature.simple.SimpleFeature;
 import org.opengis.feature.simple.SimpleFeatureType;
 import org.opengis.feature.type.AttributeDescriptor;
 
 public class GeoHashCollectionTest extends DataTestCase {
 
+    @BeforeClass
+    public static void setup() {
+        System.setProperty("org.geotools.referencing.forceXY", "true");
+    }
```
jodygarnett, line 23: As above

src/extension/geopkg-output/pom.xml
```diff
@@ -0,0 +1,153 @@
...
+    <!--      <groupId>org.geotools</groupId>-->
+    <!--      <artifactId>gt-render</artifactId>-->
+    <!--      <version>${gt.version}</version>-->
+    <!--    </dependency>-->
+
+    <!--    <dependency>-->
+    <!--      <groupId>ch.hsr</groupId>-->
+    <!--      <artifactId>geohash</artifactId>-->
+    <!--      <version>1.4.0</version>-->
+    <!--    </dependency>-->
```
jodygarnett, line 73: The commented out lines can be removed

jodygarnett (commented):
Some feedback below.

src/extension/geopkg-output/src/main/java/applicationContext.xml
```diff
@@ -0,0 +1,33 @@
...
+    <bean id="geopkgGetFeatureOutputFormat" class="org.geoserver.wfs.geopackage.GeoPackageGetFeatureOutputFormat">
+      <constructor-arg ref="geoServer" />
+    </bean>
+     
+    <bean id="geopkgDataStoreFactoryInitializer" class="org.geoserver.geopackage.GeoPkgDataStoreFactoryInitializer">
+        <property name="resourceLoader" ref="resourceLoader"/>
+    </bean>
+
+    <!-- bean so this will appear on the Modules list in the GS web interface-->
+    <bean id="GeoPKGOutputExtension"
```
jodygarnett, line 25: Thanks for including in module status list

src/community/geopkg/src/main/java/org/geoserver/geopkg/GeoPackageGetMapOutputFormatWPS.java
aaime: QA does not run in community modules, indeed. On move to extension you'll face the full force of the QA checks :-D

src/community/geopkg/src/test/java/org/geoserver/geopkg/wps/GeoHashCalculatorTest.java
aaime: Nothing changes it to false. False is the default.

src/community/geopkg/src/main/java/applicationContext.xml
```diff
@@ -7,19 +7,15 @@
 <!DOCTYPE beans PUBLIC "-//SPRING//DTD BEAN//EN" "http://www.springframework.org/dtd/spring-beans.dtd">
 <beans>
 
-    <bean id="geopkgGetMapOutputFormat" class="org.geoserver.geopkg.GeoPackageGetMapOutputFormat">
+    <bean id="geopkgGetMapOutputFormatWPS" class="org.geoserver.geopkg.wps.GeoPackageGetMapOutputFormatWPS">
```
aaime, line 10: So by pluggin this one in, GeoServer will find two of them... does that cause any misbehavior?
Why not just depend on the geopkg-output module?

src/community/geopkg/src/test/java/org/geoserver/geopkg/wps/GeoHashCalculatorTest.java
aaime: But I see what you mean, when multiple tests run in the same JVM, there is a chance a reset in setup might be needed.
It's not actually happening, since integration tests force it to true anyways, the setup was needed anyways:
-  to run the test stand alone
- if surefire happened to run the test as the first one in the build

src/extension/geopkg-output/src/main/java/applicationContext.xml
aaime: This self-review style is weird... or are you committing stuff on behalf of someone else? :D

src/extension/geopkg-output/src/main/java/org/geoserver/tiles/AbstractTilesGetMapOutputFormat.java
```diff
@@ -459,6 +493,19 @@ protected GridSubset findBestGridSubset(GetMapRequest req) {
         return gridSubsets.iterator().next();
     }
 
+    // finds the appropriate gridset zoom levels for the request
```
aaime, line 496: Why are all these (across the PR) common comments, rather than javadocs?
It's valid but weird at the same time.

src/extension/geopkg-output/src/main/java/applicationContext.xml
jodygarnett: @davidblasby has been doing the work, with some assistance from me as needed.  I am catching up near the end and submitted the PR.

src/community/geopkg/src/main/java/applicationContext.xml
jodygarnett: Different bean, this one makes the WPS output format.

src/extension/geopkg-output/src/main/java/org/geoserver/tiles/AbstractTilesGetMapOutputFormat.java
jodygarnett: David is gradually adjusting coding style.  I am focused on the public methods (not overrides) when I reviewed.

src/community/geopkg/src/test/java/org/geoserver/geopkg/wps/GeoHashCollectionTest.java
```diff
@@ -9,13 +9,28 @@
 import org.geotools.data.DataTestCase;
```
aaime, line 9: New file without copyright header. Checkstyle does not run on community modules, but it would still be nice to have headers in place.

src/community/geopkg/src/main/java/org/geoserver/geopkg/wps/GeoPackageGetMapOutputFormatWPS.java
```diff
@@ -1,33 +1,20 @@
-/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
- * (c) 2001 - 2013 OpenPlans
- * This code is licensed under the GPL 2.0 license, available at the root
- * application directory.
- */
-package org.geoserver.geopkg;
+package org.geoserver.geopkg.wps;
```
aaime, line 1: Copyright header was removed here?

src/community/geopkg/pom.xml
```diff
@@ -42,16 +42,16 @@
       <artifactId>gs-wfs</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.geoserver.extension</groupId>
+      <artifactId>gs-geopkg-output</artifactId>
```
aaime, line 47: This dependency is not reflected in the ``src/community/release/ext-geopkg.xml`` assembly file. Please update and make sure the community module package installs and runs correctly.

doc/en/user/source/extensions/geopkg-output/index.rst
```diff
@@ -0,0 +1,12 @@
+.. _geopkgoutput:
+
+GeoPackage Output
+=================
+
+This extension supports GeoPKG as an output format for both WFS (vectors) and WMS (tiles).
```
aaime, line 6: The community module documentation does not seem to have been updated, should only report the WPS output format after this change, no?
https://docs.geoserver.org/latest/en/user/community/geopkg/index.html
and
https://docs.geoserver.org/latest/en/user/community/geopkg/output.html

aaime (changes requested):
In addition to other comments, I'll note that the gs

src/community/geopkg/src/main/java/applicationContext.xml
jodygarnett: Perhaps I did not understand, is WPS output format and WFS Output format implementing the same interface?

src/community/geopkg/pom.xml
jodygarnett: I checked that, the ext-geopkg grabs this jar correctly because it uses a wildcard.

src/community/geopkg/src/main/java/applicationContext.xml
aaime: This bean that you are re-declaring in the community module is a WMS output format. The GeoPackage process generates tiles entries calling onto the WMS output format. The way it's set up, you have two classes for the WMS output format, and two beans that declare them. They are related by inheritance, the WPS one extends the WMS one.

src/extension/geopkg-output/src/main/java/org/geoserver/tiles/AbstractTilesGetMapOutputFormat.java
aaime: Right, but this is still supported code... looks like it was written by someone whose first language is not Java...

aaime:
All green but docs, which are also failing here for the :website: thing. Duh... also the Chinese docs are failing for the same reason... hmm....

jodygarnett:
Okay if we rebase we should get the docs fixed.

PR #5591, opened 2022-01-17
[GEOS-10332]-GrowableInternationalStringConverter can cause random XStream conversion failures

yashwanthanands:
[GEOS-10332](https://osgeo-org.atlassian.net/browse/GEOS-10332)

Growable International String and Simple International String conversion failure in XStreamPersister has been fixed.

  

# Checklist


src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
```diff
@@ -2683,20 +2681,44 @@ public boolean canConvert(Class aClass) {
...
+                        writer.endNode();
+                    } else if (size > 1) {
+                        writer.startNode(DEFAULT_LOCALE);
+                        writer.setValue(internationalString.toString(l));
+                        writer.endNode();
+                    }
+                }
+            } else {
+                defaultValue = source.toString();
+                Set<Locale> locales = Stream.of(Locale.getDefault()).collect(Collectors.toSet());
```
taba90, line 2707: Seems that you only need to use the Default Locale so there is no need to add it before to a Set and then iterate.

src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
```diff
@@ -2683,20 +2681,44 @@ public boolean canConvert(Class aClass) {
...
+                        if (elementName.contains(" "))
+                            elementName = elementName.replaceAll(" ", "__");
+                        writer.startNode(elementName);
+                        writer.setValue(defaultValue);
+                        writer.endNode();
+                    } else if (size > 1) {
+                        writer.startNode(DEFAULT_LOCALE);
+                        writer.setValue(defaultValue);
+                        writer.endNode();
+                    }
```
taba90, line 2721: this code block is the same used for the GrowableInternationalString case. It would be better to put it into a separate method to avoid duplication

src/main/src/test/java/org/geoserver/config/util/XStreamPersisterTest.java
```diff
@@ -1580,11 +1584,37 @@ public synchronized String toString(Locale locale) {
...
+                        .getXStream()
+                        .getConverterLookup()
+                        .lookupConverterForType(SimpleInternationalString.class);
+        assertTrue(candidate instanceof XStreamPersister.InternationalStringConverter);
+
+        persister.save(toTest, out);
+        final ByteArrayInputStream in = new ByteArrayInputStream(out.toByteArray());
+        final ObjectInputStream objectIn = new ObjectInputStream(in);
+        final Object object = objectIn.readObject();
+        objectIn.close();
```
taba90, line 1614: I would covert the byteArrayOutputStream to a String after the save method is called to assert that it contains the message variable.

taba90 (changes requested):

PR #5593, opened 2022-01-18
Fix sphinx build warnings indicating where extlinks can be used 

jodygarnett:
Most of the warnings were download instructions for extensions, so I just went through all of them to make it consistent.

# Checklist


aaime:
Noice, the docs build is back to green. However, the CN doc build is still red:
image

Not sure what to do about the latter... would it be an easy fix? Or should we turn it off for the time being, and ask for help on the contributors to the CN translation?

jodygarnett:
Noticed that also, it the zhcn docs should now build also

jodygarnett:
And we are go: 
image

aaime:
Yessir... I was trying to figure out where this magic "wiki" or "website" things were coming from, at first thought they were well-known bits of RST/Sphinx, but turns out they are [variables declared in configuration](https://github.com/geoserver/geoserver/blob/main/doc/en/user/source/conf.py#L107) and there are more of them.

Maybe I'm not the only one that had no idea about them...

doc/en/developer/source/maven-guide/index.rst
```diff
@@ -198,31 +198,4 @@ This can also be used when running the local jetty application server:
 
   mvn jetty:run -DconfigId=release -DconfigDirectory=../../../data/release
 
-You may also use an absolute path, if you have a custom data directory you would like to use.
```
jodygarnett, line 201: Looks like I removed one too many lines

aaime:
Seeing the same build failures in stable and maintenance, a backport is needed.

PR #5597, opened 2022-01-19
[GEOS-10101] Recalculating a featuretype via geoserver rest, fails if the SRS handling is set to Keep Native

yashwanthanands:
[GEOS-10101](https://osgeo-org.atlassian.net/browse/GEOS-10101)

  

# Checklist

### DIFF
--- a/doc/en/user/source/rest/api/featuretypes.rst
+++ b/doc/en/user/source/rest/api/featuretypes.rst
@@ -155,6 +155,23 @@ The ``recalculate`` parameter specifies whether to recalculate any bounding boxe
 * ``recalculate=nativebbox``: Recalculate the native bounding box, but do not recalculate the lat/long bounding box.
 * ``recalculate=nativebbox,latlonbbox``: Recalculate both the native bounding box and the lat/long bounding box.
 
+``Projection Policy``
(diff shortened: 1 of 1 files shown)


doc/en/user/source/rest/api/featuretypes.rst
```diff
@@ -155,6 +155,20 @@ The ``recalculate`` parameter specifies whether to recalculate any bounding boxe
 * ``recalculate=nativebbox``: Recalculate the native bounding box, but do not recalculate the lat/long bounding box.
 * ``recalculate=nativebbox,latlonbbox``: Recalculate both the native bounding box and the lat/long bounding box.
 
+``Projection Policy``
+^^^^^^^^^^^^^^^^^^^^^
+.. list-table::
```
taba90, line 160: Please add some context to this table so that the user can cleary understand understand the meaning of it
```suggestion
   When specifying the Projection Policy in a FeatureType defined in the request body, the internal name should be used instead of the one available on the UI. The following table shows the correspondence between display and internal names:
.. list-table::
```

taba90 (changes requested):

yashwanthanands:
@taba90 

taba90 (approved):

PR #5598, opened 2022-01-20
[GEOS-10362] Remove selected and remove role associations removes user and the roles from roles.xml

ydeliorman:
[GEOS-10362](https://osgeo-org.atlassian.net/browse/GEOS-10362)

Remove selected and remove role associations removes user and user's roles from roles.xml.

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/impl/AbstractRoleStore.java
+++ b/src/main/src/main/java/org/geoserver/security/impl/AbstractRoleStore.java
@@ -254,6 +254,9 @@ public abstract class AbstractRoleStore implements GeoServerRoleStore {
             roles.remove(role);
             setModified(true);
         }
+        if (helper.user_roleMap.get(username) != null && roles != null && roles.size() == 0) {
+            helper.user_roleMap.remove(username);
+        }
     }
 
     /* (non-Javadoc)
--- a/doc/en/user/source/security/webadmin/ugr.rst
+++ b/doc/en/user/source/security/webadmin/ugr.rst
@@ -238,6 +238,20 @@ The Groups tab provides configuration options for groups in this user/group serv
 
 .. _security_webadmin_groups:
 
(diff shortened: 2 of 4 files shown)


src/main/src/main/java/org/geoserver/security/impl/AbstractRoleStore.java
```diff
@@ -254,6 +254,9 @@ public void disAssociateRoleFromUser(GeoServerRole role, String username) throws
             roles.remove(role);
             setModified(true);
         }
+        if (helper.user_roleMap.get(username) != null && roles != null && roles.size() == 0) {
```
aaime, line 257: This change makes sure there is no rule-user association in case there are no roles left for that user. Good. But is the test checking that this is actually happening? Does not seem so.

Also, the ticket in Jira title is "Username remains in roles.xml after user removal operation".
This is clearer. It does not seem like the test is doing any check against the actual contents of roles.xml.

Maybe I'm just misunderstanding, please clarify.

ydeliorman: Removing user and roles from UI enters [here](https://github.com/geoserver/geoserver/blob/1631857aae6d1557a93591d995f7d536a89a43d4/src/web/security/core/src/main/java/org/geoserver/security/web/user/SelectionUserRemovalLink.java#L126-L129). Then it goes down to AbstractRoleStore. "roles.xml" is fetching the user data from AbstractRoleStore helper.user_roleMap variable. The problem was not updating the user_roleMap. Since the logic is happening here, I wrote the tests to mimic the for loop behavior. 

But you were right, I should be checking the XML file too. I created a new test and deleted the old one. In the end, it looks for an empty userlist tag in the XML.

aaime:
Merged. Backport as you see fit.

PR #5599, opened 2022-01-20
[GEOS-10332]-GrowableInternationalStringConverter can cause random XStream conversion failures

yashwanthanands:
[GEOS-10332](https://osgeo-org.atlassian.net/browse/GEOS-10332)

Growable International String and Simple International String conversion failure in XStreamPersister has been fixed.

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
+++ b/src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
@@ -521,31 +521,27 @@ public class XStreamPersister {
         xs.registerLocalConverter(GeneralEnvelope.class, "crs", new SRSConverter());
 
         xs.registerLocalConverter(
-                impl(ResourceInfo.class),
-                "internationalTitle",
-                new GrowableInternationalStringConverter());
+                impl(ResourceInfo.class), "internationalTitle", new InternationalStringConverter());
         xs.registerLocalConverter(
                 impl(ResourceInfo.class),
                 "internationalAbstract",
-                new GrowableInternationalStringConverter());
+                new InternationalStringConverter());
 
         xs.registerLocalConverter(
-                impl(ServiceInfo.class),
-                "internationalTitle",
-                new GrowableInternationalStringConverter());
+                impl(ServiceInfo.class), "internationalTitle", new InternationalStringConverter());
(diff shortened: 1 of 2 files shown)


src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
```diff
@@ -2683,21 +2679,32 @@ public boolean canConvert(Class aClass) {
...
+                        writer.startNode(DEFAULT_LOCALE);
+                        writer.setValue(internationalString.toString(l));
+                        writer.endNode();
+                    }
                 }
+            } else {
+                defaultValue = source.toString();
+                writer.startNode(DEFAULT_LOCALE);
+                writer.setValue(defaultValue);
+                writer.endNode();
```
taba90, line 2707: I'm a bit confused here: in the closed pr you were using the Locale.getDefault() to start the node. But now you swithced to the DEFAULT_LOCALE constant. Why this change? Note that the DEFAULT_LOCALE would result in an xml element name <default>. It seems to me that the Locale.getDefault() is the preferred option here.

src/main/src/test/java/org/geoserver/config/util/XStreamPersisterTest.java
```diff
@@ -1585,6 +1588,30 @@ public synchronized String toString(Locale locale) {
...
+        objectOut.writeObject(toTest);
+        objectOut.close();
+        XStreamPersister persister = new XStreamPersisterFactory().createXMLPersister();
+        Converter candidate =
+                persister
+                        .getXStream()
+                        .getConverterLookup()
+                        .lookupConverterForType(SimpleInternationalString.class);
+        assertTrue(candidate instanceof XStreamPersister.InternationalStringConverter);
+        assertEquals(message, toTest.toString());
```
taba90, line 1606: Doesn't have that match sense to test the equality between the String and the InternationalString value here.
What I would test instead is the byte output stream content that contains the expected string value after the persister.save method has been executed.

taba90 (changes requested):

yashwanthanands:
@taba90

taba90 (approved):

taba90:
seems good to me. Merging

PR #5600, opened 2022-01-20
Fix AttributeTypeInfoImpl.equals() and hashCode() after new source and nillable properties

groldan:
Fix AttributeTypeInfoImpl.equals() and hashCode() after new source property

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/impl/AttributeTypeInfoImpl.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/AttributeTypeInfoImpl.java
@@ -154,27 +154,24 @@ public class AttributeTypeInfoImpl implements AttributeTypeInfo {
 
     @Override
     public boolean equals(Object o) {
-        if (this == o) return true;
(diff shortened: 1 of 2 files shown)


src/main/src/main/java/org/geoserver/catalog/impl/AttributeTypeInfoImpl.java
```diff
@@ -154,27 +154,24 @@ public void setSource(String source) {
...
     public int hashCode() {
         // feature type excluded, or it's gonna go in infinite recursion
+        String source = this.source == null ? this.name : this.source;
         return Objects.hash(
-                id, name, minOccurs, maxOccurs, nillable, attribute, metadata, binding, length,
+                id,
+                name,
+                minOccurs,
+                maxOccurs,
+                isNillable(),
```
aaime, line 170: Good catch.

aaime (approved):
Looks good, thanks for the fix, merging.

aaime:
Tagged it with the initial commit, so that we can find both if there will be a backport.

PR #5602, opened 2022-01-21
[GEOS-10357]-WFS 2.0 expects namespace request parameter instead of namespaces

yashwanthanands:
[GEOS-10357](https://osgeo-org.atlassian.net/browse/GEOS-10357)

WFS 2.0 GetFeature Request allows namespace and namespaces.But it allows only namespaces for STRICT Compliance enabled.

  

# Checklist

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/kvp/BaseFeatureKvpRequestReader.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/kvp/BaseFeatureKvpRequestReader.java
@@ -88,6 +88,15 @@ public abstract class BaseFeatureKvpRequestReader extends WFSKvpRequestReader {
                 new String[] {"featureId", "resourceId", "filter", "bbox", "cql_filter"},
                 eObject);
 
+        // check strict cite compliance
+        if (getWFS().isCiteCompliant()) {
+            if (kvp.containsKey("namespace")) {
+                throw new WFSException(
+                        eObject,
+                        "Strict Cite Compliance is enabled.Request contains [namespace].Only [namespaces] is supported for WFS 2.0");
+            }
+        }
+
         // did the user supply alternate namespace prefixes?
         NamespaceSupport namespaces = null;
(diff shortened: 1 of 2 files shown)


src/wfs/src/main/java/org/geoserver/wfs/kvp/BaseFeatureKvpRequestReader.java
```diff
@@ -126,6 +135,7 @@ public Object read(Object request, Map<String, Object> kvp, Map<String, Object>
                     @SuppressWarnings("unchecked")
                     List<QName> qNames = (List<QName>) obj;
                     for (int i = 0; i < qNames.size(); i++) {
+                        namespaces = null;
```
taba90, line 138: Why this is set to null here? Not sure this is the correct way to go.

taba90 (changes requested):
Builds are failing on gs-wfs. A test for the exception being thrown when cite compliant mode is activate is needed.

yashwanthanands:
@taba90 

src/wfs/src/main/java/org/geoserver/wfs/kvp/BaseFeatureKvpRequestReader.java
```diff
@@ -333,11 +342,12 @@ QName checkTypeName(QName qName, NamespaceSupport namespaces, EObject request) {
                     // namespace
                     namespaceURI = uri;
                 }
-            } else if (namespaces.getURI(prefix) != null) {
+            } else if ((namespaces.getURI(prefix) != null) && (qName.getNamespaceURI() != null)) {
```
nmco, line 345: This change is unexpected for me, looks to me that GeoServer is trying to find in the provided namespaces a URI that matches the feature type prefix that was used in the request. For example:
```
typeNames=topp:states
namespaces=xmlns(topp,http://www.openplans.org/topp),xmlns(ns2,http://someserver.example.com/ns2)
```

... in this case he should match the ``topp`` prefix with the URI ``xmlns(topp,http://www.openplans.org/topp)``.

PR #5607, opened 2022-01-22
[GEOS-10363] Switch from itextpdf to openpdf for PDF map rendering

bradh:
[GEOS-10363](https://osgeo-org.atlassian.net/browse/GEOS-10363)

Switches the WMS dependency to OpenPDF, which is basically a drop-in replacement.

There is one code change that was needed, to dispose of a graphics entity that we were not previously handling. There is also a switch to try-with-resources on the document, required to pass PMD checks.

This PR also includes some minor documentation updates,.
  
# Checklist

- [N/A] New unit tests have been added covering the changes.
- [N/A] [Documentation](https://github.com/geoserver/geoserver/tree/main/doc/en/user/source) has been updated (if change is visible to end users).
- [N/A] The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers).

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -1026,9 +1026,14 @@
         <version>1.2.17.norce</version>
       </dependency>
       <dependency>
-        <groupId>com.lowagie</groupId>
-        <artifactId>itext</artifactId>
-        <version>2.1.5</version>
+        <groupId>com.github.librepdf</groupId>
+        <artifactId>openpdf</artifactId>
+        <version>1.3.26</version>
+      </dependency>
+      <dependency>
+        <groupId>com.twelvemonkeys.imageio</groupId>
+        <artifactId>imageio-tiff</artifactId>
+        <version>3.4.1</version>
       </dependency>
       <dependency>
         <groupId>org.eclipse.jetty</groupId>
@@ -2077,7 +2082,6 @@
            -link http://www.jump-project.org/docs/jts/1.7/api
            -link http://www.geotools.fr/javadoc/2.2/
            -link http://xmlgraphics.apache.org/batik/javadoc
-           -link http://itextdocs.lowagie.com/docs
            -tag todo:tfmc:TODO:
            -tag task:tfmc:TODO:
            -tag revisit:tfmc:TODO:
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -1031,11 +1031,6 @@
         <version>1.3.26</version>
       </dependency>
       <dependency>
-        <groupId>com.twelvemonkeys.imageio</groupId>
-        <artifactId>imageio-tiff</artifactId>
-        <version>3.4.1</version>
-      </dependency>
-      <dependency>
         <groupId>org.eclipse.jetty</groupId>
         <artifactId>jetty-server</artifactId>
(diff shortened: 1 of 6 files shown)


src/wms/src/main/java/org/geoserver/wms/map/PDFMapResponse.java
```diff
@@ -117,153 +118,155 @@ public void write(Object value, OutputStream output, Operation operation)
             // height of document-object is height*72 inches
             com.lowagie.text.Rectangle pageSize = new com.lowagie.text.Rectangle(width, height);
```
ianturton, line 119: should this change? or do the packages remain the same despite the change?

ianturton (approved):
Looks good to me, it's a bit weird that the package names haven't changed but I guess that makes the change easier.

src/wms/src/main/java/org/geoserver/wms/map/PDFMapResponse.java
mprins: this class exists in openpdf: https://github.com/LibrePDF/OpenPDF/blob/master/openpdf/src/main/java/com/lowagie/text/Rectangle.java

src/pom.xml
```diff
@@ -1026,9 +1026,14 @@
...
       <dependency>
-        <groupId>com.lowagie</groupId>
-        <artifactId>itext</artifactId>
-        <version>2.1.5</version>
+        <groupId>com.github.librepdf</groupId>
+        <artifactId>openpdf</artifactId>
+        <version>1.3.26</version>
+      </dependency>
+      <dependency>
+        <groupId>com.twelvemonkeys.imageio</groupId>
```
mprins, line 1034: this dependency does not appear to be used anywhere?  I would expect it to come back in the WMS module, but it doesn't.

it seems openpdf only needs it to _read_ TIFF

src/wms/src/main/java/org/geoserver/wms/map/PDFMapResponse.java
bradh: It looks like openpdf is forward compatible with all the old com.lowagie names. By contrast itextpdf moved to `com.itextpdf` when it went to v5+.

src/pom.xml
bradh: I left it expecting to need it for another use that didn't turn out. In some ways it serves as a reminder , in other ways it's just noise. 

Would you prefer to see it removed?

mprins: if not used, I rather see it gone, it could generate noise in the future when there are updates, it could also force unexpected transitive dependency version - though I haven't checked on that

bradh: Will address.

bradh: Have removed it.

mprins (approved):

jodygarnett:
Thanks for the update @bradh; is this change suitable to backport?

bradh:
I see no reason why it couldn't be backported. I'd like to have a bit more time though.

PR #5620, opened 2022-01-27
[GEOS-10371] Removing LayerGroup does not remove associated data access rules

groldan:
[GEOS-10371](https://osgeo-org.atlassian.net/browse/GEOS-10371)

When a `LayerGroup` that belongs to a `Workspace` is removed,
its associated data access rules are removed.

When a root `LayerGropup` (i.e. null workspace) is removed,
its associated data access rules are not removed. Instead,
a `NullPointerException` is logged.

`SecuredResourceNameChangeListener` doesn’t check if a
`LayerGroup.getWorkspace()` is null, resuling in an NPE,
logged as a warning.

This patch simplifies the logic in `SecuredResourceNameChangeListener` that
would otherwise result in even more if-else clauses, makes sure
root layer group rules are treated properly, and adds test cases
for that, and for the rename cases, which weren't covered at all.

  

# Checklist

- N/A [Documentation](https://github.com/geoserver/geoserver/tree/main/doc/en/user/source) has been updated (if change is visible to end users).
- N/A The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers).

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/SecuredResourceNameChangeListener.java
+++ b/src/main/src/main/java/org/geoserver/security/SecuredResourceNameChangeListener.java
@@ -5,10 +5,14 @@
 package org.geoserver.security;
 
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 import org.geoserver.catalog.Catalog;
 import org.geoserver.catalog.CatalogException;
 import org.geoserver.catalog.CatalogInfo;
@@ -51,37 +55,29 @@ public class SecuredResourceNameChangeListener implements CatalogListener {
 
     @Override
     public void handleRemoveEvent(CatalogRemoveEvent event) throws CatalogException {
-        String resourceWorkSpaceName = null;
-        CatalogInfo resourceInfo = event.getSource();
-        String resourceName = null;
-        if (resourceInfo instanceof LayerInfo) {
-            resourceWorkSpaceName =
-                    ((LayerInfo) resourceInfo).getResource().getStore().getWorkspace().getName();
-            resourceName = ((LayerInfo) resourceInfo).getName();
-        } else if (event.getSource() instanceof LayerGroupInfo) {
-            resourceWorkSpaceName = ((LayerGroupInfo) resourceInfo).getWorkspace().getName();
-            resourceName = ((LayerGroupInfo) resourceInfo).getName();
+        final String removedObjectName; // for logging
+        final Predicate<DataAccessRule> filter;
+        final CatalogInfo eventSource = event.getSource();
+        if (eventSource instanceof WorkspaceInfo) {
+            WorkspaceInfo ws = (WorkspaceInfo) eventSource;
+            filter = workspaceFilter(ws.getName());
+            removedObjectName = "Workspace " + ws.getName();
+        } else if (eventSource instanceof LayerInfo) {
+            LayerInfo l = (LayerInfo) eventSource;
+            WorkspaceInfo ws = l.getResource().getStore().getWorkspace();
+            filter = layerFilter(ws.getName(), l.getName());
+            removedObjectName = "Layer " + l.getName();
+        } else if (eventSource instanceof LayerGroupInfo) {
+            LayerGroupInfo lg = (LayerGroupInfo) eventSource;
+            filter = layerGroupFilter(lg.getWorkspace(), lg.getName());
+            removedObjectName = "Layer Group " + lg.getName();
+        } else {
+            return;
         }
 
-        if (resourceWorkSpaceName != null && resourceInfo != null) {
-            // if a security rule exists for the layer whose name has been changed
-            if (layerHasSecurityRule(resourceWorkSpaceName, resourceName)) {
-                LOGGER.info(
-                        String.format(
-                                "Removing Security Rules for removed layer: %s", resourceName));
-
-                List<DataAccessRule> rulesToRemove = getDataAccessRule(resourceInfo, resourceName);
-                for (DataAccessRule r : rulesToRemove) {
-                    dao.removeRule(r);
-                }
-                try {
-                    dao.storeRules();
-                } catch (IOException e) {
-                    LOGGER.log(Level.SEVERE, e.getMessage(), e);
-                    throw new CatalogException(e);
-                }
-            }
-        }
+        Supplier<String> message =
+                () -> String.format("Removing Security Rules for deleted %s", removedObjectName);
+        apply(filter, dao::removeRule, message);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


src/main/src/main/java/org/geoserver/security/SecuredResourceNameChangeListener.java
```diff
@@ -51,37 +54,32 @@ public void handleAddEvent(CatalogAddEvent event) throws CatalogException {
...
-                    throw new CatalogException(e);
-                }
+        Predicate<DataAccessRule> filter;
+        final String layerName;
+        final CatalogInfo eventSource = event.getSource();
+        if (eventSource instanceof LayerInfo) {
+            LayerInfo l = (LayerInfo) eventSource;
+            WorkspaceInfo ws = l.getResource().getStore().getWorkspace();
+            layerName = l.getName();
+            filter = filter(ws.getName()::equalsIgnoreCase, layerName::equalsIgnoreCase);
```
aaime, line 64: I see you're just replicating the original behavior but... wondering why it's case insensitive.
Layer names are case sensitive in GeoServer, and so seems the application of rules in DefaultResourceAccessManager.

Checked if there was any discussion about it in the original PR, but [there seems to have been none](https://github.com/geoserver/geoserver/pull/3644).

src/main/src/main/java/org/geoserver/security/SecuredResourceNameChangeListener.java
```diff
@@ -51,37 +54,32 @@ public void handleAddEvent(CatalogAddEvent event) throws CatalogException {
...
+        } else if (eventSource instanceof LayerGroupInfo) {
+            LayerGroupInfo lg = (LayerGroupInfo) eventSource;
+            WorkspaceInfo ws = lg.getWorkspace();
+            layerName = lg.getName();
+            if (ws == null) {
+                filter = filter(layerName::equalsIgnoreCase, r -> true);
```
aaime, line 70: I was doubtful about this one, but yes, it seems a global layer group rule puts the group name in "root" and leaves the layer name empty. Maybe a comment about it would be useful.

src/main/src/main/java/org/geoserver/security/SecuredResourceNameChangeListener.java
```diff
@@ -91,61 +89,75 @@ public void handleModifyEvent(CatalogModifyEvent event) throws CatalogException
...
-                        getDataAccessRule(event.getSource(), oldName);
-                for (DataAccessRule rule : rulesToModifyList) rule.setRoot(newName);
-            }
-        } else if (event.getSource() instanceof ResourceInfo) {
+        final CatalogInfo eventSource = event.getSource();
+        final Predicate<DataAccessRule> filter;
+        final Consumer<DataAccessRule> updater;
+        if (eventSource instanceof WorkspaceInfo) {
+            // if workspace has been renamed
+            filter = filter(oldName::equalsIgnoreCase, l -> true);
```
aaime, line 109: Wouldn't this catch also global layer groups? A workspace and a global layer group might share the same name.

src/main/src/main/java/org/geoserver/security/SecuredResourceNameChangeListener.java
```diff
@@ -91,61 +89,75 @@ public void handleModifyEvent(CatalogModifyEvent event) throws CatalogException
...
+            updater = r -> r.setLayer(newName);
+        } else if (eventSource instanceof LayerGroupInfo) {
             // if a layergroup has been renamed
             // similar layergroup names can exist in different workspaces
-            resourceInfo = event.getSource();
-            resourceWorkSpaceName = ((LayerGroupInfo) resourceInfo).getWorkspace().getName();
+            LayerGroupInfo lg = (LayerGroupInfo) eventSource;
+            WorkspaceInfo ws = lg.getWorkspace();
+            if (ws == null) {
+                filter = filter(oldName::equalsIgnoreCase, r -> true);
```
aaime, line 123: Same as above but opposite direction, might catch a workspace rule (admin rule).

src/main/src/main/java/org/geoserver/security/SecuredResourceNameChangeListener.java
```diff
@@ -91,61 +89,75 @@ public void handleModifyEvent(CatalogModifyEvent event) throws CatalogException
...
-                List<DataAccessRule> rulesToModifyList = getDataAccessRule(resourceInfo, oldName);
-                for (DataAccessRule rule : rulesToModifyList) rule.setLayer(newName);
-            }
+    void apply(
+            Predicate<DataAccessRule> filter,
+            Consumer<DataAccessRule> action,
+            Supplier<String> logMessage) {
+        List<DataAccessRule> matches = getDataAccessRules(filter);
+        if (!matches.isEmpty()) {
+            LOGGER.info(logMessage.get());
```
aaime, line 149: Better give it the supplier directly, it can skip the message production if the LOGGER level is higher.

src/main/src/test/java/org/geoserver/catalog/impl/CatalogImplTest.java
```diff
@@ -1554,13 +1555,19 @@ private boolean layerHasSecurityRule(
...
-            if (rule.getRoot().equalsIgnoreCase(workspaceName)
-                    && rule.getLayer().equalsIgnoreCase(layerName)) return true;
+            if (rule.getRoot().equalsIgnoreCase(workspaceName) && sameLayer(layerName, rule))
+                return true;
         }
 
         return false;
     }
 
+    private boolean sameLayer(String layerName, DataAccessRule rule) {
```
aaime, line 1565: Name of method and what it actually does is surprising. It works, in the context of the calling point, because the root was already checked. Maybe move the entire test in the if inside the method? Then it would make more sense.

src/main/src/test/java/org/geoserver/catalog/impl/CatalogImplTest.java
```diff
@@ -2174,9 +2184,109 @@ public void testGetLayerGroupByName() {
         assertNull(catalog.getLayerGroupByName("cite", "layerGroup2"));
     }
 
+    @Test
+    public void testRenameResourceRenamesAssociatedDataRules() throws IOException {
+        // Testing only with a FeatureType because I know SecuredResourceNameChangeListener works at
+        // the ResourceInfo level. If for any reason you change that and don't update the tests
```
aaime, line 2190: Non useful comment, nobody will remember it when changing other code. If there is something people need to remember, best put it where they might do the change that would break this assumption instead.

src/main/src/main/java/org/geoserver/security/SecuredResourceNameChangeListener.java
groldan: I don't know, just wouldn't dare changing the current behavior

src/main/src/main/java/org/geoserver/security/SecuredResourceNameChangeListener.java
groldan: That is a good catch, thanks for looking into this. In fact, they shouldn't collide, since workspace rules never have `null` layer, but an actual name or the `*` glob. e.g., if you have both a ws and a global group named `osm`, rules on them would be like this:

image

I'm adding a fix and a test case.

groldan: @aaime in writing the test case, I found that when a workspace is removed, its data rules aren't removed.
do you know if that's intentional or should they also be removed?
cause I'm just at the point where it'd be a couple lines of code to fix it

aaime: Does not seem intentional to me, more like another issue.

src/main/src/test/java/org/geoserver/catalog/impl/CatalogImplTest.java
groldan: removed.

src/main/src/test/java/org/geoserver/catalog/impl/CatalogImplTest.java
groldan: removed method and inlined check. I think at some point it was called by more methods .

src/main/src/main/java/org/geoserver/security/SecuredResourceNameChangeListener.java
groldan: good catch, done.

src/main/src/main/java/org/geoserver/security/SecuredResourceNameChangeListener.java
groldan: Fixed as mentioned above.

src/main/src/main/java/org/geoserver/security/SecuredResourceNameChangeListener.java
groldan: it is more obvious now, removed the `if`:
```
filter = layerGroupFilter(lg.getWorkspace(), lg.getName());
```
And [documented](https://github.com/geoserver/geoserver/pull/5620/files#diff-f76ab8b4ba5e8409fd7059659a76119a4f24a1dbe3aa8e0e6a45524951e8bdbaR171) the extracted method.

PR #5622, opened 2022-01-27
[GEOS-10372] Geopackage integration tests for appschema

ydeliorman:
[GEOS-10372](https://osgeo-org.atlassian.net/browse/GEOS-10372)

Geopackage integration tests are added for appschema.
  

# Checklist


src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/onlineTest/support/AbstractReferenceDataSetup.java
```diff
@@ -91,6 +93,14 @@ protected void configureFixture() {
                 if (exists == null || exists.booleanValue()) {
                     if (fixtureFile.exists()) {
                         fixture = GSFixtureUtilitiesDelegate.loadProperties(fixtureFile);
+                        // update jdbc connection url. take geopkg file from resources and use that
+                        // geopkg in the tests
+                        if (fixtureId.equals("geopkg")) {
+                            fixture.setProperty(
+                                    "url",
+                                    fixture.getProperty("url").substring(0, 11)
+                                            + GEOPKG_DIR.substring(4));
```
taba90, line 102: Instead of having a public static variable around and configuring the Fixture in the constructor of  ReferenceDataGeopackageSetup I would resolve the url property [here ](https://github.com/geoserver/geoserver/blob/e1a596a23eaecfb35b139a68ee8a8c3ce33d942b/src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java#L602) and then call the setFixture method. So we don't need to have any static attribute around.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java
```diff
@@ -773,4 +841,67 @@ private String modifyOnlineMappingFileContent(String mappingFileName) throws IOE
...
+                new ByteArrayInputStream(newIncludedTypes.getBytes(StandardCharsets.UTF_8))) {
+            copyTarget(bais, "appschema/appschema-gml3132/meteo.appschema");
+        }
+    }
+
+    private void copyGeopkgResourcesFolder() {
+        copy(
+                getClass().getClassLoader().getResourceAsStream("appschema/stations.gpkg"),
+                "appschema/stations.gpkg");
+        copy(
```
taba90, line 886: why in addition to the geopackage you need to copy also the xsd and mapping files? For what I know mappings and xsd are already handled by the addFeatureType method. Did I miss something?
In case they are not needed please amend the code and remove the additional files in the resource folder.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java
```diff
@@ -172,6 +183,15 @@ public AbstractAppSchemaMockData(Map<String, String> namespaces) {
...
 
+        appschema = new File(data, "appschema");
+        appschema.mkdir();
+
+        appschemagml3132 = new File(appschema, "appschema-gml3132");
+        appschemagml3132.mkdir();
+
+        appschemacache = new File(appschemagml3132, "app-schema-cache");
+        appschemacache.mkdir();
+
```
taba90, line 194: I guess these directories are for the mappings and xsd in the resource folder. Please check the comment below.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java
```diff
@@ -738,8 +784,30 @@ private String modifyOnlineMappingFileContent(String mappingFileName) throws IOE
...
+                                copyGeopkgResourcesFolder();
+                                File[] stations =
+                                        Objects
+                                                .requireNonNull(
+                                                        data.listFiles(
+                                                                (dir, name) ->
+                                                                        name.startsWith(
+                                                                                "appschema")))[0]
+                                                .listFiles(
+                                                        (dir, name) -> name.startsWith("stations"));
```
taba90, line 799: I would directly check for the entire file name with the gpkg extension instead of simply checking the fileName.
Also could you refact a bit this piece of code to make it more readable?

taba90 (changes requested):
There are some part of code that could be improved. Also is not clear if we really need additional mappings and xsd bought by this pr and why.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java
ydeliorman: you are right. fixed.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java
ydeliorman: fixed.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java
ydeliorman: done

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/onlineTest/support/AbstractReferenceDataSetup.java
ydeliorman: I tried to pass the string as you suggested instead of the static variable. Since we need dir path in [modifyOnlineMappingFileContent](https://github.com/geoserver/geoserver/blob/e1a596a23eaecfb35b139a68ee8a8c3ce33d942b/src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java#L805) we need to pass the string from the parameters of the function. We can make a global variable in the class to carry the string and evaluate it at the point you suggested. However, we also need to change [configureFixture](https://github.com/geoserver/geoserver/blob/e1a596a23eaecfb35b139a68ee8a8c3ce33d942b/src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/onlineTest/support/AbstractReferenceDataSetup.java#L75) parameters for every postgis and oracle and will be receiving an empty string.

We also need to have the string ready in [here](https://github.com/geoserver/geoserver/blob/e1a596a23eaecfb35b139a68ee8a8c3ce33d942b/src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java#L716). If we evaluate dir string before setup(), geopkg directory is null at this place. We can add if checks in here to move to assign public geopkg dir string to here but I think this will add too much code.

taba90: The problematic point is indeed modifyOnlineMappingFileContent this one: https://github.com/geoserver/geoserver/blob/e1a596a23eaecfb35b139a68ee8a8c3ce33d942b/src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java#L805. The current implementation is setting to a static variable in order to make the geopackage file available before inserting the data.
My point is then avoid static variable and assing it to an instance attribute. Then in [this ](https://github.com/geoserver/geoserver/blob/e1a596a23eaecfb35b139a68ee8a8c3ce33d942b/src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java#L602) else if statement you retrieve the file path and set it to the Property and then you explicitly set the properties trough setFixture.

ydeliorman: We cannot call setFixture because we are having a null value for the setup variable until [this line](https://github.com/geoserver/geoserver/blob/e1a596a23eaecfb35b139a68ee8a8c3ce33d942b/src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java#L601) is executed. If we use setFixture before that line, we are having a null pointer exception due to having setup is null. 

We cannot call setFixture after that line because we are using the geopkgDir string in the getInstance function. Therefore I passed the geopkgDir string as a parameter in getInstance function.

ydeliorman:
@taba90 applied the review about static string

taba90 (approved):

PR #5631, opened 2022-01-30
[GEOS-10374] Log exception flooding opening GWC's NewCachedLayerPage

groldan:
[GEOS-10374](https://osgeo-org.atlassian.net/browse/GEOS-10374)

Fix `java.io.NotSerializableException` log flooding at
`NewCachedLayerPage.java` due to its inner class `CachedSize`
not implementing `Serializable`, introduced at commit cf77a0baef4.

  

# Checklist

### DIFF
--- a/src/web/gwc/src/main/java/org/geoserver/gwc/web/layer/UnconfiguredCachedLayersProvider.java
+++ b/src/web/gwc/src/main/java/org/geoserver/gwc/web/layer/UnconfiguredCachedLayersProvider.java
@@ -6,6 +6,7 @@
 package org.geoserver.gwc.web.layer;
 
 import com.google.common.collect.Streams;
+import java.io.Serializable;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
@@ -74,7 +75,8 @@ class UnconfiguredCachedLayersProvider extends GeoServerDataProvider<TileLayer> 
      * Simple cache for the last computed size based on the keywords, which in turn are the sole
      * input for the filter (at the moment, at least)
      */
-    private class CachedSize {
+    private class CachedSize implements Serializable {
+        private static final long serialVersionUID = 1L;
         private static final long NOT_CACHED = Long.MIN_VALUE;
         private String[] cachedSizeKeywords;
         private long cachedSize = NOT_CACHED;
--- a/src/web/gwc/src/test/java/org/geoserver/gwc/web/layer/NewCachedLayerPageTest.java
+++ b/src/web/gwc/src/test/java/org/geoserver/gwc/web/layer/NewCachedLayerPageTest.java
@@ -5,6 +5,12 @@
  */
 package org.geoserver.gwc.web.layer;
 
+import static org.junit.Assert.assertTrue;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.ObjectOutputStream;
+import org.apache.wicket.serialize.ISerializer;
 import org.geoserver.web.GeoServerWicketTestSupport;
 import org.junit.Test;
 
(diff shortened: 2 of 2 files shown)


groldan:
QA build failing due to unrelated issue, looks like waiting for #5630 to be merged.

```
Error: 8,414 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project gs-main: Compilation failure
Error: 8,414 [ERROR] /home/runner/work/geoserver/geoserver/src/main/src/main/java/org/geoserver/security/decorators/SecuredWebMapTileServer.java: warnings found and -Werror specified
```

src/web/gwc/src/main/java/org/geoserver/gwc/web/layer/UnconfiguredCachedLayersProvider.java
```diff
@@ -74,7 +75,8 @@
      * Simple cache for the last computed size based on the keywords, which in turn are the sole
      * input for the filter (at the moment, at least)
      */
-    private class CachedSize {
+    private class CachedSize implements Serializable {
```
aaime, line 78: Yep, I see how this is happening. At the same time, I haven't noticed that trace in the logs, guess my normal usage does not force Wicket to serialize pages (grepped through the logs of the data dirs I use for testing, it's not there...).

Can you think of a way to make these easier to catch when developing? Some settings in the WicketTestSupport that would make these appear right away?

aaime:
QA issue should be addressed now, could up rebase and push, just to make sure there is nothing else? (there shouldn't be, but the QA build failed right at the beginning).

src/web/gwc/src/main/java/org/geoserver/gwc/web/layer/UnconfiguredCachedLayersProvider.java
groldan: that's an interesting thought, I'm not sure what's triggering it in my case but probably the use of `WebUIMode.DO_NOT_REDIRECT`. Won't make a difference in the simple "check page loads" test though.

The following test can be added, but not sure how/if to enforce it for all wicket tests:
```
    @Test
    public void testSerializable() throws IOException {
        login();
        NewCachedLayerPage page = new NewCachedLayerPage();
        ISerializer serializer = getGeoServerApplication().getFrameworkSettings().getSerializer();
        //this would only result in an ERROR level log
        serializer.serialize(page);
        
        try {
            new ObjectOutputStream(new ByteArrayOutputStream()).writeObject(page);
        }catch(NotSerializableException e) {
            fail(e.getMessage());
        }
    }
```
```
java.lang.AssertionError: org.geoserver.gwc.web.layer.UnconfiguredCachedLayersProvider$CachedSize
```

groldan:
rebased and pushed, QA build is happy again.

src/web/gwc/src/main/java/org/geoserver/gwc/web/layer/UnconfiguredCachedLayersProvider.java
groldan: Actually I just did add the test for this specific case and force-pushed.

aaime: Hum... looking at the test you setup there gave me an idea... the simple "check page loads" code goes though ``WicketTester.startPage``.... so maybe we could roll a subclass of it, that checks the page can be serialized, whenever startPage is used... yes, it would do that multiple times for nothing, but on the other end, it's too easy to forget doing so manually for each and every page.

Maybe the subclass could be smart enough to keep a cache of which classes it has already verified.

It still would not cover all cases, maybe the page could get something non serializable in its model after some user interaction... oh well, just  thinking out loud...

PR #5665, opened 2022-02-10
Add limitation: User Enabled checkbox does not affect Authkey module

petersmythe:
ref: https://sourceforge.net/p/geoserver/mailman/message/37602011/

# Checklist

### DIFF
--- a/doc/en/user/source/extensions/authkey/index.rst
+++ b/doc/en/user/source/extensions/authkey/index.rst
@@ -33,7 +33,7 @@ Key providers
 -------------
 
 Key providers are responsible for mapping the authentication keys to a user. The authentication key
-itself is a UUID (Universal Unique Identifier). A key provider needs a user/group service and it is
+itself is a UUID (Universally Unique IDentifier (UUID)). A key provider needs a user/group service and it is
 responsible for synchronizing the authentication keys with the users contained in this service.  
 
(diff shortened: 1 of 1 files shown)


doc/en/user/source/extensions/authkey/index.rst
```diff
@@ -26,11 +26,14 @@ Limitations
...
 
+The ``authkey`` module replaces the default basic authentication provider, therefore when editing a user
+the Enabled checkbox does not affect the ``authkey`` provider and you cannot disable users this way.
+
 Key providers
 -------------
 
 Key providers are responsible for mapping the authentication keys to a user. The authentication key
-itself is a UUID (Universal unique Identifier). A key provider needs a user/group service and it is
+itself is a UUID (Universal Unique Identifier). A key provider needs a user/group service and it is
```
bradh, line 36: I can't make a suggested change (not sure why), but it should be 
"Universally Unique IDentifier (UUID)" per https://datatracker.ietf.org/doc/html/rfc4122

Otherwise LGTM.

mprins, line 36:
```suggestion
itself is a UUID (Universally Unique IDentifier (UUID)). A key provider needs a user/group service and it is
```

mprins (approved):

PR #5667, opened 2022-02-11
[GEOS-10387] Thread updating a cache in ResourcePool might block indefinitely other threads reading from the same caches

taba90:
# Checklist


src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -1184,9 +1219,19 @@ public AttributeDescriptor getAttributeDescriptor(
...
-        featureTypeCache.remove(getFeatureTypeInfoKey(info, true));
-        featureTypeCache.remove(getFeatureTypeInfoKey(info, false));
-        featureTypeAttributeCache.remove(info.getId());
+        String id = info.getId();
+        if (id != null) {
+            String id2 = getFeatureTypeInfoKey(info, true);
+            String id3 = getFeatureTypeInfoKey(info, false);
+            featureTypeCache.remove(id2);
+            featureTypeCache.remove(id3);
+            featureTypeAttributeCache.remove(id);
```
aaime, line 1228: How do we get id into the cache?

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -1184,9 +1219,19 @@ public AttributeDescriptor getAttributeDescriptor(
...
+            featureTypeAttributeCache.remove(id2);
+            featureTypeAttributeCache.remove(id3);
+            stringCacheKeys.remove(id);
```
aaime, line 1231: Same as above, self managing.

src/main/src/test/java/org/geoserver/catalog/ResourcePoolTest.java
```diff
@@ -971,4 +980,290 @@ public void testCoverageReaderInputConverterFilePath() throws IOException {
...
     }
+
+    @Test
+    public void testGetDataStoreConcurrency() throws Exception {
+        ResourcePool pool = null;
+        try {
+            final Catalog catalog = getCatalog();
+            DataStoreInfo ds = storeInfo(catalog, "concurrencyTest1");
+            pool =
+                    new ResourcePool(getCatalog()) {
```
aaime, line 991: This anonymous class seems to be doing nothing new? Why not just use ResourcePool?

src/main/src/test/java/org/geoserver/catalog/ResourcePoolTest.java
```diff
@@ -971,4 +980,290 @@ public void testCoverageReaderInputConverterFilePath() throws IOException {
...
+
+            ResourcePoolLatchedThread.PoolBiFunction<DataStoreInfo, DataAccess> function =
+                    (resPool, info) -> resPool.getDataStore(info);
+            CountDownLatch taskLatch = new CountDownLatch(1);
+            int numberOfThreads = 5;
+            CountDownLatch completeLatch = new CountDownLatch(numberOfThreads);
+            List<ResourcePoolLatchedThread<DataStoreInfo, DataAccess>> threads =
+                    getLatchedThreads(
+                            taskLatch, completeLatch, numberOfThreads, pool, ds, function);
+            threads.forEach(t -> t.start());
```
aaime, line 1008: So you are explicitly creating threads and starting them... but I don't see you destroying them, meaning they will accumulate in VM. It's a test JVM, so no big deal, but could cause troubles in the future, as Maven surefire testing JVMs can be reused across modules. Best clean up and kill the threads, here and in the other tests. Or create runnables and submit them to a Executor (you'll still have to shut down the executor, anyways).

aaime (changes requested):

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -205,6 +207,18 @@
      */
     TransformFeatureTypeCallback transformer = new TransformFeatureTypeCallback();
 
+    /**
+     * Holds the keys for all the cache having String keys. By ensuring identity allows to
+     * synchronize on id values.
+     */
+    private CanonicalSet<String> stringCacheKeys;
```
aaime, line 214: Maybe just call it cacheKeys? And the other one, coverageCacheKeys.

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -401,6 +417,7 @@ public void setFeatureTypeCacheSize(int featureTypeCacheSize) {
             featureTypeCache = createFeatureTypeCache(featureTypeCacheSize);
             featureTypeAttributeCache.clear();
             featureTypeAttributeCache = createFeatureTypeAttributeCache(featureTypeCacheSize);
+            stringCacheKeys.clear();
```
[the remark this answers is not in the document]
taba90: thanks for your point, removed

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -488,7 +505,7 @@ public CoordinateReferenceSystem getCRS(String srsName) throws IOException {
      * <p>This method first uses {@link DataStoreInfo#getType()} to obtain the datastore. In the
      * event of a failure it falls back on {@link DataStoreInfo#getConnectionParameters()}.
      *
-     * @param info The data store metadata.
+     * @param info The data stofre metadata.
```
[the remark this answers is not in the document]
taba90: fixed

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -529,115 +546,20 @@ public DataAccessFactory getDataStoreFactory(DataStoreInfo info) throws IOExcept
...
         DataAccess<? extends FeatureType, ? extends Feature> dataStore = null;
         try {
-            String id = info.getId();
-            dataStore = dataStoreCache.get(id);
+
+            String storeId = info.getId();
+            // cache only if the id is not null, no need to cache the stores
+            // returned from un-saved DataStoreInfo objects (it would be actually
+            // harmful, NPE when trying to dispose of them)
+            if (storeId == null) return createDataAccess(info, expandedStore);
```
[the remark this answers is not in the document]
taba90: fixed

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -742,7 +763,11 @@ public DataAccessFactory getDataStoreFactory(DataStoreInfo info) throws IOExcept
      * @param info The data store metadata.
      */
     public void clear(DataStoreInfo info) {
-        dataStoreCache.remove(info.getId());
+        String id = info.getId();
+        if (id != null) {
+            dataStoreCache.remove(id);
+            stringCacheKeys.remove(id);
```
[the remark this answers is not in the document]
taba90: fixed

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
taba90: fixed

src/main/src/test/java/org/geoserver/catalog/ResourcePoolTest.java
taba90: fixed

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -529,115 +546,20 @@ public DataAccessFactory getDataStoreFactory(DataStoreInfo info) throws IOExcept
 
         DataAccess<? extends FeatureType, ? extends Feature> dataStore = null;
         try {
-            String id = info.getId();
-            dataStore = dataStoreCache.get(id);
+
+            String storeId = info.getId();
+            // cache only if the id is not null, no need to cache the stores
```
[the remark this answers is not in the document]
taba90: Netiher the previous code was putting the store in the cache. For what I saw debugging, a store with null id can comes only when saving a new store [here ](https://github.com/geoserver/geoserver/blob/71bfcd03119c2bba41e3c8133b6c520e62f691ac/src/web/core/src/main/java/org/geoserver/web/data/store/DataAccessNewPage.java#L91), because the id is set later from the catalog, and the data access get then disposed correctly.

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
taba90: fixed

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -932,31 +959,39 @@ FeatureType tryGetFeatureType(FeatureTypeInfo info, boolean handleProjectionPoli
 
     FeatureType getCacheableFeatureType(FeatureTypeInfo info, boolean handleProjectionPolicy)
             throws IOException {
-        String key = getFeatureTypeInfoKey(info, handleProjectionPolicy);
+
+        String id = info.getId();
+        if (id == null) return acquireFeatureType(info, handleProjectionPolicy);
```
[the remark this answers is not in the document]
taba90: fixed

taba90:
thanks for your review @aaime. Feedback applied.

aaime (approved):
Looks good to me

PR #5675, opened 2022-02-14
[GEOS-9785] Invalid argument type=null when trying to use gs:Download WPS identifier

sergitopo:
[GEOS-9785](https://osgeo-org.atlassian.net/browse/GEOS-9785)

WPS download fails to estimate either the response result will exceed the size limit configured or not. Then no download process is executed.

This bug occurs when there is a Coverage saved in catalog which has missing info regarding the pixel dimension type in one or more bands. Unluckily this happens for the default geoserver coverages included in demo data.

# Checklist


src/extension/wps-download/src/main/java/org/geoserver/wps/gs/download/RasterEstimator.java
```diff
@@ -200,6 +200,14 @@ public boolean execute(
 
         // Use sample info type for each output band to estimate size
         List<CoverageDimensionInfo> coverageDimensionInfoList = coverageInfo.getDimensions();
+
+
+        if (coverageDimensionInfoList.stream().anyMatch(cdi -> cdi.getDimensionType() == null)) {
+            if (LOGGER.isLoggable(Level.FINE)) {
```
aaime, line 206: The guard is not necessary when the message is a static string (no construction of new objects needed).

src/extension/wps-download/src/main/java/org/geoserver/wps/gs/download/RasterEstimator.java
```diff
@@ -250,4 +258,24 @@ public boolean execute(
...
+            GridCoverage2DReader reader =
+                    (GridCoverage2DReader)
+                            catalog.getResourcePool()
+                                    .getGridCoverageReader(
+                                            ci, nativeName, GeoTools.getDefaultHints());
+            rebuilt = cb.buildCoverage(reader, nativeName, null);
+        } else {
+            rebuilt = cb.buildCoverage(nativeName);
+        }
+        ci.getDimensions().clear();
```
aaime, line 278: This update is not really saving any new information in the catalog, and it's not intended to either (having config that gets modified as a result of a WPS process meant to download stuff would be unexpected).

Maybe it's best to just return rebuilt.getDimensions() and assign that to the dimension list instead.

aaime (changes requested):
Looks good, small issue in the code, there is also a formatting issue, see the QA build results.

src/extension/wps-download/src/main/java/org/geoserver/wps/gs/download/RasterEstimator.java
sergitopo: Sorry Andrea, where is the construction of a new object in these lines?

aaime: Indeed there is none ("no construction of new objects needed") so no need to guard  the statement

sergitopo:
> Looks good, small issue in the code, there is also a formatting issue, see the QA build results.
Solved in latest commit.

aaime:
Looks good now. Found a typo, and QA build is failing on PMD:

```
2022-02-15T10:25:54.7695676Z 10:25:54,769 [INFO] PMD Failure: org.geoserver.wps.gs.download.RasterEstimator:11 Rule:wildcards Priority:3 No Wildcard Imports.
```

Caused by:

``
import org.geoserver.catalog.*;
``

See also:
- [the coding convention](https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports)
- https://stackoverflow.com/questions/147454/why-is-using-a-wild-card-with-a-java-import-statement-bad

[You can setup IntelliJ not to create wildcard imports](https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#disable-wildcard-imports).

PR #5677, opened 2022-02-15
[GEOS-10389] Introduce ENTITY_RESOLUTION_ALLOWLIST to further restrict entity resolver

jodygarnett:
[GEOS-10389](https://osgeo-org.atlassian.net/browse/GEOS-10389)

# Checklist


src/main/src/main/java/org/geoserver/util/InternalEntityResolver.java
```diff
@@ -0,0 +1,142 @@
...
+
+    protected static final Logger LOGGER = Logging.getLogger(InternalEntityResolver.class);
+
+    /** Internal uri references */
+    private static final Pattern INTERNAL_URIS = Pattern.compile("(?i)(jar:file|vfs)[^?#;]*\\.xsd");
+
+    /** OGC Schema references */
+    private static final Pattern OGC_URIS =
+            Pattern.compile(
+                    "(?i)(http|https)://(schemas.opengis.net|www.opengis.net)/[^?#;]*\\.xsd");
```
jodygarnett, line 38: feedback: Can this list of hosts be supplied by an environment variable?

src/main/src/main/java/org/geoserver/util/EntityResolverProvider.java
```diff
@@ -26,21 +30,57 @@
...
             Boolean externalEntitiesEnabled = geoServer.getGlobal().isXmlExternalEntitiesEnabled();
             if (externalEntitiesEnabled != null && externalEntitiesEnabled) {
                 // XML parser will try to resolve entities
                 return null;
             }
-        }
 
+            if (geoServer.getSettings() != null
+                    && geoServer.getSettings().getProxyBaseUrl() != null) {
+                return INTERNAL_ENTITY_RESOLVER;
```
jodygarnett, line 62: feedback: This should be disabled by default, consider only enabling if environment variable providing list of hosts is known

src/main/src/main/java/org/geoserver/util/InternalEntityResolver.java
jodygarnett: Introduced ENTITY_RESOLUTION_ALLOWLIST property

jodygarnett:
Provided as distinct commits to ease review process, use squash and merge when ready for ease of back porting to 2.20.x.

jodygarnett:
Thanks Ian, I had intended to squash and merge (to make back porting easier, but this works also).

PR #5683, opened 2022-02-16
[GEOS-10391] Add option to limit directories that processes can write to

sikeoka:
[GEOS-10391](https://osgeo-org.atlassian.net/browse/GEOS-10391)

This PR adds the ability for GeoServer administrators to limit WPS processes to only write within a specific output directory or to completely disable user-specified output file paths.

  

# Checklist


aaime (commented):
I understand the change, and see it comes with a backwards incompatibility anyways, since by default the external storage will be null, disallowing to write outside of the data directory.

The georectify process is supported, so we should keep the ability to write on a specific path.
The other two have always been unsupported, wondering if we should just get rid of the ability to write to a random disk position... @jodygarnett what do you think? I'm asking because I believe may be the only one with actual users of those processes (well, the GeoPackage one at least).

doc/en/user/source/services/wps/administration.rst
```diff
@@ -34,6 +34,7 @@ Resource settings:
 
   * *Resource expiration timeout*: number of seconds the result of a asynchronous execution will be kept available on disk for user to retrieve. Once this time is expired these resources will be eligible for clearing (which happens at regular intervals).
   * *Resource storage directory*: where on disk the input, temporary and output resources associated to a certain process will be kept. By default it will be the ``temp/wps`` directory inside the GeoServer data directory
+  * *External output directory*: where on disk, certain processes are allowed to write their outputs outside of the resource storage so that the files will not be automatically removed based on the resource expiration timeout. Leave empty to disable writing outside of the resource storage.
```
jodygarnett, line 37: Is this option only available for process that write files? Or is it for any async response ...

jodygarnett: Oh I see: difference between "resource storage" and "external output" directory.

Still does not quite answer my question about async output. Consider using: "response" and "result" rather than "output".

```
  * :guilabel:`Resource expiration timeout`: number of seconds the result of an asynchronous execution will be kept available on disk for user to retrieve. Once this time is expired these resources will be eligible for clearing (which happens at regular intervals).
  * :guilabel:`Resource storage directory`: where on disk the temporary input resources, temporary outputs resources, and asynchronous results of a process execution will be kept. By default it will be the :file:`temp/wps` directory inside the GeoServer data directory. These resources are subject to the :guilabel:`Resource expiration timeout` above.
  * :guilabel:`External output directory`: Some processes allow execution outputs to be stored in an external output directory (not subject :guilabel:`Resource expiration timeout`). To enable this functionality provide a path to external storage with the understanding that you are responsible for managing the contents of this folder. Leave empty to disable writing outside of the resource storage.
```

src/community/geopkg/src/main/java/org/geoserver/geopkg/wps/GeoPackageProcess.java
```diff
@@ -171,9 +171,8 @@ public URL execute(
 
         String outputName = contents.getName() + ".gpkg";
         if (!remove && path != null) {
-            File urlToFile = URLs.urlToFile(path);
-            urlToFile.mkdirs();
-            file = new File(urlToFile, contents.getName() + ".gpkg");
+            String urlToFile = URLs.urlToFile(path).getPath();
+            file = resources.getExternalOutputFile(urlToFile, outputName);
```
jodygarnett, line 175: Since `resources` is responsible for both getExternalOutputFile and getOutputResource could it take responsibility for this logic to avoid possibility for processes handing this parameter in an inconsistent manner.

`resources` already knows if setExternalOutputDirectory has been provided.

sikeoka:
I updated the PR based on your feedback.

aaime:
@jodygarnett our initial question was more "can we remove the external output altoghether in the unsupported processes"?

sikeoka:
Removing the option from only some of the processes doesn't really impact the necessity of this PR so, in my opinion, the option should just be left available and secured in all of the processes that currently support it unless the plan is to remove the option from every process that has it.

aaime:
I disagree (strongly so). If possible, we should remove the options from modules that have no "user promise" attached to them (community ones), the one in extension is the problem, because there we should try hard not to break backward compatibility instead.

src/community/geopkg/src/main/java/org/geoserver/geopkg/wps/GeoPackageProcess.java
sikeoka: The code was moved into WPSResourceManager.

doc/en/user/source/services/wps/administration.rst
sikeoka: The documentation was updated based on your feedback.

jodygarnett: Thanks!

jodygarnett:
I do not have need to maintain an external output directory at present, the tmp folder subject to time out is fine for my use at present.  I am not sure I understand what "get rid of the ability to write to a random disk position" is proposing?

The GeoPackage output was only just made into an extension so now is the time to break backwards compatibility if there is reason to do so.

aaime:
@jodygarnett good point, then my take would be to get rid of ability to write on random disk positions entirely, in all processes.

jodygarnett:
I would support that.

sikeoka:
Just the be clear, the consensus is to remove the writing parameter from the GeoPackage and MBTiles processes?

aaime:
Yes, the "path" parameter in the mbtiles process and the path field in GeoPackageProcessRequest.

sikeoka:
I removed the path parameters from the Geopackage and MBTiles processes.

aaime:
Looks good to me,  merged.

PR #5687, opened 2022-02-18
[GEOS-10393] Improve WFS-T BatchManager

awaterme:
[GEOS-10393](https://osgeo-org.atlassian.net/browse/GEOS-10393)

Hi there,

here is the PR for [GEOS-10393](https://osgeo-org.atlassian.net/browse/GEOS-10393).

I have also removed methods from "org.geoserver.wfs.request.Delete" and "org.geoserver.wfs.request.Insert" classes which where introduced for BatchManager but not any longer required.

I would be glad if you could port it back when merging. I guess is much less overhead for you.

Thank you very much for taking care!

Best regards,
Andreas

PS: As a side note: I find the new code formatting rules a little hard to get used to.
  

# Checklist

- [ ] [Documentation](https://github.com/geoserver/geoserver/tree/main/doc/en/user/source) has been updated (if change is visible to end users). *Not rquired*
- [ ] The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers). *Not rquired*

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/Transaction.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/Transaction.java
@@ -14,7 +14,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.Optional;
+import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.xml.namespace.QName;
@@ -26,12 +26,9 @@ import org.geoserver.platform.GeoServerExtensions;
 import org.geoserver.platform.ServiceException;
 import org.geoserver.wfs.request.Delete;
 import org.geoserver.wfs.request.Insert;
-import org.geoserver.wfs.request.Native;
-import org.geoserver.wfs.request.Replace;
 import org.geoserver.wfs.request.TransactionElement;
 import org.geoserver.wfs.request.TransactionRequest;
 import org.geoserver.wfs.request.TransactionResponse;
-import org.geoserver.wfs.request.Update;
 import org.geotools.data.DefaultTransaction;
 import org.geotools.data.FeatureLockException;
 import org.geotools.data.FeatureSource;
@@ -295,14 +292,8 @@ public class Transaction {
 
         try {
             BatchManager batchManager =
-                    new BatchManager(request, multiplexer, stores, result, elementHandlers);
-            for (Entry<TransactionElement, TransactionElementHandler>
-                    transactionElementTransactionElementHandlerEntry : elementHandlers.entrySet()) {
-                Entry entry = transactionElementTransactionElementHandlerEntry;
-                batchManager.addToBatch(entry);
-            }
-
-            batchManager.executeRemainingBatch();
+                    createBatchManager(request, multiplexer, stores, elementHandlers, result);
+            batchManager.run();
         } catch (WFSTransactionException e) {
             LOGGER.log(Level.SEVERE, "Transaction failed", e);
 
@@ -426,6 +417,24 @@ public class Transaction {
         // we will commit in the writeTo method
         // after user has got the response
         // response = build;
+    }
+
+    /**
+     * @param request
+     * @param multiplexer
+     * @param stores
+     * @param elementHandlers
+     * @param result
+     * @return a new {@link BatchManager} batching INSERT and DELETE operations where possible.
+     */
+    protected BatchManager createBatchManager(
+            TransactionRequest request,
+            TransactionListenerMux multiplexer,
+            Map<QName, FeatureStore> stores,
+            Map<TransactionElement, TransactionElementHandler> elementHandlers,
+            TransactionResponse result) {
+        return new BatchManager(
+                request, multiplexer, stores, result, elementHandlers, DELETE_BATCH_SIZE);
     }
 
     private TransactionRequest fireBeforeTransaction(TransactionRequest request) {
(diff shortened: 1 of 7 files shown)


src/wfs/src/main/java/org/geoserver/wfs/Transaction.java
```diff
@@ -620,181 +629,106 @@ public void dataStoreChange(TransactionEvent event) throws WFSException {
...
-
-        // execute batch inserts then clean the batch
-        private void executeInsertBulkTransaction(
-                Insert insertElement, TransactionElementHandler insertElementHandler) {
-            if (insertElement != null && batchType == BatchType.INSERT) {
-                insertElementHandler.execute(insertElement, request, stores, result, multiplexer);
-                insertElement.setFeatures(new ArrayList());
-                batchType = BatchType.DELETE;
+            if (lRemoveFromRequest) {
+                // contents of the element have been added to target element. To avoid contents
```
aaime, line 722: Fair point, but the current execution model does not spare them from inconsistency anyways.
When runAggregated() is run, the listeners receive a transaction request object (part of TransactionEvent), which is being modified as you go, so each call returns a different state of it.

At the same time, DispatcherCallback can pick the parsed request object either at the leading or trailing edge of execution, and they would get two different views of the request if it gets modified.

If you want to avoid all inconsistencies, then a new TransactionRequest object would have to be built, completely restructured, and then executed at the end (giving a consistent vision to the transaction listeners) while preserving a consistent vision to the dispatcher callbacks.

src/wfs/src/test/java/org/geoserver/wfs/BatchManagerTest.java
```diff
@@ -0,0 +1,261 @@
...
+import org.opengis.feature.simple.SimpleFeature;
+import org.opengis.filter.Filter;
+import org.opengis.filter.Or;
+
+/**
+ * Tests for {@link BatchManager}.
+ *
+ * @author awaterme
+ */
+public class BatchManagerTest {
```
aaime, line 51: In GeoServer we don't have a separate group of testers that run consistent functional tests against the integrated product, everything has to be done by the developers. Can you please add a functional test for this as well, following the same pattern as the previous change, down in TransactionTest?

aaime (changes requested):
Here are some comments.
About the backports, I'm sorry, you'll have to perform it yourself, we can only offer free review time.

src/wfs/src/main/java/org/geoserver/wfs/Transaction.java
```diff
@@ -428,6 +419,24 @@ protected TransactionResponse execute(TransactionRequest request) throws Excepti
...
+     * @param result
+     * @return a new {@link BatchManager} batching INSERT and DELETE operations where possible.
+     */
+    protected BatchManager createBatchManager(
+            TransactionRequest request,
+            TransactionListenerMux multiplexer,
+            Map<QName, FeatureStore> stores,
+            Map<TransactionElement, TransactionElementHandler> elementHandlers,
+            TransactionResponse result) {
+        return new BatchManager(
```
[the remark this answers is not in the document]
awaterme: This was done by intention, because it enables replacing the BatchManager implementation without overriding the entire execute() method, which is very long.
Ok for you?

src/wfs/src/main/java/org/geoserver/wfs/Transaction.java
awaterme: I understand your point, but I would not call a modified or restructured request object *inconsistent* per se.

What I meant with "inconsistent" is that contents logically differ - unintentionally. E.g. previously batch manager *copied* features from the Transactions Insert #2 to Insert #1, *leaving* them in Insert #1 also. 

The current implementation just *moves* the features. But from a logical perspective the request object stays the same.

From my point of view it is not required that the request object stays unmodified for all listeners, etc. It is sufficient if it is logically the same.

Ok?

src/wfs/src/test/java/org/geoserver/wfs/BatchManagerTest.java
awaterme: Ok, I will add a further test case.

awaterme:
Hi Andrea, thank you for taking time to review the PR.
Ok, I will create separate PR for backports.
Best regards,
Andreas

src/wfs/src/main/java/org/geoserver/wfs/Transaction.java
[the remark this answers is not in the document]
aaime: Not really. Transaction was not meant for subclassing, I don't see usage in the mocks either. I've now noticed you also made BatchManager protected. Again, why? Do you have specific needs in a fork of GeoServer?

src/wfs/src/main/java/org/geoserver/wfs/Transaction.java
aaime: Fair enough.

src/wfs/src/test/java/org/geoserver/wfs/BatchManagerTest.java
aaime: Thank you.

src/wfs/src/main/java/org/geoserver/wfs/Transaction.java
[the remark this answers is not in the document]
awaterme: Both is to enable replacing BatchManager with a "No-Op" implementation, if required for TransactionListeners etc, which are not happy about request restructuring. 

Currently this is not required for us, but I am just in the middle of upgrading GeoServer, so I am not 100% sure yet.

I'd suggest to keep the API open. Subclassers are operating on their own risk anyway, but it is making life much easier for adoptions.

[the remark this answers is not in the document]
aaime: Replacing the usage of Transaction with a subclass is not all that easy, one needs to either modify or replace also DefaultWebFeatureService and DefaultWebFeatureService20. But ok, let's keep it. Since you are making it API, give it readable javadoc then.

awaterme:
Hi Andrea,

things are done from my point of view now. I have additionally improved the delete batch size handling and adjusted the tests accordingly.

According to the [release schedule](https://github.com/geoserver/geoserver/wiki/Release-Schedule) bugfix releases for 2.20.x and 2.19.x are planned. So I will create additional PRs for the branches 2.20.x and 2.19.x .
Is that the way to go?

Thank you again!
Andreas

aaime:
The backport to 2.19.x is optional, as there is no planned release for it anymore (the 2.19.5 one has been cut during the weekend). 2.20.x has still a few releases ahead instead. 

Let's see about the builds, I'll kick them again.

aaime:
Squash-merged, when backporting you can just cherry-pick https://github.com/geoserver/geoserver/commit/bc952cb29690faee5a3b92a3f607262340d115ee

PR #5688, opened 2022-02-19
[GEOS-10396] Downgrade jt-jdbc-teradata to unsupported

aaime:
[GEOS-10396](https://osgeo-org.atlassian.net/browse/GEOS-10396)

As per our latest PSC meeting.
Follows up with https://github.com/geotools/geotools/pull/3801

# Checklist


src/web/app/pom.xml
```diff
@@ -735,7 +735,7 @@
       <id>teradata</id>
```
bradh, line 735: Should this profile just be removed?

aaime: Should be fine to keep it, like the other profiles we have for community modules. It allows a developer in the IDE, or anyone building GeoServer from Maven, to include the module in WAR.
Will be removed once the community module is also removed (given the trajectory of this module, probably September 2022?)

dromagnoli (commented):
+1

PR #5693, opened 2022-02-22
Update production recommendations

jodygarnett:
Fixed wording on entity resolution, and then noticed how out of date the entire page was (recommending users edit ``info.xml`` file directly, when control available in web admin)

* [x] Used more specific links where available (example to global settings to change production logging)
* [x] removed recommendation to run two geoserver now that data and service security are available
* [x] fixed wording "XML entity resolution is **used to** obtain any referenced documents"

Many of the recommndations are from much earlier in the project history, before some facilities such as security were introduced.

  

# Checklist

### DIFF
--- a/doc/en/user/source/production/config.rst
+++ b/doc/en/user/source/production/config.rst
@@ -6,9 +6,9 @@ Configuration Considerations
 Use production logging
 ----------------------
 
-Logging may visibly affect the performance of your server. High logging levels are often necessary to track down issues, but by default you should run with low levels.  (You can switch the logging levels while GeoServer is running.)  
+Logging may visibly affect the performance of your server. High logging levels are often necessary to track down issues, but by default you should run with low levels.  (You can switch the logging levels while GeoServer is running.)
 
-You can change the logging level in the :ref:`web_admin`.  You'll want to choose the :guilabel:`PRODUCTION` logging configuration.
+You can change the logging level in the :ref:`config_globalsettings`.  You will want to choose the ``PRODUCTION`` logging configuration, where only problems are written to the log files.
 
 Set a service strategy
 ----------------------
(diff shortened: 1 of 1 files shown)


doc/en/user/source/production/config.rst
```diff
@@ -49,39 +49,44 @@ Make sure clients cannot request an inordinate amount of resources from your ser
...
 
-For extra security, make sure that the connection to the datastore that is open to all is through a user who has read-only permissions. This will eliminate the possibility of a SQL injection (though GeoServer is generally not vulnerable to that sort of attack).
+If you would like some users to be able to modify data, set the service level :guilabel:`Service Level` to ``Transactional`` (or ``Complete``) and use :ref:`security_service` to limit access to the `WFS.Transaction` operation.
+
+If you would like some users to be able to modify some but not all of your data, set the :guilabel:`Service Level` to ``Transactional`` (or ``Complete``), and use :ref:`security_layer` to limit write access to specific layers. Data security can be used to allow write access based on a to a workspace, datastore, or layer security.
```
bradh, line 65: If you would like some users to be able to modify some but not all of your data, set the :guilabel:`Service Level` to ``Transactional`` (or ``Complete``), and use :ref:`security_layer` to limit write access to specific layers. Data security can be used to allow write access based on workspace, datastore, or layer security.

doc/en/user/source/production/config.rst
```diff
@@ -49,39 +49,44 @@ Make sure clients cannot request an inordinate amount of resources from your ser
...
+Caching is also possible for WFS layers, in a very limited fashion. For DataStores that don't have a quick way to determine feature counts (i.e. shapefiles), enabling caching can prevent querying a store twice during a single request. To enable caching, set the Java system property ``org.geoserver.wfs.getfeature.cachelimit`` to a positive integer. Any data sets that are smaller than the cache limit will be cached for the duration of a request, which will prevent the dataset from being queried a second time for the feature count. Note that this may adversely affect some types of DataStores, as it bypasses any feature count optimizations that may exist.
```
bradh, line 74: probably `e.g. shapefiles` rather than `i.e.`.

bradh (commented):
LGTM. I couldn't make the usual github suggestions for easy application, but they're pretty trivial in any case.

PR #5703, opened 2022-02-25
2.19.0, the tile Layer cannot load tiles rightly at some levels

Hsomething:
Hi, I add a new Feature to VectorLayer by WFS's insert API, then I seed tile cache (GeoWebCache) to get new tiles which has new featrure. But , in some leave the tile cannot load rightly.
My geoserver version is 2.19.0
issue1
issue2
issue3
issue4


mprins:
Please open a bug report that includes how to reproduce in Jira or ask questions on the mailing list, see http://geoserver.org/comm/ or http://geoserver.org/issues/

PR #5709, opened 2022-03-02
[GEOS-10405] GetFeatureInfo: Support multiple featureCollections 

awaterme:
Pull request for [GEOS-10405](https://osgeo-org.atlassian.net/browse/GEOS-10405)

Thank you for considering this.

# Checklist

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/featureinfo/GeoJSONTemplateManager.java
+++ b/src/wms/src/main/java/org/geoserver/wms/featureinfo/GeoJSONTemplateManager.java
@@ -14,10 +14,12 @@ import java.io.OutputStreamWriter;
 import java.io.Writer;
 import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Collectors;
 import org.geoserver.platform.GeoServerResourceLoader;
 import org.geoserver.wfs.json.GeoJSONBuilder;
 import org.geoserver.wfs.json.GeoJSONGetFeatureResponse;
 import org.geoserver.wms.GetFeatureInfoRequest;
+import org.geoserver.wms.MapLayerInfo;
 import org.geoserver.wms.WMS;
 import org.geotools.feature.FeatureCollection;
 import org.locationtech.jts.geom.Geometry;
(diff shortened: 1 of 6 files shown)


awaterme:
Some checks failed due to connection issues with the OSGeo maven repo...  :-(

src/wms/src/main/java/org/geoserver/wms/featureinfo/GeoJSONTemplateManager.java
```diff
@@ -73,8 +75,19 @@ protected void handleContent(
             if (content == null) {
                 handleJSONWithoutTemplate(fc, osw);
             } else {
-                String typeName = request.getQueryLayers().get(i).getName();
-                processTemplate(typeName, fc, content, osw);
+                String queryLayerForErrorMsg;
+                List<MapLayerInfo> queryLayers = request.getQueryLayers();
+                if (i < queryLayers.size()) {
```
aaime, line 80: This avoids the exception, but... does not seem the right thing to do, in general.

Some context. You say you have a custom identifier that returns multiple collections. In fact ``LayerIdentifier.identify`` can return multiple ones, and we actually have one use case that does: cascaded WMS servers. 
If we are cascading a WMS, and maybe the remote service is a GeoServer supporting GML FeatureInfo, and we are cascading a layer group, I'd expect to actually get back multiple feature collections.

Say that the cascaded layer is the first, and produces 3 collections, and that then GetFeatureInfo also has local layers too ... it would shift the type names in a way that makes the logging at least confusing, some of the layer names from the remote servers would get local names, some locals would be shifted, the last of them would get the full list of names instead just because we overflowed....

Maybe it's best to use the type name in the collection instead? It's what the template lookup machinery does in the end, calling ``FeatureCollectionDecorator.getName(fc)``. Ideally it would be great to track the actual source layer that originated one particular feature collection, but we don't have such info locally.

aaime (changes requested):

src/wms/src/main/java/org/geoserver/wms/featureinfo/GeoJSONTemplateManager.java
awaterme: Hi Andrea,

thanks for your feedback.

To be honest: I already thought about the scenario with the cascading WMS with multiple collections, but the behaviour is actually different in this case: GeoServer exposes some sort of "combined featureType" having the attributes of all remote types and exploses one collection of those only. The result is not very useful, IMHO. But I did not track that down, as it is a different issue.

Regarding the queryLayer: I am of your opinion, the result might be confusing. In earlier GS versions the type name from the collection was used. I find that more reasonable.

I am happy to change the implementation to log the collection name only. As an alternative GS can additionally always log all query layers, so this information is not getting lost.

What do you think?

Best regards,
Andreas

aaime:
Merged. Do you need to backport?

awaterme:
Hi Andrea,
no backport required.
Thank you!
Andreas

PR #5712, opened 2022-03-02
[GEOS-10409] Improve deletion of WPS Execute input temp files

sikeoka:
[GEOS-10409](https://osgeo-org.atlassian.net/browse/GEOS-10409)

This PR updates the ArcGrid, GeoTIFF and ShapeZIP PPIOs to delete their temp files.  The unit tests do not actually verify that files are being deleted from the system test directory.

  

# Checklist


sikeoka:
I will try to run the build again later.  This unrelated error keeps failing the build
`Failed to execute goal on project gs-dds: Could not resolve dependencies for project org.geoserver.community:gs-dds:jar:2.21-SNAPSHOT: Could not transfer artifact gov.nasa:worldwind:jar:0.6 from/to osgeo-releases (https://repo.osgeo.org/repository/release/): GET request of: gov/nasa/worldwind/0.6/worldwind-0.6.jar from osgeo-releases failed: Premature end of Content-Length delimited message body (expected: 7,016,108; received: 6,143,487)`

sikeoka:
Any thoughts on this @aaime or @jodygarnett?

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ppio/GeoTiffPPIO.java
```diff
@@ -65,26 +65,35 @@
...
-                    "Could not find the GeoTIFF GT2 format, please check it's in the classpath");
+        File file = File.createTempFile("wps", ".tiff", root);
+        GridCoverageReaderResource resource = null;
+        try {
+            FileUtils.copyInputStreamToFile(input, file);
+            GeoTiffFormat format = new GeoTiffFormat();
+            if (!format.accepts(file)) {
+                throw new WPSException("Could not read " + getMimeType() + " coverage");
+            }
+            GeoTiffReader reader = format.getReader(file);
```
aaime, line 87: This may be a behavioral change. The old code would have ended up accepting pretty much anything for which a format is registered, but wondering if now the code can still read an improper geotiff, like, tiff without geotags along with tfw and prj files. The old code would have, using the WorldImageReader. @dromagnoli do you remember if the GeoTIFFReader can read the triplet above?

aaime (commented):
Looks pretty good, I have a hesitation about the GeoTIFF PPIO behavior change.

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ppio/GeoTiffPPIO.java
sikeoka: The WorldImage format actually doesn't work with the current GeoTIFF PPIO because the WorldImage format checks the file extension to determine whether it can read the image but the current temp file has no file extension.  There is a behavior change if coverage formats from some extensions were being used that do not rely on the file name/extension.  This change also causes another behavior change where it was previously possible for the GeoTIFF PPIO to read ArcGrid files so users wouldn't have to explicitly specify the ArcGrid mimetype in the request.  I will change this section back to the original code.

sikeoka: I fixed the GeoTIFF PPIO behavior to be the same as it is now.

sikeoka: I fixed the GeoTIFF PPIO behavior to be the same as it is now.

aaime:
Looks good, thanks! 
Merging, backport as you see fit.

PR #5730, opened 2022-03-18
Dependency upgrade: xstream 1.4.11.1 -> 1.4.19, jettison 1.0.1 -> 1.4.1

groldan:
Upgrading to XStream 1.4.19/Jettison 1.4.1.

Jettison 1.4.1 changes its deafult behavior when
encoding single-element collections as JSON.
While it used to encode such collections as a JSON
object, it now (correctly/consistently) encodes
them as single-element JSON arrays.

Such change would break backwards compatibility
in several GeoServer REST API responses, despite
JSON collection encoding being consistent, we can't
know for sure it wouldn't break exising API clients.

Therefore, measures are taken to preserve the old
behavior where single-element collections are serialized
as a JSON object instead of a single-element JSON array.

For instance:

- `XStreamCatalogListConverter$JSONXStreamListConverter` will
recognize a single-element list, and use a surrogate object to
force serialization preserving Jettison's legacy encoding format.
For example, to get the expected:
`{"coverages": {"coverage": [{ "name": "tazdem", ... }]}}`,
instead of
`{"coverages": {"coverage": { "name": "tazdem", ... }}}`
or
`{"coverages": [{"coverage": { "name": "tazdem", ... }}]}`
if just configuring the `JettisonMappedXmlDriver` with `useSerializeAsArray = false`
or `useSerializeAsArray = true`, respectively.

- A new `ResourceDirectoryInfoJSONConverter` to handle the same issue
with `ResourceController`'s `ResourceDirectoryInfo.getChildren()`.

  

# Checklist

### DIFF
--- a/src/rest/src/main/java/org/geoserver/rest/converters/XStreamJSONMessageConverter.java
+++ b/src/rest/src/main/java/org/geoserver/rest/converters/XStreamJSONMessageConverter.java
@@ -6,9 +6,7 @@ package org.geoserver.rest.converters;
 
 import com.thoughtworks.xstream.XStream;
 import com.thoughtworks.xstream.io.HierarchicalStreamWriter;
-import com.thoughtworks.xstream.io.json.JettisonMappedXmlDriver;
 import java.io.IOException;
-import org.geoserver.config.util.SecureXStream;
 import org.geoserver.config.util.XStreamPersister;
 import org.geoserver.rest.wrapper.RestHttpInputWrapper;
 import org.geoserver.rest.wrapper.RestListWrapper;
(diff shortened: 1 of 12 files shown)


groldan:
QA build failure is unrelated, modules affected by this pr succeeded:

```
09:44:33,512 [INFO] Main Module ........................................ SUCCESS [01:53 min]
09:44:33,513 [INFO] gs-rest ............................................ SUCCESS [ 24.751 s]
09:44:33,513 [INFO] gs-restconfig ...................................... SUCCESS [01:02 min]
09:44:33,514 [INFO] WMTS Multi Dimensional ............................. FAILURE [  1.366 s]
```

src/rest/src/main/java/org/geoserver/rest/converters/XStreamJSONMessageConverter.java
```diff
@@ -99,6 +97,6 @@ public void encodeCollectionLink(String link, HierarchicalStreamWriter writer) {
 
     @Override
     protected XStream createXStreamInstance() {
```
aaime, line 99: An eventual backport should probably deprecate the method.

aaime (approved):

aaime:
Merged

PR #5735, opened 2022-03-22
[GEOS-10427] Improve access check in ImportProcess

sikeoka:
[GEOS-10427](https://osgeo-org.atlassian.net/browse/GEOS-10427)

This PR adds a check to ensure that users executing the Import process have sufficient privileges to modify the target workspace.

  

# Checklist

### DIFF
--- a/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/gs/ImportProcess.java
+++ b/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/gs/ImportProcess.java
@@ -28,8 +28,13 @@ import org.geoserver.catalog.ProjectionPolicy;
 import org.geoserver.catalog.StoreInfo;
 import org.geoserver.catalog.StyleInfo;
 import org.geoserver.catalog.WorkspaceInfo;
+import org.geoserver.platform.GeoServerExtensions;
 import org.geoserver.platform.resource.Paths;
 import org.geoserver.platform.resource.Resource;
+import org.geoserver.security.AccessLevel;
+import org.geoserver.security.SecureCatalogImpl;
+import org.geoserver.security.SecureCatalogImpl.MixedModeBehavior;
+import org.geoserver.security.WrapperPolicy;
 import org.geoserver.wps.WPSException;
 import org.geotools.coverage.grid.GridCoverage2D;
 import org.geotools.coverage.grid.io.AbstractGridFormat;
@@ -66,6 +71,7 @@ import org.opengis.parameter.ParameterValueGroup;
 import org.opengis.referencing.crs.CoordinateReferenceSystem;
 import org.opengis.referencing.operation.MathTransform;
 import org.opengis.util.ProgressListener;
+import org.springframework.security.core.context.SecurityContextHolder;
 import org.vfny.geoserver.util.WCSUtils;
 
 /**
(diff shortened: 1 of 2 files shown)


aaime:
Service code should not be mixed with security, it should happen somewhere else... the likely place is the secured catalog wrapper, I'd guess.

sikeoka:
I need more information about what you want.  I was trying to avoid modifying SecureCatalogImpl just to fix an issue in one small part of an extension module.

I suppose I could move the check to org.geoserver.security.SecureCatalogImpl.add(StoreInfo) and org.geoserver.security.SecureCatalogImpl.save(StoreInfo) and see if anything breaks:
https://github.com/geoserver/geoserver/blob/main/src/main/src/main/java/org/geoserver/security/SecureCatalogImpl.java#L1302
https://github.com/geoserver/geoserver/blob/main/src/main/src/main/java/org/geoserver/security/SecureCatalogImpl.java#L1521

There are at least wo other classes that directly access SecureCatalogImpl to check permissions:
https://github.com/geoserver/geoserver/tree/main/src/web/core/src/main/java/org/geoserver/web/WorkspaceAdminComponentAuthorizer.java
https://github.com/geoserver/geoserver/tree/main/src/community/taskmanager/core/src/main/java/org/geoserver/taskmanager/util/TaskManagerSecurityUtil.java

aaime:
The WorkspaceAdminComponentAuthorizer is effectively part of the security subsystem, it's there to make sure the GUI does not list pages/components that the user cannot see (rather than getting exceptions trying to access them). It's properly factored out, no page mocks with security besides declaring (via its base class) if it's an admin page, or not.

The community module is instead breaking the pattern, shouldn't do it. But as a community module, anything goes.

aaime:
Taking a note that this is a good topic for the next PSC meeting, happening tomorrow, 6:30 CET, on Jitsi: https://meet.jit.si/GeoServerMeeting

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/gs/ImportProcess.java
```diff
@@ -247,6 +254,21 @@ public String execute(
...
+        // This is essentially the same code to check the access level as used by
+        // org.geoserver.security.SecureCatalogImpl.getDataStoreByName(String, String)
+        WrapperPolicy policy =
+                GeoServerExtensions.bean(SecureCatalogImpl.class)
+                        .buildWrapperPolicy(
+                                SecurityContextHolder.getContext().getAuthentication(),
+                                workspace,
+                                workspace.getName(),
+                                MixedModeBehavior.CHALLENGE);
+        if (policy.level != AccessLevel.READ_WRITE) {
```
aaime, line 267: As per our meeting today, moving the checks to the SecureCatalogImpl would be architecturally better, but would open a can of works vs [the REST security](https://docs.geoserver.org/maintain/en/user/security/rest.html).

So ok to do it here, but it should check whether the current user has admin access on the current workspace, something like ``((WorkspaceAccessLimits) policy.getAccessLimits()).isAdminable()``.
Which is probably similar, in practice, to what you're doing, but clearer and we hope more future proof.

sikeoka: I updated it to check for admin access instead of write access.

aaime:
Merged. Backport as you see fit.

PR #5767, opened 2022-04-01
[2.20.x Backport][GEOS-10418] Bad request sent to GeoFence when matching roles only

shadab-geo:
[GEOS-10418](https://osgeo-org.atlassian.net/browse/GEOS-10418)

  

This is backport of [5756](https://github.com/geoserver/geoserver/pull/5756) for 2.20.x

# Checklist

### DIFF
--- a/src/extension/geofence/src/main/java/org/geoserver/geofence/GeofenceAccessManager.java
+++ b/src/extension/geofence/src/main/java/org/geoserver/geofence/GeofenceAccessManager.java
@@ -575,6 +575,8 @@ public class GeofenceAccessManager
                 for (GrantedAuthority authority : user.getAuthorities()) {
                     if (config.getRoles().contains(authority.getAuthority())) {
                         role = authority.getAuthority();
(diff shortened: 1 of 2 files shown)


etj:
@taba90 pls reopen and fix this one.
The user should be set also if no role is matching.

src/extension/geofence/src/main/java/org/geoserver/geofence/GeofenceAccessManager.java
```diff
@@ -575,6 +575,8 @@ private void setRuleFilterUserOrRole(Authentication user, RuleFilter ruleFilter)
                 for (GrantedAuthority authority : user.getAuthorities()) {
                     if (config.getRoles().contains(authority.getAuthority())) {
                         role = authority.getAuthority();
+                        ruleFilter.setUser(RuleFilter.SpecialFilterType.DEFAULT);
```
etj, line 578: This `setUser` should be run everytime the condition `config.isUseRolesToFilter()` in the previous if block is true.
As it is now, it is only executed when a role is found.
Please move this line beside the setRole call at the end of the `if` block starting at line 572

PR #5775, opened 2022-04-03
[GEOS-10442] Upgrade or remove jasper dependencies

bradh:
[GEOS-10442](https://osgeo-org.atlassian.net/browse/GEOS-10442)

This removes the obsolete jasper dependency.

That is an old tomcat entity for JSP support, which we don't appear to need. It was packaged in the bin (jetty) release package, and I have verified that it the -bin package is still OK without it.

# Checklist

- [N/A] New unit tests have been added covering the changes.
- [N/A] [Documentation](https://github.com/geoserver/geoserver/tree/main/doc/en/user/source) has been updated (if change is visible to end users).
- [N/A] The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers).

This is an alternative approach to the upgrade that I originally proposed in https://github.com/geoserver/geoserver/pull/5768

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -1053,29 +1053,8 @@
       </dependency>
       <dependency>
         <groupId>org.eclipse.jetty</groupId>
-        <artifactId>jsp-2.0</artifactId>
-        <version>${jetty.version}</version>
-        <type>pom</type>
-      </dependency>
-      <dependency>
-        <groupId>org.eclipse.jetty</groupId>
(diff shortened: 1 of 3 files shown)


mprins (approved):
I agree that these are unnecessary.

tried/tested with `mvn jetty:run` src/web/app and the bin-zip package using `$GEOSERVER_HOME/bin/startup.sh`

PR #5781, opened 2022-04-04
[GEOS-10444]: Supporting REPLACE UpdateMode on rasters import

dromagnoli:
[GEOS-10444](https://osgeo-org.atlassian.net/browse/GEOS-10444)

Importer: Add support for UpdateMode.REPLACE  on raster too

# Checklist

- x ] There is an issue in the [GeoServer Jira](https://osgeo-org.atlassian.net/browse/GEOS/summary) (except for changes that do not affect administrators or end users in any way).

### DIFF
--- a/doc/en/user/source/extensions/importer/rest_reference.rst
+++ b/doc/en/user/source/extensions/importer/rest_reference.rst
@@ -447,9 +447,19 @@ Updating a task
 A PUT request over an existing task can be used to update its representation. The representation can be partial, and just contains
 the elements that need to be updated.
 
-The updateMode of a task normally starts as "CREATE", that is, create the target resource if missing. Other possible values are
-"REPLACE", that is, delete the existing features in the target layer and replace them with the task source ones, or "APPEND",
-to just add the features from the task source into an existing layer.
+The updateMode of a task may have different values. 
+
+.. list-table::
+   :header-rows: 1
+
+   * - UpdateMode
+     - Description
+   * - CREATE
+     - This is the default starting updateMode of a task, that is: create the target resource if missing.
+   * - REPLACE
+     - For vector stores: delete the existing features in the target layer and replace them with those from the task source. For raster stores: replace the underlying data. When dealing with StructuredGridCoverage reader (e.g. ImageMosaic) the new file will be harvested (replacing the old one). For single raster coverages (e.g. GeoTIFF) the name of the file should be the same so that the coverageStore layer will preserve the original name (the old file will be deleted).
+   * - APPEND
+     - Add the features from the task source into an existing layer.
 
 The following PUT request updates a task from "CREATE" to "APPEND" mode::
 
--- a/src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
+++ b/src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
@@ -12,6 +12,7 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -25,6 +26,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
 import org.geoserver.catalog.Catalog;
 import org.geoserver.catalog.CatalogBuilder;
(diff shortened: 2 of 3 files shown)


doc/en/user/source/extensions/importer/rest_reference.rst
```diff
@@ -447,9 +447,19 @@ Updating a task
...
+
+.. list-table::
+   :header-rows: 1
+
+   * - UpdateMode
+     - Description
+   * - CREATE
+     - This is the default starting updateMode of a task, that is: create the target resource if missing.
+   * - REPLACE
+     - For vector stores: delete the existing features in the target layer and replace them with the task source ones. For raster stores: replace the underlying data. When dealing with StructuredGridCoverage reader (i.e. ImageMosaic) the new file will get harvested (replacing the old one). For single raster coverages (i.e. GeoTIFF) the name of the file should be the same so that the coverageStore layer will preserve the original name (The old file will be deleted).
```
bradh, line 460:
```suggestion
     - For vector stores: delete the existing features in the target layer and replace them with those from the task source. For raster stores: replace the underlying data. When dealing with StructuredGridCoverage reader (e.g. ImageMosaic) the new file will be harvested (replacing the old one). For single raster coverages (e.g. GeoTIFF) the name of the file should be the same so that the coverageStore layer will preserve the original name (the old file will be deleted).
```

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -1494,6 +1488,125 @@ void loadIntoDataStore(
...
+        String nativeName = originalCoverage.getNativeName();
+
+        // Marking previous file for eventual delete
+        GridCoverage2DReader oldReader = ((GridCoverage2DReader) reader);
+        ServiceInfo readerInfo = oldReader.getInfo();
+        String oldFile = readerInfo.getSource().toURL().getFile();
+
+        // Setting up the new coverage
+        CatalogRepository repository = catalog.getResourcePool().getRepository();
+        Hints hints = new Hints(new RenderingHints(Hints.REPOSITORY, repository));
```
aaime, line 1556: These are nonstructured readers, why do we need to pass in the repository?

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -1494,6 +1488,125 @@ void loadIntoDataStore(
...
+        OwsUtils.copy(coverage, originalCoverage, CoverageInfo.class);
+
+        // Restore original names
+        originalCoverage.setName(name);
+        originalCoverage.setNativeName(nativeName);
+        catalog.save(originalCoverage);
+        catalog.save(store);
+
+        // Clean up the oldFile when not overwriting it
+        if (!newFile.equalsIgnoreCase(oldFile)) {
```
aaime, line 1580: Hum... maybe the "single file" thing has been taken a bit too far here.
AFAIK one can update also a WorldImage case, made of several files, and also the "single file" GeoTIFF might have external overviews, CRS forcing prjs, maybe mask files, and so on.

I think the could should work in terms of "set of files before" and "set of files after", where the set of files can be gathered using ``GridCoverage2DReader.getInfo(coverage)``.

src/extension/importer/core/src/test/java/org/geoserver/importer/ImporterDataTest.java
```diff
@@ -650,6 +651,34 @@ public void testImportGeoTIFFFromDataDir() throws Exception {
                 ((CoverageStoreInfo) task.getLayer().getResource().getStore()).getURL());
 
         runChecks("EmissiveCampania");
+        CoverageStoreInfo store = cat.getCoverageStoreByName("EmissiveCampania");
+
+        // Replacing the layer
+        File tempDir = FileUtils.getTempDirectory();
+        dir = unpack("geotiff/EmissiveCampania.tif.bz2", tempDir);
```
aaime, line 658: If it's the same source file, how do you know the replace actually worked?

aaime (changes requested):

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
dromagnoli: good catch, I didn't properly clean up the code while checking the catalogBuilder usage. Fixing

src/extension/importer/core/src/test/java/org/geoserver/importer/ImporterDataTest.java
dromagnoli: I'm actually updating the same file but from a different path and the storeURL get properly changed.

src/extension/importer/core/src/main/java/org/geoserver/importer/Importer.java
```diff
@@ -1494,6 +1488,156 @@ void loadIntoDataStore(
...
+        catalog.save(originalCoverage);
+        catalog.save(store);
+
+        // Clean up the oldFile when not overwriting it
+        cleanUp(oldReader, newReader);
+    }
+
+    private void cleanUp(GridCoverage2DReader oldReader, GridCoverage2DReader newReader)
+            throws IOException {
+        List<File> oldFiles = getFiles(oldReader);
```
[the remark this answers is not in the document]
dromagnoli: mmm. I don't think so. Do you have any particular bad-case in mind?
The replace occurs on the Catalog.CoverageStore URL whilst the reader is a GeoTools reader returning info about what has been set as input.

Anyway, I have updated the code to collect oldFiles and newFiles at the beginning and do the cleanup at the end

aaime (approved):

aaime:
Looks good, thanks!

PR #5790, opened 2022-04-08
[GEOS-10454] Make LanguageUrl mangler using rawKvp instead of HttpServletRequest params

fostash:
[GEOS-10454](https://osgeo-org.atlassian.net/browse/GEOS-10454)

[GEOS-10454] use raw kv params detached from servlet request because of tomcat request lifecycle that cause the reset of a finished request

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/ows/LanguageURLMangler.java
+++ b/src/main/src/main/java/org/geoserver/ows/LanguageURLMangler.java
@@ -4,10 +4,9 @@
  */
 package org.geoserver.ows;
 
-import java.util.Collections;
-import java.util.Enumeration;
+import java.util.HashMap;
 import java.util.Map;
-import javax.servlet.http.HttpServletRequest;
+import java.util.Optional;
 
 /** Mangles service URL's based on the AcceptLanguages and Languages parameters */
 public class LanguageURLMangler implements URLMangler {
@@ -20,24 +19,23 @@ public class LanguageURLMangler implements URLMangler {
             StringBuilder baseURL, StringBuilder path, Map<String, String> kvp, URLType type) {
         if (Dispatcher.REQUEST.get() == null || !type.equals(URLType.SERVICE)) return;
 
-        String languageParameter = null;
-        String acceptLanguagesParameter = null;
+        final Map<String, Object> rawKvp =
+                Optional.ofNullable(Dispatcher.REQUEST.get().rawKvp).orElse(new HashMap<>());
 
-        HttpServletRequest httpRequest = Dispatcher.REQUEST.get().getHttpRequest();
-        Enumeration<String> parameterNames = httpRequest.getParameterNames();
+        final Optional<String> languageParameter =
+                rawKvp.keySet().stream()
+                        .filter(param -> param.equalsIgnoreCase(LANGUAGE))
+                        .findFirst();
 
-        for (String parameter : Collections.list(parameterNames)) {
-            if (parameter.equalsIgnoreCase(LANGUAGE)) {
-                languageParameter = parameter;
-            } else if (parameter.equalsIgnoreCase(ACCEPT_LANGUAGES)) {
-                acceptLanguagesParameter = parameter;
-            }
-        }
+        final Optional<String> acceptLanguagesParameter =
+                rawKvp.keySet().stream()
+                        .filter(param -> param.equalsIgnoreCase(ACCEPT_LANGUAGES))
+                        .findFirst();
 
         String language = "";
 
-        if (acceptLanguagesParameter != null) {
-            String acceptLanguages = httpRequest.getParameter(acceptLanguagesParameter);
+        if (acceptLanguagesParameter.isPresent()) {
+            String acceptLanguages = (String) rawKvp.get(acceptLanguagesParameter.get());
             if (acceptLanguages != null) {
                 String commaSplit = acceptLanguages.split(",")[0];
                 String spaceSplit = commaSplit.split(" ")[0];
@@ -49,8 +47,8 @@ public class LanguageURLMangler implements URLMangler {
             }
         }
 
-        if (languageParameter != null) {
-            language = httpRequest.getParameter(languageParameter);
+        if (languageParameter.isPresent()) {
+            language = (String) rawKvp.get(languageParameter.get());
         }
 
         if (language != null && language.length() > 0) {
--- a/src/main/src/main/java/org/geoserver/ows/LanguageURLMangler.java
+++ b/src/main/src/main/java/org/geoserver/ows/LanguageURLMangler.java
@@ -4,45 +4,72 @@
  */
 package org.geoserver.ows;
 
+import java.util.Arrays;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


taba90:
looks good to me. Could you only replace the current commit message using the titile instead of the description from the Jira ticket? thanks!

src/main/src/main/java/org/geoserver/ows/LanguageURLMangler.java
```diff
@@ -20,24 +19,23 @@ public void mangleURL(
...
-        String acceptLanguagesParameter = null;
+        final Map<String, Object> rawKvp =
+                Optional.ofNullable(Dispatcher.REQUEST.get().rawKvp).orElse(new HashMap<>());
 
-        HttpServletRequest httpRequest = Dispatcher.REQUEST.get().getHttpRequest();
-        Enumeration<String> parameterNames = httpRequest.getParameterNames();
+        final Optional<String> languageParameter =
+                rawKvp.keySet().stream()
+                        .filter(param -> param.equalsIgnoreCase(LANGUAGE))
+                        .findFirst();
```
taba90, line 28: the raw kvp map is an HasMap implementation that is already ignore cased https://github.com/geoserver/geoserver/blob/9e40cbbccb9626b8d08516c8eacaf5a8f168189b/src/ows/src/main/java/org/geoserver/ows/util/KvpMap.java#L19. We can directly use the contains method of it without iterating before.

src/main/src/main/java/org/geoserver/ows/LanguageURLMangler.java
```diff
@@ -20,24 +19,23 @@ public void mangleURL(
...
 
-        for (String parameter : Collections.list(parameterNames)) {
-            if (parameter.equalsIgnoreCase(LANGUAGE)) {
-                languageParameter = parameter;
-            } else if (parameter.equalsIgnoreCase(ACCEPT_LANGUAGES)) {
-                acceptLanguagesParameter = parameter;
-            }
-        }
+        final Optional<String> acceptLanguagesParameter =
```
taba90, line 30: same as before

taba90 (changes requested):
Looks good only a minor change requested.

src/main/src/main/java/org/geoserver/ows/LanguageURLMangler.java
```diff
@@ -20,24 +19,13 @@ public void mangleURL(
...
-            }
-        }
+        final Map<String, Object> rawKvp =
+                Optional.ofNullable(Dispatcher.REQUEST.get().rawKvp).orElse(new KvpMap<>());
 
         String language = "";
 
-        if (acceptLanguagesParameter != null) {
-            String acceptLanguages = httpRequest.getParameter(acceptLanguagesParameter);
+        if (rawKvp.containsKey(ACCEPT_LANGUAGES)) {
```
nmco, line 27: I think this if is redundant.

src/main/src/main/java/org/geoserver/ows/LanguageURLMangler.java
```diff
@@ -20,24 +19,13 @@ public void mangleURL(
...
+            String acceptLanguages = (String) rawKvp.get(ACCEPT_LANGUAGES);
             if (acceptLanguages != null) {
```
nmco, line 29: I would check here if the string is not empty, or even if the string as a comma.

nmco:
Thank you @fostash, looks good to me, I will merge when the build is green.

src/main/src/main/java/org/geoserver/ows/LanguageURLMangler.java
```diff
@@ -4,57 +4,72 @@
...
-        }
-
-        if (languageParameter != null) {
-            language = httpRequest.getParameter(languageParameter);
-        }
-
-        if (language != null && language.length() > 0) {
-            kvp.put("Language", language.trim());
-        }
+        Optional.ofNullable(type)
```
aaime, line 32: Careful about this proliferation of Optional calls, as they come with object allocations, one for the main Optional, plus others for specific methods (e.g. map, flatmap).

In this code, it's probably not harmful, a single request processing normally builds up to a few tens of URLs, so overall the code change introduces the allocation of the order of 100 (ish) new objects that weren't there in the old code. In the economy of a OGC request, that's peanuts.

However, always consider the context in which the code is running, if it's called many times (hundreds of thousands, millions, not unusual when dealing with GIS data) best to stay away from optionals and streams due to their memory and execution costs.

fostash: I perfectly got your point, and here the `type` parameter wrapped in an Optional is for sure useless. I will revert it with the if.

aaime:
Legit formatting check failure in the QA build:

```
Error: 7,275 [ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.20.0:check (default) on project gs-main: The following files had format violations:
Error: 7,275 [ERROR]     src/main/java/org/geoserver/ows/LanguageURLMangler.java
Error: 7,275 [ERROR]         @@ -37,8 +37,8 @@
Error: 7,275 [ERROR]          
Error: 7,275 [ERROR]          ····/**
Error: 7,275 [ERROR]          ·····*·@return·Getting·the·rawKvp·from·the·Request·inside·the·thread·local·search·for·parameter
Error: 7,275 [ERROR]         -·····*·Language·or·AcceptLanguages·and·return·the·first·one·if·present·or·the·first·occurrence
Error: 7,275 [ERROR]         -·····*·of·a·language·code·(if·present)·inside·the·second·one
Error: 7,275 [ERROR]         +·····*·····Language·or·AcceptLanguages·and·return·the·first·one·if·present·or·the·first·occurrence
Error: 7,275 [ERROR]         +·····*·····of·a·language·code·(if·present)·inside·the·second·one
Error: 7,275 [ERROR]          ·····*/
Error: 7,275 [ERROR]          ····protected·Optional<Object>·processLanguageParam()·{
Error: 7,275 [ERROR]          ········return·Optional.ofNullable(Dispatcher.REQUEST.get())
Error: 7,275 [ERROR]         @@ -50,7 +50,7 @@
Error: 7,275 [ERROR]          ····/**
Error: 7,275 [ERROR]          ·····*·@param·rawKvp·Map·of·request·parameter
Error: 7,275 [ERROR]          ·····*·@return·the·value·of·Language·parameter·inside·the·{@link·Request#rawKvp}·map·or·an
Error: 7,275 [ERROR]         -·····*·Optional.empty
Error: 7,275 [ERROR]         +·····*·····Optional.empty
Error: 7,275 [ERROR]          ·····*/
Error: 7,275 [ERROR]          ····protected·Optional<Object>·getLanguage(Map<String,·Object>·rawKvp)·{
Error: 7,275 [ERROR]          ········return·Optional.ofNullable(rawKvp.get(LANGUAGE)).filter(filterEmtpyString);
Error: 7,275 [ERROR] Run 'mvn spotless:apply' to fix these violations.
```

Just run a maven build before committing, it will reformat the code for you.

PR #5772, opened 2022-04-01
[GEOS-10445] Bump spring.version from 5.1.20.RELEASE to 5.2.20.RELEASE

groldan:
[GEOS-10445](https://osgeo-org.atlassian.net/browse/GEOS-10445)

Build errors on `mvn verify -Prelease`:

---

* gs-gwc's `RESTIntegrationTest.testGetLayersList()`:

`junit.framework.AssertionFailedError: Expecting to find matches for Xpath //layers/layer/atom:link[@href = 'http://localhost:8080/geoserver/gwc/rest/layers/cdf:Deletes.xml']`

response:

```xml
<?xml version="1.0" encoding="UTF-8"?><layers>
 ...

  <layer>
    <name>cdf:Deletes</name>
    <atom:link xmlns:atom="http://www.w3.org/2005/Atom" href="http://localhost:8080/geoserver/gwc//rest/layers/cdf:Deletes.xml" rel="alternate" type="text/xml"/>
  </layer>
</layers>
```
expected vs actual:
```
http://localhost:8080/geoserver/gwc/rest/layers/cdf:Deletes.xml
http://localhost:8080/geoserver/gwc//rest/layers/cdf:Deletes.xml
```

Fix: https://github.com/GeoWebCache/geowebcache/pull/1039

---
* gs-ogcapi

```
Error: 4,875 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project gs-ogcapi-core: Compilation failure
Error: 4,875 [ERROR] /home/runner/work/geoserver/geoserver/src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/APIDispatcher.java:[134,46] method requestMappingHandlerAdapter in class org.springframework.web.servlet.config.annotation.WebMvcConfigurationSupport cannot be applied to given types;
Error: 4,875 [ERROR]   required: org.springframework.web.accept.ContentNegotiationManager,org.springframework.format.support.FormattingConversionService,org.springframework.validation.Validator
Error: 4,875 [ERROR]   found: no arguments
Error: 4,875 [ERROR]   reason: actual and formal argument lists differ in length
```
Fix: 16bf3ed5bf101a4dab4c99ac0be68c29ffdd6d3d (@aaime please verify that's correct, or suggest an alternative. Idon't know the module)

---

* `StyleControllerTest.testPutAsSLDWithExtension()`: POST request to /rest/styles/Ponds.sld
[...]

PR #5791, opened 2022-04-08
STAC tests fixes and NPE avoidance vs STAC global queryables

aaime:
First off, apologies for mixing two concerns in one PR, I went in to fix a NPE and found most of STAC tests failing.

The bulk of the failures were due to new test data added in https://github.com/geoserver/geoserver/pull/5729 , along with a NPE that was happening when trying to get queryables, either global or per collection, in the STAC API. @turingtestfail can you review the changes in the first commit? They address those issues.

An interesting set of failures was instead due to the upgrade of Spring to 5.2.20, in particular [to this one liner](https://github.com/geoserver/geoserver/pull/5772/files#diff-e2a648d6b8dcdcbc9402cf852d0073bd9953c45cad49c28eb0026f241e799177R134). The way Spring MVC wires up changed and that caused, along with the one-liner change, the handlerAdapter not to have a conversion service anymore, which resulted in inability to do most of the common conversions on request parameters, including going from comma separed list of values to lists or arrays of value (but also SortBy usage got broken).
@groldan can you have a look at the change I've made in the second commit? It seems to work and makes the handler be wired correctly, but for the life of me I cannot get APIDispatcherTest to work anymore... @groldan do you have any clue on how to get things back working there?

# Checklist


turingtestfail:
@aaime Reviewed first commit and it looks good to me.

src/community/oseo/oseo-core/src/test/java/org/geoserver/opensearch/eo/store/GeoServerOpenSearchTestSupport.java
```diff
@@ -49,8 +50,7 @@ public static void setupBasicOpenSearch(
         // configure opensearch for EO to use it
         OSEOInfo service = gs.getService(OSEOInfo.class);
         service.setOpenSearchAccessStoreId(osDs.getId());
-        String queryables = "id,geometry,collection";
-        service.setGlobalQueryables(queryables);
+        service.getGlobalQueryables().addAll(Arrays.asList("id", "goemetry", "collection"));
```
torse, line 53: Looks like a typo: `goemetry` -> `geometry`

PR #5794, opened 2022-04-10
Documentation - Add precision on community modules download availability

AlexGacon:
Following several questions and issues raised by the community about missing dependencies when working with community modules and stable versions of GeoServer, try to make the documentation more explciit.

# Checklist

### DIFF
--- a/doc/en/user/source/community/index.rst
+++ b/doc/en/user/source/community/index.rst
@@ -11,6 +11,12 @@ officially part of the GeoServer releases. They are however built along with the
 
    Community modules are generally considered experimental in nature and are often under constant development. For that reason documentation in this section should not be considered solid or final and will be subject to change.
 
+.. note::
+
+   Unlike the official extensions, these modules are not released and stored on SourceForge when an official GeoServer release is produced.
+   As a consequence, using a module built against the nightly build of GeoServer with an official release won't probably work.
+
+   If you need a community module for an official release, the only way is to build it with the sources of the GitHub tag matching the release.
 
 .. please keep the following list sorted alphabetically 
 


doc/en/user/source/community/index.rst
```diff
@@ -11,6 +11,12 @@ officially part of the GeoServer releases. They are however built along with the
 
    Community modules are generally considered experimental in nature and are often under constant development. For that reason documentation in this section should not be considered solid or final and will be subject to change.
 
+.. note::
+
+   Unlike the official extensions, these modules are not released and stored on SourceForge when an official GeoServer release is produced.
+   As a consequence, using a module built against the nightly build of GeoServer with an official release won't probably work.
```
aaime, line 17: IMHO wording is too strong... in fact they often work, but one cannot take that for granted. I'd go for "may not work due to dependency or code compatibility".

Then there is the other fact, that community modules have a very variable state of maintenance... some won't work period, not matter the match, some are semi-maintained only on the developer branch (OGC API is one of such cases), while others are periodically updated to work also on stable (often reacting to customer demand).

It is also important to understand that tests and QA procedures are not run on the community modules, breakages happen without developers even noticing: that's by design, if a module is important enough to a dev then it will be pushed to extension. "important" often equates to "I have enough business contracts going with it to justify the cost of being a maintainer for it".

AlexGacon: You are probably right : the "wont' work" is perhaps too strong.

But I think that the first sentence is important because personnaly until this week I wasn't aware of that : perhaps a comparison table between extension and modules to explain the differences could be a good complement.

Topics to cover in this table:
- Automatic tests
- Automatic QA procedures
- Regular synchronization with Transifexe
- Official release
- Maintained by

PR #5797, opened 2022-04-11
[GEOS-1906] DataAccessLimits serialization issue

fostash:
[GEOS-1906](https://osgeo-org.atlassian.net/browse/GEOS-1906)

Fix serialization of filters when access on Layer configuration page
  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/DataAccessLimits.java
+++ b/src/main/src/main/java/org/geoserver/security/DataAccessLimits.java
@@ -5,12 +5,12 @@
  */
 package org.geoserver.security;
 
+import org.geotools.filter.text.ecql.ECQL;
+import org.opengis.filter.Filter;
+
 import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
-import java.util.Collections;
-import org.geotools.filter.text.ecql.ECQL;
-import org.opengis.filter.Filter;
 
 /**
  * Base class for all AccessLimits declared by a {@link ResourceAccessManager}
@@ -64,12 +64,12 @@ public class DataAccessLimits extends AccessLimits {
     }
 
     /**
-     * Writes the non Serializable Filter object ot the ObjectOutputStream via a OGC Filter XML
+     * Writes the non Serializable Filter object to the ObjectOutputStream using ECQL
      * encoding conversion
      */
     protected void writeFilter(Filter filter, ObjectOutputStream out) throws IOException {
         if (filter != null) {
-            out.writeObject(ECQL.toCQL(Collections.singletonList(filter)));
+            out.writeObject(ECQL.toCQL(filter));
         } else {
             out.writeObject(null);
         }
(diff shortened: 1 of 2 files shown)


src/main/src/main/java/org/geoserver/security/DataAccessLimits.java
```diff
@@ -74,11 +69,7 @@ private void writeObject(ObjectOutputStream out) throws IOException {
      */
     protected void writeFilter(Filter filter, ObjectOutputStream out) throws IOException {
         if (filter != null) {
-            try (ByteArrayOutputStream bos = new ByteArrayOutputStream()) {
-                Encoder encoder = new Encoder(CONFIGURATION);
-                encoder.encode(filter, OGC.Filter, bos);
-                out.writeObject(bos.toByteArray());
-            }
+            out.writeObject(ECQL.toCQL(Collections.singletonList(filter)));
```
aaime, line 72: CQL is known to have round-tripping issues, you might serialize one filter and get back another, that is functionally equivalent, but not syntactically the same.

mprins (changes requested):
this Jira [GEOS-1906](https://osgeo-org.atlassian.net/browse/GEOS-1906)  does not at all seem related to this commit

src/main/src/main/java/org/geoserver/security/DataAccessLimits.java
fostash: I changed it using the other method that accept a single Filter. is that what you was pointing to? or I should not use the `toCQL` method?

aaime: Best not to use CQL for filter serialization at all, due to the round-tripping issues I was referring to above.

src/main/src/main/java/org/geoserver/security/DataAccessLimits.java
```diff
@@ -64,12 +69,20 @@ private void writeObject(ObjectOutputStream out) throws IOException {
     }
 
     /**
-     * Writes the non Serializable Filter object to the ObjectOutputStream using ECQL
+     * Writes the non Serializable Filter object ot the ObjectOutputStream via a OGC Filter XML
```
bradh, line 72: typo introduced.

fostash:
you're right. I will change the commit ones I have the correct JIRA ticket

src/main/src/main/java/org/geoserver/security/DataAccessLimits.java
```diff
@@ -69,15 +69,19 @@ private void writeObject(ObjectOutputStream out) throws IOException {
...
-                encoder.encode(filter, OGC.Filter, bos);
-                out.writeObject(bos.toByteArray());
+            if (filter != Filter.INCLUDE && filter != Filter.EXCLUDE) {
+                try (ByteArrayOutputStream bos = new ByteArrayOutputStream()) {
+                    Encoder encoder = new Encoder(CONFIGURATION);
+                    encoder.encode(filter, OGC.Filter, bos);
+                    out.writeObject(bos.toByteArray());
+                }
+            } else {
+                out.writeObject(filter);
```
aaime, line 84: Well well, I did not know INCLUDE and EXCLUDE were serializable... nice find.

aaime (approved):

aaime:
I've created a Jira ticket for you, https://osgeo-org.atlassian.net/browse/GEOS-10470, and squash merged.
Backport as you see fit.

PR #5811, opened 2022-04-14
[GEOS-10463]: Support for Default Deflate compression level customization

dromagnoli:
[GEOS-10463](https://osgeo-org.atlassian.net/browse/GEOS-10463)

# Checklist


src/wcs/src/main/java/org/geoserver/wcs/WCSInfoImpl.java
```diff
@@ -30,6 +30,8 @@ public class WCSInfoImpl extends ServiceInfoImpl implements WCSInfo {
 
     long maxOutputMemory = -1;
 
+    int defaultDeflateCompressionLevel = WCSInfo.DEFAULT_DEFLATE_COMPRESSION_LEVEL;
```
aaime, line 33: I'm afraid that might not work, due to the way XStream builds objects form XML (skipping the normal construction path). 

Better to:
* Make it an Integer (just the field)
* Deal with eventual nulls in the getter, defaulting to the desired compression level.
* Update hashcode/tostring as a consequence

src/web/wcs/src/main/java/org/geoserver/wcs/web/WCSAdminPage.java
```diff
@@ -67,6 +67,10 @@ protected void build(IModel info, Form form) {
         TextField<Integer> maxOutputMemory = new TextField<>("maxOutputMemory");
         maxOutputMemory.add(RangeValidator.minimum(0l));
         form.add(maxOutputMemory);
+        TextField<Integer> defaultDeflateCompressionLevel =
+                new TextField<>("defaultDeflateCompressionLevel");
+        defaultDeflateCompressionLevel.add(RangeValidator.range(1, 9));
+        form.add(defaultDeflateCompressionLevel);
```
aaime, line 73: Maybe extend WCSAdminPageTest to cover the new field and its validation?

aaime (changes requested):

dromagnoli:
@aaime feedbacks addressed

aaime (approved):

PR #5815, opened 2022-04-15
Update pom.xml to use CI-friendly versions through ${revision} property

groldan:
The project version is specified only in the root pom
as the "revision" property.

All child modules specify the parent's version as ${revision}.

This allows to change a single line in the root pom when
the project version has to be changed, instead of having
to modify all pom.xml files.

It also allows to specify the version is as command-line
argument (e.g. mvn -Drevision=2.21.0 clean install).

The flatten-maven-plugin resolves the revision variable and
installs all pom.xml files in the local maven repository with
the resolved value.

The sortpom-maven-plugin configuration is updated to set:
<pomFile>pom.xml</pomFile>
in order to format the actual pom instead of the intermediary
.flattened-pom.xml files created by flatten-maven-plugin.

See https://maven.apache.org/maven-ci-friendly.html

  

# Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,5 @@ src/release/installer/win/target/
 # spotless up to date checks
 .spotless-index
 
+#flatten-maven-plugin flattened pom files
+.flattened-pom.xml
--- a/src/community/app-schema/pom.xml
+++ b/src/community/app-schema/pom.xml
@@ -12,7 +12,7 @@
     <parent>
         <groupId>org.geoserver</groupId>
         <artifactId>community</artifactId>
-        <version>2.21-SNAPSHOT</version>
+        <version>${revision}</version>
     </parent>
 
     <groupId>org.geoserver.community</groupId>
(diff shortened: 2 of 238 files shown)


aaime:
"revision" is a name that we already used for git, using it to specify the project version seems pretty confusing.
I've looked with Google to see if there is a practice of calling these "revisions" but could not find any... can we stick to a "version" of sorts instead? "gs.version" for example?
Might be something worth discussing on the developer list, not everyone participates in PR discussions.

groldan:
> "revision" is a name that we already used for git, using it to specify the project version seems pretty confusing. I've looked with Google to see if there is a practice of calling these "revisions" but could not find any... can we stick to a "version" of sorts instead? 

Unfortunately it's a feature embedded in maven, the documentation clearly states that "revision" is treated specially and only ${revision}, ${sha1} and/or ${changelist} as placeholders for the version in your pom file:

https://maven.apache.org/maven-ci-friendly.html

"gs.version" for example? Might be something worth discussing on the developer list, not everyone participates in PR discussions.

sure, that's why I've created a draft pr, need to dig deeper into which documentation bits need upgrading and planned to discuss on the mailing list

PR #5834, opened 2022-04-27
STAC queryables and indexing handling improvements

aaime:
First two commits avoid NPEs on un-configured queryables at the global and per collection level.
The third is a set of improvements on automatic index management:
* Do a diff of queryables on reconfiguration, touch only the indexes that need to be created/destroyed
* Make sure all code paths map back to database properties (json path did, simple properties didn't)
* General code simplification (try-with-resouces, exception handling)

@turingtestfail could you have a look?

# Checklist

### DIFF
--- a/src/community/oseo/oseo-core/src/main/java/org/geoserver/opensearch/eo/store/Indexable.java
+++ b/src/community/oseo/oseo-core/src/main/java/org/geoserver/opensearch/eo/store/Indexable.java
@@ -7,6 +7,15 @@ package org.geoserver.opensearch.eo.store;
 import java.util.Objects;
 import org.opengis.filter.expression.Expression;
 
+/**
+ * Represents an indexable property essential traits:
+ *
+ * <ul>
+ *   <li>The queryable name
+ *   <li>The expression backing it
+ *   <li>The type of field, influencing how the index is built
+ * </ul>
+ */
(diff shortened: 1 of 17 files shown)


src/community/oseo/oseo-core/src/main/java/org/geoserver/opensearch/eo/store/Indexable.java
```diff
@@ -0,0 +1,83 @@
+/* (c) 2022 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.opensearch.eo.store;
+
+import java.util.Objects;
+import org.opengis.filter.expression.Expression;
+
```
turingtestfail, line 9: Javadoc

src/community/oseo/oseo-core/src/main/java/org/geoserver/opensearch/eo/store/JDBCIndex.java
```diff
@@ -0,0 +1,60 @@
+package org.geoserver.opensearch.eo.store;
+
```
turingtestfail, line 2: Javadoc

PR #5840, opened 2022-04-28
Link Loop

colintalex:
The link referring users to more details about JDBC User/Group Service and the `user_props` table takes the user to a page where the table is never mentioned. The next page in sequence takes you back to the page with the link for more details.  

Test it out here:
https://docs.geoserver.org/latest/en/user/security/usergrouprole/usergroupservices.html

I'm trying to document how this is supposed to work for my own needs first, and will propose much better documentation regarding this when complete.

Better documentation. Many items are well covered, but some items like JDBC User/Group have large gaps where there is no documentation. For example, the 'user_props' table is only mentioned as a reference, and it's purpose within JDBC/GeoServer is never explained. I've searched the internet for 2 days without any luck, including the geoserver-mailing list. 
  

# Checklist

I followed QuickFix instructions on Geoserver docs. I'm not a Java dev, but I have been using Geoserver and reading the docs for about 3 months.

### DIFF
--- a/doc/en/user/source/security/usergrouprole/usergroupservices.rst
+++ b/doc/en/user/source/security/usergrouprole/usergroupservices.rst
@@ -132,7 +132,7 @@ The JDBC user/group service persists the user/group database via JDBC, managing 
      - NO
      - PRI
 
-The ``users`` table is the primary table and contains the list of users with associated passwords. The ``user_props`` table maps additional properties to a user. (See :ref:`security_rolesystem_usergroups` for more details.)  The ``groups`` table lists all available groups, and the ``group_members`` table maps which users belong to which groups.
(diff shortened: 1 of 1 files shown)


bradh:
So just to follow-up: The table structure is on the page you linked to as the test page (https://docs.geoserver.org/latest/en/user/security/usergrouprole/usergroupservices.html#jdbc-user-group-service). The page that is linked to provides context (in that they are key-value pairs for implementation-specific purposes).

More generally, we don't have documentation for every part of geoserver, and certainly not all of it is right. We don't usually say "there is no documentation about X" because it doesn't really help people - sometimes more documentation on X is added later, or its there, and just not obvious. Corrections are always welcome.

PR #5867, opened 2022-05-16
[GEOS-10489] Add options to LDAP Role Service to configure prefixes a…

turingtestfail:
[GEOS-10489](https://osgeo-org.atlassian.net/browse/GEOS-10489)

…nd enforce capitalization

tests

make prefix blankable

null prefix becomes empty string not default value

documentation

  

# Checklist


src/security/ldap/src/main/java/org/geoserver/security/ldap/LDAPBaseSecurityServiceConfig.java
```diff
@@ -50,6 +50,8 @@ public abstract class LDAPBaseSecurityServiceConfig extends BaseSecurityNamedSer
 
     String adminGroup;
     String groupAdminGroup;
+    String rolePrefix = "ROLE_";
```
aaime, line 53: Is this class serialized/deserialized using XStream? If so, there might be issues with people upgrading from an older data directory, where the fields were missing.

XStream bypasses the normal object construction and these fields initializers won't be used, so defaults are not reliable. For the ``convertToUpperCase`` you can recognize ``null`` in the getter and make it "true" (to be done), but for the roleprefix it seems different, not sure there is a difference between the lack of XML tag, and possibly and empty tag, if someone configures it to have no prefix at all. Something you should check, I don't have a ready-made solution for it.

src/security/ldap/src/main/java/org/geoserver/security/ldap/LDAPRoleService.java
```diff
@@ -287,7 +295,7 @@ public GeoServerRole getParentRole(GeoServerRole role) throws IOException {
 
     @Override
     public GeoServerRole getRoleByName(String role) throws IOException {
-        if (role.startsWith("ROLE_")) {
+        if (role.startsWith(rolePrefix)) {
             // remove standard role prefix
             role = role.substring(5);
```
aaime, line 300: If the role prefix is configured, it might not be 5 chars long. Use its actual lenght.

src/security/ldap/src/main/java/org/geoserver/security/ldap/LDAPRoleService.java
```diff
@@ -353,7 +361,7 @@ public int getRoleCount() throws IOException {
     }
 
     private String normalizeGroupName(String role) {
-        if (role.startsWith("ROLE_")) {
+        if (role.startsWith(rolePrefix)) {
             // remove standard role prefix
             role = role.substring(5);
```
aaime, line 366: Same as above. See if there is any other place where this "5" appears and fix.

aaime (changes requested):
Looks good, I have one worry about people upgrading, and another issue with role trimming not accounting for the actual prefix length (tests should check that case too) See below.

src/security/ldap/src/main/java/org/geoserver/security/ldap/LDAPBaseSecurityServiceConfig.java
turingtestfail: Fixed so that nulls convert to defaults (and I validated that the lack of a tag in the XML is returned as null and the empty tag is an empty string) and empty string is returned by wicket when the field is empty.

src/security/ldap/src/main/java/org/geoserver/security/ldap/LDAPRoleService.java
turingtestfail: fixed

src/security/ldap/src/main/java/org/geoserver/security/ldap/LDAPRoleService.java
turingtestfail: fixed

aaime (approved):

PR #5874, opened 2022-05-20
[GEOS-10501]: Supporting auth headers forwarding to allowed remote SLD Urls

dromagnoli:
[GEOS-10501](https://osgeo-org.atlassian.net/browse/GEOS-10501)

[GEOS-10501]: When a GetMap request uses a remote URLs as SLD parameter, eventually forward auth headers to the remote SLD in case it's properly specified in the allowed list.
  

# Checklist


doc/en/user/source/services/wms/webadmin.rst
```diff
@@ -99,6 +99,17 @@ Some clients can have problems processing the large list of SRS (projections) th
...
+Authorization headers forwarding for remote SLDs
+------------------------------------------------
+A GetMap request may specify the style by referring a remote URL in the SLD parameter.
+There might be the case that the remote URL require same authorization headers as the current GetMap request.
+If that's the case a list of allowed style URLs can be specified.
+Authorization headers will be only forwarded to a remote URL when it starts with one of the specified URLs.
+
+.. figure:: img/services_WMS_remoteurlsauth.png
+
+  The list of remote URLs being allowed for authorization headers forwarding.
```
aaime, line 111: I'd add somewhere in here that the list is newline separated (other lists, like the SRS one, is comma separated). Btw, I agree newline is better in this context, URLs can be long, I'd just document it.

src/web/core/src/main/java/org/geoserver/web/wicket/HTTPURLsListTextArea.java
```diff
@@ -0,0 +1,101 @@
...
+            return (IConverter<C>) new URLsListConverter();
+        }
+        return super.getConverter(type);
+    }
+
+    private static class URLsListConverter implements IConverter<List<String>> {
+        private static final long serialVersionUID = 1083795866666107798L;
+
+        static final Pattern NEW_LINE_SEPARATED =
+                Pattern.compile("\\s*\\r?\\n\\s*", Pattern.MULTILINE);
```
aaime, line 47: Made me wonder about just "\r", but looks like OSX always used "\n" instead, and classic MacOS is long gone.

src/web/core/src/main/java/org/geoserver/web/wicket/HTTPURLsListTextArea.java
```diff
@@ -0,0 +1,101 @@
...
+                                try {
+                                    // Only valid URLs and HTTP Urls are supported
+                                    URL validURL = new URL(url);
+                                    isValid = validURL.getProtocol().startsWith("http");
+                                } catch (Exception e) {
+                                    isValid = false;
+                                }
+                                if (!isValid) {
+                                    invalid.add(url);
+                                }
```
aaime, line 89: Factoring this in a separate method and referring to it via method reference (``this::validateURL``) would improve both readability and formatting.

src/web/core/src/main/java/org/geoserver/web/wicket/HTTPURLsListTextArea.java
```diff
@@ -0,0 +1,101 @@
...
+                            });
+
+            if (!invalid.isEmpty()) {
+                IValidationError err =
+                        new ValidationError("HTTPURLsListTextArea.invalidURL")
```
aaime, line 94: This constructor takes a message rather than a i18n key... wondering if you could not just get away with the empty argument constructor here?

src/wms/src/main/java/org/geoserver/wms/WMSInfoImpl.java
```diff
@@ -24,6 +24,8 @@ public class WMSInfoImpl extends ServiceInfoImpl implements WMSInfo {
 
     List<String> srs = new ArrayList<>();
 
+    List<String> allowedURLsForAuthForwarding = new ArrayList<>();
```
aaime, line 27: XStream dodges this initialization, leaving the field null, when there is no tag in the XML (so, anyone upgrading).

src/wms/src/main/java/org/geoserver/wms/WMSInfoImpl.java
```diff
@@ -349,4 +351,13 @@ public GrowableInternationalString getInternationalRootLayerAbstract() {
     public void setInternationalRootLayerAbstract(InternationalString rootLayerAbstract) {
         this.internationalRootLayerAbstract = InternationalStringUtils.growable(rootLayerAbstract);
     }
+
+    @Override
+    public List<String> getAllowedURLsForAuthForwarding() {
+        return allowedURLsForAuthForwarding;
```
aaime, line 357: You can make check that the field is null, and set it to a new ArrayList if that's the case.
(not thread-safe, but config is not supposed to be modified concurrently... if you want belts and suspenders, also make the method synchronized or do a double-checked locking with volatile field).

aaime (changes requested):
See feedback below.
@sikeoka do you see any way this could become trouble?

src/web/core/src/main/java/org/geoserver/web/wicket/HTTPURLsListTextArea.java
dromagnoli: Sorry. I was "copying" from SRSListArea which was using the i18n key.
I will set it empty.

src/wms/src/main/java/org/geoserver/wms/WMSInfoImpl.java
dromagnoli: Ok. I simply did as a simple null check since I have found other pieces of code doing the same check)

src/web/core/src/main/java/org/geoserver/web/wicket/HTTPURLsListTextArea.java
dromagnoli: Done

doc/en/user/source/services/wms/webadmin.rst
dromagnoli: Done

src/wms/src/main/java/org/geoserver/wms/WMSInfoImpl.java
aaime: So if the collection is null, the only way to make it non-null is to cast to the Impl or use reflection to access the setter. It's not ideal... other bits of code might be doing it, but it's not really how it should be done....

dromagnoli: My fault. During a refactoring of another branch and another repo on IntelliJ I made some mistakes with git stash & co.
When I said "I simply did a null check" I was meaning that I have simply adopted the null check + new ArrayList as in other pieces of code instead of using the synchronized / volatile approach. 
However, for some reason that code change didn't end in the commit (due to the above IntelliJ mess) so I have re-updated it right now.

Sorry for the confusion :D

PR #5875, opened 2022-05-20
[GEOS-10502] GML3 output is not pretty printed when pretty print is turned on in global settings.

ianturton:
[GEOS-10502](https://osgeo-org.atlassian.net/browse/GEOS-10502)

This PR makes the GMLFormatters (both 2 and 3.0/3.2) honour the verbose/Pretty print setting in the global settings page.
  

# Checklist

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/xml/GML3OutputFormat.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/xml/GML3OutputFormat.java
@@ -28,6 +28,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.logging.Level;
+
 import javax.xml.namespace.QName;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.transform.ErrorListener;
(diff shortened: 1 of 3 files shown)


src/wfs/src/main/java/org/geoserver/wfs/xml/GML3OutputFormat.java
```diff
@@ -243,6 +243,12 @@ protected void write(
         Encoder encoder = createEncoder(configuration, ns2metas, gft);
 
         encoder.setEncoding(Charset.forName(geoServer.getSettings().getCharset()));
+        if (wfs.isVerbose() || geoServer.getSettings().isVerbose()) {
+            geoServer.getSettings().setVerbose(true);
```
aaime, line 247: For every invocation of this output format, a configuration change is made? That does not seem right....
(also, it's not persisted, but it's sort of a bonus in this situation).

ianturton: Good spot - that was when I was debugging which set of settings need to be set.

PR #5890, opened 2022-05-25
[GEOS-10487] Log4J custom logging XML configuration is case-insenstive

jodygarnett:
[GEOS-10487](https://osgeo-org.atlassian.net/browse/GEOS-10487)

The log4j configuration is case-insenstive, adjusting GeoServerXmlConfiguration to pre-process in a case-insenstive manner.

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/logging/GeoServerXMLConfiguration.java
+++ b/src/main/src/main/java/org/geoserver/logging/GeoServerXMLConfiguration.java
@@ -5,6 +5,7 @@
 package org.geoserver.logging;
 
 import java.io.File;
+import java.util.Map;
 import org.apache.logging.log4j.core.Appender;
 import org.apache.logging.log4j.core.LoggerContext;
 import org.apache.logging.log4j.core.appender.FileAppender;
@@ -29,7 +30,7 @@ public class GeoServerXMLConfiguration extends XmlConfiguration {
     /** Console node attribute name to supress for {@link #suppressStdOutLogging} */
     protected static final String STDOUT = "stdout";
 
-    /** File output node attribute name to supress for {@link #suppressFileLogging} */
+    /** File output node attribute name to suppress for {@link #suppressFileLogging} */
     protected static final String GEOSERVERLOGFILE = "geoserverlogfile";
 
     /** Node name for appender references */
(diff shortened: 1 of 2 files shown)


jodygarnett:
Backport requested to 2.21.0.

src/main/src/main/java/org/geoserver/logging/GeoServerXMLConfiguration.java
```diff
@@ -120,61 +121,133 @@ protected void stripConsoleStout(Node node) {
...
+     * not been corrected yet).
+     *
+     * <p>Use this to preform a case-insenstive lookup of attribute value: {@code
+     * node.getAttribute().get(attributeName(node,"fileName"))}.
+     *
+     * @param node Config node
+     * @param name Attribute name
+     * @return node attribute matching name, or null if not found
+     */
+    static final String attribute(Node node, String name) {
```
aaime, line 146: Seems to be used only once in a test, in a weird way... I'd suggest removing and using attributeGet instead (see below).

src/main/src/test/java/org/geoserver/logging/LoggingStartupContextListenerTest.java
```diff
@@ -174,4 +175,36 @@ public void testFilenameAndFilePattern() throws Exception {
                 GeoServerXMLConfiguration.applyPathTemplate(
                         "${baseDir}/logs/logging.log", "logs/geoserver.log"));
     }
+
+    @Test
+    public void testNodeAttributes() {
+        Node node = new Node();
+        node.getAttributes().put("fileName", "logs/geoserver.log");
+
+        assertEquals("fileName", GeoServerXMLConfiguration.attribute(node, "fileName"));
```
aaime, line 184: Something similar to this should work too, and avoid the need for a dedicated method?

```suggestion
        assertNotNull(GeoServerXMLConfiguration.attributeGet(node, "fileName"));
```

src/main/src/test/java/org/geoserver/logging/LoggingStartupContextListenerTest.java
```diff
@@ -174,4 +175,36 @@ public void testFilenameAndFilePattern() throws Exception {
...
+
+        assertEquals(
+                "attribute style",
+                "logs/geoserver.log",
+                GeoServerXMLConfiguration.attributeGet(node, "fileName"));
```
aaime, line 189: Can I suggest adding a static import, and get more concise code?

aaime (changes requested):

aaime:
Fun find: the Log4J configuration can be both "concise" (case insensitive) and "strict" (case sensitive), but I'm not sure how they tell the two cases apart... weird stuff :-D

See:
https://logging.apache.org/log4j/2.x/manual/configuration.html

src/main/src/test/java/org/geoserver/logging/LoggingStartupContextListenerTest.java
jodygarnett: Got it, double check that attributeGet is case in-sensitive and yes combine the methods.

src/main/src/test/java/org/geoserver/logging/LoggingStartupContextListenerTest.java
jodygarnett: You can suggest indeed, not generally worried about concise code for test cases.

I am trying to keep my options open for moving this node pre-processing into a distinct class (so it can be shared if we every make a GeoServerYAMLConfiguration). Perhaps static import with package visibility is a good approach.

jodygarnett:
> Fun find: the Log4J configuration can be both "concise" (case insensitive) and "strict" (case sensitive I suppose, says it can be validated with XML schema), but I'm not sure how they tell the two cases apart... weird stuff :-D
> 
> See: https://logging.apache.org/log4j/2.x/manual/configuration.html

Yeah I am deeply aware, which is why the public review was useful (turned up this bug for example).

PR #5894, opened 2022-05-25
[GEOS-10505] Display style format as new column in Styles-list, along with Style Name and Workspace

mmohiahmed:
[GEOS-10505](https://osgeo-org.atlassian.net/browse/GEOS-10505)

It's beneficial for our company (and hopefully other users), to add a new column 'Format' in Data > Styles table along with 'Name' and 'Workspace'. Formats will be SLD, ZIP, Property etc. Will you accept this as my first OSS contribution?
# Checklist


petersmythe:
@aaime @jodygarnett or other core developers, please take the time to review my AfriGIS colleague's PR.  We're trying to build up some internal developer capacity to tackle small changes, and hopefully more in time.  Thanks, Peter

jodygarnett:
@petersmythe I thought I had already provided feedback, here is a doc failure, and I am not sure how to add a test case to this one. Perhaps copy one of the other examples and confirm that the format is visible on screen?

aaime:
Should not be merged without a test. 

In order to add a test:
* Extend the existing StylePageTest
* Add one SLD 1.1 style (to have something different)
* Start StylePage, check the column is there and has the right contents (see the testWorkspaceFilter method for inspiration, you could have the style names set up so that only one style shows at a time, and check the SLD 1.0 and SLD 1.1 cases one by one.)

src/web/core/src/main/java/org/geoserver/web/wicket/StyleFormatLabel.java
```diff
@@ -0,0 +1,41 @@
...
+        if (model.getObject() != null) {
+            String format = (String) model.getObject();
+            formatDisplayName = Styles.handler(format).getName();
+
+            if (versionString != null) {
+                formatDisplayName =
+                        new StringBuilder(formatDisplayName)
+                                .append(" ")
+                                .append(versionString)
+                                .toString();
```
aaime, line 34: [No need for StringBuilder for simple concatenation](https://dzone.com/articles/string-concatenation-performacne-improvement-in-ja). Use it in loops or string builds spanning several lines, but here it's overkill and harder to read.

```suggestion
                        new formatDisplayName + " " + versionString;
```

src/web/wms/src/main/java/org/geoserver/wms/web/data/StylePage.java
```diff
@@ -63,6 +65,18 @@ protected Component getComponentForProperty(
                                             id,
                                             StyleProvider.CREATED_TIMESTAMP.getModel(itemModel));
                                 }
+                                if (property == StyleProvider.FORMAT) {
+                                    Version formatVersion =
```
aaime, line 69: Setting up the label is split among the code here and the decated StyleFormatLabel component. Please move everything inside StyleFormatLabel instead, you could give it the entire itemModel as constructor argument for example, and perform extraction and formatting in there.

src/web/wms/src/main/java/org/geoserver/wms/web/data/StylePage.java
```diff
@@ -151,4 +165,18 @@ protected static boolean isDefaultStyle(CatalogInfo catalogInfo) {
...
+        Comparable<?> major = formatVersion.getMajor();
+        Comparable<?> minor = formatVersion.getMinor();
+
+        if (major == null) {
+            return null;
+        }
+
+        if (minor == null) return major.toString();
+
+        return new StringBuilder(major.toString()).append(".").append(minor.toString()).toString();
```
aaime, line 180: Same as above, no need to use StringBuilder.

aaime (changes requested):

aaime:
Regarding the two build failures:
* The docs one is related to a new warning emitted by the latest version of Sphinx. If you rebase you branch on main (or merge) you'll get the fix required (just merged):
* The checkstyle violation is related to excessive code nesting, can normally be solved by factoring out some code bits in separate methods:

```
2022-05-30T10:42:14.9443610Z 10:42:14,943 [INFO] There is 1 error reported by Checkstyle 8.29 with /home/runner/work/geoserver/geoserver/src/../build/qa/checkstyle.xml ruleset.
2022-05-30T10:42:14.9453180Z 10:42:14,944 [ERROR] src/main/java/org/geoserver/wms/web/data/StylePage.java:[73] (regexp) RegexpSingleline: Excessive nesting found. Please try to factor out the deeply nested code in a separate method or class
```

aaime:
I've tried it locally again, found it would not display the style list anymore... this was due to some styles I have, that did not have their style handler plugged in (e..g, having CSS styles in the data dir, without the CSS plugin installed).
Before it was failing to use the styles, but would list them without issues: it's a regression, I've fixed it, merged with the latest main, and pushed the changes up. 

The build is now ongoing again, let's see if it gets to green.

aaime:
Cool, build was green, I've squash-merged the changes.

PR #5903, opened 2022-05-26
[Backport to 2.21.x][GEOS-10514]: Better capture catalog configuration issue on layergroup with a layer without resource

dromagnoli:
[GEOS-10514](https://osgeo-org.atlassian.net/browse/GEOS-10514)

Backporting GEOS-10514 to 2.21.x

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/impl/CatalogImpl.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/CatalogImpl.java
@@ -1067,14 +1067,15 @@ public class CatalogImpl implements Catalog {
                             + layerGroup.getWorkspace().getName());
         }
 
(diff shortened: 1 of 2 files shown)


src/main/src/main/java/org/geoserver/catalog/impl/CatalogImpl.java
```diff
@@ -1098,10 +1099,20 @@ private void checkLayerGroupResourceIsInWorkspace(StyleInfo style, WorkspaceInfo
...
         ResourceInfo r = layer.getResource();
+
+        if (r == null) {
+            throw new IllegalArgumentException(
+                    "Layer group "
+                            + layerGroup.getName()
+                            + " references a layer ("
+                            + layer.getId()
+                            + ") without a proper Resource attached");
+        }
```
jodygarnett, line 1115: Thanks, I really appreciate this change :)

PR #5906, opened 2022-05-27
[GEOS-10518] Partial RELINQUISH_LOG4J_CONTROL regression with WildFly

sikeoka:
[GEOS-10518](https://osgeo-org.atlassian.net/browse/GEOS-10518)

This PR moves the configureGeoServerLogging and initLogging method implementations from org.geoserver.logging.LoggingUtils into a new utility class (Log4jUtils) in order to remove all import statements of log4j-core classes from LoggingUtils.  The public API of LoggingUtils is not changed by this PR although there is a slight behavior change where those two methods now check the relinquishLog4jControl variable first.  Everything in Log4jUtils was copied directly out of LoggingUtils including the Javadocs.  The only difference with the Log4jUtils methods is the access level where the methods that are public in LoggingUtils are package private in Log4jUtils and all of the other Log4jUtils methods are private.  This was to make it clear that any code previously calling these LoggingUtils methods should still be calling the LoggingUtils methods instead of the Log4jUtils methods.

While this PR seemed to be sufficient to fix the RELINQUISH_LOG4J_CONTROL regression in WildFly, I can't guarantee that this will fix any similar regression in other application servers if they are also experiencing this issue. Any GeoServer deployment not affected by this regression, regardless of whether or not RELINQUISH_LOG4J_CONTROL is enabled, should not see any change in their logging behavior.
  

# Checklist


src/main/src/main/java/org/geoserver/logging/Log4jUtils.java
```diff
@@ -0,0 +1,705 @@
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2013 OpenPlans
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
```
jodygarnett, line 5: Where is this file from to have such an old header?

src/main/src/main/java/org/geoserver/logging/Log4jUtils.java
```diff
@@ -0,0 +1,705 @@
...
+ * -DRELINQUISH_LOG4J_CONTROL=true -Dlog4j.configuration=CUSTOM_LOGGING.properties}. This forces
+ * GeoServer to read your {@code CUSTOM_LOGGING.properties} Log4J 1.2 configuration (via Log4J
+ * bridge).
+ *
+ * <p>To troubleshoot use {@code org.apache.logging.log4j.simplelog.StatusLogger.level=DEBUG}, or
+ * adjust {@code log4j2.xml} file &lt;Configuration status=&quote;trace&quote;&gt;.
+ *
+ * @see org.geoserver.config.LoggingInfo
+ */
+public class Log4jUtils {
```
jodygarnett, line 61: A couple quick suggestions:

* Rename to LoggingUtilsDelegate (since this is kind of a delegate if I read the code correctly).
* No need to be a public class, package visibility is enough 
* Double check LoggingStartupContextListener as it should be called first (it is responsible for redirecting if needed).

jodygarnett (changes requested):
I like the approach, consider feedback below.

Can you think of any way we can test this?

aaime:
About a test... I guess it should check if the Log4j classes have been loaded.
Maybe [this S.O. suggestion](https://stackoverflow.com/a/482637/731672) has merit... set up a custom classloader, that overrides that "findLoadedClass" method, and use it to check if class loading happened. That said, seems tricky, I'd give it a try before the code modifications to make sure it actually catches the issue.

src/main/src/main/java/org/geoserver/logging/Log4jUtils.java
sikeoka: I removed the older copyright from the header.

src/main/src/main/java/org/geoserver/logging/Log4jUtils.java
sikeoka: I renamed the class and changed it to package visibility.  There were no changes made to LoggingStartupContextListener  or that need to be made to that class.

sikeoka:
I tried to create a test based on that suggestion but I could only get it to work when running the test individually and not with the whole module's tests.

aaime:
Never mind, good try. I guess for this one we can overlook the test requirement, as "too hard to test reliably".

jodygarnett (approved):

aaime:
@jodygarnett when you approve a PR it would be good if you could also merge it :rofl: 
Doing so now.

aaime:
Oh, but this PR does not contain the bot configuration yet, so probably it won't be created...

aaime:
How interesting, it did anyways:
https://github.com/geoserver/geoserver/pull/5928

I guess that workflow consider the state of the source code past merge.

jodygarnett:
Thanks for the improvement @sikeoka, and minding the backport bot aaime.

PR #5922, opened 2022-06-03
[GEOS-10528] Empty groupList tag remains after group deletion

ahmedababnehgeo:
[GEOS-10528](https://osgeo-org.atlassian.net/browse/GEOS-10528)

We noticed that, in when using XML storage for roles, pressing `Remove Selected and remove role associations`, truly removes the group and role associations, but on inspecting the `roles.xml` file, we found an empty user groups tag is still there.

This can create confusion for the system administrators, since the empty tag need not be there anymore.

In this PR, we remove the `groupList` in the above-mentioned case.
  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/impl/AbstractRoleStore.java
+++ b/src/main/src/main/java/org/geoserver/security/impl/AbstractRoleStore.java
@@ -210,7 +210,7 @@ public abstract class AbstractRoleStore implements GeoServerRoleStore {
             roles.remove(role);
             setModified(true);
         }
-        if (helper.group_roleMap.get(groupname) != null && roles != null && roles.size() == 0) {
+        if (helper.group_roleMap.get(groupname) != null && roles != null && roles.isEmpty()) {
             helper.group_roleMap.remove(groupname);
         }
     }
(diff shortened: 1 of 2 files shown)


ahmedababnehgeo:
relevant JIRA: https://osgeo-org.atlassian.net/browse/GEOS-10528

ahmedababnehgeo:
@taba90 PR ready for review

src/main/src/main/java/org/geoserver/security/impl/AbstractRoleStore.java
```diff
@@ -210,6 +210,9 @@ public void disAssociateRoleFromGroup(GeoServerRole role, String groupname) thro
             roles.remove(role);
             setModified(true);
         }
+        if (helper.group_roleMap.get(groupname) != null && roles != null && roles.size() == 0) {
```
aaime, line 213: Surprised the PMD rule on the usage of isEmpty did not catch this one, but to be consistent with the resto of the code:

```suggestion
        if (helper.group_roleMap.get(groupname) != null && roles != null && roles.isEmpty()) {
```

aaime (changes requested):
Looks good, minor nitpick for consistency, see below.

ahmedababnehgeo:
@aaime comment resolved

PR #5938, opened 2022-06-09
[GEOS-10535] WFS Update request throw NPE on bad namespace

ianturton:
# Checklist

### DIFF
--- a/src/wfs/src/test/java/org/geoserver/wfs/TransactionTest.java
+++ b/src/wfs/src/test/java/org/geoserver/wfs/TransactionTest.java
@@ -1095,4 +1095,34 @@ public class TransactionTest extends WFSTestSupport {
             fail("Should not throw an exception");
         }
     }
+
+    @Test
+    public void testBrokenUpdate() {
+
+        // perform an update
+        String update =
+                "<wfs:Transaction service=\"WFS\" version=\"1.0.0\" "
+                        + "xmlns:cgf=\"cite\" "
+                        + "xmlns:ogc=\"http://www.opengis.net/ogc\" "
+                        + "xmlns:wfs=\"http://www.opengis.net/wfs\" "
+                        + "xmlns:gml=\"http://www.opengis.net/gml\"> "
+                        + "<wfs:Update typeName=\"cgf:Polygons\" > "
+                        + "<wfs:Property>"
+                        + "<wfs:Name>id</wfs:Name>"
+                        + "<wfs:Value>t0003</wfs:Value>"
+                        + "</wfs:Property>"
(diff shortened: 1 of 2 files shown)


aaime:
Added the backport labels, made some changes, let's see if now it works (will trigger on PR merge)

src/wfs/src/test/java/org/geoserver/wfs/TransactionTest.java
```diff
@@ -1095,4 +1095,34 @@ public void testBrokenDelete() throws Exception {
...
+                        + "<ogc:Filter>"
+                        + "<ogc:PropertyIsEqualTo>"
+                        + "<ogc:PropertyName>id</ogc:PropertyName>"
+                        + "<ogc:Literal>t0002</ogc:Literal>"
+                        + "</ogc:PropertyIsEqualTo>"
+                        + "</ogc:Filter>"
+                        + "</wfs:Update>"
+                        + "</wfs:Transaction>";
+        try {
+            postAsDOM("wfs", update);
```
aaime, line 1123: You can check the contents of a service exception using one of the various ``checkOwsXYException`` methods.. I think you need ``checkOws10Exception`` but not completely sure.

ianturton: So you can :-p

aaime: @ianturton this does not look like a serious test. If you're doing this on volunteer time, please wait and I'll add a better one (hopefully this weekend). If you are on consultancy time, please remember to charge for writing proper tests too.

ianturton: All I want to test is that it doesn't throw a NPE and we get a proper error message - seems like a fine test to me.

aaime: You're sending a request to the service... even in case of a NPE, you should be getting a 200, with a service exception, telling inside the XML that a NPE happened. Or did the NPE occur inside the Dispatcher? (that's the only place where I think an exception would be able to escape out to the container).

bradh:
@aaime Looks like a problem with the bot here.

```
remote: Permission to geoserver/geoserver.git denied to geoserver-bot.
fatal: unable to access 'https://github.com/geoserver/geoserver.git/': The requested URL returned error: 403
```

aaime:
Aware, I'm discussing how it should be setup here:
https://github.com/m-kuhn/backport/pull/21#discussion_r893445680

PR #5944, opened 2022-06-09
[GEOS-10536] Security - oauth2 oidc - changes for keycloak server

davidblasby:
[GEOS-10536](https://osgeo-org.atlassian.net/browse/GEOS-10536)

I tested the current oauth2-openid-connect community module against keycloak acting as an oidc server.

There were two problems (GEOS-10536);

a) The spring oauth2 library was giving an error with the token that keycloak was returning - "enc (use) is currently not supported"
     + This is fixed in later spring oauth2 libraries (it ignores these types of tokens)
     + FIX: update to a newer (2.0.16.RELEASE -> 2.5.2.RELEASE) spring oauth2 library
     + The spring oauth2 module is only used in the oauth2 community modules, so this should be "safe" for the rest of GS

b) Keycloak, by default, will put the roles in the ID Token in a path like "resource_access.CLIENT_NAME.roles"
     + The current oauth2 library only supports a simple (one level) path for the roles list (i.e. "roles" or "group")
     + I added support for a json path
     + To do this, I added a dependency on  com.jayway.jsonpath (which I took from the main GS pom.xml)
     + I added a test case
  

# Checklist

- [n/a ] The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers).
- [GEOS-10536] There is an issue in the [GeoServer Jira](https://osgeo-org.atlassian.net/browse/GEOS/summary) (except for changes that do not affect administrators or end users in any way).

### DIFF
--- a/src/community/security/oauth2-openid-connect/pom.xml
+++ b/src/community/security/oauth2-openid-connect/pom.xml
@@ -38,11 +38,14 @@
       <artifactId>gs-sec-oauth2-core</artifactId>
       <version>${project.version}</version>
     </dependency>
-
+    <dependency>
+      <groupId>com.jayway.jsonpath</groupId>
+      <artifactId>json-path</artifactId>
+    </dependency>
     <dependency>
       <groupId>org.springframework.security.oauth</groupId>
       <artifactId>spring-security-oauth2</artifactId>
-      <version>2.0.16.RELEASE</version>
+      <version>2.5.2.RELEASE</version>
     </dependency>
     <dependency>
       <groupId>org.springframework.security</groupId>
--- a/src/community/security/oauth2-openid-connect/src/main/java/org/geoserver/security/oauth2/OpenIdConnectAuthenticationFilter.java
+++ b/src/community/security/oauth2-openid-connect/src/main/java/org/geoserver/security/oauth2/OpenIdConnectAuthenticationFilter.java
@@ -9,14 +9,13 @@ package org.geoserver.security.oauth2;
 import static org.geoserver.security.oauth2.OpenIdConnectFilterConfig.OpenIdRoleSource.AccessToken;
 import static org.geoserver.security.oauth2.OpenIdConnectFilterConfig.OpenIdRoleSource.IdToken;
 
+import com.jayway.jsonpath.JsonPath;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.logging.Level;
 import javax.servlet.http.HttpServletRequest;
-import net.sf.json.JSONObject;
-import net.sf.json.JSONSerializer;
 import org.geoserver.security.config.RoleSource;
 import org.geoserver.security.config.SecurityNamedServiceConfig;
 import org.geoserver.security.impl.GeoServerRole;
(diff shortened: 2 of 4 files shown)


jodygarnett:
The keycloak client is being phased out at the end of the year, thus the interest in using OAuth2 to connect to keycloak service.

jodygarnett:
I do not see a problem updating the spring-security-oauth2 module, however I have a question on medium term strategy.

The spring-security-oauth2 module already reached [end-of-life](https://spring.io/projects/spring-security-oauth).
Is this entire section going to need to be rewritten for to use [spring-security](https://docs.spring.io/spring-security/reference/servlet/oauth2/index.html) built-in client?

src/community/security/oauth2-openid-connect/src/main/java/org/geoserver/security/oauth2/OpenIdConnectAuthenticationFilter.java
```diff
@@ -69,17 +68,26 @@ protected Collection<GeoServerRole> getRoles(HttpServletRequest request, String
         return super.getRoles(request, principal);
     }
 
+    public static Object extractFromJSON(String json, String path) {
```
jodygarnett, line 71:
```suggestion
    static Object extractFromJSON(String json, String path) {
```

jodygarnett: Reduce visibility (suggested change) or document public method with javadoc (surprised a code QA check has not caught this?).

src/community/security/oauth2-openid-connect/src/main/java/org/geoserver/security/oauth2/OpenIdConnectAuthenticationFilter.java
```diff
@@ -69,17 +68,26 @@ protected Collection<GeoServerRole> getRoles(HttpServletRequest request, String
...
     private Collection<GeoServerRole> getRolesFromToken(String token) throws IOException {
         if (token == null) {
             LOGGER.warning("Token not found, cannot perform role extraction");
             return new ArrayList<>();
         }
         Jwt decoded = JwtHelper.decode(token);
         String claims = decoded.getClaims();
-        JSONObject json = (JSONObject) JSONSerializer.toJSON(claims);
-        String rolesAttribute =
+        // JSONObject json = (JSONObject) JSONSerializer.toJSON(claims);
```
jodygarnett, line 87:
```suggestion
```

jodygarnett (approved):
Very minor suggestions below.

davidblasby:
I expect the entire oauth2 modules will have to be re-written.

src/community/security/oauth2-openid-connect/pom.xml
```diff
@@ -38,11 +38,15 @@
       <artifactId>gs-sec-oauth2-core</artifactId>
       <version>${project.version}</version>
     </dependency>
-
+    <dependency>
+      <groupId>com.jayway.jsonpath</groupId>
+      <artifactId>json-path</artifactId>
+      <version>2.4.0</version>
```
jodygarnett, line 44:
```suggestion
```

jodygarnett: No need for version, it is covered by `src/pom.xml` dependency management already.

aaime:
@taba90 you've been recently working on OAuth2 support on Azure right? Could you have a look at this one?

taba90:
Looks good to me.

PR #5952, opened 2022-06-13
[GEOS-10542] Use headers for proxy URL on workspace level

ahmedababnehgeo:
[GEOS-10542](https://osgeo-org.atlassian.net/browse/GEOS-10542)

This PR aims at adding a new checkbox on the workspace level to allow adding headers for the proxy URL. This will allow end users to override the disabled checkbox "Use headers for Proxy URL" on the global settings page, by checking this option per workspace.
  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/ows/ProxifyingURLMangler.java
+++ b/src/main/src/main/java/org/geoserver/ows/ProxifyingURLMangler.java
@@ -97,11 +97,7 @@ public class ProxifyingURLMangler implements URLMangler {
         // (for two reasons: a) speed; b) to make the admin aware of
         // possible security liabilities)
 
-        boolean workspaceEnabled = this.geoServer.getSettings().getWorkspace() != null;
-        boolean doMangleHeaders =
-                workspaceEnabled
-                        ? this.geoServer.getSettings().isUseHeadersProxyURL()
-                        : this.geoServer.getGlobal().isUseHeadersProxyURL();
+        boolean doMangleHeaders = geoServer.getSettings().isUseHeadersProxyURL();
 
         if (proxyBase != null && doMangleHeaders) {
             this.mangleURLHeaders(baseURL, proxyBase);
--- a/src/main/src/main/java/org/geoserver/ows/ProxifyingURLMangler.java
+++ b/src/main/src/main/java/org/geoserver/ows/ProxifyingURLMangler.java
@@ -96,14 +96,22 @@ public class ProxifyingURLMangler implements URLMangler {
         // Mangles the URL base in different ways based on a flag
         // (for two reasons: a) speed; b) to make the admin aware of
         // possible security liabilities)
-
-        boolean doMangleHeaders = geoServer.getSettings().isUseHeadersProxyURL();
-
+        boolean doMangleHeaders = resolveDoMangleHeaders();
         if (proxyBase != null && doMangleHeaders) {
             this.mangleURLHeaders(baseURL, proxyBase);
         } else {
             this.mangleURLFixedURL(baseURL, proxyBase);
         }
+    }
+
+    private boolean resolveDoMangleHeaders() {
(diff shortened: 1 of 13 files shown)


ahmedababnehgeo:
@fernandor777 PR ready for review

src/main/src/main/java/org/geoserver/ows/ProxifyingURLMangler.java
```diff
@@ -95,10 +95,15 @@ public void mangleURL(
...
         // possible security liabilities)
 
-        if (this.geoServer.getGlobal().isUseHeadersProxyURL() == true && proxyBase != null) {
+        boolean workspaceEnabled = this.geoServer.getSettings().getWorkspace() != null;
+        boolean doMangleHeaders =
+                workspaceEnabled
+                        ? this.geoServer.getSettings().isUseHeadersProxyURL()
+                        : this.geoServer.getGlobal().isUseHeadersProxyURL();
+
+        if (proxyBase != null && doMangleHeaders) {
```
fernandor777, line 106: Please add some FINE level logging to know whats happening and the result of this logic conditions execution.

fernandor777 (commented):
Overall looks good, just some quick request for logging lines.

src/main/src/main/java/org/geoserver/ows/ProxifyingURLMangler.java
groldan: While it seems ok to set the flag in `SettingsInfo`, there's still an impedance mismatch with `GeoServer.getGlobal().getSettings():SeettingsInfo`, on which these would be of no use.
If `isUseHeadersProxyURL` is **moved** instead to `SettingsInfo`, everything would be consistent and this decision making unnecessary, leaving just `doMangleHeaders = this.geoServer.getSettings().isUseHeadersProxyURL()`, and letting `geoServer.getSettings()` decide whether to return the workspace or the global settings.

That of course implies deprecating `GeoServerInfo.is/setUseHeadersProxyURL` and defer to its internal `SettingsInfo` (e.g. `@Deprecated public Boolean isUseHeadersProxyURL() { return getSettings().isUseHeadersProxyURL();}`)

groldan (changes requested):

src/main/src/main/java/org/geoserver/ows/ProxifyingURLMangler.java
groldan: @aaime thoughts?

groldan: hmm thinking further, still the workspace settings may not have the value set at all and you'll still need to manually select the global settings value. For which you would need to define it as a `Boolean` instead of `boolean` and allow it to be `null`. Maybe add an `Optional<Boolean> useUseHeadersProxyURL(){return Optional.ofNullable(this.useHeadersProxyURL);}`

aaime:
I agree that ``useHeadersForProxyURL`` should not have been placed inside GeoServerInfo, but inside SettingsInfo, where the proxy base URL also is. Deprecating and delegating sounds like a good plan, but there is a hole: upgrading users.

The current setting has been around for 4 years now, it must keep on working as people upgrade from older data directories.
It is also going to break whoever is using the REST API, since XStream goes directly onto the fields, skipping the getters/setters.

One possible avenue is to write a custom XStream converter, that would migrate the value from the GlobalInfoImpl, to its settings, on deserialization. Could be a subclass of AbstractReflectionConverter in XStreamPersister, and dedicated to the serialization of GeoServerInfoImpl.

aaime:
Looks good to me, thanks!

jodygarnett:
I cam to this pull request because this is listed as a new feature for GeoServer 2.21.1 ... but I do not see docs?

Please do not merge a pull request with a user interface change with out docs.

PR #5963, opened 2022-06-14
Make ResoucePool's CacheCleanerListener usable from external code

groldan:
Convert the public CacheCleanerListener inner class to static,
receiving the ResourcePool to use as a constructor argument.

Otherwise the ResoucePool's package private `flushDataStore()`
method can't be triggered even calling `clear(FeatureTypeInfo info)`.

Allows to use it from external code. E.g.:

```
info.accpet(new ReourcePool.CacheCleanerListener(resourcePool));
```

  

# Checklist


src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -2507,48 +2515,41 @@ public void handleModifyEvent(CatalogModifyEvent event) {}
         public void handlePostModifyEvent(CatalogPostModifyEvent event) {
             CatalogInfo source = event.getSource();
             source.accept(this);
-
```
aaime, line 2510: These changes do not relate to making the module "usable from external code". Looks like this flush and the other one in ``handleRemoveEvent`` are related to JDBCConfig (guessing by the history of changes).

@NielsCharlier you added this bit of code as part of https://osgeo-org.atlassian.net/browse/GEOS-9169, could you please have a look?

aaime: I did not notice... right now there are 278 modified files, hard to comment further. @groldan can you clean up the PR again?

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -2507,48 +2515,41 @@ public void handleModifyEvent(CatalogModifyEvent event) {}
...
         public void visit(CoverageStoreInfo coverageStore) {
-            clear(coverageStore);
+            pool.clear(coverageStore);
         }
 
         @Override
         public void visit(FeatureTypeInfo featureType) {
-            clear(featureType);
+            pool.clear(featureType);
+            pool.flushDataStore(featureType);
```
aaime, line 2542: Ah ha, ok now I see it.

PR #5964, opened 2022-06-15
[GEOS-10545] Layer Group cache not initialized

ahmedababnehgeo:
[GEOS-10545](https://osgeo-org.atlassian.net/browse/GEOS-10545)

This PR resolves a cache issue that impacts the layer groups cache. We make sure that we call `buildLayerGroupCache` using a post processor bean. The bean checks whether the bean is of the desired type in order to build the cache for that bean (i.e. SecureCatalog object).
 

# Checklist


ahmedababnehgeo:
@taba90 PR ready for review

taba90:
> It looks to me like building the layer groups cache is an implementation detail for the `GeoFenceAccessManager` and it's not worth nor appropriate to extend `ResourceAccessManager`s interface with a `buildLayerGroupCache()`. Instead, `GeoFenceAccessManager` itself could be an `ApplicationListener`, or `ApplicationListenerImpl` be moved to the `org.geoserver.geofence` package (and renamed to something less generic).

Thanks for your point @groldan. I agree it doesn't look the ideal solution. However the main issue is that the LayerGroupCache is also used from the DefaultResourceAccessManager that is not a bean but is instead initialized from the [SecureCatalog](https://github.com/geoserver/geoserver/blob/fe4a8909d51df12f13dc0f3c7e368780add2a6e2/src/main/src/main/java/org/geoserver/security/SecureCatalogImpl.java#L114) if no bean of type ResourceAccessManager is found.
Another approach might be to call the reload method in the listeners attached to the catalog when application context is refreshed, since the LayerGroupCache has one that build the cache in the reloaded [method](https://github.com/geoserver/geoserver/blob/fe4a8909d51df12f13dc0f3c7e368780add2a6e2/src/main/src/main/java/org/geoserver/security/SecureCatalogImpl.java#L114 ). What do you think?

aaime:
Oh no, the LayerGroupContainmentCache has been introduced in 2016 for usage in DefaultResourceAccessManager (see GEOS-7913), GeoFence had initial support for layer groups only in 2020. 

The containment cache allows to efficiently handle [container-tree related security rules](https://docs.geoserver.org/stable/en/user/security/layer.html#rules). As such, it's an implementation detail used to speed up containment lookups.

The current approach exposes it in both security managers, which looks indeed like exposing an implementation detail (another security manager might decide to break containment-tree security handling for some reason), and does not fit in the ResouceAccessManager interface (which deals with security, not how it's implemented).

In the case of DefaultResourceAccessManager the external listener is mandatory, as it's not part of the application context (it's instantiated programmatically if no other resource access manager is found). Thinking out loud, what about a separate interface, LayerGroupContainmentCacheInitializer (yeah it's a mouthful, better names accepted) that would expose the method,  and that the listener could test and then invoke?

groldan:
hi, thanks for the insights. So to recap, the current situation is like this:

- `SecureCatalogImpl` is a bean defined in `gs-main`'s `applicationContext.xml`
- It has an overloaded constructor to receive both the catalog and the `ResourceAccessManager`
- It is bound with the one that only receives the catalog, which in turn delegates to the overloaded one with the result of the static method
```
    static ResourceAccessManager lookupResourceAccessManager() throws Exception {
        ResourceAccessManager manager = GeoServerExtensions.bean(ResourceAccessManager.class);
        if (manager == null) {
            manager = buildDefaultResourceAccessManager();
        }
        CatalogFilterAccessManager lwManager = new CatalogFilterAccessManager();
        lwManager.setDelegate(manager);
        return lwManager;
    }
    private static DefaultResourceAccessManager buildDefaultResourceAccessManager() {
        return new DefaultResourceAccessManager(
                GeoServerExtensions.bean(DataAccessRuleDAO.class),
                (Catalog) GeoServerExtensions.bean("rawCatalog"));
    }

```
- both `DefaultResourceAccessManager` and `GeofenceAccessManager` create their own `LayerGroupContainmentCache`

This is more "magic" than I'd like, all those `GeoServerExtensions.bean()` calls trigger bean initialization at a point where they may not be ready and cause cycle errors. I know I had issues with that in the past. Ideally, I'd rather have us eat out own dog food and do rely on dependency injection. Meaning:

- `LayerGroupContainmentCache` is turned into a bean, and is responsible of clearing itself up upon a context refreshed event
- `DefaultResourceAccessManager` is turned into a bean
- Both `DefaultResourceAccessManager` and `GeoFenceAccessManager` get the `LayerGroupContainmentCache` injected
- `GeoFenceAccessManager` overrides `DefaultResourceAccessManager`

aaime:
@groldan that will break other resource access managers found in customized GeoServer installations. I know our company has written one for several large customers because they had specific needs not addressed by GeoFence (which is instead used by most smaller setups). I expect to find several others out there.

I'm not totally against it, but I'd suggest you to make a PR in that direction, make sure it's still possible to install a security manager extension by just dropping the zip in WEB-INF, and finally create a GeoServer upgrade section explaining how things should be changed (afaik we don't have one, it would be brand new).

Thinking out loud, how about a simpler approach, where both beans are available in the app context, but the catalog impl prefers anything but the default resource access manager, if found? (yes, still based on GeoServerExtensions).

ahmedababnehgeo:
@taba90 I have modified the implementation taking into consideration the points raised above. Could you take a look please?

src/extension/geofence/src/main/java/org/geoserver/geofence/GeofenceAccessManager.java
```diff
@@ -1084,4 +1088,43 @@ public List<Object> parseLayers(
     public int getPriority() {
         return ExtensionPriority.LOWEST;
     }
+
+    // a container for accessInfo that can hold spatial filters as Geometry type.
+    class GeoserverAccessInfo {
```
taba90, line 1093: This addition I guess was unwanted. Was removed in another pr. Please remove it.

src/main/src/main/java/org/geoserver/security/ResourceAccessManager.java
```diff
@@ -71,4 +71,6 @@ public default LayerGroupAccessLimits getAccessLimits(
      */
     public @Nullable Filter getSecurityFilter(
             Authentication user, final Class<? extends CatalogInfo> clazz);
+
+    default void buildLayerGroupCache() {}
```
taba90, line 75: This should go away as is no more used.

src/main/src/main/java/org/geoserver/security/impl/LayerGroupContainmentCache.java
```diff
@@ -72,7 +77,7 @@ public LayerGroupContainmentCache(Catalog catalog) {
         buildLayerGroupCaches();
     }
 
-    private void buildLayerGroupCaches() {
+    public void buildLayerGroupCaches() {
```
taba90, line 80: Looks to me that this method can be left as private as is no more needed to have it public.

src/main/src/main/java/org/geoserver/security/impl/LayerGroupContainmentCache.java
```diff
@@ -190,6 +195,14 @@ private void collectContainers(LayerGroupSummary lg, Set<LayerGroupSummary> grou
         }
     }
 
+    @Override
+    public void onApplicationEvent(ApplicationEvent applicationEvent) {
+        if (applicationEvent instanceof ContextRefreshedEvent) {
+            SecureCatalogImpl secureCatalog = GeoServerExtensions.bean(SecureCatalogImpl.class);
+            secureCatalog.getResourceAccessManager().buildLayerGroupCache();
+        }
```
taba90, line 203: we don't need this logic. The LayerGroupContainment cache is a singleton bean now. You can directly invoke the buildLayerGroupCache method no need to go through the SecureCatalogImpl and the AccessManager.

taba90 (changes requested):
There are some things to fix. Please check my comments below.

ahmedababnehgeo:
This PR was closed. Here is the new PR: https://github.com/geoserver/geoserver/pull/5985

PR #5971, opened 2022-06-17
[GEOS-10551] Refactor commons-httpclient usage in the WPS module

sikeoka:
[GEOS-10551](https://osgeo-org.atlassian.net/browse/GEOS-10551)

This PR refactors the WPS module from using commons-httpclient 3.1 to httpclient 4.5.13.  The getValueInternal method was broken up from a single large method into smaller methods and try-with-resources statements are now used since the new HttpClient and HttpResponse classes implement Closeable.  There is also improved handling of HTTPS references to match HTTP references, all checks performed on references are also now applied to body references, a bug was fixed that prevented using HTTP body references, HTTP error status codes are no longer exposed to users and file body references can no longer obtain directory listings.  There is also now extensive unit testing for this class.

This PR is intended to support eventually removing the commons-httpclient library from GeoServer and does not need to be backported.

  

# Checklist


mprins (approved):

src/extension/wps/wps-core/src/test/java/org/geoserver/wps/executor/RemoteRequestInputProviderTest.java
```diff
@@ -0,0 +1,939 @@
...
+import org.geoserver.wps.ppio.ComplexPPIO;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.Test;
+import org.opengis.util.ProgressListener;
+
+public class RemoteRequestInputProviderTest {
```
aaime, line 67: That's a lot of tests!!!

aaime (commented):
Looks good... I just don't see a test using foo.zip?

src/extension/wps/wps-core/src/test/java/org/geoserver/wps/executor/RemoteRequestInputProviderTest.java
sikeoka: test.zip and foo.zip are used to test URLs that aren't file, http or https.

PR #5972, opened 2022-06-17
[GEOS-10516] WMS GetCapabilities dimension representations ignores the end attribute

sikeoka:
[GEOS-10516](https://osgeo-org.atlassian.net/browse/GEOS-10516)
This PR updates WMS GetCapabilities dimension representations to use the end attribute for feature types with a configured end attribute when the dimension presentation is not List.  I had problems running the WMSTest that uses TimeElevationWithStartEnd.properties in a machine with a non-UTC timezone so I had to update that too for all of the tests to pass on my machine.

  

# Checklist

### DIFF
--- a/src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/DimensionsVectorCapabilitiesTest.java
+++ b/src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/DimensionsVectorCapabilitiesTest.java
@@ -32,6 +32,7 @@ import org.geoserver.security.TestResourceAccessManager;
 import org.geoserver.security.VectorAccessLimits;
 import org.geoserver.security.impl.AbstractUserGroupService;
 import org.geoserver.wms.WMSDimensionsTestSupport;
+import org.junit.After;
 import org.junit.Test;
 import org.opengis.filter.Filter;
 import org.w3c.dom.Document;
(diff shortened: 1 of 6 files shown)


src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/DimensionsVectorCapabilitiesTest.java
```diff
@@ -270,6 +313,32 @@ public void testElevationDiscrerteManualResolution() throws Exception {
...
+        // check dimension has been declared
+        assertXpathEvaluatesTo("1", "count(//Layer/Dimension)", dom);
+        assertXpathEvaluatesTo("elevation", "//Layer/Dimension/@name", dom);
+        assertXpathEvaluatesTo(UNITS, "//Layer/Dimension/@units", dom);
+        assertXpathEvaluatesTo(UNIT_SYMBOL, "//Layer/Dimension/@unitSymbol", dom);
+        // check we have the extent
+        assertXpathEvaluatesTo("1", "count(//Layer/Extent)", dom);
+        assertXpathEvaluatesTo("elevation", "//Layer/Extent/@name", dom);
+        assertXpathEvaluatesTo("1.0", "//Layer/Extent/@default", dom);
+        assertXpathEvaluatesTo("1.0/3.0/2.0", "//Layer/Extent", dom);
```
aaime, line 339: Someone reading the capabilities would conclude that the specification is nonsensical, as 2 does not fit in the min/max value... but there is also no way to indicate that the dimension is based on intervals, rather than single values. Looks like something the spec did not foresee.... I searched in both WMS 1.1 and WMS 1.3 and found no indication of how to handle such case. It's probably a good idea to explain this situation in the GeoServer documentation.

aaime (commented):
Looks good in general, but the results might be surprising users. See comment below.

src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/DimensionsVectorCapabilitiesTest.java
sikeoka: GeoServer lets administrators set pretty much any resolution they want as long as it is a number so whether or not the capabilities is showing something nonsensical is really up to the administrator even without this change.

PR #5986, opened 2022-06-23
GEOS-10559 Lazy initialization of GetMapKvpRequestReader's HTTP client used to fetch remote styles

groldan:
[GEOS-10559](https://osgeo-org.atlassian.net/browse/GEOS-10559)

`GetMapKvpRequestReader` creates an apache
`CloseableHttpClient` on its constructor.

To do so, it needs to get the stles cache configuration from the
`org.geoserver.wms.WMS facade.`

Depending on the bootstrap conditions, this triggers an eager initialization
of beans, that can lead to a circular reference error.
This issue is observed in GeoServer Cloud.

Also, the eager initialization of the HTTP client can result in a waste
of resources, because it may never use it, and because there are subclasses
that definitely won’t use it.

Some of these subclasses are not spring beans, but created on-demand,
such as `WMSRequests.LayerParser`, which currently adds a
`ConfigurationListener` to `org.geoserver.config.GeoServer`
on each invocation of `WMSRequests.getGetMapParams()`, that is never
removed, hence being a memory leak.

As a solution, the HTTP client used by `GetMapKvpRequestReader` is now
lazily initialized, and the `ConfigurationListener` it uses to update
the cache configuration whenever `WMSInfo` is changed, removed whenever
the HTTP client is closed.

  

# Checklist

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/map/GetMapKvpRequestReader.java
+++ b/src/wms/src/main/java/org/geoserver/wms/map/GetMapKvpRequestReader.java
@@ -21,6 +21,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.Timer;
@@ -54,6 +55,7 @@ import org.geoserver.catalog.Styles;
 import org.geoserver.catalog.WMSLayerInfo;
 import org.geoserver.catalog.util.ReaderDimensionsAccessor;
 import org.geoserver.config.ConfigurationListenerAdapter;
+import org.geoserver.config.GeoServer;
 import org.geoserver.config.ServiceInfo;
 import org.geoserver.ows.Dispatcher;
 import org.geoserver.ows.KvpRequestReader;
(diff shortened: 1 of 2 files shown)


src/wms/src/main/java/org/geoserver/wms/map/GetMapKvpRequestReader.java
```diff
@@ -139,78 +175,87 @@ public GetMapKvpRequestReader(WMS wms) {
...
-                    HttpClientBuilder.create().setDefaultRequestConfig(requestConfig).build();
+            builder = HttpClientBuilder.create().setDefaultRequestConfig(requestConfig);
+        }
+        return builder.build();
+    }
+
+    private void closeHttpClient() throws IOException {
+        this.cacheCfg = null;
+        @SuppressWarnings("PMD.CloseResource") // closing it right below
+        CloseableHttpClient client = this._httpClient;
```
mprins, line 254: this seems a bit ... convoluted as `this._httpClient;` is defined as a CloseableHttpClient in https://github.com/geoserver/geoserver/pull/5986/files#diff-904a705d3e94b6ec62a0e1d7b2526cb8399ed952e44dcb59be56cd3a879ee360R133

so how about `null` checking and closing that instead of creating a new CloseableHttpClient instance `client` only to call close on that?

groldan: It's not creating a new CloseableHttpClient instance at all, it's storing the instance variable as a local variable, nullifying the instance variable, and then closing the local variable if it wasn't null to start with. A simple way to not try to close the instance variable twice in a multi-threaded scenario.

groldan: Though I can see how that can fail too...

groldan: Switched to good old double checked locking with volatile semantics. please re-review.

mprins (approved):

PR #6015, opened 2022-06-30
[GEOS-10539] DescribeLayer typeName is no longer workspace qualified

mprins:
[GEOS-10539](https://osgeo-org.atlassian.net/browse/GEOS-10539)

- [x] Update existing workspace qualified test to fail when typeName is not fully qualified and add test for non-qualified workspace that does the same check
- [x]  fix DescribeLayerTransformer to output fully qualified typenames

  

# Checklist

### DIFF
--- a/src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/DescribeLayerTest.java
+++ b/src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/DescribeLayerTest.java
@@ -74,5 +74,80 @@ public class DescribeLayerTest extends WMSTestSupport {
         Element e = (Element) dom.getElementsByTagName("LayerDescription").item(0);
         String attribute = e.getAttribute("owsURL");
         assertTrue(attribute.contains("sf/wfs"));
+
+        attribute = e.getAttribute("owsType");
+        assertEquals("WFS", attribute);
+
+        e = (Element) dom.getElementsByTagName("Query").item(0);
+        String typeName = e.getAttribute("typeName");
+        assertEquals("unexpected fully qualified typename", "sf:PrimitiveGeoFeature", typeName);
+    }
+
+    @Test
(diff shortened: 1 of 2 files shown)


src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/DescribeLayerTest.java
```diff
@@ -74,5 +74,74 @@ public void testWorkspaceQualified() throws Exception {
...
+
+        Element e = (Element) dom.getElementsByTagName("LayerDescription").item(0);
+        String attribute = e.getAttribute("owsURL");
+        assertTrue(attribute.contains("wcs"));
+        attribute = e.getAttribute("owsType");
+        assertEquals("WCS", attribute);
+
+        e = (Element) dom.getElementsByTagName("Query").item(0);
+        String typeName = e.getAttribute("typeName");
+        assertEquals("unexpected fully qualified typename", "wcs:World", typeName);
```
aaime, line 125: Hum... this seems correct, but at the same time, it's not completely... the reason is that the layer is not named in the same way in the various WCS versions. In WCS 1.0 and 1.1 I believe it's ``wcs:World``, but in the WCS 2.0 column are not allowed in coverage names, and the id would be ``wcs__World`` instead.

I don't think it's solvable to be honest, the DTD for DescribeLayer does not allow to specify a version. Just noting the mismatch.

Oh, the error message seems incorrect, shouldn't it have been "Unexpected un-qualified typename"?).

aaime (commented):
Looks good, the WCS situation seems to be a bit complicated, took notes below.

src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/DescribeLayerTest.java
mprins: yes... and it could also be an empty string; I'll just use "unexpected typename" instead

PR #6016, opened 2022-06-30
[GEOS-10570] Deleting a style in a Hazelcast cluster renames the styles directory

sikeoka:
[GEOS-10570](https://osgeo-org.atlassian.net/browse/GEOS-10570)

This PR adds a check to removeStyle to skip renaming the file if the style's filename points to a directory instead of a regular file.

  

# Checklist


aaime:
The fix is legit, but I'm still feeling like I'm missing something... why are we getting an event with a directory name rather than a file name? Is there also a bug someone down in the hazelcast/jdbcstore/jdbcconfig module?

sikeoka:
The event has a null filename since the hazelcast remove events create a proxy object with only the resource id and name.  Using the directory name is just a way to trigger the bug without a hazelcast cluster.

PR #6021, opened 2022-07-04
[GEOS-10568] Fix module modularity for GWC

ahmedababnehgeo:
[GEOS-10568](https://osgeo-org.atlassian.net/browse/GEOS-10568)

We separated the REST APIs into another module. The new module name is `gs-gwc-rest`.

# Checklist

### DIFF
--- a/src/gwc-rest/pom.xml
+++ b/src/gwc-rest/pom.xml
@@ -1,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.geoserver</groupId>
+    <artifactId>geoserver</artifactId>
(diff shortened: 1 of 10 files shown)


ahmedababnehgeo:
@taba90 Ready for review

taba90 (changes requested):
Looks to me that there are some things that are missing:
- the REST module has been separated by the core part but no container module for gwc-core and gwc rest was added.
- this class https://github.com/geoserver/geoserver/blob/fe4a8909d51df12f13dc0f3c7e368780add2a6e2/src/gwc/src/main/java/org/geoserver/gwc/layer/GWCGeoServerRESTConfigurationProvider.java should probably go in the new rest module as well as it is meant for serialization and deserialization of TileLayer instances.
- the dependency of gwc-rest should go also in the root pom.xml of GeoServer
- there are some community/extensions modules extending the gwc functionality, once the refactoring happened also for the gwc core the gs-gwc depedency in those modules should be updated as well in their pom.xml

@aaime any other feedback from you, or correction to what I wrote?

ahmedababnehgeo:
No association was found between the `gwc-rest` and any of the community modules. We agreed to create the `gwc-rest` module on the same level of `gwc`. The rest of the comments have been addressed.

taba90:
@aaime builds are now green and your feedback has been applied. I've tested this locally and seems to work fine.

taba90 (approved):

aaime:
Merged, thanks!

PR #6057, opened 2022-07-13
[GEOS-10584] Do not close origional inputstream when logging request body

jodygarnett:
[GEOS-10584](https://osgeo-org.atlassian.net/browse/GEOS-10584)

BufferedRequestStream now respects close() clearing internal myInputStream reference.

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/filters/BufferedRequestStream.java
+++ b/src/main/src/main/java/org/geoserver/filters/BufferedRequestStream.java
@@ -8,6 +8,7 @@ package org.geoserver.filters;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.logging.Logger;
 import javax.servlet.ServletInputStream;
 
 /**
@@ -31,6 +32,9 @@ public class BufferedRequestStream extends ServletInputStream {
 
     @Override
     public int readLine(byte[] b, int off, int len) throws IOException {
+        if (myInputStream == null) {
+            throw new IOException("Stream closed");
+        }
         int read;
         int index = off;
         int end = off + len;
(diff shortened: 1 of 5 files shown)


ianturton:
lgtm

src/main/src/test/java/org/geoserver/filters/BufferedRequestStreamTest.java
```diff
@@ -32,6 +34,32 @@ public void testReadLine() throws Exception {
...
+
+        myBRS.close();
+
+        try {
+            amountRead = myBRS.readLine(b, off, len);
+            String s3 = new String(b, 0, amountRead);
+            assertEquals(myTestString, s3);
+            fail("Buffered Request Stream should already be closed");
+        } catch (IOException closed) {
+            // expected
```
juanluisrp, line 62: You aren't really testing that the exception is triggered. Maybe you should use the JUnit `@Test(expected = IOException.class)` method annotation instead of just `@Test`

src/main/src/main/java/org/geoserver/filters/BufferedRequestStream.java
```diff
@@ -48,11 +51,61 @@ public int readLine(byte[] b, int off, int len) throws IOException {
...
+    public void close() throws IOException {
+        if (myInputStream != null) {
+            try {
+                myInputStream.close();
+            } finally {
+                myInputStream = null;
+            }
+        } else {
+            throw new IOException("Stream already closed");
+        }
```
juanluisrp, line 86: Usually closing a stream more than once doesn't throw an exception.

src/main/src/test/java/org/geoserver/filters/BufferedRequestStreamTest.java
jodygarnett: Okay; I will check the specific IOException message is present.

jodygarnett:
Thanks for review @juanluisrp, feedback addressed.

src/main/src/main/java/org/geoserver/filters/LoggingFilter.java
```diff
@@ -153,25 +153,25 @@ public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain)
...
-                        float maxBytesPerCharacter = charset.newEncoder().maxBytesPerChar();
-                        int byteSize = (int) (requestLogBufferSize * (double) maxBytesPerCharacter);
-                        byte[] reqCharacters = new byte[byteSize];
-                        BufferedInputStream bufferedStream = new BufferedInputStream(is);
-                        bufferedStream.mark(byteSize);
-                        bufferedStream.read(reqCharacters, 0, byteSize);
-                        body = new String(reqCharacters, encoding).trim();
-                        bufferedStream.reset();
-                        req = new BufferedRequestWrapper(hreq, encoding, bufferedStream);
+                    InputStream is = hreq.getInputStream();
```
aaime, line 156: Surprised PMD is not complaining (well the build is undergoing, it still might).
In any case, I'd add a comment explaining why this is not using a try-with-resource (not to me, here, I know why, but in the code, as a comment, for the future developer that might read and wonder).

src/main/src/main/java/org/geoserver/filters/LoggingFilter.java
```diff
@@ -153,25 +153,25 @@ public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain)
...
+
+                    Charset charset = Charset.defaultCharset();
+                    try {
+                        charset = Charset.forName(encoding);
+                    } catch (IllegalCharsetNameException icn) {
+                        logger.info(
+                                "Request character set not recognized, defaulting to ISO-8859-1");
```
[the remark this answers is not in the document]
aaime: Yup

aaime:
And oh, noticed that there is no test for the actual important behavioral change: LoggingFilter does not close the stream anymore. Think that should go in LoggingFilterTest, using mocks to verify close is not called.

src/main/src/main/java/org/geoserver/filters/LoggingFilter.java
jodygarnett: okay something like:
```
// Stream opened for BufferedRequestWrapper use - do not close
```

aaime: I'd go for 

```
// The HTTPServletResponse stream should not be closed (Tomcat would complain), so avoid usage or try-with-resources
```

jodygarnett:
> And oh, noticed that there is no test for the actual important behavioral change: LoggingFilter does not close the stream anymore. Think that should go in LoggingFilterTest, using mocks to verify close is not called.

Okay that is done now also; not pretty but it will catch LoggingFilter closing the original stream. Done by only buffering have then bytes; and then making a servlet to consume and echo everything.

PR #6064, opened 2022-07-15
[GEOS-10587] Allow DataStore to auto disable on connection failure

taba90:
[GEOS-10587](https://osgeo-org.atlassian.net/browse/GEOS-10587)

This pr adds a property to DataStore in order to allow, when flagged, the autodisabling when connection fails.

  

# Checklist


doc/en/user/source/data/webadmin/stores.rst
```diff
@@ -47,6 +47,7 @@ The basic information is common for all formats.
 * **Data Source Name** - the store name as listed on the view page
 * **Description** - (optional) a description that displays in the administration interface
 * **Enabled** - enables or disables access to the store, along with all datasets defined for it
+* **Auto disable on connection failure** - auto disables access to the store if any issue happens when geoserver try to acquire access to the store
```
aaime, line 50:
```suggestion
* **Auto disable on connection failure** - auto disables access to the store if any issue happens when GeoServer tries to acquire access to the store
```

src/main/src/main/java/org/geoserver/catalog/impl/StoreInfoImpl.java
```diff
@@ -47,6 +47,8 @@ public abstract class StoreInfoImpl implements StoreInfo {
 
     protected Date dateModified;
 
+    protected boolean disableOnConnFailure;
```
aaime, line 50: New field, but equals and hashCode have not been updated.

src/main/src/main/java/org/geoserver/catalog/StoreInfo.java
```diff
@@ -170,4 +170,18 @@ public interface StoreInfo extends CatalogInfo {
      */
     // <T extends Resource> Iterator<T> getResources(ProgressListener monitor) throws IOException;
 
+    /**
+     * Get the auto disable flag on connection failure.
```
aaime, line 174:
```suggestion
     * When true, the {@link ResourcePool} will automatically disable the store on connection failure. Defaults to false.
```

(I should have linked to the Resource

src/main/src/main/java/org/geoserver/security/config/BaseSecurityNamedServiceConfig.java
```diff
@@ -94,6 +94,17 @@ public SecurityConfig clone(boolean allowEnvParametrization) {
                     && gsEnvironment != null
                     && GeoServerEnvironment.allowEnvParametrization()) {
                 target.setName((String) gsEnvironment.resolveValue(name));
+                /*try {
```
aaime, line 97: Unrelated change, please remove.

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -582,6 +588,12 @@ public DataAccessFactory getDataStoreFactory(DataStoreInfo info) throws IOExcept
         }
     }
 
+    private void disableStoreInfo(WorkspaceInfo ws, String name) {
```
aaime, line 591: The documentation, GUI and REST configuration work covers all types of store. And yet, the ResoucePool/ResourcePoolTest changes seem to only cover DataStore implementations, I don't see the equivalent for the other types of store.

aaime (changes requested):
@taba90 looks like the development was cut short and some bits are missing

taba90:
@aaime thanks for the review and sorry the mess. I've pushed a branch with uncommitted changes.
Your feedback have been applied.

aaime:
Looks good, thanks. Merging!

jodygarnett:
I was not keen on this functionality/workflow - see comments on jira ticket.

aaime:
@jodygarnett you did not want the functionality to be merged and all you said on the ticket was:

```
I thought they already did this? Perhaps only during initial connection attempt?

So what is the user experience you are trying for here …
```

Next time try to be clearer. That said, I don't think you understand what the flag is doing, or what is the default behavior of it (it's off by default).

PR #6014, opened 2022-06-30
[GEOS-10567] Fix module modularity for OpenIdConnect and oauth2-core

ahmedababnehgeo:
[GEOS-10567](https://osgeo-org.atlassian.net/browse/GEOS-10567)

Refactor the OpenId Connect community module separating the `web` functionality away form the core functionality each in a separate module.

# Checklist

- [x] I have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/main/CONTRIBUTING.md).
- [x] I have sent a [Contribution Licence Agreement](https://docs.geoserver.org/latest/en/developer/policies/committing.html) (not required for small changes, e.g., fixing typos in documentation).
- [x] First PR targets the `main` branch (backports managed later; ignore for branch specific issues).
- [x] All the build checks are green ([see automated QA checks](https://docs.geoserver.org/latest/en/developer/qa-guide/index.html)).

For core and extension modules:

- [ ] New unit tests have been added covering the changes.
- [x] [Documentation](https://github.com/geoserver/geoserver/tree/main/doc/en/user/source) has been updated (if change is visible to end users).
- [x] The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers).
- [x] There is an issue in the [GeoServer Jira](https://osgeo-org.atlassian.net/browse/GEOS/summary) (except for changes that do not affect administrators or end users in any way).
- [x] Commit message(s) must be in the form ``[GEOS-XYZWV] Title of the Jira ticket``.
- [x] Bug fixes and small new features are presented as a single commit.
- [x] Each commit has a single objective (if there are multiple commits, each has a separate JIRA ticket describing its goal).

PR #6065, opened 2022-07-15
[GEOS-10588] avoid community profiles duplicating modules

jodygarnett:
[GEOS-10588](https://osgeo-org.atlassian.net/browse/GEOS-10588)

Modules need to be accessed using the same parent / child relationships regardless of the maven profile used. This fixes an inconsistency introduced by https://github.com/geoserver/geoserver/pull/6014 

# Checklist


jodygarnett:
I would like to see this back ported prior to the 2.21.1 release

bradh:
Looks OK. I'd like to see the redundant profiles removed, but that can happen later.

jodygarnett:
> Looks OK. I'd like to see the redundant profiles removed, but that can happen later.

I am not sure which (if any) profiles are redundant? Even though `community/security/pom.xml` has two identical profiles - they each have a different use.  `communityRelease` shortlists modules needed for nightly release bundles.

And `oauth2-all` happens to include the same modules *today* but that may not always be the case.

PR #6069, opened 2022-07-17
[GEOS-10533] Review startup warning, info and config logging

jodygarnett:
[GEOS-10533](https://osgeo-org.atlassian.net/browse/GEOS-10533)

Also cut back on shutdown messages

Message on the management of threads and calling of destroy() are intended for developers and have been reduced to fine logging level.

See also https://github.com/GeoWebCache/geowebcache/pull/1073

  

# Checklist

### DIFF
--- a/src/gwc/src/main/java/org/geoserver/gwc/config/GeoserverXMLResourceProvider.java
+++ b/src/gwc/src/main/java/org/geoserver/gwc/config/GeoserverXMLResourceProvider.java
@@ -13,21 +13,22 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.List;
 import java.util.Optional;
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
+import java.util.logging.Logger;
 import org.geoserver.platform.GeoServerExtensions;
 import org.geoserver.platform.resource.Resource;
 import org.geoserver.platform.resource.ResourceStore;
 import org.geoserver.platform.resource.Resources;
 import org.geoserver.util.IOUtils;
+import org.geotools.util.logging.Logging;
 import org.geowebcache.config.ConfigurationException;
 import org.geowebcache.config.ConfigurationResourceProvider;
 import org.geowebcache.config.XMLFileResourceProvider;
 import org.geowebcache.storage.DefaultStorageFinder;
 
+/** Adapt GeoWebCache ConfigurationResourceProvider to GeoServer Resource API. */
 public class GeoserverXMLResourceProvider implements ConfigurationResourceProvider {
 
-    private static Log LOGGER = LogFactory.getLog(GeoserverXMLResourceProvider.class);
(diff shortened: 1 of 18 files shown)


src/gwc/src/main/java/org/geoserver/gwc/config/GeoserverXMLResourceProvider.java
```diff
@@ -49,7 +49,7 @@ public GeoserverXMLResourceProvider(
             throws ConfigurationException {
         this.configFileName = configFileName;
         this.configDirectory = inferConfigDirectory(resourceStore, providedConfigDirectory);
-        LOGGER.info(
+        LOGGER.debug(
```
bradh, line 52:
```suggestion
        LOGGER.config(
```

src/gwc/src/main/java/org/geoserver/gwc/config/GeoserverXMLResourceProvider.java
```diff
@@ -154,7 +154,7 @@ private Resource findOrCreateConfFile() throws IOException {
         Resource xmlFile = findConfigFile();
 
         if (Resources.exists(xmlFile)) {
-            LOGGER.info("Found configuration file in " + configDirectory.path());
+            LOGGER.debug("Found configuration file in '" + configDirectory.path() + "'");
```
bradh, line 157:
```suggestion
            LOGGER.config("Found configuration file in '" + configDirectory.path() + "'");
```

bradh (approved):
Couple of suggestions for consistency in level. Otherwise OK.

src/gwc/src/main/java/org/geoserver/gwc/config/GeoserverXMLResourceProvider.java
jodygarnett: Sadly I tried that, but this is a query method which is called multiple times. So the log ends up with the same message multiple times.

src/gwc/src/main/java/org/geoserver/gwc/config/GeoserverXMLResourceProvider.java
jodygarnett: This one may be appropriate, I will test and report back.

jodygarnett: It was actually the other way around, this oone has to stay at debug.

I found that this was using the log4j 2 api directly and updated it to follow our logging standards.

src/gwc/src/main/java/org/geoserver/gwc/config/GeoserverXMLResourceProvider.java
jodygarnett: Actually this was is okay, adds a little noise in the logs but does not repeat; changed to config logging level.

PR #6071, opened 2022-07-18
Add missing swagger docs for associating users with groups.

petersmythe:
Add missing swagger docs for associating users with groups. 
Also fixed (broken) plural endpoints to (working) singular

# Checklist

### DIFF
--- a/doc/en/api/1.0.0/usergroup.yaml
+++ b/doc/en/api/1.0.0/usergroup.yaml
@@ -100,26 +100,6 @@ paths:
               description: ID of the newly created user
               type: string
   /usergroup/user/{user}:
-    get:
-      operationId: userDefaultGet
-      tags:
-       - "UserGroup"
-      summary: Query a user
-      description: Not Implemented.  Query a user in the default user/group service.
-      produces:
-        - application/xml
-        - application/json
-      parameters:       
-        - name: user
-          description: the name of the user
-          in: path
-          required: true
-          type: string
-      responses:
-        200:
-          description: OK
-          schema:
-            $ref: "#/definitions/User"  
     post:
       operationId: userDefaultPost
       tags:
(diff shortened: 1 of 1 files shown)


petersmythe:
> LGTM, but I don't have time to test/validate

Mark, thanks for your review

I hope that these sample curl commands (working on the default data) will help you, or someone else, validate the changes:

* Create one group named group_abc	
	curl -X POST http://admin:geoserver@localhost:8080/geoserver/rest/security/usergroup/group/group_abc -is	
* Ditto, test plural endpoint does not work	
	curl -X POST http://admin:geoserver@localhost:8080/geoserver/rest/security/usergroup/groups/group_def -is	
* Add the group to the admin user	
	curl -X POST http://admin:geoserver@localhost:8080/geoserver/rest/security/usergroup/user/admin/group/group_abc -is	
* Query all users for a group	
	curl -X GET http://admin:geoserver@localhost:8080/geoserver/rest/security/usergroup/group/group_abc/users -is	
* Query all groups for a user	
	curl -X GET http://admin:geoserver@localhost:8080/geoserver/rest/security/usergroup/user/admin/groups -is	
* Remove the group from the admin user	
	curl -X DELETE http://admin:geoserver@localhost:8080/geoserver/rest/security/usergroup/user/admin/group/group_abc -is	
* Delete the group	
	curl -X DELETE http://admin:geoserver@localhost:8080/geoserver/rest/security/usergroup/group/group_abc -is

doc/en/api/1.0.0/usergroup.yaml
```diff
@@ -101,15 +99,14 @@ paths:
...
               type: string
-  /usergroup/users/{user}:
+  /usergroup/user/{user}:
     get:
       operationId: userDefaultGet
       tags:
        - "UserGroup"
       summary: Query a user
-      description: Query a user in the default user/group service.
+      description: Not Implemented.  Query a user in the default user/group service.
```
aaime, line 108: I'm not sure about adding "Not implemented" in the description... wouldn't it be better to just remove it?
The reason I'm saying this, is that the default rendering does not show the description, and one could think the method is actually available, e.g.:

image

If the entry is not there instead, it's probably clearer that it's not available.

petersmythe: Yes, I agree @aaime.  I just didn't want to delete what seemed like an important method "Query a user" without flagging the deficiency in some way.  Commit added.

jodygarnett:
> I hope that these sample curl commands (working on the default data) will help you, or someone else, validate the documentation

Those are very useful examples; I recently recorded my own "testing" commands as part of the importer documentation so I could literal include the bash commands.

* https://github.com/geoserver/geoserver/blob/main/doc/en/user/source/extensions/importer/files/test.sh
* https://github.com/geoserver/geoserver/blob/main/doc/en/user/source/extensions/importer/rest_examples.rst

If you like the approach you could use it to preserve your examples

PR #6074, opened 2022-07-19
[GEOS-10591] XML view params format implementation

fernandor777:
[GEOS-10591](https://osgeo-org.atlassian.net/browse/GEOS-10591)

JIRA issue:  
https://osgeo-org.atlassian.net/browse/GEOS-10591

XML view parameters alternative format implementation.

# Checklist

### DIFF
--- a/doc/en/user/source/data/database/sqlview.rst
+++ b/doc/en/user/source/data/database/sqlview.rst
@@ -151,6 +151,29 @@ Parameters can be provided for multiple layers by separating each parameter map 
 
 The number of parameter maps must match the number of layers (featuretypes) included in the request.
 
+Using XML View parameters format
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Aside the default SQL view parameters format, an XML format is available by using the request parameter/value: 
+
+  ``&viewParamsFormat=XML``
+
+XML alternative format example:
+
+  ``&viewParams=<VP><PS><P n="mmsi">538008302,244060802,538008505</P><P n="mmsi">22,44</P></PS><PS/><PS><P n="csvInput">acv,rrp;1,0;0,7;22,1</P></PS></VP>``
+
+viewParamsFormat new optional parameter definition:
+  - Selects the view parameters format implementation to use by its identifier.
+  - Optional parameter, if not set the default character separated format will be used supporting backward compatibility.
+
+XML tags/attributes definition:
+  - VP: the root xml element tag for View Params.  This was added to comply with the XML format validation.
+  - PS: the parameters set XML element for the positional layer.  If there are no parameters for the current layer this must be set as empty element <PS/>
+  - P: the parameter definition XML element, including the parameter name as the n attributer and the value as the XML element child value text.
+  - n: the parameter name attribute inside the 'P' parameter definition element.
+
+Note: If a layer doesn't have parameters to set we just set an empty PS element as seen on the example: <PS/>
+
 Parameters and validation
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/doc/en/user/source/data/database/sqlview.rst
+++ b/doc/en/user/source/data/database/sqlview.rst
@@ -172,7 +172,7 @@ XML tags/attributes definition:
   - P: the parameter definition XML element, including the parameter name as the n attributer and the value as the XML element child value text.
   - n: the parameter name attribute inside the 'P' parameter definition element.
 
-Note: If a layer doesn't have parameters to set we just set an empty PS element as seen on the example: <PS/>
+Note: If a layer doesn't have parameters to be set, just provide an empty ``PS`` element : ``<PS/>``
 
 Parameters and validation
 ^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/doc/en/user/source/data/database/sqlview.rst
+++ b/doc/en/user/source/data/database/sqlview.rst
@@ -162,9 +162,9 @@ XML alternative format example:
 
   ``&viewParams=<VP><PS><P n="mmsi">538008302,244060802,538008505</P><P n="mmsi">22,44</P></PS><PS/><PS><P n="csvInput">acv,rrp;1,0;0,7;22,1</P></PS></VP>``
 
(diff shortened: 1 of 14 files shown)


src/ows/src/main/java/org/geoserver/ows/kvp/ViewParamsKvpParser.java
```diff
@@ -36,23 +48,61 @@ public void setApplicationContext(ApplicationContext applicationContext) throws
...
+
+    private ViewParamsFormatParser getRequestFormatParser() {
+        List<ViewParamsFormatParser> formatParsers =
+                GeoServerExtensions.extensions(ViewParamsFormatParser.class, applicationContext);
+        String formatParserParameterValue = getFormatParserParameterValue();
+        LOGGER.log(Level.FINE, "Using format parser identifier: {0}", formatParserParameterValue);
+        return formatParsers.stream()
+                .filter(fp -> fp.getIdentifier().equals(formatParserParameterValue))
+                .findFirst()
+                .get();
```
aaime, line 64: This one will fail if ``viewParamsFormat`` contains a bogus value. The end user should see a ServiceException with the right code, ``InvalidParameterValue``, and locator, ``viewParamsFormat``, but I believe with the current  code will end up returning a generic error instead.

Please add a test covering invalid values too, check the right OGC exception is returned, and see the ``ServiceException`` class and its ``INVALID_PARAMETER_VALUE`` constant.

src/ows/src/main/java/org/geoserver/ows/kvp/view/XMLViewParamsUtils.java
```diff
@@ -0,0 +1,30 @@
...
+    private XMLViewParamsUtils() {}
+
+    /**
+     * Parses the XML string into a ViewParamsRoot entity bean.
+     *
+     * @param xmlStr the XML string to parse
+     * @return the parsed instance
+     */
+    public static ViewParamsRoot parseViewParams(String xmlStr) {
+        XmlMapper mapper = new XmlMapper();
```
aaime, line 21: If it has any similarity to an ObjectMapper, it's expensive to create but also thread-safe, so it can be kept in a static field.

doc/en/user/source/data/database/sqlview.rst
```diff
@@ -151,6 +151,29 @@ Parameters can be provided for multiple layers by separating each parameter map
...
+
+  ``&viewParamsFormat=XML``
+
+XML alternative format example:
+
+  ``&viewParams=<VP><PS><P n="mmsi">538008302,244060802,538008505</P><P n="mmsi">22,44</P></PS><PS/><PS><P n="csvInput">acv,rrp;1,0;0,7;22,1</P></PS></VP>``
+
+viewParamsFormat new optional parameter definition:
+  - Selects the view parameters format implementation to use by its identifier.
+  - Optional parameter, if not set the default character separated format will be used supporting backward compatibility.
```
aaime, line 167:
```suggestion
``viewParamsFormat`` new optional parameter definition:
  - Selects the view parameters format, valid implementation values are ... <complete here>
  - It's an optional parameter, if not set the default character separated format will be used supporting backward compatibility.
```

doc/en/user/source/data/database/sqlview.rst
```diff
@@ -151,6 +151,29 @@ Parameters can be provided for multiple layers by separating each parameter map
...
+
+XML tags/attributes definition:
+  - VP: the root xml element tag for View Params.  This was added to comply with the XML format validation.
+  - PS: the parameters set XML element for the positional layer.  If there are no parameters for the current layer this must be set as empty element <PS/>
+  - P: the parameter definition XML element, including the parameter name as the n attributer and the value as the XML element child value text.
+  - n: the parameter name attribute inside the 'P' parameter definition element.
```
aaime, line 173:
```suggestion
  - ``VP``: the root XML element tag for View Params.  This ensures XML validity (an XML document must have a single root element).
  - ``PS``: contains the parameters for a given layer (by position).  If there are no parameters for the current layer this must be set as an empty element, e.g. ``<PS/>``
  - ``P``: the parameter definition XML element, including the parameter name as the ``n`` attribute and the value as its text content.
  - ``n``: the parameter name attribute inside the ``P`` element.
```

doc/en/user/source/data/database/sqlview.rst
```diff
@@ -151,6 +151,29 @@ Parameters can be provided for multiple layers by separating each parameter map
...
+
+Note: If a layer doesn't have parameters to set we just set an empty PS element as seen on the example: <PS/>
```
aaime, line 175:
```suggestion
Note: If a layer doesn't have parameters to be set, just provide an empty ``PS`` element : ``<PS/>``
```

aaime (changes requested):
Noted a few comments and suggestions below.

I'd also make sure everything is working by using an integration test, like [SQLViewTest](https://github.com/geoserver/geoserver/blob/b12a833ec3b04ff3606e023681fd90205f22417e/src/wfs/src/test/java/org/geoserver/wfs/SQLViewTest.java). 

Interesting enough, reading the above I've noticed it's also possible to provide view params in a XML POST GetFeature request. The way it's set up, as an attribute, is not easy to extend, as it would require playing with the EMF bindings and the XSD parser. Moreover, having to stick XML inside an attribute value would be really weird. 

I'd just document that XML params can only be used in GET requests.

src/ows/src/main/java/org/geoserver/ows/kvp/ViewParamsKvpParser.java
fernandor777: Applied.

src/ows/src/main/java/org/geoserver/ows/kvp/view/XMLViewParamsUtils.java
fernandor777: Good point, applied.

fernandor777:
Thanks @aaime for your review.  I applied all the hints.

jodygarnett:
Please see email "XML alternative format for view parameters enhancement proposa"  I did not agree this one

PR #6075, opened 2022-07-20
Mask AWS Secret Key in S3 blobstore (was plain text)

petersmythe:
Mask AWS Secret Key in S3 blobstore (currently it is shown in plain text)

I don't believe the AWS Access Key needs to be masked.

# Checklist

### DIFF
--- a/src/extension/gwc-s3/src/main/java/org/geoserver/gwc/web/blob/S3BlobStorePanel.html
+++ b/src/extension/gwc-s3/src/main/java/org/geoserver/gwc/web/blob/S3BlobStorePanel.html
@@ -11,7 +11,7 @@
 			<input id="awsAccessKey" wicket:id="awsAccessKey" type="text" class="text" />
 				
 			<label for="awsSecretKey"><wicket:message key="awsSecretKey"/></label>
-			<input id="awsSecretKey" wicket:id="awsSecretKey" type="text" class="text" />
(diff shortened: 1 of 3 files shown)


petersmythe:
Sample screenshot

image

aaime:
Looked good, but the change on the test picked my interest for a deeper look.

Found this change comes with its own set of challenges:
* The password field is [reset at each render](https://nightlies.apache.org/wicket/apidocs/8.x/org/apache/wicket/markup/html/form/PasswordTextField.html#setResetPassword-boolean-) because "setResetPassword" was not invoked in the form, meaning one has to re-type the password even if only  other fields where modified in a edit
* Invoking ``setResetPassword(true)`` would however make the form insecure and [fall in this trap](https://osgeo-org.atlassian.net/browse/GEOS-9127).
* [PasswordParamPanel has code handling both issues](https://github.com/geoserver/geoserver/blob/b05e92aa12fb0bf7568d28908d88d67d759919ac/src/web/core/src/main/java/org/geoserver/web/data/store/panel/PasswordParamPanel.java#L20). The WriteOnlyModel in there should probably be moved to the ``org.geoserver.web.wicket``panel for general reuse, or maybe a subclass of PasswordTextField that always wraps the user provided model into a WriteOnlyModel could be created, and used across the board.

aaime:
Not all is lost, this PR is being recovered and completed here: https://github.com/geoserver/geoserver/pull/6338

PR #6077, opened 2022-07-22
Clarify that the ENV_PROPERTIES contains the full filename

petersmythe:
Updated documentation to clarify that the ENV_PROPERTIES contains the full filename and not just the directory path, like the 3rd example used to imply.

# Checklist


aaime:
Reading the code, I see a more complex situation:
* The value of the variable is [first tried out as an absolute path](https://github.com/geoserver/geoserver/blob/b1a7f63984c47e0033b1fa7c2ebb98cb0bfaafd1/src/platform/src/main/java/org/geoserver/platform/GeoServerEnvironment.java#L117), if it exist, it's going to be used as is.
* If that lookup fails, then it's used as a [relative path inside the data directory](https://github.com/geoserver/geoserver/blob/b1a7f63984c47e0033b1fa7c2ebb98cb0bfaafd1/src/platform/src/main/java/org/geoserver/platform/GeoServerEnvironment.java#L137)

PR #6078, opened 2022-07-25
[GEOS-10598] XSS vulnerability in the email address field

ahmedababnehgeo:
[GEOS-10598](https://osgeo-org.atlassian.net/browse/GEOS-10598)

In this PR, we preform output encoding to protect against XSS attacks. Furthermore, we do validation on the email field to reduce the risk of attacks. Several techniques, to protect against these attacks can be found here: https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html 

# Checklist

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.java
+++ b/src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.java
@@ -76,7 +76,7 @@ public class GeoServerHomePage extends GeoServerBasePage implements GeoServerUnl
                             "footerMessage",
                             new StringResourceModel(
                                     "GeoServerHomePage.footer", this, new Model(params)));
-            label.setEscapeModelStrings(false);
+            label.setEscapeModelStrings(true);
             add(label);
         }
 
--- a/src/web/core/src/main/java/org/geoserver/web/StringAndInternationalStringPanel.java
+++ b/src/web/core/src/main/java/org/geoserver/web/StringAndInternationalStringPanel.java
@@ -11,6 +11,7 @@ import org.apache.wicket.markup.html.panel.Panel;
 import org.apache.wicket.model.IModel;
 import org.apache.wicket.model.PropertyModel;
 import org.apache.wicket.model.StringResourceModel;
+import org.apache.wicket.validation.IValidator;
 
 /**
  * A reusable component for values that can be provided both as a TextField and as an i18n editable
@@ -32,7 +33,8 @@ public class StringAndInternationalStringPanel extends Panel {
             String labelKey,
             String stringProperty,
             String internationalProperty,
-            WebMarkupContainer labelProvider) {
+            WebMarkupContainer labelProvider,
+            IValidator<String> validator) {
         super(id, model);
         WebMarkupContainer container = new WebMarkupContainer("labelContainer");
         add(container);
@@ -40,6 +42,9 @@ public class StringAndInternationalStringPanel extends Panel {
         TextField<String> title =
                 new TextField<>("stringField", new PropertyModel<>(model, stringProperty));
         add(title);
+        if (validator != null) {
+            title.add(validator);
+        }
 
         InternationalStringPanel<TextField<String>> internationalStringField =
                 new InternationalStringPanel<TextField<String>>(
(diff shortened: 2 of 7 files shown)


ianturton (approved):

ahmedababnehgeo:
@aaime PR is ready for review

src/main/src/main/java/org/geoserver/config/impl/ContactInfoImpl.java
```diff
@@ -152,7 +154,12 @@ public void setAddressType(String addressType) {
 
     @Override
     public String getContactEmail() {
-        return InternationalStringUtils.getOrDefault(contactEmail, internationalContactEmail);
+        String result = contactEmail;
+        if (result == null && internationalContactEmail != null)
+            result = internationalContactEmail.toString(GeoServerDefaultLocale.get());
+        String escapedResult = StringEscapeUtils.escapeHtml4(result);
```
aaime, line 160: This should be done in the UI and in the REST freemarker templates, but not here. For example, a REST call getting the email field, returned in XML or JSON, should not have escaping happening, the field should be returned as-is in those cases.

src/web/core/src/main/java/org/geoserver/web/StringAndInternationalStringPanel.java
```diff
@@ -55,4 +56,38 @@ protected TextField<String> getTextComponent(String id, IModel<String> model) {
...
+    public StringAndInternationalStringPanel(
+            String id,
+            IModel<?> model,
+            String labelKey,
+            String stringProperty,
+            String internationalProperty,
+            WebMarkupContainer labelProvider,
+            IValidator<String> validator) {
+        super(id, model);
+        WebMarkupContainer container = new WebMarkupContainer("labelContainer");
```
aaime, line 69: The addition of validators is done in the right places, but duplicating the entire existing constructor to add them is overkill. I suggest using a single constructor with actual code, a couple of ifs, and the previous constructor would call the new one passing null as the validator.

aaime (changes requested):
Why have the two tests for WMS and WMTS store pages been removed?
I don't agree with the solution, should not be affecting how the mail is saved and handed in core, but only touching the sensitive HTML parts:
- https://github.com/geoserver/geoserver/blob/fe4a8909d51df12f13dc0f3c7e368780add2a6e2/src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.java#L68
- Potentially the freemarker templates, if they are not correctly escaping already

src/web/core/src/main/java/org/geoserver/web/StringAndInternationalStringPanel.java
ahmedababnehgeo: I removed the extra constructor as suggested

ahmedababnehgeo:
@aaime PR has been updated.

src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.java
```diff
@@ -76,7 +76,7 @@ public GeoServerHomePage() {
                             "footerMessage",
                             new StringResourceModel(
                                     "GeoServerHomePage.footer", this, new Model(params)));
-            label.setEscapeModelStrings(false);
+            label.setEscapeModelStrings(true);
```
aaime, line 79: Yup, thanks

src/wfs/src/main/java/org/geoserver/wfs/CapabilitiesTransformer.java
```diff
@@ -1392,8 +1395,9 @@ protected void internationalServiceProvider(ContactInfo contact) {
                                 contact.getInternationalAddressCountry()));
                 element(
                         "ows:ElectronicMailAddress",
-                        internationalContentHelper.getNullableString(
-                                contact.getInternationalContactEmail()));
+                        StringEscapeUtils.escapeXml10(
```
aaime, line 1398: The contact mail is used in other capabilities document generators as well (WCS, WMS) but wondering if this escaping is actually useful, I think the low-level XML encoder here should already be doing the necessary escaping for you. Do you have a way to actually do harm using the XML documents?

ahmedababnehgeo: Good point ... I am unaware of a way to attack using XML document? Do you suggest rolling back this change?!

aaime: If it's not needed, better to back it out, yes (if you find a way to move an attack using XML documents instead, it needs tests and be replicated in all other places a caps document emits the mail address).

ahmedababnehgeo:
@aaime PR ready for review

PR #6086, opened 2022-08-02
[GEOS-10604] Handling empty coverage data

ahmedababnehgeo:
[GEOS-10604](https://osgeo-org.atlassian.net/browse/GEOS-10604)

In this PR we return a fake image when there is no coverage info, instead of returning an exception. We use the grid coverage factory in order to create the image.   

# Checklist

### DIFF
--- a/src/wcs1_0/src/main/java/org/geoserver/wcs/DefaultWebCoverageService100.java
+++ b/src/wcs1_0/src/main/java/org/geoserver/wcs/DefaultWebCoverageService100.java
@@ -7,7 +7,10 @@ package org.geoserver.wcs;
 
 import static org.vfny.geoserver.wcs.WcsException.WcsExceptionCode.InvalidParameterValue;
 
+import it.geosolutions.jaiext.utilities.ImageLayout2;
 import java.awt.Rectangle;
+import java.awt.RenderingHints;
+import java.awt.image.RenderedImage;
 import java.io.IOException;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
@@ -17,6 +20,8 @@ import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.media.jai.Interpolation;
+import javax.media.jai.JAI;
+import javax.media.jai.operator.ConstantDescriptor;
 import net.opengis.gml.CodeType;
 import net.opengis.gml.DirectPositionType;
 import net.opengis.gml.RectifiedGridType;
@@ -49,8 +54,10 @@ import org.geoserver.wcs.response.Wcs10CapsTransformer;
 import org.geoserver.wcs.response.Wcs10DescribeCoverageTransformer;
 import org.geoserver.wcs.responses.CoverageResponseDelegate;
 import org.geoserver.wcs.responses.CoverageResponseDelegateFinder;
+import org.geotools.coverage.CoverageFactoryFinder;
 import org.geotools.coverage.grid.GeneralGridGeometry;
 import org.geotools.coverage.grid.GridCoverage2D;
+import org.geotools.coverage.grid.GridCoverageFactory;
 import org.geotools.coverage.grid.GridEnvelope2D;
 import org.geotools.coverage.grid.GridGeometry2D;
 import org.geotools.coverage.grid.io.AbstractGridFormat;
@@ -63,6 +70,7 @@ import org.geotools.referencing.operation.transform.AffineTransform2D;
 import org.geotools.referencing.util.CRSUtilities;
 import org.geotools.util.DateRange;
 import org.geotools.util.NumberRange;
+import org.geotools.util.factory.GeoTools;
 import org.opengis.coverage.grid.GridCoverage;
 import org.opengis.coverage.grid.GridEnvelope;
 import org.opengis.filter.Filter;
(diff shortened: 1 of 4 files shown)


ahmedababnehgeo:
@dromagnoli PR ready for review

src/wcs1_0/src/main/java/org/geoserver/wcs/DefaultWebCoverageService100.java
```diff
@@ -425,10 +433,34 @@ public GridCoverage[] getCoverage(final GetCoverageType request) {
             //
             coverage = reader.read(readParameters);
             if ((coverage == null) || !(coverage instanceof GridCoverage2D)) {
-                throw new IOException(
-                        "No raster data found in the request (it may be that "
-                                + "the request bbox is outside of the coverage area, or that the filters used "
-                                + "match no portions of it.");
+                Integer[] values = new Integer[1];
```
dromagnoli, line 436: Better to check the datatype of the CoverageInfo and use proper number type based on that.
See the meta.dimensions elements.

src/wcs1_0/src/main/java/org/geoserver/wcs/DefaultWebCoverageService100.java
```diff
@@ -425,10 +433,34 @@ public GridCoverage[] getCoverage(final GetCoverageType request) {
...
+
+                values[0] = 0;
```
dromagnoli, line 438: meta.dimensions also contains null value / nil value / nodata value... Whatever.
I think that I would be better to use that value to fill the coverage.
Think about a Digital Elevation Model storing -9999 as noData.
Better to fill the coverage with -9999 so the end user or the software can recognize the pixels as no valid data.
We can use zero if there isn't any nodata / null value defined.

dromagnoli (changes requested):
I would also create a specific test to validate the new behavior.
You can configure a coverage with some small dataset, fixing the bbox to a greater extent than the available data and finally requesting a coverage in the extended part which isn't actually containing any real data.

src/wcs1_0/src/main/java/org/geoserver/wcs/DefaultWebCoverageService100.java
jratike80: Copying my comment from Jira https://osgeo-org.atlassian.net/browse/GEOS-10604 also here:

The change is against the WCS standard. At least the behavior must be configurable and the default must follow the standard.

ahmedababnehgeo: I believe that the image we are returning is not fake, as it will be filled with no data values. @dromagnoli Do you think we need to change anything in the implementation to be aligned with the WCS standard?

jratike80: I am reading the WCS 2.0 standard. It does not explicitly say what should happen is subset is outside the extent of the coverage. By these excerpts I think that it has been considered that such case should not happen:

> - Domain subsetting is subdivided into trimming and slicing. A trim operation identifies a dimension
> and a lower and upper bound (which both must lie inside the coverage’s domain)...
> - **Requirement /req/core/getCoverage-request-trim-within-extent**:
> Let the extent of the coverage’s gml:Envelope along the dimension
> specified in the trim request range from L to H. Then, for the trim bounds
> trimLow and trimHigh the following shall hold: L <= trimLow <=
> trimHigh <= H.
> 

However, it looks like a note in another requirement deals with partially overlapping subset and extent:

> **Requirement 38 /req/core/getCoverage-response-trimming**:
> ...
> NOTE This requirement does not specify the actual extent of the coverage returned. Possible options
> include: the minimal bounding box of the coverage returned, or the request bounding box. Servers are
> strongly encouraged to deliver the minimal bounding box.

Notice that it is _strongly encouraged_ to deliver the minimal bounding box like Geoserver currently does. Minimal bounding box of out-of-extend subset is nothing and I think that sending an InvalidSubsetting error HTTP 404 is what the standard considers right

>  "Operation request contains an invalid subsetting
> value; either a trim or slice parameter
> value is outside the extent of the coverage
> or, in a trim operation, a lower bound
> is above the upper bound"


But there is also a note that reveals that the case of out-of-extent has been thought to some extent:

> NOTE Extensions to this core may add further functionality over trimming and slicing, including
> further parameters which may extend such functionality. For example, a forthcoming WCS nil values
> extension may specify, by way of an additional parameter, that bounding boxes larger than the coverage
> extent are admissible.

Such nil values extension was never written but maybe Geoserver could take the suggested route  "by way of an additional parameter". Let the current behavior stay but support a vendor specific option like "&response_bbox=full" that makes Geoserver to always use the full range of subsets and fill the out-of-extent area with nodata.

jratike80: I checked also the draft of OGC API Coverages specification https://docs.opengeospatial.org/DRAFTS/19-087.html. It defines in Requirement 22:  req/collections/rc-subset-definition that "If the intervalOrPoint values fall entirely outside the range of valid values defined for the identified axis, a 204 status code SHALL be returned". It is obvious that OGC wants that request that does not hit the data area at all shall return an error.

I do not mean that this change should not be implemented but I suggest to do it in a way that is clearly a vendor option for Geoserver.

dromagnoli: Hi @jratike80.
Thanks for the feedbacks. 
The use case for this code change is to cover a scenario where you have a "dynamic / growing" dataset collection (e.g. an ImageMosaic) made of scattered pieces which can be dynamically added or removed.

So at the very beginning, at the time of publishing, the data admin imposes/specifies a Coverage extent covering a "big" area (for example -180,-90,180,90) so that "theoretically" every position inside that bbox is a valid part of the coverage extent definition, even if for some area there isn't any actual data pixel.

So in that case the subset is theoretically inside the coverage extent.
Trivial example: I define a coverage covering the whole world (imposing the above bbox at configuration time) and I request data for a big chunk of Europe (where I have some data for a few countries). I'm inside the bbox even if I only have data available for a few countries. In that case my extent is still inside the bbox.

A different case will be if I define a coverage covering North America and I'm going to ask data in Europe.
In that case, the request is totally outside the coverage extent definition so the above code change won't apply.
Does it make any sense?

@aaime any comment on this?

aaime: From the point of view of the client, and the one of the standard at the same time, what matters is what is declared in the capabilities document. My understanding is that:

 * If the client makes a request outside of the bounds declared in the capabilities document, it should get an exception
 * If the request is overlapping, the returned data should be at the intersection between the request and declared bounds
 * If the request is inside the declared bounds, then data should be returned, eventually filling with NODATA if there is no actual data in storage

As with other standards, we need to present a consistent picture, and make sure all responses are consistent the data description found in the GetCapabilities response, regardless of how it is actually structured (examples are forcing CRSs, or axis flipping in WFS 1.1, and so on).

bradh: I think I see the goal, but share a sense of unease about the change.

The SRTM example is probably a good case - there really are tiles for -180 to +180 lon (and not quite to the poles in lat). However there are not tiles for every part of that BBOX (e.g. large water bodies). So void filling / no-data is valid for the world. 

I think this is a pretty subtle distinction though, and a potential change in behaviour for an existing system, and we should at least be explaining it in the WCS docs.

jratike80: Sorry, because I cannot really read the code I read the literal description. In https://osgeo-org.atlassian.net/browse/GEOS-10604 the description does not distinct the cases "bbox totally outside extent" and "bbox inside extend but coverage data contains just nodata". The latter case is of course very common with orthophotos etc. especially when filtered by acquisition time.

I agree with the list of three cases that Adrea wrote. If I understand right, this change would affect only the third case, right?

BTW. in the the tiled option of the OGC API Coverages draft it looks like the service should return a 206 error code instead of an empty tile in the all-nodata case https://github.com/opengeospatial/ogcapi-coverages/issues/27#issuecomment-714815434. Because it is still a draft you might want to suggest an alternative for that.

PR #6090, opened 2022-08-03
[GEOS-10607] Links disappearing for the Admin user

ahmedababnehgeo:
[GEOS-10607](https://osgeo-org.atlassian.net/browse/GEOS-10607)

Whenever a user has been assigned to two roles, or even the same role, and one of those roles has ".admin" power on some data, while the other (or even the same) has ".write" power on the data, whenever you click on the "Layer Preview" link the others disappear.  

We add a new check to make sure that admin access is taken into account when retrieving Workspace Access limits.

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
+++ b/src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
@@ -447,8 +447,8 @@ public class DefaultResourceAccessManager implements ResourceAccessManager {
         CatalogMode mode = getMode();
 
         if (readable && writable) {
-            if (AdminRequest.get() == null) {
-                // not admin request, read+write means full acesss
+            if (!adminable && AdminRequest.get() == null) {
+                // It is not an admin request, read+write means full access
                 return null;
             }
         }
--- a/src/main/src/test/java/org/geoserver/security/impl/DefaultResourceAccessManagerTest.java
+++ b/src/main/src/test/java/org/geoserver/security/impl/DefaultResourceAccessManagerTest.java
@@ -1,0 +1,42 @@
+/* (c) 2022 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.security.impl;
+
+import static org.junit.Assert.assertNotNull;
+import static org.mockito.Mockito.mock;
+
+import java.util.Collection;
+import java.util.Collections;
+import org.geoserver.catalog.Catalog;
+import org.geoserver.catalog.WorkspaceInfo;
+import org.geoserver.security.WorkspaceAccessLimits;
+import org.junit.Test;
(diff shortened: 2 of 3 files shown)


taba90:
there is no test on this pr @ahmedababnehgeo. Please add one.

src/security/security-tests/src/test/java/org/geoserver/security/impl/DefaultResourceAccessManagerAuthTest.java
```diff
@@ -38,6 +38,12 @@ public void testWideOpen() throws Exception {
         checkUserAccessFlat(manager, anonymous, true, true);
     }
 
+    @Test
+    public void testAdminProperties() throws Exception {
+        ResourceAccessManager manager = buildAccessManager("admin.properties");
+        checkUserAccessFlat(manager, root, true, true);
```
taba90, line 44: this method is testing that the canAccess methods return true according to the rules you set. I guess this was happening  even before your fix. The test should also test the getAccessLimits for the workspace doesn't return null but the WorkspaceLimits instance.

taba90 (changes requested):

ahmedababnehgeo:
@taba90 Test has been updated

jodygarnett:
@taba90 has your feedback been addressed?

src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
```diff
@@ -446,7 +446,7 @@ public WorkspaceAccessLimits getAccessLimits(Authentication user, WorkspaceInfo
 
         CatalogMode mode = getMode();
 
-        if (readable && writable) {
+        if (readable && writable && !adminable) {
```
aaime, line 449: I'm confused by this change... looking at the ticket, we are dealing with a workspace administrator.
The code authorizing the components visibility should be the [WorkspaceAdminComponentAuthorizer](https://github.com/geoserver/geoserver/blob/6bb431a36a57aa1b6cf68d78bc7dea35b0958812/src/web/core/src/main/java/org/geoserver/web/WorkspaceAdminComponentAuthorizer.java#L52) which should allow visibility when there is at least one adminable workspace. Are you sure that, with this change, links are not getting visible also for users that cannot admin anything?

ahmedababnehgeo: I tested the change for users who have no admin privilege at all, and links are not visible for them. It seems that this change solves the issue without side effects. 

Anyway, I changed the if statement organization, to increase its readability.

ahmedababnehgeo: @aaime In order to make sure that REST API works well, I have tested access to the REST API page after this change. Normal users are denied access to REST API, whereas admins are allowed.

aaime:
Tested manually, everything seems to work correctly. Merging.

PR #6091, opened 2022-08-03
[GEOS-10609] Fix Issue retrieving a workspace style without prefix on JDBCConfig

fernandor777:
[GEOS-10609](https://osgeo-org.atlassian.net/browse/GEOS-10609)

This PR introduces a fix for the reported issue:

https://osgeo-org.atlassian.net/browse/GEOS-10609

"We found an issue when GeoServer use JDBC config catalog implementation, making it unable to use a workspace based style without using the prefix.  JDBC config catalog should follow the same business logic as the default GeoServer catalog implementation."

# Checklist

### DIFF
--- a/src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/catalog/JDBCCatalogFacade.java
+++ b/src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/catalog/JDBCCatalogFacade.java
@@ -650,7 +650,11 @@ public class JDBCCatalogFacade implements CatalogFacade {
     /** @see org.geoserver.catalog.CatalogFacade#getStyleByName(java.lang.String) */
     @Override
     public StyleInfo getStyleByName(String name) {
-        return getStyleByName(NO_WORKSPACE, name);
+        StyleInfo match = getStyleByName(NO_WORKSPACE, name);
+        if (match == null) {
+            match = getStyleByName(ANY_WORKSPACE, name);
+        }
+        return match;
     }
 
     /** @see org.geoserver.catalog.CatalogFacade#getStyles() */
--- a/src/community/jdbcconfig/src/test/java/org/geoserver/jdbcconfig/catalog/CatalogImplWithJDBCFacadeTest.java
+++ b/src/community/jdbcconfig/src/test/java/org/geoserver/jdbcconfig/catalog/CatalogImplWithJDBCFacadeTest.java
@@ -32,6 +32,7 @@ import org.geoserver.catalog.StoreInfo;
 import org.geoserver.catalog.StyleInfo;
 import org.geoserver.catalog.WMTSLayerInfo;
 import org.geoserver.catalog.WMTSStoreInfo;
+import org.geoserver.catalog.WorkspaceInfo;
 import org.geoserver.catalog.impl.CatalogImpl;
 import org.geoserver.catalog.util.CloseableIterator;
 import org.geoserver.jdbcconfig.JDBCConfigTestSupport;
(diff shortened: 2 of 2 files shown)


src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/catalog/JDBCCatalogFacade.java
```diff
@@ -650,7 +650,11 @@ public StyleInfo getStyle(String id) {
     /** @see org.geoserver.catalog.CatalogFacade#getStyleByName(java.lang.String) */
     @Override
     public StyleInfo getStyleByName(String name) {
-        return getStyleByName(NO_WORKSPACE, name);
+        StyleInfo match = getStyleByName(NO_WORKSPACE, name);
+        if (match == null) {
+            match = getStyleByName(ANY_WORKSPACE, name);
```
jodygarnett, line 655: Having two styles in different workspaces with the same name would cause failure.

I am not sure that is a good plan; your scenario is from a virtual web service so you should take care to search using that context.

jodygarnett (changes requested):
Solution does not fix issue

src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/catalog/JDBCCatalogFacade.java
jodygarnett: You indicate expectation that prefix be inferred from virtual web service or prefix, or from layer prefix? If that is the case have a look fixing GetMap Request processing to resolve this issue.

fernandor777: Hi @jodygarnett thanks for your feedback, indeed maybe the description of this fix is too broad and the goal here is more simple.  The idea of this fix is to give JDBC config the same behavior [is already available at GeoServer defatult file based behavior when getting an style by name without a workspace context defined as we can see at DefaultCatalogFacade code](https://github.com/geoserver/geoserver/blob/a945fad3993e8d633d3509aa20ea836f46c7bb99/src/main/src/main/java/org/geoserver/catalog/impl/DefaultCatalogFacade.java#L899-L905).  

You have a point on the suggestion of checking only the current virtual service WS styles or the layer workspace styles, but this would give us a different behavior that the one is already supported on default GeoServer catalog implementation and hence will break backward compatibility.

aaime: +1 for having consistent behavior... we could go and make the behavior "better", but it would have consequences on those upgrading with a previous data dir (stuff that used to work, may not anymore). Which is fine to some extent, but needs to be explained, have some workarounds/upgrade instructions, and we'll take some heat for it, so we need to make sure it's worth it.

fernandor777:
Fixed the scope and goal of the target issue to make JDBC config have the same logic as default GeoServer catalog implementation.

src/community/jdbcconfig/src/main/java/org/geoserver/jdbcconfig/catalog/JDBCCatalogFacade.java
jodygarnett: Making behaviour sane can be a separate improvement.

If this makes the behaviour consistent please update the description, and +1

jodygarnett: I really do not like the idea of searching different workspaces because I have customers that "lock down" workspace to a specific user; and if it finds a style in a different workspace it may not be able to access the icons in that workspace.

aside: I was under the impression for some years that searching for a style in a workspace styles would first search the workspace, and then search the global styles folder only.

sikeoka: If you really wanted to have the same behavior as the default catalog when the same style name exists in multiple workspaces, then I think that something like the following could be used to return a style instead of an exception in that case:
```
Filter filter = equal("name", name);
try (CloseableIterator<StyleInfo> it = db.query(StyleInfo.class, filter, null, 1)) {
    if (it.hasNext()) {
        match = it.next();
    }
}
```
but you would also have to add the sort by on the workspace name since the default catalog uses sorted maps.

aaime: > I really do not like the idea of searching different workspaces because I have customers that "lock down" workspace to a specific user; and if it finds a style in a different workspace it may not be able to access the icons in that workspace.

Well, to me it's even worse than that, in a workspace-specific service, a style belonging to a different workspace should not really be found, security or not. But the current code happily lets people link layers in WS1 to styles in WS2 (try that in the UI)... and since it's allowed, that situation is likely to present in a number of existing data dirs, sigh...

PR #6092, opened 2022-08-05
[GEOS-10608]:Stop flooding logs with InvalidDate (and similar) at Severe

dromagnoli:
[GEOS-10608](https://osgeo-org.atlassian.net/browse/GEOS-10608)

# Checklist

### DIFF
--- a/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
+++ b/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
@@ -1701,17 +1701,33 @@ public class Dispatcher extends AbstractController {
                 logger.log(Level.FINER, "", t);
             }
         } else {
-            logger.log(Level.SEVERE, "", t);
+            String errorCode = null;
+            String message = t.getMessage();
+            Level level = Level.SEVERE;
 
             // unwind the exception stack until we find one we know about
             Throwable cause = t;
             while (cause != null) {
                 if (cause instanceof ServiceException) {
+                    errorCode = ((ServiceException) cause).getCode();
                     break;
                 }
-
                 cause = cause.getCause();
             }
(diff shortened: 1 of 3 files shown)


src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
```diff
@@ -1701,17 +1701,34 @@ void exception(Throwable t, Service service, Request request) {
...
+                // since the service might not be known in advance. Therefore, there
+                // might be the case that it isn't a ServiceException and therefore no
+                // errorCode available. Let's use the error message in that case.
+                errorCode = message;
+            }
+
+            if (errorCode != null && errorCode.toUpperCase().startsWith("INVALID")) {
+                // Log Exceptions dealing with "Invalid" to INFO
+                level = Level.INFO;
+            }
```
aaime, line 1729: errorCode is currently used only to determine the logging level, but one might get fooled by the variable later and we might end up having an exception message in the OWS response... how about using directly the message in the logic that determines the error message instead? Something like:

```java
if ((errorCode != null && errorCode.toUpperCase().startsWith("INVALID")) || (errorCode == null && message != null && message.toUpperCase().startsWith("INVALID")))
```

aaime (changes requested):
Looks ok, and should remove a lot of invalid parameter value reports from the logs (reducing PRODUCTION logging, so that it has less noise and hopefully mostly actual problems).

One bit in the logic below worries me about future maintenance.

src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
dromagnoli: @aaime 
Sounds good. fixed

aaime (approved):

PR #6093, opened 2022-08-05
[GEOS-10610] Selective cache reset on stores and resources, via REST API

aaime:
[GEOS-10610](https://osgeo-org.atlassian.net/browse/GEOS-10610)

See [GSIP-214](https://github.com/geoserver/geoserver/wiki/GSIP-214)

@jodygarnett regarding [GEOS-6637](https://osgeo-org.atlassian.net/browse/GEOS-6637), the ability to recalculate the bounding boxes was already built into the rest API, using a "calculate" attribute on coverages, and a "recalculate" one on feature types. I've added "attributes" and "dimensions" (aka bands) for completeness, and to address some asymmetry between feature types and coverages (feature types may have attributes configured, but often just reflect on the storage setup, coverages always have band configured instead, and won't reload them unless explicitly required... I've made sure one can force reload of such config for both vectors and rasters, while doing a PUT though, since that actually changes the config, it does not simply clear caches).

# Checklist

### DIFF
--- a/doc/en/api/1.0.0/coverages.yaml
+++ b/doc/en/api/1.0.0/coverages.yaml
@@ -859,7 +859,7 @@ paths:
             $ref: "#/definitions/CoverageInfo"
         - name: calculate
           in: query
-          description: 'Comma-separated list of optional fields to calculate. Optional fields include: "nativebbox", "latlonbbox".'
+          description: 'Comma-separated list of optional fields to calculate. Optional fields include: "nativebbox", "latlonbbox", "dimensions" (that is, bands).'
           required: false
           type: array
           collectionFormat: csv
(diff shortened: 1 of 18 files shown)


dromagnoli (approved):
Looks good to me

jodygarnett:
Andrea I hope you can close GEOS-6637 as part of this PR being merged; mark it a duplicate of GEOS-10610 etc...

doc/en/api/1.0.0/coverages.yaml
```diff
@@ -859,7 +859,7 @@ paths:
             $ref: "#/definitions/CoverageInfo"
         - name: calculate
           in: query
-          description: 'Comma-separated list of optional fields to calculate. Optional fields include: "nativebbox", "latlonbbox".'
+          description: 'Comma-separated list of optional fields to calculate. Optional fields include: "nativebbox", "latlonbbox", "dimensions" (that is, bands).'
```
jodygarnett, line 862: nice!

doc/en/api/1.0.0/featuretypes.yaml
```diff
@@ -631,7 +632,7 @@ paths:
...
...roperties of feature types are automatically recalculated when necessary. In particular, the native bounding box is recalculated when the projection or projection policy are changed, and the lat/lon bounding box is recalculated when the native bounding box is recalculated, or when a new native bounding box is explicitly provided in the request. (The native and lat/lon bounding boxes are not automatically recalculated when they are explicitly included in the request.) In addition, the client may explicitly request a fixed set of fields to calculate, by including a comma-separated list of their names in the recalculate parameter.  The empty parameter 'recalculate=' is useful avoid slow recalculation when operating against large datasets as 'recalculate=' avoids calculating any fields, regardless of any changes to projection, projection policy, etc. The nativebbox parameter 'recalculate=nativebbox' is used recalculates the native bounding box, while avoiding recalculating the lat/lon bounding box. Recalculate parameters can be used in together - 'recalculate=nativebbox,latlonbbox' can be used after a bulk import to recalculate both the native bounding box and the lat/lon bounding box.
```
jodygarnett, line 634: That is a hard one to take, losing the customized attributes; any hesitation about this being the right thing to do?

jodygarnett (approved):
Great job :) I really appreciate the rest api being direct calls to resource pool with no logic of their own.

doc/en/api/1.0.0/featuretypes.yaml
aaime: No hesitation so far... let me explain.

If I just want to reset the knowledge that GeoServer has about the original attributes, I'll just use reset, and pray the 
However, if those changed, there is a significant chance the customization will not work anymore (it might referer to attributes that are no longer there). At that point there are two choices:
* Reset the customization and start fresh, maybe to customize later (I'm imagining a UI workflow to do that)
* Somehow rebuild a sensible customization right away, doing a PUT against the FeatureType resource that already contains the new list of attribute infos.

jodygarnett:
Is this waiting on anything? It is a great improvement.

aaime:
It is waiting on votes: @tbarsballe, @smithkm and @jratike80 haven't voted on the proposal yet, and the 10 days wait has not expired either. Monday it will be ready for merge, unless negative votes show up.

PR #5807, opened 2022-04-11
[GEOS-10462] Upgrade to org.apache.httpcomponents:httpclient completed

jodygarnett:
[GEOS-10462](https://osgeo-org.atlassian.net/browse/GEOS-10462)

The dependency on ``commons-httpclient`` is no longer required. In addition to building, I checked ``dependency:tree`` on main and web to look out for any remaining transitive dependencies.

This depends on [GEOT-5859](https://osgeo-org.atlassian.net/browse/GEOT-5859) and https://github.com/geoserver/geoserver/pull/5702

# Checklist

- [x] I have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/main/CONTRIBUTING.md).
- [x] I have sent a [Contribution Licence Agreement](https://docs.geoserver.org/latest/en/developer/policies/committing.html) (not required for small changes, e.g., fixing typos in documentation).
- [x] First PR targets the `main` branch (backports managed later; ignore for branch specific issues).
- [x] All the build checks are green ([see automated QA checks](https://docs.geoserver.org/latest/en/developer/qa-guide/index.html)).

For core and extension modules:

- [ ] New unit tests have been added covering the changes.
- [ ] [Documentation](https://github.com/geoserver/geoserver/tree/main/doc/en/user/source) has been updated (if change is visible to end users).
- [ ] The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers).
- [x] There is an issue in the [GeoServer Jira](https://osgeo-org.atlassian.net/browse/GEOS/summary) (except for changes that do not affect administrators or end users in any way).
- [x] Commit message(s) must be in the form ``[GEOS-XYZWV] Title of the Jira ticket``.
- [x] Bug fixes and small new features are presented as a single commit.
- [x] Each commit has a single objective (if there are multiple commits, each has a separate JIRA ticket describing its goal).

PR #6104, opened 2022-08-11
[GEOS-10462] Upgrade commons-httpclient 3.1 to org.apache.httpcomponents:httpclient:4.5.13

sikeoka:
[GEOS-10462](https://osgeo-org.atlassian.net/browse/GEOS-10462)

This PR attempts to remove the old commons-httpclient dependency from the main GeoServer package.  commons-httpclient is removed from the main module, added to the Importer extension as a test dependency and added to the SAML community module.  Assuming that there aren't any unexpected transitive dependencies, commons-httpclient should only be included when the SAML module is installed.

  

# Checklist

### DIFF
--- a/src/community/release/ext-saml.xml
+++ b/src/community/release/ext-saml.xml
@@ -22,6 +22,7 @@
         <include>openws*.jar</include>
         <include>velocity*.jar</include>
         <include>esapi*.jar</include>
+        <include>commons-httpclient*.jar</include>
       </includes>
     </fileSet>
   </fileSets>
--- a/src/community/saml/pom.xml
+++ b/src/community/saml/pom.xml
@@ -86,6 +86,10 @@
       </exclusions>
     </dependency>
     <dependency>
+      <groupId>commons-httpclient</groupId>
+      <artifactId>commons-httpclient</artifactId>
+    </dependency>
+    <dependency>
       <groupId>com.github.tomakehurst</groupId>
       <artifactId>wiremock-standalone</artifactId>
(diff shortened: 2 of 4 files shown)


sikeoka:
@aaime and @jodygarnett could you take a look at this please?  It could replace the older PR that had compile failures https://github.com/geoserver/geoserver/pull/5807

aaime:
Looks good to me.

jodygarnett:
I wondered if we  commons-httpclient as a transitive dependency from gt-http-commons: https://github.com/geotools/geotools/blob/main/modules/plugin/http-commons/pom.xml#L64

(but cannot see that used in the codebase so yeah we are managing this on our own).

jodygarnett (approved):

sikeoka:
> I wondered if we commons-httpclient as a transitive dependency from gt-http-commons: https://github.com/geotools/geotools/blob/main/modules/plugin/http-commons/pom.xml#L64
> 
> (but cannot see that used in the codebase so yeah we are managing this on our own).

sikeoka:
I hit the wrong button.

sikeoka:
The pom.xml you linked to actually has the new httpclient library which is okay, not the old commons-httpclient library.

mprins (approved):
Checked that the `commons-httpclient:commons-httpclient:jar:3.1` is removed from the dependency tree (as a runtime dependency) using `commons-httpclient:commons-httpclient:jar:3.1` (it is still there as a test dep as expected) and also checked the war file

PR #6115, opened 2022-08-16
[GEOS-10547] Integrated WMS caching without the tiled parameter might…

turingtestfail:
[GEOS-10547](https://osgeo-org.atlassian.net/browse/GEOS-10547)

… result in deep recursion

cleanup

  

# Checklist

### DIFF
--- a/src/gwc/src/main/java/org/geoserver/gwc/wms/CachingWebMapService.java
+++ b/src/gwc/src/main/java/org/geoserver/gwc/wms/CachingWebMapService.java
@@ -13,6 +13,7 @@ import java.io.ByteArrayOutputStream;
 import java.lang.reflect.Method;
 import java.nio.channels.Channels;
 import java.util.LinkedHashMap;
+import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.servlet.http.HttpServletResponse;
@@ -20,6 +21,7 @@ import org.aopalliance.intercept.MethodInterceptor;
 import org.aopalliance.intercept.MethodInvocation;
 import org.geoserver.gwc.GWC;
 import org.geoserver.gwc.config.GWCConfig;
+import org.geoserver.gwc.layer.GeoServerTileLayer;
 import org.geoserver.ows.Dispatcher;
 import org.geoserver.ows.HttpErrorCodeException;
 import org.geoserver.wms.GetMapRequest;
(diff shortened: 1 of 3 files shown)


src/gwc/src/main/java/org/geoserver/gwc/wms/CachingWebMapService.java
```diff
@@ -64,7 +66,14 @@ public WebMap invoke(MethodInvocation invocation) throws Throwable {
...
         final GetMapRequest request = getRequest(invocation);
         boolean tiled = request.isTiled() || !config.isRequireTiledParameter();
-        if (!tiled) {
+        final Map<String, String> rawKvp = request.getRawKvp();
+        boolean isSeedingRequest =
+                rawKvp != null
+                        && rawKvp.containsKey(
+                                GeoServerTileLayer
+                                        .GWC_SEED_INTERCEPT_TOKEN); // skipping seeding requests to
+        // avoid meta tile recursion
```
aaime, line 75: Making peace with the formatter:

```suggestion
        // skipping seeding requests to avoid meta tile recursion
        boolean isSeedingRequest =
                rawKvp != null && rawKvp.containsKey(GeoServerTileLayer.GWC_SEED_INTERCEPT_TOKEN);
```

src/gwc/src/test/java/org/geoserver/gwc/GWCTest.java
```diff
@@ -1195,6 +1203,43 @@ private void assertDispatchMismatch(GetMapRequest request, String expectedReason
...
+     */
+    @Test
+    public void testGetMapRequest() throws Throwable {
+        GetMapRequest request = new GetMapRequest();
+
+        @SuppressWarnings("unchecked")
+        Map<String, String> rawKvp = new CaseInsensitiveMap(new HashMap<String, String>());
+        rawKvp.put(GeoServerTileLayer.GWC_SEED_INTERCEPT_TOKEN, "true");
+        request.setRawKvp(rawKvp);
+        request.setTiled(true);
```
aaime, line 1218: The bug report was actually about not having the tiled request parameter required, e.g.:

```
mediator.getConfig().setRequireTiledParameter(false);
```

aaime (changes requested):
Tested, works fine! See my two comments below.

src/gwc/src/test/java/org/geoserver/gwc/GWCTest.java
turingtestfail: Fixed

PR #6113, opened 2022-08-15
 [GEOS-10058] Welcome Page Layout

jodygarnett:
[GEOS-10058](https://osgeo-org.atlassian.net/browse/GEOS-10058)

See https://github.com/geoserver/geoserver/wiki/GSIP-202 proposal for details

# Checklist

- [x] I have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/main/CONTRIBUTING.md).
- [x] I have sent a [Contribution Licence Agreement](https://docs.geoserver.org/latest/en/developer/policies/committing.html) (not required for small changes, e.g., fixing typos in documentation).
- [x] First PR targets the `main` branch (backports managed later; ignore for branch specific issues).
- [ ] All the build checks are green ([see automated QA checks](https://docs.geoserver.org/latest/en/developer/qa-guide/index.html)).

For core and extension modules:

- [ ] New unit tests have been added covering the changes.
- [ ] [Documentation](https://github.com/geoserver/geoserver/tree/main/doc/en/user/source) has been updated (if change is visible to end users).
- [ ] The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers).
- [x] There is an issue in the [GeoServer Jira](https://osgeo-org.atlassian.net/browse/GEOS/summary) (except for changes that do not affect administrators or end users in any way).
- [ ] Commit message(s) must be in the form ``[GEOS-XYZWV] Title of the Jira ticket``.
- [ ] Bug fixes and small new features are presented as a single commit.
- [ ] Each commit has a single objective (if there are multiple commits, each has a separate JIRA ticket describing its goal).

PR #6122, opened 2022-08-19
[GEOS-10058] Welcome Page Layout

jodygarnett:
[GEOS-10058](https://osgeo-org.atlassian.net/browse/GEOS-10058)

See https://github.com/geoserver/geoserver/wiki/GSIP-202 proposal for details, prior discussion on https://github.com/geoserver/geoserver/pull/6113

# Checklist


jodygarnett:
Added select control for workspace and layers; but the catalog.getLayers() is not returning content unless logged in....

image

jodygarnett:
Okay code complete, styling fixed up, will pass QA checks tomorrow

src/main/src/main/java/org/geoserver/ows/LocalWorkspaceCallback.java
```diff
@@ -23,7 +23,7 @@
 
 /**
  * Dispatcher callback that sets and clears the {@link LocalWorkspace} and {@link LocalPublished}
- * thread locals.
+ * thread locals. "abc", "abc/xyz"
```
bradh, line 26: I don't get the intent of this change.

src/web/core/src/main/java/org/geoserver/web/CapabilitiesHomePagePanel.java
```diff
@@ -85,10 +88,20 @@ public CapabilitiesHomePagePanel(final String id, final List<CapsInfo> capsLinks
 
         super(id);
 
+        // Group CapInfo by service id
         final Map<String, List<CapsInfo>> byService = new HashMap<>();
         for (CapsInfo c : capsLinks) {
             final String key =
                     c.getService().toLowerCase(); // to avoid problems with uppercase definitions
+            if (key.equals("wms")
```
bradh, line 96: This feels a bit brittle. Could we ask the `ServicesPanel` what `key`s it handles  or if a `key` is handled instead?

src/main/src/main/java/org/geoserver/ows/LocalWorkspaceCallback.java
jodygarnett: it took me some time to understand the thread local context was of the form "workspace/layer", or "layergroup" or whatever ...

I will improve the javadoc while I wait for a build.

src/web/core/src/main/java/org/geoserver/web/CapabilitiesHomePagePanel.java
jodygarnett: I think I can remove these now that the ``CapabilitiesHomePageLinkProvider`` instances have been removed CapabilitiesHomePageLinkProvider.

This is only still present in the codebase for any community modules

jodygarnett: It is more what keys ``ServiceInfoCapabilitiesProvider`` handles; but perhaps I can ask that extension point.

jodygarnett: Okay I was able to process a good skip list.

aaime:
Looks indeed much better. Maybe there should be a bit more vertical space between one services section and the next one?

Wondering how the new page interacts with:

- OGC API modules (currently they seamlessly integrate in the list, how does it look after the change?)
- Logged in page, which often contains warnings, and tools to create new stores/layers on the spot

jodygarnett:
> * OGC API modules (currently they seamlessly integrate in the list, how does it look after the change?)

Here is with csw and ogcapi - they use the updated ``CapabilitiesHomePagePanel`` which has had ``geoserver.css`` styles adjusted to visually match. 

image

Because wicket components are being produced I could not process it into useful "other" panel; although now that I look at it I may talk to Michel to see if something could be done.

>  Logged in page, which often contains warnings, and tools to create new stores/layers on the spot

image

jodygarnett:
Just rebased to get a small fix in without making the commit history difficult.

mprins:
I've left some feedback in https://osgeo-org.atlassian.net/browse/GEOS-10634

aaime:
Toying with the home page, I've noticed the list of "extra services" does not reduce based on the currently selected layer:

image

"Service capabilities" could use a static explanation (something saying this are "other services" I suppose?) and its contents is not sensitive to the type of layer just selected.

The boxes for the service links still seem pretty bulky to me,

aaime:
About the new NE map, labels at this scale denominator would have plenty of space, I keep on wondering if labels are broken as I zoom in and out:

image

They show up at the next zoom level, where the generalization of country borders is too visible:

image

The net result is that it's difficult to see a "nice map".

aaime (changes requested):

src/web/core/src/main/java/org/geoserver/web/ServiceDescriptionProvider.java
```diff
@@ -0,0 +1,90 @@
...
+ * @author Jody Garnett
+ * @see ServicesPanel
+ */
+public abstract class ServiceDescriptionProvider {
+
+    /**
+     * Provides service descriptions, optionally filtered by workspace and layer.
+     *
+     * <p>Filtering is forgiving: provide the global services unless the workspace exactly matches;
+     * provide workspace services unless the layer exactly matches.
```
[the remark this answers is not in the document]
jodygarnett: My thinking is that the screen should not be empty, or 404...

Perhaps we could use the base page feedback to provide a warning if the layer is not found?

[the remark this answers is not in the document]
jodygarnett: Scratch that, the real info objects are now passed in (if found) and I will update this javadoc to match.

src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.java
```diff
@@ -21,67 +27,276 @@
...
+                        if (layerName == null && layerField.getModelObject() != null) {
+                            // No input selected, check model object displayed
+                            layerName = layerField.getModelObject().prefixedName();
+                        }
+                        String workspaceName = workspaceField.getInput();
+                        if (workspaceName == null && workspaceField.getModelObject() != null) {
+                            // No input selected, check model object displayed
+                            workspaceName = workspaceField.getModelObject().getName();
+                        }
+                        setResponsePage(
```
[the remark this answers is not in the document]
jodygarnett: I think it is an area that can be improved for sure.

src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.java
```diff
@@ -141,6 +361,19 @@ protected void populateItem(ListItem<GeoServerHomePageContentProvider> item) {
...
         add(contentView);
 
+        List<ServicesPanel.ServiceDescription> serviceDescriptions = new ArrayList<>();
+        List<ServicesPanel.ServiceLinkDescription> serviceLinks = new ArrayList<>();
+        for (ServiceDescriptionProvider provider :
+                getGeoServerApplication().getBeansOfType(ServiceDescriptionProvider.class)) {
+            serviceDescriptions.addAll(provider.getServices(workspaceInfo, layerInfo));
+            serviceLinks.addAll(provider.getServiceLinks(workspaceInfo, layerInfo));
+        }
+        if (!isAdmin(auth)) {
```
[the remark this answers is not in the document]
jodygarnett: Yeah I like that much more clean

aaime:
Checks have now legit compile failures:

```
Error: 1,400 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project gs-web-core: Compilation failure
Error: 1,400 [ERROR] /home/runner/work/geoserver/geoserver/src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.java:[295,55] illegal start of type
```

src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.java
```diff
@@ -22,69 +28,277 @@
...
+            Label footerMessage =
                     new Label(
                             "footerMessage",
                             new StringResourceModel(
-                                    "GeoServerHomePage.footer", this, new Model(params)));
-            label.setEscapeModelStrings(false);
-            add(label);
+                                    "GeoServerHomePage.footer",
+                                    this,
+                                    new Model<HashMap<>>(params)));
```
mprins, line 295:
```suggestion
                                    new Model<HashMap<>(params)));
```

data/release/wmts.xml
```diff
@@ -0,0 +1,22 @@
+<wmts>
+  <id>WMTSInfoImpl--3ef31051:182c1e8b194:-8000</id>
+  <enabled>true</enabled>
+  <name>WMTS</name>
+  <title>GeoServer Web Map Tile Service</title>
+  <maintainer>http://geoserver.org/com</maintainer>
+  <abstrct>A compliant implementation of WMTS service.</abstrct>
```
bradh, line 7: Possible typo? Or do we really do this?

```suggestion
  <abstract>A compliant implementation of WMTS service.</abstract>
```

aaime: We really do it :rofl: ... this data structure is a dump of a Java class properties, and "abstract" is a reserved keyword...

src/web/gwc/src/main/java/org/geoserver/gwc/web/WMTSServiceDescriptionProvider.java
```diff
@@ -0,0 +1,155 @@
...
+import org.geoserver.util.InternationalStringUtils;
+import org.geoserver.web.GeoServerApplication;
+import org.geoserver.web.ServiceDescriptionProvider;
+import org.geoserver.web.ServicesPanel;
+import org.geotools.util.Version;
+import org.opengis.util.InternationalString;
+import org.springframework.beans.factory.NoSuchBeanDefinitionException;
+
+/** Provide description of GeoWebCache services for welcome page. */
+public class WMTSServiceDescriptionProvider extends ServiceDescriptionProvider {
```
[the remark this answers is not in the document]
jodygarnett: Sure

jodygarnett (commented):
Feedback from earlier

src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.java
```diff
@@ -179,4 +455,113 @@ private boolean isAdmin(Authentication authentication) {
         return GeoServerExtensions.bean(GeoServerSecurityManager.class)
                 .checkAuthenticationForAdminRole(authentication);
     }
+
+    /**
```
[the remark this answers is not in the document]
jodygarnett: Would it be appropriate here:

* GeoServerSession.get().getAuthentication()
*  GeoServerSession.get().getUsername()

jodygarnett:
> About the new NE map, labels at this scale denominator would have plenty of space, I keep on wondering if labels are broken as I zoom in and out:
> 
> They show up at the next zoom level, where the generalization of country borders is too visible:
> 
> The net result is that it's difficult to see a "nice map".

I am sure more can be done during RC phase; for now I would like to focus on the functionality and we can circle back to the styling during RC.

jodygarnett:
# Initial Notes and Feedback

* Reviewed the the existing ``CapabilitiesHomePageLinkProvider`` which was setup to return a component; but in reality the component was always a ``CapabilitiesHomePagePanel`` which does not have enough visibility to see ServiceInfo objects.  Decided to setup a ``ServiceDescriptionProvider`` extension to return a data structure (ServiceDescriptino and ServiceLinkDescription)  which can be processed for display.  We can deprecate ``CapabilitiesHomePageLinkProvider`` when transition is complete.

* Experimenting with using page properties to include workspace and layer information in URL (``ne``, ``ne/roads``, ``?workspace=ne``, ``?workspace=ne&layer=roads``) so we can share a URL to a workspace's  services. The logic for how this is handled in dispatcher is spread out in a number of locations.

* @bradh provided api and javadoc feedback -- resolved

* @aaime asks that the look be more in keeping with geoserver [standard appearance](https://github.com/geoserver/geoserver/wiki/GeoServer-Branding); or even the [original proposal](https://github.com/geoserver/geoserver/wiki/GSIP-202#initial-design).
  
  Changed to use of standard headings, normal font size, and reduced the size of most everything.
  
* @MichelGabriel asked that the web services be a link, rather than just the text

* Used ``Select2DropDownChoice`` to list workspaces, layer groups and layers for selection. Had some trouble with SecurityCatalogImpl which was ignoring workspace and layer read permissions, and setting read to false when user was not an admin. While this was fixable it broke too many security tests, and I found another way to do it - GeoServerHomePage clears the  AdminRequest thread locale if it is operating as an anonymous page.

The result is much improved

image


Workspace specific:

image

jodygarnett:
# 2.22-M0 Notes and Feedback

Issues address in response to testing and feedback:
* Select2DropDownChoice updates check object, input and raw input to determine updated workspace name or layer name
* Refactor large init method into smaller methods each focused on a section of the page
* Admin overview summary now updates to reflect current selection
* Replace "Welcome / Welcome" (Title / Description) with a dynamic description that updates to describe what is being shown on the page of the form.
* global services were not disabling or hiding services

Issues identified:

* [x] GEOS-10664 WMTSGetCapabilities providerSite use of online resource
* [x] GEOS-10668 HTML page default for WFS 1.0.0 GetCapabilities OnlineResource (when WFSInfo.getOnlineResource() not available)

Non-critical issues:
* [ ] GEOS-10663  Recognize ContactInfo addressDeliveryPoint to address as synonyms and update WFS 1.1 Transformer
  
**Welcome Page with Global Services OFF**

image

**Welcome Page with Global Services ON**

image

**Welcome Page with workspace selected**

Workspace settings have customized online resource, organization name, and welcome blurb.

image

**Welcome Page with layer selected**

image

**Welcome Page with LayerGroup SINGLE selected**

image

**Welcome Page with LayerGroup Named Tree selected**

image

src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.java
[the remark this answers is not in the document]
jodygarnett: Andrea this is the last bit of feedback I have not resolved, going to go ahead and add ``GeoServerSession.isAdmin()`` as I think it is a good location that developers will discover when trying to figure out what is going on...

jodygarnett:
Final checks:
- I remember the last change request from @aaime which was to stack the workspace / layer virtual service selection.
- Addressed Andrea's request to factor out admin check, moved to ``GeoServerSession.isAdmin()`` 
- Removed the `data/release/workspaces/ne` into #6198 so any feedback on cartography would not hold up this functionality PR.

aaime:
Like the stacking of the dropdowns, wondering if a little label or an icon would help people to figure out what they are... maybe just the icons used in the menu, with a hover message telling what they are, e.g.:

Workspace selector

Anyways, nicer but not that important. What is still quite bad is the list of service capabilities for OGC APIs, all stacked in the same column even when using a full-screen window (1920x1080):


services

jodygarnett:
> Anyways, nicer but not that important. What is still quite bad is the list of service capabilities for OGC APIs, all stacked in the same column even when using a full-screen window (1920x1080):

Darn I thought I tested that to use the same presentation as the service descriptions above; I will fix that and merge later today.

jodygarnett:
@aaime I double checked that all the models were detached improving page load time.

I am not sure what has gone wrong with your display of Service Capabilities; is something caching geoserver.css ?

aaime:
The developer toolbar is a set of commands coming in handy when developing against the Wicket UI, it activates when the following system variable is provided to the Java runtime:

``-Dwicket.configuration=development``

It has been there since the first Wicket UI implementation, I searched for docs but it does not seem to be documented.

aaime:
I've tried again and guess what? The layout issue does not appear unless the wicket development mode is activated LOL (I have it always on).

aaime:
PMD ain't happy:

```
2022-09-27T05:35:17.3788360Z 05:35:17,378 [INFO] PMD Failure: org.geoserver.web.GeoServerHomePage:393 Rule:UnusedLocalVariable Priority:3 Avoid unused local variables such as 'layersModel'..
```

jodygarnett:
> I've tried again and guess what? The layout issue does not appear unless the wicket development mode is activated LOL (I have it always on).

"Does not seem to be documented" <-- lol!

So the development tools must be modifying the DOM structure a bit throwing off the css styling?

aaime:
> So the development tools must be modifying the DOM structure a bit throwing off the css styling?

I have no clue, most of what they do is to change configs in Wicket itself.

jodygarnett:
> PMD ain't happy:
> 
> ```
> 2022-09-27T05:35:17.3788360Z 05:35:17,378 [INFO] PMD Failure: org.geoserver.web.GeoServerHomePage:393 Rule:UnusedLocalVariable Priority:3 Avoid unused local variables such as 'layersModel'..
> ```

Unused local variable my fav PMD warning; okay it is addressed.

jodygarnett:
> > So the development tools must be modifying the DOM structure a bit throwing off the css styling?
> 
> I have no clue, most of what they do is to change configs in Wicket itself.

Looks like it inserted a ``<wicketpanel>``, going to adjust my css rules to use ``.`` rather than ``#``:
```
#services .link-bordered, .capabilities .link-bordered {
...
}
```

jodygarnett:
From today's meeting:

* [*] Include serviceCapabilities even if global services are off so ogg-api services (for example) are listed

MichelGabriel:
> Update: Checking in browser inspect page <wicket:panel> is treated as a <div> and is throwing off the flex-display.

As far as I can see, the `<wicket:panel>` is the first `<div>` after the the `<div class="flex-display">`, which means that's the one who will be the child (the flex item), and all the other blocks will be displayed normally (in this case in one small column).

A lot of layouting with css depends on the order of elements in the DOM

PR #6124, opened 2022-08-23
[GEOS-10649] Concurrent modification to GWC style parameter filter can lead to OOM

jasminb:
[GEOS-10649](https://osgeo-org.atlassian.net/browse/GEOS-10649)

This PR fixes the issue in `StyleParameterFilter#setLayer()` method which is being called concurrently and can cause out of memory problems due to the corruption of `TreeHashMap` that can occur in scenarios when map is being modified by more than one thread.

I did not go into analysis of why the instances of this class are being concurrently updated in the first place, however the changes made in this PR are at least fixing the OOM that its current implementation can cause.

# Checklist


aaime:
As you can see from the checklist, under normal circumstances we only merge a PR if it has a test. Can you set up one? Ideally, it would be a integration type, showing how it happens, but a more unit one would also be good. If you cannot provide a test, it might be that another developer could complete the task, or if we are not that lucky, the stale bot will close the PR in a few months.

aaime:
Changes look good. Created a JIra ticket for it, pushed a change adding missing copyright header in the tests. Will squash in a single commit on merge.

PR #6129, opened 2022-08-25
[GEOS-10631] AccessManager will not be looked up if multiple beans are of type DefaultResourceAccessManager

taba90:
[GEOS-10631](https://osgeo-org.atlassian.net/browse/GEOS-10631)

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/SecureCatalogImpl.java
+++ b/src/main/src/main/java/org/geoserver/security/SecureCatalogImpl.java
@@ -104,16 +104,20 @@ public class SecureCatalogImpl extends AbstractDecorator<Catalog> implements Cat
         List<ResourceAccessManager> managers =
                 GeoServerExtensions.extensions(ResourceAccessManager.class);
         ResourceAccessManager manager = null;
-        if (managers.size() == 1) {
+        int size = managers.size();
+        if (size == 1) {
             manager = managers.get(0);
         } else {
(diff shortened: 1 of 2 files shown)


src/main/src/main/java/org/geoserver/security/SecureCatalogImpl.java
```diff
@@ -104,16 +104,20 @@ static ResourceAccessManager lookupResourceAccessManager() throws Exception {
...
-                if (!(resourceAccessManager instanceof DefaultResourceAccessManager)) {
+                if (!DefaultResourceAccessManager.class.equals(resourceAccessManager.getClass())) {
                     manager = resourceAccessManager;
                     break;
                 }
             }
         }
+        // should never happen,just in case we have multiple singleton beans
+        // of type DefaultResourceAccessManager
+        if (manager == null) manager = managers.get(size - 1);
```
aaime, line 120: Maybe better to use managers.get(0), if multiple managers are registered the first one should be the one with the highest priority (if the ExtensionPriority interface has been used)

aaime (changes requested):
Looks good, just a note about the last resort lookup below

src/main/src/main/java/org/geoserver/security/SecureCatalogImpl.java
taba90: fixed

taba90:
@aaime  thanks for the review, feedback applied

PR #6130, opened 2022-08-25
[GEOS-10632] Make sure GetLegendGraphics honors the WMS memory limits

ahmedababnehgeo:
[GEOS-10632](https://osgeo-org.atlassian.net/browse/GEOS-10632)

The legend graphic building happening BufferedImageLegendGraphicBuilder is complicated, and generally inconsistent... however there is a common trait, legends are build of of list of images, one per layer, and each layer image is built out of a list of smaller images. These collections of images are accumulated, to be then merged

# Checklist

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/legendgraphic/TalliedQueue.java
+++ b/src/wms/src/main/java/org/geoserver/wms/legendgraphic/TalliedQueue.java
@@ -1,0 +1,58 @@
+/* (c) 2014 - 2022 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2022 OpenPlans
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.wms.legendgraphic;
+
+import java.util.LinkedList;
+import java.util.Queue;
+
+public class TalliedQueue<T> implements ImageSizeComputable {
+
+    private final Queue<T> queue = new LinkedList<>();
+
+    private final Tally tally;
+
+    public TalliedQueue(Tally tally) {
+        this.tally = tally;
+    }
+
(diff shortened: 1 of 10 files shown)


src/wms/src/main/java/org/geoserver/wms/legendgraphic/TalliedQueue.java
```diff
@@ -0,0 +1,58 @@
...
+ * (c) 2001 - 2022 OpenPlans
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.wms.legendgraphic;
+
+import java.util.LinkedList;
+import java.util.Queue;
+
+public class TalliedQueue<T> implements ImageSizeComputable {
```
jodygarnett, line 11: This seems quite the heavy weight solution; how does a normal GetMap request respect time constraint?

aaime: GetMap has a simpler computation, width*height*number of feature type styles* bands.
Here due to the various layouting options and symbology sizes, it's difficult to predict how big the legend will become.

I don't see where the "heavy weight" is, the code is already using collections to keep track of the various sub-images that it will have to lay out later, these new classes do the same keeping track of how much memory has been used in the process. Please elaborate on how you'd do the same in a simpler way.

jodygarnett: Okay cool; I just wanted to make sure that introducing a new class ``TalliedQueue`` was not duplicating work already done for GetMap.

aaime:
Superceded by https://github.com/geoserver/geoserver/pull/6446

PR #6151, opened 2022-08-31
[GEOS-10640] OGC API - Features - Output to JSON fails with class cast exception

sebastianfrey:
[GEOS-10640](https://osgeo-org.atlassian.net/browse/GEOS-10640)

  

# Checklist

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java
@@ -1064,7 +1064,7 @@ public class GetFeature {
                 .collect(
                         Collectors.toMap(
                                 e -> e.getKey(),
-                                e -> (String) e.getValue(),
+                                e -> String.valueOf(e.getValue()),
                                 (u, v) -> {
                                     throw new IllegalStateException(
                                             String.format("Duplicate key %s", u));


aaime:
Won't be considered for review without a test.

src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java
```diff
@@ -1064,7 +1064,7 @@ private KvpMap<String, String> mapValuesToStrings(Map<String, Object> rawKvp) {
                 .collect(
                         Collectors.toMap(
                                 e -> e.getKey(),
-                                e -> (String) e.getValue(),
+                                e -> String.valueOf(e.getValue()),
```
aaime, line 1067: A raw KVP normally contains only strings, what's happening in OGC API that makes it contain something else? Also, what is that, a String[] maybe?

sebastianfrey: In fact it is a list of Strings. Why that happens, I actually don't know. Maybe it is related to #6076, where sorting was introduced to OGC API?

aaime: It may be related indeed, but should still not happen, the raw kvp should contain the original string found in the query string (.e.g, "att1,-att2,att3").

sebastianfrey: I think we can close this PR. I don't know why it happened, and how it happened. But after I did setup a fresh GeoServer instance the ClassCastException disappeared. Sorry for robbing your time on this one.

sebastianfrey:
I think we can close this PR. I don't know why it happened, and how it happened. But after I did setup a fresh GeoServer instance the ClassCastException disappeared. Sorry for robbing your time on this one.

aaime:
No worries, keep an eye for when it might reoccur (one never knows).

PR #6187, opened 2022-09-21
[GEOS-10665] - Parametrize Users' passwords

taba90:
[GEOS-10655](https://osgeo-org.atlassian.net/browse/GEOS-10655)

Allow to use env parametrization for users' passwords.
See related jira ticket for details.  

# Checklist

### DIFF
--- a/doc/en/user/source/security/passwd.rst
+++ b/doc/en/user/source/security/passwd.rst
@@ -120,3 +120,6 @@ Each user/group service uses a password policy to enforce these rules. The defau
 * Password minimum length
 * Password maximum length
 
+Parametrized Passwords
+-----------------------
+It is possible to parametrize users' passwords in a similar way to the catalog settings (see :ref:`datadir_configtemplate`). Parametrization is supported when the encryption method used to store the place holder in the password field is plain text or is reversible (pbe, strong pbe). Non reverisble encoding for the placeholder (eg. digest) is not supported. On the contrary the actual value can be defined in the ``geoserver-environment.properties`` with any password encoding method supported by GeoServer. 
--- a/doc/en/user/source/security/passwd.rst
+++ b/doc/en/user/source/security/passwd.rst
@@ -122,4 +122,10 @@ Each user/group service uses a password policy to enforce these rules. The defau
 
 Parametrized Passwords
 -----------------------
-It is possible to parametrize users' passwords in a similar way to the catalog settings (see :ref:`datadir_configtemplate`). Parametrization is supported when the encryption method used to store the place holder in the password field is plain text or is reversible (pbe, strong pbe). Non reverisble encoding for the placeholder (eg. digest) is not supported. On the contrary the actual value can be defined in the ``geoserver-environment.properties`` with any password encoding method supported by GeoServer. 
+It is possible to parametrize users' passwords in a similar way to the catalog settings (see :ref:`datadir_configtemplate`). Parametrization is supported when the encryption method used to store the place holder in the password field is plain text or is reversible (pbe, strong pbe). Non reversible encoding for the placeholder (e.g. digest) is not supported. On the contrary the actual value can be defined in the ``geoserver-environment.properties`` with any password encoding method supported by GeoServer. Examples are provided below:
+
+.. code-block:: properties
+
+ pwd.one=plain:clear_text_password
+ pwd.two=digest1:D9miJH/hVgfxZJscMafEtbtliG0ROxhLfsznyWfG38X2pda2JOSV4POi55PQI4tw
+ pwd.three=crypt1:xZJscMafEtbtliG0ROxhLfsznyWfG38X2pda2JOSV4POi55PQI4tw
--- a/src/main/src/main/java/org/geoserver/security/auth/EnvPwdResolverUserGroupServiceWrapper.java
+++ b/src/main/src/main/java/org/geoserver/security/auth/EnvPwdResolverUserGroupServiceWrapper.java
@@ -1,0 +1,95 @@
+/* (c) 2022 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.security.auth;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import org.apache.commons.lang.StringUtils;
+import org.geoserver.platform.GeoServerEnvironment;
+import org.geoserver.platform.GeoServerExtensions;
+import org.geoserver.security.GeoServerSecurityManager;
+import org.geoserver.security.GeoServerUserGroupService;
+import org.geoserver.security.impl.GeoServerUser;
+import org.geoserver.security.password.GeoServerPasswordEncoder;
+import org.geotools.util.logging.Logging;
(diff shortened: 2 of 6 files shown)


bradh:
Failed format check in QA. Needs local build.

doc/en/user/source/security/passwd.rst
```diff
@@ -120,3 +120,6 @@ Each user/group service uses a password policy to enforce these rules. The defau
 * Password minimum length
 * Password maximum length
 
+Parametrized Passwords
+-----------------------
+It is possible to parametrize users' passwords in a similar way to the catalog settings (see :ref:`datadir_configtemplate`). Parametrization is supported when the encryption method used to store the place holder in the password field is plain text or is reversible (pbe, strong pbe). Non reverisble encoding for the placeholder (eg. digest) is not supported. On the contrary the actual value can be defined in the ``geoserver-environment.properties`` with any password encoding method supported by GeoServer.
```
bradh, line 125:
```suggestion
It is possible to parametrize users' passwords in a similar way to the catalog settings (see :ref:`datadir_configtemplate`). Parametrization is supported when the encryption method used to store the place holder in the password field is plain text or is reversible (pbe, strong pbe). Non reversible encoding for the placeholder (e.g. digest) is not supported. On the contrary the actual value can be defined in the ``geoserver-environment.properties`` with any password encoding method supported by GeoServer. 
```

bradh (commented):
I would have guess "parameterize" and "parameterized" was a bit more common, but as long as it is consistent with other docs, seems OK.

src/main/src/main/java/org/geoserver/security/auth/EnvPwdResolverUserGroupServiceWrapper.java
```diff
@@ -0,0 +1,95 @@
...
+import org.geotools.util.logging.Logging;
+import org.springframework.security.core.userdetails.UserDetails;
+import org.springframework.security.core.userdetails.UserDetailsService;
+import org.springframework.security.core.userdetails.UsernameNotFoundException;
+
+/**
+ * A UserGroupService wrapper able to resolve password that have been parametrized. The placeholder
+ * can be encoded in plain txt or any reversible encoding.
+ */
+class EnvPwdResolverUserGroupServiceWrapper implements UserDetailsService {
```
aaime, line 27:
```suggestion
class EnvironemetUserDetailService implements UserDetailsService {
```

src/main/src/main/java/org/geoserver/security/auth/EnvPwdResolverUserGroupServiceWrapper.java
```diff
@@ -0,0 +1,95 @@
...
+
+    EnvPwdResolverUserGroupServiceWrapper(GeoServerUserGroupService userDetailsService) {
+        this.delegate = userDetailsService;
+        environment = GeoServerExtensions.bean(GeoServerEnvironment.class);
+    }
+
+    @Override
+    public UserDetails loadUserByUsername(String username) throws UsernameNotFoundException {
+        UserDetails userDetails = delegate.loadUserByUsername(username);
+        if (environment != null && GeoServerEnvironment.allowEnvParametrization()) {
```
aaime, line 42: Since this can be activated only at startup time, I'd suggest to move the check into ``UsernamePasswordAuthenticationProvider`` and just skip the wrapping if env parametrization is not enabled or not available.

src/main/src/main/java/org/geoserver/security/auth/EnvPwdResolverUserGroupServiceWrapper.java
```diff
@@ -0,0 +1,95 @@
...
+                String decodedPwd = decode(password, encoders);
+                return returnOrGetResolvedPwdUser(decodedPwd, user);
+            }
+        }
+        return userDetails;
+    }
+
+    // copy the user and set the pwd value if it was parametrized in env properties.
+    // otherwise returns the user passed as a parameter.
+    private GeoServerUser returnOrGetResolvedPwdUser(String decodedPwd, GeoServerUser user) {
```
aaime, line 58:
```suggestion
    private GeoServerUser resolvePassword(String decodedPwd, GeoServerUser user) {
```

src/main/src/main/java/org/geoserver/security/auth/EnvPwdResolverUserGroupServiceWrapper.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @Override
+    public UserDetails loadUserByUsername(String username) throws UsernameNotFoundException {
+        UserDetails userDetails = delegate.loadUserByUsername(username);
+        if (environment != null && GeoServerEnvironment.allowEnvParametrization()) {
+            if (userDetails instanceof GeoServerUser) {
+                GeoServerUser user = (GeoServerUser) userDetails;
+                String password = user.getPassword();
+                GeoServerSecurityManager manager = delegate.getSecurityManager();
+                List<GeoServerPasswordEncoder> encoders =
```
aaime, line 47: I'm not familiar with how password encoding works, but it seems you're getting a UserDetails with an encoded password, you decode it and replace it with a clear text version. But the upper levels should still suppose the password is encoded, or not? Also, I've seen in the past that plain passwords are normally prefixed by "plain:"....

aaime (changes requested):

doc/en/user/source/security/passwd.rst
taba90: fixed

src/main/src/main/java/org/geoserver/security/auth/EnvPwdResolverUserGroupServiceWrapper.java
taba90: fixed

src/main/src/main/java/org/geoserver/security/auth/EnvPwdResolverUserGroupServiceWrapper.java
taba90: fixed

src/main/src/main/java/org/geoserver/security/auth/EnvPwdResolverUserGroupServiceWrapper.java
taba90: fixed

src/main/src/main/java/org/geoserver/security/auth/EnvPwdResolverUserGroupServiceWrapper.java
taba90: The encoded password (including plain) is first decoded to retrieve the eventually present placeholder. Then the placeholder is resolved on env. The env value provided might be encoded in wathever encryption supported by GeoServer since the dao authentication provider will try all the password encoders to match the incoming password with the resolved one. The only component that will have visibility over the resolved value is the DaoAuthenticationProvider wrapped into the UserNamePasswordAtuehtnicationProvider, that will actually do the password check.
I've also added an example of passwords that might be stored in the properties on documentation.

taba90:
thanks for your reviews @bradh and @aaime. Feedbacks applied.

aaime (approved):

aaime:
Looks good, thanks!

PR #6195, opened 2022-09-23
[GEOS-10671] Parallel REST API calls failures (users)

aaime:
[GEOS-10671](https://osgeo-org.atlassian.net/browse/GEOS-10671)

See ticket

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/concurrent/LockingRoleService.java
+++ b/src/main/src/main/java/org/geoserver/security/concurrent/LockingRoleService.java
@@ -45,8 +45,14 @@ public class LockingRoleService extends AbstractLockingService
 
     @Override
     public GeoServerRoleStore createStore() throws IOException {
-        GeoServerRoleStore store = getService().createStore();
-        return store != null ? new LockingRoleStore(store) : null;
+        // the store will duplicate the role service internal data structures, so read lock
+        readLock();
+        try {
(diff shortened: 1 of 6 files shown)


src/main/src/main/java/org/geoserver/GeoServerConfigurationLock.java
```diff
@@ -38,7 +38,7 @@ public class GeoServerConfigurationLock {
      */
     public static long DEFAULT_TRY_LOCK_TIMEOUT_MS = 30000;
 
-    private static final Level LEVEL = Level.FINE;
+    private static final Level LEVEL = Level.INFO;
```
jodygarnett, line 41: Do we really want configuration lock at "info" level? Assume you had this on for troubleshooting; looks like LEVEL just controls one log message; so setting this as FINE and using log4j.xml logging of locking should be good enough.

jodygarnett (approved):

taba90 (commented):
Looks good to me.

src/main/src/main/java/org/geoserver/GeoServerConfigurationLock.java
aaime: Yep, it was for troubleshooting and forgot to back it out. NIce catch, fixing it.

PR #6205, opened 2022-09-28
[GEOS-10677] Improve cleanup of multi part form upload to the dispatcher

sikeoka:
[GEOS-10677](https://osgeo-org.atlassian.net/browse/GEOS-10677)

This PR updates the OWS dispatcher to close streams and remove temp files for multipart/form-data requests without relying on the garbage collector.  Unnecessary input streams are no longer opened for requests that contain multiple file uploads and an NPE is fixed if the dispatcher couldn't find a field to read the input from.  This PR also fixes a race condition where the GC could try to delete a file while the stream is still open which would fail on Windows or could delete the file before GeoServer finished parsing it on Linux. The commons-fileupload upgrade fixes some edge cases where invalid input could leave streams open so it is relevant to this PR.
  

# Checklist

### DIFF
--- a/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
+++ b/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
@@ -327,37 +327,49 @@ public class Dispatcher extends AbstractController {
             } else if (reqContentType != null
                     && ServletFileUpload.isMultipartContent(httpRequest)) {
                 // multipart form upload
-                ServletFileUpload up = new ServletFileUpload();
-                up.setFileItemFactory(new DiskFileItemFactory());
+                ServletFileUpload up = new ServletFileUpload(new DiskFileItemFactory());
 
                 // treat regular form fields as additional kvp parameters
                 Map<String, FileItem> kvpFileItems =
                         new CaseInsensitiveMap<>(new LinkedHashMap<>());
                 try {
-                    for (FileItem item : up.parseRequest(httpRequest)) {
+                    List<FileItem> items = up.parseRequest(httpRequest);
+                    FileItemCleanupCallback.setFileItems(items);
+                    FileItem body = null;
+                    for (FileItem item : items) {
                         if (item.isFormField()) {
-                            kvpFileItems.put(item.getFieldName(), item);
+                            FileItem old = kvpFileItems.put(item.getFieldName(), item);
+                            if (old != null) {
+                                old.delete(); // the temp file can be deleted at this point
+                            }
                         } else {
-                            request.setInput(fileItemReader(item));
+                            if (body != null) {
+                                // GeoServer only reads the last uploaded file per request so any
+                                // other file uploads can be deleted immediately
+                                body.delete();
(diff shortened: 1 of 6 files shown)


src/ows/src/main/java/org/geoserver/ows/FileItemCleanupCallback.java
```diff
@@ -0,0 +1,45 @@
...
+    public static void setFileItems(List<FileItem> fileItems) {
+        FILE_ITEMS.set(fileItems);
+    }
+
+    @Override
+    public void finished(Request request) {
+        List<FileItem> items = FILE_ITEMS.get();
+        FILE_ITEMS.remove();
+        if (!items.isEmpty()) {
+            try (Reader r = request.getInput()) {
```
aaime, line 36: Hmm... this input could also be coming from the HttpServletRequest.getInputStream(). I vaguely remember Tomcat getting real mad at us (like, hard segfault when using APR) for messing with his streams. 
In general, we should not be closing streams that we did not create ourselves, no?

src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
```diff
@@ -327,37 +327,49 @@ Request init(Request request) throws ServiceException, IOException {
...
+                            FileItem old = kvpFileItems.put(item.getFieldName(), item);
+                            if (old != null) {
+                                old.delete(); // the temp file can be deleted at this point
+                            }
                         } else {
-                            request.setInput(fileItemReader(item));
+                            if (body != null) {
+                                // GeoServer only reads the last uploaded file per request so any
+                                // other file uploads can be deleted immediately
+                                body.delete();
```
aaime, line 349: The callback is already going to delete all items at the end of the request, so this extra logic improves deletion by a little bit and only for corner cases. Is it worth the extra complication? Makes me wonder if you're trying to address some sort of attack.

aaime (commented):
Looks good in general, see below two bits that made me wonder...

src/ows/src/main/java/org/geoserver/ows/FileItemCleanupCallback.java
sikeoka: The reader is only closed for multipart/form-data requests since no other requests will have a non-empty list of file items and the input for those form requests are always from FileItem.getInputStream(), not the servlet input stream.

src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
sikeoka: Deleting temp files that could contain arbitrary contents uploaded by anyone with access to the GeoServer instance seems like a good idea but it's your call on whether or not the extra logic is worth having.  Even with this extra logic, someone with local access to the GeoServer machine could still use GeoServer to upload arbitrary files onto the machine by creating a custom client that sends a request to upload the file and then throttles the request on a subsequent field until the user has time to copy the file out of the temp directory.

PR #6210, opened 2022-10-03
[GEOS-10680] FidTransformeVisitor misspelled

dieterstueken:
[GEOS-10692](https://osgeo-org.atlassian.net/browse/GEOS-10692)

[GEOS-10680](https://osgeo-org.atlassian.net/browse/GEOS-10692)

While the file reads as FidTransformerVisitor.java, the class itself  is misspelled as FidTransformeVisitor.

### DIFF
--- a/src/main/src/main/java/org/geoserver/feature/retype/FidTransformerVisitor.java
+++ b/src/main/src/main/java/org/geoserver/feature/retype/FidTransformerVisitor.java
@@ -20,10 +20,10 @@ import org.opengis.filter.identity.FeatureId;
  *
  * @author Andrea Aime
  */
-class FidTransformeVisitor extends DuplicatingFilterVisitor {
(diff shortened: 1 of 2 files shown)


mprins (approved):
LGTM, not sure if a CLA is required, it is clearly a typo

aaime:
Nah, changes are minimal, no need for a CLA, and given it's just a typo, no need for a test either. I'm merging this one.

bradh:
Should it be backported, given its an API change? Or is this OK because the class isn't `public`?

aaime:
The second, because it's not public. In GeoServer we have more relaxed approach at API stability. Gut feeling:
* Had it been public but not widely used, we should have stopped the change on main.
* Had it been an extension point of some widely used class, I would have requested a "extend, deprecate and replace" approach, renaming the class, but keeping an extension of it with the old name, deprecated, to avoid breaking users

PR #6213, opened 2022-10-04
[GEOS-10678]: NPE if <settings> element is missing

dieterstueken:
[GEOS-10678](https://osgeo-org.atlassian.net/browse/GEOS-10678)

[GEOS-10678](https://osgeo-org.atlassian.net/browse/GEOS-10678)

When using `data/minimal/global.xml` no `<settings>` element is defined.
This causes a NPE on `XStreamPersister.doUnmarshal()`.
`XStreamPersister.doMarshal()` looks similar, so I also added a precaution null check here.

If the initialization fails, spring destroys its singleton beans again. 
This causes an other NPE on`GeoServerLoader.destroy()` since it was not fully initialized. 
So I added an other NP check here.

# Checklist

- [x]  I have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/main/CONTRIBUTING.md).

### DIFF
--- a/src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
+++ b/src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
@@ -2565,7 +2565,7 @@ public class XStreamPersister {
                 Object source, HierarchicalStreamWriter writer, MarshallingContext context) {
             GeoServerInfo geoServerInfo = (GeoServerInfo) source;
             SettingsInfo settings = geoServerInfo.getSettings();
-            if (settings.isUseHeadersProxyURL() == null) {
+            if (settings != null && settings.isUseHeadersProxyURL() == null) {
                 settings.setUseHeadersProxyURL(geoServerInfo.isUseHeadersProxyURL());
                 if (geoServerInfo instanceof GeoServerInfoImpl)
                     ((GeoServerInfoImpl) geoServerInfo).setUseHeadersProxyURLRaw(null);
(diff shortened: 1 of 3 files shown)


dieterstueken:
Unit test added, force pushed in front of the fix.
So you can verify the NPE without the fix.

Both other null tests might be obsolete now, but are still preventive (no unit tests for them now).

aaime:
@davidblasby you reported the ticket, could you review the PR?

davidblasby:
ok, looks like all the homepage issues are unrelated to this PR.

src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
```diff
@@ -2565,7 +2565,7 @@ protected void doMarshal(
                 Object source, HierarchicalStreamWriter writer, MarshallingContext context) {
             GeoServerInfo geoServerInfo = (GeoServerInfo) source;
             SettingsInfo settings = geoServerInfo.getSettings();
-            if (settings.isUseHeadersProxyURL() == null) {
+            if (settings != null && settings.isUseHeadersProxyURL() == null) {
```
jodygarnett, line 2568: While we can add a null check; the real challenge is that during the load process settings should of been created.

I suspect that this is during the migration of an "old" data directory where only global.xml is defined....

jodygarnett (changes requested):

jodygarnett:
Tested with empty folder and minimal; looks good approving and merging.

jodygarnett:
> With minimal I am still failing to start to to NPE:

Stepping through with debugger GeoServerInfoConverter --> AbstractReflectionConverter.doUnmarshal I noted that GeoServerInfoImpl.toStirng() would also fail due to NPE calling getTitle().

jodygarnett (approved):
It worked in testing today: release, minimal and empty directory

jodygarnett:
Thanks for the fix

PR #6215, opened 2022-10-04
[GEOS-10681] WPS: handle info of currently running processes

etj:
[GEOS-10681](https://osgeo-org.atlassian.net/browse/GEOS-10681)

> In order to be able to handle data authorization during chained WPS calls (see geoserver/geofence#225 ) we need to be able to find out which process is trying to access a resource. Since processes can be chained, the auth mechanism also needs to know which are the parent callers of the current executing process.

The implementation adds a new Listener that is notified whenever a WPS process is invoked or returns (either succeeding or failing).

# Checklist


aaime:
Legit QA failure, multiple files are not formatted according to the coding convention. Just build with Maven, and it will format the files for you (in the one and only accepted way). I'd recommend also disabling any form of auto-formatting in the IDE, if you have one.

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ChainedProcessListener.java
```diff
@@ -0,0 +1,21 @@
...
+import org.geoserver.wps.executor.ProcessManager;
+
+/** An interface to monitor how processes are chained.
+ * 
+ *  @author etj (Emanuele Tajariol @ GeoSolutions)
+ */
+public interface ChainedProcessListener {
+
+    /** Called right before the process is submitted into the {@link ProcessManager} */
+    void submitted(String executionId, String processName, boolean chained) throws WPSException;
```
aaime, line 16: This is a new extension point, please add javadoc to all methods.

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ChainedProcessListenerNotifier.java
```diff
@@ -0,0 +1,50 @@
+/* (c) 2022 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.wps;
+
+import java.util.List;
+import org.geoserver.wps.executor.ProcessManager;
+
+/** A notifier to fire events when processes are (un)chained.
```
aaime, line 10: What is process (un)chaining?

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ChainedProcessListenerNotifier.java
```diff
@@ -0,0 +1,50 @@
...
+ * 
+ *  @author etj (Emanuele Tajariol @ GeoSolutions)
+ */
+public class ChainedProcessListenerNotifier {
```
aaime, line 14: Please avoid public visibility for classes unless they need to be visible in other packages to work. As a support class, it could be placed in the executor package and be done with it, is there anything else that needs to use it?

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ChainedProcessListener.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+package org.geoserver.wps;
+
+import org.geoserver.wps.executor.ProcessManager;
+
+/** An interface to monitor how processes are chained.
+ * 
+ *  @author etj (Emanuele Tajariol @ GeoSolutions)
+ */
+public interface ChainedProcessListener {
```
aaime, line 13: Given the name, I was expecting to see notifications only for the chained processes, but it seems it's notifying also about the progress of the main one... we already have notifications for that one, couldn't a class implement both?
Also, since the interface is generic, why not notify about dismissal too?

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/executor/WPSExecutionManager.java
```diff
@@ -450,12 +478,15 @@ private ExecuteResponseType execute() {
                             inputPercentage + executionPercentage,
                             "Execution completed, preparing to write response");
                 }
+                chainNotifier.fireReturned();
             } catch (ProcessDismissedException e) {
                 // that's fine, move on writing the output
+                chainNotifier.fireFailed();
```
aaime, line 484: This is not a failure though, it's the user asking to dismiss an otherwise properly running process.

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/executor/WPSExecutionManager.java
```diff
@@ -450,12 +478,15 @@ private ExecuteResponseType execute() {
...
             } catch (Exception e) {
                 if (status.getPhase() != ProcessState.DISMISSING) {
                     LOGGER.log(Level.SEVERE, "Process execution failed", e);
                     notifier.fireFailed(e);
+                    chainNotifier.fireFailed();
```
aaime, line 489: From a generic usage standpoint, getting to know the exception might be useful.

src/extension/wps/wps-core/src/test/java/org/geoserver/wps/ChainedProcessListenerTest.java
```diff
@@ -0,0 +1,281 @@
...
+    @Before
+    public void oneTimeSetUp() throws Exception {
+        WPSInfo wps = getGeoServer().getService(WPSInfo.class);
+
+        // want at least two asynchronous processes to test concurrency
+        wps.setMaxAsynchronousProcesses(Math.max(2, wps.getMaxAsynchronousProcesses()));
+//        wps.setMaxSynchronousExecutionTime(Integer.MAX_VALUE); // avoid timeout while debugging
+//        wps.setMaxAsynchronousExecutionTime(Integer.MAX_VALUE); // avoid timeout while debugging
+//        wps.setMaxSynchronousTotalTime(Integer.MAX_VALUE); // avoid timeout while debugging
+//        wps.setMaxAsynchronousTotalTime(Integer.MAX_VALUE); // avoid timeout while debugging
```
aaime, line 48: Please avoid commented out code.

aaime (changes requested):

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ChainedProcessListener.java
etj: Javadoc updated to reflect for the fact that the listener is notified for both the root process and internal chained processes.
Modifying the existing listener would break its API contract.
This listener is a bit more "strict" about the subprocesses: it fires a return/fail when the process returns, while the existing listener lingers a bit on the output encoding.

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/executor/WPSExecutionManager.java
etj: It's not going to run further chained processes tho, which is what the listener is about. Maybe different method names are needed here.

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/ChainedProcessListener.java
aaime: I see the updated javadoc on the ChainedProcessListenerNotifier (which is not public, good thanks) but not here in the public interface, which just states:

```An interface to monitor how processes are chained.```

**Comment by aaime** [comment]
There is one legit PMD failure:

```
2022-10-05T17:04:46.3927786Z 17:04:46,392 [INFO] PMD Failure: org.geoserver.wps.ChainedProcessListenerTest:7 Rule:wildcards Priority:3 No Wildcard Imports.
```

PR #6216, opened 2022-10-04
[GEOS-10694] Suppress stacktrace on startup when logging.xml is missing.

dieterstueken:
[GEOS-10685](https://osgeo-org.atlassian.net/browse/GEOS-10685)

[GEOS-10694](https://osgeo-org.atlassian.net/browse/GEOS-10694)
[GEOS-10693](https://osgeo-org.atlassian.net/browse/GEOS-10693)

Starting on an empty context without any logging.xml generates an error. 
Even worse: an empty logging.xml file is generated instead. This raises future future errors.
(The possibly deeper problem here is, that `Resource.in()` creates an empty new file.)

Expected behavior: silently ignore the empty logging.xml and don't create an empty one.

Committed a unit test to demonstrate the error an a fix to resolve it.

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/logging/LoggingStartupContextListener.java
+++ b/src/main/src/main/java/org/geoserver/logging/LoggingStartupContextListener.java
@@ -112,7 +112,7 @@ public class LoggingStartupContextListener implements ServletContextListener {
         // the "change" between logging.xml and the versions stored in JDBC. KS
         // TODO find a better solution than re-initializing on JDBCCOnfig startup.
         Resource f = store.get("logging.xml");
-        if (f != null) {
+        if (f != null && f.getType() == Resource.Type.RESOURCE) {
             XStreamPersister xp = new XStreamPersisterFactory().createXMLPersister();
             try (BufferedInputStream in = new BufferedInputStream(f.in())) {
                 LoggingInfo loginfo = xp.load(in, LoggingInfo.class);
--- a/src/main/src/test/java/org/geoserver/logging/LoggingStartupContextListenerTest.java
+++ b/src/main/src/test/java/org/geoserver/logging/LoggingStartupContextListenerTest.java
@@ -142,10 +142,17 @@ public class LoggingStartupContextListenerTest {
         context.setInitParameter(
                 "GEOSERVER_LOG_LOCATION", new File(tmp, "foo.log").getAbsolutePath());
 
-        ServletContextListener listener = new LoggingStartupContextListener();
+        try (TestAppender appender = TestAppender.createAppender("quite", null)) {
+            appender.startRecording("org.geoserver.logging");
 
-        listener.contextInitialized(new ServletContextEvent(context));
-        listener.contextDestroyed(new ServletContextEvent(context));
+            appender.trigger("Could not reconfigure LOG4J loggers");
+
+            ServletContextListener listener = new LoggingStartupContextListener();
+            listener.contextInitialized(new ServletContextEvent(context));
+            listener.contextDestroyed(new ServletContextEvent(context));
+
+            appender.stopRecording("org.geoserver.logging");
+        }
     }
 
     @Test
--- a/src/main/src/test/java/org/geoserver/logging/LoggingStartupContextListenerTest.java
+++ b/src/main/src/test/java/org/geoserver/logging/LoggingStartupContextListenerTest.java
@@ -13,6 +13,7 @@ import static org.junit.Assert.assertTrue;
 
 import java.io.File;
 import javax.servlet.ServletContextEvent;
+import javax.servlet.ServletContextListener;
 import org.apache.commons.io.FileUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.core.Appender;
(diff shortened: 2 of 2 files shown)


aaime:
I don't think this fix matches the description of https://osgeo-org.atlassian.net/browse/GEOS-10685
The ticket must describe the issue you found, if there is no ticket matching it, you should create a new one.

@jodygarnett this area of code is something you worked on recently I believe, could you have a look at the changes?

dieterstueken:
This tries to solve the first topic of [GEOS-10685](https://osgeo-org.atlassian.net/browse/GEOS-10685):

> 1. Usual message when trying to configure without logging.xml

I may split it off into a separate ticket, but that does not contain any additional information.

aaime:
@dieterstueken we want tickets specifically describing the change, not something related. When this change is committed, the ticket is closed and gets into the release notes. With a vague relation, we either close the ticket, and the release notes are incorrect about the scope of changes, or we don't, and the change is not reported to end users.

src/main/src/main/java/org/geoserver/logging/LoggingStartupContextListener.java
```diff
@@ -112,7 +112,7 @@ public void contextInitialized(ServletContextEvent event) {
         // the "change" between logging.xml and the versions stored in JDBC. KS
         // TODO find a better solution than re-initializing on JDBCCOnfig startup.
         Resource f = store.get("logging.xml");
-        if (f != null) {
+        if (f != null && f.getType() == Resource.Type.RESOURCE) {
```
jodygarnett, line 115: Nice minimal change! Although I am not sure `f` can ever be ``null``.

jodygarnett (approved):
Nice minimal change; suitable for backport

jodygarnett:
> @dieterstueken we want tickets specifically describing the change, not something related. When this change is committed, the ticket is closed and gets into the release notes. With a vague relation, we either close the ticket, and the release notes are incorrect about the scope of changes, or we don't, and the change is not reported to end users.

Andrea in this case I reported the ticket, we can make "sub tasks": 

* https://osgeo-org.atlassian.net/browse/GEOS-10694 Creating a logging.xml file when starting up form an empty directory is expected and should not log a stacktrace

src/main/src/test/java/org/geoserver/logging/LoggingStartupContextListenerTest.java
```diff
@@ -127,6 +128,26 @@ public void testLogLocationFromServletContext() throws Exception {
...
+        assertTrue(logs.mkdirs());
+
+        MockServletContext context = new MockServletContext();
+        context.setInitParameter("GEOSERVER_DATA_DIR", tmp.getPath());
+        context.setInitParameter(
+                "GEOSERVER_LOG_LOCATION", new File(tmp, "foo.log").getAbsolutePath());
+
+        ServletContextListener listener = new LoggingStartupContextListener();
+
+        listener.contextInitialized(new ServletContextEvent(context));
```
jodygarnett, line 147: The test case does not seem to listen to the logs to see if an logging event is generated. I made an test appender for writing these kind of test cases; but this is a bit of a special case since there is not yet an existing logging setup to add a test appender to! 

So this test case is good and we should keep it; but it does not check that logging messages for the log file being missing are indeed suppressed.

aaime: @dieterstueken can you add the missing tests?

dieterstueken: will have a look this evening ...

dieterstueken: My initial intention was just to avoid the NPE.
(sorry for the different issues flipping back and forth, 
something with the badge processing i did not get right)

Without the fix the NPE is raised and after the fix it vanished. So, the test works fine for me.

That the logging generation must be suppressed if the config file is missing, is an additional requirement i was not aware of. But I may simply add a test if the log directory really remains empty after all.

I also observed that the test does not finally remove the directories created (but this is true for all other tests, too).

aaime: Not removing the directories created is an intentional decision, allows the developer to inspect the data dir contents used by the tests. Very important when pulling hairs debugging a complex situation. The directories are placed in ``/target``, which is meant to contain removable information anyways (so ``mvn clean`` will get rid of them).

jodygarnett: Here is the code snippet - however TestAppender was not collecting events when I tried:

```
try (TestAppender appender = TestAppender.createAppender("quite", null)) {
            appender.startRecording("org.geoserver.logging");
            
            appender.trigger("Could not reconfigure LOG4J loggers");
            
            LoggingStartupContextListener listener = new LoggingStartupContextListener();
            listener.contextInitialized(new ServletContextEvent(context));
            listener.contextDestroyed(new ServletContextEvent(context));

            appender.stopRecording("org.geoserver.logging");
        }
```

The code snippet is correct though, and when TestAppender functionality is fixed it will be correct.

jodygarnett:
When this is done building it can be squashed and merged

PR #6217, opened 2022-10-04
Just build please: address too many open files and use of osgeo-releases for third-party jars

jodygarnett:
This PR covers fixes/improvements when changing to Java 11 as my daily driver:

* [x] Too many open files encountered when doing a complete build with -Prelease.communityRelease
  Despite increasing ulimit -n value a full build will encounter "too many open files". 
  
* [x] Use of osgeo-releasees for third-party dependencies (having a single repo to check speeds up build)

These are non-functional changes to the build system only and may be back ported.

# Checklist

- [ x I have sent a [Contribution Licence Agreement](https://docs.geoserver.org/latest/en/developer/policies/committing.html) (not required for small changes, e.g., fixing typos in documentation).

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -2202,9 +2202,6 @@
           <releases>
             <enabled>true</enabled>
           </releases>
-          <snapshots>
-            <enabled>true</enabled>
-          </snapshots>
           <id>geosolutions</id>
           <name>geosolutions repository</name>
           <url>https://maven.geo-solutions.it/</url>
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -162,6 +162,7 @@
 
   <dependencyManagement>
     <dependencies>
+      <!-- GeoServer -->
       <dependency>
         <groupId>org.geoserver</groupId>
         <artifactId>gs-platform</artifactId>
@@ -262,6 +263,8 @@
         <artifactId>gs-kml</artifactId>
         <version>${gs.version}</version>
       </dependency>
+
+      <!-- GeoTools -->
       <dependency>
         <groupId>org.geotools</groupId>
         <artifactId>gt-geotools</artifactId>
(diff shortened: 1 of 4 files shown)


aaime:
Interesting, did they fix something specific or was it just a random guess?

jodygarnett:
> Interesting, did they fix something specific or was it just a random guess?

Given that changing to errorProne compiler worked I figured something was up; tried the latest.

src/pom.xml
```diff
@@ -1506,45 +1532,45 @@
...
       <name>OSGeo Nexus Snapshot Repository</name>
       <url>https://repo.osgeo.org/repository/snapshot/</url>
+      <!-- Access to OSGeo snapshots repositories                                          -->
     </repository>
 
-    <repository>
-      <releases>
-        <enabled>true</enabled>
-      </releases>
-      <snapshots>
```
aaime, line 1541: You should get most of the speedup you're looking for by just disabling snapshot support. Operationally, it would be better for us, as only a few of the front-facing developers have access to the OSGeo infrastructure for deploy (I believe only me and Daniele) while others do not even have commit access (let alone the rest), but occasionally we have to deploy a jar for quick security fixes or to get a fix on an otherwise dead dependency (e.g., jgridshift).

jodygarnett: You can give more people access to upload, they just need their osgeo id. 

Still the existing caching rules should be sufficient? And I left a profile available (in case there is a community module with dependencies I do not know about).

src/pom.xml
```diff
@@ -2139,8 +2184,36 @@
...
+      <!--   ^/sf/json-lib/.*                                          -->
+      <id>geosolutions</id>
+      <repositories>
+        <repository>
+          <releases>
+            <enabled>true</enabled>
+          </releases>
+          <snapshots>
+            <enabled>true</enabled>
+          </snapshots>
```
jodygarnett, line 2207: @aaime we can remove snapshots from here?

jodygarnett: Change made

mprins (approved):

PR #6220, opened 2022-10-05
[GEOS-10647]fixes for issues with welcome/home page

davidblasby:
[GEOS-10647](https://osgeo-org.atlassian.net/browse/GEOS-10647)

This fixes some issues on the Geoserver Home (welcome) page.

1. There were some minor issues with the urls - see https://osgeo-org.atlassian.net/browse/GEOS-10647
2. Sometime the same service would appear twice
3. For GWC-based services, it was sometimes thinking the service wasn't enabled when it was (isAvalable)
4. Urls for GWC-based getcapabilities were incorrect

# Checklist

- [x] I have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/main/CONTRIBUTING.md).
- [x] I have sent a [Contribution Licence Agreement](https://docs.geoserver.org/latest/en/developer/policies/committing.html) (not required for small changes, e.g., fixing typos in documentation).
- [x] First PR targets the `main` branch (backports managed later; ignore for branch specific issues).
- [ ] All the build checks are green ([see automated QA checks](https://docs.geoserver.org/latest/en/developer/qa-guide/index.html)).

For core and extension modules:

- [ ] New unit tests have been added covering the changes.
- [n/a] [Documentation](https://github.com/geoserver/geoserver/tree/main/doc/en/user/source) has been updated (if change is visible to end users).
- [n/a] The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers).
- [x] There is an issue in the [GeoServer Jira](https://osgeo-org.atlassian.net/browse/GEOS/summary) (except for changes that do not affect administrators or end users in any way).
- [on squash] Commit message(s) must be in the form ``[GEOS-XYZWV] Title of the Jira ticket``.
- [x] Bug fixes and small new features are presented as a single commit.
- [x] Each commit has a single objective (if there are multiple commits, each has a separate JIRA ticket describing its goal).

PR #6221, opened 2022-10-06
Slim GeoJSON WFS output format

cklein05:
This community module adds the Slim GeoJSON WFS output format. That format is a more compact and memory-friendly variant of traditional GeoJSON, which saves space by removing redundant schema information for simple feature results. For complex features there are no differences compared to traditional GeoJSON output.

See the [README.md](https://github.com/cklein05/geoserver/tree/slimgeojson/src/community/slimgeojson) for a more detailed description of the format.

This community module and format was announced on the developer's mailing list as _Compact JSON Format_. I decided to rename it to _Slim GeoJSON_ as _Slim_ is shorter and easier to pronounce...

This PR contains the community module code as well as some changes in the already existing JSON provider, which this new format extends.

Still missing are the required entries in `.../src/community/pom.xml` as well as an XML file in directory
 `.../src/community/release` (if required). Also missing is the module's help content in directory (I guess) `.../doc/en/user/source/community/slimgeojson`. Maybe the linked README.md above is a good starting point for a module description. Are you going to add all these?

# Checklist

I guess that all the build checks are green. However, for a first time Maven user this was not so easy to get sorted. AFAIK, community modules are build only if profile `communityRelease` is specified on the command line. Also, the module needs to be added to file `.../src/community/pom.xml` (which I did not within this PR). So, after a successful build of the core, extensions and community modules, I've build the new module "by hand" from its own directory.

BTW, (how) do you invoke Maven from a standard Windows command line? Or are you all on Linux/Mac? It seems that there are problems with Windows paths containing backslashes. maven-compiler-plugin was complaining about an _invalid escape sequence_ in `C:\Users\...` near index 3. Also, some tests did not work correctly (I guess maven was unable to find and read the generated output files due to path name problems). So, I had to start Maven with the -f option in order to specify the path to the root pom.xml with slashes. Still tests did not work properly so I had to skipTests as well.

### DIFF
--- a/src/community/slimgeojson/README.md
+++ b/src/community/slimgeojson/README.md
@@ -1,0 +1,109 @@
+# Slim GeoJSON WFS output format extension
+
+This module adds the Slim GeoJSON WFS output format. That format is a more compact and memory-friendly variant of traditional GeoJSON, which saves space by removing redundant schema information for simple feature results. For complex features there are no differences compared to traditional GeoJSON output.
+
+This module adds these additional WFS output formats for requesting simple features in Slim GeoJSON format:
+
+- `application/json; subtype=geojson/slim`  
+   for Slim GeoJSON
+- `text/javascript; subtype=geojson/slim`  
+   for Slim GeoJSON in JSONP requests
+
+In traditional GeoJSON, every feature in a (simple feature) feature collection has its own schema information. That is, every feature contains all its (not necessarily short) attribute names. Except the geometry name, these are used as the keys in the `"properties"` map:
+
+```json
+{
+  "type": "FeatureCollection",
+  "features": [
+    {
+      "type": "Feature",
+      "id": "areas.1",
+      "geometry": {
+        "type": "Point",
+        "coordinates": [590529, 4914625]
+      },
+      "geometry_name": "the_geom",
+      "properties": {
+        "area_no": 12,
+        "area_name": "Mainland",
+        "area_description": "grassland",
+        "area_cost_center": "0815"
+      }
+    },
+    {
+      "type": "Feature",
+      "id": "areas.2",
+      "geometry": {
+        "type": "Point",
+        "coordinates": [590215, 4913987]
+      },
+      "geometry_name": "the_geom",
+      "properties": {
+        "area_no": 17,
+        "area_name": "South region",
+        "area_description" : "meadow, pasture",
+        "area_cost_center": "0812"
+      }
+    }
+  ],
+  "totalFeatures": 2,
+  "numberMatched": 2,
+  "numberReturned": 2,
+  "timeStamp": "2022-10-03T08:44:45.118Z",
+  "crs": {
+    "type": "name",
+    "properties": {
+      "name": "urn:ogc:def:crs:EPSG::26713"
+    }
+  }
+}
+```
+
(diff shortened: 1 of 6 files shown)


cklein05:
OK, after waiting for the tests seems like not all of them are green :(

No idea why gs-security-tests fail on Mac but work on other systems. Is that a Mac setup problem?

Both QA failures seem to complain about a too long line of code. One comes from my code `src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java`. The other too long line has been added by #6220.

Is it correct, that I should build with Maven (using spotless for formatting) prior to committing? Could you please give some more advice?

Still not sure whether the "real" new code of the community module actually gets tested by these tests while not added to `.../src/community/pom.xml`? Currently I have no real idea what exactly to add (and where). Some more advice is appreciated.

jratike80:
Have you compared this format with https://www.brokjson.dev/specification?

cklein05:
> Have you compared this format with https://www.brokjson.dev/specification?

No, I didn't know the BrokJSON format. Seems like they use the same technique of using arrays for both field names and values. Of course, that's a quite obvious idea when one realizes, that storing field names for every feature is just a waste of space.

However, they've added a couple of new functions, which I do not yet understand in detail (should leave the office now). I will have a closer look tomorrow.

aaime:
Many questions... here are a few answers:
* It's up to you to modify ``community/pom.xml`` as well as the release descriptors, referencing them from the community pom, and testing that the zip generated from the build works fine with a nightly.
* While a bit of documentation is not mandatory for community modules, a page describing the format would certainly help people that want to try out the format (https://www.brokjson.dev/specification is a good example of how to push for a format, with a very visible page about it). Anyways, you're the master of the module, as long as it's a community module, it's entirely up to you.
* The [developer guide](https://docs.geoserver.org/latest/en/developer/) is a outdated in some parts, but it should answer at least some of your questions regarding building.
* Most of the developers are on *nix systems, yes, but we keep a Windows build among the GitHub build actions to make sure the code is buildable on Windows too. Not sure about the failure in Maven compile, but indeed the few times I'm forced on windows I create a "c:\devel" directory as the root of my projects. 
* Some tests do randomly fail on the Github actions, we have not found a way to reproduce locally. We just restart the builds (like I did for you right now).
* Regarding building with maven to get the right formatting, yes, it's correct. Read the developer guide, it has a section talking about the QA checks.
* Code in community modules is built without running its tests, as an incentive to push the modules up to extension status, where tests are actually run.

cklein05:
Thanks for the answers.

Moving the format name discussion to the thread with Jukka on the mailing list.

I will manage the community/pom.xml et al as well as the *.rst files under help/...

We use *nix systems on the server side only so far. Maybe it's time for a change :-)

Thanks for restarting the Mac OS CI build. Know this behavior from other projects...

Agree with the rest of your answers.

aaime:
Formatting should have been fixed, you can rebase this branch (preferred, but you can also merge main, up to you, this PR is due to for a "squash merge" anyways, to get one commit and a reference to one jira ticket... btw, please create one so that the format shows up in release notes, which are generated from the list of jira tickets closed targetting a specific release).

cklein05:
Thanks for that. However, I'd like to clear up the format and naming issues prior to merge things into GeoServer. There's a discussion on the mailing list with Jukka (you are getting these mails as well).

cklein05:
Closed. Will be replace with a new PR, soon.

PR #6228, opened 2022-10-07
typo

mirandadam:
Although the full meaning of the parameter is not clear to me yet, it seems clear that the intended word is "through". It did "throw" me off at first when trying to understand it. ;-)

  

# Checklist

- [NA] I have sent a [Contribution Licence Agreement](https://docs.geoserver.org/latest/en/developer/policies/committing.html) (not required for small changes, e.g., fixing typos in documentation).

### DIFF
--- a/doc/en/user/source/extensions/printing/configuration.rst
+++ b/doc/en/user/source/extensions/printing/configuration.rst
@@ -33,7 +33,7 @@ Here is the general structure:
     - {HOST_WHITELIST_DEFINITION}
     {...}
 
-  ?localHostForward: # For request on map.example.com we build an http request on localhost with the header Host=map.example.com, this is to don't pass throw the proxy.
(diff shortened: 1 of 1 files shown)


doc/en/user/source/extensions/printing/configuration.rst
```diff
@@ -33,7 +33,7 @@ Here is the general structure:
     - {HOST_WHITELIST_DEFINITION}
     {...}
 
-  ?localHostForward: # For request on map.example.com we build an http request on localhost with the header Host=map.example.com, this is to don't pass throw the proxy.
+  ?localHostForward: # For request on map.example.com we build an http request on localhost with the header Host=map.example.com, this is to don't pass through the proxy.
```
bradh, line 36:
```suggestion
  ?localHostForward: # For request on map.example.com we build an http request on localhost with the header Host=map.example.com, this is to not pass the request through the proxy.
```

bradh (commented):
The change is valid, although it could still be worded better. I've added a suggestion.

PR #6231, opened 2022-10-10
[GEOT-10702] Use constant to crosslink between service description and links

jodygarnett:
# Checklist

### DIFF
--- a/src/web/gwc/src/main/java/org/geoserver/gwc/web/GWCServiceDescriptionProvider.java
+++ b/src/web/gwc/src/main/java/org/geoserver/gwc/web/GWCServiceDescriptionProvider.java
@@ -24,6 +24,9 @@ import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 /** Provide description of GeoWebCache services for welcome page. */
 public class GWCServiceDescriptionProvider extends ServiceDescriptionProvider {
 
+    /** Service used to cross-link between service description and service link description. */
+    public static final String SERVICE = "gwc";
+
     private final GWC gwc;
     GeoServer geoserver;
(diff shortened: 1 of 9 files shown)


src/web/gwc/src/main/java/org/geoserver/gwc/web/GWCServiceDescriptionProvider.java
```diff
@@ -151,7 +154,7 @@ public List<ServiceLinkDescription> getServiceLinks(
             if (gwcConfig.isWMSCEnabled() && null != app.getBean("gwcServiceWMS")) {
                 links.add(
                         new ServiceLinkDescription(
-                                "wmts",
+                                SERVICE,
```
aaime, line 157: The other changes are making sense, this one seems surprising... why go from a fixed "wmts" to a fixed "gwc?

jodygarnett: This is only to establish the headings; since you had me rename WMTSServiceDescriptorProvider to GWCServiceDescriptorProvider I was going with the flow.

Really the variable names have shifted over time to accommodate GWC speaking widely different protocols and planning ahead for ogcapi services. The variable protocol in the case of gwc provides "WMS-C", "WMTS", "TMS" etc...

jodygarnett: Updated the PR description to explain the relationships better

aaime:
Ok, makes sense now, you updated both GWC description and link descriptions, so they match. Merging.

PR #6242, opened 2022-10-14
[GEOS-10700] Avoid automaticly updating logging profiles from built-in template

aaime:
[GEOS-10700](https://osgeo-org.atlassian.net/browse/GEOS-10700)

See ticket. 
Replaces https://github.com/geoserver/geoserver/pull/6234
Needs squash-merge to go back being a single commit.

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/logging/LoggingUtilsDelegate.java
+++ b/src/main/src/main/java/org/geoserver/logging/LoggingUtilsDelegate.java
@@ -5,7 +5,6 @@
 package org.geoserver.logging;
 
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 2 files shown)


aaime:
@jodygarnett I cannot get the test to work... the intention is to change the verbose logging so that it would log fine but not finer.
The test works fine if run in isolation, but run all the tests in the class, and it fails... as if there is something lingering.
I've tried issuing a "LogManager.shutdown" but it did not help. Any clue?

aaime:
Never mind, found a way to better isolate the test (the test class is inadvertently mixing up log4j and java loggers).

jodygarnett (changes requested):
Andrea I like your test case; I completed my draft PR here https://github.com/geoserver/geoserver/pull/6254 to include docs, and built-in logging profile update procedures etc...

Since we expect some changes to built-in logging procedures it will be nice to link to a documentation page from the release announcement.

I am going to take your test-case over to that branch.

Would you like to close this PR?

PR #6244, opened 2022-10-14
[GEOS-10701] Logging profiles timestamp reports minutes where it should report months

aaime:
[GEOS-10701](https://osgeo-org.atlassian.net/browse/GEOS-10701)

See ticket.

# Checklist

### DIFF
--- a/src/main/src/test/java/org/geoserver/logging/LoggingStartupContextListenerTest.java
+++ b/src/main/src/test/java/org/geoserver/logging/LoggingStartupContextListenerTest.java
@@ -5,6 +5,8 @@
  */
 package org.geoserver.logging;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.regex.Pattern.DOTALL;
 import static org.geoserver.logging.GeoServerXMLConfiguration.attributeGet;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
@@ -12,10 +14,13 @@ import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
 import java.io.File;
+import java.util.Locale;
+import java.util.regex.Pattern;
 import javax.servlet.ServletContextEvent;
 import javax.servlet.ServletContextListener;
 import org.apache.commons.io.FileUtils;
 import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.core.Appender;
 import org.apache.logging.log4j.core.LoggerContext;
 import org.apache.logging.log4j.core.appender.RollingFileAppender;
@@ -24,6 +29,8 @@ import org.apache.logging.log4j.core.config.Node;
 import org.geoserver.platform.GeoServerResourceLoader;
 import org.geoserver.platform.resource.FileSystemResourceStore;
 import org.geoserver.platform.resource.MemoryLockProvider;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
 import org.junit.Test;
 import org.springframework.mock.web.MockServletContext;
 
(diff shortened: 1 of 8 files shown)


aaime:
@jodygarnett this one also seems to suffer from poor isolation between tests... works fine if I run it alone, fails in two different ways if I run all the tests in IntelliJ, and in Maven.

In IntelliJ:

```
java.lang.AssertionError: Expected geoserverlogfile to be null.  But was: geoserverlogfile expected null, but was:<geoserverlogfile>

	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotNull(Assert.java:756)
	at org.junit.Assert.assertNull(Assert.java:738)
	at org.geoserver.logging.LoggingStartupContextListenerTest.testLogLocationFromServletContext(LoggingStartupContextListenerTest.java:89)
```

In maven:

```
[ERROR] Errors: 
[ERROR]   LoggingStartupContextListenerTest.testDateFormat:259->testExpectetFormat:273 » FileNotFound
[INFO] 
[ERROR] Tests run: 787, Failures: 0, Errors: 1, Skipped: 5
[INFO] 
```

jodygarnett:
> Is there an equivalent of `CRS.reset("all")` for the logging setup?

Yes, you can call:

```
LoggingUtils.initLogging(loader, "TEST_LOGGING.xml", false, true, null);
```

Internally these methods are doing the following to reset:
```
        @SuppressWarnings({
            "resource",
            "PMD.CloseResource"
        }) // current context, no need to enforce AutoClosable
        LoggerContext loggerContext = (LoggerContext) LogManager.getContext(false);
        ...
        ConfigurationSource source = ConfigurationSource.fromUri(configLocation);
        ...
        loggerContext.reconfigure(configuration);
```

To troubleshoot configuration stuff run with ``-Dorg.apache.logging.log4j.simplelog.StatusLogger.level=DEBUG`` as documented in LoggingUtils javadocs.

jodygarnett:
Andrea I agree this is being hard to test in isolation - testing just that class works:
```
mvn test -Dorg.apache.logging.log4j.simplelog.StatusLogger.level=DEBUG -Dtest=org.geoserver.logging.LoggingStartupContextListenerTest
```

And testing just that method:
```
mvn test -Dorg.apache.logging.log4j.simplelog.StatusLogger.level=DEBUG -Dtest=org.geoserver.logging.LoggingStartupContextListenerTest#testDateFormat
```

aaime:
Ok, found the state interference: RELINQUISH_LOG4J_CONTROL system variable and static variable value. Fixed, should build now.

aaime:
All green, good to go @jodygarnett ?

jodygarnett:
Andrea I went down a deep rabbit hole trying to figure this out - the apache docs on [testing with maven](https://logging.apache.org/log4j/log4j-2.3/manual/configuration.html#UnitTestingInMaven) are wrong; it looks like they use a newer junit now that does some kind of injection.

```
@LoggerContextSource("log4j-deadlock.xml")
@Tag("concurrency")
public class DeadlockTest {

    @Test
    public void deadlockOnReconfigure(final LoggerContext context) {
        context.reconfigure();
    }
}
```

What I got to work was making a method for testing each logging profile; I think this is an acceptable result and we can add to these over time if there is specific functionality we wish to test for an individual logging profile.

src/main/src/test/java/org/geoserver/logging/LoggingStartupContextListenerTest.java
```diff
@@ -36,6 +43,22 @@
...
 
+    static Locale systemLocale;
+
+    @BeforeClass
+    public static void initialize() {
+        systemLocale = Locale.getDefault();
+        Locale.setDefault(Locale.ENGLISH);
+        // reset static state leftover from other tests
+        System.clearProperty(LoggingUtils.RELINQUISH_LOG4J_CONTROL);
+        LoggingUtils.relinquishLog4jControl = false;
```
jodygarnett, line 54: Nicely done!

So this either means something is setting that concurrently (and it may fail). Or something failed to restore the value after completing it's test....

jodygarnett (approved):

PR #6245, opened 2022-10-14
Update Ubuntu 18.04 workflows to 20.04

aaime:
Backport bot is failing with this message:

 ```
This is a scheduled Ubuntu-18.04 brownout. The Ubuntu-18.04 environment is deprecated and will be removed on December 1st, 2022. For more details, see https://github.com/actions/runner-images/issues/6002
 ```

So trying to update the builds still using 18.04 to 20.04

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -14,7 +14,7 @@ jobs:
       contents: write
       pull-requests: write
       issues: write
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     name: Backport
     steps:
       - name: Backport Bot
--- a/.github/workflows/linux_jdk8.yml
+++ b/.github/workflows/linux_jdk8.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         include:
-          - os: ubuntu-18.04
+          - os: ubuntu-20.04
             jdk: 8
             dist: 'temurin'
     steps:


aaime:
Hmm... the backport bot fails regardless but I cannot get any sort of log, even when explicitly requesting to have debug logs... the build just gets cancelled?

mprins (commented):
I don't think the backport workflow should be backported, it should only run in the main branch.

probably it should be removed from  the 2.22.x branch

aaime:
I was thinking to have it, because just occasionally we have some fix that needs to happen only on older versions.
E.g., say we had to do a fix for an older Log4J 1.x configuration, which does not make sense for main anymore, or something similar for any sort of updated dependency with breaking changes.

aaime:
Also inquired here: https://github.com/m-kuhn/backport/issues/24

mprins:
that makes sense.

backport might fail for backport for the old branches (not 2.22.x) anyway because the file does not exist so it can't create a patch.

but it's odd that t just cancels; any chance it's running the backport workflow from main (on 18.04 that is browned out atm)

aaime:
I thought the same, but I can ask the action what workflow file it used, and:

image

mprins:
hmmm that's odd

in any case the brownout schedule is here: https://github.com/actions/runner-images/issues/6002 so another 3 hrs

mprins (approved):
Kicked it again and now it's green

PR #6257, opened 2022-10-17
New community module: SpatialJSON WFS Output Format

cklein05:
This module adds the SpatialJSON WFS output format. The SpatialJSON format is a more compact and memory-friendly variant of GeoServer's GeoJSON format. It aims to save space by applying several optimizations to traditional GeoJSON format for  simple feature results. Most of these optimizations work by removing redundand information from the JSON-encoded features.

A service exception is thrown if the result contains complex features as the SpatialJSON format does not handle those.

See the [README.md](https://github.com/cklein05/geoserver/tree/spatialjson/src/community/spatialjson) for a more detailed description of the format.

# Checklist

### DIFF
--- a/doc/en/user/source/community/index.rst
+++ b/doc/en/user/source/community/index.rst
@@ -61,6 +61,7 @@ officially part of the GeoServer releases. They are however built along with the
    saml/index
    schemaless-features/index
    smart-data-loader/index
+   spatialjson/index
    stac-datastore/index
    solr/index
    taskmanager/index
--- a/doc/en/user/source/community/spatialjson/development.rst
+++ b/doc/en/user/source/community/spatialjson/development.rst
@@ -1,0 +1,26 @@
+.. _spatialjson_development:
+
+Development Status
+==================
+
+The SpatialJSON format is still a playground for implementing several optimizations to transfer
+even huge amounts of spatial data from the server to the client efficiently:
+
+#. **Opt. 1: Removing redundant schema information**, see :doc:`topic <schema>`
+#. Opt. 2: Removing redundant attribute values (e. g. shared string table)
+#. Opt. 3: Handling sparse rows (most values are NULL) more efficiently
+#. Opt. 4: Reducing space required for geometries (e. g. differential coordinates)
+
+Bold items have already been implemented.
+
+The shown optimizations are ordered from *simple to implement* to *hard to implement* (not *really*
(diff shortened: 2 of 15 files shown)


aaime:
Looks good to me, letting the builds go to verify the changes. 
If you want a backport, you can issue backport PRs against 2.22.x or earlier (we have a backport bot, but won't work in this case, one of the pom files has the 2.23 version number in it, so cannot be blindly cherry-picked on 2.22.x)

cklein05:
> but won't work in this case, one of the pom files has the 2.23 version number in it, so cannot be blindly cherry-picked on 2.22.x

But I can manually issue a PR against 2.22 w/o modifying that pom file?

aaime:
You can manually issue a PR against 2.22, and that one pom file will have to say 2.22-SNAPSHOT instead of 2.23-SNAPSHOT.

aaime:
The PR seems ready for merge but I read on the devel list that you still want to do some changes. If you want they can come in a subsequent PR? In GeoServer we have been applying [RERO](https://en.wikipedia.org/wiki/Release_early,_release_often) for quite a while, it does not need to be fully complete at the first merge, especially for a community module.

cklein05:
You could merge right now, if you like. However, changes will be ready 
even today. So, I let this up to you :) If merge now, I will issue a 
subsequent PR.

aaime:
Yes, I can merge.
The backports are up to you, not required. Mind that every change must land on main first, and then it can be backported.
As a result, many modules in heavy development are often maintainted only on the main branch (but, up to you, if all the backporting does not bother you, there is no problem).

Unfortunately, we cannot give you write access to just a single module, but only to the entire codebase. Also, we are moving more and more towards an almost exclusive usage of PRs as a way to contribute code, leaving direct commits to emergencies. 

This means you have a lesser need for direct commit access, although we can still give it to you, under the understanding that you'll use it only to modify the community module, and will use it to merge your own PRs on that community module only.

cklein05:
OK, merge it. Still know to develop in main first... I'd like to backport to at least 2.22, since I want to use the module before 2.23 release in spring 2023...

I was not thinking about direct commits but the right to merge my own PRs without bothering anyone of you... Don't want/need to have write access to the entire codebase right now. So, I will ask you to merge if required (with a Jira issue and PR) and try not to backport too much... :)

aaime:
Done

PR #6260, opened 2022-10-17
[GEOS-10710] Features Templating backward mapping with back xpath ('../my/property/name') doesn't work

taba90:
[GEOS-10710](https://osgeo-org.atlassian.net/browse/GEOS-10710)

  

# Checklist

### DIFF
--- a/src/community/features-templating/features-templating-core/src/main/java/org/geoserver/featurestemplating/request/TemplatePathVisitor.java
+++ b/src/community/features-templating/features-templating-core/src/main/java/org/geoserver/featurestemplating/request/TemplatePathVisitor.java
@@ -7,11 +7,8 @@ package org.geoserver.featurestemplating.request;
 import com.fasterxml.jackson.databind.JsonNode;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
-import java.util.LinkedList;
 import java.util.List;
-import java.util.stream.Collectors;
-import org.apache.commons.lang.StringUtils;
+import java.util.Stack;
 import org.geoserver.featurestemplating.builders.AbstractTemplateBuilder;
 import org.geoserver.featurestemplating.builders.SourceBuilder;
 import org.geoserver.featurestemplating.builders.TemplateBuilder;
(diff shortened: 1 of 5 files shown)


src/community/features-templating/features-templating-core/src/main/java/org/geoserver/featurestemplating/request/TemplatePathVisitor.java
```diff
@@ -286,6 +292,37 @@ private boolean canCompleteXpath(Object extradata) {
                     || (extradata instanceof Boolean && ((Boolean) extradata).booleanValue());
         }
 
+        private String completeXpath(String source, String xpath) {
+            String sepSrc = "/";
+            String sepXpath = "/";
+            if (source.indexOf(sepSrc) == -1) sepSrc = "\\.";
+            if (xpath.indexOf(sepXpath) == -1) sepXpath = "\\.";
```
aaime, line 299: Not sure what you're trying to do here.... the dot becomes the separator? For paths like "a.b.c?". But at the same time, you're trying to handle ".." in xpath....

src/community/features-templating/features-templating-core/src/main/java/org/geoserver/featurestemplating/request/TemplatePathVisitor.java
```diff
@@ -286,6 +292,37 @@ private boolean canCompleteXpath(Object extradata) {
...
+            String[] xpathSteps = xpath.split(sepXpath);
+
+            List<String> sourceList = new LinkedList<>();
+            List<String> xpathList = new LinkedList<>();
+            Collections.addAll(sourceList, sourceSteps);
+            Collections.addAll(xpathList, xpathSteps);
+
+            // does the xpath contains all or some portions of the source already?
+            List<String> intersection = new LinkedList<>(xpathList);
+            intersection.retainAll(sourceList);
```
aaime, line 311: I'm confused... what if they share bits, but not in the same order? Recursive data structures might have stuff like "a/b/a/b", legit, but it does not mean that you can compare "a/a" with "a/b/a/b".

PR #6266, opened 2022-10-18
[GEOS-10756] Add support for manual configuration of start and end value in WMS Layer Dimensions

weskamm:
[GEOS-10756](https://osgeo-org.atlassian.net/browse/GEOS-10756)

This adds two new input fields in the Dimensions Tab of a layer.
The fields allow to manually specify the start and end datetime GeoServer should use when generating the capabilities for this layer:

image

Values have to be ISO8601 or the string `PRESENT` to reflect the current DateTime in the Capabilities.
The settings from the screenshot above lead to the following capabilities for that layer:

image

When specifying start and end times, the automatic detection of those which GeoServer usually does is skipped and the given values are used.

This is especially useful when working with layers that are based on very large datasets, as the auto detection may become very slow / unresponding leading to slow / unresponsive capabilities responses.
  

# Checklist

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/WMS.java
+++ b/src/wms/src/main/java/org/geoserver/wms/WMS.java
@@ -25,6 +25,7 @@ import java.util.TreeSet;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
+import org.apache.commons.lang.StringUtils;
 import org.geoserver.catalog.AttributeTypeInfo;
 import org.geoserver.catalog.Catalog;
 import org.geoserver.catalog.CoverageInfo;
@@ -55,6 +56,7 @@ import org.geoserver.util.DimensionWarning;
 import org.geoserver.util.NearestMatchFinder;
 import org.geoserver.wms.WMSInfo.WMSInterpolation;
 import org.geoserver.wms.WatermarkInfo.Position;
+import org.geoserver.wms.capabilities.DimensionHelper;
 import org.geoserver.wms.dimension.DimensionDefaultValueSelectionStrategy;
 import org.geoserver.wms.dimension.DimensionDefaultValueSelectionStrategyFactory;
 import org.geoserver.wms.featureinfo.GetFeatureInfoOutputFormat;
@@ -1273,6 +1275,17 @@ public class WMS implements ApplicationContextAware {
         FeatureCollection collection = getDimensionCollection(typeInfo, time);
 
         TreeSet<Date> result = new TreeSet<>();
+
+        String startValue = time.getStartValue();
+        String endValue = time.getEndValue();
+        if (!StringUtils.isEmpty(startValue)
+                && !StringUtils.isEmpty(endValue)
+                && time.getPresentation() != DimensionPresentation.LIST) {
+            result.add(DimensionHelper.parseTimeRangeValue(startValue));
+            result.add(DimensionHelper.parseTimeRangeValue(endValue));
+            return result;
+        }
+
         if (time.getPresentation() == DimensionPresentation.LIST) {
             final UniqueVisitor visitor = new UniqueVisitor(time.getAttribute());
             collection.accepts(visitor, null);
@@ -1354,6 +1367,16 @@ public class WMS implements ApplicationContextAware {
         FeatureCollection collection = getDimensionCollection(typeInfo, elevation);
 
         TreeSet<Double> result = new TreeSet<>();
+
+        String startValue = elevation.getStartValue();
+        String endValue = elevation.getEndValue();
+        if (!StringUtils.isEmpty(startValue)
+                && !StringUtils.isEmpty(endValue)
+                && elevation.getPresentation() != DimensionPresentation.LIST) {
+            result.add(Double.parseDouble(startValue));
+            result.add(Double.parseDouble(endValue));
+            return result;
+        }
         if (elevation.getPresentation() == DimensionPresentation.LIST
                 || (elevation.getPresentation() == DimensionPresentation.DISCRETE_INTERVAL
                         && elevation.getResolution() == null)) {
@@ -1785,7 +1808,21 @@ public class WMS implements ApplicationContextAware {
         final FeatureCollection fcollection = getDimensionCollection(typeInfo, dimensionInfo);
 
         final TreeSet<Object> result = new TreeSet<>();
-        if (dimensionInfo.getPresentation() == DimensionPresentation.LIST
+
+        String startValue = dimensionInfo.getStartValue();
+        String endValue = dimensionInfo.getEndValue();
+
+        if (dimensionInfo.getPresentation() != DimensionPresentation.LIST
+                && !StringUtils.isEmpty(startValue)
+                && !StringUtils.isEmpty(endValue)) {
+            try {
+                result.add(Double.parseDouble(startValue));
+                result.add(Double.parseDouble(endValue));
+            } catch (NumberFormatException e) {
+                result.add(DimensionHelper.parseTimeRangeValue(startValue));
+                result.add(DimensionHelper.parseTimeRangeValue(endValue));
+            }
+        } else if (dimensionInfo.getPresentation() == DimensionPresentation.LIST
                 || (dimensionInfo.getPresentation() == DimensionPresentation.DISCRETE_INTERVAL
                         && dimensionInfo.getResolution() == null)) {
             final UniqueVisitor uniqueVisitor = new UniqueVisitor(dimensionInfo.getAttribute());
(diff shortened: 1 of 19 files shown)


jodygarnett:
Q: Does the REST API need to be updated with the new fields also?

weskamm:
The two new fields are children of the `dimensionInfo` element, which is used in as an `entry` in the `metadata` set.
It seems that none of the current time configuration parameters are documented.

Anyway, i could add a snippet like
```
<featureType>
  <metadata>
    <entry key="time">
        <dimensionInfo>
            <enabled>true</enabled>
            <attribute>time</attribute>
            <presentation>DISCRETE_INTERVAL</presentation>
            <startValue>2020-05-01T00:00:00Z</startValue>
            <endValue>2021-06-01T00:00:00Z</endValue>
        </dimensionInfo>
    </entry>
  </metadata>
</featureType>
```
to the examples in `configuration.yaml` and `featuretypes.yaml` if its useful

weskamm:
Ok, i have added those snippets to the REST API docs

mprins (changes requested):
This enhancement requires a Jira issue so it shows up in release notes etc.

aaime (changes requested):
I was looking forward to this change, there is indeed a need for statically configured dimensions.

The implementation has significant shortcomings though.

The generic "start/ed of data range" name seems to imply it can be used for elevation and custom dimension as well, and indeed the UI allows for its configuration in both elevation:

image

It also breaks the ability to configure a custom vector dimension, the moment one tries to add one:

https://user-images.githubusercontent.com/325433/197509779-be7ab18a-179d-4267-b3cc-bc3a76a2e286.mp4

Exception details:

```
org.apache.wicket.WicketRuntimeException: The component(s) below failed to render. Possible reasons could be that:
	1) you have added a component in code but forgot to reference it in the markup (thus the component will never be rendered),
	2) if your components were added in a parent container then make sure the markup for the child container includes them in <wicket:extend>.

1. [WebMarkupContainer [Component id = startEndContainer, page = org.geoserver.web.data.resource.ResourceConfigurationPage, path = publishedinfo:tabs:panel:vectorCustomDimPanel:mainDiv:form:vectorCustomDimensionsView:1:vectorCustomDimension:customVectorEditor:configContainer:configs:startEndContainer, type = org.apache.wicket.markup.html.WebMarkupContainer, isVisible = true, isVersioned = false], children =  [TextField [Component id = startValue]][Component id = startValueValidationMsg][TextField [Component id = endValue]][Component id = endValueValidationMsg]]
2. [TextField [Component id = startValue, page = org.geoserver.web.data.resource.ResourceConfigurationPage, path = publishedinfo:tabs:panel:vectorCustomDimPanel:mainDiv:form:vectorCustomDimensionsView:1:vectorCustomDimension:customVectorEditor:configContainer:configs:startEndContainer:startValue, type = org.apache.wicket.markup.html.form.TextField, isVisible = true, isVersioned = false]]
3. [TextField [Component id = endValue, page = org.geoserver.web.data.resource.ResourceConfigurationPage, path = publishedinfo:tabs:panel:vectorCustomDimPanel:mainDiv:form:vectorCustomDimensionsView:1:vectorCustomDimension:customVectorEditor:configContainer:configs:startEndContainer:endValue, type = org.apache.wicket.markup.html.form.TextField, isVisible = true, isVersioned = false]]
```

I believe adding actual support for elevation be easy, please consider it. Have a look at custom dimensions too, if you really don't intend supporting it, be very clear about it in documentation and make sure the UI does not show up and trick users into entering values that will not work.

Implementation wise, it covers WMS, however, the configurations are global and there are other services dealing with dimensions, for example WCS 2.0 reports time as well (see WCSDimensionHelper). The same goes for the WMTS multidimensional extension, and for time support in OGC API.

I guess they will all carry on using queries, which may be the source for inconsistency across services (some reporting declared values, others reporting actual one).
Personally, I would not ask you to go an implement consistent support across all services, but it should be clear from documentation that the scope of these two new settings is limited to WMS and users should expect a different behavior elsewhere.
That said, it's just my opinion, I recommend discussing this topic on the developer list.

Testing is a bit limited too, can you cover WMS 1.1 and continuous interval presentation as well?

aaime:
And oh, another thing... a few months ago a colleague of mine tried to implement something similar, but based on a fixed list of values (we did not end up contributing it as the implementation was too poor).

The context was that if one has vector data with hundreds of millions of records, but relatively few unique scattered dimension values (few thousands), the call to a distinct on such dataset would have taken too much time, even on an index. So the UI allowed to enter a comma-separated list of values.

Here instead you're concentrating on start/end, which normally are pretty fast to read from an index, because they are the beginning and end of the index tree (so a few steps away from the root of the b-tree, no need to walk it all). And yet here you're trying to optimize away exactly that... wondering what's the context?
Of course not having an index would be a cause, but that would slow down every request, not just capabilities (the fix is in the database in that case). Maybe you have a lot of layers configured with start/end, and you're just trying to avoid making a lot of queries, that even if fast, would have a cost in terms of latency?

src/wms/src/main/java/org/geoserver/wms/capabilities/DimensionHelper.java
```diff
@@ -387,11 +388,41 @@ private void handleTimeDimensionRaster(
...
+                    } else {
+                        try {
+                            DateTimeParser dateTimeParser = new DateTimeParser();
+                            Date startDateTime =
+                                    (Date) ((List) dateTimeParser.parse(startValue)).get(0);
+                            temporalDomain.add(startDateTime);
+                        } catch (Exception e) {
+                            throw new ServiceException("Unable to parse the start-date", e);
+                        }
+                    }
```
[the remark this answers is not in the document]
weskamm: Done. Had to make the DimensionHelper public for this

weskamm:
So i have extended this to also cover elevation and custom dimensions.
Tests have been extended and now also cover WMS 1.1.1.
Please review again.

As for your question, yes, we had performance issues with lots of layers based on massive datasets, which in sum make the capabilities slow even when using indices

weskamm:
not sure why the jdk 17 check is failing, any hints?

src/wms/src/main/java/org/geoserver/wms/WMS.java
```diff
@@ -1273,6 +1275,15 @@ public TreeSet<Date> getFeatureTypeTimes(FeatureTypeInfo typeInfo) throws IOExce
         FeatureCollection collection = getDimensionCollection(typeInfo, time);
 
         TreeSet<Date> result = new TreeSet<>();
+
+        String startValue = time.getStartValue();
+        String endValue = time.getEndValue();
+        if (!StringUtils.isEmpty(startValue) && !StringUtils.isEmpty(endValue)) {
```
aaime, line 1281: The configuration can be changed via UI (which I see is resetting the fields contents if list presentation is chosen), but can also be modified via REST, often without full understanding of how things are supposed to work. I'd recommend adding a ``&& time.getPresentation != DimensionPresentation.LIST``, or similar condition, to make sure the presentation and configured values are compatible.

src/wms/src/main/java/org/geoserver/wms/WMS.java
```diff
@@ -1354,6 +1365,14 @@ public TreeSet<Double> getFeatureTypeElevations(FeatureTypeInfo typeInfo) throws
         FeatureCollection collection = getDimensionCollection(typeInfo, elevation);
 
         TreeSet<Double> result = new TreeSet<>();
+
+        String startValue = elevation.getStartValue();
+        String endValue = elevation.getEndValue();
+        if (!StringUtils.isEmpty(startValue) && !StringUtils.isEmpty(endValue)) {
```
aaime, line 1371: Same as above.

aaime (changes requested):
Looks quite good, thanks! Found minor issues here and there, should be quick fixes.

weskamm:
fixed

aaime:
All changes look good. While checking the code a new question came to mind: how do you see this new setting play with the default value selection strategy? 

The current documentation and implementation both say the same thing: this change affects only the capabilities. If a user selected "min" or "max" as the default value, they will still be pulled from the data source, causing a protocol violation (the value declared in the caps is not the one used in GetMap). 
This can be fixed by declaring the default value also as a fixed value, repeating the value chosen in the static min/max fields. It's inconvenient, but users could be warned about it.

Is this what you expected users to do?

aaime (approved):

weskamm:
> Actually wait... the default value declared in the capabilities does not come from min/max set in the new GUI elements, or does it?

No it doesnt. Its just that what the user configures. If one chooses to set a default value outside of the static range, it will appear like that in the caps

aaime:
Ok, merged :-D

weskamm:
@aaime what has to be done to get a backport of this PR? Could you give me a hint on this? And if this gets backported, would it be included in the upcoming 2.22.1 Release? Thanks!

aaime:
For backport on a stable series you have to ask on the devel list, showing that if the new functionality is not used, then there is no chance of regression. Also show that it will work seamlessly with existing data directories (no requirement for changes to the config contents).

PR #6280, opened 2022-10-20
[GEOS-10715] - Workspace Filter for PublishedInfo does not include LayerGroups

davidblasby:
[GEOS-10715](https://osgeo-org.atlassian.net/browse/GEOS-10715)

See [GEOS-10715](https://osgeo-org.atlassian.net/browse/GEOS-10715)

To reproduce - 

a) create a user with only access to NE workspace
b) got to GS Homepage  or Layer Preview
c) you would expect to see 5 “layers” - 4 real layers and one LayerGroup (“world”)
d) You don’t see the “world” LayerGroup

Underlying cause is the security filter for the catalog uses layer-specific check ([ resource.store.workspace.name = topp ]) that doesn’t work on a layerGroup.

Solution - use a filter that handles both LayerGroup and Layer - “[[ workspace.name = topp ] OR [ resource.store.workspace.name = topp ]]”

NOTE: one of the test cases (SecureCatalogImplTest#testSecurityFilterComplex) relies on the incorrect behaviour. 
  

# Checklist

- [x ] I have sent a [Contribution Licence Agreement](https://docs.geoserver.org/latest/en/developer/policies/committing.html) (not required for small changes, e.g., fixing typos in documentation).

### DIFF
--- a/src/security/security-tests/src/test/java/org/geoserver/security/impl/SecureCatalogImplTest.java
+++ b/src/security/security-tests/src/test/java/org/geoserver/security/impl/SecureCatalogImplTest.java
@@ -1098,24 +1098,25 @@ public class SecureCatalogImplTest extends AbstractAuthorizationTest {
         // Workspace test
         Class<? extends CatalogInfo> clazz = WorkspaceInfo.class;
         // Creating filter for anonymous user
-        Filter security = resourceManager.getSecurityFilter(anonymous, clazz);
+        Filter securityAnonymous = resourceManager.getSecurityFilter(anonymous, clazz);
         // Creating filter for read write user
-        Filter security2 = resourceManager.getSecurityFilter(rwUser, clazz);
+        Filter securityReadWriteUser = resourceManager.getSecurityFilter(rwUser, clazz);
         // Creating filter for military user
-        Filter security3 = resourceManager.getSecurityFilter(milUser, clazz);
+        Filter securityMilitary = resourceManager.getSecurityFilter(milUser, clazz);
         // anonymous and military can access only to topp
-        assertNotSame(security, Filter.EXCLUDE);
-        assertNotSame(security, Filter.INCLUDE);
-        assertSame(security2, Filter.INCLUDE);
-        assertNotSame(security3, Filter.EXCLUDE);
-        assertNotSame(security3, Filter.INCLUDE);
+        assertNotSame(securityAnonymous, Filter.EXCLUDE);
+        assertNotSame(securityAnonymous, Filter.INCLUDE);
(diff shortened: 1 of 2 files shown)


src/security/security-tests/src/test/java/org/geoserver/security/impl/SecureCatalogImplTest.java
```diff
@@ -1151,8 +1151,8 @@ public void testSecurityFilterComplex() throws Exception {
         Collection<PublishedInfo> filteredPublished =
                 Collections2.filter(publisheds, new PredicateFilter(security));
         assertEquals(4, filteredLayers.size());
-        assertEquals(0, filteredGroups.size());
-        assertEquals(4, filteredPublished.size());
+        assertEquals(2, filteredGroups.size());
+        assertEquals(6, filteredPublished.size());
```
jodygarnett, line 1155: These numbers agree with your fix; but not with the test case comment above:
```
        // Check the count does not include the groups. Since the catalog is just a mock, we
        // extract the layers and groups, and do counts using the security filter
```

So the comment is "accurate" documenting the incorrect behaviour that you have now fixed :)

I have *no idea* why the previous behaviour of the security filter was to ignore layer groups; it makes no sense as we need those to be included for layer preview page to include them?!?

jodygarnett (changes requested):
Your fix is good; I am going to make a non-functional change to the test case comment and variable names. Thanks for the prompt fix David.

src/security/security-tests/src/test/java/org/geoserver/security/impl/SecureCatalogImplTest.java
davidblasby: I think it was probably missed because this functionality is not turned on if you are admin or not filtering by workspace.  I think your changes to the homepage made this more visible and used.  Good to catch it!

jodygarnett:
David I looked at adding to DefaultResourceAccessManagerTest but all the action is in SecureCatalogImplTest. I have rebased and pushed up your branch and the test case reads much easier.

jodygarnett (commented):
While I have updated the test case (much improved) I would like someone more familiar with the catalog to review. @groldan are you available?

src/security/security-tests/src/test/java/org/geoserver/security/impl/SecureCatalogImplTest.java
jodygarnett: I agree; renaming the variables to improve code readability, and stepping through the code multiple times with the debugger has me confident the fix is now doing the right thing.

jodygarnett (approved):

PR #6289, opened 2022-10-21
[Fixes GEOS-10718] the OIDC plugin does not currently take into account the id_token_hint parameter

afabiani:
[GEOS-10718](https://osgeo-org.atlassian.net/browse/GEOS-10718)

References: https://osgeo-org.atlassian.net/browse/GEOS-10718

# Checklist

### DIFF
--- a/src/community/security/oauth2-openid-connect/oauth2-openid-connect-core/src/main/java/org/geoserver/security/oauth2/OpenIdConnectAuthenticationFilter.java
+++ b/src/community/security/oauth2-openid-connect/oauth2-openid-connect-core/src/main/java/org/geoserver/security/oauth2/OpenIdConnectAuthenticationFilter.java
@@ -25,15 +25,18 @@ import net.sf.json.JSONObject;
 import org.geoserver.platform.GeoServerExtensions;
 import org.geoserver.security.config.RoleSource;
 import org.geoserver.security.config.SecurityNamedServiceConfig;
+import org.geoserver.security.filter.GeoServerLogoutFilter;
 import org.geoserver.security.impl.GeoServerRole;
 import org.geoserver.security.impl.RoleCalculator;
 import org.geoserver.security.oauth2.bearer.TokenValidator;
 import org.geoserver.security.oauth2.services.OpenIdConnectTokenServices;
+import org.springframework.security.core.Authentication;
 import org.springframework.security.jwt.Jwt;
 import org.springframework.security.jwt.JwtHelper;
 import org.springframework.security.oauth2.client.OAuth2RestOperations;
 import org.springframework.security.oauth2.client.token.DefaultRequestEnhancer;
 import org.springframework.security.oauth2.client.token.grant.code.AuthorizationCodeAccessTokenProvider;
+import org.springframework.security.oauth2.common.OAuth2AccessToken;
 import org.springframework.security.oauth2.provider.authentication.OAuth2AuthenticationDetails;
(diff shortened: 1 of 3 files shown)


src/community/security/oauth2-openid-connect/oauth2-openid-connect-core/src/main/java/org/geoserver/security/oauth2/OpenIdConnectFilterConfig.java
```diff
@@ -100,4 +100,12 @@ protected StringBuilder buildAuthorizationUrl() {
             sb.append("&response_mode=").append(responseMode);
         return sb;
     }
+
+    protected StringBuilder buildEndSessionUrl(final String idToken) {
```
aaime, line 104: Found documentation about it here: 
https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout

It says recommended, however in other places (e.g., [Okta](https://developer.okta.com/docs/reference/api/oidc/#logout)) it was documented as mandatory.

So I guess it's a good thing to have :-D

src/community/security/oauth2-openid-connect/oauth2-openid-connect-core/src/main/java/org/geoserver/security/oauth2/OpenIdConnectAuthenticationFilter.java
```diff
@@ -265,4 +268,31 @@ private void enrichWithRoleCalculator(List<GeoServerRole> roles) throws IOExcept
...
+                    (String) request.getAttribute(OpenIdConnectAuthenticationFilter.ID_TOKEN_VALUE);
+        } else {
+            OAuth2AccessToken token = restTemplate.getOAuth2ClientContext().getAccessToken();
+            if (token != null && token.getAdditionalInformation() != null) {
+                Object maybeIdToken = token.getAdditionalInformation().get("id_token");
+                if (maybeIdToken instanceof String) {
+                    idToken = (String) maybeIdToken;
+                }
+            }
+        }
```
aaime, line 290: How comes there are two possible places where to look up this the idToken?

taba90: I might be wrong but I guess probably because the request attribute has request scope, while the restTemplate is a bean with session scope.
When logging out from UI probably is likely that the fallback logic will be applied.

taba90 (approved):
Looks good to me

src/community/security/oauth2-openid-connect/oauth2-openid-connect-core/src/main/java/org/geoserver/security/oauth2/OpenIdConnectAuthenticationFilter.java
afabiani: As @taba90 said!

src/community/security/oauth2-openid-connect/oauth2-openid-connect-core/src/main/java/org/geoserver/security/oauth2/OpenIdConnectFilterConfig.java
afabiani: Yep, currently I would say this is **strongly** recommended, in most of the cases **mandatory**.

In any case having this won't break the protocol.

aaime:
Then in it goes!

PR #6304, opened 2022-10-27
[GEOS-10723] params.extractor Resource

dieterstueken:
[GEOS-10723](https://osgeo-org.atlassian.net/browse/GEOS-10723)

Classes from `org.geoserver.params.extractor` frequently use `InputStreams` and `OutputStreams`. Doing so, it is not clear if the producer or the consumer has to close the stream. Using `Supplier<*Stream>` does not help either. This also leads to additional catch/try/resource blocks spread all over the code. This approach uses `org.geoserver.platform.resource.Resource` instead.

This way the catch/try/resource blocks can be kept small and consistent.
This also enables to consider `Type.UNDEFINED` concerning [GEOS-10705].

# Checklist

### DIFF
--- a/src/extension/params-extractor/src/main/java/org/geoserver/params/extractor/EchoParametersDao.java
+++ b/src/extension/params-extractor/src/main/java/org/geoserver/params/extractor/EchoParametersDao.java
@@ -4,7 +4,6 @@
  */
 package org.geoserver.params.extractor;
 
-import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.ArrayList;
@@ -12,7 +11,6 @@ import java.util.List;
 import java.util.UUID;
 import java.util.function.Consumer;
 import java.util.function.Function;
-import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import org.apache.commons.lang3.ArrayUtils;
@@ -50,27 +48,36 @@ public final class EchoParametersDao {
 
     public static List<EchoParameter> getEchoParameters() {
         Resource echoParameters = getDataDirectory().get(getEchoParametersPath());
-        return getEchoParameters(echoParameters.in());
+        return getEchoParameters(echoParameters);
     }
 
     private static GeoServerDataDirectory getDataDirectory() {
         return (GeoServerDataDirectory) GeoServerExtensions.bean("dataDirectory");
     }
 
-    @SuppressWarnings("PMD.UseTryWithResources") // cannot use try-with-resources here
-    public static List<EchoParameter> getEchoParameters(InputStream inputStream) {
-        try {
-            if (inputStream.available() == 0) {
-                Utils.debug(LOGGER, "Echo parameters file seems to be empty.");
-                return new ArrayList<>();
+    /**
+     * Read a list of EchoParameter from a resource. Return an empty list if the resource does not
+     * exist. This prevents Resource.in() from creating the file or throwing an exception.
+     *
+     * @param resource to read.
+     * @return a list of EchoParameter or an empty list if the resource does not exist.
+     */
+    public static List<EchoParameter> getEchoParameters(Resource resource) {
+        if (resource.getType() == Resource.Type.RESOURCE) {
+            try (InputStream inputStream = resource.in()) {
+                if (inputStream.available() == 0) {
+                    Utils.debug(LOGGER, "Echo parameters file seems to be empty.");
+                } else {
+                    EchoParametersList list = (EchoParametersList) xStream.fromXML(inputStream);
+                    return list.parameters == null ? new ArrayList<>() : list.parameters;
+                }
+            } catch (Exception exception) {
+                throw Utils.exception(exception, "Error parsing echo parameters files.");
             }
-            EchoParametersList list = (EchoParametersList) xStream.fromXML(inputStream);
-            return list.parameters == null ? new ArrayList<>() : list.parameters;
-        } catch (Exception exception) {
-            throw Utils.exception(exception, "Error parsing echo parameters files.");
-        } finally {
-            Utils.closeQuietly(inputStream);
+        } else {
+            Utils.info(LOGGER, "Echo parameters file does not exist.");
         }
+        return new ArrayList<>();
     }
 
     public static void saveOrUpdateEchoParameter(EchoParameter echoParameter) {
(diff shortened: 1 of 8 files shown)


dieterstueken:
A general problem with `Resource.in()` is, that it returns an `InputStream` to the client code which works with it in different and questionable ways. So, I thought about the functional approach again: 
Almost all usages (126!) try to convert the given `Resource` into some object using try/catch/resource with additional if/else conditions if it is a resource at all and what to do else.

All that boilerplate code can easily be avoided by using a generator `Function<InputStream, T>` and finally returning an Optional<T>. Using an Optional<T> also eases all those else{default} blocks by simply using `Optional.orElse(T)`.

I try to demonstrate it on this PR.

src/extension/params-extractor/src/main/java/org/geoserver/params/extractor/EchoParametersDao.java
```diff
@@ -50,27 +49,40 @@ public static String getTmpEchoParametersPath() {
...
-            }
+    /**
+     * Read a list of EchoParameter from a resource. Return an empty list if the resource does not
+     * exist. This prevents Resource.in() from creating the file or throwing an exception.
+     *
+     * @param resource to read.
+     * @return a list of EchoParameter or an empty list if the resource does not exist.
+     */
+    public static List<EchoParameter> getEchoParameters(Resource resource) {
+        return resource.optionalRead(EchoParametersDao::read).orElseGet(ArrayList::new);
```
dieterstueken, line 67: this looks much clearer than the deeply cascaded if/try blocks before.

src/extension/params-extractor/src/main/java/org/geoserver/params/extractor/EchoParametersDao.java
```diff
@@ -83,52 +95,44 @@ public static void saveOrUpdateEchoParameter(EchoParameter echoParameter) {
 
     public static void saveOrUpdateEchoParameter(
             EchoParameter echoParameter, Resource input, Resource output) {
-        try (InputStream inputStream = input.in();
```
dieterstueken, line 86: combining input and output will also leave an empty output file if reading fails. Better to separate this. It's also easier to read.

src/extension/params-extractor/src/test/java/org/geoserver/params/extractor/TestSupport.java
```diff
@@ -59,18 +58,12 @@ private void deleteTestDirectoryQuietly() {
         }
     }
 
-    protected static void doWork(String resourcePath, Consumer<InputStream> consumer)
```
dieterstueken, line 62: simplified this a bit. This avoids too excessive lambda expressions and significantly improves the readability.

OK, not strictly related to the original topic; split this off into a separate PR?

aaime:
One legit QA failure:

```
2022-10-28T14:25:43.0874191Z 14:25:43,087 [INFO] PMD Failure: org.geoserver.platform.resource.Resource:8 Rule:wildcards Priority:3 No Wildcard Imports.
```

(wildcard imports are disallowed by the Google Java coding convention we adopted).

Besides that, the changes to the params-extractor module look good to me, but I'd like to have @jodygarnett review the changes to Resource. Thinking out loud, it looks like a good addition, but it's also asymmetric, probably the same methods should be offered on the write side? Anyways, let's hear from @jodygarnett

dieterstueken:
Yes, sorry.

I have to convince my IDE (Intellj)  not to do so.
And I’m too impatient in pushing 😊
I may try to find a spotless plugin or something.
Currently my IDE takes another 7 minutes to compile everything  again after a spoltless:apply.
And if I’m unheeding it bricks it again. Hope I can improve this soon.

Dieter.
Von: Andrea Aime ***@***.***>
Gesendet: Montag, 31. Oktober 2022 12:28
An: geoserver/geoserver ***@***.***>
Cc: Dieter Stüken - con terra GmbH ***@***.***>; Author ***@***.***>
Betreff: Re: [geoserver/geoserver] [GEOS-10723] params.extractor Resource (PR #6304)


One legit QA failure:

2022-10-28T14:25:43.0874191Z 14:25:43,087 [INFO] PMD Failure: org.geoserver.platform.resource.Resource:8 Rule:wildcards Priority:3 No Wildcard Imports.

(wildcard imports are disallowed by the Google Java coding convention we adopted).

Besides that, the changes to the params-extractor module look good to me, but I'd like to have @jodygarnett<https://eur04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fjodygarnett&data=05%7C01%7C%7Cc1b7c2c43fbc40fd257d08dabb32ebe4%7C6e0bfede3fcb4518a16565dc14fe5620%7C0%7C0%7C638028124624166419%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=mkFbjIlaE65ed%2BbhwVskZrraFYaW8eA6Nhq3gn%2BEUKE%3D&reserved=0> review the changes to Resource. Thinking out loud, it looks like a good addition, but it's also asymmetric, probably the same methods should be offered on the write side? Anyways, let's hear from @jodygarnett<https://eur04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fjodygarnett&data=05%7C01%7C%7Cc1b7c2c43fbc40fd257d08dabb32ebe4%7C6e0bfede3fcb4518a16565dc14fe5620%7C0%7C0%7C638028124624322646%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=j7GdhSXsKsnkAsq7gMr8rEHMvEgomRqjDjrK8leZRTw%3D&reserved=0>

src/extension/params-extractor/src/main/java/org/geoserver/params/extractor/EchoParametersDao.java
jodygarnett, line 67: This makes a strong case for your api change I like it

src/extension/params-extractor/src/main/java/org/geoserver/params/extractor/RulesDao.java
```diff
@@ -43,40 +43,53 @@ public static String getTempRulesPath() {
...
 
-    public static List<Rule> getRules(Supplier<InputStream> iss) {
-        try (InputStream inputStream = iss.get()) {
-            if (inputStream.available() == 0) {
-                Utils.info(LOGGER, "Parameters extractor rules file seems to be empty.");
-                return new ArrayList<>();
-            }
+    /**
+     * Read a list of Rule from a resource. Return an empty list if the resource does not exist.
+     * This prevents Resource.in() from creating the file or throwing an exception.
```
jodygarnett, line 55: This comment is no is not part of the api; if you wish to added it below you can.
```suggestion
```

src/extension/params-extractor/src/main/java/org/geoserver/params/extractor/RulesDao.java
```diff
@@ -43,40 +43,53 @@ public static String getTempRulesPath() {
...
+     *
+     * @param resource to read.
+     * @return a list of Rule or an empty list if the resource does not exist.
+     */
+    public static List<Rule> getRules(Resource resource) {
```
jodygarnett, line 60:
```suggestion
    public static List<Rule> getRules(Resource resource) {
        // This prevents Resource.in() from creating the file or throwing an exception.
```

src/extension/params-extractor/src/test/java/org/geoserver/params/extractor/TestSupport.java
jodygarnett: I think it is fine here; it is part of your theme of trying to make resource access less difficult; and TestSupport is not an api change for the web application

src/platform/src/main/java/org/geoserver/platform/resource/Resource.java
```diff
@@ -197,6 +196,44 @@ default byte[] getContents() throws IOException {
         }
     }
 
+    /**
+     * Read and convert this resource into a result T using a converter function. The Reader
+     * function may (but does not need to) throw an IOException. Returns null if this resource does
+     * not exist. Reading a {@code TYPE.DIRECTORY} resource will raise an IOException.
```
jodygarnett, line 202: Thanks for writing clear javadocs; consider adding a small code example here

jodygarnett (commented):
Some small feedback; I rally like the change and clear style of programming it encourages.

dieterstueken:
Pushed the suggested cleanup.

> @aaime: ... but it's also asymmetric, probably the same methods should be offered on the write side?

Yes, I think the same. But this may better be solved in context of [GEOS-10705] in general, as it is not related to `params.extractor`. After this gets pushed, I can continue with some other issues related to `.in()`.

aaime:
@dieterstueken we try not to have work done half-way, because the other half is not guaranteed to show up. Could you also add the methods for the out side?

dieterstueken:
Ok,
I added a simple `write` method using an `IOConsumer` and rewrote `setContents` to demonstrate its usage similar to `getContents`.

But I am not too happy about by now. It would be better to have some functional interface like:
`void write(T data, OutputStream out) throws IOException;`. But this is missing on `org.apache.commons.io`.

I have additional suggestions about this, but it involves some more extensive refactoring going beyond the topic of this issue.

jodygarnett:
> Ok, I added a simple `write` method using an `IOConsumer` and rewrote `setContents` to demonstrate its usage similar to `getContents`.
> 
> But I am not too happy about by now. It would be better to have some functional interface like: `void write(T data, OutputStream out) throws IOException;`. But this is missing on `org.apache.commons.io`.
> 
> I have additional suggestions about this, but it involves some more extensive refactoring going beyond the topic of this issue.

Would it help to create your own interface? I have a small hesitation about using org.apache.commons.io interfaces directly in our API. I know we use such libraries internally; but it is different including them on api boundaries.

aaime:
@dieterstueken that's why we strongly suggest to have a single focus on a single PR. In this one you are already working on two.
Options:
- Roll back the changes to FileResource and just focus on the params extractor module change
- Create a separate commit for the changes to FileReosurce and make that complete

The preferred approach is the first, but since you just recently started to contributed, I'm happy to go with the second too. Just not happy to merge a work done half-way.

dieterstueken:
> @dieterstueken that's why we strongly suggest to have a single focus on a single PR.

You are right. I was tempted to introduce the functional approach within this PR, which was possibly off topic.
I may try to focus on the main aspect again without changing Resource at first.
Is it OK to perform a rebase, even if the comments on it by @jodygarnett may vanish then?

After the original approach was finished, I can start again introducing the functional approach as a separate PR.
Since I'm just coming home drunken, I have to shift it to tomorrow🍺

> ... since you just recently started to contribute ...

yes, I'm still on the learning curve and to impatient, too.

aaime:
@dieterstueken there is a way to keep both:
- Close this PR, without further modification
- Create a new clean one, with focused changes

You can refer to this one from newer PRs and nothing is lost. Just an idea!

dieterstueken:
This pull request became too expansive concerning functional extensions of `Resource`.
So, it is closed and replaced by [#6331](https://github.com/geoserver/geoserver/pull/6331).
The aspects of functional extensions of `Resource` will then go to a separate issue.

PR #6312, opened 2022-10-27
[GEOS-10725] : [WFS] : GetFeature Performance with jdbc plugin

eropartz:
[GEOS-10725](https://osgeo-org.atlassian.net/browse/GEOS-10725)

  

# Checklist

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/CatalogNamespaceSupport.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/CatalogNamespaceSupport.java
@@ -6,10 +6,9 @@
 package org.geoserver.wfs;
 
 import java.util.Enumeration;
+import java.util.Iterator;
 import org.geoserver.catalog.Catalog;
 import org.geoserver.catalog.NamespaceInfo;
-import org.geoserver.catalog.Predicates;
-import org.geoserver.catalog.util.CloseableIterator;
 import org.xml.sax.helpers.NamespaceSupport;
 
 /** NamespaceContext based on GeoServer catalog. */
@@ -30,16 +29,13 @@ public class CatalogNamespaceSupport extends NamespaceSupport {
 
     @Override
     public Enumeration getPrefixes() {
-        @SuppressWarnings("PMD.CloseResource") // best effort closing
-        final CloseableIterator<NamespaceInfo> it =
-                catalog.list(NamespaceInfo.class, Predicates.acceptAll());
+        final Iterator<NamespaceInfo> it = catalog.getNamespaces().iterator();
         return new Enumeration() {
             @Override
             public boolean hasMoreElements() {
                 if (it.hasNext()) {
                     return true;
                 } else {
-                    it.close();
                     return false;
                 }
             }


aaime:
@groldan could you have a look at this one?

eropartz:
Hi @aaime, did you have time to look this PR? It's important for us to fix this performance problem on jdbcconfig
Many Thanks

aaime:
I want someone that knows about JDBCConfig better, as noted above, this change reverts another commit that was done specifically to make JDBCConfig work with large number of layers (so it seems it would make one thing better at the expense of a regression in another use case).

If @groldan or @NielsCharlier could have a look it would be great. I won't merge the PR myself.

NielsCharlier:
I do not think this change is harmful, but I do not understand how it will help.

As far as I understand, the original commit introduced the entire class to reduce the amount of queries to the db, this is not being undone. But I do not see how getNamespaces() would be more efficient than list(). As far as I can see, the first method ends up in ConfigDatabase.getAll and the second ends up in ConfigDatabase.query. Both do a query to the databse to get the ids, and then use the cache to load the existing objects from the id's. But I could be missing something.

@eropartz Can you clarify how this change is supposed to reduce db access?

NielsCharlier:
An approach that I would expect to help is to let CatalogNamespaceSupport keep the namespaces in memory and make it a CatalogListener that updates itself when namespaces are added/removed/changed

aaime:
@NielsCharlier the difference I see is that the previous code was iterating over a database resultset (not evident maybe, but that's what's happening when one gathers an iterator that way no?) while the new code loads all namespaces in an in-memory list first, and then opens an iterator over the list. Which is, what I believe GEOS-6597 was trying to prevent, I believe...

NielsCharlier:
@aaime I don't think that the iterator iterates over the db resultset. If I read DbConfig.query, it seems that it executes a query for ID's which are all loaded in memory. Then the iterator will, each time, check the cache and if not present execute another query to get the object. I don't understand yet the purpose of this change.

NielsCharlier:
the db is used for querying the ID's because normally the entire catalog is always in memory and if you are clustering you use hz-cluster to reload from the database only when a change has been made on another node.

sikeoka:
The query for `catalog.list(NamespaceInfo.class, Predicates.acceptAll());` is like:
`select id from object where type_id in (2) AND ((1=1)) ORDER BY oid`
and the query for `catalog.getNamespaces()` is like:
`select id from object where type_id in (2) order by id`
and the extra `1=1` is irrelevant, at least in postgresql.  The additional database queries for each namespace all work with the cache regardless of which catalog method is called.  

The only difference is that the existing code will lazily load each namespace from the cache/database as it is read from the iterator and the new code will load all of the namespaces at once.  It seems to me like the existing code should work better in certain cases and I'm not sure how this change is fixing the described problem.

eropartz:
Hi @aaime, @NielsCharlier, @sikeoka 

Thanks for the review.  I'm agree with you that both methods ( list and query) in ConfigDatabase use the cache to load objects.

After some investigation:
-  The method list is not using Predicate and the SQL is more simpler
-  The iteration over namespaces seems to be faster (list use Iterator whereas query use CloseableIterator)

**SQL generated by list when CatalogNamespaceSupport.getPrefixes is called** 
`14 déc. 17:46:06 DEBUG [jdbcconfig.internal] - select id from object where type_id in ( 8 ) order by id`
`14 déc. 17:46:06 DEBUG [jdbcconfig.internal] - query returned 384 records in 118.9 ms`
`14 déc. 17:46:07 DEBUG [jdbcconfig.internal] - select id from object where type_id in ( 2 ) order by id`
`14 déc. 17:46:07 DEBUG [jdbcconfig.internal] - query returned 4 records in 97.23 ms`

**SQL generated by query when CatalogNamespaceSupport.getPrefixes is called**
`14 déc. 17:19:21 DEBUG [jdbcconfig.internal] - select id from object where type_id in ( 8 ) order by id`
`14 déc. 17:19:21 DEBUG [jdbcconfig.internal] - query returned 384 records in 138.4 ms`
`14 déc. 17:19:22 DEBUG [jdbcconfig.internal] - Original filter: [Filter.INCLUDE AND [ true = InternalFunctionImpl() ] AND Filter.INCLUDE AND [ true = InternalFunctionImpl() ] AND Filter.INCLUDE]`
`14 déc. 17:19:22 DEBUG [jdbcconfig.internal] - Supported filter: Filter.INCLUDE`
`14 déc. 17:19:22 DEBUG [jdbcconfig.internal] - Unsupported filter: [[ true = InternalFunctionImpl() ] AND [ true = InternalFunctionImpl() ]]`
`14 déc. 17:19:22 DEBUG [jdbcconfig.internal] - select id from object where type_id in (2) AND (
(1=1) /* INCLUDE */
)
 ORDER BY oid`
`14 déc. 17:19:22 DEBUG [jdbcconfig.internal] - query returned 4 records in 26.30 ms`


** Response time for the same WFS request in my environment** 
- List : 7s
- Query : 4,5s

Regards

NielsCharlier:
@eropartz how do you mean exactly 'SQL is more simpler", the condition 1=1 has no effect

I want to understand the cause of the delay better... if it is indeed the closeableiterator, which is possible, I don't know, can we fix that then? I also notice there is an Unsupported filter: "true = InternalFunctionImpl()" -> what is the cause and purpose of this and could this be causing delay?

PR #6314, opened 2022-10-27
[GEOS-10724] SpringResourceAdaptor should throw a FileNotFoundException instead of creating any missing file

dieterstueken:
[GEOS-10724](https://osgeo-org.atlassian.net/browse/GEOS-10724)

The `SpringResourceAdaptor` wraps an `org.geoserver.platform.resource.Resource` into a `org.springframework.core.io.Resource`.

Both are quite similar, but while the geoserver Resource wraps any `IOException` into an `IllegalStateException`  
(which is fine to use it in λ-expressions), the spring Resource however can and should throw the `IOException` itself.

Looking at [GeoServerPropertyConfigurer:104](https://github.com/geoserver/geoserver/blob/190f8849b9a2da54f709f227f3ad3e93a1194f41/src/main/src/main/java/org/geoserver/config/GeoServerPropertyConfigurer.java#L104) it seems, the missing config file shall be created and filled.
However, `super.loadProperties` indirectly creates some empty config file instead. 

Since this situation was not covered before, the first commit adds some unit test to show up the problem.

The second commit modifies `SpringResourceAdaptor`  to throw `FileNotFoundException` for `Resource.Type.UNDEFINED`.

Remaining question: what about `Resource.Type.DIRECTORY`?

# Checklist

- [x] There is an issue in the [GeoServer Jira](https://osgeo-org.atlassian.net/browse/GEOS/summary)

### DIFF
--- a/src/main/src/main/java/org/geoserver/config/SpringResourceAdaptor.java
+++ b/src/main/src/main/java/org/geoserver/config/SpringResourceAdaptor.java
@@ -1,4 +1,4 @@
-/* (c) 2015 Open Source Geospatial Foundation - all rights reserved
+/* (c) 2022 Open Source Geospatial Foundation - all rights reserved
  * This code is licensed under the GPL 2.0 license, available at the root
  * application directory.
  */
@@ -107,9 +107,16 @@ public class SpringResourceAdaptor implements org.springframework.core.io.Resour
         return file;
     }
 
+    /**
+     * {@link org.springframework.core.io.Resource#contentLength()} suggests to throw an IOException
+     * while {@link java.io.File#length()} suggests to return 0L instead, which is used here to
+     * kinder the client code. Directory resources also return 0L.
+     *
+     * @return the content length for this resource or 0L if it is not a file.
+     */
     @Override
-    public long contentLength() throws IOException {
-        return getFile().length();
+    public long contentLength() {
+        return resource.getType() == RESOURCE ? resource.file().length() : 0;
     }
 
     @Override
@@ -118,8 +125,7 @@ public class SpringResourceAdaptor implements org.springframework.core.io.Resour
     }
 
     @Override
-    public org.springframework.core.io.Resource createRelative(String relativePath)
-            throws IOException {
+    public org.springframework.core.io.Resource createRelative(String relativePath) {
         return new SpringResourceAdaptor(resource.get(relativePath));
     }
 
--- a/src/main/src/main/java/org/geoserver/config/SpringResourceAdaptor.java
+++ b/src/main/src/main/java/org/geoserver/config/SpringResourceAdaptor.java
@@ -4,7 +4,11 @@
  */
 package org.geoserver.config;
 
+import static org.geoserver.platform.resource.Resource.Type.RESOURCE;
+import static org.geoserver.platform.resource.Resource.Type.UNDEFINED;
+
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
@@ -54,6 +58,9 @@ public class SpringResourceAdaptor implements org.springframework.core.io.Resour
 
     @Override
     public InputStream getInputStream() throws IOException {
+        if (resource.getType() != RESOURCE) {
+            throw new FileNotFoundException(resource.path());
+        }
         return resource.in();
     }
 
@@ -63,8 +70,13 @@ public class SpringResourceAdaptor implements org.springframework.core.io.Resour
     }
 
     @Override
+    public boolean isFile() {
+        return resource != null && resource.getType() == RESOURCE;
+    }
+
+    @Override
     public boolean isReadable() {
-        return Resources.canRead(resource);
+        return resource.getType() != UNDEFINED && Resources.canRead(resource);
     }
 
     @Override
@@ -74,22 +86,30 @@ public class SpringResourceAdaptor implements org.springframework.core.io.Resour
 
     @Override
     public URL getURL() throws IOException {
-        return Resources.find(resource).toURI().toURL();
+        return getFile().toURI().toURL();
     }
 
     @Override
     public URI getURI() throws IOException {
-        return Resources.find(resource).toURI();
+        return getFile().toURI();
     }
 
+    /**
+     * @return a File handle for this resource.
+     * @throws FileNotFoundException if the resource is UNDEFINED.
+     */
     @Override
     public File getFile() throws IOException {
-        return Resources.find(resource);
+        File file = Resources.find(resource);
+        if (file == null) {
+            throw new FileNotFoundException(resource.path());
+        }
+        return file;
     }
 
     @Override
     public long contentLength() throws IOException {
-        return Resources.find(resource).length();
+        return getFile().length();
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


src/main/src/main/java/org/geoserver/config/SpringResourceAdaptor.java
```diff
@@ -74,22 +86,30 @@ public boolean isOpen() {
...
     @Override
     public URL getURL() throws IOException {
-        return Resources.find(resource).toURI().toURL();
+        return getFile().toURI().toURL();
     }
 
     @Override
     public URI getURI() throws IOException {
-        return Resources.find(resource).toURI();
+        return getFile().toURI();
```
jodygarnett, line 94: Okay I see what you did there; using getFile() to do the null check in a single location

src/main/src/main/java/org/geoserver/config/SpringResourceAdaptor.java
```diff
@@ -74,22 +86,30 @@ public boolean isOpen() {
...
+        if (file == null) {
+            throw new FileNotFoundException(resource.path());
+        }
+        return file;
     }
 
     @Override
     public long contentLength() throws IOException {
-        return Resources.find(resource).length();
+        return getFile().length();
```
jodygarnett, line 112: It is these kind of method where I want to check the javadocs; and see if we could return ``0`` or ``-1`` or something to be kinder to client code. Still this is a judgement call and you are the person doing the work so I trust your judgement :)

src/main/src/test/java/org/geoserver/config/SpringResourceAdaptorTest.java
```diff
@@ -0,0 +1,95 @@
+/* (c) 2015 Open Source Geospatial Foundation - all rights reserved
```
jodygarnett, line 1: Probably ``(c) 2022`` ?

jodygarnett (commented):
I like the direction this is heading

src/main/src/main/java/org/geoserver/config/SpringResourceAdaptor.java
dieterstueken: yes, I see. I was too fast to replace them all. I will check again.

dieterstueken:
I was unable to figure out the reason for those  _PMD violations_. How can I reproduce it locally?
Also, the repeated  _bean lookup error_ occurs randomly. 
Did not find the cause and it seems not to be related to my changes.

aaime:
Regarding PMD, see https://docs.geoserver.org/latest/en/developer/qa-guide/index.html (linked from the checklist too).

src/main/src/test/java/org/geoserver/config/GeoServerPropertyConfigurerTest.java
```diff
@@ -35,6 +37,17 @@ public void tearDown() throws Exception {
         ctx.close();
     }
 
+    @Test
+    public void testConfigFileCreated() throws IOException {
```
sikeoka, line 41: Is this testing anything relevant to the rest of the PR?

sikeoka:
> I was unable to figure out the reason for those _PMD violations_. How can I reproduce it locally? Also, the repeated _bean lookup error_ occurs randomly. Did not find the cause and it seems not to be related to my changes.

PMD Failure: org.geoserver.config.SpringResourceAdaptorTest:8 Rule:wildcards Priority:3 No Wildcard Imports.

jodygarnett:
> > I was unable to figure out the reason for those _PMD violations_. How can I reproduce it locally?
> 
> PMD Failure: org.geoserver.config.SpringResourceAdaptorTest:8 Rule:wildcards Priority:3 No Wildcard Imports. PMD Failure: org.geoserver.config.SpringResourceAdaptorTest:9 Rule:wildcards Priority:3 No Wildcard Imports.

Sorry I should of seen this earlier - I struggled with this also and added some examples on how to run PMD to the user guide: https://docs.geoserver.org/latest/en/developer/qa-guide/index.html


```
mvn verify -Ppmd
```

or
```
mvn initialize pmd:check -Ppmd
```

src/main/src/test/java/org/geoserver/config/GeoServerPropertyConfigurerTest.java
jodygarnett: Did we decided if this is needed here?

jodygarnett (approved):
There is one question about GeoServerPropertyConfigurerTest, but the change here seems good and has test coverage.

src/main/src/test/java/org/geoserver/config/GeoServerPropertyConfigurerTest.java
dieterstueken: Took a while to realize it again. If you roll back to b14f54e1 testConfigFileCreated you can see the test to snap:

Without throwing a `FileNotFoundException` [GeoServerPropertyConfigurer.loadProperties()](https://github.com/geoserver/geoserver/blob/main/src/main/src/main/java/org/geoserver/config/GeoServerPropertyConfigurer.java#L104) won't ever enter the catch block, since an empty file was silently created, which did not hurt anybody else before.

The (otherwise failed) creation takes place within [GeoServerPropertyConfigurerTest.setUp()](https://github.com/geoserver/geoserver/pull/6314/files#diff-6e0586e9277979b3436a1ee5aa890935b97522db04e285e655279ce2d4ee236fR31) outside of testConfigFileCreated() and is hard to locate, since the exception is caught. Here is a stacktrace of the hidden exeption:

<pre>
java.io.FileNotFoundException: foo.properties
	at org.geoserver.config.SpringResourceAdaptor.getInputStream(SpringResourceAdaptor.java:62)
	at org.springframework.core.io.support.EncodedResource.getInputStream(EncodedResource.java:159)
	at org.springframework.core.io.support.PropertiesLoaderUtils.fillProperties(PropertiesLoaderUtils.java:99)
	at org.springframework.core.io.support.PropertiesLoaderSupport.loadProperties(PropertiesLoaderSupport.java:182)
--->	at org.geoserver.config.GeoServerPropertyConfigurer.loadProperties(GeoServerPropertyConfigurer.java:103)
	at org.springframework.core.io.support.PropertiesLoaderSupport.mergeProperties(PropertiesLoaderSupport.java:163)
	at org.springframework.context.support.PropertySourcesPlaceholderConfigurer.postProcessBeanFactory(PropertySourcesPlaceholderConfigurer.java:144)
	at org.springframework.context.support.PostProcessorRegistrationDelegate.invokeBeanFactoryPostProcessors(PostProcessorRegistrationDelegate.java:291)
	at org.springframework.context.support.PostProcessorRegistrationDelegate.invokeBeanFactoryPostProcessors(PostProcessorRegistrationDelegate.java:167)
	at org.springframework.context.support.AbstractApplicationContext.invokeBeanFactoryPostProcessors(AbstractApplicationContext.java:707)
	at org.springframework.context.support.AbstractApplicationContext.refresh(AbstractApplicationContext.java:533)
	at org.springframework.context.support.ClassPathXmlApplicationContext.<init>(ClassPathXmlApplicationContext.java:202)
	at org.springframework.context.support.ClassPathXmlApplicationContext.<init>(ClassPathXmlApplicationContext.java:177)
	at org.springframework.context.support.ClassPathXmlApplicationContext.<init>(ClassPathXmlApplicationContext.java:163)
	at org.geoserver.config.GeoServerPropertyConfigurerTest.setUp(GeoServerPropertyConfigurerTest.java:31)
....
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:54)
</pre>

So, after introducing the FNF Exception, creation of the missing property file works now and the unit test verifies it. I think this was hidden bug and went wrong before.

aaime:
@jodygarnett are you happy to merge the request now? Dieter followed up with your question.

dieterstueken:
yes, please!
no back port necessary for me.

PR #6325, opened 2022-11-04
[GEOS-10731] GWC variable Parameterization does not work with geoserver-environmen…

turingtestfail:
[GEOS-10731](https://osgeo-org.atlassian.net/browse/GEOS-10731)

…t.properties due to the bean initialization order

missed adds

cleanup

added test

  

# Checklist

### DIFF
--- a/src/gwc/src/main/java/org/geoserver/gwc/GWC.java
+++ b/src/gwc/src/main/java/org/geoserver/gwc/GWC.java
@@ -191,6 +191,8 @@ public class GWC implements DisposableBean, InitializingBean, ApplicationContext
 
     private final GridSetBroker gridSetBroker;
 
+    private GWCSynchEnv gwcSynchEnv;
+
     private DiskQuotaMonitor monitor;
 
     private CatalogLayerEventListener catalogLayerEventListener;
@@ -233,6 +235,7 @@ public class GWC implements DisposableBean, InitializingBean, ApplicationContext
      * @param jdbcConfigurationStorage GeoServer integrator for GeoWebCache DiskQuota {@link
      *     JDBCConfiguration}
      * @param blobStoreAggregator GeoWebCache BlobStore Aggregator
+     * @param gwcSynchEnv GeoServer integrator for GeoWebCache environment synchronization
      */
     public GWC(
             final GWCConfigPersister gwcConfigPersister,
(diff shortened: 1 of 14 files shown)


src/gwc/src/main/java/org/geoserver/gwc/GWCSynchEnv.java
```diff
@@ -0,0 +1,85 @@
...
+import org.geowebcache.GeoWebCacheEnvironment;
+import org.springframework.beans.BeansException;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+
+/** This class is responsible for synchronizing the GeoServer environment for GWC. */
+public class GWCSynchEnv implements ApplicationContextAware {
+    private GeoWebCacheEnvironment gwcEnvironment;
+
+    GeoServerEnvironment gsEnvironment = GeoServerExtensions.bean(GeoServerEnvironment.class);
```
aaime, line 20: Why not inject this one with the constructor?

src/gwc/src/test/java/org/geoserver/gwc/GWCSynchEnvTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+        GWCSynchEnv gwcSynchEnv = new GWCSynchEnv();
+        gwcSynchEnv.setGsEnvironment(genv);
+
+        GeoWebCacheEnvironment gwcEnv = new GeoWebCacheEnvironment();
+        gwcSynchEnv.setGwcEnvironment(gwcEnv);
+
+        gwcSynchEnv.setForceSync(
+                true); // force sync because system setting for allowing parametrization keeps
+        // getting reset
```
aaime, line 38: Play ball with the formatter

```suggestion
        // force sync because system setting for allowing parametrization keeps getting reset
        gwcSynchEnv.setForceSync(true); 
```

src/gwc/src/main/java/org/geoserver/gwc/GWC.java
```diff
@@ -316,6 +330,18 @@ public static void set(GWC instance) {
...
     }
 
+    /**
+     * Only to aid in unit testing for the places where a mock GWC mediator and GeoServer
+     * environment synchhronizer are needed
+     *
+     * @param instance the GWC mediator instance
+     * @param gwcSynchEnv the GeoServer environment synchronizer instance
+     */
+    public static void set(GWC instance, GWCSynchEnv gwcSynchEnv) {
```
aaime, line 340: Thinking out loud, is the other ``set(GWC)`` method still getting any usage? (I'm asking because I see tests using it being switched to this new method).

aaime (changes requested):
Looks good, see few minor comments

src/gwc/src/main/java/org/geoserver/gwc/GWCSynchEnv.java
turingtestfail: Fixed

src/gwc/src/main/java/org/geoserver/gwc/GWC.java
turingtestfail: It was in other places, but cleaner to switch to two argument form everywhere.  Fixed.

aaime (approved):

PR #6332, opened 2022-11-08
GEOS-10738 Better logging for OIDC

davidblasby:
[GEOS-10738](https://osgeo-org.atlassian.net/browse/GEOS-10738)

This adds some logging for troubleshooting OIDC issues.

Adds button on UI to turn on the logging.  If this is turned on, then more logging will take place (turn on VERBOSE logging).

This attempts to not log too sensitive information - 
+ CODE is logged AFTER its been used (its single-use)
+ JWT ID/ACCESS tokens only log the Claims (not the header/signature)
+ non-JWT tokens are not logged (that would be insecure)
+ some personal information will likely be logged (i.e. token claims include email/name/etc...)

Do not turn this on in a production environment.

# Checklist

- [n/a] New unit tests have been added covering the changes.
- [x Bug fixes and small new features are presented as a single commit.

### DIFF
--- a/src/community/security/oauth2-openid-connect/oauth2-openid-connect-core/src/main/java/org/geoserver/security/oauth2/OpenIdConnectAuthenticationFilter.java
+++ b/src/community/security/oauth2-openid-connect/oauth2-openid-connect-core/src/main/java/org/geoserver/security/oauth2/OpenIdConnectAuthenticationFilter.java
@@ -18,6 +18,7 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.logging.Level;
+import java.util.logging.Logger;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
@@ -30,6 +31,7 @@ import org.geoserver.security.impl.GeoServerRole;
 import org.geoserver.security.impl.RoleCalculator;
 import org.geoserver.security.oauth2.bearer.TokenValidator;
 import org.geoserver.security.oauth2.services.OpenIdConnectTokenServices;
+import org.geotools.util.logging.Logging;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.jwt.Jwt;
 import org.springframework.security.jwt.JwtHelper;
@@ -42,6 +44,8 @@ import org.springframework.security.oauth2.provider.token.RemoteTokenServices;
 
 /** Authenticate using OpenID Connect. */
 public class OpenIdConnectAuthenticationFilter extends GeoServerOAuthAuthenticationFilter {
+
+    private static final Logger LOGGER = Logging.getLogger(OpenIdConnectAuthenticationFilter.class);
 
     static final String ID_TOKEN_VALUE = "OpenIdConnect-IdTokenValue";
     TokenValidator bearerTokenValidator;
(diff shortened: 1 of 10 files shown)


src/community/security/oauth2-openid-connect/oauth2-openid-connect-core/src/main/java/org/geoserver/security/oauth2/OpenIdConnectAuthenticationFilter.java
```diff
@@ -119,38 +123,43 @@ protected Collection<GeoServerRole> getRoles(HttpServletRequest request, String
...
         }
-        return super.getRoles(request, principal);
+        if (filterConfig.isAllowUnSecureLogging()) {
+            if (result == null) {
+                LOGGER.log(Level.FINE, "OIDC: roles returned null (unexpected)");
+            } else if (result.isEmpty()) {
+                LOGGER.log(Level.FINE, "OIDC: roles returned NO ROLES");
+            } else {
+                for (GeoServerRole role : result) {
+                    LOGGER.log(Level.FINE, "OIDC: Geoserver Roles: " + role.getAuthority());
```
jodygarnett, line 157: Is only one role returned here?

src/community/security/oauth2-openid-connect/oauth2-openid-connect-web/src/main/resources/GeoServerApplication.properties
```diff
@@ -89,6 +89,11 @@ OpenIdConnectAuthProviderPanel.sendClientSecretHelp.title=Send Client Secret in
 OpenIdConnectAuthProviderPanel.sendClientSecretHelp= Check this option if the OpenId server requires a confidential client \
   to send the client_secret in the token response (eg. ADFS and Azure AD).
 
+OpenIdConnectAuthProviderPanel.allowUnSecureLogging=Log Sensitive Information (do not use in production)
+OpenIdConnectAuthProviderPanel.allowUnSecureLoggingHelp.title=Log Sensitive Information
+OpenIdConnectAuthProviderPanel.allowUnSecureLoggingHelp=Check this option if you want to log more sensitive information \
+  (i.e. tokens).  This is useful when debugging an OIDC configuration that isn't working correctly.
```
jodygarnett, line 95: Appreciate the direction because the user manual is too far away.

You still may wish to update the user manual page as part of this pull request.

jodygarnett:
General feedback; this is a good change (and is better than commented out code).

When making a change to the configuration should document and update the rest api if needed. OIDC is not yet an extension but it is good to add difficulty to making it an extension later.

Having this be a configuration option, rather than an environmental variable flag is unusual for this kind of feature. Many similar security adjacent flags are controlled via system property or environmental flag. If I understand that approach requires less code change (faster to write) and gate keeps the functionality to system administrators who can control the environment the application is running in:

```
boolean = Boolean.valueOf(GeoServerExtensions.getProperty(LOG_ODIC_DETAILS);
```

Recommending waiting for input from @taba90

src/community/security/oauth2-openid-connect/oauth2-openid-connect-core/src/main/java/org/geoserver/security/oauth2/OpenIdConnectAuthenticationFilter.java
davidblasby: no - several.

davidblasby:
I'm not sure if the configuration should be exclusively in environment variables - especially when the rest of the configuration is not.  Changing environment variables is hard (and can be quite difficult in dockerized systems) - but its really easy to change on the GUI.

src/community/security/oauth2-openid-connect/oauth2-openid-connect-web/src/main/resources/GeoServerApplication.properties
davidblasby: I will wait for @taba90

PR #6333, opened 2022-11-09
[GEOS-10740] Introduce ServiceInfo.getType() for use with filtering and enablement

jodygarnett:
[GEOS-10740](https://osgeo-org.atlassian.net/browse/GEOS-10740)

Although only common with older data directories, or ones that have been modified by hand, ServiceInfo.getName() can be useful to modify and is reflected in the GetCapabilities document produced by web services.

This value being available for modication by users makes it unsitable for layer disabled service list as too many sections of the codebase assume these are simple ``WMS``, ``WFS``, ``WCS`` identifiers (even if they are not documented as such).

# Checklist

### DIFF
--- a/src/community/ogcapi/ogcapi-images/src/main/java/org/geoserver/ogcapi/images/ImagesServiceInfoImpl.java
+++ b/src/community/ogcapi/ogcapi-images/src/main/java/org/geoserver/ogcapi/images/ImagesServiceInfoImpl.java
@@ -12,7 +12,7 @@ public class ImagesServiceInfoImpl extends ServiceInfoImpl implements ImagesServ
 
     @Override
     public String getType() {
-        return "ImagesService";
+        return "Images";
     }
 
     @Override
--- a/src/community/ogcapi/ogcapi-images/src/main/java/org/geoserver/ogcapi/images/ImagesServiceInfoImpl.java
+++ b/src/community/ogcapi/ogcapi-images/src/main/java/org/geoserver/ogcapi/images/ImagesServiceInfoImpl.java
@@ -11,6 +11,11 @@ public class ImagesServiceInfoImpl extends ServiceInfoImpl implements ImagesServ
     int maxImages = 1000;
 
     @Override
+    public String getType() {
+        return "ImagesService";
+    }
+
+    @Override
     public int getMaxImages() {
         return maxImages;
     }
--- a/src/main/src/main/java/org/geoserver/config/impl/ServiceInfoImpl.java
+++ b/src/main/src/main/java/org/geoserver/config/impl/ServiceInfoImpl.java
@@ -95,7 +95,7 @@ public class ServiceInfoImpl implements ServiceInfo {
         } else {
             // this default, while incorrect, has the greatest chance of
             // success across data directories
-            return getName().toUpperCase();
+            return getName() != null ? getName().toUpperCase() : null;
         }
     }
 
--- a/src/main/src/main/java/org/geoserver/config/impl/ServiceInfoImpl.java
+++ b/src/main/src/main/java/org/geoserver/config/impl/ServiceInfoImpl.java
@@ -94,7 +94,7 @@ public class ServiceInfoImpl implements ServiceInfo {
             return simpleName.substring(0, truncate);
         } else {
             // this default, while incorrect, has the greatest chance of
-            // success acorss data directories
+            // success across data directories
             return getName().toUpperCase();
         }
     }
--- a/src/main/src/main/java/org/geoserver/config/impl/ServiceInfoImpl.java
+++ b/src/main/src/main/java/org/geoserver/config/impl/ServiceInfoImpl.java
@@ -80,6 +80,25 @@ public class ServiceInfoImpl implements ServiceInfo {
         this.id = id;
     }
 
+    /**
+     * Default implementation attempts to determine service type based on class naming convention.
+     * Subclasses are encouraged to override.
+     *
+     * @return service type based on class name, truncating ServiceInfo.
+     */
+    @Override
+    public String getType() {
+        String simpleName = getClass().getSimpleName();
+        int truncate = simpleName.indexOf("ServiceInfo");
+        if (truncate > 0) {
+            return simpleName.substring(0, truncate);
+        } else {
+            // this default, while incorrect, has the greatest chance of
+            // success acorss data directories
+            return getName().toUpperCase();
+        }
+    }
+
     @Override
     public WorkspaceInfo getWorkspace() {
         return workspace;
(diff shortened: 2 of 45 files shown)


jodygarnett:
I am trying for the brute force fixing of them problem rather than something tricky with default methods. I did create a ``ServiceInfoImpl.getType()`` implementation which tries to guess something useful based on class name so existing code will not break.

First cut added a constant TYPE to each service info but xstream overwrites this value with ``null`` leading to introduction of a method as we see here.

src/web/gwc/src/main/java/org/geoserver/gwc/web/GWCServiceDescriptionProvider.java
```diff
@@ -25,7 +25,7 @@
 public class GWCServiceDescriptionProvider extends ServiceDescriptionProvider {
 
     /** Service used to cross-link between service description and service link description. */
-    public static final String SERVICE = "gwc";
+    public static final String SERVICE = "wmts";
```
jodygarnett, line 28: Seeking consistency with ``GwcServiceProxy`` in use of ``wmts`` as service type.

The string ``WMTS`` is used by GWCResourceServiceVoter to avoid showing this as a service enablement.

jodygarnett:
Feedback:
- There should be constants for these strings; or even a ``CodeList``.
- The ``WMTS`` term is required by GWCResourceServiceVoter ... to remove it from user interface list (since the tiles tab already has a toggle for that). If the user interface needs to special case GWC then it should do so; or this ResourceServiceVoter should be a bit more clear that it is about offering the user control to disable (and does not itself control if the service is disabled).
- Actually checking if a tile cache is available for the layer is not done

src/web/core/src/main/java/org/geoserver/web/ServiceDescription.java
```diff
@@ -103,14 +103,14 @@ public ServiceDescription(
...
-            String service,
+            String serviceType,
             InternationalString title,
             InternationalString description,
             boolean available,
             boolean admin,
             String workspace,
             String layer) {
-        this.service = service.toLowerCase();
+        this.service = serviceType.toLowerCase();
```
davidblasby, line 113: I think (??) this might be a bit inconsistent with rest of the code (which seems to be based on uppercase).  I don't think this will be an issue since you use `.equalsIgnoreCase`...

src/web/core/src/main/java/org/geoserver/web/ServiceLinkDescription.java
```diff
@@ -35,18 +35,18 @@ public class ServiceLinkDescription implements Serializable, Comparable<ServiceL
...
     public ServiceLinkDescription(
-            String service,
+            String serviceType,
             Version version,
             String link,
             String workspace,
             String layer,
             String protocol) {
-        this.service = service.toLowerCase();
+        this.service = serviceType.toLowerCase();
```
davidblasby, line 49: I think (??) this might be a bit inconsistent with rest of the code (which seems to be based on uppercase).  I don't think this will be an issue since you use `.equalsIgnoreCase`...

src/web/core/src/main/java/org/geoserver/web/ServiceDescription.java
jodygarnett: Yeah I am considering changing ServiceDescription and ServiceLinkDescription to upper case just so there is less guess work.

It is good feedback I should do this; and now is the correct time to do so (when trying to make things consistent with service enablement controls).

src/web/core/src/main/java/org/geoserver/web/ServiceLinkDescription.java
jodygarnett: Thanks for your feedback David; can you think of a good location we could add string constants for these values?

jodygarnett:
@davidblasby I have updated based on your feedback; going to look at how to test codebase resistance to services having non-default names. I think it would be best to adjust the test configuration service names; this would provide minimal test case changes and help ensure there are no side effects.

src/community/ogcapi/ogcapi-images/src/main/java/org/geoserver/ogcapi/images/ImagesServiceInfoImpl.java
```diff
@@ -10,6 +10,11 @@ public class ImagesServiceInfoImpl extends ServiceInfoImpl implements ImagesServ
 
     int maxImages = 1000;
 
+    @Override
+    public String getType() {
+        return "ImagesService";
```
aaime, line 15:
```suggestion
        return "Images";
```

src/main/src/main/java/org/geoserver/config/impl/ServiceInfoImpl.java
```diff
@@ -80,6 +80,25 @@ public void setId(String id) {
...
+     */
+    @Override
+    public String getType() {
+        String simpleName = getClass().getSimpleName();
+        int truncate = simpleName.indexOf("ServiceInfo");
+        if (truncate > 0) {
+            return simpleName.substring(0, truncate);
+        } else {
+            // this default, while incorrect, has the greatest chance of
+            // success acorss data directories
```
aaime, line 97:
```suggestion
            // success across data directories
```

aaime (changes requested):
Overall looks ok (see minor comments below)... with maybe one surprise for me.
I was hoping to see the name of the services without the uppercasing in the welcome page, and yet, see below:

image

I'm guessing the GUI is not populated using the service type (says "IMAGES" not "ImagesService" and yet the current PR has the second string as the type).... wondering, should it be using the type there?

jodygarnett:
> Overall looks ok (see minor comments below)... with maybe one surprise for me. I was hoping to see the name of the services without the uppercasing in the welcome page, and yet, see below:
> 
> image
> 
> I'm guessing the GUI is not populated using the service type (says "IMAGES" not "ImagesService" and yet the current PR has the second string as the type).... wondering, should it be using the type there?

I was also not expecting that; let me look. I think we are both agreed we want to see ``WMS``, ``Features`` Images`` (like I have documented in the javadocs).

src/main/src/test/java/org/geoserver/config/util/XStreamPersisterTest.java
```diff
@@ -208,6 +208,11 @@ static class MyServiceInfo extends ServiceInfoImpl {
 
         String foo;
 
+        @Override
+        public String getType() {
+            return "My";
```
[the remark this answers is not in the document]
jodygarnett: Yes, just so: https://disney.fandom.com/wiki/Seagulls

jodygarnett:
Andrea just as an aside - you are getting  much different visual to me.

image

Is this due to your wicket debug setting (which I cannot remember the details of)

jodygarnett:
Adjusted Service Capabilities to show service type as is: 
image

Comparison with service description skip list is still case insensitive.

PR #6334, opened 2022-11-09
[GEOS-10741] Remove deprecated YUI usage

taba90:
[GEOS-10741](https://osgeo-org.atlassian.net/browse/GEOS-10741)

This pr:
- remove wicket date time dependencies
- provides new components where needed namely: for datepicker in monitor and metadata modules; for Date Label in wicket core (System status).

Below a video showing the new datepicker.

https://user-images.githubusercontent.com/46672505/200799679-22e0eb73-87bf-435e-83b2-c1192d4979ad.mp4

  

# Checklist


src/web/core/src/main/java/org/geoserver/web/wicket/DatePicker.js
```diff
@@ -0,0 +1,116 @@
+// below some variable and functions to map date format tokens from java to moment js eg. d -> D for day of month.
```
aaime, line 1: This is something you added ? Should have the GeoServer license header then.

aaime (changes requested):
Overall looks good, just noting that we often put bits of javascript coming from other libraries in a dedicated "js" subfolder:

https://github.com/geoserver/geoserver/tree/main/src/web/core/src/main/java/org/geoserver/web/wicket/js

At the same time, bridge javascript bits that we write ourselves are on the side of the Java class using them, instead.

src/web/core/src/main/java/org/geoserver/web/wicket/DatePicker.js
taba90: fixed

taba90:
@aaime thanks for the review. Feedback applied.

PR #6335, opened 2022-11-09
[GEOS-10743] read missing style files

dieterstueken:
[GEOS-10743](https://osgeo-org.atlassian.net/browse/GEOS-10743)

[GEOS-10743](https://osgeo-org.atlassian.net/browse/GEOS-10743)

[ResourcePool.readStyle](https://github.com/geoserver/geoserver/blob/main/src/main/src/main/java/org/geoserver/catalog/ResourcePool.java#L2165) calls `Resource.in()` without any checking if the file exists.

The decision what to do, if the resource does not exist should be done at higher level. Unfortunately, it is directly turned into a character reader (even a BufferedReader) already. I think this not a good idea to do that at this level. For instance, the encoding is unclear at this point especially for SLD/XML files (same for writing). But I think this is again a refactoring going far beyond this issue.

To keep changes minimal and defensive, I simply returned an empty reader, which is compatible to the previous behavior, without opening (and thus creating) the file.

In general: ResourcePool seems to be a fat [god class](https://refactoring.guru/smells/large-class) which needs some slimming.
  
# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
+++ b/src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
@@ -2167,6 +2167,11 @@ public class ResourcePool {
         if (styleResource == null) {
             throw new IOException("No such resource: " + style.getFilename());
         }
+
+        // return an empty content if the file is missing.
+        if (styleResource.getType() == Resource.Type.UNDEFINED)
+            return new BufferedReader(new org.apache.commons.io.input.NullReader());
+
         return new BufferedReader(new InputStreamReader(styleResource.in()));
     }
 
--- a/src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
+++ b/src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
@@ -11,6 +11,7 @@ import java.io.BufferedWriter;
 import java.io.Closeable;
 import java.io.File;
 import java.io.FileInputStream;
+import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
@@ -2164,13 +2165,9 @@ public class ResourcePool {
      */
     public BufferedReader readStyle(StyleInfo style) throws IOException {
         Resource styleResource = dataDir().style(style);
-        if (styleResource == null) {
-            throw new IOException("No such resource: " + style.getFilename());
+        if (styleResource.getType() != Resource.Type.RESOURCE) {
+            throw new FileNotFoundException("No such resource: " + style.getFilename());
         }
-
-        // return an empty content if the file is missing.
-        if (styleResource.getType() == Resource.Type.UNDEFINED)
-            return new BufferedReader(new org.apache.commons.io.input.NullReader());
 
         return new BufferedReader(new InputStreamReader(styleResource.in()));
     }
--- a/src/main/src/test/java/org/geoserver/catalog/ResourcePoolTest.java
+++ b/src/main/src/test/java/org/geoserver/catalog/ResourcePoolTest.java
@@ -28,6 +28,7 @@ import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import java.awt.image.RenderedImage;
+import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
@@ -709,6 +710,25 @@ public class ResourcePoolTest extends GeoServerSystemTestSupport {
             // writeStyleFile throws an IOException
         }
         Assert.assertFalse("foo.sld should not exist on disk after a failure.", sldFile.exists());
+    }
+
+    /**
+     * Since GEOS-10743 readStyle() must throw a FileNotFoundException if the style file does not
+     * exist. (Before an empty file was returned.)
+     *
+     * @throws IOException
+     */
+    @Test(expected = FileNotFoundException.class)
+    public void testMissingStyleThrowsException() throws IOException {
+        Catalog catalog = getCatalog();
+        StyleInfo missing = catalog.getFactory().createStyle();
+        missing.setName("missing");
+        missing.setFilename("missing.sld");
+
+        ResourcePool pool = new ResourcePool(catalog);
+        try (BufferedReader reader = pool.readStyle(missing)) {
+            fail("FileNotFoundException expected for missing style files");
+        }
     }
 
     @Test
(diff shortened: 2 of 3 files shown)


src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -2167,6 +2167,11 @@ public BufferedReader readStyle(StyleInfo style) throws IOException {
         if (styleResource == null) {
```
dieterstueken, line 2167: This did not work anyway since: `assert r != null;` [here](https://github.com/geoserver/geoserver/blob/c3fc435088348df4096ed4dc00035c0994deaba0/src/main/src/main/java/org/geoserver/config/GeoServerDataDirectory.java#L670)

src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -2167,6 +2167,11 @@ public BufferedReader readStyle(StyleInfo style) throws IOException {
...
             throw new IOException("No such resource: " + style.getFilename());
         }
+
+        // return an empty content if the file is missing.
+        if (styleResource.getType() == Resource.Type.UNDEFINED)
+            return new BufferedReader(new org.apache.commons.io.input.NullReader());
```
jodygarnett, line 2173: Why not throw the same exception as above? The code path is already expecting an IOException if the resource is not available ...

dieterstueken: It seems this was the intention here. But since it never worked before, this was not taken into account. To do this right, it currently causes additional ugly catch/try brocks at different places. This is just a quick fix, to be able to finish the main idea: throw an exception for r.in() on missing files.

After that was finished, I can propose a more general solution here. It's the same as with [GEOS-10723](https://github.com/geoserver/geoserver/pull/6304#issuecomment-1307330343) and the functional approach. It's still on my list.

sikeoka: StyleAdminPanel and SLDNamedLayerRenameHelper are the only non-community module classes that call this method and they seem to already expect the exception and have the appropriate try/catch blocks.

sikeoka: If there are unit tests that depend on a bug then fixing the bug should require changing those tests.  Could you specifically list places that would require impractical changes?  The two places that I found already seem to be expecting the exception to be thrown and should probably be changed if you are fixing this in a way that doesn't throw the exception.

dieterstueken: To throw the exception is the right solution. But it definitely changes the behavior. This affects much more possible use cases and there are a lot of unit tests missing to verify all these cases. So, I don't want to fix this in the current PR to keep it small.

This PR has another topic: to avoid calling res.in(), and this way it preserves the original behavior (even if it seems wrong). This is currently important, since it blocks [GEOS-10705].

After all pending subtasks have been resolved, I can come back and review the Exception problem the way you proposed. But that will be another issue then.
This one is only an intermediate minimal fix.

dieterstueken:
Sorry I was too busy with my other projects. I just resolved the remaining wildcard problems of [GEOS-10724](https://github.com/geoserver/geoserver/pull/6314). Hope I can resolve this one very soon now.

dieterstueken:
I found, that the failed `StylesExtensionTest`,  I noticed,  had nothing to do with my changes.
So I was finally convinced to throw an `FileNotFoundException` for missing resource files.

However, I get an error on `MBStyleEditPageTest` now. 
I happens during setup of the page to edit some `mbstyle.json` style, which does not exist. Before it was returned as an empty bogus style. Now the page fails with:

> ...
>          at org.geoserver.community.mbstyle.web.MBStyleEditPageTest.setUp(MBStyleEditPageTest.java:45)
> Caused by: java.util.MissingResourceException: Unable to find property: 'styleNotFound' for component:  [class=org.geoserver.wms.web.data.StyleAdminPanel]. Locale: null, style: null
> 	at org.apache.wicket.Localizer.getString(Localizer.java:268)
> 	at org.apache.wicket.model.StringResourceModel.getString(StringResourceModel.java:445)
> 	at org.apache.wicket.model.StringResourceModel.getString(StringResourceModel.java:424)
> 	at org.geoserver.wms.web.data.StyleAdminPanel.initUI(StyleAdminPanel.java:236)
> 	at org.geoserver.wms.web.data.StyleAdminPanel.<init>(StyleAdminPanel.java:106)
> 	at org.geoserver.wms.web.data.AbstractStylePage$2.getPanel(AbstractStylePage.java:248)
> 	at org.geoserver.wms.web.data.AbstractStylePage$2.getPanel(AbstractStylePage.java:244)
> 	at org.apache.wicket.extensions.markup.html.tabs.PanelCachingTab.getPanel(PanelCachingTab.java:67)
> 	at org.apache.wicket.extensions.markup.html.tabs.TabbedPanel.setCurrentTab(TabbedPanel.java:422)
> 	at org.apache.wicket.extensions.markup.html.tabs.TabbedPanel.setSelectedTab(TabbedPanel.java:397)
> 	at org.apache.wicket.extensions.markup.html.tabs.TabbedPanel.onBeforeRender(TabbedPanel.java:268)
> 	at org.apache.wicket.Component.internalBeforeRender(Component.java:950)
> 	at org.apache.wicket.Component.beforeRender(Component.java:1018)
> 	at org.apache.wicket.MarkupContainer.onBeforeRenderChildren(MarkupContainer.java:1826)
> 	... 58 more

I think this discloses a separate problem with `MBStyleEditPageTest`. It tries to show an error page about the missing file, but the resource must be `AbstractStylePage.styleNotFound` instead.

No idea how to resolve this.

dieterstueken:
Seems to work now:

Throwing the exception now causes `StyleEditTabPanel` to load an `"styleNotFound"` error page. 
Choosing the right component here (`AbstractStylePage`) succeeds in locating the resource and the failed test vanished.

Seems this was an unrelated bug which was not caught by any other unit test before.

jodygarnett:
> Seems this was an unrelated bug which was not caught by any other unit test before.

Thanks for the fix; so how can we add a unit test to prove your fix? Ensuring changes have a test case is part of our check list / review process.

jodygarnett (changes requested):
A test case is needed, to verify IOException is thrown when reading a missing style file.

jodygarnett (approved):
Thank you very much for this improvement

jodygarnett:
Do you wish to see this improvement back ported?

dieterstueken:
Thank you for the review.

I see no need for a back port, since this is only one part of [GEOS-10705](https://osgeo-org.atlassian.net/browse/GEOS-10705).

If [PR#6314](https://github.com/geoserver/geoserver/pull/6314#pullrequestreview-1264857719) can also be finished we can close all three of them. 
Please note my [comment regarding GeoServerPropertyConfigurerTest](https://github.com/geoserver/geoserver/pull/6314#discussion_r1083565872).

After all three problems with IOExceptions get solved I can continue working on ResourceStore,

PR #6349, opened 2022-11-14
[GEOS-10729] Concurrent access on data access rules (authorization) can lead to loss of configured catalog mode, and lost rules. Follow up, more fixes

aaime:
[GEOS-10729](https://osgeo-org.atlassian.net/browse/GEOS-10729)

Follows up to https://github.com/geoserver/geoserver/pull/6318, with more concurrency fixes. I could not write a test, runnable in the build, that would successfully replicate the issue. I have a large integration one, that works against tens of thousands of rules, and tens of concurrent threads, that managed to make the issue occur a handful of times over the course of a 30 minutes run. After these fixes, I cannot get it to lose rules anymore (finger crossed).

For reference, here is the test class I've used (it's really a REST client talking to a live GeoServer):
https://gist.github.com/aaime/4bc441ece59a7cb31a0abf71110a5d51

  

# Checklist

### DIFF
--- a/src/restconfig/src/main/java/org/geoserver/rest/security/AbstractAclController.java
+++ b/src/restconfig/src/main/java/org/geoserver/rest/security/AbstractAclController.java
@@ -126,9 +126,11 @@ public abstract class AbstractAclController<
         }
 
         try {
-            ruleDAO.removeRule(rule);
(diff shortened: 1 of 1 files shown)


taba90 (approved):
Looks good to me.

PR #6361, opened 2022-11-17
Fix --mount option

raphaeljolivet:
In the docker documentation, the mount option was incorrectly -mount instead of --mount.

Also, the release name of docker container is incorrect :
`docker.osgeo.org/geoserver:2.22-SNAPSHOT` does not exist and should be replaced by `docker.osgeo.org/geoserver:2.22-RC` but I did not find where to change this.


jodygarnett (approved):

jodygarnett:
Thanks!

PR #6370, opened 2022-11-19
[GEOS-10638] Drop Java 8 support

aaime:
[GEOS-10638](https://osgeo-org.atlassian.net/browse/GEOS-10638)

Updated maven and github actions (docs have been updated in a previous PR)

# Checklist


aaime:
According to this ticket, the build should now fail on Windows:
https://osgeo-org.atlassian.net/browse/GEOS-10683

bradh (approved):
LGTM

bradh:
A couple of spotbugs errors breaking the build:

```
Error: 7,867 [ERROR] Read of unwritten field selectedRequestSRSs in org.geoserver.wcs.web.publish.WCSLayerConfig$1.onSubmit() [org.geoserver.wcs.web.publish.WCSLayerConfig$1] At WCSLayerConfig.java:[line 72] NP_UNWRITTEN_FIELD
Error: 7,867 [ERROR] Read of unwritten field selectedResponseSRSs in org.geoserver.wcs.web.publish.WCSLayerConfig$3.onSubmit() [org.geoserver.wcs.web.publish.WCSLayerConfig$3] At WCSLayerConfig.java:[line 102] NP_UNWRITTEN_FIELD
```

aaime:
> A couple of spotbugs errors breaking the build:
> 
> ```
> Error: 7,867 [ERROR] Read of unwritten field selectedRequestSRSs in org.geoserver.wcs.web.publish.WCSLayerConfig$1.onSubmit() [org.geoserver.wcs.web.publish.WCSLayerConfig$1] At WCSLayerConfig.java:[line 72] NP_UNWRITTEN_FIELD
> Error: 7,867 [ERROR] Read of unwritten field selectedResponseSRSs in org.geoserver.wcs.web.publish.WCSLayerConfig$3.onSubmit() [org.geoserver.wcs.web.publish.WCSLayerConfig$3] At WCSLayerConfig.java:[line 102] NP_UNWRITTEN_FIELD
> ```

Ah, it's a false positive, Wicket uses reflection to access the field. Added suppressions.

jodygarnett:
> According to this ticket, the build should now fail on Windows:
> https://osgeo-org.atlassian.net/browse/GEOS-10683

Congrats - it *does* fail on windows.  Would it be enough to not do the root path magic if on windows? The special treatment of absolute paths seems to be linux specific based on the comment....

```
        File file = Paths.toFile(null, path);
        if (!file.isAbsolute()) {
            // in linux, an absolute path might appear relative if the root slash has been removed.
            // This can also occur with the root path if java.io.tmpdir is relative.
```

aaime:
I've checked a bit and I can't quite believe what I'm reading... Windows has a "drive-specific current directory" so "c:test" and "c:\test" are not the same path and might be pointing to different places.

Unfortunately when we build the full File object in Paths.toFile, creating ``new File("c:")`` and ``new File("c:\")`` is not the same thing, the first one points to the drive-specific current directory (and it's a relative path), the second to the root of the drive (and it's an absolute path). 
Long story short, it seems we cannot ignore the path separator any longer when we are on Windows..

jodygarnett:
> ... Windows has a "drive-specific current directory" so "c:test" and "c:\test" are not the same path and might be pointing to different places.
> 
> ... Long story short, it seems we cannot ignore the path separator any longer when we are on Windows..

That sounds horrible! 

* Still we should be able to do something because these paths right here are our own Resource API which is really trying to be relative to the data directory, and are similar to file URLs??

Yes folks can put in an absolute URL, but we should be able to detect that somehow?

Checking URL approach can help: https://en.wikipedia.org/wiki/File_URI_scheme

* ``file://localhost/c:/path/to/the%20file.txt``
* ``file:///c:/path/to/the%20file.txt``

aside: Apparently ``///` is almost the same as ``//localhost/`:

So if we assume ``file://localhost/` for our Resource API absolute path we end up with ``c:/path/to/the%20file.txt``

From your note, and I am then going to assume from practice, folks have probable used windows ``\`` on occasion?

So to be forgiving ``C:\path\to\the file.txt``, but not ``C:path\to\the file.txt``

Agreed?

aaime:
``Jody I have to admit I've lost you... the reason why the issue is happening is that doing `new File("c:", "test")` results in a file with path `c:test`, in order to get an absolute path one has to use `new File("c:\", "test")` instead.

This change to `Paths.toFile,` works...  it's not pretty but it's a fix for a platform + java version-specific behavior so... not sure it can be made any prettier:

```
    public static File toFile(File base, String path) {
        for (String item : Paths.names(path)) {
            if (SystemUtils.isJavaVersionAtLeast(JavaVersion.JAVA_1_1)
                    && SystemUtils.IS_OS_WINDOWS
                    && base == null
                    && item.matches("\\w\\:")) {
                base = new File(item + "\\");
            } else {
                base = new File(base, item);
            }
        }
        return base;
    }
```

I've also considered applying it in `Paths.names`, but we'd miss the context that the base is empty, and thus we should try to consider the possibility the path is absoluted. Probably a better fix is to check the original path and see if it starts with "letter:/" or "letter:\\" to verify the original intention of the path was actually  to use an absolute path, and not the "drive-specific current directory".

By the way, the "drive-specific current directory" is not necessarily where GeoServer is running from, I could set the current directory in c:, then switch to d:, and start GeoServer using an absolute reference... the current directory in c: would be the one set before, even if GeoServer has been started from d: and may be in an entirely other place in c: (other than the drive specific current dir).

jodygarnett:
Sorry if I was not clear, I wanted to point out we are providing our own Paths API so we have a little more freedom.
I would like to confirm we are allowing ``letter:\\`` (or are we sticking to ``letter:/``).

Your code example seems good, should change to ``JavaqVersion.JAVA_11``).

I understand the drive-specific current directory. I am assuming that the environment that is running geoserver has defined a current directory for each available drive. And that it is the admin running geoserver that is responsible for the environment.

jodygarnett:
Reading more I think a lot of this is "just" a problem in the test case - it is testing the Files.ResourceAdaptor which works with absolute paths. The result shows a problem path ``Paths.convert(filePath)`` which indicates it handles absolute paths but is having trouble.

Some effort to hack around these problems are in ``FileWrapperResourceTheoryTest.getResource(path)`` in order to reuse the existing relative path tests; however this workaround have run into trouble on Windows Java 11.

jodygarnett:
Andrea after review I think you have the right interpretation; it would be better to be clear about absolute paths. The paths code eating the leading ``/`` (or converting ``C:\`` to ``C:``) is messing things up.

jodygarnett:
Andrea I am pushing up a fix to the test case which avoids tripping on absolute paths.

I also experimented with the following:
```java
    public static File toFile(File base, String path) {
        for (String item : Paths.names(path)) {
            if (base == null) {
                // absolute path
                if (SystemUtils.IS_OS_WINDOWS) {
                    if (item.matches("\\w\\:")) {
                        // absolute path, driver letter reference
                        base = new File(item + "\\");
                    } else {
                        base = new File(new File(System.getenv("SystemDrive")), item);
                    }
                } else {
                    if (item.equals("/")) {
                        base = new File(item);
                    } else {
                        base = new File("/" + item);
                    }
                }
            } else {
                // relative path
                base = new File(base, item);
            }
        }
        return base;
    }
```

Idea is to make absolute paths for our api explicit with either a ``/`` or a ``C:`` drive letter.

aaime:
Windows is green, yeah!
Do we need anything else in this PR?

Before merging we need to warn people on the devel list, or they will wonder why their usual Java 8 setup is not building anymore.

jodygarnett:
I think we can work backwards; change GeoServer to Java 8 first.

src/platform/src/test/java/org/geoserver/platform/resource/FileWrapperResourceTheoryTest.java
```diff
@@ -14,6 +14,15 @@
...
 import org.junit.rules.TemporaryFolder;
 
+/**
+ * Test handling of file wrapper (Files.ResourceAdaptor) which works with absolute paths internally.
+ *
+ * <p>This ResourceAdaptor is intended to allow GeoServer to work with absolute paths outside of the
+ * data direct. As a result stores an absolute file internally.
+ *
+ * <p>As ResourceTheoryTest covers a lot of relative paths {@link #getResource()} has been adjusted
+ * to be forgiving when absolute paths are used.
```
jodygarnett, line 24:
```suggestion
```

I adjusted the ResourceTheoryTest so broken absolute path would not be generated, as a result this comment no longer applies.

src/platform/src/test/java/org/geoserver/platform/resource/FileWrapperResourceTheoryTest.java
```diff
@@ -14,6 +14,13 @@
 import org.junit.experimental.theories.DataPoints;
 import org.junit.rules.TemporaryFolder;
 
+/**
+ * Test handling of file wrapper (Files.ResourceAdaptor) which works with absolute paths internally.
+ *
+ * <p>This ResourceAdaptor is intended to allow GeoServer to work with absolute paths outside of the
+ * data direct. As a result stores an absolute file internally.
+ *
```
jodygarnett, line 22:
```suggestion
```

PR #6382, opened 2022-11-24
[GEOS-10757] CITE: WMS <Style> has elements in wrong order (DTD validation)

davidblasby:
[GEOS-10757](https://osgeo-org.atlassian.net/browse/GEOS-10757)

The WMS DTD specifies that the order of tags inside the <Style> element must be Name, Title, ...
However, Geoserver was putting them in Title, Abstract, Name order.

<!ELEMENT Style ( Name, Title, Abstract?,
                  LegendURL*, StyleSheetURL?, StyleURL? ) >

This PR just puts the Name (correctly) first.

This is tested by CITE tests WMS 1.1.1.

I have added a test case that will validate the GetCapabilities against the DTD.

While doing this I noticed another problem - for nested Layers the nested Layer tag must be the last tag.

`Element Layer content does not follow the DTD, expecting (Name? , Title , Abstract? , KeywordList? , SRS* , LatLonBoundingBox? , BoundingBox* , Dimension* , Extent* , Attribution? , AuthorityURL* , Identifier* , MetadataURL* , DataURL* , FeatureListURL* , Style* , ScaleHint? , Layer*), got (Title Abstract KeywordList SRS LatLonBoundingBox BoundingBox Layer Style )`

In this, specific, example, the Style tag is AFTER the (nested layergroup) Layer tag.  It should be before.  Same for the ScaleHints.

  

# Checklist

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformer.java
+++ b/src/wms/src/main/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformer.java
@@ -1150,6 +1150,7 @@ public class GetCapabilitiesTransformer extends TransformerBase {
         }
 
         private void handleCommonStyleElements(StyleInfo defaultStyle) {
+            element("Name", defaultStyle.prefixedName());
             Style ftStyle;
             try {
                 ftStyle = defaultStyle.getStyle();
@@ -1157,7 +1158,6 @@ public class GetCapabilitiesTransformer extends TransformerBase {
             } catch (IOException e) {
                 throw new RuntimeException(e);
             }
-            element("Name", defaultStyle.prefixedName());
         }
 
         private void handleStyleTitleAndAbstract(String prefixedName, Style ftStyle) {
--- a/src/wms/src/main/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformer.java
+++ b/src/wms/src/main/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformer.java
@@ -1344,6 +1344,14 @@ public class GetCapabilitiesTransformer extends TransformerBase {
             }
             handleMetadataList(metadataLinks);
 
+            if (CapabilityUtil.encodeGroupDefaultStyle(wmsConfig, layerGroup))
+                handleLayerGroupDefaultStyle(layerName);
+
+            if (isSingleOrOpaque(layerGroup))
+                handleLayerGroupStyles(layerName, layerGroup.getLayerGroupStyles());
+
+            handleScaleHint(layerGroup);
+
             // handle children layers and groups
             if (!LayerGroupInfo.Mode.OPAQUE_CONTAINER.equals(layerGroup.getMode())
                     && !LayerGroupInfo.Mode.SINGLE.equals(layerGroup.getMode())) {
@@ -1358,13 +1366,6 @@ public class GetCapabilitiesTransformer extends TransformerBase {
                     }
                 }
             }
-            if (CapabilityUtil.encodeGroupDefaultStyle(wmsConfig, layerGroup))
-                handleLayerGroupDefaultStyle(layerName);
-
-            if (isSingleOrOpaque(layerGroup))
-                handleLayerGroupStyles(layerName, layerGroup.getLayerGroupStyles());
-
-            handleScaleHint(layerGroup);
 
             end("Layer");
         }
--- a/src/wms/src/test/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformerTest.java
+++ b/src/wms/src/test/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformerTest.java
@@ -466,7 +466,7 @@ public class GetCapabilitiesTransformerTest extends WMSTestSupport {
         // get the wms 1.1.1 DTD
         URL dtdURL =
                 GetCapabilitiesTransformer.class.getResource(
-                        "/schemas/wms/1.1.1/wms_ms_capabilities.dtd");
+                        "/schemas/wms/1.1.1/WMS_MS_Capabilities.dtd");
         String dtd = Resources.toString(dtdURL, StandardCharsets.UTF_8);
 
         try (InputStream dtdInputStream = new ByteArrayInputStream(dtd.getBytes())) {
--- a/src/wms/src/test/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformerTest.java
+++ b/src/wms/src/test/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformerTest.java
@@ -13,14 +13,25 @@ import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
 import com.google.common.base.Strings;
+import com.google.common.io.Resources;
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.StringWriter;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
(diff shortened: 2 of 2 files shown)


jodygarnett:
David this will need a test case; but perhaps rather than test for this specific outcome we could try ... validating against the DTD? It would catch this category of CITE failures.

src/wms/src/main/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformer.java
```diff
@@ -1150,14 +1150,14 @@ protected void handleLayer(final LayerInfo layer, boolean isRoot) throws IOExcep
         }
 
         private void handleCommonStyleElements(StyleInfo defaultStyle) {
+            element("Name", defaultStyle.prefixedName());
```
jodygarnett, line 1153: Moving the element name seems ... too easy :)

Is this same code used by WMS 1.1.1 and WMS 1.3.0? With our luck they have a different order or something terrible.

jodygarnett (changes requested):
We need a test case; either for this specific change to make sure it worked. Or it would be amazing to validate against the DTD (to shut down the entire category of CITE test failures)

src/wms/src/main/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformer.java
jodygarnett: aside: This is an excellent fix; especially if it let the project pass cite tests.

jodygarnett:
So it seems two kind of errors:

1) ``WARNING: Unable to load EntityResolverProvider. Entity resolution will be enabled: java.lang.ClassNotFoundException: org.geoserver.util.EntityResolverProvider``

This may make sense if DTD is configured and ``EntityResolverProvider`` is somehow not available. I wonder if you need to restore the setup of the SAX parser after your test (via try/finally).

2) Some cleanup errors when GeoServerExtensions is being used when program is already crashing?

src/wms/src/test/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformerTest.java
```diff
@@ -415,6 +431,89 @@ private void checkVendorSpecificCapsProviders(GetCapabilitiesTransformer tr) thr
...
+                    new EntityResolver() {
+                        @Override
+                        public InputSource resolveEntity(String publicId, String systemId)
+                                throws SAXException, IOException {
+                            if (systemId.endsWith("WMS_MS_Capabilities.dtd")) {
+                                return new InputSource(dtdInputStream);
+                            }
+                            return null;
+                        }
+                    });
```
jodygarnett, line 493: This looks legit; and I would expect both factory and builder to be isolated. 

Cannot tell why other threads are unhappy and failing.

jodygarnett:
David it runs locally for me on macOS with Java 11 so I cannot reproduce the failure locally.

The second issue ``BeanFactory not initialized or already closed - call 'refresh' before accessing beans via the ApplicationContext`` trying to be used when spring is unhappy seems to be a generic error message when spring is misconfigured. Stack exchange [lists many options](https://stackoverflow.com/questions/20114955/beanfactory-not-initialized-or-already-closed-call-refresh-before) for how this can occur

jodygarnett:
@aaime do you have any guidance on this one; I cannot see what is causing the failure?

src/wms/src/test/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformerTest.java
```diff
@@ -415,6 +431,89 @@ private void checkVendorSpecificCapsProviders(GetCapabilitiesTransformer tr) thr
         assertEquals(1, list.getLength());
     }
 
+    /**
+     * Gets a capabilities document as an XML string Also, adds: <!DOCTYPE WMT_MS_Capabilities
+     * SYSTEM "WMS_MS_Capabilities.dtd"> at the top of the document.
+     */
+    String getCapabilitiesXML() throws Exception {
+        // WMS wms = new WMS(getGeoServer());
+        WMS wms = (WMS) applicationContext.getBean("wms");
```
davidblasby, line 440: This is the only place I can see where spring would be unhappy...  I copied this from another test...

src/wms/src/test/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformerTest.java
```diff
@@ -415,6 +431,89 @@ private void checkVendorSpecificCapsProviders(GetCapabilitiesTransformer tr) thr
...
+    @Test
+    public void testValidatesAgainstDTD() throws Exception {
+        // get a capabilities document
+        String getCapXML = getCapabilitiesXML();
+
+        // get the wms 1.1.1 DTD
+        URL dtdURL =
+                GetCapabilitiesTransformer.class.getResource(
+                        "/schemas/wms/1.1.1/wms_ms_capabilities.dtd");
+        String dtd = Resources.toString(dtdURL, StandardCharsets.UTF_8);
```
davidblasby, line 470: I see this error in the logs:
2022-11-25T01:25:32.0940789Z 01:25:32,093 [ERROR] testValidatesAgainstDTD(org.geoserver.wms.capabilities.GetCapabilitiesTransformerTest)  Time elapsed: 0.836 s  <<< ERROR!
2022-11-25T01:25:32.0941352Z java.lang.NullPointerException
2022-11-25T01:25:32.0942044Z 	at org.geoserver.wms.capabilities.GetCapabilitiesTransformerTest.testValidatesAgainstDTD(GetCapabilitiesTransformerTest.java:470)

It works on Mac and Windows, but not on linux.

davidblasby:
> The second issue `BeanFactory not initialized or already closed - call 'refresh' before accessing beans via ...

Not sure if this is being caused by my changes???

davidblasby:
> So it seems two kind of errors:
> 
> 1. `WARNING: Unable to load EntityResolverProvider. Entity resolution will be enabled: java.lang.ClassNotFoundException: org.geoserver.util.EntityResolverProvider`

I don't think this is a problem with this PR - you'll see the issue with other PRs.

src/wms/src/test/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformerTest.java
davidblasby: I think this is the issue - I don't think the other issues you looked at are caused by this PR.  
Please see the linux output of other (recent) PR builds...

src/wms/src/test/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformerTest.java
```diff
@@ -415,6 +431,89 @@ private void checkVendorSpecificCapsProviders(GetCapabilitiesTransformer tr) thr
...
+
+    @Test
+    public void testValidatesAgainstDTD() throws Exception {
+        // get a capabilities document
+        String getCapXML = getCapabilitiesXML();
+
+        // get the wms 1.1.1 DTD
+        URL dtdURL =
+                GetCapabilitiesTransformer.class.getResource(
+                        "/schemas/wms/1.1.1/wms_ms_capabilities.dtd");
```
jodygarnett, line 469:
```suggestion
                        "/schemas/wms/1.1.1/WMS_MS_Capabilities.dtd");
```

Linux is case sensitive, mACos and wInDoWS are not.

src/wms/src/test/java/org/geoserver/wms/capabilities/GetCapabilitiesTransformerTest.java
jodygarnett: Yes; windows and macOS are case in-sensitive. Linux is the sensitive one....

jodygarnett:
@davidblasby I used GitHub suggestions to fix the ``WMS_MS_Capabilities.dtd`` reference; I think when you wake up this PR will build.

jodygarnett (approved):
Test case added; this looks good to me.

davidblasby:
@jodygarnett  - thanks for finding & fixing that issue.

PR #6392, opened 2022-11-29
[GEOS-10759] Welcome page autocomplete mode

jodygarnett:
[GEOS-10759](https://osgeo-org.atlassian.net/browse/GEOS-10759)

  

# Checklist

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.html
+++ b/src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.html
@@ -5,11 +5,11 @@
   <form wicket:id="form">
     <div class="select">
 			<img class="icon" src="#" wicket:id="workspace.icon" wicket:message="title:workspace"/>
-			<select wicket:id="workspace"/>
+			<input wicket:id="workspace" type="text" size="20" />
 	  </div>
     <div class="select">
 			<img class="icon" src="#" wicket:id="layer.icon" wicket:message="title:layer"/>
-		  <select wicket:id="layer"/>
+			<input wicket:id="layer" type="text" size="20" />
 	  </div>
     <div id="refresh">
 		  <a href="#" wicket:id="refresh">refresh</a>
--- a/src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.java
+++ b/src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.java
@@ -10,7 +10,9 @@ import static org.geoserver.catalog.Predicates.acceptAll;
 import com.google.common.base.Stopwatch;
 import java.text.NumberFormat;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
 import java.util.logging.Level;
@@ -20,10 +22,11 @@ import org.apache.wicket.Component;
 import org.apache.wicket.Localizer;
 import org.apache.wicket.ajax.AjaxRequestTarget;
 import org.apache.wicket.ajax.form.AjaxFormComponentUpdatingBehavior;
+import org.apache.wicket.extensions.ajax.markup.html.autocomplete.AutoCompleteTextField;
 import org.apache.wicket.markup.html.basic.Label;
 import org.apache.wicket.markup.html.form.Form;
-import org.apache.wicket.markup.html.form.IChoiceRenderer;
 import org.apache.wicket.markup.html.form.SubmitLink;
+import org.apache.wicket.markup.html.form.TextField;
 import org.apache.wicket.markup.html.image.Image;
 import org.apache.wicket.markup.html.link.BookmarkablePageLink;
 import org.apache.wicket.markup.html.list.ListItem;
@@ -49,26 +52,23 @@ import org.geoserver.catalog.PublishedInfo;
 import org.geoserver.catalog.StoreInfo;
 import org.geoserver.catalog.WorkspaceInfo;
 import org.geoserver.catalog.impl.DefaultCatalogFacade;
+import org.geoserver.catalog.util.CloseableIterator;
 import org.geoserver.config.ContactInfo;
 import org.geoserver.config.GeoServer;
 import org.geoserver.config.ServiceInfo;
 import org.geoserver.config.SettingsInfo;
+import org.geoserver.platform.GeoServerExtensions;
 import org.geoserver.security.AdminRequest;
 import org.geoserver.util.InternationalStringUtils;
 import org.geoserver.web.data.layer.LayerPage;
 import org.geoserver.web.data.layer.NewLayerPage;
-import org.geoserver.web.data.layer.PublishedChoiceRenderer;
-import org.geoserver.web.data.layer.PublishedInfosModel;
 import org.geoserver.web.data.layergroup.LayerGroupEditPage;
 import org.geoserver.web.data.layergroup.LayerGroupPage;
 import org.geoserver.web.data.store.NewDataPage;
 import org.geoserver.web.data.store.StorePage;
-import org.geoserver.web.data.workspace.WorkspaceChoiceNameRenderer;
 import org.geoserver.web.data.workspace.WorkspaceNewPage;
 import org.geoserver.web.data.workspace.WorkspacePage;
-import org.geoserver.web.data.workspace.WorkspacesModel;
 import org.geoserver.web.spring.security.GeoServerSession;
-import org.geoserver.web.wicket.Select2DropDownChoice;
 import org.geotools.feature.NameImpl;
 import org.opengis.filter.Filter;
 import org.opengis.util.InternationalString;
@@ -97,6 +97,44 @@ import org.opengis.util.InternationalString;
 public class GeoServerHomePage extends GeoServerBasePage implements GeoServerUnlockablePage {
 
     /**
+     * System property used to externally define {@link #SelectionMode}.
+     *
+     * <p>If provided this setting will override any configuration option.
+     */
+    public static String SELECTION_MODE = "GeoServerHomePage.selectionMode";
+
+    enum SelectionMode {
+        /** Workspace and layer auto-complete suitable for small catalogues. */
+        AUTOCOMPLETE,
+        /**
+         * Layer autocomplete is only available when workspace prefix provided. Suitable for large
+         * catalogues with many workspaces
+         */
+        PREFIX,
+        /**
+         * Workspace autocomplete only, disabling layer autocomplete.
+         *
+         * <p>Suitable for large/slow catalogues.
+         */
+        WORKSPACE,
+        /** Disable autocomplete, use plain text-field for workspace and layer selection */
+        DISABLE;
+    }
+
+    SelectionMode selectionMode = SelectionMode.WORKSPACE;
+
+    {
+        try {
+            String mode = GeoServerExtensions.getProperty(SELECTION_MODE);
+            if (!Strings.isEmpty(mode)) {
+                selectionMode = SelectionMode.valueOf(mode.toUpperCase());
+            }
+        } catch (IllegalArgumentException ignore) {
+            LOGGER.fine("GeoServer home page selection mode not set:" + ignore);
+        }
+    }
+
+    /**
      * Optional workspace context for displayed web services, or {@code null} for global services.
      *
      * <p>This field matches the page parameter and is used to populate the raw text input of the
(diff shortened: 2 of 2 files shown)


jodygarnett:
@aaime here is something for you to consider; personally I hate the user experience

jodygarnett:
Instructions for testing: ``-DGeoServerHomePage.selectionMode=AUTOCOMPELTE``, possible values:

* AUTOCOMPLETE: Workspace and layer auto-complete suitable for small catalogues
* PREFIX: Layer autocomplete is only available when workspace prefix provided. Suitable for large catalogues with many workspaces
* WORKSPACE: Workspace autocomplete only, disabling layer autocomplete. Suitable for large/slow catalogues.
* DISABLE: Disable autocomplete, use plain text-field for workspace and layer selection

This is good enough for @aaime to try performance testing; the updates are not perfect (selecting a layer clears the workspace selection shown etc...).

I do not wish to see this merged; but it is the best option if we cannot find another way.

aaime:
Just noticed the commit message... LOL!! :rofl:

aaime:
So I've tested my local 50k layers data directory... the autocompletes are working quite reasonably IMHO. I made the following changes to make it more user friendly:

```
diff --git a/src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.java b/src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.java
index 399794eed8..bf5751d1df 100644
--- a/src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.java
+++ b/src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.java
@@ -22,6 +22,7 @@
 import org.apache.wicket.Localizer;
 import org.apache.wicket.ajax.AjaxRequestTarget;
 import org.apache.wicket.ajax.form.AjaxFormComponentUpdatingBehavior;
+import org.apache.wicket.extensions.ajax.markup.html.autocomplete.AutoCompleteSettings;
 import org.apache.wicket.extensions.ajax.markup.html.autocomplete.AutoCompleteTextField;
 import org.apache.wicket.markup.html.basic.Label;
 import org.apache.wicket.markup.html.form.Form;
@@ -43,6 +44,7 @@
 import org.apache.wicket.request.resource.PackageResourceReference;
 import org.apache.wicket.util.string.StringValue;
 import org.apache.wicket.util.string.Strings;
+import org.checkerframework.checker.units.qual.A;
 import org.geoserver.catalog.Catalog;
 import org.geoserver.catalog.LayerGroupInfo;
 import org.geoserver.catalog.LayerInfo;
@@ -375,12 +377,15 @@ public void onSubmit() {
         form.add(refresh);
         form.setDefaultButton(refresh);
 
+        AutoCompleteSettings settings = new AutoCompleteSettings();
+        settings.setShowListOnEmptyInput(true);
+        settings.setShowListOnFocusGain(true);
         this.workspaceField =
                 new AutoCompleteTextField<String>(
-                        "workspace", new PropertyModel<>(this, "workspace")) {
+                        "workspace", new PropertyModel<>(this, "workspace"), settings) {
                     @Override
                     protected Iterator<String> getChoices(String input) {
-                        if (Strings.isEmpty(input) || selectionMode == SelectionMode.DISABLE) {
+                        if (selectionMode == SelectionMode.DISABLE) {
                             return Collections.emptyIterator();
                         }
                         Filter filter = Predicates.contains("name", input);
@@ -389,7 +394,7 @@ protected Iterator<String> getChoices(String input) {
                         clearAdminRequestIfNotAdmin();
                         long TIME_LIMIT = System.currentTimeMillis() + 3000;
                         try (CloseableIterator<WorkspaceInfo> search =
-                                getCatalog().list(WorkspaceInfo.class, filter, 0, 10, null)) {
+                                getCatalog().list(WorkspaceInfo.class, filter, 0, 20, null)) {
                             while (System.currentTimeMillis() < TIME_LIMIT && search.hasNext()) {
                                 suggestions.add(search.next().getName());
                             }
@@ -410,12 +415,14 @@ protected void onUpdate(AjaxRequestTarget target) {
                     }
                 });
         form.add(workspaceField);
+        
         this.layerField =
-                new AutoCompleteTextField<String>("layer", new PropertyModel<>(this, "layer")) {
+                new AutoCompleteTextField<String>("layer", new PropertyModel<>(this, "layer"), String.class,
+                        settings) {
                     @Override
                     protected Iterator<String> getChoices(String input) {
-                        if (Strings.isEmpty(input)
-                                || (selectionMode == SelectionMode.DISABLE)
+                        if (
+                                (selectionMode == SelectionMode.DISABLE)
                                 || (selectionMode == SelectionMode.PREFIX
                                         && workspaceInfo == null)) {
                             return Collections.emptyIterator();
@@ -426,7 +433,7 @@ protected Iterator<String> getChoices(String input) {
                         clearAdminRequestIfNotAdmin();
                         long TWO_SECONDS = System.currentTimeMillis() + 1000;
                         try (CloseableIterator<PublishedInfo> search =
-                                getCatalog().list(PublishedInfo.class, filter, 0, 10, null)) {
+                                getCatalog().list(PublishedInfo.class, filter, 0, 20, null)) {
                             while (System.currentTimeMillis() < TWO_SECONDS && search.hasNext()) {
                                 if (workspaceInfo == null) {
                                     suggestions.add(search.next().prefixedName());
```

The dropdown unfortunately does not have a scroll, so having more than 20 items is not particuarly useful (quickly goes beyong the screen height).. that said, it might be possible to fix it, the compoenent accepts custom CSS.

I'll see if I can test on a more realistic scenario, too.

jodygarnett:
Okay so now I need to figure out how to do this for real:

1. AutoComplete with show list on focus gain and empty input
2. Use the header panel thing to swap between panels
3. Preserve use of WorkspaceInfo and PublishedInfo as parameters (rather than just strings like in this prototype).
4. Make global settings for auto complete options with number of suggestions and timeout

aaime:
Point number one is covered in the diff I've shared.

aaime:
So I've tested against the "difficult" data directory and .... the PR as is, is not helping. The data dir has 20k layers, which is not outrageous, but that is paired with 50k security rules in ``layers.properties``.

The moment we start scrolling over layers, the DefaultResourceAccessManager builds a very complex filter in the ``getSecurityFilter`` method, which is then evaluated in memory.... so basically multiply 20k layers by a filter that has 20k rules (it comes up with a filter based on the name of the layer, for each layer)... and it clearly becomes too big.
Ah, for reference, there are 2 workspaces, so any attempt at workspace filtering is going to be ineffective.

Can we make the filter better, say, avoid filtering at all if all layers are visible? Sure, we could, but we'll still be stuck in situations where the number of excluded layers is half/half (so I could have a filtering rule that is in the order of 10k conditions).

So where is the time spent now? Counting the layers mainly (takes several minutes to get a result)... The count needs some sort of limit too, and if it's more than that, we'll need to just say "hey, too many".... basically move the loading of the lists before we need the counts, and get the counts from the lists themselves, plus some info on whether we went beyond limits (took too much time, or we went beyond the maximum number of layers to be loaded).

I've done some quick tests, and just removing the count makes the home page show up quickly (rather than having a blank page for minutes)... the autocomplete on layers is still very slow though, well over 30 seconds... I'm guessing that the issue is that it takes a lot of time to get the first result because due to the rules evaluation, it might take a lot of time to just go through layers that are getting skipped (by default the system returns 190 layers out of the 20k available when using anonymous access).

The current DefaultResourceAccessManager rule building is geared towards JDBCConfig, it allows encoding it down in SQL, so lots of logic combos. I guess we could turn it into a more efficient "in" filter with a function, but that would also make JDBCConfig regress unless it learns how to deal with an "in" filter and write it down in SQL (just thinking out loud).

As another option... well... what I said at the beginning: have a way to just turn this new functionality completely.

jodygarnett:
Okay Andrea thanks for the extensive testing; that is what I needed to know for planning. To be honest I added the count because it was already present on the page (for administrators) and David asked for it (wanting to know what was going on when looking at a workspace for which no layers were available).

I am not comfortable turning off the functionality completely - as I want people to be able to share book markable pages to available virtual services.  But we can turn off the auto-complete and so on ...  I assume looking up a *single* workspace or layer would not have such a security penalty. 

Questions:
1. If the security check is so slow - was the GeoServer 2.21.x welcome page slow? I am guessing not as it only obtained a count when used as an administrator....
2. If the security check is so slow - is the GetCapabilities documents also very slow? I am guessing Catalog Mode is designed to help here.... Is respecting catalog mode useful? 
3. Is the layer preview page also broken for this large catalog with long security checks? I am using the same query structure ...  I am going to guess that it is.
4. Is catalog mode going to be enough control? On the surface it is: default - works as written; challenge - is fine to list everything; mixed - this is where a no auto complete should be available

Okay, so plan of attack:
1. Make it so we can substitute the workspace / layer selection between at least two strategies. The default select2choice and a plain text field (that does a single name based lookup, possibly respecting catalog mode).
2. Remove the count (or make it depend on the above

aaime:
> To be honest I added the count because it was already present on the page (for administrators) and David asked for it (wanting to know what was going on when looking at a workspace for which no layers were available).

Administrators have the count at a cheap price, [full admins can see everything](https://github.com/geoserver/geoserver/blob/6d68b80fca96ceac227f8a814b1501652172f0c6/src/main/src/main/java/org/geoserver/security/SecureCatalogImpl.java#L1711), so that was not a problem. Exposing it to all users makes security restrictions kick in and... well, you see where we ended up.
 
> I am not comfortable turning off the functionality completely - as I want people to be able to share book markable pages to available virtual services. But we can turn off the auto-complete and so on ... I assume looking up a _single_ workspace or layer would not have such a security penalty.

You cannot assume that in general. In the data directory I'm playing with, there are only two workspaces, one contains the bulk of the layers. Very large installations tend to have everything thrown into a single workspace and managed automatically.

> Questions: 0. What catalog mode is your example using?

HIDE. Mind it's just an example. Users have been adopting GeoServer over the span of 20 years. It's safe to assume their tried anything that the software even remotely allowed to do. Look here for some big servers, many of which are GeoServers (GeoServer somehow invites people to accumulate many layers, maybe because it's easy, maybe because there is a REST API):
https://www.geoseer.net/stats/#10+WFSes+with+the+most+Datasets

And... well apart from that one WFS with 200k layers, the others are somehow still in a "sane" position (would start up in a few minutes, caps return in a few tens of seconds). It's however the norm, for large installations, that the caps document is dodged entirely and the "one client" they have uses some other access mode (separate catalog, naming conventions, something like that).

> 
> 1. If the security check is so slow - was the GeoServer 2.21.x welcome page slow? I am guessing not as it only obtained a count when used as an administrator....

It was responding instantly indeed.

> 2. If the security check is so slow - is the GetCapabilities documents also very slow? I am guessing Catalog Mode is designed to help here.... Is respecting catalog mode useful?

See above, caps documents are irrelevant in these large installations. OGC never designed them to scale up, users just stopped using them, GeoServer is based on standards but users use it as they see fit... not necessarily following standards. That happens for other aspects as well, e.g. WMS dimensions are useless past a certain size, they are either too big (list mode) or incorrect (assumption of regular cadence for data, when everything realistic has holes).

> 3. Is the layer preview page also broken for this large catalog with long security checks? I am using the same query structure ...  I am going to guess that it is.

Yep, quite likely. Again, these large installations do not care. The GUI is often not reachable, the "one client" they provide is the only access point.

> 4. Is catalog mode going to be enough control? On the surface it is: default - works as written; challenge - is fine to list everything; mixed - this is where a no auto-complete should be available

Catalog mode is not going to be helpful I'm afraid, you cannot assume it has a relation to size. The mode is chosen based on the security needs people have (do I have a common client that needs a challenge? HIDE is not for me, CHALLENGE is the only way to go).

> Okay, so plan of attack:
> 
> 1. Make it so we can substitute the workspace / layer selection between at least two strategies. The default select2choice and a plain text field (that does a single name based lookup, possibly respecting catalog mode).
> 2. Remove the count (or make it depend on the above

Makes sense I guess... but how do you know in which case you are? There should be some flag to force behavior, automation is nice but it won't work in all cases (see my previous comment, even trying to scroll and limit the time one has to wait for the first result... which can take a lot of time if most layers are not accessible). Maybe the catalog scroll API should be extended to have a timeout?

aaime:
I'd say, superceded by https://github.com/geoserver/geoserver/pull/6539 ?
Could not get the approach in this one to work with the larger catalogs, but recovered as much as possible, including the brilliant idea of limiting the loading from the catalog in both time and size (the remaining hiccup is that even gathering the first result might be too long, which the new PR allows to avoid by using a system variable)

PR #6393, opened 2022-11-29
[GEOS-10760] GeoFence XML REST API broken: wrong element names

groldan:
[GEOS-10760](https://osgeo-org.atlassian.net/browse/GEOS-10760)

Fix a regression in encoding GeoFence JAXB message payloads introduced by [GEOS-10591](https://osgeo-org.atlassian.net/browse/GEOS-10591) at commit cf57b56edcfc22080fe73c3951b01b3408088a10 (`2.22.x` series), which adds the `jackson-dataformat-xml` dependency to `gs-ows`.

As a result, `WebMvcConfigurationSupport.addDefaultHttpMessageConverters()` will not add a Jaxb2RootElementHttpMessageConverter`, but a `MappingJackson2XmlHttpMessageConverter` instead, rendering the JAXB2 annotations in Geofence's REST object model useless.

This patch makes `org.geoserver.rest.RestConfiguration` extend `DelegatingWebMvcConfiguration` instead of WebMvcConfigurationSupport`, which is a subclass itself, to detect and delegate to all beansof type `WebMvcConfigurer`, allowing them to customize the configuration provided by `WebMvcConfigurationSupport`.

A `WebMvcConfigurer` (`GeofenceJaxbRestConfiguration`) is then contributed by the `geofence-server` module, to implement `configureMessageConverters()` making sure a `Jaxb2RootElementHttpMessageConverter` is provided with higher precedence than `MappingJackson2XmlHttpMessageConverter`.

This is safe, since the Jaxb converter will only engage for message payloads annotated with JAXB's `@XmlRootElement`.

  

# Checklist

### DIFF
--- a/src/extension/geofence/geofence-server/src/main/java/org/geoserver/geofence/server/rest/GeofenceJaxbRestConfiguration.java
+++ b/src/extension/geofence/geofence-server/src/main/java/org/geoserver/geofence/server/rest/GeofenceJaxbRestConfiguration.java
@@ -1,0 +1,46 @@
+/* (c) 2022 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.geofence.server.rest;
+
+import java.util.List;
+import java.util.logging.Logger;
+import javax.xml.bind.annotation.XmlRootElement;
+import org.geotools.util.logging.Logging;
+import org.springframework.http.converter.HttpMessageConverter;
+import org.springframework.http.converter.xml.Jaxb2RootElementHttpMessageConverter;
+import org.springframework.stereotype.Component;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurationSupport;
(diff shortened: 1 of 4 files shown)


groldan:
@aaime @taba90 note the change in `RestConfig`, extending the `DelegatingWebMvcConfiguration` subclass of `WebMvcConfigurationSupport` instead of `WebMvcConfigurationSupport` itself.

This is the cleanest way I found to allow decoupled modules to contribute to the rest configuration, after trying a thousand things.

I've noted the `ogc-api` `APIConfigurationSupport` class extends `RestConfig`, you may want to just implement `WebMvcConfigurer` instead? Unless it's completely unrelated like in a spring sub-context, which I don't know.

```
/**
 * Subclass of {@link WebMvcConfigurationSupport} adding support for dispatching {@link
 * DispatcherCallback#operationDispatched} events to callbacks
 */
// Not @Configuration on purpose, or it will interfere with the REST API.
// @Component is good enough to allow auto-wiring.
@Component
public class APIConfigurationSupport extends RestConfiguration {
...
```

taba90 (approved):
looks good to me

PR #6396, opened 2022-11-30
[GEOS-10783] CITE WFS 1.1 - unable to lock features

davidblasby:
[GEOS-10783](https://osgeo-org.atlassian.net/browse/GEOS-10783)

While running CITE tests, i was unable to lock any features.

JIRA: https://osgeo-org.atlassian.net/browse/GEOS-10783

Tracking this down, what was happening is that this was being called:

` fs = (SimpleFeatureSource) transformer.wrapFeatureSource(info, fs);`

This was returning a non-locking SimpleFeatureSource.  This was then wrapped in another non-lock SimpleFeatureSource.

This was because the Feature Info had a list of attributes in it.  I'm not sure if this is particular to the CITE configuration or if GS locking is completely broken for everyone.

Test cases have getAttributes() returning an empty list, while the actual catalog returns a full list of attributes.

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
+++ b/src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
@@ -1307,7 +1307,9 @@ public class ResourcePool {
         SimpleFeatureSource fs = dataStore.getFeatureSource(nativeName);
 
         // if feature type customization is there, apply it first
-        if (info.getAttributes() != null && !info.getAttributes().isEmpty()) {
+        if (info.getAttributes() != null
+                && !info.getAttributes().isEmpty()
+                && !info.getFeatureType().equals(fs.getSchema())) {
             fs = (SimpleFeatureSource) transformer.wrapFeatureSource(info, fs);
         }
         // then check name, and other customizations (e.g., projection policy)
--- a/src/main/src/test/java/org/geoserver/catalog/ResourcePoolTest.java
+++ b/src/main/src/test/java/org/geoserver/catalog/ResourcePoolTest.java
@@ -80,8 +80,10 @@ import org.geotools.coverage.grid.io.StructuredGridCoverage2DReader;
 import org.geotools.coverage.util.CoverageUtilities;
 import org.geotools.data.DataAccess;
 import org.geotools.data.DataStore;
+import org.geotools.data.FeatureSource;
 import org.geotools.data.Query;
 import org.geotools.data.simple.SimpleFeatureCollection;
+import org.geotools.data.simple.SimpleFeatureLocking;
 import org.geotools.data.simple.SimpleFeatureSource;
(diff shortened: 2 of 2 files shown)


src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -1307,7 +1307,9 @@ public void clear(CoverageInfo info) {
         SimpleFeatureSource fs = dataStore.getFeatureSource(nativeName);
 
         // if feature type customization is there, apply it first
-        if (info.getAttributes() != null && !info.getAttributes().isEmpty()) {
+        if (info.getAttributes() != null
+                && !info.getAttributes().isEmpty()
+                && !info.getFeatureType().equals(fs.getSchema())) {
```
jodygarnett, line 1312: This seems a fair check if a transform is required; this code must of been added since the last time cite tests were run

aaime: Indeed, it happened later. Here we have "portion a fix", the ability to lock customized feature types remains, it just so happens that the list of attributes in CITE did not customize anything. A proper fix would require creating a TransformFeatureLocking down in gt-transform. Do you have a ticket to track this issue?

davidblasby: https://osgeo-org.atlassian.net/browse/GEOS-10786

jodygarnett: I think the fix as provided is good; fixing regression.

Adding a TransformFeatureLocking could be a separate GeoTools improvement.

aaime: It's an improvement over the current situation, for sure. 

The ticket, "Cannot lock Customized FeatureTypes", should be reworded, this fix does not allow locking customized types, it just makes the specific situation of the WFS 1.1 data directory work.

I cannot stress it enough, tickets are not bureaucracy, they are our public communication to users, we derive release notes and release announcements from them, and users look them up later when trying to figure out when and how to upgrade, or when a specific "fix" caused their odd usage of GeoServer break (one of the common cases of issues when upgrading in my experience, is that we fixed some bug users used to rely onto... think for example the WCS store parameter one, which turns a lenient behavior into a strict one).

jodygarnett: Suggestion, split into two tickets:
- Improvement: Check customized feature type to determine if transform wrapper needed
- Improvement: Add WFS-T Locking for customized feature types

davidblasby: I changed the name of the 2 tickets

jodygarnett (approved):
Ticket has been updated to reflect the improvement made.

PR #6397, opened 2022-11-30
[GEOS-10784] CITE WFS 1.1 - don't do illegal geometry conversions

davidblasby:
[GEOS-10784](https://osgeo-org.atlassian.net/browse/GEOS-10784)

One of the CITE tests attempts to put the wrong geometry type (multicurve (linestring) into a polygon).

JIRA: https://osgeo-org.atlassian.net/browse/GEOS-10784

Geoserver would convert this linestring to a polygon and do the insert.

The Polygon that it created just appended all the line's coordinates together, so it wasnt a very good conversion.

CITE test explicitly check that this situation throws an exception. 

This change allows for most geometry xforms, but doesn't allow dimensional changes (i.e. 2d to 1d).

Here is the CITE test:

https://github.com/opengeospatial/ets-wfs11/blob/master/src/main/scripts/ctl/wfs-transaction/Transaction/Transaction-XML.xml#L815

 

# Checklist

- [x ] All the build checks are green ([see automated QA checks](https://docs.geoserver.org/latest/en/developer/qa-guide/index.html)).

### DIFF
--- a/src/wfs/src/test/java/org/geoserver/wfs/UpdateElementHandlerTest.java
+++ b/src/wfs/src/test/java/org/geoserver/wfs/UpdateElementHandlerTest.java
@@ -1,0 +1,40 @@
+/* (c) 2022 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.wfs;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import org.geotools.geometry.jts.WKTReader2;
+import org.junit.Test;
+import org.locationtech.jts.geom.Geometry;
+import org.locationtech.jts.geom.LineString;
+import org.locationtech.jts.geom.MultiPoint;
+import org.locationtech.jts.geom.Point;
+import org.locationtech.jts.geom.Polygon;
+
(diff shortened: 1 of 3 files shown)


aaime:
Do you mean "illegal"? :rofl:

src/wfs/src/test/java/org/geoserver/wfs/UpdateElementHandlerTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.geotools.geometry.jts.WKTReader2;
+import org.junit.Test;
+import org.locationtech.jts.geom.Geometry;
+import org.locationtech.jts.geom.LineString;
+import org.locationtech.jts.geom.MultiPoint;
+import org.locationtech.jts.geom.Point;
+import org.locationtech.jts.geom.Polygon;
+
+public class UpdateElementHandlerTest {
```
aaime, line 18: Nice to have a unit test, but I'd feel more comfortable merging if there was also a end-to-end test on transaction. Pick any TransactionTest (there are 3, one per version) and adapt an existing test using the wrong type of geometry, check that the right service exception is returned.

jodygarnett (approved):

jodygarnett:
Thanks David; I reviewed this earlier when troubleshooting the problem. This fix is a good approach.

PR #6399, opened 2022-11-30
Fix GeoServerDataDirectoryTest clean up after running tests

sikeoka:
GeoServerDataDirectoryTest doesn't properly clean up the application context after it is finished.  This causes more than 2,000 stack traces in the logs about "BeanFactory not initialized or already closed" when running the gs-main unit tests.

  

# Checklist

### DIFF
--- a/src/main/src/test/java/org/geoserver/config/GeoServerDataDirectoryTest.java
+++ b/src/main/src/test/java/org/geoserver/config/GeoServerDataDirectoryTest.java
@@ -45,6 +45,7 @@ public class GeoServerDataDirectoryTest {
 
     @AfterClass
     public static void afterClass() {
+        GeoServerExtensionsHelper.init(null);
         factory = null;
     }
 


sikeoka:
Could you take a look at this @jodygarnett?  This is related to work you did last year https://github.com/geoserver/geoserver/commit/dc38d4a4cb35b7dd002bc394f3e6431ef843a042#diff-025b0d07d7e155fbe0aead6cb69df188f2f3f1746c0a5cbbc2606b3c7d40a415

src/main/src/test/java/org/geoserver/config/GeoServerDataDirectoryTest.java
```diff
@@ -45,6 +45,7 @@ public static void beforeClass() {
 
     @AfterClass
     public static void afterClass() {
+        GeoServerExtensionsHelper.init(null);
```
jodygarnett, line 48: So what effect does that have? Do we need to update the javadoc to explain?

jodygarnett:
Thanks for looking at this! Is there anything else we need to do to clean up the context?

src/main/src/test/java/org/geoserver/config/GeoServerDataDirectoryTest.java
sikeoka: This removes the ApplicationContext from GeoServerExtensions when this test class is finished so that the next test class that calls GeoServerExtensionsHelper.singleton won't modify the old ApplicationContext which will cause the stack trace every time GeoServerExtensions tries to look up a bean after that.  

This is done in several existing unit tests like:
https://github.com/geoserver/geoserver/blob/main/src/gwc/src/test/java/org/geoserver/gwc/GWCTest.java#L1646

sikeoka:
Any additional thoughts on this PR?  It's a simple fix that reduced the size of a successful build log from ~15MB to ~3.5MB.

aaime:
Oh right, I've started the build locally, to look at the output, and then got a call and forgot about it. Yep, it looks good IMHO. Merging.

PR #6402, opened 2022-12-02
[GEOS-10578] OGCAPI - Features - Add `storageCrs` property for Collections

sebastianfrey:
[GEOS-10758](https://osgeo-org.atlassian.net/browse/GEOS-10758)

  

# Checklist

### DIFF
--- a/src/community/ogcapi/ogcapi-features/src/main/resources/org/geoserver/ogcapi/features/collection_include.ftl
+++ b/src/community/ogcapi/ogcapi-features/src/main/resources/org/geoserver/ogcapi/features/collection_include.ftl
@@ -1,10 +1,13 @@
 <div class="card-body">
   <ul>
-    <#if collection.title??> 
+    <#if collection.title??>
       <li><b>Title</b>: <span id="${collection.htmlId}_title">${collection.title}</span><br/></li>
       </#if>
       <#if collection.description??>
       <li><b>Description</b>: <span id="${collection.htmlId}_description">${collection.description!}</span><br/></li>
+      </#if>
+      <#if collection.storageCrs??>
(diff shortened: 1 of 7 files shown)


src/community/ogcapi/ogcapi-features/src/main/resources/org/geoserver/ogcapi/features/collection_include.ftl
```diff
@@ -1,11 +1,14 @@
...
   <ul>
-    <#if collection.title??> 
+    <#if collection.title??>
       <li><b>Title</b>: <span id="${collection.htmlId}_title">${collection.title}</span><br/></li>
       </#if>
       <#if collection.description??>
       <li><b>Description</b>: <span id="${collection.htmlId}_description">${collection.description!}</span><br/></li>
       </#if>
+      <#if collection.storageCrs??>
+      <li><b>Storage CRS</b>: <span id="${collection.htmlId}_storageCrs">${collection.storageCrs!}</span><br/></li>
```
sebastianfrey, line 10: I am unsure about this line, because the current implementation renders the full CRS-URI. Maybe it would be better to render only the shortened string version of the CRS, for example `CRS-84` or `EPSG:3857.

aaime: I suggest avoid it, as shortened CRS codes have had a specific axis order meaning for years:
https://docs.geoserver.org/latest/en/user/services/wfs/axis_order.html

sebastianfrey: Sounds valid. Than I will leave it as it is right now and render the full CRS-URI.

PR #6427, opened 2022-12-12
[GEOS-10787] CITE WCS 1.1.1 - Throw exception on bad 'store' parameter

davidblasby:
[GEOS-10787](https://osgeo-org.atlassian.net/browse/GEOS-10787)

According to WFS 1.1 spec, the “store” parameter must be "True" or "False". 

CITE test "GetCoverage_Store_Bogus" tests that an exception is thrown when its set to something else.

Geoserver treats an unknown value (ie. “bogus”) as false.

JIRA: https://osgeo-org.atlassian.net/browse/GEOS-10787  

NOTE: an alternative place to fix this would be in `BooleanKvpParser`. However, I think this would cause some side-effects.

# Checklist

### DIFF
--- a/src/wcs1_1/src/main/java/org/geoserver/wcs/kvp/GetCoverageRequestReader.java
+++ b/src/wcs1_1/src/main/java/org/geoserver/wcs/kvp/GetCoverageRequestReader.java
@@ -5,6 +5,7 @@
  */
 package org.geoserver.wcs.kvp;
 
+import static org.vfny.geoserver.wcs.WcsException.WcsExceptionCode.InvalidParameterValue;
 import static org.vfny.geoserver.wcs.WcsException.WcsExceptionCode.MissingParameterValue;
 
 import java.util.Map;
@@ -17,6 +18,7 @@ import net.opengis.wcs11.TimeSequenceType;
 import net.opengis.wcs11.Wcs111Factory;
 import org.geoserver.catalog.Catalog;
 import org.geoserver.ows.kvp.EMFKvpRequestReader;
+import org.geoserver.ows.util.KvpUtils;
 import org.vfny.geoserver.wcs.WcsException;
 import org.vfny.geoserver.wcs.WcsException.WcsExceptionCode;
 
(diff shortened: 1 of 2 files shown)


src/wcs1_1/src/main/java/org/geoserver/wcs/kvp/GetCoverageRequestReader.java
```diff
@@ -45,6 +47,17 @@ public Object read(Object request, Map<String, Object> kvp, Map<String, Object>
         // build output element
         getCoverage.setOutput(parseOutputElement(kvp));
 
+        String store = KvpUtils.caseInsensitiveParam(rawKvp, "store", null);
+        if (store != null) {
```
mprins, line 51: This null check can be omitted if you swap the String comparison placing the static first

aaime: Some users out there might be relying on this lenient behavior. Wondering if the stricter approach should be enabled when the strict OGC compliance flag is raised?
Mostly thinking out loud, as I don't personally believe many people are using store=true to start with, and setting store to other values would be the same as not setting it at all.

davidblasby: @mprins - I don't understand what you are saying...

@aaime - I can add a check for strictness (I'll need the bean).  However, very few people will be using `store=...` let alone to an invalid value.  Right now, setting `store=foo` is the same as `store=false` which seems a bit strange that someone would actually set it but expect it to be false...

aaime: As said, "mostly thinking out loud", wanted to check if you or @jodygarnett had an opinion on the topic.

davidblasby: I'm fine either way - but I think its ok the way it is.  I can change the ticket title to "WCS STORE parameter - but be missing (false), 'true', or 'false'" as that will make it more obvious to people looking at release notes.

jodygarnett: @David I think @mprins is providing a coding style feedback: 
```
if (!"TRUE".equalsIgnoreCase(store) && !"FALSE"..equalsIgnoreCase(store)) {
  ...
}
```
Personally I find the ``store != null`` check to be clear.

aaime:
Let's give it a run and see how it goes. I'm not sure about backport labels due to the potential backwards incompatibilty. Maybe 2.22.x but not 2.21.x?

PR #6428, opened 2022-12-13
[GEOS-10788] CITE WCS 1.1.1 -  Empty InterpolationMethod should throw exception

davidblasby:
[GEOS-10788](https://osgeo-org.atlassian.net/browse/GEOS-10788)

While running the WCS 1.1.1 CITE tests, GS wasn't passing some of the interpolationMethod tests.

This was discussed with OGC CITE - https://github.com/opengeospatial/ets-wcs11/issues/76

Because the spec is not specific, the CITE test will be changed to only check for an exception report.

The PR only addresses the outstanding problem - ensuring that GS throws an exception where there is a blank interpolationMethod.

Here is an example CITE test;

https://github.com/opengeospatial/ets-wcs11/blob/master/src/main/scripts/ctl/GetCoverage.xml#L5586

  

# Checklist

### DIFF
--- a/src/wcs1_1/src/main/java/org/geoserver/wcs/DefaultWebCoverageService111.java
+++ b/src/wcs1_1/src/main/java/org/geoserver/wcs/DefaultWebCoverageService111.java
@@ -957,6 +957,14 @@ public class DefaultWebCoverageService111 implements WebCoverageService111 {
                     || interpolation.equalsIgnoreCase("bilinear")) {
                 interpolation = "bilinear";
             }
+            if (interpolation.trim().isEmpty()) {
+                // ie. "contents:"  WCS Spec doesn't specify the exact error message/locator
+                //  This satisfies the WCS 1.1 CITE Tests
+                throw new WcsException(
+                        "RangeSubset parameter - InterpolationMethod is empty",
+                        InvalidParameterValue,
+                        "RangeSubset");
+            }
 
             for (String method : info.getInterpolationMethods()) {
                 if (method.toLowerCase().startsWith(interpolation)) {
--- a/src/wcs1_1/src/test/java/org/geoserver/wcs/GetCoverageTest.java
+++ b/src/wcs1_1/src/test/java/org/geoserver/wcs/GetCoverageTest.java
@@ -75,6 +75,42 @@ public class GetCoverageTest extends AbstractGetCoverageTest {
     }
 
     @Test
+    public void testMissingInterpolation() throws Exception {
+        String request =
+                "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\r\n"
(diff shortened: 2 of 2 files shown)


jodygarnett (commented):
David this looks good to me, but I am not comfortable approving a change  in the WCS module.

aaime:
The changes look good in general, what is surprising me is the syntax of the locator. Normally the locator is the name of the key that would be involved in the error, if the request was made using a KVP GET request. This one looks more similar to an XML XPath instead, which I don't recall seeing before. @davidblasby what are the CITE tests actually looking for? Specifically for the form you used?

Also, according to the spec, the interpolation specification is not mandatory:

image

The test seems to be checking whether the interpolation is present but an empty string, which of course is not valid... but why not check for invalid values instead? The change seems more of a CITE tests workaround, but mind, often times the tests themselves are wrong (I've reported many to OGC while playing with CITE tests, over the years).

davidblasby:
Andrea,

I was surprised at the syntax of the locator as well. 

Here an example of one of their tests;
https://github.com/opengeospatial/ets-wcs11/blob/master/src/main/scripts/ctl/GetCoverage.xml#L5586

That test case has the request look like `content:`, which is invalid.

That same test has a XML Post:
https://github.com/opengeospatial/ets-wcs11/blob/master/src/main/scripts/ctl/GetCoverage.xml#L5573

 `<InterpolationType/>` - which is "present but null"

There are a few other tests that are similar.

GS was, previously, ignoring some of these issues.  I also made the errors a bit more specific.  

The specification is ambiguous on what the actual error message should be.  I expect the CITE test is incorrect (the text looks too specific), but there isn't anything in the specification that says what it should be.

davidblasby:
I put a bug report in the OGC github:
https://github.com/opengeospatial/ets-wcs11/issues/76

@aaime - perhaps comment there?

davidblasby:
Looks like the OGC will be discussing this - lets put this PR to sleep until they give us guidance.

davidblasby:
I am making an ogc-wcs CITE PR to change the test case behaviour.

However, there still needs to be changes in the GS code base because its not throwing exceptions when it should.

I'm going to add part of this pull request in and have it throw an exception if they have the missing info.

davidblasby:
see https://github.com/opengeospatial/ets-wcs11/pull/77

davidblasby:
DO NOT MERGE UNTIL OGC MERGES

davidblasby:
OGC has merged https://github.com/opengeospatial/ets-wcs11/pull/77

This can be merged now.

jodygarnett (approved):
This looks good to me, OGC test is now just testing for an exception. The exception message provided by GeoServer is clear enough.

PR #6430, opened 2022-12-14
[GEOS-10792] acceptVersions negotiation isn't working with WCS

davidblasby:
[GEOS-10792](https://osgeo-org.atlassian.net/browse/GEOS-10792)

When using the acceptVersions negotiation with WCS (not sure about other systems), the system gets into a “half-2.0.2” and “half-1.1.1” situation.   You get an error.

For example,

https://how2map.geocat.live/geoserver/ows?service=WCS&request=getcapabilities&acceptVersions=1.1.1

JIRA: https://osgeo-org.atlassian.net/browse/GEOS-10792

Looking here;

[geoserver/GetCapabilities.java at main · geoserver/geoserver](https://github.com/geoserver/geoserver/blob/main/src/wcs2_0/src/main/java/org/geoserver/wcs2_0/GetCapabilities.java#L49)

You can see it trying to “re-dispatch” a 1.1.1 result in the 2.0.2 service.  This causes issues during output.

This PR does parsing of the acceptVersion (KVP and XML) and puts them in the Request object.
It then does version negotiation.  

# TO DO

1. feedback to make sure this is ok
2. Test case
3. remove this old code.  Is there anywhere else that needs removing?
 [geoserver/GetCapabilities.java at main · geoserver/geoserver](https://github.com/geoserver/geoserver/blob/main/src/wcs2_0/src/main/java/org/geoserver/wcs2_0/GetCapabilities.java#L49)

# Checklist

### DIFF
--- a/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
+++ b/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
@@ -22,6 +22,7 @@ import java.lang.reflect.Method;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
@@ -47,7 +48,6 @@ import javax.xml.stream.XMLStreamReader;
 import javax.xml.transform.TransformerFactory;
 import javax.xml.transform.dom.DOMSource;
 import javax.xml.transform.stream.StreamResult;
-import net.opengis.ows11.AcceptVersionsType;
 import org.apache.commons.fileupload.FileItem;
 import org.apache.commons.fileupload.disk.DiskFileItemFactory;
 import org.apache.commons.fileupload.servlet.ServletFileUpload;
@@ -513,16 +513,15 @@ public class Dispatcher extends AbstractController {
         if (req.getService().isEmpty() || !req.getRequest().equalsIgnoreCase("getCapabilities")) {
             return;
         }
-        AcceptVersionsType acceptVersions = (AcceptVersionsType) req.getKvp().get("AcceptVersions");
+        String acceptVersions = (String) req.getRawKvp().get("AcceptVersions");
         if (acceptVersions == null) {
             return;
         }
-        @SuppressWarnings(value = "unchecked")
+
         List<Version> acceptVersionsList =
-                ((List<String>) acceptVersions.getVersion())
-                        .stream()
-                                .map(x -> new Version(normalizeVersion(x)))
-                                .collect(Collectors.toList());
+                Arrays.asList(acceptVersions.split(",")).stream()
+                        .map(x -> new Version(normalizeVersion(x)))
+                        .collect(Collectors.toList());
 
         req.setAcceptVersions(acceptVersionsList);
     }
--- a/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
+++ b/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
@@ -598,9 +598,12 @@ public class Dispatcher extends AbstractController {
                 throw new ServiceException(msg, "InvalidParameterValue", "service");
             }
         }
-        req.setVersion(serviceDescriptor.getVersion().toString());
         req.setServiceDescriptor(serviceDescriptor);
-        return fireServiceDispatchedCallback(req, serviceDescriptor);
+        Service result = fireServiceDispatchedCallback(req, serviceDescriptor);
+        if (req.getVersion() == null) {
+            req.setVersion(serviceDescriptor.getVersion().toString());
+        }
+        return result;
     }
 
     Service fireServiceDispatchedCallback(Request req, Service service) {
--- a/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
+++ b/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
@@ -48,6 +48,8 @@ import javax.xml.stream.XMLStreamReader;
 import javax.xml.transform.TransformerFactory;
 import javax.xml.transform.dom.DOMSource;
 import javax.xml.transform.stream.StreamResult;
+import net.opengis.ows10.AcceptVersionsType;
+import net.opengis.ows10.Ows10Factory;
 import org.apache.commons.fileupload.FileItem;
 import org.apache.commons.fileupload.disk.DiskFileItemFactory;
 import org.apache.commons.fileupload.servlet.ServletFileUpload;
@@ -258,6 +260,8 @@ public class Dispatcher extends AbstractController {
             }
 
             // dispatch the operation
+            updateAcceptVersions(request, service);
+
             Operation operation = dispatch(request, service);
             request.setOperation(operation);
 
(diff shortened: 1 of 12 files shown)


jodygarnett (commented):
Thanks David I much prefer to see version negotiation handled at Dispatcher / Request level (rather than WFS and WCS trying to sort it out on their own).

davidblasby:
Added test case specific to the changes (verify parsing the acceptVersion in KVP and XML).  The existing test cases test the rest.

jodygarnett:
@davidblasby with respect to:

> feedback to make sure this is ok

I think this is a good change, and we discussed briefly at a geoserver meeting.

src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
```diff
@@ -500,6 +503,29 @@ BufferedReader fileItemReader(FileItem fileItem) throws IOException {
...
+        }
+        String acceptVersions = (String) req.getRawKvp().get("AcceptVersions");
+        if (acceptVersions == null) {
+            return;
+        }
+
+        List<Version> acceptVersionsList =
+                Arrays.asList(acceptVersions.split(",")).stream()
+                        .map(x -> new Version(normalizeVersion(x)))
+                        .collect(Collectors.toList());
```
jodygarnett, line 524: Do you need to sort this list, or is it documented to be in the order preferred by the client?

davidblasby: yes - the order specified in the request is important.

jodygarnett:
Can this be taken out of draft @davidblasby ?

davidblasby:
I think this is the only thing remaining - 

3. remove this old code. Is there anywhere else that needs removing?
[geoserver/GetCapabilities.java at main · geoserver/geoserver](https://github.com/geoserver/geoserver/blob/main/src/wcs2_0/src/main/java/org/geoserver/wcs2_0/GetCapabilities.java#L49)

I'll look at removing it tomorrow.

src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
```diff
@@ -1159,6 +1197,20 @@ Service findService(String id, String ver, String namespace) throws ServiceExcep
 
         Service sBean = null;
 
+        if ((version == null) && (acceptsVersions != null)) {
+            // version negotiation
+            // 06-121r3: The server, upon receiving a GetCapabilities request, shall scan through
+            // this list and find the first version number that it supports.
```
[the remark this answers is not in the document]
aaime: Agree. I would like to see a test of this behavior though. The tests seem to check parsing occurs correctly, but I don't see a test checking "the first version the server supports" has been chosen. I guess it would be easy-ish to write in the WFS module, which sports 3 different WFS versions in a single module.

Guess what, I've found a VersionNegotiationTest in WFS, so I'm guessing the WFS machinery also has code to do negotiation... oh, it's actually in a RequestUtils method it seems (so possibly some duplication with the work happening here):
https://github.com/geoserver/geoserver/blob/de184b2f9ed9a844b5291dd163e772132a07689b/src/wfs/src/main/java/org/geoserver/wfs/GetCapabilities.java#L97

davidblasby:
@aaime -
1. removed some redundant code in the wcs 2.0 module
2. now use RequestUtils.getVersionPreOws()

src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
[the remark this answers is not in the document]
jodygarnett: Trying to remove this with david showed a number different things:

* Some version negotiation happens in KVP parsing; producing OWS11Version data structures which do not work correctly when talking to WFS 1.0
* Reducing everything to a single acceptsVersion method; and then updating the KVP works ... but then a whole bunch of test cases across the codebase fail - where their are testing the accepts version functionality that has been implemented separately in the individual modules. While this is expected as the code is obsolete / removed
* When WFS 0.8 is provided OWS1.0 version negotiation expects 1.0 (closest), while OWS1.1 expects 2.0 (highest) so some changes were needed.

I love the idea of Dispatcher handling acceptsVersions; but I think each protocol that has its own logic would need to migrate over to trusting dispatcher to do the right thing.

I am not clear how (or if) fixing the KVP Parser version handling should be done; WFS 1.0 was producing a strongly typed OWS11 Version object rather than a string.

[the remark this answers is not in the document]
jodygarnett: I expect that WFS code was used to process the ``http://localhost/geoserver/wfs?`` URLs (when ``SERVICE=WFS`` is not included). In any case dispatcher change is a good approach; but the other modules doing their own accceptsVersions negotiation may need to be migrated over time; and have their test cases adjusted to match.

davidblasby:
Converting back to draft.
There are too many problems in other parts of the code base - like the XML parser and KVP parsers.  There's also quite a bit of ambiguity in the spec (and cite tests) for version negotiation. 

Also, I think a lot of users are using (incorrect) functionality (i.e. using version= when they should be using acceptVersions=).

davidblasby:
ok, spent a few more days on this.

This doesn't capture every situation, but it captures all the important/common issues.  

I noticed a few issues with test cases, which I've updated.  I also added more tests.

@jodygarnett  @aaime

src/ows/src/main/java/org/geoserver/ows/util/RequestUtils.java
```diff
@@ -63,7 +63,7 @@ public static String getVersionPreOws(List<String> providedList, List<String> ac
         if (acceptedList == null || acceptedList.isEmpty()) return provided.last().toString();
 
         // next figure out what the client accepts (and check they are good version numbers)
-        TreeSet<Version> accepted = new TreeSet<>();
+        List<Version> accepted = new ArrayList<>();
```
aaime, line 66: This change, along with the method javadoc,  prompted me to look at the specifications better.

As you can see, RequestUtils has two negotiation methods, one for pre-OWS 1.1 services, and one for OWS 1.1+ services, while ``getVersionOws11`` should be used for those based on OWS 1.1.

AcceptVersions was introduced in OWS 1.1, some services simply do not have such parameter in their spec: WMS 1.1 and 1.3, WFS 1.0, WCS 1.0 just don't have it, and have their own negotiation mechanism that is different from what OWS 1.1 mandates. In particular, in those protocols "version" is the parameter to be used, and the algorithm is the following (picked from WMS 1.3):

> 
>  *  If a version unknown to the server and higher than the lowest supported version is requested, the server shall send the highest version it supports that is less than the requested version.
> * If a version lower than any of those known to the server is requested, then the server shall send the lowest version it supports.
> * If the client does not support the version sent by the server, it may either cease communicating with the server or send a new request with a different version number that the client does support. 

WFS 1.1 is a strange beast, in that it has an ``acceptVersions`` parameter, but still mandates the above version negotiation algorithm.

Since OWS 1.1 instead, it works differently, again from the spec:

> The server, upon receiving a GetCapabilities request, shall scan through this list and find
the first version number that it supports. It shall then return a service metadata document
conforming to that version of the specification, and containing that value of the ―version‖
parameter. If the list does not contain any version numbers that the server supports, the
server shall return an Exception with exceptionCode="VersionNegotiationFailed".

Long story short, we need to find a mechanism to figure out which of the two negotiation mechanism to use. Checking if AcceptVersions is present is tempting, but the WFS 1.1 seems to break that approach... wondering if we can simply ignore it and treat it as if it was a OWS 1.1 service....

src/web/app/src/test/java/org/geoserver/integration/VersionNegotiationKVPIntegrationTest.java
```diff
@@ -0,0 +1,297 @@
...
+package org.geoserver.integration;
+
+import static org.junit.Assert.assertEquals;
+
+import org.geoserver.test.GeoServerSystemTestSupport;
+import org.junit.Test;
+import org.w3c.dom.Document;
+
+/** Tests version negotiation for GetCapabilities Requests (KVP). */
+public class VersionNegotiationKVPIntegrationTest extends GeoServerSystemTestSupport {
```
aaime, line 14: Nice cross service integration test :D

src/wfs/src/test/java/org/geoserver/wfs/v1_1/VersionNegotiationTest.java
```diff
@@ -45,7 +45,10 @@ public void test0() throws Exception {
         request.getAcceptVersions().getVersion().add("1.1.0");
 
         TransformerBase tx = getCaps.run(GetCapabilitiesRequest.adapt(request));
-        assertTrue(tx instanceof CapabilitiesTransformer.WFS1_1);
+        assertTrue(
+                tx
+                        instanceof
+                        CapabilitiesTransformer.WFS1_0); // choose first appropriate AcceptVersions
```
aaime, line 51: The formatter mandates 100 columns, best not put trailing comments unless very short

```suggestion
        // choose first appropriate AcceptVersions
        assertTrue(tx instanceof CapabilitiesTransformer.WFS1_0); 
```

aaime (changes requested):

davidblasby:
I'm going to have to down-tools on this and hand it over to @jodygarnett .

jodygarnett:
Hello bot; I will be your friend and keep you company.

aaime:
@jodygarnett I don't see the reason of keeping a stale PR open... one can link it from the ticket and close it, it will be found by interested parties willing to complete the work, no?

jodygarnett:
I was mostly just being funny; although I am interested in making this PR work.

PR #6447, opened 2022-12-20
[GEOS-10804] Update Resource API with representation absolute paths

jodygarnett:
[GEOS-10804](https://osgeo-org.atlassian.net/browse/GEOS-10804)

This was found by a Java 11 highlighting the difference between ``C:folder`` relative paths and ``C:\folder`` absolute paths.

* [x] This fix adjusts Paths.names(String) list allowing the fist path to be a Linux ``/`` or Windows ``C:/`` absolute path.
* [x] Documentation updated (there was not a lot so I copied javadocs)
* [x] Resource REST API ``rest/resources/directory`` is preferred ( ``rest/resources//directory`` still allowed)
* [x] Resource Browser tree model also updated to avoid ``//`` double slashes, paths are now validated on move / rename / creation
* [x] A lot of validation checks added to confirm handling of paths is consistent

# Checklist


jodygarnett:
This issue is a carry over from discussion on PR https://github.com/geoserver/geoserver/pull/6370

src/platform/src/main/java/org/geoserver/platform/resource/Paths.java
```diff
@@ -459,12 +495,60 @@ public static String convert(String path, String filename) {
...
     }
 
+    /**
+     * Pattern used to recognize absolute path in Windows, as indicated by driver letter reference
+     * (included {@code :}), followed by as slash character.
+     *
+     * <p>Aside: A drive letter reference on its own results in a relative path (relative to the
+     * current directory for that drive).
+     */
+    static final Pattern WINDOWS_DRIVE_LETTER = Pattern.compile("^\\w\\:/.*$");
```
aaime, line 505: Hum... I think that now I can enter a path with backward slashes in the UI and it will work. 

I read somewhere in your comments (and hopefully I just misunderstood) that we can have our own conventions for paths... that would be wrong IMHO, the paths are visible to users, they enter them also manually, so we should be respecting the conventions of the OS they are already using.

src/platform/src/main/java/org/geoserver/platform/resource/Resources.java
```diff
@@ -711,7 +711,7 @@ public static URL toURL(final Resource res) {
                     "resource",
                     null,
                     -1,
-                    String.format(res.getType() == Type.DIRECTORY ? "/%s/" : "/%s", res.path()),
+                    String.format(res.getType() == Type.DIRECTORY ? "%s/" : "%s", res.path()),
```
aaime, line 714: Is this going to change anything for existing data directories?

src/platform/src/test/java/org/geoserver/platform/resource/ResourceTheoryTest.java
```diff
@@ -486,10 +486,10 @@ public void theoryRecursiveDelete(String path) throws Exception {
     }
 
     @Theory
-    public void theoryRootSlashIsIgnored(String path) throws Exception {
-        final Resource res = getResource(path);
-        final Resource res2 = getResource("/" + path);
-        assertEquals(res, res2);
-        assertEquals(res.path(), res2.path());
+    public void theoryRootSlashIsAbsoluteOnLinux(String path) throws Exception {
+        if (!SystemUtils.IS_OS_WINDOWS) {
```
aaime, line 490: I don't think this is what we want, just ignoring a test on Windows makes the test pass but will the weird behavior remain?

src/platform/src/main/java/org/geoserver/platform/resource/Paths.java
jodygarnett: I added some docs to the PR to try and make it more clear

doc/en/developer/source/programming-guide/config/resource.rst
```diff
@@ -0,0 +1,202 @@
...
+* `GSIP-137 - ResourceStore Rest API <https://github.com/geoserver/geoserver/wiki/GSIP-137>`__
+* `GSIP-136 - Resource Notification Dispatcher <https://github.com/geoserver/geoserver/wiki/GSIP-136>`__
+* `GSIP-132 - GSIP 132 - Resource Store changes <https://github.com/geoserver/geoserver/wiki/GSIP-132>`__
+
+Before we start
+---------------
+
+* ``path`` relative location of resource within the data directory
+* ``file`` a java file reference
+* ``location`` (deprecated) a relative location within the data directory, or an absolute location on disk
```
jodygarnett, line 25: This is the important distinction for these APIs

src/platform/src/main/java/org/geoserver/platform/resource/Paths.java
```diff
@@ -459,12 +495,60 @@ public static String convert(String path, String filename) {
...
+     * is some support for absolute paths.
+     *
+     * <p><b>Linux</b>: Linux absolute paths start with a leading {@code /} character. As this slash
+     * character is also used as the path separator special handling is required. Notably {@link
+     * #names(String)} will represent an absolute path as: {@code { "/", "srv" "gis", "cadaster",
+     * "district.geopkg"}}
+     *
+     * <p><b>Windows</b>: Windows absolute paths start with a drive letter, colon, and slash
+     * characters.{@link #names(String)} will represent an absolute path on windows as: {@code {
+     * "D:/, "gis", "cadaser", "district.geopkg"}}
```
bradh, line 518:
```suggestion
     * "D:/, "gis", "cadaster", "district.geopkg"}}
```

bradh (commented):
A few doc suggestions.

doc/en/developer/source/programming-guide/config/catalog.rst
```diff
@@ -0,0 +1,123 @@
...
+   //get all feature types in a namespace
+   NamespaceInfo ns = catalog.getNamespaceByURI( "http://myNamespace.org" );
+   List<FeatureTypeInfo> featureTypes = catalog.getResourcesByNamespace( ns, FeatureTypeINfo.class );
+
+## Layers
+
+A layers is used to publish a resource. The ``getLayers()`` provides access to all layers
+in the catalog. A layer is uniquely identified by its name. The ``getLayerByName(String)``
+method provides access to a layer by its name. The ``getLayers(ResourceInfo)`` return all
+the layers publish a specific resource. Some examples:
```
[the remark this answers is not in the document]
jodygarnett: Some of this is straight from the javadocs, this section seems confusing.

src/platform/src/main/java/org/geoserver/platform/resource/Resources.java
jodygarnett: This changes what is being passed around internally at runtime (to make clear difference between (File, Path, and Location).

The internal resource URL was introducing a ``/`` prefix here presenting an odd situation where the generated path would get double slash ``//`` as part of the URL. The processing of Path was covering for this by ignoring the empty ``""`` directory name between the two slashes ``//``.

By removing this double slash the code is easier to follow.

src/platform/src/test/java/org/geoserver/platform/resource/ResourceTheoryTest.java
jodygarnett: I believe I setup a test that covers the windows behaviour also. Using the fact that java ``user.home`` will be located on a drive letter in order to have something to test against.

src/platform/src/test/java/org/geoserver/platform/resource/PathsTest.java
```diff
@@ -148,23 +162,35 @@ public void parentTest() {
...
         File file1 = new File("file1");
         File file2 = new File(folder, "file2");
 
         assertEquals("folder", Paths.convert(folder.getPath()));
         assertEquals("folder/file2", Paths.convert(file2.getPath()));
         assertEquals("file1", Paths.convert(file1.getPath()));
+
+        // absolute paths
+        File home = new File(System.getProperty("user.home")).getCanonicalFile();
+        assertTrue(isAbsolute(Paths.convert(home.getPath())));
```
jodygarnett, line 193: @aaime this is the test that covers use of windows absolute path (to ensure ability to handle windows drive letter correctly).

src/platform/src/test/java/org/geoserver/platform/resource/ResourceTheoryTest.java
jodygarnett: Adjusted this test so it uses an absolute path on both linux and windows.

jodygarnett:
Okay I added a bunch more testing to increase my confidence; added some validation to web-resource browser extension.

aaime (commented):
Made a bunch of interactive tests on Windows and Linux:
- Play with topp:states store URL param (relative to data dir, but also as absolute path), 
- Play with sf:sfdem store URL (relative to data dir)
- Check "burg" style works (relative reference to svg in the style)
- Changing the location of the log file within the data directory and outside with an absolute path
- Played with the resource browser a bit

Found no issues. See comments below.

aaime:
Oh, unrelated issue... I was wondering how to create a directory in the browser, using "new resource" it suggest to type a path in order to create a directory... but if I write "temp" it just creates a file, and if I type "/temp" it does the same... duh?
Eventually figured out I have to create an actual file in a path, and then the path gets created... weird (habits from the file system, one first creates a directory, and only then create a file, if a file creation right away is wanted at all, that is).

doc/en/developer/source/programming-guide/config/resource.rst
```diff
@@ -0,0 +1,202 @@
...
+A Resource can also be extracted to a file if needed:
+
+.. code-block:: java
+
+   File file = resourceStore.get("module/logo.png");
+   BufferedImage img = ImageIO.read( file );
+
+The base directory is available using ``Paths.BASE`` (as ``""`` but relative paths (``.`` and
+``..`` are not supported). Path assumes a unix like file system, all paths are relative and use forward slash
+{@code /} as the separator.
```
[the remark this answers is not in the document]
jodygarnett: It is from the javadocs, that what I used as a starting point for these pages.

doc/en/developer/source/programming-guide/config/catalog.rst
```diff
@@ -0,0 +1,123 @@
+.. _config_catalog:
+
+Catalogue API
```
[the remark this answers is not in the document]
jodygarnett: Thanks; the entire topic was not covered so I used the javadocs to bootstrap some material. 

In particular we seem to have lost the original catalog api design as it was on the old wiki. There were some handy diagrams there.

doc/en/developer/source/programming-guide/config/geoserver.rst
```diff
@@ -0,0 +1,98 @@
+.. _config_geoserver:
+
+GeoServer
+=========
+
+Configuration is managed by a top-level ``GeoServer`` object providing ``save()`` and ``reload()`` functionality.
```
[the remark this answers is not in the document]
jodygarnett: Yes and I agree. I tried to bootstrap this section of the developers guide so I could have some context and a section on the Resource API.

doc/en/developer/source/programming-guide/config/resource.rst
```diff
@@ -0,0 +1,202 @@
...
+
+Paths are broken down into a sequence of names, as listed by ``Paths.names(path)``:
+
+* ``Path.names("data/tasmania/roads.shp")`` is represented as a list of ``data``, ``tasmania``, ``roads.shp``.
+* On linux ``Path.names("/src/gis/cadaster/district.geopkg")`` starts with a marker to indicate an absolute path, resulting in ``/``, ``src``, ``gis``, ``cadaster``, ``district.geopkg``.
+* On windows ``Path.names("D:/gis/cadaster/district.geopkg")`` starts with a marker to indicate an absolute path, resulting in ``D:/``, ``gis``, ``cadaster``, ``district.geopkg``.
+
+The ``convert`` methods are used to process file references into resource paths:
+
+* ``Paths.convert(base,file)`` - uses URI relativize to determine relative path
```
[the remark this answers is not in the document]
jodygarnett: It is the name of the URI method - https://docs.oracle.com/javase/7/docs/api/java/net/URI.html#relativize(java.net.URI)

Indeed a bunch of the Java URI docs talk about management of relative paths.

jodygarnett, line 104:
```suggestion
* ``Paths.convert(base,file)`` - uses URI relativize method to determine relative path (between file and base)
```

doc/en/developer/source/programming-guide/config/resource.rst
```diff
@@ -0,0 +1,202 @@
...
+   for (Resource svg : Resources( resource, new ExtensionFilter("svg"), true )) {
+      ...    
+   }
+
+Files
+-----
+
+The ``Files`` facade provides some powerful methods for working with file objects, and one method of critical importace to the Resource API.
+
+The ``Files.asResource(file)`` method creates a ``ResourceAdapter`` wrapper around an absolute file location. This wrapper is not perfect, but it allows the ease-of-use benefits of the Resource API when working with content outside of the data directory. The definition of a path has been expanded to work with the external locations (with ``Paths.isAbsolute(path)`` and ``Paths.names(path)``).
```
jodygarnett, line 153:
```suggestion
The ``Files.asResource(file)`` method creates a ``ResourceAdapter`` wrapper around an absolute file location. Allows the use of Resource API when working with content outside of the data directory. The definition of a path has been expanded to work with the external locations (with ``Paths.isAbsolute(path)`` and ``Paths.names(path)``).
```

jodygarnett:
> Oh, unrelated issue... I was wondering how to create a directory in the browser, using "new resource" it suggest to type a path in order to create a directory... but if I write "temp" it just creates a file, and if I type "/temp" it does the same... duh?
> Eventually figured out I have to create an actual file in a path, and then the path gets created... weird (habits from the file system, one first creates a directory, and only then create a file, if a file creation right away is wanted at all, that is).

That is a good topic to mention in the developers guide also. 

I wonder if the prompt in the resource browser can be improved to indicate "provide the complete path"? One strange side effect of this is that you cannot have an empty directory in the resource browser or rest api.

aaime:
> I wonder if the prompt in the resource browser can be improved to indicate "provide the complete path"? One strange side effect of this is that you cannot have an empty directory in the resource browser or rest api.

Not being able to have an empty directory is not unseen (git for example) but it's pretty weird given most users are working with the file system storage and there directories can be managed. So I'm guessing that in jdbc-store one cannot store directories? Or it's a limitation at the API level? I was not aware of it, besides, ResourceType includes DIRECTORY so... duh?

jodygarnett:
> Not being able to have an empty directory is not unseen (git for example) but it's pretty weird given most users are working with the file system storage and there directories can be managed. So I'm guessing that in jdbc-store one cannot store directories? Or it's a limitation at the API level? I was not aware of it, besides, ResourceType includes DIRECTORY so... duh?

That is exactly it, the JDBCStore stores resources (blobs) only, and the paths (including each directory along the way) are inferred. So the difference an empty directory, and an undefined path location, end up being the same idea.

aaime:
> That is exactly it, the JDBCStore stores resources (blobs) only, and the paths (including each directory along the way) are inferred. So the difference an empty directory, and an undefined path location, end up being the same idea.

Hum... I checked quickly the Resource Javadoc and I don't see anything about directories being non-creatable in isolation. If I had to judge by that, the behavior you're describing would seem more like a bug in jdbc-store?

jodygarnett:
> Hum... I checked quickly the Resource Javadoc and I don't see anything about directories being non-creatable in isolation. If I had to judge by that, the behavior you're describing would seem more like a bug in jdbc-store?

I was about to come here and say the same thing; you are correct there is a method ``Resource.dir()`` to create an empty directory.

jodygarnett:
rebased and cleaned up history, will merge when completed

PR #6454, opened 2022-12-24
[GEOS-10808] Relax reprojection delta for IEEE differences on M1

jodygarnett:
[GEOS-10808](https://osgeo-org.atlassian.net/browse/GEOS-10808)

The following two tests show differences in floating point implementation on ARM / Apple M1. Switch to using assertEquals(expected,actual,delta) rather than string comparison.

There is no functional change here.

  

# Checklist

### DIFF
--- a/src/wcs2_0/src/test/java/org/geoserver/wcs2_0/WCSTestSupport.java
+++ b/src/wcs2_0/src/test/java/org/geoserver/wcs2_0/WCSTestSupport.java
@@ -54,6 +54,7 @@ import org.geotools.referencing.operation.matrix.XAffineTransform;
 import org.geotools.wcs.v2_0.WCSConfiguration;
 import org.geotools.xsd.Parser;
 import org.junit.After;
+import org.locationtech.jts.geom.CoordinateXY;
 import org.opengis.coverage.Coverage;
 import org.opengis.coverage.grid.GridCoverage;
 import org.opengis.coverage.grid.GridGeometry;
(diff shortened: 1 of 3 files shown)


jodygarnett:
@davidblasby you are the only other developer with an M1 that I know; have you been seeing these failures also? Or just building with an translated intel JVM.

src/wcs2_0/src/test/java/org/geoserver/wcs2_0/WCSTestSupport.java
```diff
@@ -343,6 +347,15 @@ public void cleanCoverages() {
         }
     }
 
+    public static void assertXpathCoordinate(
```
aaime, line 350: Good idea! 
This could be useful elsewhere too... what about pushing it up to GeoServerAbstractTestSupport, with a configurable tolerance? 1E-9 makes some sense for geographic coordinate tests ([mandatory XKCD](https://xkcd.com/2170/)) but it's too strict for projected ones.

jodygarnett: Sure good idea; note I am a little hampered testing locally util this is in okay.

jodygarnett: It fit better in ``GeoServerSystemTestSupport`` along side other DOM manipulation and assert methods.

PR #6455, opened 2022-12-26
[GEOS-10809] Keycloak : add support for usernames with spaces

rafmcpaf:
[GEOS-10809](https://osgeo-org.atlassian.net/browse/GEOS-10809)

When using geoserver secured via keycloak, an error is thrown if the user has a username containing at least one space. 
This PR should fix  this.  

# Checklist


src/community/security/keycloak/src/main/java/org/geoserver/security/keycloak/KeycloakUrlBuilder.java
```diff
@@ -43,7 +47,11 @@ KeycloakUrlBuilder client(String clientId) {
      */
     KeycloakUrlBuilder userByName(String userName) {
         users();
-        sb.append("?exact=true&username=").append(userName);
+        try {
+            sb.append("?exact=true&username=").append(URLEncoder.encode(userName, "UTF-8"));
```
bradh, line 51: You can probably avoid the try / catch if you do something like:

```
sb.append("?exact=true&username=").append(URLEncoder.encode(userName, StandardCharsets.UTF_8));
```

bradh:
Per the checklist, this needs to go into `main` first, and you might need to do a CLA.

src/community/security/keycloak/src/main/java/org/geoserver/security/keycloak/KeycloakUrlBuilder.java
rafmcpaf: Good thinking, I changed it

PR #6485, opened 2023-01-09
[GEOS-10816] OGC API Features complex features test fails since introduction of <meta> tag in HTML templates

aaime:
[GEOS-10816](https://osgeo-org.atlassian.net/browse/GEOS-10816)

See ticket for details

# Checklist

### DIFF
--- a/src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java
+++ b/src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java
@@ -125,6 +125,7 @@ import org.geotools.util.PreventLocalEntityResolver;
 import org.geotools.util.logging.Log4J2LoggerFactory;
 import org.geotools.util.logging.Logging;
 import org.geotools.xsd.XSD;
+import org.jsoup.Jsoup;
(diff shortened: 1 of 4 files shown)


src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java
```diff
@@ -2013,6 +2014,19 @@ protected int countNonBlankPixels(String testName, BufferedImage image, Color bg
         return pixelsDiffer;
     }
 
+    // TODO: consider moving to GeoServerSystemTestSupport
```
taba90, line 2017: isn't indeed moved?

taba90 (changes requested):
Looks good but there is a legit QA failure on gsr ControllerTest class

src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java
aaime: Oh LOL yes, at first I was not sure about moving it and then I did and forgot the comment.

PR #6534, opened 2023-01-23
GEOS-6313 Lifecycle handlers during shutdown

dieterstueken:
[GEOS-6313](https://osgeo-org.atlassian.net/browse/GEOS-6313)

This is another attempt to resolve the shutdown problems of GEOS-6313 which are around since 2014. 

Similar problems are also described by [GEOS-8122]. Those are fixed by manually calling [getGeoServer().dispose()](https://github.com/geoserver/geoserver/commit/849bced5b29c2919310372af42b8d591c17e984f).

This fixed the unit tests but not the source of the problem: Calling `Geoserver.dispose()` from `applicationContext.close()` was too late to acquire the necessary `GeoServerLifecycleHandler` beans and throws an exception. This inhibits calling `catalog.dispose()` due to a missing finally block. This can be fixed easily.

A unit test `GeoServerLifecycleTests` verifies calling the various `GeoServerLifecycleHandler` callbacks.
However there is still a problem with `Geoserver.dispose()`  if it is called _without_ a previous call to `Geoserver.reset()`.
This is exposed by the base `GeoServerLifecycleTest`.

To resolve this, the shutdown procedure is triggered by `ApplicationListener<ContextClosedEvent>` instead of  `DisposableBean` now, to allow to acquire the beans.

# Checklist

### DIFF
--- a/src/main/src/test/java/org/geoserver/GeoServerLifecycleTest.java
+++ b/src/main/src/test/java/org/geoserver/GeoServerLifecycleTest.java
@@ -1,0 +1,79 @@
+/* (c) 2017 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver;
+
+import static org.junit.Assert.assertTrue;
+
+import org.geoserver.config.impl.GeoServerLifecycleHandler;
+import org.geoserver.data.test.SystemTestData;
+import org.geoserver.test.GeoServerSystemTestSupport;
+import org.junit.Test;
+import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
+
+public class GeoServerLifecycleTest extends GeoServerSystemTestSupport {
+
+    @Override
+    protected void onSetUp(SystemTestData data) throws Exception {
+        super.onSetUp(data);
+        ConfigurableListableBeanFactory beanFactory = applicationContext.getBeanFactory();
+        beanFactory.registerSingleton("lifecycleWatcher", new LifecycleWatcher());
+    }
+
+    protected LifecycleWatcher getLifecycleWatcher() {
+        return applicationContext.getBean(LifecycleWatcher.class);
+    }
+
+    @Test
+    public synchronized void testRunning() {
+        assertTrue(applicationContext.isRunning());
+    }
+
+    @Override
+    protected void destroyGeoServer() {
+        LifecycleWatcher watcher = getLifecycleWatcher().reset();
+        super.destroyGeoServer();
+        assertTrue(watcher.didDispose);
+    }
+
+    protected static class LifecycleWatcher implements GeoServerLifecycleHandler {
+
+        boolean didReload = false;
+
+        boolean didBeforeReload = false;
+
+        boolean didReset = false;
+
+        boolean didDispose = false;
+
+        public LifecycleWatcher reset() {
+            didBeforeReload = false;
+            didReload = false;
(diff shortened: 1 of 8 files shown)


jodygarnett (commented):
This looks good to me and makes sense

src/main/src/test/java/org/geoserver/GeoServerLifecycleTest.java
```diff
@@ -0,0 +1,79 @@
+/* (c) 2017 Open Source Geospatial Foundation - all rights reserved
```
mprins, line 1:
```suggestion
/* (c) 2023 Open Source Geospatial Foundation - all rights reserved
```

src/main/src/test/java/org/geoserver/GeoServerLifecycleTests.java
```diff
@@ -0,0 +1,27 @@
+/* (c) 2017 Open Source Geospatial Foundation - all rights reserved
```
mprins, line 1:
```suggestion
/* (c) 2023 Open Source Geospatial Foundation - all rights reserved
```

mprins (changes requested):
the new files should have 2023 as copyright year, apart from that LGTM

src/main/src/test/java/org/geoserver/GeoServerLifecycleTests.java
```diff
@@ -0,0 +1,27 @@
...
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver;
+
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
+
+public class GeoServerLifecycleTests extends GeoServerLifecycleTest {
```
aaime, line 11: Confusing and wasteful:
- Since it extends another class, the tests in the base class are run again (so testRunning is really run twice)
- Having two classes named GeoServerLifecycleTest and GeoServerLifcycleTests is also confusing, does not show what is the reason for the two classes, the explanation provided in the PR will be lost, git history does not retain it.

It seems that what you need is a clean restart of the full class at every method run, which you can get here:

- https://github.com/geoserver/geoserver/blob/main/src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java#L172
I

If that does not solve the issue, the explanation needs to go in the test classes javadoc and better names should be chosen to explain the reason. GeoServer is a long lived project, code must be written so that it's somewhat understandable years down the line, by someone else that would not have been around when this PR was reviewed.

mprins (approved):
What about the CLA? https://docs.geoserver.org/latest/en/developer/policies/committing.html

dieterstueken:
Yes, I realized it.

My current contributions are not directly commissioned by con terra, so an individual CLA may be sufficient.
But it may change in future, so I'm about to ask for a CCLA. After we are separated from ESRI again since a few years, 
we have a raising interest on open source again and I like to support that officially.

If you are curios, I may tell you more details about what what we are currently doing with a GeoServer.
Perhaps on on a separate channel.

jodygarnett:
@dieterstueken please let us know about sending the CLA in. It is available as a PDF (or text) file with you can email to osgeo.

You are welcome to dropping to one of our bi-weekly meetings if you wish to discuss what you are currently doing with GeoServer (our door is always open for collaboration)

aaime:
@jodygarnett CLA has been sent to OSGeo, I was cc'ed.

aaime:
Merged, asked bot to create backports.

PR #6536, opened 2023-01-24
Shutdown.bat: support java path with spaces

jratike80:
The original `for /f %f in ('where java')` truncates "Program Files" into "Program". PR adds "tokens=*"
Fixes https://osgeo-org.atlassian.net/browse/GEOS-7506

### DIFF
--- a/src/release/bin/shutdown.bat
+++ b/src/release/bin/shutdown.bat
@@ -32,7 +32,7 @@ goto shutdown
 
 :trySystemJava
   echo The JAVA_HOME environment variable is not defined, trying to use System Java
-for /f %%i in ('where java') do set RUN_JAVA=%%i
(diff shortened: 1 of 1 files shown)


aaime:
Interesting, never seen options in the batch for loop. Some reference docs:

> tokens=n     Specifies which numbered items to read from each line (default = 1)
> tokens=* will cause all items on each line to be processed.

jratike80:
> never seen options in the batch for loop.
So not taken a look at the startup.bat lately :)

aaime:
Hah, not really... I might even have been the one to add it for what I know, but when it comes to Windows I'm "monkey see, monkey does... monkey forgets!"

aaime:
Thanks, merged, and asked the bot to create backports.

PR #6537, opened 2023-01-24
[GEOS-10837] GeoPackage output locking mode normal

jodygarnett:
[GEOS-10837](https://osgeo-org.atlassian.net/browse/GEOS-10837)

  

# Checklist

### DIFF
--- a/src/extension/geopkg-output/src/main/java/org/geoserver/geopkg/GeoPkg.java
+++ b/src/extension/geopkg-output/src/main/java/org/geoserver/geopkg/GeoPkg.java
@@ -43,12 +43,15 @@ public class GeoPkg {
      */
     public static GeoPackage getGeoPackage(File file) throws IOException {
         SQLiteConfig config = new SQLiteConfig();
-        config.setSharedCache(true);
+        config.setSharedCache(false);
         config.setJournalMode(SQLiteConfig.JournalMode.OFF);
-        config.setPragma(SQLiteConfig.Pragma.SYNCHRONOUS, "OFF");
-        config.setLockingMode(SQLiteConfig.LockingMode.EXCLUSIVE);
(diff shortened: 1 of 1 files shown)


aaime:
Ouch, it was set to exclusive to get performance gains, sqlite can avoid internal locking machinery if it has assurances it can work in isolation. Did you try to write anything significant, gigabytes or more, and compare?

src/extension/geopkg-output/src/main/java/org/geoserver/geopkg/GeoPkg.java
```diff
@@ -43,12 +43,15 @@ public class GeoPkg {
...
     public static GeoPackage getGeoPackage(File file) throws IOException {
         SQLiteConfig config = new SQLiteConfig();
-        config.setSharedCache(true);
+        config.setSharedCache(false);
         config.setJournalMode(SQLiteConfig.JournalMode.OFF);
-        config.setPragma(SQLiteConfig.Pragma.SYNCHRONOUS, "OFF");
-        config.setLockingMode(SQLiteConfig.LockingMode.EXCLUSIVE);
+        config.setSynchronous(SQLiteConfig.SynchronousMode.OFF);
+        config.setLockingMode(SQLiteConfig.LockingMode.NORMAL);
+        config.setTempStore(SQLiteConfig.TempStore.MEMORY);
```
aaime, line 50: This worries me too, this class is shared with the WPS process, which can dump massive amounts of data (and well, the WFS output format can too, it will dump to disk before trying to write on the output, so the HTTP timeout won't save it).

jodygarnett:
Conveying this PR to draft as I wish to test the performance and effectiveness of the change.

jodygarnett:
Closing this attempt as a bad idea (tm)

PR #6538, opened 2023-01-25
Documentation: Filter function is notEqualTo iso notEqual.

emacgillavry:
XML element containing the list of functions differs for the different WFS versions. Also the function to compare values not being equal, the correctfunction name is `notEqualTo` instead of `notEqual`.


jodygarnett (approved):
Thanks!

PR #6539, opened 2023-01-25
Improve homepage performance, fall back on simpler selection modes for large catalogs

aaime:
Based on two separate items:
* Allow the home page controls and description to fall back on [simpler, cheaper to initialize GUI controls](https://osgeo-org.atlassian.net/browse/GEOS-10833)
* [Speed up DefaultResourceAccessManager filters](https://osgeo-org.atlassian.net/browse/GEOS-10838) to speed up the above counts/fills (and along with it, layer preview and capabilities document generation)

[GEOS-10838](https://osgeo-org.atlassian.net/browse/GEOS-10838) makes the current GUI viable when relatively few items are available out of a large set of layers due to security rules. 

[GEOS-10833](https://osgeo-org.atlassian.net/browse/GEOS-10833) instead makes sure that if too many items are available, we can still have a working homepage, getting an acceptable response time (max 5 seconds by default) in exchange for a very simple user experience (plain text boxes, no autofill). 

I did not go for auto-complete text because a full kill-switch was needed anyways, due to the pluggability of both Catalog and ResourceAccessManager there was no way to guarantee that even getting the first item for an autocomplete with timeout would have happened in a suitable time. 

Adding the middle ground UI is still possible extending the enumeration in HomeSelection, if someone wants to go there. The "auto" mode is still implemented in such a way that if the current filter picks a small enough number of items, a dropdown will appear for them. Here is an example from a setup with 1000 workspaces and 40k layers.
The workspace is still a drop-down (matches the maximum in terms of items, still works decently), and when a specific workspace is chosen, then the layer chooser also becomes a drop-down and the description provides counts:

https://user-images.githubusercontent.com/325433/214547915-5e2520c4-ab45-4ec9-b22d-06e1e1e362c0.mp4

About the DefaultResourceAccessManager speed up, it's almost 2 orders of magnitude for complex data directories with many layers and many rules. Tests with a setup having around 20k layers and 40k security rules shows the following for a WMS 1.3 GetCapabilities request:
* 188 seconds with the baseline GeoServer
* 6 seconds with the changes in GeoServer, but without the In function static lookup
* 2.5 seconds with [GEOT-7301: Static lookup for In function](https://osgeo-org.atlassian.net/browse/GEOT-7301)

However, the change might come with a slowdown in some JDBCConfig installations, as the SQL generator for it cannot handle the "in" function (cases with many layers, and most not visible due to security rules). 
As a community module, there is no requirement to care for it, but I still left a system variable to get back the old filter building just in case. @sikeoka @NielsCharlier if you have a large set of layers and restrictive rules that limit access to them, implementing support for encoding the "in" function could prove quite beneficial, in terms of SQL time execution, compared to the current verbose and convoluted security filter. I had a look at doing it, but don't have enough experience with JDBCConfig, nor a good use case to test it against.

# Checklist

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/BoundedCatalogLoader.java
+++ b/src/web/core/src/main/java/org/geoserver/web/BoundedCatalogLoader.java
@@ -1,0 +1,67 @@
+/* (c) 2023 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.web;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+import org.geoserver.catalog.Catalog;
+import org.geoserver.catalog.CatalogInfo;
+import org.geoserver.catalog.util.CloseableIterator;
+import org.opengis.filter.Filter;
+
+/**
+ * Loads {@link org.geoserver.catalog.CatalogInfo} objects from the catalog, with a timeout and a
+ * maximum number of objects to load.
+ */
+class BoundedCatalogLoader<T extends CatalogInfo> implements Serializable {
+
+    List<T> result = new ArrayList<>();
+    boolean boundExceeded = false;
+    long residualTime;
+
+    public BoundedCatalogLoader(T item) {
+        this.result.add(item);
+        this.boundExceeded = false;
+    }
+
+    public BoundedCatalogLoader(
+            Catalog catalog, Filter filter, Class<T> target, long timeout, int max) {
+        // perform a bounded load of the catalog
+        // (single call, doing count + list is slower and count cannot be stopped mid-way)
+        long limit = System.currentTimeMillis() + timeout;
+        try (CloseableIterator<T> search = catalog.list(target, filter, 0, max + 1, null)) {
+            while (System.currentTimeMillis() < limit && search.hasNext()) {
+                result.add(search.next());
+            }
+        }
+        // check if bounds have been exceeded
+        this.residualTime = limit - System.currentTimeMillis();
+        boundExceeded = residualTime <= 0 || result.size() > max;
+        // if went beyond the limit, remove the last elements
+        if (result.size() > max) {
+            // creating new list because sublist is not serializable
+            result = new ArrayList<>(result.subList(0, max));
+        }
+    }
+
+    /** Returns the list of elements loaded from the catalog */
+    public List<T> getResult() {
+        return result;
+    }
+
+    /**
+     * Returns true if the bounds were exceeded (either timed out, or went beyond maximum count),
+     * false otherwise
+     */
+    public boolean isBoundExceeded() {
+        return boundExceeded;
+    }
+
+    /** Returns the residual time, that is the time left after the load operation was completed. */
+    public long getResidualTime() {
+        return residualTime;
+    }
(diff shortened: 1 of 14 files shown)


aaime:
Ah @jodygarnett I forgot, the weird business with admin requests was simplified significantly by fixing a bug in the Wicket callback that was tasked to handle it (but did not do so properly).

src/web/core/src/main/java/org/geoserver/web/BoundedCatalogLoader.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import org.geoserver.catalog.Catalog;
+import org.geoserver.catalog.CatalogInfo;
+import org.geoserver.catalog.util.CloseableIterator;
+import org.opengis.filter.Filter;
+
+/**
+ * Loads {@link org.geoserver.catalog.CatalogInfo} objects from the catalog, with a timeout and a
+ * maximum number of objects to load.
```
jodygarnett, line 17: Really nice approach @aaime

jodygarnett (commented):
Commenting for now; going to test locally next.

jodygarnett:
Testing with the default:
- no problems found

Testing with ``text``
- trying ``sf:roads`` and watched it update workspace and layer
- trying with ``foo:bar`` and the URL updates, but the text fields blank out (so that is good, want a stable URL to share even if intended user does not have permissions yet)

Testing with ``auto``
- ⚠️ Selecting workspace ``ne``, layer ``coastlines`` works. Then changing layer to `` `` (nothing) does not work coastlines remains selected
 - (unrelated) Testing ``nl``, service description respects language setting but welcome message does not.

jodygarnett:
> Ah @jodygarnett I forgot, the weird business with admin requests was simplified significantly by fixing a bug in the Wicket callback that was tasked to handle it (but did not do so properly).

Yes I saw an appreciated the change

jodygarnett:
Andrea I am still having trouble with the selection not changing as described above

aaime:
Your PR got merged and included in the GeoServer code. I've tried this morning from the main branch, it looks like it's working to me, although the resulting URL is sometimes a bit questionable?

Here is a video, running on main, to make it switch to another page I'm either tabbing or pressing enter:

https://user-images.githubusercontent.com/325433/216029744-d3db24fe-0399-462e-94d6-f66ec384a7f7.mp4

jodygarnett:
The video cannot be closed because the file is corrupt. I will try testing again and report back.

aaime:
Weird, Chrome plays it without issues, but Firefox does not.

jodygarnett:
Here is a screen snap after:

1. Change layer to ``sf:sfdem``
2. Page updates to workspace ``sf`` and layer ``sfdem``
3. Change layer to `` `` (empty) 
4. Page updates to workspace ``sf`` and layer `` `` (empty)
5. But URL and page contents remains unchanged ``&workspace=sf&layer=sfdem``

image

Note editing the URL to have ``&workspace=sf`` (removing ``&layer=sfdem``) displayed the correct contents

aaime:
Oh, with the dropdowns.... but the code for the drop downs was just moved around, not modified.. strange?

aaime:
Confirmed it's a regression, does not happen on 2.22.1. Working on a fix.

PR #6540, opened 2023-01-26
[GEOS-10689] OSHISystemInfoCollector holds non daemon threads, prevents clean shutdown of Tomcat

dieterstueken:
[GEOS-10689](https://osgeo-org.atlassian.net/browse/GEOS-10689)

[GEOS-10689]

`OSHISystemInfoCollector`  is a singleton bean to collect system statistics. It creates some `SystemInfo` context and starts a thread to work upon them. A thread is created using an inline lambda expression to implement the collector loop.

This manifests several flaws:

- the implemented loop was taken from [some sample code](https://github.com/oshi/oshi/issues/372) to monitor a _separate_ process. However, in our loop we are observing _our self_. So, the `while (processExists)` loop runs forever.
- the thread is not a _daemon_ thread. Thus the VM will not terminate voluntarily.
- the bean runs on a different thread, so some variables must be volatile.
- the bean must be `Serializable`, thus the objects used by the thread can not be final. Even worse: they have to be transient.
- even if the bean saves its state, the `SystemInfo` context is never restored again.

So, I split it into two different classes: 

- A separate collector thread to hold the necessary `SystemInfo` context and statistics results on its own.
- A bean to manage the status and the life cycle of the thread accordingly.

This way both life cycles are decoupled cleanly and the bean may be a `DisposableBean` now. The only remaining transient value of the bean is the collector thread, which does not harm and is restored on demand.

I decided to create a new class for the bean named `OSHISystemInfoMonitor` and use the former class `OSHISystemInfoCollector` to implement the thread, since it finally performs the actual collection. This way also the history of most of the code remains stable.

I added unit tests to verify that the collector thread was (re)started and is finally stopped.

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/system/status/OSHISystemInfoCollector.java
+++ b/src/main/src/main/java/org/geoserver/system/status/OSHISystemInfoCollector.java
@@ -10,7 +10,6 @@ import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.geotools.util.logging.Logging;
-import org.springframework.stereotype.Component;
 import oshi.SystemInfo;
 import oshi.hardware.CentralProcessor;
 import oshi.hardware.GlobalMemory;
@@ -25,35 +24,31 @@ import oshi.software.os.OperatingSystem;
 /**
  * Retrieve real system information metrics defined in {@link MetricInfo} using OSHI library.
  *
- * <p>It's possible to extend this class to change or add other low level APIs and use this new
- * class as main collector after register using autowiring.
- *
  * @author sandr
  * @see <a href="https://github.com/oshi/oshi">OSHI library </a>
  */
-@Component
-public class OSHISystemInfoCollector extends BaseSystemInfoCollector {
+public class OSHISystemInfoCollector extends Thread {
 
-    private static final long serialVersionUID = 502867203324474735L;
+    public static final int INTERVAL = 1000;
 
-    private static Logger logger = Logging.getLogger(OSHISystemInfoCollector.class);
+    private static final Logger logger = Logging.getLogger(OSHISystemInfoCollector.class);
 
-    private transient OperatingSystem os;
+    private final OperatingSystem os;
 
-    private transient HardwareAbstractionLayer hal;
+    private final HardwareAbstractionLayer hal;
 
-    private transient CentralProcessor pr;
+    private final CentralProcessor pr;
 
-    private transient GlobalMemory mm;
+    private final GlobalMemory mm;
 
-    private transient Sensors ss;
+    private final Sensors ss;
 
-    private transient FileSystem fs;
+    private final FileSystem fs;
 
-    private volatile double cpuUsage = 0;
+    private double cpuUsage = 0;
 
-    private volatile long[][] oldTicks;
-    private volatile long[] oldLoadTicks;
+    private long[][] oldTicks;
+    private long[] oldLoadTicks;
 
     public OSHISystemInfoCollector() {
         SystemInfo si = new SystemInfo();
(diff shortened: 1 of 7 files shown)


dieterstueken:
The second commit adds some tiny (possibly unrelated) unrelated cleanups, i.e. make some variables final.

But I'm quite unhappy with the generated diff of `OSHISystemInfoCollector`. The former lambda expression and the new `run()` method should be side by side. Instead the code of `getMetricValues()` gets shifted apart and shows lots of changes now which are not real! Only very few lines are really changed if aligned right. My Intellij  does a much better job here.

aaime:
QA failed:

```
Error: 2,716 [ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.22.1:check (default) on project gs-main: The following files had format violations:
Error: 2,716 [ERROR]     src/main/java/org/geoserver/system/status/OSHISystemInfoCollector.java
Error: 2,716 [ERROR]         @@ -427,8 +427,7 @@
Error: 2,716 [ERROR]          ····························MetricValue·mv·=·new·MetricValue(info);
Error: 2,717 [ERROR]          ····························mv.setValue(ni.getBytesRecv());
Error: 2,717 [ERROR]          ····························mv.setAvailable(true);
Error: 2,717 [ERROR]         -····························mv.setDescription(
Error: 2,717 [ERROR]         -····································"Network·interface·["·+·ni.getName()·+·"]·received");
Error: 2,717 [ERROR]         +····························mv.setDescription("Network·interface·["·+·ni.getName()·+·"]·received");
Error: 2,717 [ERROR]          ····························mv.setPriority(info.getPriority()·+·(i++)·*·3);
Error: 2,717 [ERROR]          ····························mv.setIdentifier(ni.getName());
Error: 2,717 [ERROR]          ····························si.add(mv);
Error: 2,717 [ERROR] Run 'mvn spotless:apply' to fix these violations.
Error: 2,717 [ERROR] -> [Help 1]
Error: 2,717 [ERROR] 
```

You really need to make a habit of building with Maven before pushing the PR. It sets the right formatting for you
There is only one valid format with Google format, don't try to mimic from the IDE, while you have an option of using it in the IDE too, we are a few versions behind because latest releases require Java 11. We'll upgrade to them once all 3 branches are on Java 11 (that is, in March 2024).

src/main/src/main/java/org/geoserver/system/status/OSHISystemInfoBean.java
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * *
+ *
+ * <p>It's possible to extend this class to change or add other low level APIs and use this new
+ * class as main collector after register using autowiring.
+ *
+ * @author d.stueken
+ */
+@Component
+public class OSHISystemInfoBean extends BaseSystemInfoCollector implements DisposableBean {
```
aaime, line 23: Please don't annotate as @Component, gs-main is not using autowiring and package scan (newer modules do, but when they do, the whole module is auto wired).

Also, please don't introduce the "Bean" naming convention, check the existing applicationContext.xml files to double-check that indeed we do not normally call Spring beans "xyzBean".

I'm not saying we should not do it, but if we did, it would have to be uniform across the project, we don't want each developer to go their merry way.

src/main/src/main/java/org/geoserver/system/status/OSHISystemInfoCollector.java
```diff
@@ -25,516 +21,505 @@
...
 
-    private transient GlobalMemory mm;
+    private long previousTime;
 
-    private transient Sensors ss;
+    private double cpuUsage = 0;
 
-    private transient FileSystem fs;
+    private long[][] oldTicks;
+    private long[] oldLoadTicks;
```
aaime, line 48: Why were some fields re-ordered? It just adds confusion to the diff.

aaime (changes requested):

dieterstueken:
> QA failed:

sorry, I really tried to do it right this time. 
But after 42 tries of rebase and reordering of my commits to make the diff happy, I finally confused myself, too.

aaime:
> > QA failed:
> 
> sorry, I really tried to do it right this time. But after 42 tries of rebase and reordering of my commits to make the diff happy, I finally confused myself, too.

Eh, thank you for trying anyways, not everyone will check out the PR and run diffs locally with a different tool

aaime (approved):

PR #6542, opened 2023-01-27
[GEOS-10726] ogc-api: add encoding tag with charset taken from geoserver settings

NielsCharlier:
[GEOS-10726](https://osgeo-org.atlassian.net/browse/GEOS-10726)

  

# Checklist

### DIFF
--- a/src/community/ogcapi/ogcapi-core/src/main/resources/org/geoserver/ogcapi/common-header.ftl
+++ b/src/community/ogcapi/ogcapi-core/src/main/resources/org/geoserver/ogcapi/common-header.ftl
@@ -7,7 +7,7 @@
       <#elseif model?? && model.htmlTitle?has_content>
         <title>${model.htmlTitle}</title>
       </#if>
-      <meta charset="UTF-8">
+      <meta http-equiv="content-type" content="text/html; charset=${.output_encoding}">
       <link rel="stylesheet" href="${resourceLink("apicss/bootstrap.min.css")}" type="text/css" media="all" />
       <link rel="stylesheet" href="${resourceLink("apicss/geoserver.css")}" type="text/css" media="all" />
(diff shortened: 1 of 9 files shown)


src/community/ogcapi/ogcapi-core/src/main/resources/org/geoserver/ogcapi/common-header.ftl
```diff
@@ -2,6 +2,7 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
+      <meta http-equiv="content-type" content="text/html; charset=${.output_encoding}">
```
aaime, line 5: There is a fixed ``<meta charset="UTF-8">`` [a few lines below](https://github.com/geoserver/geoserver/blob/main/src/community/ogcapi/ogcapi-core/src/main/resources/org/geoserver/ogcapi/common-header.ftl#L10
)... not sure what happens if there are two, but if you did manual testing, that could have interfered.

src/community/ogcapi/ogcapi-tiled-features/src/main/java/org/geoserver/ogcapi/features/tiled/TiledFeaturesExtension.java
```diff
@@ -59,20 +60,20 @@ public TiledFeaturesExtension(FreemarkerTemplateSupport templateSupport) {
     }
 
     @Override
-    public String getExtension(Request dr, Map<String, Object> model, List htmlExtensionArguments)
+    public String getExtension(Request dr, Map model, Charset charset, List htmlExtensionArguments)
```
aaime, line 63: Why was ``Map<String, Object>`` relaxed backj to ``Map``?

aaime (commented):
Overall does not look bad, but some changes look strange... it seems like a change that was done on an older branch and somehow forward ported to main?

NielsCharlier:
Indeed, this still needed updating. I fixed it

PR #6560, opened 2023-02-02
STAC extension: fix item.json path

ktiits:
I believe the template for item.json path was wrong in documentation, it gave 404.


aaime:
Yep, thanks, looks like there was a copy/paste error. Merging.

PR #6561, opened 2023-02-02
[GEOS-10833] GeoServerHomePage unresponsive against large catalogs - fix regression on layer dropdown

aaime:
[GEOS-10833](https://osgeo-org.atlassian.net/browse/GEOS-10833)

@jodygarnett this should fix the layer selection clearing issue you saw. While I was at it, I've also made the URLs better looking (instead of an empty key, the key is removed altogether).

# Checklist

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.java
+++ b/src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.java
@@ -138,7 +138,6 @@ public class GeoServerHomePage extends GeoServerBasePage implements GeoServerUnl
 
     private void homeInit() {
         GeoServer gs = getGeoServer();
-        initFromPageParameters(getPageParameters());
 
         boolean admin = getSession().isAdmin();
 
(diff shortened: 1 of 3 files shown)


jodygarnett:
It appears that initFromPageParameters is already handled by already handled by commonBaseInit:

```
    private void homeInit() {
        GeoServer gs = getGeoServer();
        initFromPageParameters(getPageParameters());
```

jodygarnett (approved):
Andrea this is working well, thank you for the improvement. 

See above comment for a small suggestion (I think we are calling ``initFromPageParameters(getPageParameters());`` twice and it can be removed from ``homeInit()``).

```
    private void homeInit() {
        GeoServer gs = getGeoServer();
        // initFromPageParameters(getPageParameters());
```

jodygarnett:
> How is that related to this change?

I found it while confirming correct handling of page parameters?

aaime:
@jodygarnett good to go?

jodygarnett:
> @jodygarnett good to go?

Yes, I approved earlier. The PR was just something I noticed and not a requirement.

PR #6565, opened 2023-02-03
GEOS-10802 OGC API: queryables page crashed on complex features

NielsCharlier:
[GEOS-10802](https://osgeo-org.atlassian.net/browse/GEOS-10802)

  

# Checklist

### DIFF
--- a/src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/QueryablesBuilder.java
+++ b/src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/QueryablesBuilder.java
@@ -4,6 +4,8 @@
  */
 package org.geoserver.ogcapi;
 
+import static org.geotools.data.complex.util.ComplexFeatureConstants.FEATURE_CHAINING_LINK_NAME;
+
 import io.swagger.v3.oas.models.media.Schema;
 import java.io.IOException;
 import java.util.LinkedHashMap;
@@ -18,8 +20,8 @@ import org.locationtech.jts.geom.MultiPoint;
 import org.locationtech.jts.geom.MultiPolygon;
 import org.locationtech.jts.geom.Point;
 import org.locationtech.jts.geom.Polygon;
-import org.opengis.feature.simple.SimpleFeatureType;
-import org.opengis.feature.type.AttributeType;
+import org.opengis.feature.type.FeatureType;
+import org.opengis.feature.type.PropertyType;
 
 public class QueryablesBuilder {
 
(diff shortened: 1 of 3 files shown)


src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/QueryablesBuilder.java
```diff
@@ -55,15 +56,21 @@ public QueryablesBuilder forType(FeatureTypeInfo ft) throws IOException {
...
+    public QueryablesBuilder forType(FeatureType ft) {
         Map<String, Schema> properties =
-                ft.getAttributeDescriptors().stream()
+                ft.getDescriptors().stream()
+                        .filter(
+                                ad -> // ignore feature chaining links, they might be duplicated
+                                !ad.getName()
+                                                .equals(
+                                                        ComplexFeatureConstants
+                                                                .FEATURE_CHAINING_LINK_NAME))
```
aaime, line 70: Ugly indenting, need to work a bit along with the formatter. For example, a static import on the FEATURE_CHAINING_LINK_NAME, and/or moving the inline comment before ``ad ->``

src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/QueryablesBuilder.java
```diff
@@ -55,15 +56,21 @@ public QueryablesBuilder forType(FeatureTypeInfo ft) throws IOException {
...
                         .collect(
                                 Collectors.toMap(
-                                        ad -> ad.getLocalName(),
+                                        ad -> ad.getName().getLocalPart(),
```
aaime, line 73: Hum... so the queryables are the top-level attributes?
What if one of the attributes is complex though, like holding a feature or a collection? It would not be usable in CQL2.

aaime (changes requested):

src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/QueryablesBuilder.java
NielsCharlier: Fixed formatting issue

src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/QueryablesBuilder.java
NielsCharlier: Admitted, this is merely a fix to get rid of the ugly stack trace, not proper complex feature functionality. The problem however is that it is not possible to provide an exhaustive list of possible xpaths for complex features, because there may be infinite recursion (and there usually is)

aaime: Makes sense, it's a community module after all. I did not mean to have full queriables, more like skipping attributes that cannot work because of their complex nature.

aaime:
Btw, could you add a test for this? We are gearing up to push this module in supported land in the next few months (hopefully!) and already did work to make sure it passes QA and the like.

NielsCharlier:
@aaime sorry for late response. Thanks for the review, a test has been added.

aaime:
Merged, thanks!

PR #6570, opened 2023-02-03
[10705] Loading missing resources creates unsolicited empty files

dieterstueken:
[GEOS-10705](https://osgeo-org.atlassian.net/browse/GEOS-10705)

After merging [GEOS-10723] [GEOS-10724] and [GEOS-10743], we now can realize the idea to throw an `Exception` in `Resource.in()` if the file does not exist. 

Finally `XStreamPersister.load(InputStream)` is affected by this change, too. All usages already have some error handling block. But they are sensitive on `IOException` but not on `IllegalStateException`. So I introduced a new `XStreamPersister.load(Resource)` method which throws an `FileNotFoundException` if the resource is not of `Type.RESOURCE`.
This replaces the boilerplate try/resource blocks and integrates smoothly, since an `IOException` is already expected.

I noticed some cases where the `InputStream` was wrapped by an `BufferedInputStream`, so I used it in general (to be discussed, see also [GSIP 155](https://github.com/geoserver/geoserver/pull/2060#pullrequestreview-18679072)).

I don't have a unit test for the new load method yet, since I was not sure where to place it. But I can provide some easily I think.

# Checklist

### DIFF
--- a/src/platform/src/main/java/org/geoserver/platform/resource/FileSystemResourceStore.java
+++ b/src/platform/src/main/java/org/geoserver/platform/resource/FileSystemResourceStore.java
@@ -201,7 +201,8 @@ public class FileSystemResourceStore implements ResourceStore {
 
         @Override
         public InputStream in() {
-            File actualFile = file();
+            // just take the File as is, don't create it.
+            File actualFile = this.file;
             if (!actualFile.exists()) {
                 throw new IllegalStateException("File not found " + actualFile);
             }
--- a/src/platform/src/main/java/org/geoserver/platform/resource/Files.java
+++ b/src/platform/src/main/java/org/geoserver/platform/resource/Files.java
@@ -77,7 +77,8 @@ public final class Files {
 
         @Override
         public InputStream in() {
-            final File actualFile = file();
+            // just take the File as is, don't create it.
+            final File actualFile = this.file;
             if (!actualFile.exists()) {
                 throw new IllegalStateException("Cannot access " + actualFile);
             }
--- a/src/platform/src/test/java/org/geoserver/platform/resource/ResourceTheoryTest.java
+++ b/src/platform/src/test/java/org/geoserver/platform/resource/ResourceTheoryTest.java
@@ -126,15 +126,15 @@ public abstract class ResourceTheoryTest {
     }
 
     @Theory
-    public void theoryUndefinedHaveIstreamAndBecomeResource(String path) throws Exception {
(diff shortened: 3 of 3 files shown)


src/main/src/main/java/org/geoserver/config/GeoServerLoader.java
```diff
@@ -1096,9 +1096,7 @@ void persist(XStreamPersister xp, Object obj, Resource f) throws Exception {
 
     /** Helper method which uses xstream to depersist an object as xml from disk. */
     <T> T depersist(XStreamPersister xp, Resource f, Class<T> clazz) throws IOException {
-        try (InputStream in = new ByteArrayInputStream(f.getContents())) {
-            return xp.load(in, clazz);
-        }
+        return xp.load(f, clazz);
```
dieterstueken, line 1099: I think the BufferedInputStream does the same job and is even more efficient for large files.

src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
```diff
@@ -725,6 +728,32 @@ public <T> T load(InputStream in, Class<T> clazz) throws IOException {
...
+     * @throws FileNotFoundException if the file does not exist.
+     */
+    public <T> T load(Resource resource, Class<T> clazz) throws IOException {
+        Resource.Type type = resource.getType();
+
+        if (type == Resource.Type.RESOURCE) {
+            try (InputStream in = resource.in()) {
+                // use a BufferedInputStream (if it not already is).
+                InputStream bin =
+                        in instanceof BufferedInputStream ? in : new BufferedInputStream(in, 512);
```
dieterstueken, line 746: the instanceof is not really necessary until no one has the idea to implement it into Resource.in() itself. (See discussion on [GSIP 155](https://github.com/geoserver/geoserver/pull/2060#pullrequestreview-18679072))

dieterstueken:
I just truncated the PR to the first commit only. So, it does not change the Resource API.
I also rebased it onto the current main to exclude any interference with recent changes.
(Hope all checks will pass this time).

Other suggestions proposed by GEOS-10705 to extend the API should go to a separate GISP instead.

jodygarnett (approved):
Thank you this is a clean addition.

You may consider adding a code example clarifying this behaviour to [developers guide documentation](https://docs.geoserver.org/latest/en/developer/programming-guide/config/resource.html).

PR #6585, opened 2023-02-13
[Backport 2.22.x] [GEOS-10851] GWC S3 Blobstore Parameters Get Converted back to plain …

aaime:
[GEOS-10851](https://osgeo-org.atlassian.net/browse/GEOS-10851)

Backport #6582
 **Authored by:** @turingtestfail


jodygarnett:
Appears to be a compilation failure on GWC S4; probably an API change?
```
incompatible types: java.lang.String cannot be converted to java.lang.Integer
```

aaime:
Yes, an API change to allow environment parametrization (specific to the GWC-S3 module). I've tested it locally.

PR #6588, opened 2023-02-13
[Backport 2.21.x] [GEOS-10857] Improve shape-zip output format to handle feature types with multiple geometry columns

aaime:
Backport #6581
 **Authored by:** @aaime


aaime:
Manually tested (relies on new API just introduced in GT 27.x)

PR #6527, opened 2023-01-22
[GEOS-10624] Mapping update

jodygarnett:
[GEOS-10624](https://osgeo-org.atlassian.net/browse/GEOS-10624)

See also:

* [GEOS-10836](https://osgeo-org.atlassian.net/browse/GEOS-10836) The demo styles in "ne" workspace do not validate

Data directory and documentation update, upgrading to 50m features to include disputed area in ``ne:world`` map. This scale also includes internationalized country names allowing demonstration of language function.

The GeoPackage increases from 1 MB to 4.5 MB with no additional processing. Reducing the number of attributes brings this down to 4.3 MB.

The result shows disputed regions in grey, providing a different to the mapcolor9 driven theme.

image
  

Adding ``&LANGUAGE=fr`` to GetMap request provides french country names (ideally we would have a language chooser to work with).

image

# Checklist

- [x] I have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/main/CONTRIBUTING.md).
- [x] I have sent a [Contribution Licence Agreement](https://docs.geoserver.org/latest/en/developer/policies/committing.html) (not required for small changes, e.g., fixing typos in documentation).
- [x] First PR targets the `main` branch (backports managed later; ignore for branch specific issues).
- [ ] All the build checks are green ([see automated QA checks](https://docs.geoserver.org/latest/en/developer/qa-guide/index.html)).

For core and extension modules:

- [ ] New unit tests have been added covering the changes.
- [ ] [Documentation](https://github.com/geoserver/geoserver/tree/main/doc/en/user/source) has been updated (if change is visible to end users).
- [ ] The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers).
- [ ] There is an issue in the [GeoServer Jira](https://osgeo-org.atlassian.net/browse/GEOS/summary) (except for changes that do not affect administrators or end users in any way).
- [x] Commit message(s) must be in the form ``[GEOS-XYZWV] Title of the Jira ticket``.
- [x] Bug fixes and small new features are presented as a single commit.
- [ ] Each commit has a single objective (if there are multiple commits, each has a separate JIRA ticket describing its goal).

PR #6598, opened 2023-02-14
Update countries.sld

jratike80:
Fix order of min/maxdenominator


jodygarnett:
Thanks, included in https://github.com/geoserver/geoserver/pull/6527

PR #6600, opened 2023-02-14
Update boundary_lines.sld

jratike80:
Fix order of min/maxdenominator


jodygarnett:
Included in https://github.com/geoserver/geoserver/pull/6527 thanks!

PR #6604, opened 2023-02-15
[GEOS-10861] Hello world fix

jodygarnett:
[GEOS-10861](https://osgeo-org.atlassian.net/browse/GEOS-10861)

Use InternationalStringUtils to combine contact info welcome and internationalWelcome to reflect the session Locale.

  

# Checklist

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.java
+++ b/src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.java
@@ -154,7 +154,10 @@ public class GeoServerHomePage extends GeoServerBasePage implements GeoServerUnl
 
         Locale locale = getLocale();
 
-        String welcomeText = contactInfo.getWelcome();
+        InternationalString welcome =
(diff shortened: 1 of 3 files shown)


mprins (approved):
lgtm

jodygarnett:
While the test case is passing; the behaviour is a little odd changing between languages. Writing explicit translations for ``fr`` shows up when changing locale selector to ``fr``. Changing it back to `` `` (or ``en``) does not seem to reliably update the screen - although WMS i18n shows no such problem and the code is now using the same technique.

src/web/core/src/main/java/org/geoserver/web/GeoServerHomePage.java
```diff
@@ -154,7 +154,10 @@ private void homeInit() {
 
         Locale locale = getLocale();
```
jodygarnett, line 155: I am trying this line in homeInit() to retrieve the correct locale - however it is being called too earlier?

PR #6605, opened 2023-02-15
[GEOS-3978] Layer configuration allows admin to enter a zero area bound…

Judge73:
[GEOS-3978](https://osgeo-org.atlassian.net/browse/GEOS-3978)

Also fixes problem with running QA build on Windows bacause of forward slashes (/) in path (https://github.com/google/error-prone/issues/2394)

Goal of this task is to add validation to bounding box axis fields so that users could not create boxes with 0 or negative area size. Added validators for min axis fields that check if if value is smaller then max value for that axis.
  

# Checklist


aaime:
@Judge73 is covered by GeoSolutions's CLA

src/web/core/src/main/resources/GeoServerApplication.properties
```diff
@@ -216,6 +216,7 @@ EnvelopePanel.minY  = Min Y
 EnvelopePanel.minZ  = Min Z
 EnvelopePanel.title = Envelope Panel
 EnvelopePanel.crs = Coordinate Reference System
+EnvelopePanel.validation.boundingBoxAxisNonPositive=The provided bouding box boundaries values are invalid! A provided minimum boundary value for axis {0} is higher or equal to the provided maximum boundary value.
```
mprins, line 219:
```suggestion
EnvelopePanel.validation.boundingBoxAxisNonPositive=The provided bounding box boundary values are invalid. A provided minimum boundary value for axis {0} is higher or equal to the provided maximum boundary value.
```

mprins (changes requested):

jodygarnett:
Not clear about:

> QA build on Windows bacause of forward slashes (/) in path

Should that be a distinct PR? Does it need to be back ported ...

src/pom.xml
```diff
@@ -2506,7 +2506,7 @@
               <fork>false</fork>
               <compilerArgs>
                 <arg>-XDcompilePolicy=simple</arg>
-                <arg>-Xplugin:ErrorProne -XepExcludedPaths:${project.build.directory}/generated-sources/.* ${errorProneFlags}</arg>
+                <arg>-Xplugin:ErrorProne -XepExcludedPaths:\Q${project.build.directory}\E.* ${errorProneFlags}</arg>
```
jodygarnett, line 2509: Curious why this PR and not other encountered trouble?

aaime: Because @Judge73 is the first developer to run the build with QA on Windows in years (if not ever). He was trying to get QA running on Windows and I've helped him find a way to address the build failure that came with it.

Judge73: Maybe because other people didn't run QA builds on windows without windows subsystem for linux (guess).

aaime:
@Judge73 seems that not everything was build with maven before issuing the pr (or as an alternative, the IDE is a tad too eager to reformat code, maybe automatically):

```
Error: 4,106 [ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.22.1:check (default) on project gs-web-core: The following files had format violations:
Error: 4,106 [ERROR]     src/test/java/org/geoserver/web/wicket/EnvelopePanelBoundingBoxTest.java
Error: 4,106 [ERROR]         @@ -74,4 +74,4 @@
Error: 4,106 [ERROR]          ········panel.setCRSFieldVisible(true);
Error: 4,106 [ERROR]          ········return·panel;
Error: 4,106 [ERROR]          ····}
Error: 4,106 [ERROR]         -}
Error: 4,106 [ERROR]         +}
Error: 4,106 [ERROR] Run 'mvn spotless:apply' to fix these violations.
Error: 4,106 [ERROR] -> [Help 1]
Error: 4,106 [ERROR] 
```

Judge73:
> @Judge73 seems that not everything was build with maven before issuing the pr (or as an alternative, the IDE is a tad too eager to reformat code, maybe automatically):
> 
> ```
> Error: 4,106 [ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.22.1:check (default) on project gs-web-core: The following files had format violations:
> Error: 4,106 [ERROR]     src/test/java/org/geoserver/web/wicket/EnvelopePanelBoundingBoxTest.java
> Error: 4,106 [ERROR]         @@ -74,4 +74,4 @@
> Error: 4,106 [ERROR]          ········panel.setCRSFieldVisible(true);
> Error: 4,106 [ERROR]          ········return·panel;
> Error: 4,106 [ERROR]          ····}
> Error: 4,106 [ERROR]         -}
> Error: 4,106 [ERROR]         +}
> Error: 4,106 [ERROR] Run 'mvn spotless:apply' to fix these violations.
> Error: 4,106 [ERROR] -> [Help 1]
> Error: 4,106 [ERROR] 
> ```

I'm looking into in. But on my local machine running mvn spotless:check in this module doesn't show any violations.

aaime:
@Judge73 it may be related to windows newlines?
Some of this stuff is bound to pop up as so few GS developers work on Windows, let's try to fix it best as we can.... this document might help: https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

mprins (approved):

Judge73:
> @Judge73 it may be related to windows newlines? Some of this stuff is bound to pop up as so few GS developers work on Windows, let's try to fix it best as we can.... this document might help: https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

Looks like the problem was actually with the code. I've tried couple of steps like change line separator in Intellij and working with git directly instead of using Intellij plugin, but that didn't provide any result.
Plus this file is new and it would be strange if only one specific line had incorrect line endings (because they all were written around same time).

src/web/core/src/main/java/org/geoserver/web/wicket/EnvelopePanel.java
```diff
@@ -115,6 +121,35 @@ void initComponents() {
...
+    private void validateAxis(
+            IValidatable<Double> validatable, DecimalTextField field, String axis) {
+        Double convertedValue = converter.convertToObject(field.getValue(), field.getLocale());
+        if (convertedValue != null && validatable.getValue() >= convertedValue) {
+            ValidationError error = new ValidationError();
+            error.setMessage(
+                    new ParamResourceModel("validation.boundingBoxAxisNonPositive", this, axis)
+                            .getObject());
+            validatable.error(error);
+        }
```
aaime, line 150: Found a weird thing while interactively testing the change. The error message appears twice when I save, even if only one axis had invalid coordinates (and indeed the message is the same):

image

Can you debug and figure out why the message is added twice? I was expecting to see it also in the tests below, but the count seems right there, e.g., 3 errors, 3 messages.

Also strange, if I  start clean and put the error on the Y axis, I'm still receiving a message for the X one: 

image

Oh... found it with a little debugging... the conversion in the validator is performed using the current locale, which for me is italian, where the "." is the thousands separator, not the decimal separator, so I get the wrong number. These fields are not meant to be edited in a locale-dependent way, the number conversion should be done using english locale.

aaime (changes requested):

src/web/core/src/main/java/org/geoserver/web/wicket/EnvelopePanel.java
Judge73: Some clarifiacation from me after debug. The values indeed wasn't converted right, but it works the same on English locale too, since by default, the input fields for bounding boxes use commas as decimal separator (for example when computing) which is invalid for some locales like English and Italian for example and valid for others like Russian and French (it was working fine for me because I forgot to switch from Russian). When panel is created using constructor the fields will have dots as separator (until computed). Plus user can actually input value with dots in those fields. From my point of view (knowing almost nothing about project internals) if input should work with english locale then it shouldn't be allowed to use comma (computed result with dot plus replacing comma with dot when input is submitted for example). But for now I'll just parse both cases.

Judge73: So I checked and if using dot there's another validation after that which doesn't allow that.

src/web/core/src/test/java/org/geoserver/web/wicket/EnvelopePanelBoundingBoxTest.java
```diff
@@ -0,0 +1,93 @@
...
+        final ReferencedEnvelope e =
+                new ReferencedEnvelope3D(180, 180, 90, 90, 20, 20, CRS.decode("EPSG:7415"));
+        tester.startPage(
+                new FormTestPage(
+                        (ComponentBuilder)
+                                id -> {
+                                    EnvelopePanel panel = new EnvelopePanel(id, e);
+                                    panel.setCRSFieldVisible(true);
+                                    return panel;
+                                }));
```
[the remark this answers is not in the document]
Judge73: done

src/web/core/src/test/java/org/geoserver/web/wicket/EnvelopePanelBoundingBoxTest.java
```diff
@@ -0,0 +1,109 @@
...
+import org.apache.wicket.markup.html.form.Form;
+import org.apache.wicket.util.tester.FormTester;
+import org.geoserver.web.FormTestPage;
+import org.geotools.geometry.jts.ReferencedEnvelope;
+import org.geotools.geometry.jts.ReferencedEnvelope3D;
+import org.geotools.referencing.CRS;
+import org.junit.Test;
+import org.opengis.referencing.FactoryException;
+
+public class EnvelopePanelBoundingBoxTest extends EnvelopePanelTest {
```
aaime, line 21: Please don't extend an existing test class, the subclass will just end up running the tests of the base class one more time:

image

src/web/core/src/test/java/org/geoserver/web/wicket/EnvelopePanelBoundingBoxTest.java
```diff
@@ -0,0 +1,109 @@
...
+        and values provided by panel after construction use dot
+        */
+        ft.setValue("panel:maxX", "-59,01");
+        ft.setValue("panel:maxY", "-57,01");
+        ft.setValue("panel:maxZ", "-55,01");
+
+        ft.submit();
+
+        assertEquals(0, tester.getFeedbackMessages(null).size());
+    }
```
aaime, line 102: Tested with italian locale and proposed change above, this behavior mixing two locales was not needed anymore: suggest removing the test.

aaime (changes requested):
I've checked the PR locally again, the current setup seems like it would work only for selected locales. What I did instead is to use Wicket own parsing abilities, in interactive tests it seems to be working fine. See here:

```
diff --git a/src/web/core/src/main/java/org/geoserver/web/wicket/EnvelopePanel.java b/src/web/core/src/main/java/org/geoserver/web/wicket/EnvelopePanel.java
index b08b23b37c..643608ea82 100644
--- a/src/web/core/src/main/java/org/geoserver/web/wicket/EnvelopePanel.java
+++ b/src/web/core/src/main/java/org/geoserver/web/wicket/EnvelopePanel.java
@@ -32,13 +32,6 @@
 public class EnvelopePanel extends FormComponentPanel<ReferencedEnvelope> {
 
     private static final long serialVersionUID = -2975427786330616705L;
-    private static final DecimalFormat DOUBLE_WITH_COMMA = new DecimalFormat();
-
-    static {
-        DecimalFormatSymbols sfs = new DecimalFormatSymbols();
-        sfs.setDecimalSeparator(',');
-        DOUBLE_WITH_COMMA.setDecimalFormatSymbols(sfs);
-    }
 
     protected Label minXLabel, minYLabel, maxXLabel, maxYLabel, minZLabel, maxZLabel;
     protected Double minX, minY, maxX, maxY, minZ, maxZ;
@@ -148,10 +141,10 @@ private void addBoundingBoxValidators() {
      */
     private void validateAxis(
             IValidatable<Double> validatable, DecimalTextField maxValueField, String axis) {
-        String value = maxValueField.getValue();
-        // Skipping dot because it's not valid input and will be validated further
-        if (!value.contains(".")) {
-            Double convertedValue = NumberUtils.parseNumber(value, Double.class, DOUBLE_WITH_COMMA);
+        // if max is not valid, form will fail anyway, and there is nothing to compare against
+        if (maxValueField.isValid()) {
+            maxValueField.convertInput();
+            Double convertedValue = maxValueField.getConvertedInput();
             if (convertedValue != null && validatable.getValue() >= convertedValue) {
                 ValidationError error = new ValidationError();
                 error.setMessage(
```

Can you please double check with the extra locales you have access to and verify it's working properly?

src/web/core/src/test/java/org/geoserver/web/wicket/EnvelopePanelBoundingBoxTest.java
Judge73: Removed

src/web/core/src/test/java/org/geoserver/web/wicket/EnvelopePanelBoundingBoxTest.java
Judge73: Switched parent to GeoServerWicketTestSupport

aaime (approved):

PR #6607, opened 2023-02-15
[GEOS-10837] geopackage output fails when java.io.tmpdir on network share

davidblasby:
[GEOS-10837](https://osgeo-org.atlassian.net/browse/GEOS-10837)

This allows the user to set a property (i.e. `-Dgeopackage.getfeatureoutput.tempdir=/path/to/temp`) that uses an alternative temp dir for the geopackage.  For people with the standard java temp dir being a mounted network drive, they should use this (pointing to a local drive).

GeoPKG (SQL Lite) has a lot of problems locking when writing to cloud (or network-mounted) file systems.

NOTE: I moved the test cases into the "proper" package (to match the class-under-test `GeoPackageGetFeatureOutputFormat`).
  

# Checklist

### DIFF
--- a/src/extension/geopkg-output/src/main/java/org/geoserver/geopkg/wfs/GeoPackageGetFeatureOutputFormat.java
+++ b/src/extension/geopkg-output/src/main/java/org/geoserver/geopkg/wfs/GeoPackageGetFeatureOutputFormat.java
@@ -47,7 +47,7 @@ public class GeoPackageGetFeatureOutputFormat extends WFSGetFeatureOutputFormat 
     protected static Logger LOGGER = Logging.getLogger(GeoPackageGetFeatureOutputFormat.class);
 
     public static final String PROPERTY_INDEXED = "geopackage.wfs.indexed";
-    public static final String CUSTOM_TEMP_DIR_PROPERTY = "geopackage.getfeatureoutput.tempdir";
+    public static final String CUSTOM_TEMP_DIR_PROPERTY = "geopackage.wfs.tempdir";
 
     public GeoPackageGetFeatureOutputFormat(GeoServer gs) {
         super(gs, Sets.union(Sets.newHashSet(MIME_TYPES), Sets.newHashSet(NAMES)));
--- a/src/extension/geopkg-output/src/main/java/org/geoserver/geopkg/wfs/GeoPackageGetFeatureOutputFormat.java
+++ b/src/extension/geopkg-output/src/main/java/org/geoserver/geopkg/wfs/GeoPackageGetFeatureOutputFormat.java
@@ -18,11 +18,13 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.List;
+import java.util.logging.Logger;
 import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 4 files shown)


src/extension/geopkg-output/src/main/java/org/geoserver/geopkg/wfs/GeoPackageGetFeatureOutputFormat.java
```diff
@@ -71,14 +78,42 @@ protected String getExtension(FeatureCollectionResponse response) {
...
+     * @return
+     */
+    File createTempFile(String prefix, String suffix) throws IOException {
+        String customTempDir = GeoServerExtensions.getProperty(CUSTOM_TEMP_DIR_PROPERTY);
+        if (!StringUtils.hasText(customTempDir)) {
+            return File.createTempFile("geopkg", ".tmp.gpkg");
+        }
+        File tempDir = new File(customTempDir);
+        if (!tempDir.exists() || !tempDir.isDirectory()) {
+            throw new IOException(
```
jodygarnett, line 101:
```suggestion
            throw new FileNotFoundException(
```

Will need to adjust import also

src/extension/geopkg-output/src/main/java/org/geoserver/geopkg/wfs/GeoPackageGetFeatureOutputFormat.java
```diff
@@ -71,14 +78,42 @@ protected String getExtension(FeatureCollectionResponse response) {
...
+     *
+     * <p>If the CUSTOM_TEMP_DIR_PROPERTY property is set, then use that as the temp directory.
+     * Otherwise (unset), use the standard File.createTempFile() temp directory.
+     *
+     * <p>If the CUSTOM_TEMP_DIR_PROPERTY directory doesn't exist (or isnt a directory), then an
+     * exception is thrown
+     *
+     * @param prefix
+     * @param suffix
+     * @return
```
jodygarnett, line 92:
```suggestion
     * @param prefix filename prefix, must be three characters or more
     * @param suffix filename suffix, or if null {@code null} suffix {@code ".tmp") used
     * @return path to newly created temporary file
```

mprins:
I'd rather use `java.io.tmpdir` or  `CATALINA_TMPDIR` instead of adding a property that only fixes this for one format.

jodygarnett:
> I'd rather use `java.io.tmpdir` or `CATALINA_TMPDIR` instead of adding a property that only fixes this for one format.

This is explicitly to avoid a ``java.io.tmpdir`` setting which as been set to a network drive. Apparently that is an acceptable approach when running some kind of docker image with hardly any local disk.

This approach is somewhat similar to how WPS has a resource storage directory.

mprins:
OK, I knew it had to be something exotic 😺

jodygarnett:
@davidblasby I am taking this out of draft. The change you have done looks clean; and I do understand why it is required. I am going to add a note for the documentation (which is my only feedback).

jodygarnett (changes requested):
Only useful feedback is to add a note to the docs which I will do now before merge and backport. Thank you for the improvement.

PR #6608, opened 2023-02-16
GEOS-10500 WFS-T unable to delete more than 30 features in a single t…

Judge73:
[GEOS-10500](https://osgeo-org.atlassian.net/browse/GEOS-10500)

[GEOS-10500 WFS-T unable to delete more than 30 features in a single transaction when the data source is PostGIS]

Deleting reasonably large batch of features (in this case 30) took enormous amounts of time
due to the way filters are aggregated on deletion. A new filter is added each time, and if added filter has nested filters (children) such process creates filters with deeply nested structure.
Example:
[[ 30 ] OR [[ 29 ] OR [[ 28 ] OR [[ 27 ] OR [[ 26 ] OR [[ 25 ] OR [[ 24 ] OR [[ 23 ] OR [[ 22 ] OR [[ 21 ] OR [[ 20 ] OR [[ 19 ] OR [[ 18 ] OR [[ 17 ] OR [[ 16 ] OR [[ 15 ] OR [[ 14 ] OR [[ 13 ] OR [[ 12 ] OR [[ 11 ] OR [[ 10 ] OR [[ 9 ] OR [[ 8 ] OR [[ 7 ] OR [[ 6 ] OR [[ 5 ] OR [[ 4 ] OR [[ 3 ] OR [[ 2 ] OR [ 1 ]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]

Changes were added to join all "children" on one layer instead of creating nested structure.

# Checklist

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/request/Delete.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/request/Delete.java
@@ -16,6 +16,7 @@ import org.geotools.factory.CommonFactoryFinder;
 import org.geotools.xsd.EMFUtils;
 import org.opengis.filter.Filter;
 import org.opengis.filter.FilterFactory;
+import org.opengis.filter.Or;
 
 /**
  * Delete element in a Transaction request.
@@ -80,12 +81,9 @@ public abstract class Delete extends TransactionElement {
         Filter currentFilter = (Filter) EMFUtils.get(obj, property);
 
         List<Filter> filters = new ArrayList<>();
-        if (newFilter != null) {
-            filters.add(newFilter);
-        }
-        if (currentFilter != null) {
-            filters.add(currentFilter);
-        }
+
+        flattenFilter(newFilter, filters);
+        flattenFilter(currentFilter, filters);
 
         Filter result;
         if (filters.isEmpty()) {
@@ -100,6 +98,16 @@ public abstract class Delete extends TransactionElement {
             ((DeleteElementTypeImpl) obj).setFilter(result);
         } else if (obj instanceof DeleteTypeImpl) {
             ((DeleteTypeImpl) obj).setFilter(result);
+        }
+    }
+
+    private void flattenFilter(Filter filter, List<Filter> filters) {
+        if (filter != null) {
+            if (filter instanceof Or) {
+                filters.addAll(((Or) filter).getChildren());
+            } else {
+                filters.add(filter);
+            }
         }
     }
 }
--- a/src/wfs/src/test/java/org/geoserver/wfs/BatchManagerTest.java
+++ b/src/wfs/src/test/java/org/geoserver/wfs/BatchManagerTest.java
@@ -177,6 +177,44 @@ public class BatchManagerTest {
         testAggregationWithDeleteBatchSize(-1);
     }
 
+    @Test
+    public void testLargeBatchDeletion() {
+
+        final int batchSizeForDeletion = 40;
+
+        Map<TransactionElement, TransactionElementHandler> element2Handlers = new LinkedHashMap<>();
+
+        Delete delete1 = newDelete(type1Name, filter1);
+        TransactionElementHandler delete1Handler = mock(TransactionElementHandler.class);
+
+        element2Handlers.put(delete1, delete1Handler);
+
+        for (int i = 1; i < batchSizeForDeletion; i++) {
+            Delete delete = newDelete(type1Name, filter1);
+            TransactionElementHandler deleteHandler = mock(TransactionElementHandler.class);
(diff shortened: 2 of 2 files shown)


aaime:
A clarification, the filter as set-up is ugly and clumsy, but the reason why it's taking so much time is in a filter visitor that has poor performance dealing with this type of nested filter. Changing the way the filter is built is the simplest approach to address this issue, although something remains down in GeoTools. @Judge73 could you also open a Jira ticket against that GeoTools class that was spending much time processing this filter?

mprins:
Not sure if it's the same problem; @matthijsln recently discovered a problem where he ran into a problem where `JDBCFeatureSource.splitFilter()` hangs with `O(n^2)` execution time in `PostPreProcessFilterSplittingVisitor`

There is a testcase in https://github.com/B3Partners/tailormap-api/pull/283 which breaks at about 26 `OR`s in an ECQL

I think we worked around this creating an `IN` instead

aaime:
@mprins I believe it was in another class but can't recall the name... @Judge73 do you still have it handy?

src/wfs/src/main/java/org/geoserver/wfs/request/Delete.java
```diff
@@ -102,4 +97,20 @@ protected void eAddForDelete(EObject obj, String property, Filter newFilter) {
...
+        if (newFilter != null) {
+            filters.add(newFilter);
+        }
+
+        if (currentFilter != null) {
+            if (currentFilter instanceof BinaryLogicOperator) {
+                filters.addAll(((BinaryLogicOperator) currentFilter).getChildren());
+            } else {
+                filters.add(currentFilter);
+            }
```
aaime, line 112: Check seems too broad... the filters in the list are re-combined with or, so it should be checking for a "or" filter, rather than BinaryLogicOperator (which covers And as well):

image

src/wfs/src/test/java/org/geoserver/wfs/BatchManagerTest.java
```diff
@@ -177,6 +177,45 @@ public void testAggregationNoDeleteAgg() throws Exception {
         testAggregationWithDeleteBatchSize(-1);
     }
 
+    @Test
+    public void testLargeBatchDeletion() {
+
+        final int batch_size_for_deletion = 40;
```
aaime, line 183: Our coding convention uses underscores only for constants. So either:
``batchSizeForDeletion``
or
``BATCH_SIZE_FOR_DELETION`` (since it's effectively a constant)

Judge73:
> @mprins I believe it was in another class but can't recall the name... @Judge73 do you still have it handy?
Only thing left is this screenshot.
stack

aaime:
That's enough @Judge73 , it seems to be happening inside NullHandlingVisitor. Can you open a Jira ticket against GeoTools for that one?

Judge73:
> [[ 30 ] OR [[ 29 ] OR [[ 28 ] OR [[ 27 ] OR [[ 26 ] OR [[ 25 ] OR [[ 24 ] OR [[ 23 ] OR [[ 22 ] OR [[ 21 ] OR [[ 20 ] OR [[ 19 ] OR [[ 18 ] OR [[ 17 ] OR [[ 16 ] OR [[ 15 ] OR [[ 14 ] OR [[ 13 ] OR [[ 12 ] OR [[ 11 ] OR [[ 10 ] OR [[ 9 ] OR [[ 8 ] OR [[ 7 ] OR [[ 6 ] OR [[ 5 ] OR [[ 4 ] OR [[ 3 ] OR [[ 2 ] OR [ 1 ]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]



> That's enough @Judge73 , it seems to be happening inside NullHandlingVisitor. Can you open a Jira ticket against GeoTools for that one?

Done https://osgeo-org.atlassian.net/browse/GEOT-7309

src/wfs/src/main/java/org/geoserver/wfs/request/Delete.java
```diff
@@ -102,4 +97,20 @@ protected void eAddForDelete(EObject obj, String property, Filter newFilter) {
...
+            filters.add(newFilter);
+        }
+
+        if (currentFilter != null) {
+            if (currentFilter instanceof Or) {
+                filters.addAll(((Or) currentFilter).getChildren());
+            } else {
+                filters.add(currentFilter);
+            }
+        }
```
aaime, line 113: I'd suggest to make this symmetrical, and apply the same logic to newFilter as well, just in case.
So the method could become a "addFilters(Filter, List<Filter> filters)" and called in sequence with both filter and currentFilter.

PR #6623, opened 2023-02-23
GEOS-10611 Uploading application/zip to styles endpoint does not clea…

Judge73:
[GEOS-10611](https://osgeo-org.atlassian.net/browse/GEOS-10611)

https://osgeo-org.atlassian.net/browse/GEOS-10611

Uploading styles using zip file didn't remove temp files after request is ended.
Added deletion of temp dir and files in the end of request. Initially deletion didn't work because multiple streams are created, but not closed during request (had to find and close them).
Added checking if files were deleted to [testPostToWorkspaceSLDPackage](https://github.com/Judge73/geoserver/blob/c0d3671f9f503a910c57ea1ee73d79e962994f63/src/restconfig/src/test/java/org/geoserver/rest/catalog/StyleControllerTest.java#L1259)
  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/SLDHandler.java
+++ b/src/main/src/main/java/org/geoserver/catalog/SLDHandler.java
@@ -153,7 +153,7 @@ public class SLDHandler extends StyleHandler {
             EntityResolver entityResolver)
             throws IOException {
         if (version == null) {
-            Object[] versionAndReader = getVersionAndReader(input);
+            Object[] versionAndReader = getVersionAndReader(input, true);
             version = (Version) versionAndReader[0];
             input = versionAndReader[1];
         }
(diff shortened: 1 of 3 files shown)


src/platform/src/main/java/org/geoserver/util/IOUtils.java
```diff
@@ -373,24 +373,27 @@ public static File getZipOutputFile(File destDir, ZipEntry entry) throws IOExcep
     }
 
     public static void decompress(InputStream input, File destDir) throws IOException {
```
aaime, line 375: Generally the PR looks good, but this change needs to be reverted, the responsibility of closing the stream provided as a parameter is on the caller of this method. Also, if the inputStream has been provided by a servelet, it should not be closed at all, it's the duty of the web container to do so

Judge73: Reverted

aaime:
Looks good, merged

PR #6630, opened 2023-02-24
[GEOS-10871] Force use of AntRun plugin 3.1.0

jusabatier:
[GEOS-10871](https://osgeo-org.atlassian.net/browse/GEOS-10871)

This PR force the use of the 3.1.0 version of AntRun plugin.

As in last version <tasks> were replaced with <task>, if we don't force the version it use the old 1.3 that doesn't support <tasks>. 

It results for example in @project.version@ not replaced in some web/core properties files.

Solve : https://osgeo-org.atlassian.net/browse/GEOS-10871

### DIFF
--- a/src/community/netcdf-ghrsst/pom.xml
+++ b/src/community/netcdf-ghrsst/pom.xml
@@ -72,6 +72,7 @@
     <plugins>
       <plugin>
         <artifactId>maven-antrun-plugin</artifactId>
+        <version>3.1.0</version>
         <executions>
           <execution>
             <id>process-resources</id>
--- a/src/community/release/pom.xml
+++ b/src/community/release/pom.xml
@@ -443,6 +443,7 @@
       </plugin>
       <plugin>
         <artifactId>maven-antrun-plugin</artifactId>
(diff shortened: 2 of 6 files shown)


bradh:
Instead of duplicating the version number, can you use a property, or even better, put it in managed dependencies?

mprins (changes requested):
Please add the m-a-p with version  to the pluginManagement section of the parent pom file and remove the version element from any m-a-p in build sections

mprins:
should fix [GEOS-10871](https://osgeo-org.atlassian.net/browse/GEOS-10871)

jodygarnett:
When’s e get this sorted I will cherry-pick the change against the 2.22.2 and redeploy.

jodygarnett:
Okay I noticed this was branching from 2.22.x rather than ``main``. Going to cherry pick the commit and clean things up. Thanks for getting this activity started!

jodygarnett:
Please see https://github.com/geoserver/geoserver/pull/6635 for PR against ``main``. I also amended your commit to follow our naming convention which helps when back porting.

PR #6643, opened 2023-02-27
[GEOS-10868] Customize FeatureType with editable description

turingtestfail:
[GEOS-10868](https://osgeo-org.atlassian.net/browse/GEOS-10868)

* editable descriptor for customize feature type table
* unit tests, null enforcement, callback
* spotless cleanup
* documentation
* copyright
* missed a test

  

# Checklist


src/main/src/main/java/org/geoserver/catalog/impl/AttributeTypeInfoImpl.java
```diff
@@ -26,6 +26,7 @@ public class AttributeTypeInfoImpl implements AttributeTypeInfo {
     protected Class<?> binding;
     protected Integer length;
     protected String source;
+    protected String description;
```
aaime, line 29: Should be part of hashCode and equalsIgnoreFeatureType

src/web/core/src/test/java/org/geoserver/web/data/resource/ResourceConfigurationPageTest.java
```diff
@@ -926,7 +927,7 @@ public void testCustomizeAttributes() {
         form.setValue(formEdit1 + "2:component:text", "abstract");
         String cql = "Concatenate(description, ' and more!')";
         form.setValue(formEdit1 + "4:component:area", cql);
```
aaime, line 929: Could you customize the description and check the update worked below, in "check saving happened"?

src/main/src/test/java/org/geoserver/catalog/TransformFeatureTypeCallbackTest.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import org.geoserver.catalog.impl.FeatureTypeInfoImpl;
+import org.geoserver.test.GeoServerSystemTestSupport;
+import org.geotools.data.simple.SimpleFeatureSource;
+import org.junit.Before;
+import org.junit.Test;
+import org.opengis.feature.simple.SimpleFeatureType;
+
+public class TransformFeatureTypeCallbackTest extends GeoServerSystemTestSupport {
```
aaime, line 21: Good to have, but I'd also like to see a functional test, for example, customize a property in the various DescribeFeatureTypeTest and checks it's actually producing the desired output.

aaime (changes requested):
Manually tested, looks good. See comments below.

src/web/core/src/test/java/org/geoserver/web/data/resource/ResourceConfigurationPageTest.java
turingtestfail: done

src/main/src/main/java/org/geoserver/catalog/impl/AttributeTypeInfoImpl.java
```diff
@@ -152,6 +153,16 @@ public void setSource(String source) {
         this.source = source;
     }
 
+    @Override
+    public String getDescription() {
```
jodygarnett, line 157: Prefer if this was InternationalString right from the start; indicates human readable text.

src/web/core/src/main/java/org/geoserver/web/data/resource/AttributeTypeInfoEditor.html
```diff
@@ -11,6 +11,9 @@
     <wicket:fragment wicket:id="area">
         <textarea wicket:id="area" class="cql"></textarea>
     </wicket:fragment>
+    <wicket:fragment wicket:id="description">
+        <textarea wicket:id="description" class="cql"></textarea>
+    </wicket:fragment>
```
jodygarnett, line 16: Making the table wider and wider is not working out so well. Description in particular invites an entire sentence to be written (if not a paragraph). 

I do not have a good/cheap suggestion on how to address this feedback - any ideas?

jodygarnett: Is this really ``class="cal"`` should it not be text of some form?

src/main/src/main/java/org/geoserver/catalog/TransformFeatureTypeCallback.java
```diff
@@ -72,7 +73,8 @@ private Definition toDefinition(AttributeTypeInfo ati) {
             String name = ati.getName();
             Expression source = ECQL.toExpression(ati.getSource());
             Class<?> binding = ati.getBinding();
-            return new Definition(name, source, binding);
+            String description = StringUtils.trimToNull(ati.getDescription());
+            return new Definition(name, source, binding, null, description);
```
jodygarnett, line 77: GeoTools AttributeTypeInfo.getDescription() is already working with an InternationalString here.

jodygarnett (commented):
Commenting for feedback only.

src/web/core/src/main/java/org/geoserver/web/data/resource/AttributeTypeInfoEditor.html
aaime: We don't have resources to make a better editor, but something radically different will be needed in case international descriptions are actually used (thinking out loud, a pop-up editor of some sort, or remove the table based editor and go for a list-based one instead).

We can go as far as using InternationalString in the API, changing the editor and making actual usage of international strings can be implemented later by parties having both interest and funds. This will involve also figuring out the current language in the DescribeFeatureType response and act accordingly, that is also not currently internationalized.

jodygarnett: Thanks, changing the api is a good step (and good communication of intent).

jodygarnett:
If this can get in tomorrow for the 2.23-RC it would be ideal

src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
```diff
@@ -2710,6 +2718,19 @@ public Object doUnmarshal(
         }
     }
 
+    class SimpleInternationalStringConverter extends AbstractSingleValueConverter {
```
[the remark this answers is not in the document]
aaime: The other international strings do not have such problem, but I've found the implemnetation is always using GrowableInternationalString for the fields, and performing conversions on the fly:

 ```
this.internationalTitle = InternationalStringUtils.growable(internationalTitle);
```

So I guess the "class" attribute is not there because the implementation class is fixed.

**Review by aaime — CHANGES_REQUESTED** [review]
Works fine, but I've discovered an annoying discrepancy in the REST representation of description vs all the other international strings managed by GeoServer

File: src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java (hunk 7 above)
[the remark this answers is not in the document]
Comment (turingtestfail) [reply src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java]: switched to growable and original converter

File: src/main/src/main/java/org/geoserver/catalog/AttributeTypeInfo.java [hunk 8]
```diff
@@ -99,4 +101,18 @@ public interface AttributeTypeInfo extends Serializable {
      * @param source A valid CQL expression, in the simple case, the attribute name
      */
     void setSource(String source);
+
+    /**
+     * Get Attribute description
+     *
+     * @return Attribute description
+     */
+    GrowableInternationalString getDescription();
```
aaime, line 110:
```suggestion
    InternationalString getDescription();
```

We just need the implementation to use GrowableInternationalString, XStream reflects directly on the fields of the Impl class, does not work against bean properties.

src/main/src/main/java/org/geoserver/catalog/impl/AttributeTypeInfoImpl.java
```diff
@@ -26,6 +29,7 @@ public class AttributeTypeInfoImpl implements AttributeTypeInfo {
     protected Class<?> binding;
     protected Integer length;
     protected String source;
+    protected GrowableInternationalString description;
```
aaime, line 32: This is what XStream is really using

aaime (approved):

PR #6665, opened 2023-03-09
[GEOS-10882] Add an option to remove trailing slash match in OGC APIs

turingtestfail:
[GEOS-10882](https://osgeo-org.atlassian.net/browse/GEOS-10882)

cleanup

testing

global settings test

documentation

moved in page

documentation and move to others

cleanup

remove dispatcher test

included restart note

restconfig test fixes

is online issue

undid changes

test fixes

more tests

more test cleanup

more test cleanup

cleanup

cleanup

  

# Checklist


src/main/src/main/java/org/geoserver/config/GeoServerInfo.java
```diff
@@ -120,6 +120,10 @@ public interface GeoServerInfo extends Info {
      */
     Boolean isXmlExternalEntitiesEnabled();
 
+    void setTrailingSlashMatch(Boolean trailingSlashMatch);
+
+    Boolean isTrailingSlashMatch();
+
```
aaime, line 126: Javadocs missing

src/main/src/main/java/org/geoserver/config/impl/GeoServerInfoImpl.java
```diff
@@ -50,6 +50,8 @@ public class GeoServerInfoImpl implements GeoServerInfo {
 
     protected Boolean xmlExternalEntitiesEnabled = Boolean.FALSE;
 
+    protected Boolean trailingSlashMatch = Boolean.FALSE;
```
aaime, line 53: Should be true by default, we might decide to set it to false down the line, in preparation for a Spring 6 upgrade (where doing a trailing slash match will require more than just configuration)

src/rest/src/main/java/org/geoserver/rest/RestConfiguration.java
```diff
@@ -208,6 +211,13 @@ public void configurePathMatch(PathMatchConfigurer configurer) {
         helper.setAlwaysUseFullPath(true);
         configurer.setUrlPathHelper(helper);
         configurer.setUseSuffixPatternMatch(true);
+        if (geoServer != null
+                && geoServer.getGlobal() != null
+                && geoServer.getGlobal().isTrailingSlashMatch() != null) {
+            configurer.setUseTrailingSlashMatch(geoServer.getGlobal().isTrailingSlashMatch());
+        } else {
+            configurer.setUseTrailingSlashMatch(true);
+        }
```
aaime, line 220:
```suggestion
        configurer.setUseTrailingSlashMatch(Optional.ofNullable(geoServer).map(g -> g.getGlobal()).map(g -> g.isTrailingSlashMatch()).orElse(true);
```

src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/APIDispatcher.java
```diff
@@ -133,6 +146,8 @@ public UrlPathHelper getUrlPathHelper() {
         APIConfigurationSupport configurationSupport =
                 context.getBean(APIConfigurationSupport.class);
         configurationSupport.setCallbacks(callbacks);
+        PathMatchConfigurer configurer = new PathMatchConfigurer();
+        configurationSupport.configurePathMatch(configurer);
```
aaime, line 150: Why is this needed?

src/web/core/src/main/resources/GeoServerApplication.properties
```diff
@@ -348,6 +348,7 @@ GlobalSettingsPage.title             = Global Settings
 GlobalSettingsPage.verboseExceptions = Include stack trace in service exception
 GlobalSettingsPage.verboseMessaging  = Verbose XML output (pretty print) 
 GlobalSettingsPage.xmlExternalEntitiesEnabled = Unrestricted XML External Entity Resolution (security risk)
+GlobalSettingsPage.trailingSlashMatch = Match URLs with trailing slash (Requires restart)
```
aaime, line 351: Nice that you added it to the label, as it's unusual. The screenshot is missing that bit though.

aaime (changes requested):
Overall looks good, found just a few minor bits to comment on.
The default should be that we match trailing slash though, at least for now, we'll see what to do about it when we upgrade to Spring 6.

Changing all the tests so that they don't use a trailing match is probably a good idea, as they work with both settings, and will avoid having lots of overhead during an upgrade.
I'd leave just one, to make sure the current code can, by default, match trailing slashes.

Can we also have a test of flipping the behavior? GeoServer rebuilds the dispatchers at spring startup, so we need the test to be isolated from others.... the test should leverage data dir setup in order to change the config files before the dispatchers are built, I suppose, it can be done overriding setUpTestData in a GeoServerSystemTestSupport .

A possible alternative would be to have the default value (both in the field declaration, and in the setter) be controlled by a system variable (dynamic lookup, not a static final one) that you could flip on/off during the test, on a data dir that has the flag set to null.

src/main/src/main/java/org/geoserver/config/impl/GeoServerInfoImpl.java
aaime: This could be controlled by a system variable, e.g. ``Boolean.getValue(System.getProperty("myProperty", "true))``

src/main/src/main/java/org/geoserver/config/impl/GeoServerInfoImpl.java
```diff
@@ -281,6 +283,16 @@ public Boolean isXmlExternalEntitiesEnabled() {
...
     }
 
+    @Override
+    public void setTrailingSlashMatch(Boolean trailingSlashMatch) {
+        this.trailingSlashMatch = trailingSlashMatch;
+    }
+
+    @Override
+    public Boolean isTrailingSlashMatch() {
+        return trailingSlashMatch == null ? Boolean.TRUE : trailingSlashMatch;
```
aaime, line 293: And then have this Boolean.TRUE use ``Boolean.getValue(System.getProperty("myProperty", "true))`` as well (move that one-liner in a utility method)

src/main/src/main/java/org/geoserver/config/GeoServerInfo.java
turingtestfail: fixed

src/main/src/main/java/org/geoserver/config/impl/GeoServerInfoImpl.java
turingtestfail: changed

src/rest/src/main/java/org/geoserver/rest/RestConfiguration.java
turingtestfail: changed

src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/APIDispatcher.java
turingtestfail: removed

src/web/core/src/main/resources/GeoServerApplication.properties
turingtestfail: screenshot updated

src/main/src/main/java/org/geoserver/config/impl/GeoServerInfoImpl.java
turingtestfail: changed

aaime (approved):

PR #6723, opened 2023-03-27
Update cog.rst

goxzy:
Environment variables that should be defined when connecting to S3-compatible service run by a third party using  it.geosolutions.imageioimpl.plugins.cog.S3RangeReader


aaime:
Looks good to me, thank you!

PR #6755, opened 2023-04-05
Java 11 minimum for running GeoServer

jodygarnett:
# Checklist

### DIFF
--- a/src/release/src/markdown/RUNNING.txt
+++ b/src/release/src/markdown/RUNNING.txt
@@ -12,7 +12,7 @@ In order to install and run this application, follow these steps:
 
 1. Download and install a Java Development Kit
 
-    * Download a Java Runtime Environment (Java 8 or Java 11) from:
+    * Download a Java Runtime Environment (Java 11 or Java 17) from:
 
         https://adoptium.net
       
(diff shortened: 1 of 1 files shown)


src/release/src/markdown/RUNNING.txt
```diff
@@ -12,7 +12,7 @@ In order to install and run this application, follow these steps:
 
 1. Download and install a Java Development Kit
```
mprins, line 13:
```suggestion
1. Download and install a Java Runtime Environment
```

aaime: Hum... JRE won't provide the tools to use the heap dump tool in the UI though (it cannot be done programmatically, we have to call jmap on the command line). Nothing wrong with using a JRE otherwise, but people need to be aware that they'll have less diagnostics available from the GUI.

jodygarnett: We document as only JRE needed; if that is changed by the heap dump tool then it should fail gracefully or something.

aaime: Yes, you'll get an error message rather than the heap dump summary.

PR #6756, opened 2023-04-05
Features templating html double escaping fix

Mike7311:
https://osgeo-org.atlassian.net/browse/GEOS-10921

- Fix for double escaping of html when features-templating module is enabled. Changed used for features-templating XMLFactory [here](https://github.com/geosolutions-it/geoserver/blob/44c5b94bc172e116b7bddf74e31935d5966c24eb/src/community/features-templating/features-templating-ows/src/main/java/org/geoserver/featurestemplating/ows/wfs/TemplateGetFeatureResponseHelper.java#L73) to default, because previous version doesn't support non escaping property. Problem was with degree sign being escaped to \&deg; and then to \&amp;deg; by factory.
- Fixed existing tests to use Jsoup for HTML. Previously HTML responses were validated using XPath and requests were processed with getAsDom which escaped incorrect (for HTML) character \&amp;deg; from response to correct \&deg;. Because of tests that were checking escaping were passing, but actual requests outside tests had incorrect symbol in response.
Had to change couple of tests to Jsoup because &deg is not valid XML entity (but valid HTML) and getAsDom throws an exception when encountering it.

# Checklist

### DIFF
--- a/src/community/features-templating/features-templating-ows/src/main/java/org/geoserver/featurestemplating/ows/wfs/TemplateGetFeatureResponseHelper.java
+++ b/src/community/features-templating/features-templating-ows/src/main/java/org/geoserver/featurestemplating/ows/wfs/TemplateGetFeatureResponseHelper.java
@@ -70,7 +70,7 @@ public class TemplateGetFeatureResponseHelper {
             case GML31:
             case GML32:
             case HTML:
-                XMLOutputFactory xMLOutputFactory = XMLOutputFactory.newInstance();
+                XMLOutputFactory xMLOutputFactory = XMLOutputFactory.newDefaultFactory();
                 // we do ourselves the escaping in the template writer since the one provided in the
                 // XMLStreamWriter implementation
                 // doesn't handle all the characters.
--- a/src/community/features-templating/features-templating-ows/src/main/java/org/geoserver/featurestemplating/ows/wfs/TemplateGetFeatureResponseHelper.java
+++ b/src/community/features-templating/features-templating-ows/src/main/java/org/geoserver/featurestemplating/ows/wfs/TemplateGetFeatureResponseHelper.java
@@ -4,15 +4,19 @@
  */
 package org.geoserver.featurestemplating.ows.wfs;
 
+import com.ctc.wstx.stax.WstxOutputFactory;
 import com.fasterxml.jackson.core.JsonEncoding;
 import com.fasterxml.jackson.core.JsonFactory;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
 import java.util.Optional;
 import javax.xml.namespace.QName;
 import javax.xml.stream.XMLOutputFactory;
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamWriter;
+import org.codehaus.stax2.io.EscapingWriterFactory;
 import org.geoserver.catalog.Catalog;
 import org.geoserver.catalog.FeatureTypeInfo;
 import org.geoserver.featurestemplating.configuration.TemplateIdentifier;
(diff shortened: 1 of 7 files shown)


src/community/features-templating/features-templating-ows/src/main/java/org/geoserver/featurestemplating/ows/wfs/TemplateGetFeatureResponseHelper.java
```diff
@@ -70,7 +70,7 @@ TemplateOutputWriter getOutputWriter(OutputStream output, String version) throws
             case GML31:
             case GML32:
             case HTML:
-                XMLOutputFactory xMLOutputFactory = XMLOutputFactory.newInstance();
+                XMLOutputFactory xMLOutputFactory = XMLOutputFactory.newDefaultFactory();
```
aaime, line 73: @Mike7311 can you elaborate on this change (how and why the change works).

I'm also noting that the [newDefaultFactory method](https://docs.oracle.com/en/java/javase/11/docs/api/java.xml/javax/xml/stream/XMLOutputFactory.html#newDefaultFactory()) is available since Java 9, so the change will work on main and 2.23.x, but won't compile back on 2.22.x and 2.21.x.

Which of course may be a problem, or not, depending on what backport targets you have.

aaime:
@mprins I believe it's the same fix, the previous attempt did not actually work but the way testing was done (with DOM rather than a dedicated HTML parser) was hiding the issue.

src/community/features-templating/features-templating-ows/src/main/java/org/geoserver/featurestemplating/ows/wfs/TemplateGetFeatureResponseHelper.java
Mike7311: Of course. newInstance() finds some implementation of XMLOutputFactory, but in this case this implementation didn't support ESCAPE_CHARS property, which is set on the next lines, thus making factory escape special characters.
image

Mike7311: As for backtracking I will discuss it

src/community/features-templating/features-templating-ows/src/test/java/org/geoserver/featurestemplating/response/HTMLGetComplexFeatureResponseTest.java
```diff
@@ -118,20 +119,17 @@ public void getMappedFeatureHTMLWithJSONLDTemplateMissing() throws Exception {
...
-        Document doc =
-                getAsDOM(
+    public void testEscaping() throws Exception {
+        org.jsoup.nodes.Document doc =
+                getAsJSoup(
                         "wfs?request=GetFeature&version=1.1.0&typename=gsml:MappedFeature"
                                 + "&outputFormat=text/html"
                                 + MF_HTML_PARAM);
-        assertXpathMatches(
-                "60&deg;",
```
nmco, line 128: This assertion is missing.

src/community/features-templating/features-templating-ows/src/main/java/org/geoserver/featurestemplating/ows/wfs/TemplateGetFeatureResponseHelper.java
Mike7311: Changed factory for java 8 support

src/community/features-templating/features-templating-ows/src/test/java/org/geoserver/featurestemplating/response/HTMLGetComplexFeatureResponseTest.java
Mike7311: added back

PR #6757, opened 2023-04-06
[GEOS-10922] Features templating exception on text/plain format

Mike7311:
[GEOS-10922](https://osgeo-org.atlassian.net/browse/GEOS-10922)

Submitting GetCapabilities request with features-templating enabled results in exception due to 
WFSGetFeatureOutputFormat#getCapabilitiesElementName not being overriden for child classes present in features-templating (Fixed). Added test.
Features templating callback is executed for layers which doesn’t have template (for example during GetFeatureInfo) and without defined INFO_FORMAT param results in NullPointerException. Added additinal checkes to TemplateCallback's operationDispatched and responseDispatched methods to support only WMS GetFeatureInfo and WFS GetFeature requests.
Added check for ignoring template if INFO_FORMAT is text/plain (default if empty) because templates do not support it.
Added test for exception on text/plain.
# Checklist

### DIFF
--- a/src/community/features-templating/features-templating-ows/src/main/java/org/geoserver/featurestemplating/ows/TemplateCallback.java
+++ b/src/community/features-templating/features-templating-ows/src/main/java/org/geoserver/featurestemplating/ows/TemplateCallback.java
@@ -176,8 +176,8 @@ public class TemplateCallback extends AbstractDispatcherCallback {
     private boolean operationSupported(Operation operation) {
         String id = operation.getId();
         String serviceId = operation.getService().getId();
-        return id.equals("GetFeatureInfo") && serviceId.equalsIgnoreCase("wms")
-                || id.equals("GetFeature") && serviceId.equalsIgnoreCase("wfs");
+        return "GetFeatureInfo".equalsIgnoreCase(id) && "wms".equalsIgnoreCase(serviceId)
+                || "GetFeature".equalsIgnoreCase(id) && "wfs".equalsIgnoreCase(serviceId);
     }
 
     private Response findResponse(Object param1) {
@@ -202,34 +202,56 @@ public class TemplateCallback extends AbstractDispatcherCallback {
         return response;
     }
 
+    /**
+     * Helper method to find the response for a WMS GetFeatureInfo request.
+     *
+     * <p>If none of the requested layers uses a feature template for the requested info format,
+     * then we delegate on whatever is the default response for the requested info format.
+     *
+     * <p>If at least one of the requested layers uses a feature template for the requested info
+     * format, then all the requested layers need to have a matching feature template for the
+     * requested info format, otherwise an exception will be thrown.
+     */
     private Response getTemplateFeatureInfoResponse(GetFeatureInfoRequest request) {
-        List<MapLayerInfo> layerInfos = request.getQueryLayers();
         String infoFormat = request.getInfoFormat();
         TemplateIdentifier identifier =
                 TemplateIdentifier.fromOutputFormat(request.getInfoFormat());
-        Response result = null;
-        if (identifier != null) {
-            int nTemplates = 0;
-            for (MapLayerInfo li : layerInfos) {
-                if (li != null && li.getResource() instanceof FeatureTypeInfo) {
-                    if (ensureTemplatesExist(li.getFeature(), identifier.getOutputFormat()) != null)
-                        nTemplates++;
-                }
-            }
-            if (nTemplates > 0) {
-                int size = layerInfos.size();
-                if (size != nTemplates)
-                    throw new ServiceException(
-                            "To get a features templating getFeatureInfo a template is needed for every FeatureType but "
-                                    + (size - nTemplates)
-                                    + " among the requested ones are missing a template");
-
-                result =
-                        OWSResponseFactory.getInstance()
-                                .featureInfoResponse(identifier, infoFormat);
+        if (identifier == null) {
+            // features templating doesn't support the requested info type,
+            // hence we delegate on whatever is the default response
+            return null;
+        }
+        // the request info format is supported by features templating
+        // let's check if the requested layers all have a template supporting
+        // the requested info format
+        List<MapLayerInfo> layers = request.getQueryLayers();
+        int matchingTemplates = 0;
+        for (MapLayerInfo layer : layers) {
+            if (layer != null
+                    && layer.getResource() instanceof FeatureTypeInfo
+                    && ensureTemplatesExist(layer.getFeature(), identifier.getOutputFormat())
+                            != null) {
+                matchingTemplates++;
             }
         }
-        return result;
+        // no requested layer is using features templating, hence we delegate in whatever is the
+        // default info format
+        if (matchingTemplates == 0) {
+            return null;
+        }
+        // if one of the requested layer uses a features template, then all the other need to have
+        // one too
+        int matchTemplatesDiff = layers.size() - matchingTemplates;
+        if (matchTemplatesDiff != 0) {
+            // there is at least one requested layer that doesn't have template matching the
+            // requested info format
+            // this situation is currently not supported
+            throw new ServiceException(
+                    "To get a features templating getFeatureInfo a template is needed for every FeatureType but "
+                            + matchTemplatesDiff
+                            + " among the requested ones are missing a template");
+        }
+        return OWSResponseFactory.getInstance().featureInfoResponse(identifier, infoFormat);
     }
 
     private Response getTemplateFeatureResponse(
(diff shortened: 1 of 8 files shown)


src/community/features-templating/features-templating-ows/src/main/java/org/geoserver/featurestemplating/ows/TemplateCallback.java
```diff
@@ -159,18 +160,38 @@ private void replaceTemplatePath(Query q, FeatureTypeInfo fti, RootBuilder root)
...
+    }
+
     private Response findResponse(Object param1) {
         Response response = null;
         if (param1 instanceof GetFeatureInfoRequest) {
-            response = getTemplateFeatureInfoResponse((GetFeatureInfoRequest) param1);
+            GetFeatureInfoRequest request = (GetFeatureInfoRequest) param1;
+            // Checking that template format is specified and that it's not text/plain because it's
+            // not supported,
+            // but is default if nothing is specified
```
aaime, line 190: Bit of fight with the formatter here?

```suggestion
            // Checking that template format is specified and that it's not text/plain because it's
            // not supported, but is default if nothing is specified
```

src/community/features-templating/features-templating-ows/src/main/java/org/geoserver/featurestemplating/ows/TemplateCallback.java
```diff
@@ -159,18 +160,38 @@ private void replaceTemplatePath(Query q, FeatureTypeInfo fti, RootBuilder root)
...
+            if (request.getInfoFormat() != null
+                    && !request.getInfoFormat().equals(MediaType.TEXT_PLAIN_VALUE)) {
+                response = getTemplateFeatureInfoResponse(request);
+            }
```
aaime, line 194: Output formats are pluggable, meaning others not supported could be added tomorrow and this exclusion code would stop working. Can we have a white list of supported formats instead? Do you think it would work better, or would it also break?

src/community/features-templating/features-templating-ows/src/test/java/org/geoserver/featurestemplating/wfs/GetCapabilitiesTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+        Set<String> s1 = new TreeSet<>();
+        for (int i = 0; i < formats.getLength(); i++) {
+            String format = formats.item(i).getNodeName();
+            s1.add(format);
+        }
+
+        assertTrue(s1.contains("JSON-LD"));
+        assertTrue(s1.contains("HTML"));
+    }
```
aaime, line 41: Makes sense for WFS... should we have the same for WMS too (in another test eventually)? The GetFeatureInfo description in GetCapabilities also provides a list of formats.

src/community/features-templating/features-templating-ows/src/test/java/org/geoserver/featurestemplating/wms/WmsGetFeatureInfoTest.java
```diff
@@ -0,0 +1,48 @@
...
+     * Checks if submitting features templating request without format doesn't throw exception
+     *
+     * @throws Exception
+     */
+    @Test
+    public void testGetFeatureInfoEmptyFormat() throws Exception {
+        String response = getAsString(URL);
+        assertFalse(response.contains("NullPointerException"));
+    }
+    /**
```
aaime, line 37: Empty newline missing between two methods:

```suggestion
    }

    /**
```

aaime (changes requested):

src/community/features-templating/features-templating-ows/src/main/java/org/geoserver/featurestemplating/ows/TemplateCallback.java
Mike7311: Yes. I guess...

src/community/features-templating/features-templating-ows/src/main/java/org/geoserver/featurestemplating/ows/TemplateCallback.java
Mike7311: I think that best solution would be something dynamic, similar to GeoServerExtensions.extensions() to return all available formats and compare against them. But I need to check it.

src/community/features-templating/features-templating-ows/src/test/java/org/geoserver/featurestemplating/wfs/GetCapabilitiesTest.java
Mike7311: Will discuss this

src/community/features-templating/features-templating-ows/src/main/java/org/geoserver/featurestemplating/ows/TemplateCallback.java
Mike7311: Removed this comment due to code changes

src/community/features-templating/features-templating-ows/src/test/java/org/geoserver/featurestemplating/wfs/GetCapabilitiesTest.java
```diff
@@ -0,0 +1,42 @@
...
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.NodeList;
+
+/** GetCapabilities test for features templating module */
+public class GetCapabilitiesTest extends WFSTestSupport {
+
+    /**
+     * Checks if features templating specific result formats are present in output and represented
+     * correctly
```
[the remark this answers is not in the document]
Mike7311: fixed

src/community/features-templating/features-templating-ows/src/test/java/org/geoserver/featurestemplating/wms/WmsGetFeatureInfoTest.java
Mike7311: fixed

src/community/features-templating/features-templating-ows/src/main/java/org/geoserver/featurestemplating/ows/TemplateCallback.java
```diff
@@ -191,23 +207,27 @@ private Response getTemplateFeatureInfoResponse(GetFeatureInfoRequest request) {
...
+                    throw new ServiceException(
+                            "To get a features templating getFeatureInfo a template is needed for every FeatureType but "
+                                    + (size - nTemplates)
+                                    + " among the requested ones are missing a template");
 
-            result = OWSResponseFactory.getInstance().featureInfoResponse(identifier, infoFormat);
+                result =
+                        OWSResponseFactory.getInstance()
+                                .featureInfoResponse(identifier, infoFormat);
+            }
```
nmco, line 230: This method is quite complicated and hard to read, I would suggest the following re-writing:
```java
   /**
     * Helper method to find the response for a WMS GetFeatureInfo request.
     * <p/>
     * If none of the requested layers uses a feature template for the requested info format, then we delegate on
     * whatever is the default response for the requested info format.
     * <p/>
     * If at least one of the requested layers uses a feature template for the requested info format, then all the
     * requested layers need to have a matching feature template for the requested info format, otherwise an
     * exception will be thrown.
     */
    private Response getTemplateFeatureInfoResponse(GetFeatureInfoRequest request) {
        // let's check if feature templating supports the requested info format
        String infoFormat = request.getInfoFormat();
        TemplateIdentifier identifier =
                TemplateIdentifier.fromOutputFormat(request.getInfoFormat());
        if (identifier == null) {
            // features templating doesn't support the requested info type,
            // hence we delegate on whatever is the default response
            return null;
        }
        // the request info format is supported by features templating
        // let's check if the requested layers all have a template supporting
        // the requested info format
        List<MapLayerInfo> layers = request.getQueryLayers();
        int matchingTemplates = 0;
        for (MapLayerInfo layer : layers) {
            if (layer != null
                    && layer.getResource() instanceof FeatureTypeInfo
                    && (ensureTemplatesExist(layer.getFeature(), identifier.getOutputFormat()) != null)) {
                // there is an associated template with the layer matching the requested output format
                matchingTemplates++;
            }
        }
        if (matchingTemplates == 0) {
            // no layers is using features templating, hence we delegate in whatever is the default info format
            return null;
        }
        // if one of the requested layer uses a features template, then all the other need to have one too
        int matchTemplatesDiff = layers.size() - matchingTemplates;
        if (matchTemplatesDiff != 0) {
            // there is at least one requested layer that doesn't have template matching the requested info format
            // this situation is currently not supported
            throw new ServiceException(
                    "In a WMS GetFeatureInfo all requested layers need to have a template matching " +
                            "the expected info format, " + matchTemplatesDiff +
                            " among the requested ones are missing such template.");
        }
        // we are good, all the requested layers have a template matching the requested info format
        return OWSResponseFactory.getInstance().featureInfoResponse(identifier, infoFormat);
    }
```

Mike7311: fixed

PR #6813, opened 2023-04-29
[GEOS-10958] Upgrade Spotbugs to 4.7.3

aaime:
[GEOS-10958](https://osgeo-org.atlassian.net/browse/GEOS-10958)

Follow up to GT/GWC upgrades

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/password/RandomPasswordProvider.java
+++ b/src/main/src/main/java/org/geoserver/security/password/RandomPasswordProvider.java
@@ -45,6 +45,13 @@ public class RandomPasswordProvider {
     public static int DefaultPasswordLength = 40;
 
     /**
+     * Cryptographically strong random number generator. As a subclass of {@link java.util.Random},
+     * it's guaranteed to be thread safe (although not necessarily scalable, but for this use case
+     * it should not be a problem).
+     */
+    SecureRandom random = new SecureRandom();
+
+    /**
      * Creates a random password of the specified length, if length <=0, return <code>null</code>
      */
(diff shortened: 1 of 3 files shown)


src/main/src/main/java/org/geoserver/security/password/RandomPasswordProvider.java
```diff
@@ -44,6 +44,13 @@ public class RandomPasswordProvider {
      */
     public static int DefaultPasswordLength = 40;
 
+    /**
+     * Cryptographically strong random number generator. As a subclass of {@link java.util.Random},
+     * it's guaranteed to be thread safe (although not necessarily scalable, but for this use case
+     * it should not be a problem).
+     */
+    SecureRandom random = new SecureRandom();
```
bradh, line 52: Is this driven by new rules in 4.7.3, or just coincidental?

aaime: Driven by new rules, quoting:

> ### DMI: Random object created and used only once (DMI_RANDOM_USED_ONLY_ONCE)
> 
> This code creates a java.util.Random object, uses it to generate one random number, and then discards the Random object. This produces mediocre quality random numbers and is inefficient. If possible, rewrite the code so that the Random object is created once and saved, and each time a new random number is required invoke a method on the existing Random object to obtain it.
> 
> If it is important that the generated Random numbers not be guessable, you must not create a new Random for each random number; the values are too easily guessable. You should strongly consider using a java.security.SecureRandom instead (and avoid allocating a new SecureRandom for each random number needed).

PR #6832, opened 2023-05-08
[GEOS-10929] isolated workspace may cause broken links between catalo…

NielsCharlier:
[GEOS-10929](https://osgeo-org.atlassian.net/browse/GEOS-10929)

…g objects

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/impl/CatalogImpl.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/CatalogImpl.java
@@ -15,7 +15,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 import java.util.Stack;
-import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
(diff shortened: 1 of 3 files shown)


src/community/wps-sextante/src/main/java/org/geoserver/wps/sextante/GTVectorLayer.java
```diff
@@ -26,6 +25,7 @@
 import org.geotools.feature.FeatureCollection;
 import org.geotools.feature.simple.SimpleFeatureBuilder;
 import org.geotools.feature.simple.SimpleFeatureTypeBuilder;
+import org.geotools.util.logging.Logging;
```
aaime, line 28: What's happening here? Unrelated reformats here and in a few other files.

src/main/src/main/java/org/geoserver/catalog/impl/CatalogImpl.java
```diff
@@ -141,6 +141,7 @@ public Iterable<CatalogValidator> getValidators() {
     }
 
     public void setFacade(CatalogFacade facade) {
+        rawFacade = facade;
```
aaime, line 144: So the actual important change is here... the refactor of the event management obfuscates it (still a good thing, but it's a distraction when trying to find what makes the fix actually work).

src/main/src/main/java/org/geoserver/catalog/impl/CatalogImpl.java
```diff
@@ -2277,4 +2246,12 @@ public <T extends CatalogInfo> T get(Class<T> type, Filter filter)
     public CatalogCapabilities getCatalogCapabilities() {
         return facade.getCatalogCapabilities();
     }
+
+    public CatalogImpl getRawCatalog() {
```
aaime, line 2250: Bit of Javadoc with an explanation of why the original CatalogImpl is not "raw"?

aaime (changes requested):

NielsCharlier:
@aaime thanks for the review, made the changes

PR #6833, opened 2023-05-10
Allow basic configuration of data in CRS with authority other than EPSG

aaime:
These three changes allow configuration of a shapefile and a geotiff in GeoServer. Still plenty of problems left, but enough to configure data and use it in the OpenLayers preview (and in that one only, mind).
The IAU authority is used only as a test support for the moment, not given yet a usable module of its own.

Needs https://github.com/geotools/geotools/pull/4284

# Checklist


src/main/src/main/java/org/geoserver/catalog/ResourcePool.java
```diff
@@ -499,6 +501,30 @@ public CoordinateReferenceSystem getCRS(String srsName) throws IOException {
...
     }
 
+    /**
+     * Looks up the identifier of a given {@link CoordinateReferenceSystem} object, giving a
+     * preference to EPSG codes expressed as <code>EPSG:xyzw</code>, when possible, and returning
+     * another code, otherwise. This behavior is specific to GeoServer, thus, it has not been
+     * included in the GeoTools {@link CRS} facade.
+     *
+     * <p>TODO: method placed in ResourcePool since it already has some CRS handling. Should we roll
+     * a GeoServer CRSUtils class instead? (calling it just CRS
```
dromagnoli, line 511: Is this statement ended? (calling it just CRS ....

dromagnoli (approved):
Looks good to me

PR #6841, opened 2023-05-16
[GEOS-10981] Slow CSW GetRecords requests with JDBC Configuration

sikeoka:
[GEOS-10981](https://osgeo-org.atlassian.net/browse/GEOS-10981)

This PR adds an interface for an object that splits catalog filters into supported and unsupported components. The JDBCConfig code to do this was moved out of QueryBuilder into a new Spring bean implementing this interface. The CSW store now loads this bean from the context and uses it to split the request filter into the supported and unsupported components before modifying the filter and passing it to the catalog.

  

# Checklist

### DIFF
--- a/src/community/jms-cluster/jms-geoserver/src/main/java/org/geoserver/cluster/impl/JMSXStreamFactory.java
+++ b/src/community/jms-cluster/jms-geoserver/src/main/java/org/geoserver/cluster/impl/JMSXStreamFactory.java
@@ -13,6 +13,7 @@ import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.List;
+import java.util.Locale;
 import org.geoserver.catalog.event.CatalogEvent;
 import org.geoserver.config.GeoServer;
 import org.geoserver.config.util.XStreamPersister;
@@ -20,7 +21,9 @@ import org.geoserver.config.util.XStreamPersisterFactory;
 import org.geoserver.config.util.XStreamServiceLoader;
 import org.geoserver.platform.GeoServerExtensions;
 import org.geotools.coverage.grid.GridGeometry2D;
+import org.geotools.geometry.jts.ReferencedEnvelope;
 import org.opengis.coverage.grid.GridGeometry;
+import org.opengis.util.InternationalString;
 
 public class JMSXStreamFactory {
 
@@ -66,6 +69,11 @@ public class JMSXStreamFactory {
         this.xs.allowTypeHierarchy(GridGeometry.class);
         this.xs.registerConverter(
                 persisterXStream.getConverterLookup().lookupConverterForType(GridGeometry2D.class));
+        // allow envelopes (show up directly in proxies when changing the bbox)
+        this.xs.allowTypeHierarchy(ReferencedEnvelope.class);
+        // allow international strings
+        this.xs.allowTypeHierarchy(InternationalString.class);
+        this.xs.allowTypeHierarchy(Locale.class);
     }
 
     public XStream createXStream() {
--- a/src/community/jms-cluster/jms-geoserver/src/test/java/org/geoserver/cluster/integration/IntegrationTest.java
+++ b/src/community/jms-cluster/jms-geoserver/src/test/java/org/geoserver/cluster/integration/IntegrationTest.java
@@ -18,16 +18,20 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Locale;
 import java.util.UUID;
 import org.geoserver.catalog.AttributionInfo;
 import org.geoserver.catalog.Catalog;
 import org.geoserver.catalog.CoverageInfo;
 import org.geoserver.catalog.CoverageStoreInfo;
+import org.geoserver.catalog.DataLinkInfo;
 import org.geoserver.catalog.DataStoreInfo;
 import org.geoserver.catalog.FeatureTypeInfo;
+import org.geoserver.catalog.Keyword;
 import org.geoserver.catalog.LayerGroupInfo;
 import org.geoserver.catalog.LayerGroupInfo.Mode;
 import org.geoserver.catalog.LayerInfo;
+import org.geoserver.catalog.MetadataLinkInfo;
 import org.geoserver.catalog.NamespaceInfo;
 import org.geoserver.catalog.ProjectionPolicy;
 import org.geoserver.catalog.PublishedType;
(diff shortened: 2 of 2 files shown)


aaime:
I don't understand why JDBCConfig cannot do filter splitting on its own? If it does not support spatial filters, it should just move them into a post-filtering code, without requiring upper levels in the archictures to participate into it.

To draw a parallel, it's as if WFS would have to participate (via extension point) into the handling of filters against a JDBCDataStore, while the store is currently handling everything transparently instead.

sikeoka:
I don't think this is something that could be done in the filter splitter because this modified CSW code actually duplicates a part of the supported filter (the IsInstanceOf function) into both the supported and unsupported filters to create a filter that is logically the same but can be properly split by filter splitter in JDBC Config.

This is still simplified from the actual CSW filter but the old filter is like:
```
(
 (IsInstanceOf(LayerGroupInfo) AND GroupFilter)
 OR
 (IsInstanceOf(LayerInfo) AND LayerInfo)
)
```
and the new filter is like:
```
(
 (
  (IsInstanceOf(LayerGroupInfo) AND SupportedGroupFilter)
  OR
  (IsInstanceOf(LayerInfo) AND SupportedLayerInfo)
 )
 AND
 (
  (IsInstanceOf(LayerGroupInfo) AND UnsupportedGroupFilter)
  OR
  (IsInstanceOf(LayerInfo) AND UnsupportedLayerFilter)
 )
)
```

aaime:
Hum interesting... so in general the conversion you're trying to do is not leading to the same truth table, so no matter what the skill level is, it cannot be implemented as a general conversion in the filter splitter.
In practice, for this case, it's working, because the only places where the truth tables differ are impossible (cases where isInstance of layer and group both return true).

For reference, try entering these two:

```
(isGroup and (sgf and ugf)) or (isLayer and (slf and ulf))

((isGroup and sgf) or (isLayer and slf)) and ((isGroup and ugf) or (isLayer and ulf))
```

in this online tool: https://web.stanford.edu/class/cs103/tools/truth-table-tool/

That said, the solution is still too ah-hoc, and makes a supported module more complex to support a community module that's rarely used.

Changes for changes... would you be amenable to a simpler solution, where the code in CSW does two queries, one for layers, one for groups, in sequence? It should be easy to wrap into a single iterator that runs the second query once the first one is done. For the default in-memory catalog the performance difference should be negligible. For the JDBC one, you should still get a good improvement.
And add a comment to specify why it's done that way.

Personally, I prefer either of the above, to a new extension point that we'd have to keep around long term, even if JDBC config eventually gets removed one day (with two interested parties that refuse to bring the module up in supported land, its future is hanging by a thread... or two :rofl: )

NielsCharlier:
I agree with @aaime that this is not ideal, it seems to go against the 'separation of concerns' principle.
Splitting the request in two I remember brought huge inconveniences... so that would not be my preferred solution.

I wonder if it really is not possible to resolve the problem in the filter splitting algorithm.  
Would it not be possible deal with "OR" filters by splitting their parts separately and then reconstruct a logically equivalent filter by duplicating the necessary parts?

sikeoka:
Making two requests would prevent sorting and paging from working in the database when the filter is fully supported.  I could keep the same interface and use it to decide whether to split the request but that doesn't seem much different than this implementation for your concerns.

I could also mostly leave CSW alone and just move everything into JDBC Config where it would be able to detect what the current CSW catalog filters look like and have special code to break the filter apart, split the portion that came from the user request and then put everything back together but that would be a lot messier and more fragile than this approach.

NielsCharlier:
Note it is not absolute necessary to _split_ a filter into mutually exclusive separate parts, in fact the post-filter could simply be the entire filter, with probably negligible loss in performance

So, in we get a filter of the form

a OR b 

where a and b both contain unsupported parts. Then why not simply say:

preFilter = supportedPart(a) OR supportedPart(b)
postFilter = a OR b

Does not have to be part of the current splitting visitor, but could be implemented in a generic way that makes no assumptions about a very specific structure

sikeoka:
Splitting the CSW filter only works because there is essentially a boolean variable with a different value on both sides of the OR like ((X AND Y) OR (!X AND Z)).  This technically isn't true and the code is making an assumption that it will always be true.

aaime:
Interesting consideration by Niels... I've tried to verify it with the truth table generator I linked to before, if I get it right, the comparison is between these two (in case anyone wonders, gf is group filter, lf is layer filter, s is supported, u is unsupported):

```
(isGroup and (sgf and ugf)) or (isLayer and (slf and ulf))

((isGroup and sgf) or (isLayer and slf)) and (isGroup and (sgf and ugf)) or (isLayer and (slf and ulf))
```

Indeed the generated truth table is exactly the same, so this splitting approach is indeed going to work.

NielsCharlier:
Indeed, @sikeoka we really do not need the assumption of a "different value on both sides of the OR like ((X AND Y) OR (!X AND Z))"

the point is (as @aaime confirmed using a truth table generator), that 
((s1 and u1) or (s2 and u2))

 logically implies

 (s1 or s2)

Therefore it is safe to generate a pre-filter like this, and then use the entire filter as a post-filter. In this specific case this would lead to the desired result since the "IsInstanceOf()" parts are supported and would end up in both filters.

sikeoka:
@aaime, your example is not accurate.  You need to find a way to split (x1 and y1 and z1) or (x2 and y2 and z2) into (x1 and y1) or (x2 and y2) and something else.

It's also not just the logic part.  I'm not even sure where to begin with modifying https://github.com/geotools/geotools/blob/main/modules/library/main/src/main/java/org/geotools/filter/visitor/CapabilitiesFilterSplitter.java to be able to do that.  Doing it filter splitter could also cause it to unnecessarily split OR expressions that will end up in the unsupported filter anyways if users are sending complex filters to JDBC data stores.  I have limited resources to work on this but I would be fine if @NielsCharlier wants to try implementing this idea.

I could do something specifically for CSW filters in JDBC Config but again that would be a lot messier than what this PR is doing in CSW.

sikeoka:
https://github.com/geoserver/geoserver/pull/6855 shows what it might look like with CSW making separate layer group and layer requests when the filter is partially supported and partially unsupported.  The changes in that PR are untested.

sikeoka:
Any additional comments on this?

aaime:
We are not paid to follow up with this ticket, which has already consumed a fair amount of time (in my case, time tracked in my company's timesheet). We already gave you two alternative approaches: keeping the garbage in a community module is acceptable, figuring out if the top level filter is an OR and do a partial pre-filter with full filter as a post filter is another option. If neither satisfies you, it's up to you to come up with a different solution.

NielsCharlier:
@sikeoka I'm sorry but I really do not have the time to write this.  

My first preference would be a generic solution which I really do believe is possible: if it is too hard to make this a part the existing filter splitter, you could write a new, additional filter 'splitter' on top of it that only deals with an OR structure at the root.

Then: I'd rather not have 'garbage' in jdbcconfig, but I suppose a hack in jdbcconfig that checks for the specific filter generated by CSW is acceptable at this point (which then hopefully will be replaced one day).

Changing CSW to do separate requests I think is my least preferred solution for the reasons you gave.

PR #6867, opened 2023-05-26
[GEOS-10992] Make GWC UI for disk quota expose HSQLDB, remove H2, automatically update existing installations - fix community build

canmehteroglu:
[GEOS-10992](https://osgeo-org.atlassian.net/browse/GEOS-10992)

# Checklist

### DIFF
--- a/src/community/taskmanager/pom.xml
+++ b/src/community/taskmanager/pom.xml
@@ -28,6 +28,11 @@
       <version>1.8.5</version>
       <scope>compile</scope>
     </dependency>
+    <dependency>
+      <groupId>com.h2database</groupId>
+      <artifactId>h2</artifactId>
+      <version>1.1.119</version>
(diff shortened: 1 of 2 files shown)


src/community/taskmanager/pom.xml
```diff
@@ -28,6 +28,11 @@
       <version>1.8.5</version>
       <scope>compile</scope>
     </dependency>
+    <dependency>
+      <groupId>com.h2database</groupId>
+      <artifactId>h2</artifactId>
+      <version>1.1.119</version>
```
aaime, line 34: Do we need the version number? Should be managed in the dependency management of the main pom.xml file of GeoServer

canmehteroglu: You're right. I missed that community project is also a child of geoserver parent pom. Removed.

PR #6869, opened 2023-05-30
[GEOS-10994] OOM due to too many dimensions when range requested

turingtestfail:
[GEOS-10994](https://osgeo-org.atlassian.net/browse/GEOS-10994)

extended to getfeatureinfo

unit tests

getmap integration tests

getfeatureinfo integration tests

clean

cleanup

  

# Checklist

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/WMS.java
+++ b/src/wms/src/main/java/org/geoserver/wms/WMS.java
@@ -19,6 +19,7 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
+import java.util.ListIterator;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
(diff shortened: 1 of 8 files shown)


src/wms/src/main/java/org/geoserver/wms/WMS.java
```diff
@@ -324,6 +324,77 @@ public String getVersion() {
...
+                        == DimensionInfo.DEFAULT_MAX_REQUESTED_DIMENSION_VALUES) {
+            // max is default, which is unlimited
+            return;
+        }
+        TreeSet<Object> treeSet = new TreeSet<>();
+        int maxDimensionsToTest = getServiceInfo().getMaxRequestedDimensionValues();
+        if (times != null && !times.isEmpty() && times.get(0) instanceof DateRange) {
+            // list is null for default value, contains Date(s) for individual values, we are only
+            // checking range here
+            DateRange range = (DateRange) times.get(0);
```
aaime, line 353: It's a bit optimistic, one can provide a comma-separated list of single points and ranges (various of them). E.g., for elevation if could be something like, ``10,15/30,25,5/17,100/200`` (e.g. mix of types, multiple ranges, eventually out of order). So this needs to loop over the values, add the single values in the set, grab the range contents, and check the limit at the end of each loop for good measure.

src/wms/src/test/java/org/geoserver/wms/WMSTest.java
```diff
@@ -212,6 +214,36 @@ public void testCacheConfiguration() {
...
+    @Test(expected = ServiceException.class)
+    public void testCheckMaxDimensionsTime() throws Exception {
+        WMSInfo info = wms.getServiceInfo();
+        info.setMaxRequestedDimensionValues(1);
+        getGeoServer().save(info);
+        setupStartEndTimeDimension(
+                TIME_WITH_START_END.getLocalPart(), "time", "startTime", "endTime");
+        String name = TIME_WITH_START_END.getLocalPart();
+        MapLayerInfo mapLayerInfo = new MapLayerInfo(getCatalog().getLayerByName(name));
+        SimpleDateFormat formatter = new SimpleDateFormat("yyyy-MM-dd", Locale.ENGLISH);
```
aaime, line 226: Good thinking about the Locale, wondering if we need to fix a timezone too (other tests seem to be doing that locally in the date format,).

aaime (changes requested):
Good start, needs a small push on generality.

aaime (approved):

PR #6871, opened 2023-05-30
[GEOS-10998] LayerGroupContainmentCache is being rebuilt on each ApplicationEvent

Mike7311:
[GEOS-10998](https://osgeo-org.atlassian.net/browse/GEOS-10998)

[[GEOS-10998] LayerGroupContainmentCache is being rebuilt on each ApplicationEvent](https://osgeo-org.atlassian.net/browse/GEOS-10998)

Layer group cache was being rebuild on most of GS actions, which resulted in performance drop. Changed code to rebuild only on ContextRefreshedEvent

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/impl/LayerGroupContainmentCache.java
+++ b/src/main/src/main/java/org/geoserver/security/impl/LayerGroupContainmentCache.java
@@ -29,8 +29,8 @@ import org.geoserver.catalog.event.CatalogModifyEvent;
 import org.geoserver.catalog.event.CatalogPostModifyEvent;
 import org.geoserver.catalog.event.CatalogRemoveEvent;
 import org.geoserver.catalog.impl.LayerGroupStyleListener;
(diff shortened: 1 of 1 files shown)


Mike7311:
Didn't add the test, because there was already [test](https://github.com/geoserver/geoserver/blob/d362a38d9d0cc8d496188901f84359074bd9ca74/src/main/src/test/java/org/geoserver/security/impl/LayerGroupContainmentCacheTest.java#L191) for this and I just narrowed method parameter type (from abstract class to it's concrete implementation).

dromagnoli:
> Didn't add the test, because there was already [test](https://github.com/geoserver/geoserver/blob/d362a38d9d0cc8d496188901f84359074bd9ca74/src/main/src/test/java/org/geoserver/security/impl/LayerGroupContainmentCacheTest.java#L191) for this and I just narrowed method parameter type (from abstract class to it's concrete implementation).

I see. Ok thanks. Going to merge it.

PR #6877, opened 2023-05-31
[GEOS-10997] Data Security Layer groups

turingtestfail:
[GEOS-10997](https://osgeo-org.atlassian.net/browse/GEOS-10997)

fixed reversed superclass

more fixed super

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
+++ b/src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
@@ -674,7 +674,7 @@ public class DefaultResourceAccessManager implements ResourceAccessManager {
         }
     }
 
-    private Filter buildInFunctionResourceFilter(
+    protected Filter buildInFunctionResourceFilter(
             Authentication user, Class<? extends CatalogInfo> clazz) {
         // base access
         boolean rootAccess = canAccess(user, root);
@@ -705,9 +705,12 @@ public class DefaultResourceAccessManager implements ResourceAccessManager {
                 }
                 boolean layerAccess = canAccess(user, layerNode);
                 if (layerAccess != wsAccess) {
-                    if (clazz.isAssignableFrom(published.getClass()))
+                    if (ResourceInfo.class.isAssignableFrom(clazz)
+                            && published instanceof LayerInfo) {
+                        layerExceptionIds.add(((LayerInfo) published).getResource().getId());
+                    } else {
                         layerExceptionIds.add(published.getId());
-                    else layerExceptionIds.add(((LayerInfo) published).getResource().getId());
+                    }
                 }
             }
 
--- a/src/security/security-tests/src/test/java/org/geoserver/security/impl/DefaultDataAccessManagerTreeTest.java
+++ b/src/security/security-tests/src/test/java/org/geoserver/security/impl/DefaultDataAccessManagerTreeTest.java
@@ -12,6 +12,8 @@ import static org.junit.Assert.assertTrue;
 
 import java.util.Properties;
 import java.util.Set;
+import org.geoserver.catalog.ResourceInfo;
+import org.geoserver.catalog.WorkspaceInfo;
 import org.geoserver.security.AccessMode;
 import org.junit.Before;
 import org.junit.Test;
@@ -90,7 +92,7 @@ public class DefaultDataAccessManagerTreeTest extends AbstractAuthorizationTest 
         assertEquals(2, root.children.size());
         SecureTreeNode topp = root.getChild("topp");
         assertNotNull(topp);
-        assertEquals(3, topp.children.size());
+        assertEquals(4, topp.children.size());
         SecureTreeNode states = topp.getChild("states");
         SecureTreeNode landmarks = topp.getChild("landmarks");
         SecureTreeNode bases = topp.getChild("bases");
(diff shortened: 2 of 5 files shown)


src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
```diff
@@ -705,9 +705,11 @@ private Filter buildInFunctionResourceFilter(
                 }
                 boolean layerAccess = canAccess(user, layerNode);
                 if (layerAccess != wsAccess) {
-                    if (clazz.isAssignableFrom(published.getClass()))
+                    if (ResourceInfo.class.isAssignableFrom(clazz)) {
+                        layerExceptionIds.add(((LayerInfo) published).getResource().getId());
+                    } else {
                         layerExceptionIds.add(published.getId());
-                    else layerExceptionIds.add(((LayerInfo) published).getResource().getId());
+                    }
```
aaime, line 712: The change works (tested manually) and yet I'm perplexed... if I had a situation where I'm listing ResourceInfo.class, and there is a workspace specific LayerGroup that has access different than the workspace, it would still throw a ClassCastException.

Yes, manually reproduced that case. I had to set up the limited role access on a workspace specific layer group, and then asked for the list of all collections in OGC API - Features, and here we go:

```
	at freemarker.ext.beans.BeansWrapper.invokeMethod(BeansWrapper.java:1552)
	at freemarker.ext.beans.BeanModel.invokeThroughDescriptor(BeanModel.java:233)
	at freemarker.ext.beans.BeanModel.get(BeanModel.java:152)
	... 130 more
Caused by: java.lang.ClassCastException: class jdk.proxy2.$Proxy53 cannot be cast to class org.geoserver.catalog.LayerInfo (jdk.proxy2.$Proxy53 is in module jdk.proxy2 of loader 'app'; org.geoserver.catalog.LayerInfo is in unnamed module of loader 'app')
	at org.geoserver.security.impl.DefaultResourceAccessManager.buildInFunctionResourceFilter(DefaultResourceAccessManager.java:709)
	at org.geoserver.security.impl.DefaultResourceAccessManager.getSecurityFilter(DefaultResourceAccessManager.java:562)
	at org.geoserver.security.CatalogFilterAccessManager.getSecurityFilter(CatalogFilterAccessManager.java:178)
	at org.geoserver.security.SecureCatalogImpl.securityFilter(SecureCatalogImpl.java:1721)
	at org.geoserver.security.SecureCatalogImpl.list(SecureCatalogImpl.java:1627)
	at org.geoserver.catalog.impl.AbstractFilteredCatalog.list(AbstractFilteredCatalog.java:933)
	at org.geoserver.catalog.impl.LocalWorkspaceCatalog.list(LocalWorkspaceCatalog.java:588)
	at org.geoserver.catalog.impl.LocalWorkspaceCatalog.list(LocalWorkspaceCatalog.java:569)
	at org.geoserver.ogcapi.v1.features.CollectionsDocument.getCollections(CollectionsDocument.java:55)
```

src/security/security-tests/src/test/java/org/geoserver/security/impl/DefaultDataAccessManagerTreeTest.java
```diff
@@ -135,4 +135,17 @@ public void testComplex() throws Exception {
...
+    @Test
+    public void testBuildInFunctionResourceFilter() throws Exception {
+        DefaultResourceAccessManager defaultResourceAccessManager =
+                (DefaultResourceAccessManager) buildManager("complex.properties");
+        // validate that the filter is built correctly and there are no casting issues when
+        // populating layerExceptionIds
+        assertEquals(
+                "[[ NOT [ in([id], [arc.grid-lid]) = true ] ] AND [ NOT [ in([id], [bases-lid]) = true ] ]]",
+                defaultResourceAccessManager
+                        .buildInFunctionResourceFilter(rwUser, toppWs.getClass())
```
aaime, line 148: toppWs is a proxy, it has a really weird class. Maybe better to have tests using explicit classes, like ResourceInfo.class, WorkspaceInfo.class

aaime (changes requested):

aaime:
And oh, if you want to try a more integration style test, SecureCatalogImplTest should have the ingredients needed.

src/security/security-tests/src/test/java/org/geoserver/security/impl/DefaultDataAccessManagerTreeTest.java
turingtestfail: fixed

src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
turingtestfail: fixed

src/gwc/-u
```diff
@@ -0,0 +1,27 @@
+Note: Unnecessary use of -X or --request, GET is already inferred.
```
aaime, line 1: I don't think this was meant to be included in the PR :rofl:

aaime (changes requested):
One extra file slipped into the PR

src/gwc/-u
turingtestfail: fixed

PR #6879, opened 2023-06-01
[GEOS-11005] Make sure H2 dependencies are included in the packages of optional modules that still need it

canmehteroglu:
[GEOS-11005](https://osgeo-org.atlassian.net/browse/GEOS-11005)

(only compile dependencies)

# Checklist


src/release/ext-importer.xml
```diff
@@ -19,6 +19,7 @@
         <include>*fileupload*</include>
         <include>*commons-vfs2*</include>
         <include>*commons-exec*</include>
+        <include>h2-*.jar</include>
```
aaime, line 22: I've checked, the compile dependency towards h2 in the importer module turned out to be not really necessary, it's really just a test dependency, the module has no actual compile dependency towards it:

```
> cd extension/importer
>  git grep -i h2 "**/*.java" | grep -v Test
core/src/main/java/org/geoserver/importer/Database.java:            // file based databases might be a full path to a file (sqlite, h2, etc..) use only
core/src/main/java/org/geoserver/importer/Importer.java:                    // sniff for postgis (h2 is used in tests and will cause failure if this occurs)
```

Could you downgrade that dependency in the importer module, to test scope, and remove this line?

canmehteroglu: overrode h2 dependency scope in importer project and removed h2 from ext-importer.xml

PR #6883, opened 2023-06-01
[GEOS-10981] Slow CSW GetRecords requests with JDBC Configuration

sikeoka:
The PR adds a workaround to JDBC Configuration for the current GeoTools filter splitter's limitations when handling partially supported OR filters.  It is intended specifically to split the filters that CSW GetRecords requests send to the JDBC Configuration catalog into a component that can be translated into SQL and a component that is evaluated in-memory when the requested filter is partially supported and partially unsupported.

  

# Checklist


sikeoka:
@NielsCharlier, this is the PR that fixes CSW performance issues in JDBC Config.

sikeoka:
Any thoughts on this @NielsCharlier?

NielsCharlier:
I had a look, I think it is fine. I still wonder though if it would be that hard to modify this code to make it more generic: simply drop checking for the isinstanceof filter and duplicating all supported filters.

sikeoka:
I will go ahead and merge this.  The outer structure of filters passed to JDBC Configuration is essentially hardcoded so I don't really see any benefit in trying to implement a more generic solution here unless there's a practical example where it would be helpful.

NielsCharlier:
Well, that it would be more robust against changes in the future as you mentioned earlier. Perhaps some time

PR #6908, opened 2023-06-08
[GEOS-10979] metadata: don't crash on missing properties file

NielsCharlier:
[GEOS-10979](https://osgeo-org.atlassian.net/browse/GEOS-10979)

  

# Checklist

### DIFF
--- a/src/extension/metadata/src/test/java/org/geoserver/metadata/EmptyConfigurationTest.java
+++ b/src/extension/metadata/src/test/java/org/geoserver/metadata/EmptyConfigurationTest.java
@@ -1,0 +1,45 @@
+/* (c) 2023 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+
+package org.geoserver.metadata;
+
+import org.apache.wicket.util.file.File;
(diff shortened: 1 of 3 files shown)


aaime:
As a supported module, a test is required for a PR to be merged against it.

src/extension/metadata/src/test/java/org/geoserver/metadata/EmptyConfigurationTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @BeforeClass
+    public static void init() throws Exception {
+        // this method shadows the init method from AbstractMetadataTest so that there
+        // is no metadata directory in the data directory.
+
+        // set data directory
+        MockData emptyDataDirectory = new MockData();
+        System.setProperty(
+                "GEOSERVER_DATA_DIR", emptyDataDirectory.getDataDirectoryRoot().toString());
```
aaime, line 28: The JVMs in tests are re-used for the tests of the same module, but also across modules in a multi-module build. So please, make sure you don't leave sys var set up, clean them in a @AfterClass.

NielsCharlier: Indeed I was having trouble with consecutive tests in maven  so I went for another approach

PR #6916, opened 2023-06-09
[GEOS-10949] URL Check UI and documentation improvements

jodygarnett:
[GEOS-10949](https://osgeo-org.atlassian.net/browse/GEOS-10949)

This pull request benefits from https://github.com/geotools/geotools/pull/4337 to allow locations to be normalized prior to checking.

First up I love this new functionality, and had a chance to check the user interface and documentation today.

* [x] UI layerout
* [x] User Interface page naming
* [x] Documentation updates reflecting the above UI changes 
* [x] Invalid localhost default for Regex Check
* [x] Hardening the default regex

This PR addresses some feedback discovered while reviewing:

* main: https://github.com/geoserver/geoserver/pull/6790
* 2.23.x: https://github.com/geoserver/geoserver/pull/6901
* 2.22.x: https://github.com/geoserver/geoserver/pull/6905

This also addresses the @sikeoka observation that an additional URL Check could be used in to verify RemoteRequestInputProvider.getBodyReferenceHref() reference.

Ideally this can be reviewed and merged onto main prior the the above back ports being accepted.

# Checklist


jodygarnett:
@canmehteroglu you prepared the back ports - are you interested in checking my work here?

aaime:
@jodygarnett we need better coordination to avoid work duplication, @canmehteroglu was looking into the same issues.

As far as I can tell, a critical point from @sikeoka review is not yet addressed, that is, hardening the default regex to avoid path traversal. Can you confirm?

jodygarnett:
> As far as I can tell, a critical point from @sikeoka review is not yet addressed, that is, hardening the default regex to avoid path traversal. Can you confirm?

I did not look at hardening the default regex; only corrected it for ``geosever``.

I looked at @sikeoka feedback with respect to WPS POST body by href ...

aaime:
Ok, I'll ask @canmehteroglu to complete the regex work in a separate PR then. We are going to squash merge this one and then @canmehteroglu will have to add this commit and the one he's going to prepare on top of the backport PRs.

jodygarnett:
Sounds good; what is your timeframe? I am happy to take over tomorrow am.

I am keen to share this improvement with my customers.

aaime:
@jodygarnett had a quick look, seems good, just a note about the documentation: we can't have images reported at "W: 1614px | H: 760px", they are just too big. An image for the documentation should be at most 700px, or around that size.

doc/en/user/source/security/urlchecks.rst
```diff
@@ -3,65 +3,110 @@
...
+   URL Checks table
 
-The top check box allows to enable or disable the URL checks. If disabled, no URL checks
-will be performed. This is not a secure setting, but could be used to upgrade existing 
-applications where GeoServer is not directly exposed to the public internet, and the
-requests are controlled by a front application validating all inputs.
+Use the :guilabel:`Enable/Disable URL Checks` enable this safety feature:
 
-The list of checks allows to add, remove and edit the checks.
+* When checkbox enabled, :guilabel:`URL checks are enabled`, URL checks are performed to limit GeoServer access to remote resources as outlined above.
```
bradh, line 29:
```suggestion
* When the :guilabel:`URL checks are enabled` checkbox is enabled, URL checks are performed to limit GeoServer access to remote resources as outlined above.
```

aaime:
Ok @canmehteroglu can work on this starting Monday, so if you want to take over the regex hardening as well sooner, that would be appreciated.

doc/en/user/source/security/urlchecks.rst
```diff
@@ -3,65 +3,110 @@
...
 
-Currently, regular expression based checks can be configured, with the following parameters for each check:
+The buttons for adding and removing URL checks can be found at the top of the :guilabel:`URL Check list` table.
+
+To add a URL Check, press the :guilabel:`Add new URL check` button. You will be prompted to enter URL check details (as described in :ref:`security_urlchecks_edit` below).
+
+Removing a regular expression based check
+-----------------------------------------
+
+To remove a URL Checks, select the checkbox next to one or more rows in the :guilabel:`URL Check list` table.
```
bradh, line 47:
```suggestion
To remove a URL Check, select the checkbox next to one or more rows in the :guilabel:`URL Check list` table.
```

bradh: Or "To remove URL checks..."

doc/en/user/source/security/urlchecks.rst
```diff
@@ -3,65 +3,110 @@
...
-* https://www.freeformatter.com/java-regex-tester.html 
+``^https://safeWFS/geoserver/ows/.*SERVICE=WFS.*$``
 
+.. figure:: images/urlchecks-edit.png
+   
+   Configure Regular Expression URL check
+    
+.. note::
+
+   Web sites are available to can help define a valid Java regular expression pattern. These tools can be used to interpret, explain and test regular expressions. For example:
```
bradh, line 86:
```suggestion
   Web sites are available to help define a valid Java regular expression pattern. These tools can be used to interpret, explain and test regular expressions. For example:
```

doc/en/user/source/security/urlchecks.rst
```diff
@@ -3,65 +3,110 @@
...
+
+   * https://regex101.com/ (enabled the Java 8 flavor)
```
bradh, line 88:
```suggestion
   * https://regex101.com/ (enable the Java 8 flavor)
```

doc/en/user/source/security/urlchecks.rst
```diff
@@ -3,65 +3,110 @@
...
+.. list-table::
+   :widths: 30 70 
+   :header-rows: 1
+
+   * - Field
+     - Description
+   * - URL to check 
+     - Supply URL of external resource to check if access is allowed
+
+Press :guilabel:`Test URL` button to perform the checks. If at least one URL Check matches the URL, it will be allowed and the test will indicate the URL Check permitting access. Otherwise it will be rejected and the test will indicate that no URL Check matched.
```
bradh, line 107:
```suggestion
Press the :guilabel:`Test URL` button to perform the checks. If at least one URL Check matches the URL, it will be allowed and the test will indicate the URL Check permitting access. Otherwise it will be rejected and the test will indicate that no URL Check matched.
```

doc/en/user/source/security/urlchecks.rst
```diff
@@ -3,65 +3,110 @@
...
 
-More functionality might go under this kind of scrutiny in the future, check this page
-for eventual updates.
+Check this page for any additional remote service access checks added in the future.
 
 Configuration of URL checks
 ---------------------------
 
-The "URL checks" page under the Security section allows configuration of the URL checks.
+Navigate to :menuselection:`Data > URL Checks` page to manage and configure URL checks.
```
bradh, line 21:
```suggestion
Navigate to :menuselection:`Data > URL Checks` page to manage and configure URL Checks.
```

bradh (dismissed):
Minor doc edits

jodygarnett:
> @jodygarnett had a quick look, seems good, just a note about the documentation: we can't have images reported at "W: 1614px | H: 760px", they are just too big. An image for the documentation should be at most 700px, or around that size.

Image size reduced, and documentation corrections from @bradh merged.

jodygarnett:
With respect to hardening the regex .. I am not sure if rejecting relative path ``../`` references is useful:

- https://styles.server.net/icons/tree.png
- https://styles.server.net/icons/arrow.png
- https://styles.server.net/sld/forestry/habitat.sld

When evaluating relative paths in an SLD document ``habitat.sld`` above I would expect external graphic ``../../icons/tree.png`` to work?

So the question becomes if the external graphic is determined as a:

- relative to base url: ``https://styles.server.net/sld/forestry/../../icons/tree.png`` or
- normalized url: ``https://styles.server.net/icons/tree.png`` ?

Hardening RegEx:

* no hardening: ``^https?://styles\.server\.net/.*$``
* hardening: ``^https?://styles\.server\.net/((?!\.\./).)*$`` - fails when ``../`` used

Use URL.normalize():

* ``^https?://styles\.server\.net/.*$`` works unmodified

jodygarnett:
Turns out both approaches are needed:

1. URI.normalize() to resolve relative paths prior to checking

   ``https://styles.server.net/sld/forestry/../../icons/tree.png`` normalized ``https://styles.server.net/icons/tree.png``

2. Hardening RegEx to fail any relative paths left over after normalization

   ``https://styles.server.net/sld/forestry/../../../icons/tree.png`` normalized ``https://styles.server.net/../icons/tree.png``

sikeoka:
I did a test and relative paths for icons in dynamic styles don't work for me at all.  It just gives me: `java.net.MalformedURLException: no protocol: foo.png`

jodygarnett:
PR is updated with RegEx hardening, and allowance for URL normalization. 

aside: I updated the URL Check tester to make use of the full set of URL Checks expecting ``http://localhost:8080/geoserver/styles/grass_fill.png`` to validate against ``StyleURLChecker`` (however StyleURLChecker is setup to expect the catalogue to have processed these into a file URL)

aaime:
> I did a test and relative paths for icons in dynamic styles don't work for me at all. It just gives me: `java.net.MalformedURLException: no protocol: foo.png`

Can you describe the use case better? The current code supports relative paths for local styles, and absolute references for dynamic styles (e.g. the cases that have tests in the codebase and examples in the release data directory).

I haven't seen a case for dynamic styles with relative references, is that supposed to work according to the documentation, or was it accidental behavior? Also, what's the expected behavior? 
Relative lookup using the style URL as a base or lookup in the data directory, using the current workspace as a reference?
The first may seem the most reasonable, but the parsing code in GetMapKvpRequestReader is not even getting a reference to the original style location (which might not even be there with SLD_BODY).

I'm overall quite confused, can you show us a specific example of what was working before, and is not working anymore?

jodygarnett:
I believe @sikeoka was testing the use case I described above of a remote SLD file with relative paths to its remote icons. 

While the idea makes sense he checked if the functionality even works. And no it does not work …

jodygarnett:
I believe the generated example is fine, and other examples have been added to the docs to minimize questions. 

@sikeoka are you content that the provided RegEx limits relative the paths as intended?

aaime:
> I believe @sikeoka was testing the use case I described above of a remote SLD file with relative paths to its remote icons.
> 
> While the idea makes sense he checked if the functionality even works. And no it does not work …

Yep, it never worked, as far as I know...

aaime:
I've tried to set up an example but I'm not getting very far. Here is a SLD + external graphic that would work against the poi demo layer:
* https://demo.geo-solutions.it/remote/remote-svg.sld
* https://demo.geo-solutions.it/remote/check.svg

``https://demo.geo-solutions.it`` has been whitelisted on gs-main. Sample request here:
https://gs-main.geosolutionsgroup.com/geoserver/wms?service=WMS&version=1.1.1&request=GetMap&layers=tiger%3Apoi&bbox=-74.0118315772888%2C40.70754683896324%2C-74.00153046439813%2C40.719885123828675&width=641&height=768&srs=EPSG%3A4326&sld=https://demo.geo-solutions.it/remote/remote-svg.sld&format=application/openlayers

I've also tested with a 2.22.1 locally with a similar URL, so that I can compare with a case that does not have the URL checks yet.

Performed tests with variations of the relative URL:
* Using just  ``check.svg`` and "./check.svg", they both fail with a service exception reporting the reference has no protocol. This happens on both gs-main and local 2.22.1
* I've tried to feed a weird URL just to make the protocol check happy ``http://./check.svg``, in both cases I get a map with the fallback gray square. Setting up logging in geotols-developer mode, gs-main complains that ``http://./check.svg`` does not pass the validation (correct), while 2.22.1 complains that "." is not a valid host.

I've then tried also with PNG, with the following:
* https://demo.geo-solutions.it/remote/remote-png.sld
* https://demo.geo-solutions.it/remote/check.png

Results are the same.

Long story short, if you want relative icon references to work in remote styles, you'll first have to implement a mechanism to resolve URLs as relative (meaning, you need to provide the base URL to the graphic factories somehow). 
If you're interested, mind not to break dynamic symbolizers, that is, CQL expressions that can be included in the path and have to be evaluated against the current feature to get the final location (I'd say, that has to happen first, and then process relative URLs).

jodygarnett:
Thank you both for the extensive testing on dynamic sld documentation location.

I still think the normalization code has value and believe it is a better test than just the ``((?!\.\./).)*`` regex on its own.

doc/en/user/source/security/urlchecks.rst
```diff
@@ -3,65 +3,172 @@
 URL Checks
 ==========
 
-The URL checks page controls the checks that are performed on user provided URLs that
+The :guilabel:`URL External Access Checks` page controls the checks that are performed on user provided URLs that
```
aaime, line 6: I keep on being surprised about usage of qualifiers like guilabel... why spent time adding them, when the rendering is confusing at best? It renders like a inline code block, the last think I go searching is GUI elements...

image

doc/en/user/source/security/urlchecks.rst
```diff
@@ -3,65 +3,172 @@
...
 
-More functionality might go under this kind of scrutiny in the future, check this page
-for eventual updates.
+Check this page for any additional remote service access checks added in the future.
 
 Configuration of URL checks
 ---------------------------
 
-The "URL checks" page under the Security section allows configuration of the URL checks.
+Navigate to :menuselection:`Data > URL Checks` page to manage and configure URL Checks.
```
aaime, line 21: This one at least has a sensible rendering, bold with little arrowheads.

src/main/src/main/java/org/geoserver/security/urlchecks/GeoServerURLChecker.java
```diff
@@ -50,12 +52,34 @@ public boolean confirm(String url) {
 
             // if enabled but no checks configured, deny access
             List<AbstractURLCheck> enabledUrlList = getEnabledChecks();
-            if (enabledUrlList.isEmpty()) return false;
-
-            for (AbstractURLCheck u : enabledUrlList) {
-                if (u.confirm(url)) {
-                    LOGGER.log(Level.FINE, () -> "URL " + url + " was matched by " + u);
-                    return true;
+            if (!enabledUrlList.isEmpty()) {
```
aaime, line 55: Oh no, a long nested block to avoid an early exit statement? I'm horrified.... :rofl:  (seriously, nested blocks should not be big).

src/main/src/main/java/org/geoserver/security/urlchecks/GeoServerURLChecker.java
```diff
@@ -50,12 +52,34 @@ public boolean confirm(String url) {
...
+                                                        + url
+                                                        + " was normalized to "
+                                                        + normalized
+                                                        + " and matched by "
+                                                        + u);
+                                return true;
+                            }
+                        }
+                    }
+                } catch (URISyntaxException nonURI) {
```
aaime, line 82: Eaten exception, should at least log something.

src/main/src/main/java/org/geoserver/security/urlchecks/GeoServerURLChecker.java
```diff
@@ -50,12 +52,34 @@ public boolean confirm(String url) {
...
+                            if (u.confirm(normalized)) {
+                                LOGGER.log(
+                                        Level.FINE,
+                                        () ->
+                                                "URL "
+                                                        + url
+                                                        + " was normalized to "
+                                                        + normalized
+                                                        + " and matched by "
+                                                        + u);
```
aaime, line 77: Deep nesting is visible at odds with our coding conventions, please avoid. Thinking out loud, the loop could be a separate method accepting a Function, first you run it with the identity function, then with one normalizing the URL.

That said... this normalization is not where it should be... GeoTools code using URL checks in other applications would end up suffering the same problem. Better to do it in URLCheckers instead?

src/web/security/core/src/main/java/org/geoserver/security/web/url/RegexCheckPage.java
```diff
@@ -90,10 +90,34 @@ private void initUI(IModel<RegexURLCheck> bean) {
...
+     *   <li>Allow anything after {@code ?} query
+     * </ul>
+     *
+     * @return example regular expression
+     */
     private String getDefaultRegex() {
-        return "^" + getOWSURL() + ".*$";
+        String url = getOWSURL();
+
+        url = url.replace("http:", "https?:");
```
aaime, line 110: Disagree... what if a host is only available via http? The default regex should be allowing for both, to give a good example, but don't assume the admin will enter only https urls.

src/web/security/core/src/main/java/org/geoserver/security/web/url/RegexCheckPage.java
```diff
@@ -90,10 +90,34 @@ private void initUI(IModel<RegexURLCheck> bean) {
...
+        url = url.replace(".", "\\.");
+
+        return "^" + url + "((?!\\.\\./).)*(\\?.*)?$";
```
aaime, line 113: It seems this is a match ``http://geoserver.org/test/..`` but I think it should not be

src/web/security/core/src/main/java/org/geoserver/security/web/url/URLChecksPage.java
```diff
@@ -112,14 +117,43 @@ public void onSubmit(AjaxRequestTarget target, Form form) {
...
                             if (check.isEnabled()) {
                                 if (check.confirm(test)) return check;
                             }
                         }
+                        // Check 2: Check using normalized test URL if different
+                        try {
+                            // URL normalization to resolve relative paths prior to checking
+                            String normalized = new URI(test).normalize().toString();
+                            if (!normalized.equals(test)) {
+                                for (AbstractURLCheck check : getUrlCheckDAO().getChecks()) {
```
aaime, line 134: Why go to the DAO again, you already have the list of checks?

aaime (changes requested):
Jody I like the idea of normalization. However I'd suggest pushing it down to GeoTools, so that every application using URL checks can benefit from it.

jodygarnett:
> Jody I like the idea of normalization. However I'd suggest pushing it down to GeoTools, so that every application using URL checks can benefit from it.

I agree that would be preferable, define it as part of the javadoc contract or something.

doc/en/user/source/security/urlchecks.rst
jodygarnett: I always write my page navigation text that way :)

doc/en/user/source/security/urlchecks.rst
aaime: So you're making that effort only because your vendor needs. Makes sense as long as GeoCat personnel does the effort. But I would remove it from the documentation guidelines then...

src/web/security/core/src/main/java/org/geoserver/security/web/url/RegexCheckPage.java
jodygarnett: You are not reading this correctly, this is ensuring that the default allows both?

I am taking the case where ``http:`` and changing it to ``https?:`` (so the ``s`` is optional).

jodygarnett: Do you want to also add so both ``http:`` and ``https:`` are covered:
```
url = url.replace("http:", "https?:");
url = url.replace("https:", "https?:");
```

src/web/security/core/src/main/java/org/geoserver/security/web/url/RegexCheckPage.java
jodygarnett: The initial URL is ``http://localhot:8080/geoserver/ows`` so this turns into ``^https?://localhot:8080/geoserver/ows((?!\\.\\./).)*(\\?.*)?$``

So it should not match ``http://geoserver.org/test/..`` ?

src/web/security/core/src/main/java/org/geoserver/security/web/url/RegexCheckPage.java
aaime: I indeed misread it. Still... it's magic. The admin enters a checker that only allows http, and the implementation also allows https. It should be documented, or even better, just be part of the sample regex, and let the admin decide what to do.

src/web/security/core/src/main/java/org/geoserver/security/web/url/URLChecksPage.java
jodygarnett: I was trying to get all the checks, the geotools ones also.

Previously we only had the geoserver checks.

src/web/security/core/src/main/java/org/geoserver/security/web/url/RegexCheckPage.java
aaime: You're being too literal here, I just typed a test regex and a test string in regex101.com, in Java mode.
``^https?://localhost:8080/geoserver/ows((?!\\.\\./).)*(\\?.*)?$```
matches:
``http://localhost:8080/geoserver/ows/..``

A trailing ``..`` without a "/" after it is allowed. I don't think it should be.

See here for a playground with the above regex and exception: https://regex101.com/r/cHMq3W/1

jodygarnett: Good point!

If I understand the consequence of this ... if a normalization is available it should always be used. So that ``http://localhost:8080/geoserver/ows/..`` normalizes to ``http://localhost:8080/geoserver`` which would indeed fail checks.

jodygarnett:
https://github.com/geoserver/geoserver/blob/main/src/main/src/main/java/org/geoserver/security/urlchecks/StyleURLChecker.java also needs to be updated. It can potentially be bypassed using a `?` or `#` character like: `file:///path/to/some/file#/../../../../path/to/datadir/styles/icon.png` I'm also wondering if StyleURLChecker should limit file extensions although it may not be a huge deal that the XML and style files will pass the check.

After my changes in https://github.com/geotools/geotools/pull/4337 the StyleURLChecker can no longer be bypassed in this manner. Locations are normalized prior to being checked. 
 
> `#` works in Linux and Windows and `?` only works in Linux and you will want to remove those before calling `new File`. `?` is used in parameterized SVG icons but is an illegal Windows filename character so this check could potentially be breaking parameterized SVG in Windows (although unit tests seemed to all pass).

I think the code that processes SVG icons is smart enough to figure that out?

doc/en/user/source/security/urlchecks.rst
jodygarnett: I wonder if we can update the geoserver theme to be based on rtd theme like everything else these days :)

src/web/security/core/src/main/java/org/geoserver/security/web/url/RegexCheckPage.java
jodygarnett: Removed, this is only an example and we have better things to do.

src/web/security/core/src/main/java/org/geoserver/security/web/url/RegexCheckPage.java
jodygarnett: I covered this case in the geotools test cases for https://github.com/geotools/geotools/pull/4337 to ensure it is covered. 

Thanks for thinking through this awkward corner case.

doc/en/user/source/security/urlchecks.rst
aaime: Hum... maybe? We'd need to keep colors and logos, but if that makes full usage of the RST extra qualifiers, why not?

src/web/security/core/src/main/java/org/geoserver/security/web/url/RegexCheckPage.java
jodygarnett: After this change the new default is ``^http://localhost:8080/geoserver/ows((?!\.\./).)*(\?.*)?$``

With normalization handled this can be simplified to ``^http://localhost:8080/geoserver/ows(\?.*)?$``

However I would prefer to simplify to ``^http://localhost:8080/geoserver/ows.*$`` as a better starting point for folks creating their own patterns to lock down external services...

aaime: Which is basically going full circle and using the original regex?

jodygarnett:
Okay all feedback is accounted for, this is ready to go (I do not expect tests to pass as this requires https://github.com/geotools/geotools/pull/4337 to be merged at the same time)

aaime:
The backport to `2.22.x` failed:
```
```
<details>
<summary>stderr</summary>

```
error: could not apply 3e90760b9f... [GEOS-10949] URL check page layout and default ows URL

```
</details>

<details>
<summary>stdout</summary>

```
CONFLICT (modify/delete): src/web/security/core/src/main/java/org/geoserver/security/web/url/RegexCheckPage.java deleted in HEAD and modified in 3e90760b9f ([GEOS-10949] URL check page layout and default ows URL).  Version 3e90760b9f ([GEOS-10949] URL check page layout and default ows URL) of src/web/security/core/src/main/java/org/geoserver/security/web/url/RegexCheckPage.java left in tree.
CONFLICT (modify/delete): src/web/security/core/src/main/java/org/geoserver/security/web/url/URLChecksPage.html deleted in HEAD and modified in 3e90760b9f ([GEOS-10949] URL check page layout and default ows URL).  Version 3e90760b9f ([GEOS-10949] URL check page layout and default ows URL) of src/web/security/core/src/main/java/org/geoserver/security/web/url/URLChecksPage.html left in tree.

```
</details>

```bash
git fetch
# Create a new working tree
git worktree add .worktrees/backport-2.22.x 2.22.x
cd .worktrees/backport-2.22.x
# Create a new branch
git switch --create backport-6916-to-2.22.x
git cherry-pick 3e90760b9f8ec2decc891440a6d6d25651457ebe,3ecb44d899a0ddb7cf4adab8995be59b69a9c0b2,41354ef0660d8bf6698a3baa307c42a90719b51d,021b7225eaf3c661a0de303fd0bd55559ebc0991,8b4c952a5e59e2c918a64ef0e93899ed91123ba0,ee2813a1a6e030cc78c07af46e2d10d2a6105586,df518b7ff3a4a0b83cea3ddce5b197667a0e9dd5,f77e6f8d1d233c965c1560c3d9d65abb5749cd33
# Push it to GitHub
git push --set-upstream origin backport-6916-to-2.22.x
cd ../..
# Delete the working tree
git worktree remove .worktrees/backport-2.22.x
```
Then, create a pull request where the `base` branch is `2.22.x` and the `compare`/`head` branch is `backport-6916-to-2.22.x`.

aaime:
```
```
<details>
<summary>stderr</summary>

```
error: could not apply 3e90760b9f... [GEOS-10949] URL check page layout and default ows URL

```
</details>

<details>
<summary>stdout</summary>

```
CONFLICT (modify/delete): src/web/security/core/src/main/java/org/geoserver/security/web/url/RegexCheckPage.java deleted in HEAD and modified in 3e90760b9f ([GEOS-10949] URL check page layout and default ows URL).  Version 3e90760b9f ([GEOS-10949] URL check page layout and default ows URL) of src/web/security/core/src/main/java/org/geoserver/security/web/url/RegexCheckPage.java left in tree.
CONFLICT (modify/delete): src/web/security/core/src/main/java/org/geoserver/security/web/url/URLChecksPage.html deleted in HEAD and modified in 3e90760b9f ([GEOS-10949] URL check page layout and default ows URL).  Version 3e90760b9f ([GEOS-10949] URL check page layout and default ows URL) of src/web/security/core/src/main/java/org/geoserver/security/web/url/URLChecksPage.html left in tree.

```
</details>

```bash
git fetch
# Create a new working tree
cd .worktrees/backport-2.23.x
# Create a new branch
git switch --create backport-6916-to-2.23.x
git cherry-pick 3e90760b9f8ec2decc891440a6d6d25651457ebe,3ecb44d899a0ddb7cf4adab8995be59b69a9c0b2,41354ef0660d8bf6698a3baa307c42a90719b51d,021b7225eaf3c661a0de303fd0bd55559ebc0991,8b4c952a5e59e2c918a64ef0e93899ed91123ba0,ee2813a1a6e030cc78c07af46e2d10d2a6105586,df518b7ff3a4a0b83cea3ddce5b197667a0e9dd5,f77e6f8d1d233c965c1560c3d9d65abb5749cd33
# Push it to GitHub
git push --set-upstream origin backport-6916-to-2.23.x
cd ../..
# Delete the working tree
```
Then, create a pull request where the `base` branch is `2.23.x` and the `compare`/`head` branch is `backport-6916-to-2.23.x`.

jodygarnett:
@canmehteroglu you are up!

sikeoka:
I did more testing and found that malicious file URLs can bypass `URLCheckers.normalize` by using invalid octet codes in the query or fragment (e.g., `file:///allowed/file/path/../../../etc/passwd?%`).  My recommended fix for that would be to call `getRawUserInfo`, `getRawPath`, `getRawQuery`, and `getRawFragment` instead of `getUserInfo`, `getPath`, `getQuery` and `getFragment` which will attempt to decode those URI sections and throw an exception if it can't be decoded.

As mentioned in the previous comment, the URI path should be decoded before calling `java.net.URI.normalize()` (do not call normalize twice) and the entire location variable should be decoded before calling `Paths.get`.

PR #6919, opened 2023-06-09
[Backport 2.22.x] [GEOS-10994] OOM due to too many dimensions when range requested

aaime:
[GEOS-10994](https://osgeo-org.atlassian.net/browse/GEOS-10994)

Backport #6869
 **Authored by:** @turingtestfail


aaime:
Legit failure due to different mime type returned by earlier Spring version.

PR #6953, opened 2023-06-16
[GEOS-11033] Take some care when updating bounds to protect against missing crs

jodygarnett:
[GEOS-11033](https://osgeo-org.atlassian.net/browse/GEOS-11033)

* [x] If bounds only includes crs, preserve original extent
* [x] if bounds only includes extent update, preserve original crs 

Scripts and code like importer and backup / restore that automate interaction with the catalog should be aware of this change.
 

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/ResourceInfo.java
+++ b/src/main/src/main/java/org/geoserver/catalog/ResourceInfo.java
@@ -197,7 +197,7 @@ public interface ResourceInfo extends CatalogInfo {
     /**
      * Returns the bounds of the resource in lat / lon.
      *
-     * <p>This value represents a "fixed value" and is not calulated on the underlying dataset.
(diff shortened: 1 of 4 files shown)


jodygarnett:
This should be good to go :)

jodygarnett:
Converting to draft as I am concerned that:
a) the the UI provides no ability to change missing crs; 
b) Change would prevent REST API users from publishing when they are unable to determine CRS? While they could use <crs>EPSG:404000</crs> that has not been defined as a requirement of ReferencedEnvelope

GeoSander:
@jodygarnett I (also) doubt if this is a good idea. We probably should not introduce too much "smartness" into the REST API.

jodygarnett:
This PR has been replaced by https://github.com/geoserver/geoserver/pull/6962

PR #6962, opened 2023-06-20
[GEOS-11033] WCS 2.0.1 DescribeCoverage axis order using layer projection policy

jodygarnett:
[GEOS-11033](https://osgeo-org.atlassian.net/browse/GEOS-11033)

The coverage.xml nativeBoundingBox is not required to have an srs. WCS 2.0.1 DescribeCoverage and WCS 1.1 DescribeCoverage assumed this information was available.

This patch makes use of layer projection policy to help determine axis order.

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/impl/ResourceInfoImpl.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/ResourceInfoImpl.java
@@ -265,11 +265,8 @@ public abstract class ResourceInfoImpl implements ResourceInfo {
         } else if (nativeCRS != null
                 && !CRS.equalsIgnoreMetadata(nativeBox.getCoordinateReferenceSystem(), nativeCRS)) {
             LOGGER.log(Level.FINE, "The native bounding box srs does not match native crs");
-            nativeBox = ReferencedEnvelope.create(nativeBox, nativeCRS);
         }
 
-        // make sure that in no case the actual field value is returned to the client, this
-        // is not a getter, it's a derivative, thus ModificationProxy won't do a copy on its own
         if (!CRS.equalsIgnoreMetadata(declaredCRS, nativeCRS)
                 && php == ProjectionPolicy.REPROJECT_TO_DECLARED) {
             if (nativeBox.getCoordinateReferenceSystem() == null) {
@@ -278,8 +275,14 @@ public abstract class ResourceInfoImpl implements ResourceInfo {
                         "Unable to reproject to declared crs (native bounding box srs and native crs are not defined)");
                 return null;
             }
+            // transform below makes a copy, in no case the actual field value is returned to the
+            // client, this
+            // is not a getter, it's a derivative, thus ModificationProxy won't do a copy on its own
             return nativeBox.transform(declaredCRS, true);
         } else if (php == ProjectionPolicy.FORCE_DECLARED) {
+            // create below makes a copy, in no case the actual field value is returned to the
+            // client, this
+            // is not a getter, it's a derivative, thus ModificationProxy won't do a copy on its own
             return ReferencedEnvelope.create((Envelope) nativeBox, declaredCRS);
         } else {
             if (nativeBox == null || nativeBox.getCoordinateReferenceSystem() == null) {
@@ -288,6 +291,9 @@ public abstract class ResourceInfoImpl implements ResourceInfo {
                         "Unable to determine native bounding crs (both native bounding box srs and native crs are not defined)");
                 // return null;
             }
+            // create below makes a copy, in no case the actual field value is returned to the
+            // client, this
+            // is not a getter, it's a derivative, thus ModificationProxy won't do a copy on its own
             return ReferencedEnvelope.create(nativeBox);
         }
     }
--- a/src/main/src/main/java/org/geoserver/catalog/impl/ResourceInfoImpl.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/ResourceInfoImpl.java
@@ -241,30 +241,55 @@ public abstract class ResourceInfoImpl implements ResourceInfo {
         CoordinateReferenceSystem nativeCRS = getNativeCRS();
         ProjectionPolicy php = getProjectionPolicy();
 
-        ReferencedEnvelope nativeBox = this.nativeBoundingBox;
+        ReferencedEnvelope nativeBox = getNativeBoundingBox();
         if (nativeBox == null) {
             // back project from lat lon
             try {
-                nativeBox = getLatLonBoundingBox().transform(declaredCRS, true);
+                if (declaredCRS != null) {
+                    nativeBox = getLatLonBoundingBox().transform(declaredCRS, true);
+                } else {
+                    LOGGER.log(
+                            Level.WARNING,
+                            "Failed to derive native bbox, there is no declared CRS provided");
+                    return null;
+                }
             } catch (Exception e) {
                 LOGGER.log(Level.WARNING, "Failed to derive native bbox from declared one", e);
                 return null;
             }
-        }
-
-        ReferencedEnvelope result;
-        if (!CRS.equalsIgnoreMetadata(declaredCRS, nativeCRS)
-                && php == ProjectionPolicy.REPROJECT_TO_DECLARED) {
-            result = nativeBox.transform(declaredCRS, true);
-        } else if (php == ProjectionPolicy.FORCE_DECLARED) {
-            result = ReferencedEnvelope.create((Envelope) nativeBox, declaredCRS);
-        } else {
-            result = nativeBox;
+        } else if (nativeBox.getCoordinateReferenceSystem() == null) {
+            if (nativeCRS != null) {
+                // Ensure provided nativeBox using nativeCRS
+                nativeBox = ReferencedEnvelope.create(nativeBox, nativeCRS);
+            }
+        } else if (nativeCRS != null
+                && !CRS.equalsIgnoreMetadata(nativeBox.getCoordinateReferenceSystem(), nativeCRS)) {
+            LOGGER.log(Level.FINE, "The native bounding box srs does not match native crs");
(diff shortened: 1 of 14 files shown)


jodygarnett:
Converting this to a draft while I look at WCS 1.1 also

jodygarnett:
This PR replaces https://github.com/geoserver/geoserver/pull/6953

davidblasby (approved):
Looks good - test cases and comments!

jodygarnett:
I am discouraged by the overlap between ResourceInfo.boundingBox() and WCSUtils.toNativeBounds( CoverageInfo ).

If boundingBox() was more safe, with less assumptions it may be suitable for use with WCS.

src/wcs2_0/src/main/java/org/geoserver/wcs2_0/GetCoverage.java
```diff
@@ -1605,7 +1605,7 @@ private Map<String, InterpolationPolicy> extractInterpolation(
             CoverageInfo ci, Map<String, ExtensionItemType> extensions) {
         // preparation
         final Map<String, InterpolationPolicy> returnValue = new HashMap<>();
-        final Envelope envelope = ci.getNativeBoundingBox();
+        final Envelope envelope = WCSUtils.toNativeBounds(ci);
```
jodygarnett, line 1608: By adding some try/catch statements ``ci.boundingBox()`` can also be used here; however it is willing to make a lot more assumptions to produce a native bounding box.

WCSUtils.toNativeBounds(ci) is more of a minimal drop-in replacement.

jodygarnett:
This pull-request solves the problem I have encountered, however I would like an opinion/feedback on use of ResourceInfo.boundingBox() vs WCSUtil.toNativeBounds(CoverageInfo).

If ResourceInfo.boundingBox() is a suitable replacement we can avoid adding and maintaining a new method.

jodygarnett:
rebasing to get the gs-metadata fix; I trust automated tests will pass now

src/main/src/main/java/org/geoserver/catalog/impl/ResourceInfoImpl.java
```diff
@@ -241,30 +241,55 @@ public ReferencedEnvelope boundingBox() throws Exception {
...
             }
+        } else if (nativeBox.getCoordinateReferenceSystem() == null) {
+            if (nativeCRS != null) {
+                // Ensure provided nativeBox using nativeCRS
+                nativeBox = ReferencedEnvelope.create(nativeBox, nativeCRS);
+            }
+        } else if (nativeCRS != null
+                && !CRS.equalsIgnoreMetadata(nativeBox.getCoordinateReferenceSystem(), nativeCRS)) {
+            LOGGER.log(Level.FINE, "The native bounding box srs does not match native crs");
+            nativeBox = ReferencedEnvelope.create(nativeBox, nativeCRS);
```
aaime, line 268: I have doubts about this bit... those two are initially set up together, but then one can request to recompute the bbox through the REST API I believe? And then the bbox might end up being more up-to-date than the native CRS?
I haven't checked in detail.

src/main/src/main/java/org/geoserver/catalog/impl/ResourceInfoImpl.java
```diff
@@ -241,30 +241,55 @@ public ReferencedEnvelope boundingBox() throws Exception {
...
+                        Level.WARNING,
+                        "Unable to determine native bounding crs (both native bounding box srs and native crs are not defined)");
+                // return null;
+            }
+            return ReferencedEnvelope.create(nativeBox);
         }
-
-        // make sure that in no case the actual field value is returned to the client, this
-        // is not a getter, it's a derivative, thus ModificationProxy won't do a copy on its own
-        return ReferencedEnvelope.create(result);
```
aaime, line 267: This was an important cue to future developers, we should not lose it (I'm talking about the comment regarding ModificationProxy)

src/main/src/test/java/org/geoserver/catalog/impl/CatalogBuilderTest.java
```diff
@@ -323,6 +323,28 @@ public void testNativeBoundsDefensiveCopy() throws Exception {
...
+        // simulate missing srs
+        ci.setNativeBoundingBox(ReferencedEnvelope.create(bbox, null));
+        assertEquals(ci.getNativeCRS(), ci.boundingBox().getCoordinateReferenceSystem());
+
+        // simulate everything being missing
+        ci.setNativeBoundingBox(null);
+        ci.setSRS(null);
+
+        // this will attempt to back-project form LatLon
+        assertNull(ci.boundingBox());
```
aaime, line 345: Hmm... an exception might be more useful here, we'd just risk getting a NPE in another part of the code and have trouble tracking it down?

src/wcs/src/main/java/org/vfny/geoserver/util/WCSUtils.java
```diff
@@ -878,4 +878,43 @@ private static GridGeometry2D simpleGridGeometryFit(
...
+     * are only determined using native bounding box and native crs information. It is not willing
+     * to back-transform from lat lon bounding box. It also unwilling to grow the native bounding
+     * box while re-projecting to declared crs.
+     *
+     * @param meta CoverageInfo configuration
+     * @return Native bounds, using native bounding box crs
+     * @throws IllegalStateException if coverage info does not define a native bounding box with
+     *     native crs information
+     */
+    public static ReferencedEnvelope toNativeBounds(CoverageInfo meta) {
```
aaime, line 899: If at all possible I'd avoid code duplication and improve ResourceInfo.boundingBox instead (which also has much larger exposure to the code base, thus better overall testing)

src/wcs1_1/src/test/java/org/geoserver/wcs/DescribeCoverageTest.java
```diff
@@ -171,6 +188,37 @@ public void testDescribeDemCoverageKvp() throws Exception {
...
     }
 
+    @Test
+    public void testDescribeNoEnvelopeSRS() throws Exception {
+        Document dom =
+                getAsDOM(
+                        BASEPATH
+                                + "?request=DescribeCoverage&service=WCS&version=1.1.1&identifiers="
+                                + getLayerId(NO_ENVELOPE_SRS));
+        print(dom);
```
aaime, line 198: We normally comment out these guys. They do not dump anything during the maven build, but show up in a run from IDE (which can be annoying as we have test classes with many tests, due to the high start-up time of each).

src/wcs2_0/src/test/java/org/geoserver/wcs2_0/GetCoverageTest.java
```diff
@@ -58,6 +60,8 @@ public class GetCoverageTest extends WCSTestSupport {
     private static final QName RAIN = new QName(MockData.SF_URI, "rain", MockData.SF_PREFIX);
     private static final QName TIMESERIES =
             new QName(MockData.SF_URI, "timeseries", MockData.SF_PREFIX);
+    public static QName NO_ENVELOPE_SRS =
```
aaime, line 63: Bit confused here... I thought the case was about not having a usable CRS rather than not having a native bbox?
Also, I believe you're contributing an improvement to read the CRS from ESRI WKT, if that's committed are the tests still proving anything here?

Oh, got it now... it has an envelope without CRS. Might want to add a clarifying comment here.

aaime (changes requested):

src/main/src/main/java/org/geoserver/catalog/impl/ResourceInfoImpl.java
jodygarnett: The UI only allows changing of nativeCRS so I have to trust that one :)

aaime: The UI does not, at least not in core GeoServer. Here is a screenshot from ArcSample: 

image

Try typing into the "native srs" field... it won't work. Only declared SRS can be modified.
The UI used to have a different color for the native SRS making it clear that it was not editable, it might have broken with the CSS update?

src/main/src/test/java/org/geoserver/catalog/impl/CatalogBuilderTest.java
jodygarnett: The boundingBox() method is already returning null if it cannot figure things out so I figured that this would be okay?

src/main/src/main/java/org/geoserver/catalog/impl/ResourceInfoImpl.java
jodygarnett: The line is included further up

src/main/src/main/java/org/geoserver/catalog/impl/ResourceInfoImpl.java
jodygarnett: @MichelGabriel see above!

jodygarnett: > I have doubts about this bit... those two are initially set up together, but then one can request to recompute the bbox through the REST API I believe? And then the bbox might end up being more up-to-date than the native CRS?

You are correct:
```
        if (fieldsToCalculate.contains("nativebbox")) {
            CatalogBuilder builder = new CatalogBuilder(catalog);
            try {
                message.setNativeBoundingBox(builder.getNativeBounds(message));
            } catch (IOException e) {
                String errorMessage = "Error while calculating native bounds for layer: " + message;
                throw new RestException(errorMessage, HttpStatus.INTERNAL_SERVER_ERROR, e);
            }
        }
```

No attempt to update the native crs to match newly calculated native bounds.

jodygarnett:
Feedback has been addressed, this should be squashed and merged to avoid complicated commit history.

PR #6977, opened 2023-06-28
[GEOS-10901] GetCapabilities lists the same style multiple times when…

pratikbandalg:
[GEOS-10901](https://osgeo-org.atlassian.net/browse/GEOS-10901)

… used as both a default and alternate style

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/LayerInfo.java
+++ b/src/main/src/main/java/org/geoserver/catalog/LayerInfo.java
@@ -5,6 +5,7 @@
  */
 package org.geoserver.catalog;
 
+import java.util.HashSet;
 import java.util.Set;
 
 /**
@@ -65,6 +66,18 @@ public interface LayerInfo extends PublishedInfo {
      *     inverse="resourceInfo:org.geoserver.catalog.StyleInfo"
      */
     Set<StyleInfo> getStyles();
+
+    /**
+     * The non default styles available for the layer.
+     *
+     * @return non default Style Information
+     */
+    default Set<StyleInfo> styles() {
+        HashSet<StyleInfo> styles = new HashSet<>();
+        if (getStyles() != null) styles.addAll(getStyles());
+        if (getDefaultStyle() != null) styles.remove(getDefaultStyle());
+        return styles;
+    }
 
     /**
      * The resource referenced by this layer.
--- a/src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/CapabilitiesTest.java
+++ b/src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/CapabilitiesTest.java
@@ -405,6 +405,16 @@ public class CapabilitiesTest extends WMSTestSupport {
         assertTrue(href.contains("GetLegendGraphic"));
         assertTrue(href.contains("layer=cdf%3AFifteen"));
         assertTrue(href.contains("style=point"));
+
+        // Default style set to point
+        layer.setDefaultStyle(pointStyle);
+        getCatalog().save(layer);
+        Document document = getAsDOM("wms?service=WMS&request=getCapabilities&version=1.1.1", true);
+        assertXpathEvaluatesTo("1", "count(//Layer[Name='cdf:Fifteen'])", document);
+        assertXpathEvaluatesTo("1", "count(//Layer[Name='cdf:Fifteen']/Style)", document);
+        // getStyles() has the default style, styles() doesnt have default style
+        assertTrue(layer.getStyles().contains(pointStyle));
+        assertFalse(layer.styles().contains(pointStyle));
     }
 
     @Test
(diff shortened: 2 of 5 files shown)


src/wms/src/test/java/org/geoserver/wms/wms_1_1_1/CapabilitiesTest.java
```diff
@@ -405,6 +405,16 @@ public void testAlternateStyles() throws Exception {
         assertTrue(href.contains("GetLegendGraphic"));
```
nmco, line 405: Can you add a style for WMS 1.3.0.

aaime: You mean a test?

src/main/src/main/java/org/geoserver/catalog/impl/LayerInfoImpl.java
```diff
@@ -184,13 +184,16 @@ public Set<StyleInfo> getStyles() {
...
+                styles.remove(this.defaultStyle);
+                return remoteStyles;
+            } else if (LOGGER.isLoggable(Level.FINE)) {
                 LOGGER.fine(
                         "No Default Styles found on cascaded WMS Resource"
                                 + getResource().getName());
+            }
         }
-
+        if (styles != null) styles.remove(this.defaultStyle);
```
aaime, line 196: @jodygarnett what do you think of this change?

On one side, it's the most confined change to solve the problem, and auto-fixes a situation that seems just broken. On the other hand, it's changing the configuration.

Maybe compromise, and return a copy of the styles structure, that has the potential default style duplication removed? Just thinking out loud.

jodygarnett: This "conflict" between the default style and the alternate style list is just a WMS protocol concern - so yes we should make a specific method to make this easier for WMS protocol - and leave the configuration as is.

* I know I have personally configured a lot of servers with several alternate styles, and then completely independently adjusted the default style. It would be frustrating if when trying different default styles the alternate styles were modified. :)
* It is only really the WMS protocol that has this limitation, the ogcapi-api situation may be different.

Can you confirm that the idea of a default style is just a GeoServer configuration idea, the protocol just has an empty ``style=`` and we wanted a way to indicate what was to be used...

Let me check:

> 7.2.4.6.5 Style
> 
> Zero or more Styles may be advertised for a Layer or collection of layers using <Style> elements, each of which shall have <Name> and <Title> elements. The style's Name is used in the Map request STYLES parameter. The Title is a human-readable string. If only a single style is available, that style is known as the “default” style and need not be advertised by the server.

So the idea of a default style is in the spec, and indicates when we could skip listing the style.

jodygarnett: Andrea I like your proposed approach, it could be a utility method, or added to LayerInfo interface as a default method:

```
default Set<StyleInfo> toStyles(){
    HashSet<StyleInfo> styles = new HashSet<>();
    if (getStyles() != null) styles.addAll(getStyles());
    if (getDefaultStyle() != null) styles.remove(getDefaultStyle());
    return styles;
}
```

jodygarnett: The PR looks to just cover external styles from WMSLayerInfo; is it worth fixing this for all layers?

aaime: @jodygarnett if we want to follow pattern with the other interfaces, it should be called just "styles()", all other interfaces remove the getter prefix to indicate that the method returns derivative information (e.g., ``FeatureTypeInfo.attributes()`` vs ``FeatureTypeInfo.getAttributes()``).

jodygarnett: I agree `styles()` would be appropriate.

jodygarnett:
@pratikbandalg that looks much better and is a clean change. I have triggered a run to ensure the QA checks are performed. 

Thank you for your work.

src/main/src/main/java/org/geoserver/catalog/LayerInfo.java
```diff
@@ -66,6 +67,20 @@ enum WMSInterpolation {
...
     Set<StyleInfo> getStyles();
 
+    /**
+     * The non default styles available for the layer.
+     *
+     * @uml.property name="styles"
+     * @uml.associationEnd multiplicity="(0 -1)" container="java.util.Set"
+     *     inverse="resourceInfo:org.geoserver.catalog.StyleInfo"
+     */
+    default Set<StyleInfo> styles() {
```
jodygarnett, line 77: I do not think we need a UML annotation here - this is a query method that does not store any information on its own. 

You may also find that the QA check requires a return javadoc in order to pass.

pratikbandalg: Updated Javadoc with return param

PR #6979, opened 2023-06-28
[GEOS-11048] Improve URL checking

sikeoka:
[GEOS-11048](https://osgeo-org.atlassian.net/browse/GEOS-11048)

This PR contains the following changes:
1. StyleURLChecker removes the URL query and fragment before doing its check.
1. URLChecker is applied to the WMS GetLegendGraphic SLD parameter.
1. Updated documentation to mention GetFeatureInfo and GetLegendGraphic requests.

  

# Checklist

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/legendgraphic/GetLegendGraphicKvpReader.java
+++ b/src/wms/src/main/java/org/geoserver/wms/legendgraphic/GetLegendGraphicKvpReader.java
@@ -50,6 +50,8 @@ import org.geoserver.wms.WMS;
 import org.geoserver.wms.capabilities.CapabilityUtil;
 import org.geotools.coverage.grid.io.GridCoverage2DReader;
 import org.geotools.coverage.util.FeatureUtilities;
+import org.geotools.data.ows.URLCheckerException;
+import org.geotools.data.ows.URLCheckers;
 import org.geotools.data.simple.SimpleFeatureCollection;
 import org.geotools.data.util.NullProgressListener;
 import org.geotools.factory.CommonFactoryFinder;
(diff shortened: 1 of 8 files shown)


src/wms/src/main/java/org/geoserver/wms/legendgraphic/GetLegendGraphicKvpReader.java
```diff
@@ -728,12 +730,15 @@ private Style findStyle(String styleName, Style[] styles) throws NoSuchElementEx
     private Style[] loadRemoteStyle(String sldUrl) throws ServiceException {
         try {
             URL url = new URL(sldUrl);
+            URLCheckers.confirm(url);
```
jodygarnett, line 733: Check string directly before creating URL, since URLCheckers.confirm(URL) just calls out to external form ...
```suggestion
            URLCheckers.confirm(sldUrl);
            URL url = new URL(sldUrl);
```

jodygarnett (changes requested):
This is looking better and better - thank you!

src/wms/src/main/java/org/geoserver/wms/legendgraphic/GetLegendGraphicKvpReader.java
jodygarnett: Note the change of order, to do the confirm(sldUrl) before creating a new URL, is on purpose.

See https://osgeo-org.atlassian.net/browse/GEOS-10384 for context.

sikeoka: I made the change.

PR #6986, opened 2023-07-05
[GEOS-10856] geoserver monitor plugin - scaling troubles

hyperman2:
[GEOS-10856](https://osgeo-org.atlassian.net/browse/GEOS-10856)

  

# Checklist

### DIFF
--- a/doc/en/user/source/extensions/monitoring/configuration.rst
+++ b/doc/en/user/source/extensions/monitoring/configuration.rst
@@ -136,6 +136,18 @@ is added::
 
    /wfs
 
+Monitoring threads
+------------------
+You can choose the number of post processor threads by configuring the ``postprocessorthreads``
+property in the ``monitor.properties`` file.  The default is 2.
+
+DNS cache configuration
+-----------------------
+The reverseDNS post processor caches its result.  You can modify the cache configuration
+by configuring the ``dnscacheconfiguration`` property in the ``monitor.properties`` file.
+The default policy is ``expireAfterWrite=15m`` .  Consult the `guava cache builder documentation <https://guava.dev/releases/32.0.0-jre/api/docs/com/google/common/cache/CacheBuilderSpec.html>`_
+for all possibilities.
+
 How to determine the filter path
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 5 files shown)


doc/en/user/source/extensions/monitoring/configuration.rst
```diff
@@ -136,6 +136,18 @@ is added::
 
    /wfs
 
+Monitoring threads
+------------------
+You can choose the number of post processor threads by configuring the ``postprocessorthreads``
```
aaime, line 141:
```suggestion
You can choose the number of post processor threads by configuring the ``postProcessorThreads``
```

The file is using camelCase for all other properties.

doc/en/user/source/extensions/monitoring/configuration.rst
```diff
@@ -136,6 +136,18 @@ is added::
...
+property in the ``monitor.properties`` file.  The default is 2.
+
+DNS cache configuration
+-----------------------
+The reverseDNS post processor caches its result.  You can modify the cache configuration
+by configuring the ``dnscacheconfiguration`` property in the ``monitor.properties`` file.
```
aaime, line 147: Same as above

```suggestion
by configuring the ``dnsCacheConfiguration`` property in the ``monitor.properties`` file.
```

aaime (changes requested):
Looks good in general, see comments below.

src/extension/monitor/core/src/main/java/org/geoserver/monitor/MonitorConfig.java
```diff
@@ -65,6 +65,10 @@ public static enum BboxMode {
     Exception error;
     private GeoServerResourceLoader loader;
 
+    static final int POSTPROCES_THREADS_DEFAULT = 2;
+
+    static final String DNS_CACHE_DEFAULT = "expireAfterWrite=15m";
```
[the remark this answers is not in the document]
hyperman2: Thanks, I'll look at this in the evening

Regarding DNS_CACHE_DEFAULT: I see a few options:  
1) Change to "expireAfterWrite=15m,softValues" should provide the same behaviour as soft references.
2) Change to "expireAfterWrite=15m,maximumSize=100" should provide the same maximum as in the current code.  

As the current code only reads the cache, never writes it, the current cache is always empty, so I can't look at what happens right now.    I presume soft refs are more heavy on the GC as they need special handling, and the cost of a String is very low, so I propose option 2, maybe even raising the maximum to 1000.

hyperman2:
Sorry, didn't see the 'Request re-review button'.  All remarks have been adapted

aaime (approved):

hyperman2:
Thank you very much

PR #7021, opened 2023-07-11
DRAFT: Add option to disable GetFeatureInfo transforming raster layers

sikeoka:
NOTE: This is just a draft PR to get some feedback on the concept and is not intended to be merged. If the idea is acceptable, a separate PR will be created to replace this one that will have the UI updates, doc updates and new unit tests but I didn't want to work on a full PR if you guys don't like this. A GeoTools PR will also be needed since SLD vendor options are added as String constants to the relevant GeoTools style interface.

This PR adds a WMS service setting and an SLD FeatureTypeStyle vendor option to disable applying rendering transformations to raster layers for WMS GetFeatureInfo requests. This was the behavior prior to GeoServer 2.21.0 which changed the behavior to always apply rendering transformations to raster layers. This PR provides a way to revert to the old behavior on a global, per-workspace or per-style basis. Only raster layer sources are affected by this PR.

There are three potential issues with the current behavior:

1. There may use cases where it is preferred for GetFeatureInfo to return the underlying raster data rather than the rendering transformation output. My customers actually only want to see the non-transformed data.
2. The rendering transformation can significantly impact the performance of WMS GetFeatureInfo requests. I have cases where the requests can be more than 50 times slower with the transformation compared to a regular RasterSymbolizer style even I though I completely hide the transformed data from users.
3. There are cases where rendering transformations (particularly raster-to-vector) change the namespace/prefix of the output which makes it not work with workspace, store or layer-specific FreeMarker templates.

This is the workaround in YSLD format that I've been using to get the raster data using a dummy raster symbolizer but it still has significant performance issues because of the transformation:
```
feature-styles:
- transform:
  # rest of main style...
- rules:
  - symbolizers:
    - raster:
        opacity: 0.0
  x-inclusion: mapOnly
```

I would like to get this into the 2.22.5 release if possible since I am supporting customers still using Java 8.

  

# Checklist


aaime:
Some feedback:
* According to the specification "The GetFeatureInfo operation is designed to provide clients of a WMS with more information about *features in the pictures of maps* that were returned by previous Map requests". The features the user sees are the transformed ones, not the original ones. I get there might be a need to just return the original data, but I'd also add a warning in the documentation that the behavior might not be as the specification intended it.
* I like the idea of having a style-specific flag to disable the RT in feature info, that's a place where there is enough context to decide whether to return the original data vs the transformed data. On the downside, "mapOnly" now indicates a symbolizer that should not be included in the legend, but that is used in both GetMap/GetFeatureInfo... not sure how you intend to go from there to "use in map and legend, but not in feature info". I can't make sense of the YAML/YSLD ... at which level is that x-include being used? Is that inside a rule? However, it should be excluding the transformation, not the symbolizer itself (if the portion that renders the data is excluded, nothing should be returned by feature info).
* The global flag is blunt, but it can be seen as restoring previous behavior, and does not have the same complications as above.

sikeoka:
Regarding the YSLD example, the original style has a single FeatureTypeStyle with a Transformation and I didn't include the rest of the main style in the example. As a workaround to get the raw raster value in the GetFeatureInfo response, I added a second FeatureTypeStyle with a 0 opacity (so nothing extra is actually drawn on the map) and the mapOnly inclusion is on this dummy FeatureTypeStyle so that it doesn't change the legend either. Moving it onto the rule or symbolizer would probably work too since there's only one of everything. Custom FreeMarker templates hide the transformed output so the user doesn't see it but the transformation is still being execute which slows down the request.

I do agree that that the freemarker template issue should be a separate bug (although I'm actually relying on this bug for my workaround) but it would be a lot of work to test and fix every rendering transformation with this issue (vector-to-raster transformations have this problem too).

PR #7027, opened 2023-07-13
[GEOS-11074] GeoFence may not load property file at boot

etj:
[GEOS-11074](https://osgeo-org.atlassian.net/browse/GEOS-11074)

The values in `geofence/geofence-server.properties` are not always read into the configuration class.

It happens randomly, so it seems it may a be a race condition.

The proposed fix is to programmatically load the property file using an InitializingBean:afterPropertiesSet() callback. In this way we'll also provide an entrypoint method to reload the configuration on request -- pls note that the GeoFence plugin is not realoading the configuration at runtime yet.

The basic changes are only related to the adding of the `afterPropertiesSet()` and `loadConfiguration()`, anyway prop names and save procedure have been refactored a lot, in order to allow furter clean up ASAP; we dont want to backport future cleanup if not required..

# Checklist

### DIFF
--- a/src/extension/geofence/geofence/src/main/java/org/geoserver/geofence/config/GeoFenceConfigurationManager.java
+++ b/src/extension/geofence/geofence/src/main/java/org/geoserver/geofence/config/GeoFenceConfigurationManager.java
@@ -8,15 +8,18 @@ import java.io.BufferedWriter;
 import java.io.IOException;
 import java.io.OutputStreamWriter;
 import java.io.Writer;
+import java.util.Properties;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import org.apache.commons.lang3.BooleanUtils;
 import org.geoserver.geofence.GeofenceAccessManager;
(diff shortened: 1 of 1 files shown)


aaime:
Changes look reasonable. Some doubts:

* There is quite a bit of new code for loading properties... before it was leveraging Spring property configuration instead? Is the GeoFencePropertyPlaceholderConfigurer class still needed?
* For extension modules, a test verifying the changes is normally required, or an explanation must be provided, to show why it's impossible to write one.

etj:
As explained in GEOS-11074 the problem is a race condition, and cannot be replicated or tested properly.

The class is still a PropertyPlaceholderConfigurer in order not to break any compatibility; the added logic forces the property assignment even if the propertyConfigurer fails to do its work.

At the moment this fix has been backported by hand to a 2.20.7, and is happily running.
I'd leave this commit as is, since it's running properly, and improve its implementation (remove propertyconfigurer inheritance, split the properties parsing and setting in proper places) in a future work.

aaime (approved):
Fair enough, thank you!

PR #7036, opened 2023-07-20
[GEOS-10248] Check WPSInitializer interaction with reload

jodygarnett:
[GEOS-10248](https://osgeo-org.atlassian.net/browse/GEOS-10248)

 
During training / workshops I like to use the WPS module as an example of installing an extension. Sadly when this is done reloading the configuration shows a WPSInitialization error (as the listener cannot find a WPSInfo until it is available).

This pull request asks WPSInitializer to stop reinitializing itself during each global configuration event during a reload.

# Checklist


jodygarnett:
@aaime sharing this for feedback, as it is not worth writing a test case if you can think of a better / more elegant way to handle this....

jodygarnett:
It works. It is not pretty but it works.

aaime (approved):
Yep, seems sane, +1.

PR #7038, opened 2023-07-21
Clarification concerning layers security

pratikbandalg:
# Checklist

### DIFF
--- a/doc/en/user/source/security/layer.rst
+++ b/doc/en/user/source/security/layer.rst
@@ -159,10 +159,12 @@ The mapping of roles to permissions is as follows:
      - r/w
      - r
    * - (All other users)
+     - (none)
      - r
      - r
      - r
-     - r
+
+.. note:: Specific workspace rule ``private.*.r=TRUSTED_ROLE`` will take precedence over the more generic rule ``*.*.r=*``. When a request is made to read a layer ``private:vulnerable_infrastructure`` the most specific rule available is used to control access. In this case the workspace rule ``private.*.r=TRUSTED_ROLE`` is the most specific and only users that have TRUSTED_ROLE will be granted ``r`` access and be able to read the ``private:vulnerable_infrastructure`` layer.  Other users that do not have the TRUSTED_ROLE will not be granted ``r`` access and will be unable to access the ``private:vulnerable_infrastructure`` layer.
 
 Locking down GeoServer
 ~~~~~~~~~~~~~~~~~~~~~~


doc/en/user/source/security/layer.rst
```diff
@@ -164,6 +164,9 @@ The mapping of roles to permissions is as follows:
      - r
      - r
 
+.. note:: Specific workspace rule will take precedence over the more generic rule.When *.*.r = * is set and more specific rule for workspace private is assigned as private.*.r = TRUSTED_ROLE,
+	Then specific rule for workspace would take precedence. Hence private workspace layers would be read by TRUSTED_ROLE users only and other users wont be able to read private workspace layers:
+
```
jodygarnett, line 169: The clarification being provided here makes me more confused.  I think this could be better communicated with an example; or a diagram.

jodygarnett, line 169:
```suggestion
.. note:: Specific workspace rule ``private.*.r=TRUSTED_ROLE`` will take precedence over the more generic rule ``*.*.r=*``. When a request is made to read a layer ``private:vulnerable_infrastructure`` the most specific rule available is used to control access. In this case the workspace rule ``private.*.r=TRUSTED_ROLE`` is the most specific and only users that have TRUSTED_ROLE will be granted ``r`` access and be able to read the ``private:vulnerable_infrastructure`` layer.  Other users that do not have the TRUSTED_ROLE will not be granted ``r`` access and will be unable to access the ``private:vulnerable_infrastructure`` layer.
```

pratikbandalg: @jodygarnett : I have updated the table and notes accordingly

jodygarnett: Thanks!

doc/en/user/source/security/layer.rst
```diff
@@ -159,10 +159,12 @@ The mapping of roles to permissions is as follows:
      - r/w
      - r
    * - (All other users)
+     - (none)
      - r
      - r
      - r
-     - r
+
+.. note:: Specific workspace rule ``private.*.r=TRUSTED_ROLE`` will take precedence over the more generic rule ``*.*.r=*``. When a request is made to read a layer ``private:vulnerable_infrastructure`` the most specific rule available is used to control access. In this case the workspace rule ``private.*.r=TRUSTED_ROLE`` is the most specific and only users that have TRUSTED_ROLE will be granted ``r`` access and be able to read the ``private:vulnerable_infrastructure`` layer.  Other users that do not have the TRUSTED_ROLE will not be granted ``r`` access and will be unable to access the ``private:vulnerable_infrastructure`` layer.
```
jodygarnett, line 167: I guess I am not american so I only just noticed that ``congress_district`` must be a layer name in the table.

jodygarnett (approved):

PR #7049, opened 2023-07-31
[GEOS-11087] Fix IsolatedCatalogFacade unnecessary performance overhead

groldan:
[GEOS-11087](https://osgeo-org.atlassian.net/browse/GEOS-11087)

* Filter the iterator created by list(...) directly instead of building an ArrayList first
* Apply in-process filter to iterator in count() only if required, delegate to the proxied facade's count() method if `Dispatcher.REQUEST.get() == null`

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/impl/IsolatedCatalogFacade.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/IsolatedCatalogFacade.java
@@ -4,7 +4,8 @@
  */
 package org.geoserver.catalog.impl;
 
-import java.util.ArrayList;
+import com.google.common.base.Predicate;
+import com.google.common.collect.Iterators;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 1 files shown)


src/main/src/main/java/org/geoserver/catalog/impl/IsolatedCatalogFacade.java
```diff
@@ -670,15 +673,19 @@ private boolean canSeeWorkspace(WorkspaceInfo workspace) {
...
-                unwrapped.stream()
-                        .filter(store -> filter.apply(store) != null)
-                        .collect(Collectors.toList()),
-                type);
+        if (isOwsRequest()) {
+            // unwrap the catalog objects list
+            List<T> unwrapped = ModificationProxy.unwrap(objects);
+            // filter the non visible catalog objects and wrap the resulting list with a
+            // modification
+            // proxy
```
aaime, line 681:
```suggestion
            // modification proxy
```

aaime (commented):
Looks good to me. Please play nice with the automated formatter :rofl:

src/main/src/main/java/org/geoserver/catalog/impl/IsolatedCatalogFacade.java
groldan: oops. Fixed.

aaime (approved):
Yep, looks good now, thanks!

PR #7050, opened 2023-07-31
[GEOS-11088] 30% of startup time spent in XstreamPersister's CRSConverter

groldan:
[GEOS-11088](https://osgeo-org.atlassian.net/browse/GEOS-11088)

Significant overhead added by `CRSConverter` while starting up off a datadir with a large number of layers, parsing the same small number of CRS WKT over and over.

Add a short-lived cache to `CRSConverter` to limit the number of calls to `CRS.parseWKT()`. Guava's cache also handles concurrency out of the box.

The cache contains up to 10 entries and values are SoftReferences.

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
+++ b/src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
@@ -5,6 +5,9 @@
  */
 package org.geoserver.config.util;
 
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.Multimap;
 import com.thoughtworks.xstream.XStream;
@@ -54,6 +57,8 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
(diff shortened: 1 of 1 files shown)


aaime:
In the PSC meeting, we were wondering about pushing cachiNg down in CRS.parseWKT(), using a GeoTools soft cache instead (which can be cleared when CRS.reset is called) .Rationale: it would benefit every usage of WKT, rather than just the GeoServer XStream usage, and fits with the rest of the GeoTools CRS handling (the buffered authority factory is caching CRS lookups too, but only by code)

aaime:
There isn't, but you got official PSC feedback above. 
If you want to merge this PR as is nonetheless, I suggest you escalate discussion on the geoserver-devel list.

groldan:
I see. I didn't take your initial comment "we were wondering about pushing caching down in CRS.parseWKT()" as a recommendation for me to do it. I'll see, this is not something I really need myself, I can apply the patch to geoserver-cloud, and it was just a nice contribution given I'm on vacations.
If I have time and interest I will make a GT pr, otherwise it's up to you, the PSC.
Cheers,

jodygarnett:
It was me with the idea Gabe, indeed we are shocked at how well your loader is doing to place such stress on the system :)

aaime:
Quick PR with GeoTools based replacement, following the existing patterns in CRS class:
[Jotted ](https://github.com/geotools/geotools/pull/4423)

@groldan would that achieve the same performance as this one?

groldan:
I'll give it a shot. It won't have the same concurrency handling as the guava cache, but at the same time it might be a bit of an edge case (concurrent requests hitting a miss will all parse the wkt).
Besides that, performance wise it should be the same.

groldan:
Closing as superseeded by https://github.com/geotools/geotools/pull/4423

PR #7051, opened 2023-07-31
[GEOS-11089] Performance penalty adding namespaces while loading catalog

groldan:
[GEOS-11089](https://osgeo-org.atlassian.net/browse/GEOS-11089)

Use a specialization of `CatalogInfoLookup` for `NamespaceInfo` that speeds up lookup of namespaces by URI.

With this patch and the one for [GEOS-11088](https://github.com/geoserver/geoserver/pull/7050), starting up from a catalog with 120k workspaces and 600k layers, goes from 18 to 4 minutes with the default geoserver loader, and down to less than a minute with the [datadir-catalog-loader](https://github.com/geoserver/geoserver/tree/580221af6da9aedcd4e0333ea205e92277ca4c44/src/community/datadir-catalog-loader) community module.

```
                                     before   after
default loader                        1113s    294s
datadir-catalog-loader (4 trheads)     587s     98s
datadir-catalog-loader (8 trheads)     334s     69s
datadir-catalog-loader (16 trheads)    245s     57s
```

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/impl/CatalogInfoLookup.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/CatalogInfoLookup.java
@@ -138,15 +138,17 @@ class CatalogInfoLookup<T extends CatalogInfo> {
      * things going on)
      */
     <U extends CatalogInfo> List<U> list(Class<U> clazz, Predicate<U> predicate) {
-        ArrayList<U> result = new ArrayList<>();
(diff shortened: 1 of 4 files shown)


aaime:
Looking good. So this helps in case there is a ton of workspaces, which may be the case for a SAAS setup (someone selling thin  slices of a GeoServer installation, each to a customer with a small amount of layers).

PR #7053, opened 2023-08-01
[GEOS-11090] Use Catalog streaming API in WorkspacePage

groldan:
[GEOS-11090](https://osgeo-org.atlassian.net/browse/GEOS-11090)

Make webui's `WorkspaceProvider` use streaming catalog access

  

# Checklist

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/data/workspace/WorkspaceProvider.java
+++ b/src/web/core/src/main/java/org/geoserver/web/data/workspace/WorkspaceProvider.java
@@ -5,24 +5,60 @@
  */
 package org.geoserver.web.data.workspace;
 
-import java.util.Arrays;
+import static org.geoserver.catalog.Predicates.sortBy;
+
+import com.google.common.collect.Streams;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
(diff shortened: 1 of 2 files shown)


src/web/core/src/main/java/org/geoserver/web/data/workspace/WorkspaceProvider.java
```diff
@@ -53,29 +86,121 @@ public WorkspaceProvider() {
...
         return new WorkspaceDetachableModel(object);
     }
+
+    private void validate(Long first, Long count) {
+        if (first > Integer.MAX_VALUE
+                || first < Integer.MIN_VALUE
+                || count > Integer.MAX_VALUE
+                || count < Integer.MIN_VALUE) {
+            throw new IllegalArgumentException(); // TODO Possibly change catalog API to use long
+        }
```
aaime, line 184: Saw the same on the LayerProvider. Well, we'll consider it when GeoServer can easily load 2 billion layers or workspaces. I think we'll need some hardware advancements, and possibly more code improvements :rofl:

aaime (approved):
Looks good. I've also tested it with a large data dir I have locally (50K layers) and gave the parallel data dir loader a spin while I was at it (34 seconds normal loader, 19 seconds parallel one).

src/web/core/src/main/java/org/geoserver/web/data/workspace/WorkspaceProvider.java
groldan: lol yeah this one started as a c&p of that one

PR #7076, opened 2023-08-14
[GEOS-11100] Add opacity parameter to the layer definitions in WPS-Download download maps

turingtestfail:
[GEOS-11100](https://osgeo-org.atlassian.net/browse/GEOS-11100)

…wnload download maps

opacity value check

first unit test and apply alpha and first layer blank

cleanup

more cleanup

doc

  

# Checklist


doc/en/user/source/extensions/wps-download/mapAnimationDownload.rst
```diff
@@ -26,6 +26,7 @@ A layer specification is a XML structure made of three parts:
  * Name: a comma separated list of layer names (eventually just one)
  * Capabilities: link to a capabilities document (optional, used when targetting remote WMS layers)
  * Parameter (key, value): an extra parameter to be added in the WMS request represented by this layer (e.g., ``elevation``, ``CQL_FILTER``, ``env``)
+ * Opacity: an optional parameter, ranging from 0 to 100, that controls the layer image level of transulucency during the merge process.  When not set the layer image is fully opaque.  Note that this is seperate from the overall map ``transparent`` boolean parameter.
```
mprins, line 29:
```suggestion
 * Opacity: an optional parameter, ranging from 0 to 100, that controls the layer image level of translucency during the merge process.  When not set the layer image is fully opaque.  Note that this is separate from the overall map ``transparent`` boolean parameter.
```

mprins (changes requested):

src/extension/wps-download/src/test/resources/org/geoserver/wps/gs/download/mapRemoteSimple11OpacityOutOfRange.xml
```diff
@@ -0,0 +1,50 @@
...
+      </wps:Data>
+    </wps:Input>
+    <wps:Input>
+      <ows:Identifier>layer</ows:Identifier>
+      <wps:Data>
+        <wps:ComplexData>
+          <dwn:Layer xmlns:dwn="http://geoserver.org/wps/download">
+            <dwn:Capabilities>http://geoserver.org/geoserver/wms?request=GetCapabilities&amp;version=1.1.0</dwn:Capabilities>
+            <dwn:Name>cite:BasicPolygons</dwn:Name>
+            <dwn:Opacity>101</dwn:Opacity>
```
aaime, line 39: If anything this should throw some error, opacity should be between 0 and 100

src/extension/wps-download/src/test/resources/org/geoserver/wps/gs/download/mapRemoteSimple11Opacity.xml
```diff
@@ -0,0 +1,50 @@
...
+      </wps:Data>
+    </wps:Input>
+    <wps:Input>
+      <ows:Identifier>layer</ows:Identifier>
+      <wps:Data>
+        <wps:ComplexData>
+          <dwn:Layer xmlns:dwn="http://geoserver.org/wps/download">
+            <dwn:Capabilities>http://geoserver.org/geoserver/wms?request=GetCapabilities&amp;version=1.1.0</dwn:Capabilities>
+            <dwn:Name>cite:BasicPolygons</dwn:Name>
+            <dwn:Opacity>50</dwn:Opacity>
```
aaime, line 39: I was expecting to see a partially transparent output, but it's not happening, both mapSimple.png and mapSimpleWithAlpha.png lack transparency, in that the alpha channel values are all 255.

In other words, picking a file as an example

```
aaime@colossus ~/devel/git-gs/src/extension/wps-download/src/test/resources/org/geoserver/wps/gs/download ((7e1cfd9515...)) $ gdalinfo -stats mapSimpleWithAlpha.png
Driver: PNG/Portable Network Graphics
Files: mapSimpleWithAlpha.png
Size is 256, 256
Image Structure Metadata:
  INTERLEAVE=PIXEL
Corner Coordinates:
Upper Left  (    0.0,    0.0)
Lower Left  (    0.0,  256.0)
Upper Right (  256.0,    0.0)
Lower Right (  256.0,  256.0)
Center      (  128.0,  128.0)
Band 1 Block=256x1 Type=Byte, ColorInterp=Red
  Minimum=0.000, Maximum=255.000, Mean=113.648, StdDev=126.471
  Mask Flags: PER_DATASET ALPHA 
  Metadata:
    STATISTICS_MAXIMUM=255
    STATISTICS_MEAN=113.64826965332
    STATISTICS_MINIMUM=0
    STATISTICS_STDDEV=126.47050543937
    STATISTICS_VALID_PERCENT=100
Band 2 Block=256x1 Type=Byte, ColorInterp=Green
  Minimum=0.000, Maximum=255.000, Mean=113.648, StdDev=126.471
  Mask Flags: PER_DATASET ALPHA 
  Metadata:
    STATISTICS_MAXIMUM=255
    STATISTICS_MEAN=113.64826965332
    STATISTICS_MINIMUM=0
    STATISTICS_STDDEV=126.47050543937
    STATISTICS_VALID_PERCENT=100
Band 3 Block=256x1 Type=Byte, ColorInterp=Blue
  Minimum=0.000, Maximum=255.000, Mean=216.206, StdDev=42.429
  Mask Flags: PER_DATASET ALPHA 
  Metadata:
    STATISTICS_MAXIMUM=255
    STATISTICS_MEAN=216.20637512207
    STATISTICS_MINIMUM=0
    STATISTICS_STDDEV=42.429166719178
    STATISTICS_VALID_PERCENT=100
Band 4 Block=256x1 Type=Byte, ColorInterp=Alpha
  Minimum=255.000, Maximum=255.000, Mean=255.000, StdDev=0.000
  Metadata:
    STATISTICS_MAXIMUM=255
    STATISTICS_MEAN=255
    STATISTICS_MINIMUM=255
    STATISTICS_STDDEV=0
    STATISTICS_VALID_PERCENT=100
```

See the stats for the alpha band (nr 4), min and max are both 255. With a transparency of 50%, I was expecting to see the alpha value at 122 instead (or 123...).

aaime (commented):
The outputs are strange, see below, they should not be fully opaque (even if with alpha band, they don't have actual translucent pixels).

src/extension/wps-download/src/test/resources/org/geoserver/wps/gs/download/mapRemoteSimple11OpacityOutOfRange.xml
turingtestfail: It throws a WPSException, which gets returned as part of a text/xml response.  I can also have it throw something outside of what WPS handles, if needed?

src/extension/wps-download/src/test/resources/org/geoserver/wps/gs/download/mapRemoteSimple11Opacity.xml
turingtestfail: Incorrect test file provided and the test was not properly detecting the mismatch (same file other than the opacity).  Correct file now provided and the test changed to test a pixel Alpha.

turingtestfail:
> The outputs are strange, see below, they should not be fully opaque (even if with alpha band, they don't have actual translucent pixels).

The problem was with an incorrect test resource - correct one added along with a check to confirm the alpha for a pixel in the middle.  Test resource (and test output) validated with GDALInfo.

doc/en/user/source/extensions/wps-download/mapAnimationDownload.rst
aaime: transulucency is still there

src/extension/wps-download/src/test/java/org/geoserver/wps/gs/download/DownloadMapProcessTest.java
```diff
@@ -248,9 +249,49 @@ public void downloadMapMars() throws Exception {
...
+    @Test
+    public void downloadRemoteSimple11Opacity() throws Exception {
+        byte[] bytes =
+                testRemoteSimple(
+                        "mapSimpleWithAlpha.png",
+                        "mapRemoteSimple11Opacity.xml",
+                        "caps111.xml",
+                        "image/png");
+        BufferedImage image = ImageIO.read(new ByteArrayInputStream(bytes));
+        ImageAssert.assertEquals(new File(SAMPLES + "mapSimpleWithAlpha.png"), image, 100);
```
aaime, line 268: This bit works, and I've verified that the mapSimpleWithAlpha has 128 in the alpha channel, so the output should have 50% of it. What's confusing is the ImageAssert passing anyways, even if it's comparing against the original image. I guess it happens because the ImageAssert was built to be tolerant of antialising differences between images, which does use various degrees of alpha channel value to achieve its goal (and thus the image assert tolerates differences in alpha values, I presume).

That said... I had to try it out by hand to figure out what was going on, some explanation here in the test would be good.

aaime (changes requested):

doc/en/user/source/extensions/wps-download/mapAnimationDownload.rst
turingtestfail: resolved

src/extension/wps-download/src/test/java/org/geoserver/wps/gs/download/DownloadMapProcessTest.java
turingtestfail: comments added

mprins (approved):

aaime (approved):

PR #7079, opened 2023-08-18
[GEOS-11102]:Allow configuration of the CSV date format

pratikbandalg:
[GEOS-11102](https://osgeo-org.atlassian.net/browse/GEOS-11102)

  

# Checklist


doc/en/user/source/services/wfs/webadmin.rst
```diff
@@ -66,6 +66,65 @@ A list of values to be declared in all feature types can be provided in the WFS
...
+  h- Hour in 12-hour format (1 to 12) 
+  hh- Hour in 12-hour format (2 digits, with leading zero) 
+  m- Minute (1 or 2 digits) 
+  mm- Minute (2 digits, with leading zero) 
+  s- Second (1 or 2 digits) 
+  ss- Second (2 digits, with leading zero) 
+  a- AM/PM marker 
+  n- Nanosecond 
+  Z- Time zone offset (e.g., “+0800”) 
+  zzzz- Time zone full name (e.g., “Pacific Standard Time”)
```
bradh, line 123: Weird looking quote marks.

doc/en/user/source/services/wfs/webadmin.rst
```diff
@@ -66,6 +66,65 @@ A list of values to be declared in all feature types can be provided in the WFS
...
+     - Wed, Jul 4, '01
+   * - ``yyyy-MM-dd'T'HH:mm:ss.SSSZ``
+     - 2001-07-04T12:08:56.235-0700 
+   * - ``yyyy-MM-dd'T'HH:mm:ss.SSSXXX``
+     - 2001-07-04T12:08:56.235-07:00
+
+
+Here, **yyyy-MM-dd'T'HH:mm:ss.SSS'Z'** pattern represents the year, month, day, hour, minute, second and milliseconds. The components are separated by a hyphen character. A literal 'T' character is used to separate the date and time parts. A literal 'Z' character represents the UTC time zone. For instance, yyyy represents the year with four digits, MM represents the month with leading zeros, dd represents the day with leading zeros, and so on.
+
+Similarly, patterns can be formed by using the characters provided below
```
bradh, line 97: The `SSS` pattern is used above, but not shown below. It might be worth including that, since it seems like a pretty common need.

src/wfs/src/main/java/org/geoserver/wfs/response/CSVOutputFormat.java
```diff
@@ -88,7 +90,6 @@ protected void write(
             FeatureCollectionResponse featureCollection, OutputStream output, Operation getFeature)
             throws IOException, ServiceException {
         // write out content here
-
```
bradh, line 91: The whitespace changes probably are not needed and could be worth cleaning up if you are making other changes. Otherwise noise.

pratikbandalg:
@bradh Did the changes

src/wfs/src/main/java/org/geoserver/wfs/response/CSVOutputFormat.java
```diff
@@ -269,7 +271,14 @@ private AttrFormatter[] getFormatters(FeatureType schema) {
...
                 } else if (java.util.Date.class.isAssignableFrom(binding)) {
-                    formatters[i] = juDateFormatter;
+                    if (gs.getService(WFSInfo.class) != null
+                            && gs.getService(WFSInfo.class).getCsvDateFormat() != null) {
+                        formatters[i] =
+                                new CustomDateFormatter(
+                                        gs.getService(WFSInfo.class).getCsvDateFormat());
+                    } else {
+                        formatters[i] = juDateFormatter;
+                    }
```
aaime, line 281: Better, but still unnecessarily complicated, you're extracting wfsInto three times and the csv format two times...
Suggestion (not sure it compiles, but should give you an idea):

```
formatters[i] = Optional.ofNullable(gs.geService(WFSInfo.class))
                                        .map(wfs -> wfs.getCsvDateFormat())
                                        .map(format -> new CustomDateFormatter(format))
                                        .orElse(juDateFormatter)
```

pratikbandalg: Done

PR #7093, opened 2023-08-25
GeoServer code should not use "Impl" classes from the style package

aaime:
Internal change, not user-visible (hopefully!) that makes GeoServer code stop using Impl classes from the GeoTools style package. Propedeutic to the opengis removal work. 
Also fixed a couple of tests that could not work offline, as it was developed in a time when we had no working wifi.

# Checklist


src/wms/src/test/java/org/geoserver/wms/map/RenderedImageMapOutputFormatTest.java
```diff
@@ -1391,20 +1393,14 @@ public void testStyleUsingChannelsFromCoverageView() throws Exception {
...
-        green.setChannelName("2");
-        blue.setChannelName("1");
-
-        cs.setRGBChannels(red, green, blue);
+        RasterSymbolizer symbolizer = styleBuilder.createRasterSymbolizer();
+        SelectedChannelType red = SF.createSelectedChannelType("3", null);
+        SelectedChannelType green = SF.createSelectedChannelType("2", null);
+        SelectedChannelType blue = SF.createSelectedChannelType("1", null);
+        ChannelSelection cs = SF.createChannelSelection(red, green, blue);
         symbolizer.setChannelSelection(cs);
```
jodygarnett, line 1404: Expect RasterSymbolizer.setChannelSelector(cs) is a RasterSybolizerImpl method.

jodygarnett (changes requested):
Accidental use of org.geotools.styling.RasterSymbolizer

jodygarnett (approved):
Thanks!

PR #7106, opened 2023-09-08
GEOS-11118: Added new way to read wkt

921000:
[GEOS-11118](https://osgeo-org.atlassian.net/browse/GEOS-11118) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

From GEOS-11118: 
      Geoserver provides an interface for clipping WMS services, which can be achieved by adding the parameter 'clip' to the WMS request. We need to tailor the tif service based on the boundaries of the administrative region. In this usage scenario, the length of the passed parameter will exceed the limit of the get request, resulting in the inability to realize the clipping function.
     In order to solve this problem, a processing method is added by reading the wkt data in the file. When requesting, the wkt file path is added to the parameter 'clipPath'. The server reads the wkt data in the file and implements the clipping function.

# Checklist

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/map/GetMapKvpRequestReader.java
+++ b/src/wms/src/main/java/org/geoserver/wms/map/GetMapKvpRequestReader.java
@@ -15,6 +15,12 @@ import java.io.InputStreamReader;
 import java.io.Reader;
 import java.io.StringReader;
 import java.net.URL;
+import java.nio.ByteBuffer;
+import java.nio.channels.FileChannel;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.FileSystems;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


mprins:
You appear to have referenced the wrong jire ticket 11118 states "Add Temporal Extents domain and datetime parameter to Maps API"
Also; what is wrong with using POST fot your request instead?

mprins:
You appear to have referenced the wrong jire ticket 11118 states "Add Temporal Extents domain and datetime parameter to Maps API"
Also; what is wrong with using POST for your request instead?

921000:
>


> You appear to have referenced the wrong jire ticket 11118 states "Add Temporal Extents domain and datetime parameter to Maps API" Also; what is wrong with using POST for your request instead?

Sorry, I just realized that the ticket I quoted was wrong.
We refer to the examples in the official documentation. In the documentation, wms clipping is done based on get requests. Can post requests be supported? If it is supported, how should I use it?
In addition, using some wkt data similar to area boundaries is much larger than ordinary http requests. I feel that using file path transfer can bring less performance and bandwidth loss.

Special thanks for your reply.

mprins (changes requested):
Also this approach is prone to open up Geoserver to arbitrary file reading vulnerabilities

mprins:
this blogpost should get you started on using POST: https://www.how2map.com/2013/10/geoserver-getmap-with-http-post.html as well as the official docs: https://docs.geoserver.org/latest/en/user/services/wms/reference.html#getmap 
Any other question should be directed at the respective channels (geoserver user list or [stackoverflow](https://gis.stackexchange.com/))

921000:
This is not a good solution, may introduce security holes, and geoserver already has other solutions, so close

921000:
This is not a good solution, may introduce security holes, and geoserver already has other solutions, so close

PR #7155, opened 2023-10-04
[GEOS-11142] Add mime type mapping for yaml files

sikeoka:
[GEOS-11142](https://osgeo-org.atlassian.net/browse/GEOS-11142) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This PR adds a mime type mapping for yaml files so that the Style Publisher will return a "Content-Type: text/plain" response header for YSLD styles instead of no Content-Type header.  There is no unit test since this is a web.xml change.

  

# Checklist

### DIFF
--- a/src/web/app/src/main/webapp/WEB-INF/web.xml
+++ b/src/web/app/src/main/webapp/WEB-INF/web.xml
@@ -290,6 +290,10 @@
       <extension>json</extension>
       <mime-type>application/json</mime-type>
     </mime-mapping>
+    <mime-mapping>
+      <extension>yaml</extension>
+      <mime-type>text/plain</mime-type>
+    </mime-mapping>
   
     <welcome-file-list>
         <welcome-file>index.html</welcome-file>


sikeoka:
That mime type is relatively new and I didn't know it existed since I had set this on my end a long time ago but only recently saw a mention of this.  The point of this PR is to register ANY mime type to resolve a potential vulnerability with YSLD styles.  Interestingly the OGC API modules use application/x-yaml which may be better to be consistent throughout the GeoServer codebase.  I will test out all three formats and update this PR tomorrow.

sikeoka:
I tested out the other mime types and the one advantage that text/plain has is that it still allows YSLD (and other .yaml) files to be rendered in a browser instead of forcing a download.  There are only a few cases where GeoServer uses the servlet mime type mappings so I don't see this as a big deal.  I'll let @jodygarnett decide on the mime type since he is also using the YSLD extension.

aaime:
We already have cases where the respose is not readable by browsers, because OGC decided on odd mime types, e.g., WMS 1.1 Capabilities response, and GML 3.2 in any WFS version (which also means WFS 2.0 responses by default are downloaded rather than shown).

mprins:
I'd prefer using the IANA registered string for compliance reasons, for the OGC API as well (`application/x-yaml` is listed as deprecated in the document I linked to)

Aside from that I don't see much value in having a browser render any yaml file as text, but that is just my 2&cent;

sikeoka:
The application/yaml media type is actually still in the Draft stage so it probably shouldn't be used yet:
https://datatracker.ietf.org/doc/draft-ietf-httpapi-yaml-mediatypes/

I will change this to application/x-yaml.

sikeoka:
I changed this back to text/plain and my opinion is to just merge this.  It can be changed later after an official YAML media type is added to Spring if someone complains about it.

jodygarnett (approved):
Excellent, is a backport desired for this change?

PR #7156, opened 2023-10-05
[GEOS-11050] Refactor Resources and Paths API to support alternative …

NielsCharlier:
Following the rather difficult discussion between Jody and I on the mailing list, I have attempted to find a solution for this issue that addresses each of our concerns. As a guideline I have implemented the points we ended up agreeing on, with the exception of one unresolved matter (the API distinction between two kinds of paths), that can still be modified after discussing an alternative solution. The main ideas behind the suggested changes can be read in the documentation changes/additions inside this PR, but I will add some more notes here regarding my motivation. The two main concerns that are somewhat conflicting and needed to be reconciled are:

1) (Niels' main concern, the matter of proper generic development) The resource API should be entirely independent of matters related to Operating Systems (such as the distinction absolute/relative paths), and make no assumption about being part of a File System. The resource store is an entity of its own and has its own 'root'. Developers should be informed not to use the file system but rather consciously use the resource API in the intended, completely generic way. 

Note that the usage of the resource store does not limit itself to internal matters. For instance, style sheets that refer to images must still work on alternative resource stores (such as JDBC or otherwise), which is why point 2 is relevant.

2) (Jody's main concern, the matter of intuitive and predictable user experience without technical knowledge requirements) What looks like an absolute file system path, should be an absolute file system path. Only pure 'relative' paths (on linux this implies starting with a name, not a slash) should be sent to the resource store; in order to assure predictability for the end user.

My proposed solution is 
1) Formally distinguish 'file paths' from 'resource paths' on the API and implementation level. To avoid any confusion with developers and to encourage proper API usage, they are no longer lumped together as 'paths'. The resource store + paths API is entirely separate from OS-dependent matters such as the relative/absolute distinction. A separate 'FilePaths' utility class was created to address the distinction between absolute/relative file paths.  The ResourceStore takes only resource paths, is OS independent, and always starts from its own root.
Note that Files.asResource does not return a 'true' resource (from the resource store) but a wrapper that allows files to be used in resource aware API. Its only proper usage is for user provided paths/urls (see point 2), and really should never be called by developers directly.
In order to distinguish between true resources and wrappers, without having to rely on the OS-dependent relative/absolute distinction, a method 'isInternal' was added to the Resource class to verify a resource is indeed, part of the resource store and its path can be interpreted as such.

2) For any URL or path provided by the end user (in configuration, templates etc) the method URLs.asResource is used to translate to a 'resource' (either a true resource from the resource store, or some kind of wrapper).
This method follows for file: URLs (or simple paths) the principle of Jody: only paths that are considered relative in the Operating System are interpreted as resources, anything else are files (such as a path starting with leading slash on linux).
Note: since file paths without leading slash violate the URL standard, the original idea was to replace them with "resource:" URL's. This is an option that may be less confusing for some people, but is entirely optional and backwards compatibility of relative file: URLs will continue to be respected. (This is not new)

[GEOS-11050](https://osgeo-org.atlassian.net/browse/GEOS-11050) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

…ResourceStore implementations

  

# Checklist

### DIFF
--- a/doc/en/developer/source/programming-guide/config/resource.rst
+++ b/doc/en/developer/source/programming-guide/config/resource.rst
@@ -21,8 +21,8 @@ Reference:
 
    The methods in the Resource API use String parameter names consistently:
 
- * ``resource path`` path to a resource in the resource store. (For instance in the case of the default FileSystemResourceStore, this is file path that is relative with respect to the data directory, but for perserving generic behaviour compatible with any resource store, developers should not assume this to be the case). Resource paths do not support the `.` and `..` relative directory names. Resource paths use forward slashes, similar to URL's and unix style file paths, and are OS-independent.
- * ``file path`` absolute path to a file in the file system. While these are OS dependent (with regard to the root of the absolute path) but they must always forward slashes, as supported by all operating systems and compatible with resource paths as well as file URL's. Note that ``Resource.path()`` for resources obtained by ``Files.asResource(file)`` will return a file path rather than a resource path.  
+ * ``resource path`` path to a resource in the resource store. (For instance in the case of the default FileSystemResourceStore, this is file path that is relative with respect to the data directory, but to preserve generic behaviour compatible with any resource store, developers should not assume this to be the case). Resource paths do not support the `.` and `..` relative directory names. Resource paths use forward slashes, similar to URL's and unix style file paths, and are OS-independent.
+ * ``file path`` absolute path to a file in the file system. While these are OS dependent (with regard to the root of the absolute path) but they must always use forward slashes, as supported by all operating systems and compatible with resource paths as well as file URL's. Note that ``Resource.path()`` for resources obtained by ``Files.asResource(file)`` will return a file path rather than a resource path.  
  * ``file`` a java File reference.  
  * ``url`` a location resolved with respect to the resource store. A number of special cases developed over time distilled into ``Resources.fromUrl(base,url)`` and ``Files.url(base,url)`` methods.  
 
--- a/doc/en/developer/source/programming-guide/config/resource.rst
+++ b/doc/en/developer/source/programming-guide/config/resource.rst
@@ -21,17 +21,21 @@ Reference:
 
    The methods in the Resource API use String parameter names consistently:
 
-   * ``path`` relative location of resource within the data directory.
-  
-     Paths are represented similar to a File URL, ``file://`` prefix used internally and
-     by the resource REST API.
-   
-   * ``file`` a java File reference.
-  
-   * ``url`` a location resolved with respect to the data directory.
-  
-     A large number of special cases developed over time distilled into  ``Resources.fromUrl(base,url)`` and ``Files.url(base,url)`` methods.
-  
+ * ``resource path`` path to a resource in the resource store. (For instance in the case of the default FileSystemResourceStore, this is file path that is relative with respect to the data directory, but for perserving generic behaviour compatible with any resource store, developers should not assume this to be the case). Resource paths do not support the `.` and `..` relative directory names. Resource paths use forward slashes, similar to URL's and unix style file paths, and are OS-independent.
+ * ``file path`` absolute path to a file in the file system. While these are OS dependent (with regard to the root of the absolute path) but they must always forward slashes, as supported by all operating systems and compatible with resource paths as well as file URL's. Note that ``Resource.path()`` for resources obtained by ``Files.asResource(file)`` will return a file path rather than a resource path.  
+ * ``file`` a java File reference.  
+ * ``url`` a location resolved with respect to the resource store. A number of special cases developed over time distilled into ``Resources.fromUrl(base,url)`` and ``Files.url(base,url)`` methods.  
+
+General Guidelines
+------------------
+
+All geoserver developers should be wary of the following general principles when contributing or reviewing:
+
+ * Avoid as much as possible using the file system directly. The only acceptable exception is when third party libraries require this, and even then the Resources API should be used maximally.
+ * For custom configuration files with a fixed location, always use ``ResourceStore``. ``GeoServerResourceLoader`` and ``GeoServerDataDirectory`` are legacy and should not be used in new code.
+ * For URL's provided by user configuration (such as templates, style sheets, etc), use ``Resources.fromURL``.
+ * For input/output, always use ``Resource.in()`` and ``Resource.out()``.
+ * Avoid the usage of ``Resource.file()`` and ``Resource.directory()``. These methods are only necessary for third party libraries that require usage of the file system, and only for input.  They should never be used for permanent output: Since there are alternative implementations of the ResourceStore that do not use the file system as underlying storage device, modifying them does not necessarily have a lasting effect
 
 ResourceStore
 -------------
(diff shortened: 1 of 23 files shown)


src/platform/src/main/java/org/geoserver/platform/resource/Resources.java
```diff
@@ -685,12 +685,30 @@ public static Resource fromURL(Resource baseDirectory, URL url) {
...
         } else if (url.getProtocol().equalsIgnoreCase("file")) {
-            return Files.asResource(URLs.urlToFile(url));
+            return fromFile(baseDirectory, URLs.urlToFile(url));
         } else {
             return null;
         }
     }
 
+    /**
+     * @param filePath
```
bradh, line 695: This is causing a QA failure. Needs to be removed or (better) made relevant.

doc/en/developer/source/programming-guide/config/resource.rst
```diff
@@ -21,17 +21,21 @@ Reference:
...
-     Paths are represented similar to a File URL, ``file://`` prefix used internally and
-     by the resource REST API.
-   
-   * ``file`` a java File reference.
-  
-   * ``url`` a location resolved with respect to the data directory.
-  
-     A large number of special cases developed over time distilled into  ``Resources.fromUrl(base,url)`` and ``Files.url(base,url)`` methods.
-  
+ * ``resource path`` path to a resource in the resource store. (For instance in the case of the default FileSystemResourceStore, this is file path that is relative with respect to the data directory, but for perserving generic behaviour compatible with any resource store, developers should not assume this to be the case). Resource paths do not support the `.` and `..` relative directory names. Resource paths use forward slashes, similar to URL's and unix style file paths, and are OS-independent.
```
bradh, line 24:
```suggestion
 * ``resource path`` path to a resource in the resource store. (For instance in the case of the default FileSystemResourceStore, this is file path that is relative with respect to the data directory, but to preserve generic behaviour compatible with any resource store, developers should not assume this to be the case). Resource paths do not support the `.` and `..` relative directory names. Resource paths use forward slashes, similar to URL's and unix style file paths, and are OS-independent.
```

doc/en/developer/source/programming-guide/config/resource.rst
```diff
@@ -21,17 +21,21 @@ Reference:
...
+ * ``file path`` absolute path to a file in the file system. While these are OS dependent (with regard to the root of the absolute path) but they must always forward slashes, as supported by all operating systems and compatible with resource paths as well as file URL's. Note that ``Resource.path()`` for resources obtained by ``Files.asResource(file)`` will return a file path rather than a resource path.
```
bradh, line 25:
```suggestion
 * ``file path`` absolute path to a file in the file system. While these are OS dependent (with regard to the root of the absolute path) but they must always use forward slashes, as supported by all operating systems and compatible with resource paths as well as file URL's. Note that ``Resource.path()`` for resources obtained by ``Files.asResource(file)`` will return a file path rather than a resource path.  
```

src/platform/src/main/java/org/geoserver/platform/resource/FilePaths.java
```diff
@@ -0,0 +1,178 @@
...
+     * directory name.
+     *
+     * <p><b>Relative</b>: Relative paths are represented in a straight forward fashion with: {@code
+     * Paths.names("data/tasmania/roads.shp"} --> {"data","tasmania","roads.shp"}}.
+     *
+     * <p><b>Absolute path</b>: When working with an absolute path the list starts with a special
+     * marker.
+     *
+     * <p>Linux absolute paths are start with leading slash character ({@code / } ). <br>
+     * {@code convert("/srv/gis/cadaster/district.geopg") --> "/srv/gis/cadaster/district.geopg" <br>
```
bradh, line 45:
```suggestion
     * {@code convert("/srv/gis/cadaster/district.geopkg") --> "/srv/gis/cadaster/district.geopkg" <br>
```

src/platform/src/main/java/org/geoserver/platform/resource/Files.java
```diff
@@ -300,7 +305,9 @@ private Files() {
      * @param baseDirectory Optional base directory used to resolve relative file URLs
      * @param url File URL or path relative to data directory
      * @return File indicated by provided URL location
+     * @deprecated use resources.
      */
+    @Deprecated
     public static File url(File baseDirectory, String url) {
```
bradh, line 311: Looks like there are a couple of users of this, and those are failing the QA checks.

bradh (changes requested):

src/platform/src/main/java/org/geoserver/platform/resource/Files.java
NielsCharlier: Hmmm, isn't the whole point of deprecation that you allow transition? Does the QA rule not effectively make it useless?

aaime: No, it just forces the developers deprecating a method to remove its usage everywhere in the code base. It's the usage that triggers QA, not the annotation on the method. 
This also helps to validate that there is indeed a good migration path, and eventually helps improving the javadoc/instructions about what to do to move away from the deprecated method.

jodygarnett:
Yes I will look at this in the week ahead; sorry I have had my hands full. This architecture api topics always seem to be important but not urgent.

jodygarnett:
This has been squashed and merged, with minor clarifications to documentation and javadocs. 

Thanks for your patience Niels, the Files.asResource(path) wrapper was useful during migration, but caused too much grief over time.

aaime:
I believe merging this PR broke the developer guide build:

```
aaime@colossus ~/devel/git-gs/doc/en (main) $ ant developer
Buildfile: /home/aaime/devel/git-gs/doc/en/build.xml

init:
     [echo] Base directory is /home/aaime/devel/git-gs/doc/en
     [echo] Build directory is /home/aaime/devel/git-gs/doc/en/target
     [echo] project version is 2.25-SNAPSHOT

developer:

sphinx:
     [echo] Running sphinx-build -D release=2.25-SNAPSHOT -q -W --keep-going -b html -d "/home/aaime/devel/git-gs/doc/en/target/developer/doctrees" . "/home/aaime/devel/git-gs/doc/en/target/developer/html"
     [exec] /home/aaime/devel/git-gs/doc/en/developer/source/programming-guide/config/resource.rst:63: WARNING: Bullet list ends without a blank line; unexpected unindent.

BUILD FAILED
/home/aaime/devel/git-gs/doc/en/build.xml:121: The following error occurred while executing this line:
/home/aaime/devel/git-gs/doc/en/build.xml:143: exec returned: 1

Total time: 1 second

```

According to git , @jodygarnett is to "blame" (LOL):

```
5cef7397742 (Jody Garnett  2024-02-28 22:36:37 -0800  60) * For URL's provided by user configuration (such as templates, style sheets, etc), use ``Resources.fromURL``.
5cef7397742 (Jody Garnett  2024-02-28 22:36:37 -0800  61) 
5cef7397742 (Jody Garnett  2024-02-28 22:36:37 -0800  62) * For input/output, always use ``Resource.in()`` and ``Resource.out()``.
5cef7397742 (Jody Garnett  2024-02-28 22:36:37 -0800  63) x  
5cef7397742 (Jody Garnett  2024-02-28 22:36:37 -0800  64)   This approach is compatible with java try-with-resource making for easy to follow code.
18281f98fcd (Jody Garnett  2022-12-19 23:58:09 -0800  65) 
18281f98fcd (Jody Garnett  2022-12-19 23:58:09 -0800  66) ResourceStore
```

See that "x" on line 63?

jodygarnett:
Yeah that build was not covered by doc workflow check. 

I fixed it.

NielsCharlier:
Fantastic! Thanks @jodygarnett

PR #7157, opened 2023-10-05
[GEOS-11145] The GUI "wait spinner" is not visible any longer

aaime:
Regression in 2.23.0 due to [GSIP-213](https://github.com/geoserver/geoserver/wiki/GSIP-213), simple fix with explanation. Can't think of a way to test this, as all the action is happening client side, in javascript.

# Checklist

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/GeoServerBasePage.html
+++ b/src/web/core/src/main/java/org/geoserver/web/GeoServerBasePage.html
@@ -97,7 +97,9 @@
     </div><!-- /#page -->
     </div><!-- /.wrap> -->
   </div><!-- /#main -->
-  <div id="ajaxFeedback" class="d-none">
+  <!-- The feebdback component needs a local style so that the Wicket Ajax support can flip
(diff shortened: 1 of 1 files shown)


src/web/core/src/main/java/org/geoserver/web/GeoServerBasePage.html
```diff
@@ -97,7 +97,9 @@ <h1 wicket:id="gbpTitle" id="header-title"></h1>
     </div><!-- /#page -->
     </div><!-- /.wrap> -->
   </div><!-- /#main -->
-  <div id="ajaxFeedback" class="d-none">
+  <!-- The feebdback component needs a local style so that the Wicket Ajax support can flip
+       its value to "display: true". Will not work with a class, do not use one here -->
+  <div id="ajaxFeedback" style="display: none">
```
jodygarnett, line 102: @MichelGabriel ping :)

jodygarnett (approved):
It makes sense to me; thanks for the fix.

PR #7162, opened 2023-10-06
[GEOS-11149] Update response headers for the Style Publisher

sikeoka:
[GEOS-11149](https://osgeo-org.atlassian.net/browse/GEOS-11149) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This PR sets the "Content-Disposition: attachment" response header to style resources with certain content types and also now uses the same default content type as the Resources REST API to mitigate a security issue.  The Style Publisher also no longer allows access to the XML configuration files even though they don't currently contain any sensitive information.

Note that with this PR and https://github.com/geoserver/geoserver/pull/7161, the Classpath/File/Style Publisher and Resources REST API will always return the same Content-Type for a particular resource.

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/ows/StylePublisher.java
+++ b/src/main/src/main/java/org/geoserver/ows/StylePublisher.java
@@ -39,6 +39,16 @@ public class StylePublisher extends AbstractURLPublisher {
     }
 
     @Override
+    protected boolean isAttachment(URL url, String filename, String mime) {
+        // prevent stored XSS using malicious style resources with the
+        // text/html, text/xml, application/xml or image/svg+xml mime types
+        String lowerCaseMime = mime.toLowerCase();
+        return lowerCaseMime.contains("xml")
+                || lowerCaseMime.contains("html")
+                || super.isAttachment(url, filename, mime);
+    }
+
+    @Override
     protected URL getUrl(HttpServletRequest request) throws IOException {
         String ctxPath = request.getContextPath();
         String reqPath = request.getRequestURI();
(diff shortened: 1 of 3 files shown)


src/main/src/main/java/org/geoserver/ows/StylePublisher.java
```diff
@@ -75,7 +85,10 @@ protected URL getUrl(HttpServletRequest request) throws IOException {
 
             switch (resource.getType()) {
                 case RESOURCE:
-                    return URLs.fileToUrl(resource.file());
+                    // don't allow access to the style .xml files
+                    return resource.name().endsWith(".xml")
+                            ? null
+                            : URLs.fileToUrl(resource.file());
```
jodygarnett, line 91: There is no easy way to tell if an xml document in the styles directory is part of the catalogue configuration, or just an xml document. So checking for "xml" extension is a fair assumption.

Technically an SLD file is just an xml document, and it is would be okay to upload such a file as a style for use.

jodygarnett (approved):

jodygarnett:
I am okay with this change, would appreciate a second reviewer had a look also.

src/main/src/main/java/org/geoserver/ows/StylePublisher.java
sikeoka: There really shouldn't be any .xml files in the style directories that isn't a valid StyleInfo file since every one is going to be an XStream stack trace in the logs on every GeoServer restart/reload.

jodygarnett: Wow okay :)

PR #7171, opened 2023-10-09
[GEOS-11150] add geoserver-monitor-kafka extension

spike83:
[GEOS-11150](https://osgeo-org.atlassian.net/browse/GEOS-11150) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

If we want to analyze the data produced by the monitoring extension we had the option to write to a database using hibernate. Going forward I’d like also to be able to stream the monitoring data into a Kafka topic.

With stream processing we can build the aggregates we want e.g. with KsqlDB, Flink or Kafka Streams Application and provide reports in near real time… But foundation is to have the data streamed into a topic…

With this extension we can stream the data into a Kafka topic and configure all the producer options as we want to.
  

# Checklist


aaime:
Whoops, looks like it has accumulated conflicts. Personally I have no experience with Kafka, cannot review the details, but the direction looks like a good one to me. Anyone else that wants to have a look before merge?

src/community/pom.xml
```diff
@@ -568,6 +570,25 @@
         <module>monitor-hibernate</module>
       </modules>
     </profile>
+    <profile>
+      <id>monitor-kafka</id>
+      <build>
+        <plugins>
+          <plugin>
+            <artifactId>maven-assembly-plugin</artifactId>
+            <version>3.2.0</version>
```
mprins, line 579: this version is already provided through pluginManagment of the parent pom:
https://github.com/geoserver/geoserver/blob/f5355bc54f1a09b2c429406002dddd09f2f5fac3/src/pom.xml#L1699-L1702

```suggestion
```

src/web/app/pom.xml
```diff
@@ -629,8 +629,23 @@
       </activation>
       <dependencies>
         <dependency>
-          <groupId>org.geoserver.extension</groupId>
+          <groupId>org.geoserver.community</groupId>
           <artifactId>gs-monitor-hibernate</artifactId>
+          <version>2.25-SNAPSHOT</version>
```
mprins, line 634:
```suggestion
          <version>${project.version}</version>
```

mprins (changes requested):
I've provided some small changes to the pom files.

Aside there seems to be Kafka 3.6.0; but since I haven't used kafka I'm not sure if using that is opportune (compared to 3.5.1)

spike83:
@mprins thanks for your review. I updated the deps. Kafka 3.6.0 was released in the mean time. Not critical but for sure a good idea to update.

mprins (approved):

aaime:
Merged! If you want to backport, that will have to be a manual PR, as you'll need a different version number in the pom.xml file (and potentially, I have not checked, some different API in GeoTools, if you want to go down to 2.23.x).

PR #7174, opened 2023-10-10
[GEOS-11153] Improve handling special characters in the WMS OpenLayers Format

sikeoka:
This PR updates the WMS OpenLayers Format to properly escape special characters from GeoServer catalog information.

  

# Checklist


src/wms/src/main/java/org/geoserver/wms/map/AbstractOpenLayersMapOutputFormat.java
```diff
@@ -292,7 +293,7 @@ private List<String> styleNames(WMSMapContent mapContent) {
             MapLayerInfo info = mapContent.getRequest().getLayers().get(0);
             result = info.getOtherStyleNames();
         }
-        return result;
+        return result.stream().map(StringEscapeUtils::escapeHtml4).collect(Collectors.toList());
```
aaime, line 296: Why are the other parameters escaped in the template, while this one is escaped in Java code?

aaime (commented):
Confused about escaping done in different places, see below.

src/wms/src/main/java/org/geoserver/wms/map/AbstractOpenLayersMapOutputFormat.java
aaime: I see, the explanation makes sense, thanks!

PR #7214, opened 2023-10-24
[GEOS-11170] OGC API Features "/api" link is broken in OpenAPI page

bradh:
[GEOS-11170](https://osgeo-org.atlassian.net/browse/GEOS-11170) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

As identified at tutorial session today. Just a left-over from the change from api -> openapi.

I also found the descriptive text was a link to the Conformance stuff, updated that too. It is a bit weak, but at least its no longer wrong.

# Checklist

### DIFF
--- a/src/community/ogcapi/ogcapi-features/src/main/resources/org/geoserver/ogcapi/v1/features/openapi.yaml
+++ b/src/community/ogcapi/ogcapi-features/src/main/resources/org/geoserver/ogcapi/v1/features/openapi.yaml
@@ -51,7 +51,7 @@ paths:
           $ref: '#/components/responses/LandingPage'
         '500':
           $ref: '#/components/responses/ServerError'
-  '/api':
+  '/openapi':
     get:
       tags:
         - Capabilities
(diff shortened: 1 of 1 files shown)


src/community/ogcapi/ogcapi-features/src/main/resources/org/geoserver/ogcapi/v1/features/openapi.yaml
```diff
@@ -1156,6 +1156,10 @@ components:
         text/html:
           schema:
             type: string
+    OpenAPI:
```
aaime, line 1159: Wondering... is it legit to only have a description without a schema? Maybe it is, dunno, otherwise I guess we can link an external, online schema for OpenAPI itself.

aaime:
Looks good, see my comment about the component declaration. Also thinking out loud, I guess the same issue is likely popping up in other openapi.yml documents in the other services. At the same time, that does not need to be solved in this PR either.

bradh:
> Looks good, see my comment about the component declaration. Also thinking out loud, I guess the same issue is likely popping up in other openapi.yml documents in the other services. At the same time, that does not need to be solved in this PR either.

I have another ticket (GEOS-11172) to check for "/api". At this stage I think there is only one other yaml file that has an "/api" entry and the others don't have "/api" or "/openapi" - another issue, which we'll likely hit for CITE level verification.

PR #7275, opened 2023-11-16
[GEOS-11203] WMS GetFeatureInfo bad WKT exception for label-geometry

petersmythe:
[GEOS-11203](https://osgeo-org.atlassian.net/browse/GEOS-11203) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

https://osgeo-org.atlassian.net/browse/GEOS-11203

# Checklist

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/featureinfo/FeatureInfoStylePreprocessor.java
+++ b/src/wms/src/main/java/org/geoserver/wms/featureinfo/FeatureInfoStylePreprocessor.java
@@ -61,7 +61,7 @@ class FeatureInfoStylePreprocessor extends SymbolizerFilteringVisitor {
 
     private PropertyName defaultGeometryExpression;
 
-    private boolean addSolidLineSymbolier;
+    private boolean addSolidLineSymbolizer;
 
     public FeatureInfoStylePreprocessor(FeatureType schema) {
         this.schema = schema;
@@ -204,10 +204,10 @@ class FeatureInfoStylePreprocessor extends SymbolizerFilteringVisitor {
         geometriesOnPolygonSymbolizer.clear();
         geometriesOnPointSymbolizer.clear();
         geometriesOnTextSymbolizer.clear();
-        addSolidLineSymbolier = false;
+        addSolidLineSymbolizer = false;
         super.visit(rule);
         Rule copy = (Rule) pages.peek();
-        if (addSolidLineSymbolier) {
+        if (addSolidLineSymbolizer) {
             // add also a black line to make sure we get something in output even
             // if the user clicks in between symbols or dashes
             LineSymbolizer ls = sb.createLineSymbolizer(Color.BLACK);
@@ -319,7 +319,7 @@ class FeatureInfoStylePreprocessor extends SymbolizerFilteringVisitor {
             List<Expression> dashArray = stroke.dashArray();
             Graphic graphicStroke = stroke.getGraphicStroke();
             if (graphicStroke != null || dashArray != null && !dashArray.isEmpty()) {
-                addSolidLineSymbolier = true;
+                addSolidLineSymbolizer = true;
             }
         }
     }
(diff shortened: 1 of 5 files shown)


src/wms/src/main/java/org/geoserver/wms/featureinfo/FeatureInfoStylePreprocessor.java
```diff
@@ -300,13 +298,19 @@ private Rule buildDynamicGeometryRule(
...
+        try {
+            Object descriptor = geom.evaluate(schema);
+            if (!(descriptor instanceof GeometryDescriptor)) {
+                // we don't know what this will be, we probably evaluated a filter function
+                return Geometry.class;
+            } else {
+                // see if we are dealing with a polygon
+                return ((GeometryDescriptor) descriptor).getType().getBinding();
+            }
+        } catch (IllegalArgumentException e) {
```
aaime, line 311: I would catch Exception, different functions might fail in different ways. Besides that, we need a test.

petersmythe: Thank you @aaime.  Do you have any guidance as to which existing test I can extend, please?  I find it difficult to understand the overall structure of the unit tests.

aaime: Most of these classes are tested via integration testing. It's however possible to build also a unit test, and for this case, probably appropriate. I've asked Github copilot to come up with a test, what it spit out is not usable as a final version, but you could use it as a starting point (I've modified the output a bit to add my own comments):

```java
// copyright header here please
package org.geoserver.wms.featureinfo;

import static org.junit.Assert.*;

import org.geotools.api.feature.type.FeatureType;
import org.geotools.api.style.Style;
import org.junit.Test;

public class FeatureInfoStylePreprocessorTest {

    @Test
    public void testGeomFromWkt() throws Exception {
        // Fix this to use concatenation of a x and y
        String sld = "<StyledLayerDescriptor xmlns=\"http://www.opengis.net/sld\">" +
            "<NamedLayer><Name>Layer</Name>" +
            "<UserStyle><FeatureTypeStyle>" +
            "<Rule><TextSymbolizer>" +
            "<Geometry><ogc:Function name=\"geomFromWkt\">" +
            "<ogc:Literal>POINT(0 0)</ogc:Literal>" +
            "</ogc:Function></Geometry>" +
            "<Label><ogc:PropertyName>name</ogc:PropertyName></Label>" +
            "</TextSymbolizer></Rule>" +
            "</FeatureTypeStyle></UserStyle>" +
            "</NamedLayer></StyledLayerDescriptor>";

        // Create a mock FeatureType
        FeatureType mockFeatureType = createMockFeatureType();

        // Create an instance of FeatureInfoStylePreprocessor
        FeatureInfoStylePreprocessor preprocessor = new FeatureInfoStylePreprocessor(mockFeatureType);

        // Apply the SLD to the preprocessor
        Style style = readSLD(sld, getClass());
        style.accept(preprocessor);

        // Check the result of preprocessing is the expected one, 
       // in addition to the obvious "do not throw an exception"
    }

    // Helper method to create a mock FeatureType
    private FeatureType createMockFeatureType() {
        // Use DataUtilities to build a FeatureType that has x and y columns
    }
}
```

petersmythe: I think it might belong in org.geoserver.wms.featureinfo.RenderingBasedFeatureInfoTest, if I add the special breaking style as a new SLD in \src\wms\src\test\resources\org\geoserver\wms\featureinfo ?

aaime: Yes, you can also write an integration test there. In addition to the style, you'll have to prepare a sample data set with a x and y column then (as a property file, see what is done with the grid/repeated_lines/giantPolygon" cases then.

petersmythe: Thank you, I had already proceeded with the second option (integration test) before I saw your reply, so I trust this is acceptable.

aaime (approved):
LGTM

petersmythe:
Appreciate your guidance

PR #7283, opened 2023-11-21
[GEOS-11209] Proof Key of Code Exchange (PKCE)

jodygarnett:
[GEOS-11209](https://osgeo-org.atlassian.net/browse/GEOS-11209) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

* Add PKCE configuration option for open id connect module
* Include OIDC_LOGGING profile to aid in troubleshooting
* PKCEAuthenticationEntryPoint to generate and store code_verifier, and then include code_challenge in authorization request
* PKCERequestEnhancer to retrieve code_verifier and include in access token request

[GEOS-11212](https://osgeo-org.atlassian.net/browse/GEOS-11212):

* Use JWK URI to check access token

Some protected methods introduced into oauth2-core classes to allow override for PKCE workflow.
  

# Checklist


doc/en/user/source/community/oauth2/oidc.rst
```diff
@@ -0,0 +1,257 @@
...
+Reference:
+
+* `rfc7636 Proof Key for Code Exchange by OAuth Public Clients <https://datatracker.ietf.org/doc/html/rfc7636>`_
+
+JSON Web Key set URI
+^^^^^^^^^^^^^^^^^^^^
+
+The ``JSON Web Key set URI`` provides the location of a document of public keys that can be used to check the signature of the provided accessToken.
+
+Optional: When using a role service to obtain the roles associated with access token, it is no longer required to use ``Check Token Endpoint URL``. If you leave that field blank you may rely only on the ``JSON Web Key set URI`` signature check.
```
jodygarnett, line 107: These are the new docs for PKCE work and JWKS URI use.

afabiani:
@jodygarnett thanks for this update. Overall it looks very good to me. Just two minor comments:

 - A couple of test cases are missing the headers I guess
 - Did you test this code with a custom KID? From what I can see it should work, just to confirm
 - It would be nice (if you have the chance, otherwise I can eventually contribute) to add some more details on the docs about validating an OpenID Connect ID Token, since most of the modern providers requires it nowdays

jodygarnett:
> * It would be nice (if you have the chance, otherwise I can eventually contribute) to add some more details on the docs about validating an OpenID Connect ID Token, since most of the modern providers requires it nowdays

My experience with security topics is limited so I do not feel comfortable offering a clear explanation. 

This pull request only adds validation in one location - where JWKs is being used to check the token. 

The other place GeoServer uses the JWKS URI has a comment indicated it will be used by JwkTokenStore to "verify" ... but I see no evidence of validation being performed by this JwkTokenStore:
```
    private void validate(OAuth2AccessToken token) {
        Object maybeIdToken = token.getAdditionalInformation().get("id_token");
        if (maybeIdToken instanceof String) {
            String idToken = (String) maybeIdToken;
            setAsRequestAttribute(OpenIdConnectAuthenticationFilter.ID_TOKEN_VALUE, idToken);
            // among other things, this verifies the token
            if (store != null) store.readAuthentication(idToken);
            // TODO: the authentication just read could contain roles, could be treated as
            // another role source... but needs to be made available to role computation
        }
    }
```
 
If validation needs to be added here we can make a new issue and pull request (and document how validation works at that time).

afabiani:
> > * It would be nice (if you have the chance, otherwise I can eventually contribute) to add some more details on the docs about validating an OpenID Connect ID Token, since most of the modern providers requires it nowdays
> 
> My experience with security topics is limited so I do not feel comfortable offering a clear explanation.
> 
> This pull request only adds validation in one location - where JWKs is being used to check the token.
> 
> The other place GeoServer uses the JWKS URI has a comment indicated it will be used by JwkTokenStore to "verify" ... but I see no evidence of validation being performed by this JwkTokenStore:
> 
> ```
>     private void validate(OAuth2AccessToken token) {
>         Object maybeIdToken = token.getAdditionalInformation().get("id_token");
>         if (maybeIdToken instanceof String) {
>             String idToken = (String) maybeIdToken;
>             setAsRequestAttribute(OpenIdConnectAuthenticationFilter.ID_TOKEN_VALUE, idToken);
>             // among other things, this verifies the token
>             if (store != null) store.readAuthentication(idToken);
>             // TODO: the authentication just read could contain roles, could be treated as
>             // another role source... but needs to be made available to role computation
>         }
>     }
> ```
> 
> If validation needs to be added here we can make a new issue and pull request (and document how validation works at that time).

Ok yes, I see. Unfortunately yes, letting the store validating the `id_token` most probably will fail in case of a custom `KIDs` I'm afraid.
However I agree that this is not something that must/should be addressed by this PR.
I'll try to open a new issue for that, since this will become a problem for sure in the future.

jodygarnett:
> * Did you test this code with a custom KID? From what I can see it should work, just to confirm

@afabiani I am not sure what this means; can you explain? kid is key identifier, is the an option to specify a different key identifier to look up in JWKs ?!

jodygarnett:
Feedback addressed, rebased and cleaned up commit history. Thanks for feedback @davidblasby and @afabiani .

PR #7308, opened 2023-12-01
[GEOS-10225] Fix for missing getImageLayout implementation

Eneuman:
[GEOS-10225](https://osgeo-org.atlassian.net/browse/GEOS-10225) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This PR fixes a problem with publishing layers from GeoPackage Mosaic files in GeoServer.

To reproduce the problem:
1. Download the example file from https://github.com/ngageoint/GeoPackage/blob/master/docs/examples/java/example.gpkg
2. In GeoServer, add a new DataSource and select GeoPackage Mosaic
3. Select the file and click okey.
4. Try and publish a layer from the Datasource.
5. GeoServer web application shows a NullPoint exception.

This happens because the class `SingleGridCoverage2DReader` relies on that the inherited class has `layout `set. This is not the case. `return delegate.getImageLayout(coverageName);` will always return null.

This should resolve issue: https://osgeo-org.atlassian.net/browse/GEOS-10225

(This PR also fixes a typo in one of the null checks.)

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java
+++ b/src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java
@@ -1239,7 +1239,7 @@ public class CatalogBuilder {
                                 + format.getName());
             }
             SampleModel sm = imageLayout.getSampleModel(null);
-            if (cm == null) {
+            if (sm == null) {
                 throw new Exception(
                         "Unable to acquire test coverage and sample model for format:"
                                 + format.getName());
--- a/src/main/src/main/java/org/geoserver/catalog/SingleGridCoverage2DReader.java
+++ b/src/main/src/main/java/org/geoserver/catalog/SingleGridCoverage2DReader.java
@@ -6,6 +6,10 @@
 package org.geoserver.catalog;
 
 import it.geosolutions.imageio.maskband.DatasetLayout;
+import java.awt.Dimension;
+import java.awt.image.BufferedImage;
+import java.awt.image.ColorModel;
+import java.awt.image.SampleModel;
 import java.io.IOException;
 import java.util.List;
 import java.util.Set;
@@ -24,6 +28,7 @@ import org.geotools.coverage.grid.io.GridCoverage2DReader;
 import org.geotools.coverage.grid.io.OverviewPolicy;
 import org.geotools.coverage.grid.io.StructuredGridCoverage2DReader;
 import org.geotools.geometry.GeneralBounds;
+import org.geotools.image.util.ImageUtilities;
 
 /** A wrapper restricting the provided reader to return a single grid coverage */
 public class SingleGridCoverage2DReader implements GridCoverage2DReader {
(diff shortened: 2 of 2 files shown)


aaime:
Please fix the issue in the [GeoTools GeoPackage reader](https://github.com/geotools/geotools/blob/main/modules/plugin/geopkg/src/main/java/org/geotools/geopkg/mosaic/GeoPackageReader.java), rather than recomputing the image layout for types of wrapper readers.

The typo fix is welcomed and should be paired with a test.

Eneuman:
@aaime After a bit more debugging I have tracked down the issue to theese two lines of code:

https://github.com/geoserver/geoserver/blob/7f1bd11d0d2c15f8c8ab8b56e51d4f781cbfd7aa/src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java#L1179
https://github.com/geoserver/geoserver/blob/7f1bd11d0d2c15f8c8ab8b56e51d4f781cbfd7aa/src/main/src/main/java/org/geoserver/catalog/CatalogBuilder.java#L1180

By setting the envelop to just 5x5 pixels makes the resolution calculations in the "`read`" method in `GeoPackageReader` to become all wrong. This in turn makes the TileBoundsCalculator to always return left, bottom, right and top tile as 0.
But the zoomlevel is calculated to 17. So when the SQL runs, the result is empty and it crashes.

`SELECT * FROM "bit_systems" WHERE zoom_level = 17 AND tile_column = 0 AND tile_row = 0`

image

My GeoPackage file is 110Gb large so if the the code is going to try and figure our the sample dimension by looking at the data, I think the GeoPackage `read` method needs to have alimit.

The other aproche I can think of is similar to the one in this PR, just to manually craft a layout with good enough values,

What do you think?

aaime:
@Eneuman sorry for the late reply, too much going on these days, so paid work gets priority.

Yes, I believe the fix should be in the GeoPackage reader code, it must be able to read images of any sort of dimension and target resolution... it should have some limits so that at least one existing tile is read.

Looking at the data though, the issue seems to be that the GeoPackage, as it often happens, covers a wider area at low zoom levels, and then refines/restricts it at higher zoom levels, but there is no way to communicate this variable bbox situation. Hum... and yet... I'm wondering if the reader is just not returning the [right bbox here](https://github.com/geotools/geotools/blob/8ebbe3273bcc3ff1497eeee311ea781437224446/modules/plugin/geopkg/src/main/java/org/geotools/geopkg/mosaic/GeoPackageReader.java#L136).

Shouldn't that be the extent reported in the gpkg_contents instead?  The spec says about the coordinates there " If tiles, this is informational and the tile matrix set should be used for calculating tile coordinates." 
Hum... yeah, that tile matrix set table contains the same contents. How does that work in your case? Do you have a bbox that actually fits the data at every zoom level, or does it go outside?

And then again in the spec it says, for that last table: "The gpkg_tile_matrix_set table defines the spatial reference system (srs_id) and the maximum bounding box (min_x, min_y, max_x, max_y) for all possible tiles in a tile pyramid user data table."

So not quite useful either... maybe one should pick the gpkg_tilemamtrix and use the intersection of all bboxes there instead? Generally speaking, we'd want some bbox that is always valid, as "the bbox of the raster layer".

If this route is not convincing, then the reader should make sure it can find a tile, somehow, when a request is made within the declared bbox.

PR #7328, opened 2023-12-22
[GEOS-11216] Graticule datastore improvements

aaime:
[GEOS-11216](https://osgeo-org.atlassian.net/browse/GEOS-11216) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

A few changes for the store:

* General code simplification, removing redundant methods and javadocs
* Making it possible to generate styles for multi-level graticules whose progression does not allow to re-use lines (e.g., 2,5,10,20,30, where 2 and 5 generate lines that are out of synch)
* Add extra attributes the labelling process to simplify the style, and provide example of compact styles for multi-level graticules


src/community/graticule/src/main/java/org/geotools/data/graticule/GraticuleFeatureSource.java
```diff
@@ -9,58 +9,69 @@
...
-        this.schema = schema;
+    }
+
+    List<Double> getSteps() {
+        return steps;
     }
 
     @Override
     protected ReferencedEnvelope getBoundsInternal(Query query) throws IOException {
+        if (query.getFilter() != Filter.INCLUDE) {
```
ianturton, line 41: are there not filters that should return bounds?

ianturton (approved):

src/community/graticule/src/main/java/org/geotools/data/graticule/GraticuleFeatureSource.java
aaime: According to the javadoc of FeatureSource:

```
@returns The bounding envelope of the feature data; or {@code null} if the bounds are unknown or too costly to calculate.
``` 

So the idea is to return bounds if it's fast, otherwise not. In this case, we are in a debatable situation... features are generated in memory so the computation should be fast unless there is a massive number of lines.. which some madman could indeed configure (and risk bringing GeoServer to OOM).

The implementation here takes the easy route. Did you ask me for everything? Configured bounds. Otherwise, if the user really wants it, you can use ``DataUtilities.bounds``, which will scan every single feature.

PR #7331, opened 2023-12-26
[GEOS-11242] Remove the Xalan library

sikeoka:
[GEOS-11242](https://osgeo-org.atlassian.net/browse/GEOS-11242) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This PR contains the following changes to completely remove the xalan:xalan and xalan:serializer dependencies:

1. WMS: Replaced the Xalan XMLChar class in RasterLayerIdentifier with the Eclipse EMF XMLChar class (covered by org.geoserver.wms.featureinfo.RasterLayerIdentifierTest). Updated the WMS GetCapabilities DTD unit test to be more lenient with the XML prolog.
2. OWS: Refactored unit tests to use the XPath API from XMLUnit instead of Xalan (like other tests do).
3. CSW: Refactored main code to remove xalan:serializer dependency (covered by org.geoserver.csw.GetRecordsTest.testValidatePost).
4. WPS: Refactored main code to remove xalan:serializer dependency and added new unit tests (the UTF-16 to UTF-8 change was necessary for unit tests to pass).
5. SLD Service: Upgraded XOM dependency to remove transitive Xalan dependency and updated unit tests to be more lenient with leading/trailing whitespace.
  

# Checklist

### DIFF
--- a/src/extension/csw/core/pom.xml
+++ b/src/extension/csw/core/pom.xml
@@ -25,16 +25,6 @@
       <artifactId>gs-csw-api</artifactId>
       <version>${project.version}</version>
     </dependency>
-    <dependency>
-      <groupId>xalan</groupId>
-      <artifactId>serializer</artifactId>
-      <exclusions>
-        <exclusion>
-          <artifactId>xml-apis</artifactId>
-          <groupId>xml-apis</groupId>
-        </exclusion>
-      </exclusions>
-    </dependency>
 
     <!-- test dependencies -->
     <dependency>
--- a/src/extension/csw/core/src/main/java/org/geoserver/csw/response/AcknowledgementTransformer.java
+++ b/src/extension/csw/core/src/main/java/org/geoserver/csw/response/AcknowledgementTransformer.java
@@ -10,8 +10,10 @@ import java.util.Date;
 import javax.servlet.http.HttpServletRequest;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.transform.TransformerFactory;
+import javax.xml.transform.dom.DOMSource;
+import javax.xml.transform.sax.SAXResult;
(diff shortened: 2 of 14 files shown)


aaime:
As far as I remeber, web-wps also uses Xalan serializer in the WPS demo request builder. It might be not in your plans, but in that case, the title should be qualified, now it seems like you're trying to remove it from the whole GeoServer.

groldan:
Ideally for good XSLT support we should move to Saxon, which is maintained https://mvnrepository.com/artifact/net.sf.saxon/Saxon-HE

Note however, Xalan had a patch release after 9 years: https://mvnrepository.com/artifact/xalan/xalan/2.7.3

sikeoka:
xalan and serializer are two different dependencies.  This PR is removing xalan but unnecessary serializer references are also being removed (since xalan depends on serializer even though the transitive dependency in the xalan pom was removed for some reason).  The serializer dependencies in csw-core and web-wps are not being touched.  GeoServer did not include xalan (or serializer except in the wps package) until 2.15 and both CSW and WPS existed prior to that.

sikeoka:
The point of this PR is to completely remove the Xalan dependency from the build which it does.  The build and unit tests no longer require the Xalan dependency and it is no longer included in the WAR file.  The internal JDK Xalan will be used if another XML library isn't added to the classpath.  This is the same behavior as prior to GeoServer 2.15.

aaime:
For me the surprise point is that [Xalan serializer does not depend on Xalan at all](https://mvnrepository.com/artifact/xalan/serializer/2.7.1), despite not having a separate repository. Seems like poor naming, but it's not Steve's fault.

sikeoka:
I updated this PR to completely remove both the xalan and serializer dependencies.

sikeoka:
Any additional comments on this PR?

aaime:
Nope, looks good to me.

PR #7336, opened 2024-01-06
[GEOS-11051] Env parametrization does not save correctly in AuthKey extension

petersmythe:
[GEOS-11051](https://osgeo-org.atlassian.net/browse/GEOS-11051) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

**A parameterised webServiceUrl, when loaded from XML is not displayed as parameterised and when saved again, overwrites the parameterisation.  See the above issue for more details/screenshots.**

The problem is that when the filter config is cloned in order to be displayed for editing, the mapperParameters are resolved (i.e. `${...}` placeholders are converted into their real values).  This real value is then saved to the XML file, instead of the unresolved placeholders.

The solution identified [exactly where](https://github.com/geoserver/geoserver/blob/95d0ad588e7d43a58e09a7672ca70c55deac0cbf/src/web/security/core/src/main/java/org/geoserver/security/web/auth/AuthenticationFiltersProvider.java#L26C79-L26C79) the filter configs are created for editing purposes, and only this instance of `loadFilterConfig` is modified to not resolve placeholders, leaving all other instances to resolve, as before.  

This means that the edit box displays the unresolved placeholders, as expected, which is consistent with other edit boxes that typically contain ENV parameter placeholders, e.g. Data store JDBC connection strings, Global proxy base URL, etc.

# Checklist

### DIFF
--- a/src/community/security/oauth2-openid-connect/oauth2-openid-connect-core/src/test/java/org/geoserver/security/oauth2/OpenIdConnectIntegrationTest.java
+++ b/src/community/security/oauth2-openid-connect/oauth2-openid-connect-core/src/test/java/org/geoserver/security/oauth2/OpenIdConnectIntegrationTest.java
@@ -220,7 +220,7 @@ public class OpenIdConnectIntegrationTest extends GeoServerSystemTestSupport {
     public void testClientConfidential() throws Exception {
         GeoServerSecurityManager manager = getSecurityManager();
         OpenIdConnectFilterConfig config =
-                (OpenIdConnectFilterConfig) manager.loadFilterConfig("openidconnect");
+                (OpenIdConnectFilterConfig) manager.loadFilterConfig("openidconnect", true);
         config.setSendClientSecret(true);
         manager.saveFilter(config);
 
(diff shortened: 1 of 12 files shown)


petersmythe:
@aaime I believe that `testMapperParametersFromEnv` already tests the ENV parameterisation correctly.  This bug is more about **displaying** the unresolved placeholders correctly **in the GUI**.  Do you have any suggestions of how to design a test case for that?

petersmythe:
Also, is propagating the `allowEnvParametrization` parameter through the various methods the best way to achieve this fix?

petersmythe:
I am aware that some other tests are currently failing, and I will fix those tomorrow.

> src/community/security/oauth2-openid-connect/oauth2-openid-connect-core/src/test/java/org/geoserver/security/oauth2/OpenIdConnectIntegrationTest.java:[223,52] no suitable method found for loadFilterConfig(java.lang.String)

aaime:
I believe this one is better reviewed by @afabiani, if memory serves me well he worked both on the authkey module and the initial implementation of environment parametrization.

afabiani (commented):
Hi, so far the patch looks good. I would just add a small test case that verifies that the config file placeholder is not overwritten anymore in the case the env parametrization has been enabled for the SecurityManager.

petersmythe:
> Hi, so far the patch looks good. I would just add a small test case that verifies that the config file placeholder is not overwritten anymore in the case the env parametrization has been enabled for the SecurityManager.

Thanks for the review Alessio.  I assumed that you meant when ENV parameterisation has been disabled (or bypassed, with `allowEnvParametrization=false`) so I added this test case.

afabiani (approved):

PR #7337, opened 2024-01-07
OGC API Spring integration cleanups

aaime:
Removing a couple of customizations from APIBodyMethodProcessor, getting closer to a normal Spring one. 
Still cannot use a basic one due a couple of lingering needs:

* OGC API of customized mime type matching
* DispatcherCallback support for responseDispatcher, which is in turn used to create generic HTTP header links support

Will try later to see if we can get rid of those two as well.

# Checklist


src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/AnnotatedHTMLMessageConverter.java
```diff
@@ -0,0 +1,133 @@
...
+    private static final String ANNOTATION = "annotation";
+    private static final String BASE_CLASS = "baseClass";
+
+    private static final String SERVICE_INFO_CLASS = "serviceInfoClass";
+
+    /** Builds the message converter */
+    public AnnotatedHTMLMessageConverter(FreemarkerTemplateSupport support, GeoServer geoServer) {
+        super(support, geoServer);
+    }
+
```
turingtestfail, line 45: Method doc?

turingtestfail (changes requested):

PR #7343, opened 2024-01-10
[GEOS-11250] WFS GeoJSON encoder fails with an exception if an infinity number is used in the geometry

mirco-romagnoli:
[GEOS-11250](https://osgeo-org.atlassian.net/browse/GEOS-11250) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This commit avoids a JSONException in case of a coordinate with infinite value (positive or negative). A coordinate with an infinite value will be encoded as a string "Infinity" or "-Infinity", depending on the coordinate sign.

# Checklist

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
@@ -201,9 +201,9 @@ public class GeoJSONBuilder extends JSONBuilder {
      *
      * <p>If the value is {@link Double#NaN} then the value will not be written.
      *
-     * <p>If the value is {@link Double#POSITIVE_INFINITY} or {@link Double#NEGATIVE_INFINITY} then it will be encoded as
-     * {@code "Infinity"} or {@code "-Infinity"}, respectively, this avoids a {@link JSONException} in case the value
-     * is infinite.
+     * <p>If the value is {@link Double#POSITIVE_INFINITY} or {@link Double#NEGATIVE_INFINITY} then
+     * it will be encoded as {@code "Infinity"} or {@code "-Infinity"}, respectively, this avoids a
+     * {@link JSONException} in case the value is infinite.
      *
      * @param value value to encode
      * @see #setNumberOfDecimals(int)
--- a/src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
@@ -171,7 +171,7 @@ public class GeoJSONBuilder extends JSONBuilder {
      * @param y X ordinate
      * @param z Z ordinate, can be {@code NaN}
      * @param m M ordinate, can be {@code NaN}
-     * @return the JSON builder instance, this allow chained calls
+     * @return the JSON builder instance, this allows chained calls
      */
     private JSONBuilder writeCoordinate(double x, double y, double z, double m) {
         // start encoding JSON array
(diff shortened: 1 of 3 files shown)


ianturton:
Pending the tests passing this looks fine but I can't quite think of an instance when you would need to store an infinity in a coordinate.

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
```diff
@@ -171,39 +171,46 @@ private JSONBuilder writeCoordinates(CoordinateSequence coordinates) throws JSON
...
         // encode M value if available
-        if (!Double.isNaN(m)) {
-            roundedValue(m);
-        }
+        encodeCoordinateIfAvailable(m);
         // we are done with the array
         return this.endArray();
     }
 
+    private void encodeCoordinateIfAvailable(double value) {
```
aaime, line 199:
```suggestion
    private void encodeOrdinate(double value) {
```

Notes:
* This is encoding just a number. It's an "ordinate" not a coordinate. 
* The details of why and how encoding is performed can be provided in the method  javadoc instead. If we wanted to be correct, the name would otherwise be "encodeOrdinateIfAvailableAndInfinityAsString".

aaime (changes requested):
Looks good in general, see one small remark below

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
mirco-romagnoli: Right, I fixed this [in the next commit](https://github.com/geoserver/geoserver/pull/7343/commits/916d424e20988f2ed9add4c46532b85162c5e985)

aaime:
QA failure, formatting issue:

```
Error: 3,761 [ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.22.1:check (default) on project gs-wfs: The following files had format violations:
Error: 3,761 [ERROR]     src/main/java/org/geoserver/wfs/json/GeoJSONBuilder.java
Error: 3,762 [ERROR]         @@ -201,9 +201,9 @@
Error: 3,762 [ERROR]          ·····*
Error: 3,762 [ERROR]          ·····*·<p>If·the·value·is·{@link·Double#NaN}·then·the·value·will·not·be·written.
Error: 3,762 [ERROR]          ·····*
Error: 3,762 [ERROR]         -·····*·<p>If·the·value·is·{@link·Double#POSITIVE_INFINITY}·or·{@link·Double#NEGATIVE_INFINITY}·then·it·will·be·encoded·as
Error: 3,762 [ERROR]         -·····*·{@code·"Infinity"}·or·{@code·"-Infinity"},·respectively,·this·avoids·a·{@link·JSONException}·in·case·the·value
Error: 3,762 [ERROR]         -·····*·is·infinite.
Error: 3,762 [ERROR]         +·····*·<p>If·the·value·is·{@link·Double#POSITIVE_INFINITY}·or·{@link·Double#NEGATIVE_INFINITY}·then
Error: 3,762 [ERROR]         +·····*·it·will·be·encoded·as·{@code·"Infinity"}·or·{@code·"-Infinity"},·respectively,·this·avoids·a
Error: 3,762 [ERROR]         +·····*·{@link·JSONException}·in·case·the·value·is·infinite.
Error: 3,762 [ERROR]          ·····*
Error: 3,765 [ERROR]          ·····*·@param·value·value·to·encode
Error: 3,765 [ERROR]          ·····*·@see·#setNumberOfDecimals(int)
Error: 3,765 [ERROR] Run 'mvn spotless:apply' to fix these violations.
```

Formatting is applied automatically as you build the project with Maven. Just remember to always run a build before committing.

PR #7353, opened 2024-01-22
[GEOS-11263] hideEmptyRules not working in JSON LegendGraphic

iaunzu:
[GEOS-11263](https://osgeo-org.atlassian.net/browse/GEOS-11263) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

Hi

This a a PR for [GEOS-11263](https://osgeo-org.atlassian.net/browse/GEOS-11263)

It applies the same strategy in ``BufferedImageLegendGraphicBuilder`` but in JSONLegendGraphicBuilder to filter out all rules with no features in the BBOX.
  
# Checklist

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/legendgraphic/JSONLegendGraphicBuilder.java
+++ b/src/wms/src/main/java/org/geoserver/wms/legendgraphic/JSONLegendGraphicBuilder.java
@@ -292,6 +292,10 @@ public class JSONLegendGraphicBuilder extends LegendGraphicBuilder {
                 applicableRules = LegendUtils.getApplicableRules(ftStyles, scaleDenominator);
             }
 
+            if (countProcessor != null && !forceLabelsOff) {
+                applicableRules = updateRuleTitles(countProcessor, legend, applicableRules);
+            }
+
             ArrayList<JSONObject> jRules = new ArrayList<>();
             if (legend instanceof CascadedLegendRequest) {
                 CascadedLegendRequest cascadedLegend = (CascadedLegendRequest) legend;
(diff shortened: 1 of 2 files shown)


aaime:
Tests are failing:

```
2024-01-22T14:45:10.4781526Z 14:45:10,477 [INFO] Results:
2024-01-22T14:45:10.4782381Z 14:45:10,477 [INFO] 
2024-01-22T14:45:10.4783130Z 14:45:10,477 [ERROR] Errors: 
2024-01-22T14:45:10.4789750Z 14:45:10,477 [ERROR]   JSONLegendGraphicOutputFormatTest.testLanguageRaster:1757->checkLanguageRule:1795 » JSON
2024-01-22T14:45:10.4791786Z 14:45:10,478 [ERROR]   JSONLegendGraphicOutputFormatTest.testLanguageVector:1747->checkLanguageRule:1795 » JSON
```

iaunzu:
Ok, I pushed a commit attempting to pass those tests. It turns out that hideEmptyRules serves two purposes. One is to filter out those rules with a <ogc:Filter>. The other is to exclude rules not painted on the map and it is heavily dependent on the BBOX. This change targets the latter, performing the process only when a BBOX is specified, assuming that all rules will be painted if no BBOX is present in the request. This should probably be changed for image legends as well.

Also, rules with RasterSymbolizers would always be filtered out and now they are skipped from the process.

aaime (changes requested):
Overall looks good, see comment about making the test more readable.

aaime:
Looks good thanks, merging and scheduling for backport.

PR #7362, opened 2024-01-24
[GEOS-11274] Fix JSON Legend with external reference

iaunzu:
This a a PR for [GEOS-11274](https://osgeo-org.atlassian.net/browse/GEOS-11274)

It prevents NullPointerException providing links in the form of URI Data for unresolvable urls and excluding ``url`` and ``external-graphic-url`` when not possible.
  
# Checklist

- [X] The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers).

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/icons/IconProperties.java
+++ b/src/wms/src/main/java/org/geoserver/wms/icons/IconProperties.java
@@ -8,8 +8,11 @@ package org.geoserver.wms.icons;
 import java.io.File;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.util.Base64;
 import java.util.Collections;
 import java.util.Map;
 import org.geoserver.config.GeoServerDataDirectory;
@@ -44,6 +47,10 @@ public abstract class IconProperties {
     public abstract String getIconName(Style style);
 
     public abstract boolean isExternal();
+
+    public String uriData() {
+        return null;
+    }
 
     public static IconProperties generator(
             final Double opacity,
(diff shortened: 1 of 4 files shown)


src/wms/src/main/java/org/geoserver/wms/icons/IconProperties.java
```diff
@@ -166,8 +173,15 @@ public String href(String baseURL, String workspace, String styleName) {
...
-                                file = null;
+                                // other dirs are not published, serialize uri-data
+                                Path path = file.toPath();
+                                String contentType = Files.probeContentType(path);
+                                byte[] bytes = Files.readAllBytes(path);
+                                return new StringBuilder("data:")
+                                        .append(contentType)
+                                        .append(";base64,")
+                                        .append(Base64.getEncoder().encodeToString(bytes))
+                                        .toString();
```
aaime, line 184: Relative references should work too though, the contents of the styles directory is available to the users.
So for example, the "grass_poly.sld" [file in the release data directory](https://github.com/geoserver/geoserver/blob/main/data/release/styles/grass_poly.sld) contains this relative reference:

                    <OnlineResource xlink:type="simple" xlink:href="grass_fill.png" />

which is accessible using a link to the origin GeoServer, e.g.:

https://gs-main.geosolutionsgroup.com/geoserver/styles/grass_fill.png

Long story short, it seems to me the "file.isAbsolute" test is too strict

aaime (changes requested):

aaime (changes requested):

iaunzu:
Very good point. This is indeed a serious security breach. At the very least, I should have verified that the resource is in the styles directory, just as it is checked for the main styles directory. This condition should be verified as a security check if the resource is serialized, or as a simple precaution if an URL is going to be generated.

On the other hand, I just realized that the contents of the styles directory for specific workspaces ARE published. However, the URL is not what I expected. For example, _geoserver/styles/sf/ExternalGraphicIcon.png_ does work.

I am going to push some changes with this in mind.

aaime:
I did not notice your last change. Could you update the PR and align it with the main branch? (there are a couple of conflicts)

iaunzu:
Done. Also fixed the test.

src/wms/src/test/resources/org/geoserver/wms/legendgraphic/ExternalReferenceGraphic.sld
```diff
@@ -0,0 +1,28 @@
+<StyledLayerDescriptor xmlns="http://www.opengis.net/sld" xmlns:ogc="http://www.opengis.net/ogc" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
aaime, line 1: While toppSample below is used, I don't see this new file being used anywhere

PR #7363, opened 2024-01-26
Use threads to read altitude in wps-longitudinal-profile community module

jusabatier:
This PR modify the wps-longitudinal-profile community module to use thread for altitude reading.
This way it allow to process more points in less time, depending on available number of cores.

# Checklist

### DIFF
--- a/src/community/wps-longitudinal-profile/src/main/java/org/geoserver/wps/longitudinal/AltitudeReaderThread.java
+++ b/src/community/wps-longitudinal-profile/src/main/java/org/geoserver/wps/longitudinal/AltitudeReaderThread.java
@@ -6,8 +6,8 @@ package org.geoserver.wps.longitudinal;
 
 import java.math.BigDecimal;
 import java.math.RoundingMode;
-import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.logging.Logger;
@@ -27,19 +27,19 @@ import org.locationtech.jts.geom.Geometry;
 import org.locationtech.jts.geom.GeometryFactory;
 import org.locationtech.jts.geom.Point;
 
-public class AltitudeReaderThread implements Callable<ArrayList<ProfileVertice>> {
+public class AltitudeReaderThread implements Callable<List<ProfileVertice>> {
     static final Logger LOGGER = Logging.getLogger(AltitudeReaderThread.class);
 
     private static final GeometryFactory GEOMETRY_FACTORY = new GeometryFactory();
 
-    private ArrayList<ProfileVertice> pvs;
+    private List<ProfileVertice> pvs;
     GridCoverage2D gridCoverage2D;
     private int altitudeIndex;
     FeatureSource adjustmentFeatureSource;
     String altitudeName;
 
     public AltitudeReaderThread(
-            ArrayList<ProfileVertice> pvs,
+            List<ProfileVertice> pvs,
             int altitudeIndex,
             FeatureSource adjustmentFeatureSource,
             String altitudeName,
@@ -52,7 +52,7 @@ public class AltitudeReaderThread implements Callable<ArrayList<ProfileVertice>>
     }
 
     @Override
-    public ArrayList<ProfileVertice> call() throws Exception {
+    public List<ProfileVertice> call() throws Exception {
         Coordinate[] coords =
                 pvs.stream()
                         .map(ProfileVertice::getCoordinate)
@@ -60,23 +60,20 @@ public class AltitudeReaderThread implements Callable<ArrayList<ProfileVertice>>
                         .toArray(new Coordinate[pvs.size()]);
         Geometry geometry = GEOMETRY_FACTORY.createLineString(coords);
 
-        Map<Geometry, Double> adjGeomValues = new HashMap<Geometry, Double>();
+        Map<Geometry, Double> adjGeomValues = new HashMap<>();
         if (adjustmentFeatureSource != null) {
             Query query;
             Filter filter;
             filter = CQL.toFilter("INTERSECTS(the_geom, " + geometry.toText() + ")");
             query = new Query(adjustmentFeatureSource.getSchema().getName().getLocalPart(), filter);
-            FeatureIterator<?> featureIterator =
-                    adjustmentFeatureSource.getFeatures(query).features();
-            try {
+            try (FeatureIterator<?> featureIterator =
+                    adjustmentFeatureSource.getFeatures(query).features()) {
                 while (featureIterator.hasNext()) {
                     Feature f = featureIterator.next();
                     Geometry g = (Geometry) f.getDefaultGeometryProperty().getValue();
                     Double altitude = (Double) f.getProperty(altitudeName).getValue();
                     adjGeomValues.put(g, altitude);
                 }
-            } finally {
-                featureIterator.close();
             }
         }
 
--- a/src/community/wps-longitudinal-profile/src/main/java/org/geoserver/wps/longitudinal/AltitudeReaderThread.java
+++ b/src/community/wps-longitudinal-profile/src/main/java/org/geoserver/wps/longitudinal/AltitudeReaderThread.java
@@ -7,7 +7,7 @@ package org.geoserver.wps.longitudinal;
 import java.io.IOException;
 import java.math.BigDecimal;
 import java.math.RoundingMode;
-import java.util.Vector;
+import java.util.ArrayList;
 import java.util.concurrent.Callable;
 import java.util.logging.Logger;
 import org.geotools.api.data.FeatureSource;
@@ -21,17 +21,17 @@ import org.geotools.filter.text.cql2.CQLException;
 import org.geotools.geometry.Position2D;
 import org.geotools.util.logging.Logging;
 
-public class AltitudeReaderThread implements Callable<Vector<ProfileVertice>> {
+public class AltitudeReaderThread implements Callable<ArrayList<ProfileVertice>> {
     static final Logger LOGGER = Logging.getLogger(AltitudeReaderThread.class);
 
-    private Vector<ProfileVertice> pvs;
+    private ArrayList<ProfileVertice> pvs;
     GridCoverage2D gridCoverage2D;
     private int altitudeIndex;
     FeatureSource adjustmentFeatureSource;
     String altitudeName;
 
     public AltitudeReaderThread(
-            Vector<ProfileVertice> pvs,
+            ArrayList<ProfileVertice> pvs,
             int altitudeIndex,
             FeatureSource adjustmentFeatureSource,
             String altitudeName,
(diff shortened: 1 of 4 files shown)


jusabatier:
Curently I use `Runtime.getRuntime().availableProcessors();` to set number of cores to use.
Maybe change this using a GS configuration option if it exists ?

src/community/wps-longitudinal-profile/src/main/java/org/geoserver/wps/longitudinal/LongitudinalProfileProcess.java
```diff
@@ -174,27 +170,79 @@ public LongitudinalProfileProcessResult execute(
...
+            }
+            if (overflow > 0) {
+                chunk.add(vertices.pop());
+                overflow = overflow - 1;
+            }
+            chunks.add(chunk);
+        }
+
+        // Create threads executor
+        ExecutorService executor = Executors.newFixedThreadPool(cores);
```
aaime, line 204: Creating a separate pool for each process invocation is source for trouble, if a few WPS request pile up you'll get ``N*cores`` threads running. Also, each thread queries the data sources, which would easily lead to starvation of the connection pools. Long story short, as written this change is surely going to benefit a server with light load, but would crush it once the load goes up.

A better version would be to have a single thread pool for the whole process (held as a field), so that light load gets more cores, but higher loads cannot diverge. 

But generally speaking, it's best not to have extra threads. For example, if the time is mostly spent querying the data sources, better to accumulate a batch of coordinates to adjust (not too many, to avoid OOMs) and do a single query grabbing all adjustment features instead.

jusabatier: > A better version would be to have a single thread pool for the whole process (held as a field), so that light load gets more cores, but higher loads cannot diverge.

I'm trying to move the thread pool to class level as you suggest.

Have you some suggestions for manage the shutdown ?
Should I just put it in finalise() ? 
Or use application event like there : https://github.com/geoserver/geoserver/blob/45de065ff2219ac754fb567949169f57c629b74f/src/extension/wps/wps-core/src/main/java/org/geoserver/wps/executor/WPSExecutionManager.java#L330-L339 ?

> For example, if the time is mostly spent querying the data sources, better to accumulate a batch of coordinates to adjust (not too many, to avoid OOMs) and do a single query grabbing all adjustment features instead.

Not sure to understand this part. The code for the adjustment part is the same as already existing one. I just put it in thread to perform the adjustment directly after altitude reading. But the amount of vertice to read and adjust is exactly the same as current code. It depends on the step and geometry params. Maybe it can be improved, but not related to this PR.

aaime: There are two choices here:
* Keep the algorithm as is, use multiple threads. This is the brute-force way of addressing the issue.
* Improve the existing algorithm so that it does not do as many queries, thus, less back and forth between the process and the adjustment data source. This should lead to improved performance without having to call for extra threads (work smarter, not harder)

The current code adjusts one point at a time, and as a result, makes a query to the adjustment source per point. Each query has a communication latency and an execution time. The total latency here is num points * single query latency, can pile up and become dominant.
If you collect a bunch of points (e.g., 100) and query the adjustment source for an intersection over all of them, you pay 1 latency only for that query: the overall latency is reduced by a factor of 100.

The downside of this approach is that it requires more complex code, the upside is that it will speed up execution without requiring multiple threads (overall, it will scale up better).

One final note: while I'd guess the second approach will be better overall (better performance, better scalability) , I cannot be sure of it. A profiling session would be needed, to check where the time is actually spent (if you see it's spent communicating with the database, then you have proof that latency is the source of slowness).

jusabatier: Ok, but the aim of my PR is to improve reading of the altitude from coverage by using multiple threads.
In the original WPS process I develop (on which this one is based) I manage to improve the process duration with this method.

The aim of this PR is to introduce threading to process vertices faster.

The adjustment function is introduced with this WPS process, and I agree it could be improved, but is not related to this PR.

I think there are multiple vector of improvment that are not exclusive : 

- Use multiple threading to read raw altitude (what this PR intend to do), becaus we must read this value for all vertices
- Improve the adjustment process by trying to perform it on group of vertices (to do in another dedicated PR)

src/community/wps-longitudinal-profile/src/main/java/org/geoserver/wps/longitudinal/AltitudeReaderThread.java
```diff
@@ -0,0 +1,130 @@
+package org.geoserver.wps.longitudinal;
```
aaime, line 1: Needs the GeoServer copyright header

src/community/wps-longitudinal-profile/src/main/java/org/geoserver/wps/longitudinal/LongitudinalProfileProcess.java
```diff
@@ -174,27 +170,79 @@ public LongitudinalProfileProcessResult execute(
...
+            treated.add(
+                    executor.submit(
+                            new AltitudeReaderThread(
+                                    chunks.pop(),
+                                    altitudeIndex,
+                                    adjustmentFeatureSource,
+                                    altitudeName,
+                                    gridCoverage2D)));
+        }
+        Vector<ProfileVertice> result = new Vector<ProfileVertice>();
```
aaime, line 222: Usage of Vector is not allowed by GeoServer QA rules. This is a community module, so the PMD checks are not allowed, but it would be best to avoid its usage anyways, if you have any goal to see this module become a supported extension, in the future. A simple ArrayList will do.

src/community/wps-longitudinal-profile/src/main/java/org/geoserver/wps/longitudinal/ProfileVertice.java
```diff
@@ -0,0 +1,42 @@
+package org.geoserver.wps.longitudinal;
```
aaime, line 1: Copyright header missing

aaime (changes requested):
Ok, did a second round of review assuming you're going for the multithreading.
Please make sure the copyright headers are available in all new files, and avoid usage of outdated Java classes (Vector in particular, while Stack is not yet disallowed in GeoServer, though it might become in the future).

For shutting down the pool, the process is already a Spring bean, so you can implement DisposableBean and shut down the pool inside the destroy() method.

aaime:
And oh, regarding the number of threads in the process executor... Ideally, it should be configurable with a GUI panel, but that takes time. I'd recommend defaulting to the number of cores, yes, but also allow to change the value by defining a system variable, or if you want a more complete treatment, use GeoServerExtensions.getProperty, which also looks in environment variables and in web.xml.

jusabatier:
I corrected all that you suggested.

Thanks for your review, since it's my first contrib to GS, I am not yet comfortable with all the procedures to follow.

I will keep in mind your suggestion to improve altitude adjustment process and submit another PR for this when I've time.

aaime:
I did not look at the algorithm, but from the description, it suffers from the same problem as the original thread implementation: you cannot have all the resources of the server for a single invocation. 
In this case, if the adjustment geometries are too many (we have no control over that layer) this will eventually lead to an out-of-memory. 
GeoServer never loads all the inputs in memory for that very reason. During map rendering, we load inputs in streams/batches, so we do for WFS output, the data is never held in memory at the same time.

Going back to the original suggestion:

> For example, if the time is mostly spent querying the data sources, better to accumulate a batch of coordinates to adjust (not too many, to avoid OOMs) and do a single query grabbing all adjustment features instead.

That's what I meant for "accumulate a batch". A batch is a set of features to adjust, controllable with some parameter (e.g., 1000 points by default). This is already good enough to remove the chattiness of the original algorithm, as it reduces the number of queries to the adjustment data source from "num points" to "num points / batch size".

aaime:
Thanks, that's better. I did a round of build to verify the tests and QA (they are not run for community modules) and noticed the module was passing QA before, but it's not anymore:

```
[INFO] --- pmd:3.14.0:check (default) @ gs-wps-longitudinal-profile ---
[INFO] PMD version: 6.42.0
[INFO] PMD Failure: org.geoserver.wps.longitudinal.AltitudeReaderThread:63 Rule:UseDiamondOperator Priority:3 Explicit type arguments can be replaced by Diamond Operator.
[INFO] PMD Failure: org.geoserver.wps.longitudinal.AltitudeReaderThread:71 Rule:UseTryWithResources Priority:3 Consider using a try-with-resources statement instead of explicitly closing the resource.
[INFO] PMD Failure: org.geoserver.wps.longitudinal.LongitudinalProfileProcess:194 Rule:UseDiamondOperator Priority:3 Explicit type arguments can be replaced by Diamond Operator.
[INFO] PMD Failure: org.geoserver.wps.longitudinal.LongitudinalProfileProcess:212 Rule:UseDiamondOperator Priority:3 Explicit type arguments can be replaced by Diamond Operator.
[INFO] PMD Failure: org.geoserver.wps.longitudinal.LongitudinalProfileProcess:229 Rule:UseDiamondOperator Priority:3 Explicit type arguments can be replaced by Diamond Operator.
[INFO] PMD Failure: org.geoserver.wps.longitudinal.LongitudinalProfileProcess:410 Rule:UseDiamondOperator Priority:3 Explicit type arguments can be replaced by Diamond Operator.
[INFO] PMD Failure: org.geoserver.wps.longitudinal.AltitudeReaderThread:63 Rule:UseDiamondOperator Priority:3 Explicit type arguments can be replaced by Diamond Operator.
[INFO] PMD Failure: org.geoserver.wps.longitudinal.AltitudeReaderThread:71 Rule:UseTryWithResources Priority:3 Consider using a try-with-resources statement instead of explicitly closing the resource.
[INFO] PMD Failure: org.geoserver.wps.longitudinal.LongitudinalProfileProcess:194 Rule:UseDiamondOperator Priority:3 Explicit type arguments can be replaced by Diamond Operator.
[INFO] PMD Failure: org.geoserver.wps.longitudinal.LongitudinalProfileProcess:212 Rule:UseDiamondOperator Priority:3 Explicit type arguments can be replaced by Diamond Operator.
[INFO] PMD Failure: org.geoserver.wps.longitudinal.LongitudinalProfileProcess:229 Rule:UseDiamondOperator Priority:3 Explicit type arguments can be replaced by Diamond Operator.
[INFO] PMD Failure: org.geoserver.wps.longitudinal.LongitudinalProfileProcess:410 Rule:UseDiamondOperator Priority:3 Explicit type arguments can be replaced by Diamond Operator.
```

Also noticed that there is a style regression, streams have been replaced by loops, variables are declared using implementation classes rather than interfaces (ArrayList vs List)... that's a very Java 5 programming style, but in GeoServer we want Java 11+. One last cleanup, please, and then this PR is ready to be merged.

You can manually run the QA on the module using "mvn clean install -Dqa`` on the command line (the above are PMD failures, there might be others found by the other QA tools).

jusabatier:
All points you mentioned should be fixed.

aaime (approved):

aaime:
Good, merged.

aaime:
Oh, noticed it did not have a Jira ticket. Next time please create one and follow the convention on commit messages.
For this time, I've created one for you:
https://osgeo-org.atlassian.net/browse/GEOS-11293

PR #7369, opened 2024-02-01
[GEOS-11289] Enable Spring Security StrictHttpFirewall by default

sikeoka:
[GEOS-11289](https://osgeo-org.atlassian.net/browse/GEOS-11289) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This PR enables using the StrictHttpFirewall by default and adds a system property to disable it.  See the ticket for additional details about this.

This PR may introduce backwards-compatibility issues for some users and should not be backported.
  

# Checklist

### DIFF
--- a/doc/en/user/source/production/config.rst
+++ b/doc/en/user/source/production/config.rst
@@ -219,6 +219,17 @@ GeoServer provides a number of facilities to control external entity resolution:
   
   This setting prevents ``ENTITY_RESOLUTION_ALLOWLIST`` from being used.
 
+.. _production_config_spring_firewall:
+
+Spring Security Firewall
+------------------------
+
+GeoServer defaults to using Spring Security's StrictHttpFirewall to help improve protection against potentially malicious
+requests. However, some users will need to disable the StrictHttpFirewall if the names of GeoServer resources (workspaces,
+layers, styles, etc.) in URL paths need to contain encoded percent, encoded period or decoded or encoded semicolon characters.
+The ``GEOSERVER_USE_STRICT_FIREWALL`` property can be set to false either via Java system property, command line argument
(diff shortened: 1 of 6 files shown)


src/main/src/main/java/org/geoserver/security/GeoServerHttpFirewall.java
```diff
@@ -0,0 +1,86 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletRequestWrapper;
+import javax.servlet.http.HttpServletResponse;
+import org.geoserver.platform.GeoServerExtensions;
+import org.springframework.security.web.firewall.DefaultHttpFirewall;
+import org.springframework.security.web.firewall.FirewalledRequest;
+import org.springframework.security.web.firewall.HttpFirewall;
+import org.springframework.security.web.firewall.StrictHttpFirewall;
+
+public class GeoServerHttpFirewall implements HttpFirewall {
```
aaime, line 17: Have Javadoc on public classes.

doc/en/user/source/production/config.rst
```diff
@@ -219,6 +219,15 @@ GeoServer provides a number of facilities to control external entity resolution:
   
   This setting prevents ``ENTITY_RESOLUTION_ALLOWLIST`` from being used.
 
+Spring Security Firewall
```
aaime, line 222: Good to have information here, but we should have a mention in the upgrade notes as well, e.g.:

https://docs.geoserver.org/stable/en/user/installation/upgrade.html#notes-on-upgrading-specific-versions

aaime (changes requested):
Looks good to me, just a couple of quick changes and it's good to go.

PR #7420, opened 2024-02-12
[GEOS-11300] Centralize access to static web files

sikeoka:
[GEOS-11300](https://osgeo-org.atlassian.net/browse/GEOS-11300) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This PR changes the text/html and application/javascript content types to text/plain in the REST Resources API, Style Publisher and File Publisher.  `/index.html` is a special case that File Publisher will allow the text/html content type and a system property was added to re-enable those mime types in the File Publisher only for the www subdirectory.  This will prevent stored XSS in the File Publisher using HTML files by default and will also prevent `<iframe src="foo.html">` and  `<script src="foo.js">` from remotely accessing static files in the data directory except through `www`.

The `Content-Disposition: attachment` for XML content types that was previously added to the Style Publisher was also added to the File Publisher to prevent stored XSS using XML or SVG files there.

Also, the line about a 415 error status was removed since it isn't used by the current code.

This is a backward-incompatible change for anyone currently using the `www` directory to serve HTML and JavaScript files and should not be backported.
  

# Checklist


aaime:
The ``www`` directory is designed to package a HTML/js application in the data directory, that's [its goal in life](https://docs.geoserver.org/stable/en/user/datadirectory/structure.html#www), see also [serving static files](https://docs.geoserver.org/stable/en/user/tutorials/staticfiles.html#tutorials-staticfiles). 

We might as well just remove the "www" publishing at this point? Just thinking out loud.
This is a most serious compatibility break.

sikeoka:
Well it can be used for other types of files too. I've put pdf and docx files in there.

Completely removing this feature seems a little too extreme and disabling this by default is a lot less of a compatibility break.

jodygarnett:
If I can recommend a counter proposal, please leave `www` as is (most people use this for small webapps - usually an open layers map).

And make a setting to turn on the `www` folder a system admin does not trust the administrator to be responsible this feature.

sikeoka:
I updated the PR so that the files will be allowed by default and an administrator will have to disable them.

jodygarnett:
Thanks @sikeoka, while I have never used the `www` folder to store PDF or word files and not web pages before.
Your change does make that possible!

Reference: https://en.wikipedia.org/wiki/World_Wide_Web#WWW_prefix

src/ows/src/main/java/org/geoserver/ows/FilePublisher.java
```diff
@@ -61,12 +70,32 @@ protected void initServletContext(ServletContext servletContext) {
...
-        reqPath = reqPath.substring(ctxPath.length());
+    protected String getMimeType(String reqPath, String filename) {
+        String mimeType = super.getMimeType(reqPath, filename);
+        if (!"/index.html".equals(reqPath)) {
+            String lowerCaseMime = mimeType.toLowerCase();
+            // force using the static web files directory for html/javascript files and only allow
+            // the html/javascript mime types when the system property is set to true to mitigate
+            // stored XSS vulnerabilities
+            if (lowerCaseMime.contains("html") || lowerCaseMime.contains("javascript")) {
+                String prop = GeoServerExtensions.getProperty(DISABLE_STATIC_WEB_FILES);
```
jodygarnett, line 81: It may be worth looking this value up on construction; although this is not very performance sensitive.

jodygarnett (approved):
Some notes for discussion; but I like the change and would be happy to see it merged.

doc/en/user/source/tutorials/staticfiles.rst
```diff
@@ -7,6 +7,11 @@ You can place static files in the ``www`` subdirectory of the GeoServer :ref:`da
 
 This approach has some limitations:
 
-* GeoServer can only serve files whose MIME type is recognized. If you get an HTTP 415 error, this is because GeoServer cannot determine a file's MIME type.
 * This approach does not make use of accelerators such as the `Tomcat APR library <http://tomcat.apache.org/tomcat-7.0-doc/apr.html>`_. If you have many static files to be served at high speed, you may wish to create your own web app to be deployed along with GeoServer or use a separate web server to serve the content.
 
+The ``GEOSERVER_DISABLE_STATIC_WEB_FILES`` property can be set to true convert the ``text/html`` and ``application/javascript``
```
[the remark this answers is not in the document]
sikeoka: I added some info about this to the production considerations page.

src/ows/src/main/java/org/geoserver/ows/FilePublisher.java
sikeoka: I don't think that would be too much of a concern here. At worst, GeoServerExtensions.getProperty is basically just doing a lookup in four different maps which is not a big deal. Also, it is known at this point that there is a valid HTML/JS file that will need to be read from the file system and transmitted to the user so calling getProperty is just a tiny fraction of the request runtime.

jodygarnett:
Thanks for addressing my comments/questions - if there is nothing else holding this back we should merge and backport.

aaime:
The backport to `2.24.x` failed:
```
```
<details>
<summary>stderr</summary>

```
error: could not apply 2878f49acf... [GEOS-11300] Centralize access to static web files

```
</details>

<details>
<summary>stdout</summary>

```
Auto-merging doc/en/user/source/production/config.rst
CONFLICT (content): Merge conflict in doc/en/user/source/production/config.rst

```
</details>

```bash
git fetch
# Create a new working tree
git worktree add .worktrees/backport-2.24.x 2.24.x
cd .worktrees/backport-2.24.x
# Create a new branch
git switch --create backport-7420-to-2.24.x
git cherry-pick 2878f49acf6abd5f12a78db44c711d7605d217fb
# Push it to GitHub
git push --set-upstream origin backport-7420-to-2.24.x
cd ../..
# Delete the working tree
git worktree remove .worktrees/backport-2.24.x
```
Then, create a pull request where the `base` branch is `2.24.x` and the `compare`/`head` branch is `backport-7420-to-2.24.x`.

PR #7428, opened 2024-02-16
[GEOS-11306] API Fix for GetFeature lazy JDBC count(*) in Java 21 and Java 17

jodygarnett:
[GEOS-11306](https://osgeo-org.atlassian.net/browse/GEOS-11306) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This required an update to apache-commons-lang3 covered by [GEOS-11307].

This issue is with respect to operating in Java 17 and Java 21 environments; relying on GitHub workflows to check regressions here (rather than new test cases).
  

# Checklist

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java
@@ -17,6 +17,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Supplier;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
@@ -26,7 +27,6 @@ import net.opengis.wfs.XlinkPropertyNameType;
 import net.opengis.wfs20.QueryType;
 import net.opengis.wfs20.ResultTypeType;
 import net.opengis.wfs20.StoredQueryType;
-import org.apache.commons.lang3.SystemUtils;
 import org.geoserver.catalog.AttributeTypeInfo;
 import org.geoserver.catalog.Catalog;
 import org.geoserver.catalog.FeatureTypeInfo;
@@ -120,8 +120,6 @@ import org.geotools.referencing.crs.DefaultGeographicCRS;
 import org.geotools.util.factory.Hints;
 import org.geotools.xsd.Encoder;
 import org.locationtech.jts.geom.Polygon;
-import org.springframework.cglib.proxy.Enhancer;
-import org.springframework.cglib.proxy.LazyLoader;
 import org.xml.sax.helpers.NamespaceSupport;
 
 /**
@@ -305,7 +303,7 @@ public class GetFeature {
 
         boolean isNumberMatchedSkipped = false;
         int count = 0; // should probably be long
-        BigInteger totalCount = BigInteger.ZERO;
+        Supplier<BigInteger> totalCount = () -> BigInteger.ZERO;
 
         // offset into result set in which to return features
         int totalOffset = request.getStartIndex() != null ? request.getStartIndex().intValue() : -1;
@@ -682,7 +680,20 @@ public class GetFeature {
         }
     }
 
-    private BigInteger updateTotalCount(
+    /**
+     * Total count represents the total count of the features matched for this query in cases where
+     * the client has limited the result set size, so we compute it lazily.
+     *
+     * @param maxFeatures
+     * @param isNumberMatchedSkipped
+     * @param count
+     * @param totalOffset
+     * @param calculateSize
+     * @param totalCountExecutors
+     * @return Lazy calculation of total count, or {@code null} if isNumberMatchedSkipped
+     * @throws IOException
+     */
+    private Supplier<BigInteger> updateTotalCount(
             int maxFeatures,
             boolean isNumberMatchedSkipped,
             int count,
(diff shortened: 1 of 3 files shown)


src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java
```diff
@@ -703,7 +703,11 @@ private BigInteger updateTotalCount(
             long total = getTotalCount(totalCountExecutors);
             totalCount = BigInteger.valueOf(total);
         } else {
-            if (SystemUtils.IS_JAVA_11 || SystemUtils.IS_JAVA_1_8) {
+            if (SystemUtils.IS_JAVA_21
+                    || SystemUtils.IS_JAVA_17
+                    || SystemUtils.IS_JAVA_11
+                    || SystemUtils.IS_JAVA_1_8) {
```
mprins, line 709: Java 8 is no longer a supported platform

aaime: That's a good point indeed. Perhaps this is more compact:

```
     if (SystemUtils.isJavaVersionAtMost(SystemUtils.IS_JAVA_21)
```

jodygarnett:
Folks I am suspicious that both JDK 17 and JDK 21 build are failing, consistently in `gs-app-schema-test`.

I am starting to suspect that just updating to compatible library is not sufficient.

aaime:
@jodygarnett we recently merged a couple of test fixes from @groldan, exactly in app-schema... is this branch up to date? Maybe you just need to rebase.

jodygarnett:
I will rebase, however since the failure is to do with count receiving 0 (rather than 2) I expect it has to do with this change directly.

src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java
```diff
@@ -703,7 +703,8 @@ private BigInteger updateTotalCount(
             long total = getTotalCount(totalCountExecutors);
             totalCount = BigInteger.valueOf(total);
         } else {
-            if (SystemUtils.IS_JAVA_11 || SystemUtils.IS_JAVA_1_8) {
+            if (SystemUtils.isJavaVersionAtMost(SystemUtils.IS_JAVA_21)) {
```
aaime, line 706:
```suggestion
            if (SystemUtils.isJavaVersionAtMost(JavaVersion.JAVA_21)) {
```

jodygarnett:
Testing with Java 17, and we have an answer:

```
org.springframework.cglib.core.CodeGenerationException: java.lang.IllegalAccessException-->module java.base does not open java.math to unnamed module @1068e947
```

While GeoServer does not log this exception it prevents this approach from working.

jodygarnett:
Adding ``--add-opens=java.base/java.math=ALL-UNNAMED`` does not help:
```
java.lang.IllegalArgumentException: $java.math.BigInteger$$EnhancerByCGLIB$$e4f5dd18 not in same package as lookup class
```

jodygarnett:
I am pushing up some defensive programming here, but the issue of a performance regression on Java 17 remains open.

I think we will need to make an API change here to use Number (rather than BigInteger).

jodygarnett:
Yeah, and the time of the release cycle is good for an API change *why does everything happen right before release candidate....*

jodygarnett:
@aaime I have some possible funding to address this; but it is unlikely to make the release candidate deadline:

Option: I could make the API change to Number now, leaving room for future development.

Option: I could wait and ask for permission to make an API and backport due to how impactful the regression is.

The availability of "run in Java 11" mitigation does take some pressure off.

aaime:
Random ideas here... the reason to have the proxy is that the number is not used in some output formats, so it's computed lazily. If you go for an API change, rather than Number it could be a Provider<BigInteger>, makes things explicit.

That could be done at the level of the wrapper, FeatureCollectionResponse, where you set the provider, and force the computation only when the property is actually requested (few formats do) or when unadapt over WFS 2.0 requests is called, which is done for GML only, and update the WFS 2.0 adaptee at that point (for WFS 1.1 the total number is not require). 
That would make for a more isolated change I believe?

jodygarnett:
I like it, that sounds cleaner then just changing to Number, and being forced to implement my own Number class.

Good thinking Andrea!

jodygarnett:
I really like the Supplier<BigInteger> change proposed; I do not have customer approval this week so risk requesting a API change backport. 

If anyone else wishes to pick this up and make the API change I do have capacity to assist / review / test.

jodygarnett:
Jotting down some notes on how to implement delegating getTotalNumberOfFeatures() by default.
```
public FeatureCollectionResult extends RequestObject {
    Supplier<BigInteger> lazyTotalNumberOfFeatures = () -> getTotalNumberOfFeatures();
    public Supplier<BigInteger> getLazyTotalNumberOfFeatures() { ... }
    public void setLazyTotalNumberOfFeatures(Supplier<BigInteger>) { ... }
    public abstract BigInteger getTotalNumberOfFeatures();
    public void setNumberOfFeatures(BigInteger n;)
}
```

aaime:
Yes. Mind you need to also update "unadapt" so that it sets the property in the EMF model, for the GML encoders to work properly.

jodygarnett:
Gak, I have recieved the go ahead to do this API change this morning. Going to update this PR and ask for review.

I am considering this a pretty large performance regression so I hope it can be treated as a bug with respect to back porting.

jodygarnett:
> Yes. Mind you need to also update "unadapt" so that it sets the property in the EMF model, for the GML encoders to work properly.

@aaime I did not see the need to do this as I have implemented setTotalCount(BigInteger) and getTotalCount() to respect the EMF model. Perhaps I am not understanding EObject workflow in sufficient detail.

src/wfs/src/main/java/org/geoserver/wfs/request/FeatureCollectionResponse.java
```diff
@@ -53,15 +62,84 @@ public void setTimeStamp(Calendar timeStamp) {
...
     public abstract BigInteger getNumberOfFeatures();
 
+    /**
+     * Number of features included in response.
+     *
+     * @param n Number of features included in response
+     */
     public abstract void setNumberOfFeatures(BigInteger n);
 
-    public abstract BigInteger getTotalNumberOfFeatures();
```
aaime, line 62: I'd keep the existing getter and setter based on the lazy supplier

src/wfs/src/main/java/org/geoserver/wfs/request/FeatureCollectionResponse.java
```diff
@@ -53,15 +62,84 @@ public void setTimeStamp(Calendar timeStamp) {
...
+     * Used to calculate total number of features on demand (only if needed). This allows formats
+     * that do not need the total to avoid calculating this expensive result, it also may be that
+     * some data stores can better estimate this total is obtained after traversing results.
+     *
+     * @param totalNumberOfFeatures Delayed calculation of total number of featuers.
+     */
+    public void setLazyTotalNumberOfFeatures(Supplier<BigInteger> totalNumberOfFeatures) {
+        if (lazyTotalNumberOfFeatures != null) {
+            throw new IllegalStateException("Total number of features has already been set");
+        }
```
aaime, line 98: No need to check this, especially if you remove the getter for the supplier, which does not seem to be required anyways.

aaime (dismissed):

aaime:
```
```
<details>
<summary>stderr</summary>

```
error: could not apply dc9094a3e1... [GEOS-11306] Java 17 does not support GetFeature lazy JDBC count(*)

```
</details>

<details>
<summary>stdout</summary>

```
Auto-merging src/pom.xml
[backport-7428-to-2.23.x 062021fa02] [GEOS-11307] Update apache-commons-lang3 to version 3.14.0 for Java 17 and Java 21 support
 Author: Jody Garnett <jody.garnett@gmail.com>
 Date: Fri Feb 16 14:07:06 2024 -0800
 1 file changed, 1 insertion(+), 1 deletion(-)
Auto-merging src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java
CONFLICT (content): Merge conflict in src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java

```
</details>

```bash
git fetch
# Create a new working tree
cd .worktrees/backport-2.23.x
# Create a new branch
git switch --create backport-7428-to-2.23.x
git cherry-pick 6d77ecb96199190671939508606f96b26eca1853,dc9094a3e1110a1f6601bbee227b77616f1cacc6
# Push it to GitHub
git push --set-upstream origin backport-7428-to-2.23.x
cd ../..
# Delete the working tree
```
Then, create a pull request where the `base` branch is `2.23.x` and the `compare`/`head` branch is `backport-7428-to-2.23.x`.

aaime:
@jodygarnett found the new lazy counter is not quite working as expected, it's true that it's delaying the count, but it's not saving the result. The output machinery can ask for the value multiple times, and as a result, it gets computed multiple times. 
Long story short, it made things better for output formats that do not need the count, but worse for those that do...

klbjerg:
Could be improved by wrapping in Guava Supplier memoize( ) - see https://guava.dev/releases/31.0-jre/api/docs/com/google/common/base/Suppliers.html#memoize(com.google.common.base.Supplier)

aaime:
We already have guava in the classpath, just FYI.

KorewaLidesu:
> This PR does not address the issue, but does provide a debug message to the logs now to indicate the problem.
> 
> As such it could be accepted "as is", leaving the issue open.
> 
> (Of course if anyone knows how to open `java.math` to reflection and constructor call from a different package we may be able to address the issue with a documentation fix)

You can check out this project
https://github.com/Rongmario/ImagineBreaker

PR #7431, opened 2024-02-19
[GEOS-11266] csw-iso: support multiple paths mapped to queryable

NielsCharlier:
[GEOS-11266](https://osgeo-org.atlassian.net/browse/GEOS-11266) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

  

# Checklist

### DIFF
--- a/doc/en/user/source/extensions/csw-iso/queryables.rst
+++ b/doc/en/user/source/extensions/csw-iso/queryables.rst
@@ -1,0 +1,14 @@
+CSW ISO Metadata Profile Queryables
+===================================
+
+Mapping
+~~~~~~~
+The ISO Metadata standard (see `OGC Implementation Specification 07-045 <http://www.opengeospatial.org/standards/specifications/catalog>`_) specifies a list of so called 'queryables'. These are simple property names such as 'Title' or 'Abstract' that can be used in filters and will automatically be translated to a corresponding XPath.
+
+In some case it might be required to have an alternative mapping of queryables (to other than the default XPaths). For instance when records represent services rather than data, the corresponding XPaths are different.
+
+The ISO Metadata queryables mapping can be found in the file ``csw/MD_Metadata.queryables.properties`` inside the data directory. It follows the format of a properties file, where each value can be a comma-separated list of XPaths. When a queryable is linked to multiple XPaths, filters will include records that have a match for any of them.
+
+Bounding Box
+~~~~~~~~~~~~
+The 'BoundingBox' queryable has an additional functionality. Changing it will not only alter the queryable as such, but also how the layers are mapped to ISO metadata records. The first XPath will be the place where the bounding box of the layers are encoded in each of the records.
--- a/doc/en/user/source/extensions/csw-iso/queryables.rst
+++ b/doc/en/user/source/extensions/csw-iso/queryables.rst
@@ -3,11 +3,11 @@ CSW ISO Metadata Profile Queryables
 
 Mapping
 ~~~~~~~
-The ISO Metadata standard (see `OGC Implementation Specification 07-045 <http://www.opengeospatial.org/standards/specifications/catalog>`_) specifies a list of so called 'queryables'. These are simple property names such as 'Title' or 'Abstract' that can be used in filters and will automatically be translated to a corresponding XPath.
+The ISO Metadata standard (see `OGC Implementation Specification 07-045 <http://www.opengeospatial.org/standards/specifications/catalog>`_) specifies a mapping of XPaths to CSW queryables. These are simple property names such as 'Title' or 'Abstract' that can be used in filters and will automatically be translated to a corresponding XPath.
 
-In some case it might be required to have an alternative mapping of queryables (to other than the default XPaths). For instance when records represent services rather than data, the corresponding XPaths are different.
+In some case it might be required to have an alternative mapping of queryables (to a different, non-default XPaths). For instance when records represent services rather than data, the corresponding XPaths are different.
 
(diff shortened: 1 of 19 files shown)


doc/en/user/source/extensions/csw-iso/queryables.rst
```diff
@@ -0,0 +1,14 @@
+CSW ISO Metadata Profile Queryables
+===================================
+
+Mapping
+~~~~~~~
+The ISO Metadata standard (see `OGC Implementation Specification 07-045 <http://www.opengeospatial.org/standards/specifications/catalog>`_) specifies a list of so called 'queryables'. These are simple property names such as 'Title' or 'Abstract' that can be used in filters and will automatically be translated to a corresponding XPath.
+
+In some case it might be required to have an alternative mapping of queryables (to other than the default XPaths). For instance when records represent services rather than data, the corresponding XPaths are different.
```
aaime, line 8:
```suggestion
In some case it might be required to have an alternative mapping of queryables (to a different, non-default XPath). For instance when records represent services rather than data, the corresponding XPaths are different.
```

doc/en/user/source/extensions/csw-iso/queryables.rst
```diff
@@ -0,0 +1,14 @@
+CSW ISO Metadata Profile Queryables
+===================================
+
+Mapping
+~~~~~~~
+The ISO Metadata standard (see `OGC Implementation Specification 07-045 <http://www.opengeospatial.org/standards/specifications/catalog>`_) specifies a list of so called 'queryables'. These are simple property names such as 'Title' or 'Abstract' that can be used in filters and will automatically be translated to a corresponding XPath.
```
aaime, line 6: I remember things slightly differently, "queriables" are a construct of CSW that advertises a uniform name for the queryable regardless of the underlying language being used (e.g., 'Title' should work with dublin core as well).

This is from the CSW 2.0 core spec:

image

And quoting the text:

> The goal of defining core queryable properties is query interoperability among catalogues
that implement the same protocol binding and query compatibility among catalogues that
implement different protocol bindings perhaps through the use of “bridges” or protocol
adapters. Defining a set of core queryable properties also enables simple cross-profile
discovery, where the same queries can be executed against any catalogue service without
modification and without detailed knowledge of the catalogue's information model

Long story short, a queryable mapping mechanism should be general.

src/extension/csw/api/src/main/java/org/geoserver/csw/store/AbstractCatalogStore.java
```diff
@@ -62,28 +61,35 @@ public CloseableIterator<String> getDomain(Name typeName, final Name attributeNa
...
         // do we have such attribute?
-        final PropertyName property = rd.translateProperty(attributeName);
-        AttributeDescriptor ad = (AttributeDescriptor) property.evaluate(rd.getFeatureType());
-        if (ad == null) {
+        final List<PropertyName> properties = rd.translateProperty(attributeName);
+        boolean hasProperty = false;
+        for (PropertyName property : properties) {
+            AttributeDescriptor ad = (AttributeDescriptor) property.evaluate(rd.getFeatureType());
+            hasProperty |= ad != null;
+        }
```
aaime, line 69: This could be a nice self-contained method "hasProperty"

src/extension/csw/api/src/main/java/org/geoserver/csw/store/AbstractCatalogStore.java
```diff
@@ -62,28 +61,35 @@ public CloseableIterator<String> getDomain(Name typeName, final Name attributeNa
...
-                                                UTF_8));
+                            for (PropertyName property : properties) {
+                                Property prop = (Property) property.evaluate(feature);
+                                if (prop != null) {
+                                    values.add(
+                                            new String(
+                                                    ((String) prop.getValue()).getBytes(ISO_8859_1),
+                                                    UTF_8));
+                                    break;
+                                }
```
aaime, line 92: Wondering what happens if multiple xpaths for a given property happen to exist at the same time?
The documentation should probably say something about this case.

src/extension/csw/csw-iso/src/main/java/org/geoserver/csw/records/iso/QueryableMappingFilterVisitor.java
```diff
@@ -0,0 +1,792 @@
...
+import org.geotools.api.filter.temporal.MetBy;
+import org.geotools.api.filter.temporal.OverlappedBy;
+import org.geotools.api.filter.temporal.TContains;
+import org.geotools.api.filter.temporal.TEquals;
+import org.geotools.api.filter.temporal.TOverlaps;
+import org.geotools.data.complex.util.XPathUtil;
+import org.geotools.factory.CommonFactoryFinder;
+import org.geotools.util.factory.GeoTools;
+
+public class QueryableMappingFilterVisitor implements FilterVisitor, ExpressionVisitor {
```
aaime, line 76: Wow, big new class. Some javadoc explaining what it does could help.

doc/en/user/source/extensions/csw-iso/queryables.rst
NielsCharlier: You are right, I didn't even know. It is a separate issue though, so I created one on jira: https://osgeo-org.atlassian.net/browse/GEOS-11323

NielsCharlier: I have changed the docs a bit too

NielsCharlier:
Thanks for the review @aaime. The requested changes were made

aaime:
Merged on main. Does it need to be backported on 2.25.x and 2.24.x?

NielsCharlier:
Thanks, I added backports labels

PR #7449, opened 2024-03-01
Mention need for config edit for root/masterpw login

kannes:
I tried to log in to a misconfigured GeoServer as `root` with the masterpw and was horrified when it did not work. Only via Google I found out that one needs to explicitely enable this special login: https://docs.geoserver.org/2.24.x/en/user/security/passwd.html#keystore-password

This PR simply adds another line to the `masterpw.info` file so that there is a better chance for discovery. :)

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/GeoServerSecurityManager.java
+++ b/src/main/src/main/java/org/geoserver/security/GeoServerSecurityManager.java
@@ -1930,7 +1930,7 @@ public class GeoServerSecurityManager implements ApplicationContextAware, Applic
             w.write("Test the master password by logging in as user \"root\"");
             w.newLine();
             w.newLine();
-            w.write("Note: Login with the master password must be enabled in security/masterpw/default/config.xml");
+            w.write("Note: To test 'Login with the master password' setting must be enabled for Master Password Provider.");
             w.newLine();
             w.newLine();
             w.write("This file should be removed after reading !!!.");
--- a/src/main/src/main/java/org/geoserver/security/GeoServerSecurityManager.java
+++ b/src/main/src/main/java/org/geoserver/security/GeoServerSecurityManager.java
@@ -1930,6 +1930,9 @@ public class GeoServerSecurityManager implements ApplicationContextAware, Applic
             w.write("Test the master password by logging in as user \"root\"");
             w.newLine();
             w.newLine();
+            w.write("Note: Login with the master password must be enabled in security/masterpw/default/config.xml");
+            w.newLine();
+            w.newLine();
             w.write("This file should be removed after reading !!!.");
             w.newLine();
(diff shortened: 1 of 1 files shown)


src/main/src/main/java/org/geoserver/security/GeoServerSecurityManager.java
```diff
@@ -1930,6 +1930,9 @@ void writeMasterPasswordInfo(Resource file, String message, char[] masterPasswor
             w.write("Test the master password by logging in as user \"root\"");
             w.newLine();
             w.newLine();
+            w.write("Note: Login with the master password must be enabled in security/masterpw/default/config.xml");
```
jodygarnett, line 1933: There is a way to enable this using the user interface, hand editing config.xml is not necessary or recommended.

You could link to the appropriate UI setting...

jodygarnett, line 1933:
```suggestion
            w.write("Note: To test 'Login with the master password' setting must be enabled for Master Password Provider.");
```

jodygarnett (changes requested):
The user interface setting is **Allow "root" user to login as Admin** when editing the default Master Password Provider.

jodygarnett (approved):

src/main/src/main/java/org/geoserver/security/GeoServerSecurityManager.java
kannes: The config.xml method is mentioned on https://docs.geoserver.org/main/en/user/security/webadmin/passwords.html#keystore-password-providers and https://docs.geoserver.org/main/en/user/security/passwd.html#keystore-password . In my case it was the only way to get access to the admin interface as there was no admin user available (so I could not have set anything in the GUI) and I imagine that others very likely might end up in such a situation as well.

Right now a person in this situation would have zero luck finding the config.xml option with a search engine.

aaime:
This PR was merged with failing QA, now the main branch is tainted and all PRs will fail QA as well.
The code formatting check failed, the code was not built with maven (see the QA documentation, linked from the checklist, which was not even compiled).

@jodygarnett there is no need to merge every single outstanding PR when doing a release, especially not one that fails some of the github actions check.

I'll close the backport and manually fix main....

aaime:
Btw, is there a Jira ticket for this? Or think change won't show up in release notes/announcements.

kannes:
> Btw, is there a Jira ticket for this? Or think change won't show up in release notes/announcements.

Nope, all I got from atlassian.net was a loop of Sign In -> Spinning Icon -> Sign In. I'm sorry but I cannot invest time with troubleshooting yet another account on yet another proprietary website. :\

PR #7458, opened 2024-03-06
[GEOS-11325] Add properties to set additional security headers

sikeoka:
[GEOS-11325](https://osgeo-org.atlassian.net/browse/GEOS-11325) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This PR adds system properties to allow GeoServer administrators to configure the X-XSS-Protection and Strict-Transport-Security HTTP response headers similar to how the X-Content-Type-Options and X-Frame-Options headers currently work.  All four of these headers are normally enabled by default by Spring Security but I left the two that are being added here disabled by default so there's no backwards compatibility issues.  I could enable them by default if that is the desired behavior.

Also, while it is probably unnecessary, to address previous concerns about calling GeoServerExtensions.getProperty so many times on every request, the property values are now cached.  I am working on something else that may add a couple more properties here.

  

# Checklist

### DIFF
--- a/doc/en/user/source/production/config.rst
+++ b/doc/en/user/source/production/config.rst
@@ -159,8 +159,17 @@ kinds of security vulnerabilities. See the `OWASP Clickjacking entry <https://ww
 
 If you wish to change this behavior you can do so through the following properties:
 
-* ``geoserver.xframe.shouldSetPolicy``: controls whether the X-Frame-Options filter should be set at all. Default is true.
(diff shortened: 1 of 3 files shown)


aaime:
"I am working on something else that may add a couple more properties here." --> will it be added here or is it going to be a separate PR?

sikeoka:
Sorry I should have been clearer about that.  It's the Content Security Policy work that I sent a preview of to geoserver-security which will be a separate PR.

aaime:
Looks good, thanks!

PR #7459, opened 2024-03-07
[GEOS-11327] Add warning about using embedded data directories

sikeoka:
[GEOS-11327](https://osgeo-org.atlassian.net/browse/GEOS-11327) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This PR adds a warning message when administrators log in to the GeoServer home page if GeoServer is using a data directory embedded in the web application directory.  It also adds a vague clarification to the documentation that there are security reasons for using an external data directory.

  

# Checklist

### DIFF
--- a/src/web/security/core/src/main/java/org/geoserver/security/web/SecurityHomePageContentProvider.java
+++ b/src/web/security/core/src/main/java/org/geoserver/security/web/SecurityHomePageContentProvider.java
@@ -8,9 +8,11 @@ package org.geoserver.security.web;
 import static org.geoserver.security.impl.GeoServerUser.ADMIN_USERNAME;
 import static org.geoserver.security.impl.GeoServerUser.DEFAULT_ADMIN_PASSWD;
 
+import java.io.File;
 import java.io.IOException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import org.apache.commons.io.FilenameUtils;
 import org.apache.wicket.Component;
 import org.apache.wicket.Page;
 import org.apache.wicket.behavior.AttributeAppender;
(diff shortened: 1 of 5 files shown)


mprins (approved):
LGTM

src/web/security/core/src/main/java/org/geoserver/security/web/SecurityHomePageContentProvider.java
```diff
@@ -46,6 +48,25 @@ public Component getPageBodyComponent(String id) {
...
     }
 
+    /** Tests if the data directory is embedded in the GeoServer web archive. */
+    public static boolean isEmbeddedDataDirectory(GeoServerApplication app) {
+        try {
+            String webRootPath = app.getServletContext().getRealPath("/");
+            if (webRootPath == null) {
+                return false; // this should never happen
+            }
+            File dataDir = app.getResourceLoader().getBaseDirectory().getAbsoluteFile();
```
aaime, line 58: Thinking out loud here... what if the resource store is externalized in a database?
A data directory should be there anyways, at least the JDBC configuration files are in it, right? So there is a risk of losing configuration on upgrade if not externalized. Is it still a security risk though?

sikeoka: I don't use the jdbcstore module but I did some investigation and even when using that module, a data directory still has to exist somewhere on the file system to tell geoserver how to connect to the jdbcstore database so both risks you mentioned would still exist.

PR #7472, opened 2024-03-12
[GEOS-11322] MapML WMS Vector Representation include query filter

turingtestfail:
[GEOS-11322](https://osgeo-org.atlassian.net/browse/GEOS-11322) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

wildcard import removed

abstract parent needs the mime

file license

cleanup

fixed test issues with projection

validating sld filters

added getStyleQuery test

Moved to util class

cleanup

test fix

removed unnecessary inheritance

cleanup

pr response

PR responses and fix to resource vs layer metadata issue

html version now adds feature kv

html source link should return mapml not features

added cql-filter test to sld filter test

switched to allowing mixed raster and vector based on layer settings

added check for mixed vector, raster in different extents

cleanup

added change to all layers vector documentation

  

# Checklist


aaime:
Overall it looks good, by interactive testing I believe the request filters are still getting ignored.

For example, if I set ``topp:states`` to generate vectors, and add ``&cql_filter=PERSONS>5000000`` to the request, I still get all the states showing in the vector map.

image

And oh, it seems the same is happening when using images as well:

image

But with an openlayer preview instead:

image

Generally speaking the image links should try to replicate all parameters found in the original GetMap. In other words, just fixing cql_filter will not be enough, as there are other parameters, e.g. ``env``, ``viewparams``, and some parameters could be data source specific as well.

prushforth (approved):
I used the cql_filter=5000000 successfully with youre recent commit.  I was also able to get an sld to perform the same filtering. So that's pretty neat!

LGTM

aaime (approved):
Thanks, LGTM

PR #7474, opened 2024-03-14
[Community Plugin] Features-Autopopulate

afabiani:
Features-Autopopulate Extension
===============================

The Features Autopopulate plug-in listens to transactions (so far only issued by WFS), and autopopulates the feature type attributes according to the values retrieved from the properties file.

The plugin uses a custom TransactionCallback that alters the insert/update WFS-T operations, forcing in specific values into them, based on configuration files.

To support configuration for multiple layers, the easiest thing is to place a configuration, file in the directories of the layers themselves, pretty much like the featureinfo templates.

A "transactionCustomizer.properties" file that contains a set of names and CQL expressions
 e.g.:

 ```
 UTENTE=env('GSUSER') # this will be replaced with the current user see @EnviromentInjectionCallback
 AGGIORNAMENTO=now()  # this will be replaced with the current date
 ```

To keep things simple, the expressions will just use environment variables, but not see the other values provided in the update/insert, and will not be differentiated by insert/update cases.


doc/en/user/source/community/features-autopopulate/index.rst
```diff
@@ -0,0 +1,20 @@
+.. _community_wfsautopopulate:
+
+Features-Autopopulate Extension
```
aaime, line 3: Thank you for the docs, they are important for users to try out the module!

src/community/features-autopopulate/features-autopopulate-core/src/main/java/org/geoserver/autopopulate/AutopopulateTemplate.java
```diff
@@ -0,0 +1,118 @@
...
+ * AutopopulateTemplate class is used to load the properties from the file and store them in a map
+ * for further use.
+ *
+ * <p>It also provides the methods to get the properties and set the properties.
+ *
+ * @author Alessio Fabiani, GeoSolutions SRL, alessio.fabiani@geosolutionsgroup.com
+ */
+public class AutopopulateTemplate {
+    /** logger */
+    private static final Logger log = Logging.getLogger(AutopopulateTransactionCallback.class);
```
aaime, line 29: By common naming convention (usage points need to be updated as well):

```suggestion
    private static final Logger LOGGER = Logging.getLogger(AutopopulateTransactionCallback.class);
```

src/community/features-autopopulate/features-autopopulate-core/src/main/java/org/geoserver/autopopulate/AutopopulateTemplateLoader.java
```diff
@@ -0,0 +1,127 @@
...
+                // try global supplementary files
+                template = Resources.file(dd.getWorkspaces(path));
+            }
+
+            if (template != null) {
+                return new AutopopulateTemplate(template.getAbsolutePath());
+            }
+        }
+
+        if (workspace != null) {
```
aaime, line 111: You already have a resource, which is placed in a workspace. This seems to allow loading from a workspace different than the resource one... what's the use case for this?

src/community/features-autopopulate/features-autopopulate-core/src/test/java/org/geoserver/autopopulate/AutopopulateTransactionCallbackTest.java
```diff
@@ -0,0 +1,151 @@
+/* (c) 2024 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.autopopulate;
+
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.*;
```
aaime, line 8: No star imports

src/community/features-autopopulate/features-autopopulate-core/src/main/java/org/geoserver/autopopulate/AutopopulateTransactionCallback.java
```diff
@@ -0,0 +1,439 @@
...
+                        log.info("Updating feature: " + feature);
+                        SimpleFeature transformed = dataStoreChangeInternal(feature);
+                        List<Property> properties = ((Update) element).getUpdateProperties();
+                        for (org.geotools.api.feature.Property p : transformed.getProperties()) {
+                            if (properties.stream()
+                                    .anyMatch(
+                                            prop ->
+                                                    prop.getName()
+                                                            .getLocalPart()
+                                                            .equals(p.getName().getLocalPart()))) {
```
aaime, line 191: This bit is used twice and is fighing with the max number of columns in the formatter. Maybe have a simple support method, ``match(p1, p2)`` method that compares the local names instead?

src/community/features-autopopulate/features-autopopulate-core/src/main/java/org/geoserver/autopopulate/AutopopulateTransactionCallback.java
```diff
@@ -0,0 +1,439 @@
...
+                ((Insert) element).setFeatures(newFeatures);
+                newElements.add(element);
+            } else if (element instanceof Update) {
+                FeatureTypeInfo featureTypeInfo =
+                        getFeatureTypeInfo(new NameImpl(element.getTypeName()));
+                try {
+                    SimpleFeatureCollection features = getTransactionFeatures((Update) element);
+                    SimpleFeatureIterator featuresIterator = features.features();
+                    while (featuresIterator.hasNext()) {
+                        SimpleFeature feature = featuresIterator.next();
```
aaime, line 181: Since you're not allowing feature attribute usage anyways (by documentation) why not sure check if the property is part of the update, if it's there, override, otherwise create it, done.

src/community/features-autopopulate/features-autopopulate-core/src/main/java/org/geoserver/autopopulate/AutopopulateTransactionCallback.java
```diff
@@ -0,0 +1,439 @@
...
+                        }
+                    }
+                }
+                ((Insert) element).setFeatures(newFeatures);
+                newElements.add(element);
+            } else if (element instanceof Update) {
+                FeatureTypeInfo featureTypeInfo =
+                        getFeatureTypeInfo(new NameImpl(element.getTypeName()));
+                try {
+                    SimpleFeatureCollection features = getTransactionFeatures((Update) element);
```
aaime, line 178: There are several casts of "element" to Update here, cast it just once in a local property instead.

src/community/features-autopopulate/features-autopopulate-core/src/main/java/org/geoserver/autopopulate/AutopopulateTransactionCallback.java
```diff
@@ -0,0 +1,439 @@
...
+     *
+     * @param update
+     * @return
+     */
+    private SimpleFeatureCollection getTransactionFeatures(Update update) throws IOException {
+        QName typeName = update.getTypeName();
+        Filter filter = update.getFilter();
+        SimpleFeatureSource source = getTransactionSource(update);
+        try {
+            List<SimpleFeature> recent = DataUtilities.list(source.getFeatures(filter));
```
aaime, line 279: Ouch no, the list of features affected by an Update might be... all of them. This is inviting a OOM to happen.
Also see my comment above about avoiding scanning the updated features at all.

aaime (changes requested):
Overall looks good, there is one bit that worries me in terms of Out of Memory. The rest is minor comments.

src/community/features-autopopulate/features-autopopulate-core/src/main/java/org/geoserver/autopopulate/AutopopulateTransactionCallback.java
```diff
@@ -0,0 +1,439 @@
...
+        // if not found, load the template
+        GeoServerResourceLoader resourceLoader =
+                GeoServerExtensions.bean(GeoServerResourceLoader.class);
+        Catalog catalog = (Catalog) GeoServerExtensions.bean("catalog");
+        FeatureTypeInfo feature = catalog.getFeatureTypeByName(featureType.getTypeName());
+        AutopopulateTemplateLoader templateLoader =
+                new AutopopulateTemplateLoader(
+                        resourceLoader, feature.getStore().getWorkspace(), feature);
+        t = templateLoader.loadTemplate(path);
+        templateCache.put(key, t);
```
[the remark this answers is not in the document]
afabiani: Do we have some utilities on GeoServer code base already to listen for file changes?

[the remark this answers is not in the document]
aaime: Yes, that would be PropertyFileWatcher. This allows to catch changes in source files as they happen, without even waiting for a reload. You can have a look at a usage example here:
https://github.com/geoserver/geoserver/blob/2a9358be33a65b22afe734749df67383bd0837cb/src/extension/control-flow/src/main/java/org/geoserver/flow/config/DefaultControlFlowConfigurator.java

It also means you cannot trust just the cache, but you'll have to pair each template with its own file watcher and see if there has been any changes in the meantime.

To make things simpler, you can just implement reload listening in this class instead, by implementing GeoServerLifecycleHandler, and drop the cache on reload/reset.

src/community/features-autopopulate/features-autopopulate-core/src/main/java/org/geoserver/autopopulate/AutopopulateTemplateLoader.java
afabiani: The original intent was to look for a template valid at workspace level. I modev the check to look inside the resource workspace.

src/community/features-autopopulate/features-autopopulate-core/src/test/java/org/geoserver/autopopulate/AutopopulateTransactionCallbackTest.java
afabiani: Done

src/community/features-autopopulate/features-autopopulate-core/src/main/java/org/geoserver/autopopulate/AutopopulateTransactionCallback.java
afabiani: Done

src/community/features-autopopulate/features-autopopulate-core/src/main/java/org/geoserver/autopopulate/AutopopulateTransactionCallback.java
afabiani: More or less done... I just load the first feature available in order to reuse the template lokkup code. But now I avoid looping through the features and, most important, I don't load them in memory anymore.
Thanks for the hint!

src/community/features-autopopulate/features-autopopulate-core/src/main/java/org/geoserver/autopopulate/AutopopulateTransactionCallback.java
afabiani: Done

src/community/features-autopopulate/features-autopopulate-core/src/main/java/org/geoserver/autopopulate/AutopopulateTransactionCallback.java
afabiani: Removed.

src/community/features-autopopulate/features-autopopulate-core/src/main/java/org/geoserver/autopopulate/AutopopulateTransactionCallback.java
```diff
@@ -0,0 +1,439 @@
...
+     * Check/alter feature collections and filters before a change hits the datastore.
+     *
+     * <p><b>Note</b> that caution should be exercised when relying on feature identifiers from a
+     * {@link TransactionEventType#POST_INSERT} event. Depending on the type of store, those
+     * identifiers may be reliable. Essentially, they can only be relied upon in the case of a
+     * spatial dbms (such as PostGIS) is being used.
+     *
+     * @return
+     */
+    private SimpleFeature dataStoreChangeInternal(SimpleFeature source) throws IOException {
```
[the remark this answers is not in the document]
afabiani: Done

src/community/features-autopopulate/features-autopopulate-core/src/main/java/org/geoserver/autopopulate/AutopopulateTransactionCallback.java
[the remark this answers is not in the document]
afabiani: Done thanks

src/community/features-autopopulate/features-autopopulate-core/src/main/java/org/geoserver/autopopulate/AutopopulateTransactionCallback.java
```diff
@@ -0,0 +1,439 @@
...
+            this.template = template;
+        }
+
+        @Override
+        public int hashCode() {
+            final int PRIME = 31;
+            int result = 1;
+            result = PRIME * result + ((template == null) ? 0 : template.hashCode());
+            result = PRIME * result + ((type == null) ? 0 : type.hashCode());
+            return result;
```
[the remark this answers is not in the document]
afabiani: Done

src/community/features-autopopulate/features-autopopulate-core/src/main/java/org/geoserver/autopopulate/AutopopulateTransactionCallback.java
```diff
@@ -0,0 +1,439 @@
...
+        }
+
+        @Override
+        public boolean equals(Object obj) {
+            if (this == obj) return true;
+            if (obj == null) return false;
+            if (getClass() != obj.getClass()) return false;
+            final TemplateKey other = (TemplateKey) obj;
+            if (template == null) {
+                if (other.template != null) return false;
```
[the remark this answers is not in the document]
afabiani: Done

afabiani:
@aaime changes done, thanks for the review.

src/community/features-autopopulate/features-autopopulate-core/src/main/java/org/geoserver/autopopulate/AutopopulateTransactionCallback.java
```diff
@@ -0,0 +1,407 @@
...
+                Insert insertElement = (Insert) element;
+                insertElement.setFeatures(newFeatures);
+                newElements.add(element);
+            } else if (element instanceof Update) {
+                FeatureTypeInfo featureTypeInfo =
+                        getFeatureTypeInfo(new NameImpl(element.getTypeName()));
+                try {
+                    Update updateElement = (Update) element;
+                    SimpleFeature feature =
+                            getTransactionSource(updateElement).getFeatures().features().next();
```
aaime, line 189: This will leak a feature iterator, eventually starving connection pools. Use ``DataUtilities.first(getTransactionSource(updateElement).getFeatures())`` instead.

afabiani: Yessir, ty. Done!

aaime (approved):

PR #7501, opened 2024-03-19
[GEOS-11337] Support feature tiles in MapML

aaime:
[GEOS-11337](https://osgeo-org.atlassian.net/browse/GEOS-11337) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

See ticket for details

# Checklist


prushforth:
The vector tiles work perfectly with the states shapefile population style. I created a style for a Canada layer I have and the tiles _sometimes_ show through:

image

It's harder to test / debug than just regular features because the actual \<map-feature> is rendered and discarded.

I'll post more info when I have it.

prushforth:
I believe it's because the problem geometries are being rendered exclusively as linstrings, whereas they work fine when they are  rendered as polygons.  For example, in [this request](http://localhost:8080/geoserver/nrcan/gwc/service/wmts?format_options=mapmlfeatures:true;mapmlskipattributes:true&request=GetTile&tilematrixset=OSMTILE&tilematrix=6&TileRow=15&service=WMTS&format=text/mapml&style=&TileCol=10&version=1.0.0&layer=nrcan:provinces) (if you load up that geoserver_data_dir) you'll see both cases.

prushforth:
This image might be a good replacement for the existing images/mapml_tile_caching_panel.png tbd:
image

aaime:
@prushforth should be fixed

prushforth:
Was about to report that I couldn't see a difference, then remembered to re-create the GWC cache.

With that it works very very well, thanks!

The only thing I think we should do, maybe not here, is to move the CSS out of each tile.  IIRC there was going to be a `format_options=mapmltile:true` (maybe this is similar in purpose to the existing `mapmlskipattributes:true`) to allow GWC to tell the WMS that it wanted the response conditioned to be a tile i.e. with the \<map-span>s and so on.  Perhaps when the source  `<mapml->` containing the `<map-extent>...<map-link rel="tile" type="text/mapml" ...>` is created, the system could embed the relevant CSS in the `<map-head>`, either as a `<map-link rel="stylesheet" href="geoserver/styles??">` or if that's not possible or simple, as an inline `<map-style>` containing the generated CSS.  This would reduce server and client processing, bandwidth and storage required to inline the CSS for each tile.

aaime:
@prushforth I'm trying to figure out how to do that. Here is some context to inform the discussion:

* A single GetMap request can contain multiple layers, if we are talking about features, each one has a separate link (single links with multiple layers generate raster tiles instead)
* The CSS class names are unique only within the same style,  they follow naming conventions such as "ruleX_symbolizerY", but a multi-layer request has multiple styles, so it's guaranteed there will be overlap

If we go and put the styles inline, in the map head, with multiple layers, then the class names will have to be prefixed with the style name.
If we go and add the stylesheet link in each extent, we need to prepare a new output format, that I presume generates a CSS output. It could be a "text/css+mapml" or something like that.

Would the existing map client able to support both implementations?

prushforth:
The client can support this:

```html
<mapml- xmlns="...">
<map-head>
<map-link rel="stylesheet" href="..."></map-link>
...
</map-head>
...
</mapml->
```
or this:

```html
<mapml- xmlns="...">
<map-head>
<map-style>CSS RULES GO HERE</map-style>
</map-head>
...
</mapml->
```
or a combination of both.

> It could be a "text/css+mapml" or something like that.

No need for `+mapml` in this case, it's just CSS afaik.

Regarding multi-layer requests, I don't believe that we currently support \<map-link rel=stylesheet> or \<map-style> within the \<map-extent>, but that would be a good addition.  I will investigate and report back.

aaime:
The CSS generated by the MapML extension is specific to it (another client will likely want a CSS that can be used to style client-side based on the attribute values, for example), so we need to qualify it somehow. 

Better to just embed the styles in the head, adding the style name prefixes, then.

prushforth:
That's a good solution. Instead of a mime type, one could use a parameter to get the MapML format specific CSS

prushforth:
> > Instead of a mime type, one could use a parameter to get the MapML format specific CSS
> 
> Nope, that's not how output format lookup works. It's based solely on the contents of the "format" parameter.

ok so it is a parameter, in that case it would be ok to use it.  I was just concerned that I would be trying to mint a new vocabulary of CSS just for mapml, called `text/css+mapml` which is not the intent.

aaime:
Found an issue in the prefixing approach... a valid GetMap request can use dynamic styling with the SLD and SLD_BODY parameters. Those styles are provided in the request and won't have an identifier recognizable to GeoServer.
At the same time, MapML right now has no support for SLD or SLD_BODY to start with (the link builders don't know what these parameters are).

Possible approaches:
- Ignore them, the rest of the MapML subsystem does not know what to do with them anyways. Just support local styles, that have a full prefixed name known to GeoServer.
- Keep the door open to support them in the future. However, this requires to have a separate stylesheet link per map extent, so that we don't have to qualify the CSS class names by style prefix (an extent with a feature link uses a single style anyways).

prushforth:
> Just support local styles, that have a full prefixed name known to GeoServer.

Sounds good

> Keep the door open to support them in the future. However, this requires to have a separate stylesheet link per map extent, so that we don't have to qualify the CSS class names by style prefix (an extent with a feature link uses a single style anyways)

Also good.  I will log the issue in our client repo.

aaime:
@prushforth styles have been centralized in the top-level mapml head section. Seems to be working fine and saves some size from the tiles.

prushforth:
> @prushforth styles have been centralized in the top-level mapml head section. Seems to be working fine and saves some size from the tiles.

I will check it out a bit later today

turingtestfail (approved):

src/extension/mapml/src/main/java/org/geoserver/mapml/MapMLConstants.java
```diff
@@ -31,8 +31,13 @@ public final class MapMLConstants {
...
 
-    /** MapML format options */
-    public static final String MAPML_FEATURE_FORMAT_OPTIONS = "mapmlfeatures:true";
+    /** MapML format option enabling features */
+    public static final String MAPML_FEATURE_FO = "mapmlfeatures";
+
+    /** MapML format option enabling attribute skipping */
+    public static final String MAPML_SKIP_ATTRIBUTES_FO = "mapmlskipattributes";
+
+    public static final String MAPML_SKIP_STYLES_FO = "mapmlskiphstyles";
```
prushforth, line 40: `mapmlskiphstyles` possibly a typo? -> `mapmlskipstyles`

prushforth (approved):
This is magnificent work.  I have started [adding scoping of \<map-style> and \<map-link rel=stylesheet> to the client](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/945), but it won't land in time for this work.  

Future work could include that, plus an idea I had to improve performance would be to transform the coordinates to integer `tile` coordinates (0-255) on the server when caching.  Something I noticed is that by not including the styles in the tile, we perhaps have the opportunity to apply different styles to the same cache without changing it, via linked stylesheets and so on.

Thanks again for this.

src/extension/mapml/src/main/java/org/geoserver/mapml/MapMLConstants.java
aaime: Not a typo, skipstyles would imply, in my mind, just wireframe, without any style. What we don't want is style definitions in the "head", while keeping the styles references in the features. 
"skipheadstyles" seemed too long though but quite a bit clearer. 
Open to suggestions.

prushforth: I have been opting for longer but obvious variable names recently as I'm getting older and seem to always forget what I was up to when I originally wrote some piece of code in the past

aaime: Switched to the longer name

PR #7154, opened 2023-10-04
[GEOS-11275] Wicket 9 upgrade

bradh:
[GEOS-11275](https://osgeo-org.atlassian.net/browse/GEOS-11275) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

Work in progress on Wicket 8 migration (https://github.com/geoserver/geoserver/wiki/Jakarta-EE#wicket)

Here is content to fix found during testing:

* [Wicket Test Plan](https://docs.google.com/spreadsheets/d/1pQmncG4zxpgJnHxeI4myFfOBD17U2CIMcy59II4XAfo/edit?usp=sharing) (Google Doc)

* Look for "TODO WICKET8" for remaining work. Mostly this is verification, which should occur at the end. However there are broken modules and broken tests that need to be debugged.

# Checklist

- [ ] I have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/main/CONTRIBUTING.md).
- [ ] I have sent a [Contribution Licence Agreement](https://docs.geoserver.org/latest/en/developer/policies/committing.html) (not required for small changes, e.g., fixing typos in documentation).
- [ ] First PR targets the `main` branch (backports managed later; ignore for branch specific issues).
- [ ] All the build checks are green ([see automated QA checks](https://docs.geoserver.org/latest/en/developer/qa-guide/index.html)).

For core and extension modules:

- [ ] New unit tests have been added covering the changes.
- [ ] [Documentation](https://github.com/geoserver/geoserver/tree/main/doc/en/user/source) has been updated (if change is visible to end users).
- [ ] The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers).
- [ ] There is an issue in the [GeoServer Jira](https://osgeo-org.atlassian.net/browse/GEOS/summary) (except for changes that do not affect administrators or end users in any way).
- [ ] Commit message(s) must be in the form ``[GEOS-XYZWV] Title of the Jira ticket``.
- [ ] Bug fixes and small new features are presented as a single commit.
- [ ] Each commit has a single objective (if there are multiple commits, each has a separate JIRA ticket describing its goal).

PR #7514, opened 2024-03-26
[GEOS-11346] Add a configurable Content-Security-Policy header

sikeoka:
[GEOS-11346](https://osgeo-org.atlassian.net/browse/GEOS-11346) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This PR adds the Content-Security-Policy header to GeoServer responses based on the request path and query and server properties, a default configuration that should support existing GeoServer functionality, a web interface for administrators to update the configuration and documentation updates. This PR is **NOT** intended to be backported.

The primary benefit of this work is to provide mitigations for security issues with WMS GetFeatureInfo HTML templates (also applies to the new WFS GetFeature HTML output), static HTML files and the OGC API community modules and to encourage developers contributing new HTML functionality to follow more secure coding practices.

See https://github.com/geoserver/geoserver/wiki/GSIP-227
  

# Checklist


src/main/src/main/java/org/geoserver/filters/XFrameOptionsFilter.java
```diff
@@ -50,7 +52,16 @@
...
+ * Two properties control the Content-Security-Policy header to prevent cross-site scripting
+ * attacks:<br>
+ * - geoserver.csp.shouldSetPolicy: controls whether to set the Content-Security-Policy header.
+ * Default is true.<br>
+ * - geoserver.csp.policy: controls the value of the Content-Security-Policy header. Default is to
+ * use the configuration file that will set the header based on the request and server environment
+ * but this property allows the administrator to set a fixed header value, which can be useful in
+ * cases where JavaScript is not needed at all.
  */
 public class XFrameOptionsFilter implements Filter {
```
aaime, line 66: This class started as a way to add XFrameOptions, but over the last few months you pushed more and more headers into it. Can you think of a more appropriate name, given its current function?

doc/en/user/source/security/csp.rst
```diff
@@ -0,0 +1,295 @@
+.. _security_csp:
+
+Content Security Policy
+=======================
+
+The :guilabel:`Content Security Policy` page controls the checks that are performed on GeoServer
+requests that GeoServer will use to set an appropriate Content-Security-Policy (CSP) HTTP
+response header to mitigate cross-site scripting and clickjacking attacks.
```
aaime, line 8:
```suggestion
The :guilabel:`Content Security Policy` page controls how GeoServer prepares the Content-Security-Policy (CSP) HTTP response headers, used to mitigate cross-site scripting and clickjacking attacks.
```

doc/en/user/source/security/csp.rst
```diff
@@ -0,0 +1,295 @@
...
+
+.. note::
+    The settings on this page do not affect anything if the CSP header has been disabled using the
```
aaime, line 11:
```suggestion
    The settings on this page do not affect anything if the CSP headers have been disabled using the
```

doc/en/user/source/security/csp.rst
```diff
@@ -0,0 +1,295 @@
...
+    ``geoserver.csp.shouldSetPolicy`` property or if the ``geoserver.csp.policy`` property is used
+    to set a fixed header value.
+
+.. note::
+    Starting with version 9, Wicket will set its own CSP header with a very strict policy that will
+    overwrite what is set by GeoServer although there is a workaround in place to merge directives
```
aaime, line 17:
```suggestion
    overwrite what is set by GeoServer, although there is a workaround in place to merge directives
```

aaime, line 17: This is a useful note for me (I was about to ask indeed), but not for an end user, that does not know which version of Wicket we are using. Until the Wicket update branch is merged, we should inform the user about the behavior, based on the current Wicket version.

Suggestion: update this comment, and once it's merged, propose a documentation change in the Wicket upgrade branch.

doc/en/user/source/security/csp.rst
```diff
@@ -0,0 +1,295 @@
...
+The default CSP configuration is intended to support many GeoServer use cases and allow users to
+securely run GeoServer without having to modify the configuration. It may be updated in future
+releases to fix bugs, support new features or enhance security.
+
+The default header value for most GeoServer requests will be (without line breaks):
+
+::
+    base-uri 'self'; form-action 'self'; default-src 'none'; child-src 'self'; connect-src 'self';
+    font-src 'self'; img-src 'self' data:; style-src 'self' 'unsafe-inline'; script-src 'self';,
+    frame-ancestors 'self';
```
aaime, line 34: Not sure what you wanted to get here, but the rendering is weird (happens in the other examples in the pages as well):

image

Also, is this comma intended? ``script-src 'self';,``

doc/en/user/source/security/csp.rst
```diff
@@ -0,0 +1,295 @@
...
+commas rather than setting multiple Content-Security-Policy headers. If a value for a specific
+directive is defined in multiple policies, web browsers will use the strictest value set for that
+directive.
+
+The button for adding policies can be found at the top of the :guilabel:`Policy List` table and a
+policy can be edited by clicking on its name in the table or removed by clicking on the remove icon
+at the end of the policy's row in the table. Policy positions can also be changed by using the
+up/down arrows or by dragging and dropping the rule's row in the table.
+
+.. figure:: images/csp-policy.png
```
aaime, line 125: The screenshot is scaled down and becomes unreadable.

aaime (commented):
While I'm providing some comments on this PR, it should not be considered a review. I just don't have enough knowledge to understand what is going on it this PR. 
Even compared to other security parts, this seems very complex, and I would expect most end users to leave it as-is. 
If anything, I'd urge to simplify and clarify the system as much as possible, a tool for experts is normally dangerous in the hands of uneducated users.

aaime:
Some more notes on the configuration:

* The configuration refers directly to GeoServer class names, something an end user will have no knowledge about, and introduces refactoring issues: if any class name of package changes, the default CSP policy will have to be updated (but developers responsible for that change are unlikely to know about it)
* The configuration uses a simple predicate language. The parser seems to be susceptible to extra spaces (might be wrong here). I guess the reason for the language is to avoid having to build a GUI for every test type? A GUI for each type backed by objects that then serialize to XML would have been more natural in GeoServer.

jodygarnett:
This is getting ... very complicated! We need some assurance for users that the defaults are sensible.

src/main/src/main/java/org/geoserver/filters/XFrameOptionsFilter.java
sikeoka: I considered renaming it to SecurityHeadersFilter but that would be a pretty significant change and would complicate backporting if the renaming wasn't backported.

doc/en/user/source/security/csp.rst
sikeoka: Other than referencing specific Wicket versions, this comment would still be useful to users after the Wicket update. It lets them know that disabling GeoServer's CSP will not disable the CSP on Wicket pages and also explains that the actual CSP header for GeoServer's Wicket pages will be a combination of Wicket's and GeoServer's CSP.

doc/en/user/source/security/csp.rst
sikeoka: The rendering is fixed now.

sikeoka:
> Some more notes on the configuration:
> 
> * The configuration refers directly to GeoServer class names, something an end user will have no knowledge about, and introduces refactoring issues: if any class name of package changes, the default CSP policy will have to be updated (but developers responsible for that change are unlikely to know about it)

The class name filtering isn't very important to this feature so it can be removed to simplify this a little.

> * The configuration uses a simple predicate language. The parser seems to be susceptible to extra spaces (might be wrong here). I guess the reason for the language is to avoid having to build a GUI for every test type? A GUI for each type backed by objects that then serialize to XML would have been more natural in GeoServer.

Making an even more complicated GUI for this seems unnecessary especially since most of the work is in writing regular expressions which would still have to be done manually by an administrator.

aaime:
The current setup seems geared towards developers that know the internal data structures and the system variables, on top of good knowledge of CSP. Wondering if the situation should be flipped: anything in the code that has HTML output declares its presence and checks if CSP should be activated at all (sysvar checks and the like), and describes itself in terms that an end user can understand (e.g., "OGC API Features HTML representation"), there is a lookup that finds all these bits, and presents them to the user with a simple setup: do you want to enable CSP on it (yes by default), and do you need to customize it, or use a default?

sikeoka:
An administrator obviously needs to understand CSP in order to modify this and I don't think it would be possible to expect someone who doesn't understand CSP to create proper and secure policies.

Besides that, an administrator would really only need to know is how to create regular expression to match the URL path and query parameters for specific HTML outputs and this should only be necessary if someone is working on new GeoServer HTML output using unsafe JavaScript that is not a Wicket page, FreeMarker template or in the www directory. It should probably be preferred that new HTML output being added to GeoServer not use unsafe JavaScript if that is feasible.

There are new system properties for this feature but if an administrator were configuring this, they would probably just configure it directly to their specific needs rather than using system properties.

doc/en/user/source/security/csp.rst
sikeoka: The documentation no longer refers to specific Wicket version. The comment may need to be updated if a Wicket upgrade does not happen in the 2.26 release.

sikeoka: I removed the comment from the documentation for now and added a note in CSPDefaultConfiguration.java that the default configuration and documentation need to be updated after the Wicket 9 upgrade.

sikeoka:
Do you have any additional feedback on this PR? I think that all of your initial feedback has been addressed and the configuration has also been significantly simplified from the original version although it now relies on additional JavaScript refactoring PRs. The Wicket rule can also be removed after the Wicket 9 upgrade which will further simplify the default config. With the JavaScript refactoring, unsafe JavaScript is now only enabled on index.html and the www directory by default. Adding this header will not only mitigate existing XSS issues but also help prevent new XSS issue from getting into future releases.

jodygarnett:
I think my feedback has been addressed

jodygarnett:
I guess I join @aaime in feeling out of my depth reviewing this feature.

Reading above this is still a very tricky subject, a system administrator that has secured their system cannot really predict what new URLs will be available when upgrading GeoServer. As an example when installing OGCAPI services a whole host of new URLs is now available (that are not so static in nature as they are based on workspace and layer names...

Andrea your idea of turning the problem inside out (forcing any code output html to be checked)  - could that be done at the dispatcher level if it see html output?

aaime:
I guess that all the javascript refactors are merged now, and you simplified the configuration too. 
@sikeoka is this one ready to be reviewed and then merged in time for 2.26.x? Or do we need further refactors?

sikeoka:
Yes it is ready to be reviewed.

aaime:
First thing I did, was to search any leftover inline scripts:

```
> git grep "<script>" -- "**/*html"
community/features-templating/features-templating-core/src/test/resources/org/geoserver/featurestemplating/response/HTMLFilteredMappedFeature.xhtml:        <script>
community/features-templating/features-templating-core/src/test/resources/org/geoserver/featurestemplating/response/HTMLHtmlFeature.xhtml:        <script>
community/features-templating/features-templating-core/src/test/resources/org/geoserver/featurestemplating/response/HTMLMappedFeature.xhtml:        <script>
community/features-templating/features-templating-core/src/test/resources/org/geoserver/featurestemplating/response/HTMLMappedFeatureWithJSONLD.xhtml:        <script>
community/gsr/src/main/resources/demos/dynamic_map_layer.html:    <script>
community/gsr/src/main/resources/demos/layers-featurelayer-polygon.html:    <script>
community/jwt-headers/gs-jwt-headers/src/main/java/org/geoserver/security/jwtheaders/web/JwtHeadersAuthFilterPanel$JsonClaimPanel.html:    <script>
community/jwt-headers/gs-jwt-headers/src/main/java/org/geoserver/security/jwtheaders/web/JwtHeadersAuthFilterPanel$JsonClaimPanel.html:    <script>
community/jwt-headers/gs-jwt-headers/src/main/java/org/geoserver/security/jwtheaders/web/JwtHeadersAuthFilterPanel.html:    <script>
community/ogcapi/ogcapi-core/src/main/resources/swagger-ui/oauth2-redirect.html:<script>
web/demo/src/main/java/org/geoserver/web/demo/SRSDescriptionPage.html:        <script>
```

Observations: 
* The GSR stuff I think can be safely ignored, it's demos in a community module that has seen no action in a while
* The SRSDescriptionPage looks like it would stop working... or not? The script is behind the map preview showing the valid area of a SRS.
* The JWT headers community module is new, and in active usage. Perhaps the CSP activity and the new module happened in parallel. Let's also hear from @davidblasby 
* The Features Templating bits are just a test, but point to a larger issue... how does one create an interactive HTML response page from GetFeature in the age of CSP? I guess the javascript bits could be placed inside "www" for static publishing? Is it ok to have an inline style, or that also needs to be moved to an externalized CSS?

davidblasby:
@aaime - please see some discussion in the https://github.com/geoserver/geoserver/pull/7154  (Wicket 9) PR.  It has this merged in. 

I noticed CSP issues with the main GS page (fixed - Wicket 9 PR), as well as the 3 demo requests (fixed - Wicket 9 PR).
I sent an email to list ("Wicket and JS (and content-security-policy)").

Basically:
1. move any inline styles to a .css file and have it included via wicket
2. move any `<script>` to an included (via wicket) script file 
3. move any  on-element event handlers (`onClick='...'`) to an on-dom-ready wicket on-render JS block;
```
        script = "$('#linkSubmit').on('click',function() {...} );";
        response.render(OnDomReadyHeaderItem.forScript(script));
```

Since there was a problem on the main page, I expect there will be problems elsewhere.
Definitely `jwt-headers` (should be easy to fix), and a few other places (like `oauth2-`). 
I suspect there are `style="..."` in the code as well as `onClick=".."` and `onChange="..."`.

Its a bit of a pain to fix, but it should be straightforward (unless its wicket putting in the event handlers).

I would recommend going through the entire application with the browser debugger open and look for CSP issues in the console.  Also, might want to not merge this and just merge the Wicket 9 PR...

sikeoka:
The jwt-headers stuff and SRSDescriptionPage are all Wicket pages. Wicket has its own strict CSP policy that is completely separate from this PR.

I could not find anything in ogcapi that uses oauth2-redirect.html. I think it's just a file copied over from the swagger library that isn't being used by geoserver.

Enabling inline javascript for WMS GetFeatureInfo freemarker templates is just a matter of setting a system property. Enabling inline javascript for the features-templating module would require adding custom CSP rules. As you mentioned, the www directory could be used to store external javascript files so that inline javascript isn't needed for both of these features if that is preferred.

Wicket sets a CSP that also restricts inline styles but this PR is only focused on restricting inline javascript for the rest of geoserver.

sikeoka:
@aaime and @jodygarnett
Now that the Wicket 9 stuff is merged, do you guys have any additional thoughts about merging this to give people time to test it out before the next release? I added a setting to allow switching to the report only header.

jodygarnett:
I really like that idea, we can include it in the QA run to look for and resolve warnings :)

jodygarnett:
> @aaime and @jodygarnett Now that the Wicket 9 stuff is merged, do you guys have any additional thoughts about merging this to give people time to test it out before the next release? I added a setting to allow switching to the report only header.

Steve can we make a proposal? It would be handy to actually Communale this key design decision change.

doc/en/user/source/security/csp.rst
```diff
@@ -0,0 +1,270 @@
...
+found that has directives and no directives will be added if no such rule exists.
+
+The button for adding rules can be found at the top of the :guilabel:`Rule List` table and a rule
+can be edited by clicking on its name in the table or removed by clicking on the remove icon at the
+end of the rule's row in the table. Rule positions can also be changed by using the up/down arrows
+or by dragging and dropping the rule's row in the table.
+
+.. figure:: images/csp-rule.png
+
+   CSP Rule Configuration
```
jodygarnett, line 140: We need a new example as the WFS Test Servlet is not included any more.

doc/en/user/source/production/config.rst
```diff
@@ -331,6 +346,16 @@ If you wish to change this behavior you can do so through the following properti
 These properties can be set either via Java system property, command line argument (-D), environment
 variable or web.xml init parameter.
 
+.. _production_config_csp:
+
+Content-Security-Policy
+'''''''''''''''''''''''
+
+In order to mitigate cross-site scripting and clickjacking attacks GeoServer defaults to setting
+the Content-Security-Policy HTTP header based on rules loaded from a configuration file. See the
```
jodygarnett, line 355: For user's of the web admin console they do not see a configuration file.
```suggestion
the Content-Security-Policy HTTP header based on rules loaded from :ref:`security_csp` page.  See the
```

doc/en/user/source/production/config.rst
```diff
@@ -274,17 +274,32 @@ If you wish to change this behavior you can do so through the following properti
 * ``geoserver.xframe.shouldSetPolicy``: controls whether the X-Frame-Options header should be set at all. Default is true.
 * ``geoserver.xframe.policy``: controls what to set the X-Frame-Options header to. Default is ``SAMEORIGIN``. Valid options are ``DENY``, ``SAMEORIGIN`` and ``ALLOW-FROM [uri]``.
 
+These properties can be set either via Java system property, command line argument (-D), environment
+variable or web.xml init parameter.
```
jodygarnett, line 278: Is this still useful? We now have a page dedicated to this sentence: https://docs.geoserver.org/latest/en/user/configuration/properties/index.html

doc/en/user/source/production/config.rst
```diff
@@ -274,17 +274,32 @@ If you wish to change this behavior you can do so through the following properti
...
+the ``X-Frame-Options`` header.
+
+* ``SAMEORIGIN`` will be ``frame-ancestors 'self'``
+* ``DENY`` will be ``frame-ancestors 'none'``
+* if the ``X-Frame-Options`` header is not set or has any other value, the ``frame-ancestors``
+  directive will be omitted
+
+When both ``frame-ancestors`` and ``X-Frame-Options`` are present, browsers that support
+``frame-ancestors`` should **enforce** the ``frame-ancestors`` policy and **ignore** the
+``X-Frame-Options`` policy.
```
jodygarnett, line 302: These new options should be listed on [Application Properties](https://docs.geoserver.org/latest/en/user/configuration/properties/index.html) page. 

I only read this to look for the option to change the default mentioned in the PR and cannot figure out you were trying to say.

doc/en/user/source/security/csp.rst
```diff
@@ -0,0 +1,270 @@
...
+  Example: ``PROP(GEOSERVER_CONSOLE_DISABLED,(?i)^(?!true$).*$)``
+
+.. note::
+    The ``(?i)`` at the beginning of the regular expression will use case-insensitive matching and
+    enclosing the pattern inside of the ``^$`` characters will match the entire string. See the
+    `Regular Expressions Tutorial <https://docs.oracle.com/javase/tutorial/essential/regex/>`_ for
+    more information about how to use Java regular expressions.
+
+Leaving the filter blank will cause this rule to match all requests and should only be used on the
+last rule in a policy since any additional rules would never be checked.
```
jodygarnett, line 187: That seems really hard to understand; if blank is aways used then it does not need to be shown (or maybe indicated another way).

jodygarnett:
@sikeoka I have been reading trying to find out this information for the proposal:

> Enabling inline javascript for WMS GetFeatureInfo freemarker templates is just a matter of setting a system property.

doc/en/user/source/production/config.rst
sikeoka: It probably isn't necessary but that sentence exists in the current documentation. It just got moved around and shows up as new in the diff.

doc/en/user/source/production/config.rst
sikeoka: I will remove the reference here to a configuration file.

doc/en/user/source/production/config.rst
sikeoka: Basically, this is trying to say that CSP frame-ancestors is more powerful than X-Frame-Options and frame-ancestors will override X-Frame-Options if both are set and the browser supports both. The frame-ancestors also defaults to having the same behavior as X-Frame-Options to prevent compatibility issues. Using 'self' as the default frame-ancestors would significantly simplify that configuration but would break frames if an administrator had disabled X-Frame-Options and allow frames from the same origin if an administrator had set X-Frame-Options to DENY.

sikeoka:
> @sikeoka I have been reading trying to find out this information for the proposal:
> 
> > Enabling inline javascript for WMS GetFeatureInfo freemarker templates is just a matter of setting a system property.

https://github.com/geoserver/geoserver/pull/7514/files#diff-4189e15f1a68d8fe29c95f1201354dd5cef9f954c3b6b2330e0a941b64537b3b

jodygarnett:
@sikeoka reading through this with an eye of making GSIP-227 make sense ... 

Do I have the correct understanding:

- The long term goal should be to make CSP restricted by default
- Provide a user-interface for rare case (like GetFeatureInfo `HTML`) where user can follow a manual process

doc/en/user/source/security/csp.rst
sikeoka: Done

sikeoka:
@jodygarnett Yes, that is correct.

sikeoka:
@jodygarnett Do you need any more information about this?

jodygarnett:
I am sorry @sikeoka I do need some more information. I created a proposal (https://github.com/geoserver/geoserver/wiki/GSIP-227) in order to communicate the strategic direction for introducing CSP headers. This decision was made on the geoserver-security list (understandably) but it also means that not many folks were aware of what was going on and why until we were ready to disclose the first round of fixes.

It has been an "action" item to ask you for a clarification on this proposal for the last couple geoserver meetings:

> Actions: 
> - Jody: Send CSP proposal to vote when Steve has confirmed default functionality is “fine”, and regex is only required if user wants to allow something

So if you can confirm that I can update the proposal :)

jodygarnett:
Thanks, I updated the proposal with your confirmation, and have [sent it to vote](https://discourse.osgeo.org/t/gsip-227-content-security-policy-headers/111180).

At least now this change will not be a surprise!

sikeoka:
@jodygarnett Do you have an update on this?

PR #7531, opened 2024-04-01
[GEOS-11349] MapML Use WMS Resource Consumption Limit to specify max image size

turingtestfail:
[GEOS-11349](https://osgeo-org.atlassian.net/browse/GEOS-11349) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

…image size

  

# Checklist

### DIFF
--- a/src/extension/mapml/src/main/java/org/geoserver/mapml/MapMLDocumentBuilder.java
+++ b/src/extension/mapml/src/main/java/org/geoserver/mapml/MapMLDocumentBuilder.java
@@ -105,6 +105,9 @@ public class MapMLDocumentBuilder {
     public static final String MAPML_MULTILAYER_AS_MULTIEXTENT = "mapmlMultiLayerAsMultiExtent";
 
     protected static final Boolean MAPML_MULTILAYER_AS_MULTIEXTENT_DEFAULT = Boolean.FALSE;
+    public static final String MINIMUM_WIDTH_HEIGHT = "1";
+    private static final int BYTES_PER_PIXEL_TRANSPARENT = 4;
+    private static final int BYTES_PER_KILOBYTE = 1024;
     private final WMS wms;
 
     private final GeoServer geoServer;
@@ -1427,21 +1430,7 @@ public class MapMLDocumentBuilder {
         input.setMax(Double.toString(bbbox.getMaxY()));
         extentList.add(input);
 
-        // width
-        input = new Input();
-        input.setName("w");
-        input.setType(InputType.WIDTH);
-        input.setMin("1");
-        input.setMax("10000");
-        extentList.add(input);
-
-        // height
-        input = new Input();
-        input.setName("h");
-        input.setType(InputType.HEIGHT);
-        input.setMin("1");
-        input.setMax("10000");
-        extentList.add(input);
+        createMinAndMaxWidthHeight();
 
         // image link
         Link imageLink = new Link();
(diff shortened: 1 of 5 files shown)


prushforth:
When I load up GeoServer with (what I believe to be) the default WMS request memory which was around 65000KB (I messed around with it, don't recall the exact number but it was around 65K), I get width and height inputs as follows (using the states layer, population style):

```html
<map-input name="w" type="width" min="1" max="11401"></map-input>
<map-input name="h" type="height" min="1" max="11401"></map-input>
```

So I took the preview WMS request for text/mapml and [bumped it up to 10000x10000](http://localhost:8080/geoserver/topp/wms?request=GetMap&crs=MapML:CBMTILE&service=WMS&bbox=-4943032.241131149,-7441968.503937006,4780424.705782741,4609826.4329862&layers=states&format=image/png&width=10000&styles=population&language=en&version=1.3.0&transparent=true&height=10000), and I get this in response:

```xml
<ServiceExceptionReport xmlns="http://www.opengis.net/ogc" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.3.0" xsi:schemaLocation="http://www.opengis.net/ogc http://localhost:8080/geoserver/schemas/wms/1.3.0/exceptions_1_3_0.xsd">
<ServiceException code="MaxMemoryExceeded"> Rendering request would use 390625KB, whilst the maximum memory allowed is 65000KB </ServiceException>
</ServiceExceptionReport>
```

Seems to me that the values for `max` should be a lot lower?  When I set the max memory to 0, I get a reeeeallly big image back.

turingtestfail:
> When I load up GeoServer with (what I believe to be) the default WMS request memory which was around 65000KB (I messed around with it, don't recall the exact number but it was around 65K), I get width and height inputs as follows (using the states layer, population style):
> 
> ```
> <map-input name="w" type="width" min="1" max="11401"></map-input>
> <map-input name="h" type="height" min="1" max="11401"></map-input>
> ```
> 
> So I took the preview WMS request for text/mapml and [bumped it up to 10000x10000](http://localhost:8080/geoserver/topp/wms?request=GetMap&crs=MapML:CBMTILE&service=WMS&bbox=-4943032.241131149,-7441968.503937006,4780424.705782741,4609826.4329862&layers=states&format=image/png&width=10000&styles=population&language=en&version=1.3.0&transparent=true&height=10000), and I get this in response:
> 
> ```
> <ServiceExceptionReport xmlns="http://www.opengis.net/ogc" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.3.0" xsi:schemaLocation="http://www.opengis.net/ogc http://localhost:8080/geoserver/schemas/wms/1.3.0/exceptions_1_3_0.xsd">
> <ServiceException code="MaxMemoryExceeded"> Rendering request would use 390625KB, whilst the maximum memory allowed is 65000KB </ServiceException>
> </ServiceExceptionReport>
> ```
> 
> Seems to me that the values for `max` should be a lot lower? When I set the max memory to 0, I get a reeeeallly big image back.

You are absolutely correct.  I was making a faulty assumption (and not properly testing the final validation) that the image depth was 4 bits.  In fact, GeoServer does a check that assumes 3-4 bytes depth.  I have added a more complete validation test and changed the logic to match the existing check.

prushforth (approved):
Works very well now

turingtestfail:
> Seems to me that the values for `max` should be a lot lower? When I set the max memory to 0, I get a reeeeallly big image back.

Quick question - with this PR, when the max memory is set to 0, the max MapML XML value is set to null because there is no max from GeoServer's perspective.  Is that ok for the client?  How does the client handle that scenario?

prushforth:
If the max memory is set to 0, and there's no limit as far as GeoServer is concerned, it would be ok to omit the max attribute in that case. They are just advisory limits, and the client currently ignores them.

aaime:
Looks good. At the same time, I feel like the usage of optional parameters and long method signatures got a bit out of hand here. I've pushed an alternative using a builder instead (hammered it mostly with IntelliJ refactors, with some manual changes in the mix). @turingtestfail can you have a look?

turingtestfail:
> Looks good. At the same time, I feel like the usage of optional parameters and long method signatures got a bit out of hand here. I've pushed an alternative using a builder instead (hammered it mostly with IntelliJ refactors, with some manual changes in the mix). @turingtestfail can you have a look?

Thanks - looks good.  Rebased with the new commit.

aaime (approved):

PR #7545, opened 2024-04-12
[GEOT-7422] Adapt to GeoTools upgrade of java units "indriya" to version 2.2

soc:
[GEOT-7422](https://osgeo-org.atlassian.net/browse/GEOT-7422) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

Companion change to https://github.com/geotools/geotools/pull/4690.

# Checklist

### DIFF
--- a/src/extension/netcdf/src/main/java/org/geoserver/netcdf/NetCDFUnitsConfigurator.java
+++ b/src/extension/netcdf/src/main/java/org/geoserver/netcdf/NetCDFUnitsConfigurator.java
@@ -40,7 +40,7 @@ public class NetCDFUnitsConfigurator implements GeoServerLifecycleHandler {
             LinkedHashMap<String, String> aliases =
                     getMapResource(NETCDF_UNIT_ALIASES, NetCDFUnitFormat.NETCDF_UNIT_ALIASES);
             if (aliases != null) {
-                NetCDFUnitFormat.setAliases(aliases);
+                NetCDFUnitFormat.getInstance().setAliases(aliases);
             }
         } catch (IOException e) {
             LOGGER.log(Level.WARNING, "Failed to load NetCDF unit aliases", e);
@@ -51,7 +51,7 @@ public class NetCDFUnitsConfigurator implements GeoServerLifecycleHandler {
                     getMapResource(
                             NETCDF_UNIT_REPLACEMENTS, NetCDFUnitFormat.NETCDF_UNIT_REPLACEMENTS);
             if (replacements != null) {
-                NetCDFUnitFormat.setReplacements(replacements);
+                NetCDFUnitFormat.getInstance().setReplacements(replacements);
             }
         } catch (IOException e) {
             LOGGER.log(Level.WARNING, "Failed to load NetCDF unit replacements", e);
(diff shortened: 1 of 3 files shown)


src/extension/netcdf/src/main/java/org/geoserver/netcdf/NetCDFUnitFormatInstance.java
```diff
@@ -0,0 +1,18 @@
+/* (c) 2018 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+
+package org.geoserver.netcdf;
+
+import org.geotools.imageio.netcdf.NetCDFUnitFormat;
```
aaime, line 8: This is not going to work. The units configuration is there to support reads too (see VariableAdapter usage in GeoTools), otherwise the class would be placed in netcdf-out.

However, I don't see why you can't use NetCDFUnitFormat directly, and go for a different instance instead.
If mutability is the issue, thinking out loud, we could have a way to build a new NetCDFUnitFormat internal instance, and just throw away the old one: this way the unit format would be immutable, for as long as it lives. Just inform the callers not to cache the formatter instance, but always get it fresh from NetCDFUnitFormat.

soc: Thanks! You mean that we also use that instance from `VariableAdapter` then?

aaime: I mean the configuration file in GeoServer should affect also VariableAdapter parsing. In other words, the configuration is used also to parse NetCDF files with uncommon units, and not only to control the generation of new NetCDFs in the GeoServer output format.

aaime:
Another thing I've noticed, gs-main won't build unless I control the unit-api dependency in the main pom file, adding it to the dependency management. 
Otherwise, the transitive dependency from systems.uom takes over and brings me back to an older version of javax.units, that does not have the QUETTA constant.

soc:
Thanks, will try addressing these things!

soc:
Reverted the undesired changes, added the unit-api dependency.

src/extension/netcdf/src/main/java/org/geoserver/netcdf/NetCDFUnitsConfigurator.java
```diff
@@ -40,7 +40,7 @@ private void configure() {
             LinkedHashMap<String, String> aliases =
                     getMapResource(NETCDF_UNIT_ALIASES, NetCDFUnitFormat.NETCDF_UNIT_ALIASES);
             if (aliases != null) {
-                NetCDFUnitFormat.setAliases(aliases);
+                NetCDFUnitFormat.getInstance().setAliases(aliases);
```
aaime, line 43: The changes look good but [setAliases is private](https://github.com/geotools/geotools/pull/4690/files#diff-ea578c094576d0dbfc82e0a8f1e124d9c468944092fcac75f71e33c79e687121R154) in the GeoTools PR. 

Did you forget to push some changes over there?

src/extension/netcdf/src/main/java/org/geoserver/netcdf/NetCDFUnitsConfigurator.java
```diff
@@ -51,7 +51,7 @@ private void configure() {
                     getMapResource(
                             NETCDF_UNIT_REPLACEMENTS, NetCDFUnitFormat.NETCDF_UNIT_REPLACEMENTS);
             if (replacements != null) {
-                NetCDFUnitFormat.setReplacements(replacements);
+                NetCDFUnitFormat.getInstance().setReplacements(replacements);
```
aaime, line 54: Same as above, this is also private in the corresponding GeoTools PR.

aaime (changes requested):

aaime:
Looks good, thanks, merging!

PR #7553, opened 2024-04-18
[GEOS-11306] Chace GetFeature total using AtomicLong for thread safety

jodygarnett:
[GEOS-11306](https://osgeo-org.atlassian.net/browse/GEOS-11306) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This addresses a regression that was introduded with API change to avoid calculating total for formats that do not require such a value.

# Checklist

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java
@@ -17,6 +17,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Supplier;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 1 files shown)


jodygarnett:
Since this is release blocker I am submitting as-is trusting to test coverage by existing tests.

I am trying to figure out if I can write a specific test to ensure value is not being called twice.

src/wfs/src/main/java/org/geoserver/wfs/GetFeature.java
```diff
@@ -711,10 +712,14 @@ private Supplier<BigInteger> updateTotalCount(
...
             return () -> BigInteger.valueOf(total);
         } else {
+            AtomicLong cache = new AtomicLong(Long.MIN_VALUE);
             return () -> {
                 try {
-                    long totalCount1 = getTotalCount(totalCountExecutors);
-                    return BigInteger.valueOf(totalCount1);
+                    if (cache.get() == Long.MIN_VALUE) {
+                        long totalCount1 = getTotalCount(totalCountExecutors);
+                        return BigInteger.valueOf(totalCount1);
```
aaime, line 720: The cache should be updated here, no?

aaime, line 720: The cache should be updated here no?

Something  like this (untested):

```suggestion
                        long totalCount1 = getTotalCount(totalCountExecutors);
                        cache.set(totalCount1);
                        return BigInteger.valueOf(totalCount1);
```

jodygarnett: Yep, updated.

PR #7560, opened 2024-04-19
[GEOS-11371] Refactor inline JavaScript in the GetMap OpenLayers format

sikeoka:
[GEOS-11371](https://osgeo-org.atlassian.net/browse/GEOS-11371) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This PR moves the inline JavaScript in the WMS GetMap Openlayers output format into an external file. This PR is related to Content-Security-Policy work for GeoServer 2.26.0 and should **NOT** be backported.

All Java inputs that were previously written directly into the script tag are now written to hidden input fields that the external JavaScript file will read. The previous code had to mix HTML and JavaScript escaping depending on where inputs were being written to but the new code only needs to use HTML escaping.

The GetFeatureInfo output of the OpenLayers2 format was previously handled with an OpenLayers AJAX call and writing the response into the DOM but is now handled by an iframe like the OpenLayers3 output format. The reason for this is that the iframe uses its own CSP which allows the GetFeatureInfo HTML to use a more lenient CSP than the OpenLayers page. For example, the GetFeatureInfo CSP can enable inline scripts even if the OpenLayers page CSP disables inline scripts.

This PR is split into two commits for easier reviewing. The first commit is a straight cut and paste of the script tag into a separate file and the second commit has the actual refactoring. Viewing the details of the second commit with the "Hide whitespace" diff option enabled will make it easier to see what was changed in the JavaScript code.

This PR uses the new webresources path added to gs-main by https://github.com/geoserver/geoserver/pull/7554 which does not affect unit tests but is required if attempting to manually run GeoServer with these changes.
  

# Checklist


mprins (approved):
lgtm

petersmythe:
@sikeoka Steve, can we merge this into main (I see not to backport this to 2.25.x, 2.24.x), or does it need to remain open?

Ditto for #7587 and #7554

aaime:
@petersmythe PRs need a review before getting merged, unless they are trivial ones. For a change like this one, whose behavior is not covered by any test, I'd expect a bit of manual testing.
This one has been checked by @mprins already, the other two aren't.

aaime:
@petersmythe when a PR gets a positive review with no comments left out to process, you can merge it.

PR #7580, opened 2024-04-25
[GEOS-11332] Renaming style with uppercase/downcase empty the sld file

aaime:
[GEOS-11332](https://osgeo-org.atlassian.net/browse/GEOS-11332) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

Nasty little bug that causes "data loss" (as in, stylesheet loss)

# Checklist

### DIFF
--- a/src/platform/src/main/java/org/geoserver/util/IOUtils.java
+++ b/src/platform/src/main/java/org/geoserver/util/IOUtils.java
@@ -30,6 +30,7 @@ import java.util.zip.ZipFile;
 import java.util.zip.ZipInputStream;
 import java.util.zip.ZipOutputStream;
 import org.apache.commons.io.FileUtils;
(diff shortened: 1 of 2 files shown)


aaime:
@jodygarnett do you have 5 minutes for a review? It's loosely in Resource area, while it does not change it directly, the method is being used by it to handle renames. The original code was written with case-insensitive file systems in mind, and causes data loss on Linux (style body is lost, at least from the UI, takes a little digging into the data dir to recover it).

jodygarnett (approved):
Wow the is a challenge, there is no way I know in java to detect if a file system is case sensitive.

aaime:
Yes, neither do I. But Files.move should figure it out by itself

PR #7590, opened 2024-05-02
[GEOS-11382] The interceptor "CiteComplianceHack" never gets invoked by the Dispatcher Servlet

afabiani:
[GEOS-11382](https://osgeo-org.atlassian.net/browse/GEOS-11382) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

Fixes: https://osgeo-org.atlassian.net/browse/GEOS-11382
  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/ows/CiteComplianceHack.java
+++ b/src/main/src/main/java/org/geoserver/ows/CiteComplianceHack.java
@@ -69,7 +69,7 @@ public class CiteComplianceHack implements HandlerInterceptor {
             return dispatcher.service(req);
         } catch (Throwable t) {
             LOGGER.log(Level.FINE, "Exception while looking for the 'Service' from the request", t);
-            // load from teh context
+            // load from the context
             try {
                 UriComponentsBuilder builder =
                         UriComponentsBuilder.fromUriString(request.getServletPath());
--- a/src/main/src/main/java/org/geoserver/ows/CiteComplianceHack.java
+++ b/src/main/src/main/java/org/geoserver/ows/CiteComplianceHack.java
@@ -13,7 +13,6 @@ import javax.servlet.http.HttpServletResponse;
 import org.geoserver.config.GeoServer;
 import org.geoserver.config.ServiceInfo;
 import org.geoserver.platform.Service;
-import org.geoserver.platform.ServiceException;
 import org.geotools.util.logging.Logging;
 import org.springframework.web.servlet.HandlerInterceptor;
 import org.springframework.web.servlet.ModelAndView;
@@ -44,10 +43,10 @@ public class CiteComplianceHack implements HandlerInterceptor {
             throws Exception {
         if (handler instanceof Dispatcher) {
             Dispatcher dispatcher = (Dispatcher) handler;
-            Service service = findService(dispatcher, request, response);
+            String service = findService(dispatcher, request, response);
             if (service != null
-                    && (service.getId().equalsIgnoreCase(getInfo().getId())
-                            || service.getId().equalsIgnoreCase(getInfo().getName()))) {
+                    && (service.equalsIgnoreCase(getInfo().getId())
+                            || service.equalsIgnoreCase(getInfo().getName()))) {
                 dispatcher.setCiteCompliant(getInfo().isCiteCompliant());
             }
         }
@@ -55,7 +54,7 @@ public class CiteComplianceHack implements HandlerInterceptor {
         return true;
     }
 
-    private Service findService(
+    private String findService(
             Dispatcher dispatcher, HttpServletRequest request, HttpServletResponse response) {
         // create a new request instance
         Request req = new Request();
@@ -66,9 +65,10 @@ public class CiteComplianceHack implements HandlerInterceptor {
 
         // find the service
         try {
-            return dispatcher.service(req);
-        } catch (Throwable t) {
-            LOGGER.log(Level.FINE, "Exception while looking for the 'Service' from the request", t);
+            return dispatcher.getServiceFromRequest(req);
+        } catch (Exception ex1) {
+            LOGGER.log(
+                    Level.FINE, "Exception while looking for the 'Service' from the request", ex1);
             // load from the context
             try {
                 UriComponentsBuilder builder =
@@ -82,23 +82,14 @@ public class CiteComplianceHack implements HandlerInterceptor {
                                     req.getVersion(),
                                     req.getNamespace());
                     if (serviceDescriptor != null) {
-                        req.setServiceDescriptor(serviceDescriptor);
-                        try {
-                            return dispatcher.fireServiceDispatchedCallback(req, serviceDescriptor);
-                        } catch (ServiceException se) {
-                            LOGGER.log(
-                                    Level.FINE,
-                                    "Exception while searching for the 'Service Descriptor'",
-                                    se);
-                            return null;
-                        }
+                        return serviceDescriptor.getId();
                     }
                 }
-            } catch (Exception e) {
+            } catch (Exception ex2) {
                 LOGGER.log(
                         Level.FINE,
                         "Exception while decoding OWS URL " + request.getServletPath(),
-                        e);
+                        ex2);
             }
             return null;
         }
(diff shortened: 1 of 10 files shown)


src/wms/src/main/java/applicationContext.xml
```diff
@@ -408,10 +414,12 @@
...
-    <bean id="wmsCiteComplianceHack" class="org.geoserver.ows.CiteComplianceHack">
-        <constructor-arg ref="geoServer"/>
-        <constructor-arg value="org.geoserver.wms.WMSInfo"/>
-    </bean>
+    <mvc:interceptors>
+        <bean id="wmsCiteComplianceHack" class="org.geoserver.ows.CiteComplianceHack">
+            <constructor-arg ref="geoServer"/>
+            <constructor-arg value="org.geoserver.wms.WMSInfo"/>
+        </bean>
+    </mvc:interceptors>
```
aaime, line 422: I heard that in order to make WMS compliant, one had to make all the other services compliant... but this change seems to be related to having the interceptor be actually invoked.  The same interceptor is declared for other services as well, why has it been changed only here?

I've done a quick test and here are some thoughts.

The situation is confusing. Each OGC service should have its own compliance hack declaration, but only a few actually do, WMS, WFS, and CSW (but not WCS for example, maybe it's strict by default):

```xml
C:\Users\andre\devel\geoserver>git grep -i citecompliancehack -- "*.xml"
src/extension/csw/core/src/main/java/applicationContext.xml:          <ref bean="cswCiteComplianceHack"/>
src/extension/csw/core/src/main/java/applicationContext.xml:    <bean id="cswCiteComplianceHack" class="org.geoserver.ows.CiteComplianceHack">
src/wfs/src/main/java/applicationContext.xml:   <bean id="wfsCiteComplianceHack" class="org.geoserver.ows.CiteComplianceHack">
src/wfs/src/main/java/applicationContext.xml:                           <ref bean="wfsCiteComplianceHack"/>
src/wms/src/main/java/applicationContext.xml:    <bean id="wmsCiteComplianceHack" class="org.geoserver.ows.CiteComplianceHack">
```

Originally the all used to run, maybe some Spring upgrade broke that? In any case, looking at the compliance hack source code, it makes no attempt to check which service is being called. So if all 3 of the above hacks are called, the last one wins, which is a random choice picking the cite compliance configuration of the last hack that gets run.

The hack should be doing some attempt at figuring out if the service is a match, e.g., check the last element of the path, and/or check if there is a service=<id> kvp parameter that matches the current service configuration.

My other worry is that the mvc configuration is something new, and it might end up interacting with the other dispatchers, like the REST API and the OGC API one... hopefully not, the cite compliance hack specifically checks that the handler is the OWS Dispatcher. But I'd recommend doing some light manual testing of the REST API and the OGC API in a running GeoServer.

Long story short:
* Update the hack to check its own configured service only
* Update other OGC services with CITE hacks as well
* Add a test to ensure the flag is working, it's evidently missing or we would not have had this regression

afabiani: Yessir!

afabiani: @aaime turns out that there's no need to introduce the `mvc:interceptors` tag. There was just a missing initialized element on the `wmsURLMapping` bean.

I've been added few more tests and checks for the current service.

src/main/src/main/java/org/geoserver/ows/CiteComplianceHack.java
```diff
@@ -36,12 +44,66 @@ public boolean preHandle(
...
+        // create a new request instance
+        Request req = new Request();
+
+        // set request / response
+        req.setHttpRequest(request);
+        req.setHttpResponse(response);
+
+        // find the service
+        try {
+            return dispatcher.service(req);
```
aaime, line 69: Good idea, but problematic. The service method at the end calls the dispatcher callbacks, which should not happen:
* they would receive events out of order
* they would receive events multiple times (when the dispatcher actually parses the request)

That said, it should be easy to refactour a method that only computes the service descriptor, make it static, and call it from both the existing Dispatcher.service(...) and from here.

I'm worried about POST requests, as the XML body is read while computing the service... it is properly "pushed back" so that the dispatcher can parse it again? You might want to test a GetFeature in XML format, interactively and/or as part of tests.

src/main/src/main/java/org/geoserver/ows/CiteComplianceHack.java
```diff
@@ -36,12 +44,66 @@ public boolean preHandle(
...
+        } catch (Throwable t) {
```
aaime, line 70: Never catch a Throable, you want Error instances (e..g, OOM) to go free. Catch Exception instead.

src/main/src/main/java/org/geoserver/ows/CiteComplianceHack.java
```diff
@@ -36,12 +44,66 @@ public boolean preHandle(
...
+                        req.setServiceDescriptor(serviceDescriptor);
+                        try {
+                            return dispatcher.fireServiceDispatchedCallback(req, serviceDescriptor);
+                        } catch (ServiceException se) {
+                            LOGGER.log(
+                                    Level.FINE,
+                                    "Exception while searching for the 'Service Descriptor'",
+                                    se);
+                            return null;
+                        }
```
aaime, line 94: Why go through the callbacks, and not just return the service found?

src/main/src/main/java/org/geoserver/ows/CiteComplianceHack.java
```diff
@@ -36,12 +44,66 @@ public boolean preHandle(
...
+        // set request / response
+        req.setHttpRequest(request);
+        req.setHttpResponse(response);
+
+        // find the service
+        try {
+            return dispatcher.service(req);
+        } catch (Throwable t) {
+            LOGGER.log(Level.FINE, "Exception while looking for the 'Service' from the request", t);
+            // load from teh context
```
aaime, line 72:
```suggestion
            // load from the context
```

src/wfs/src/main/java/applicationContext.xml
```diff
@@ -1,11 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- 
+<!--
  Copyright (C) 2014 - 2016 Open Source Geospatial Foundation. All rights reserved.
  This code is licensed under the GPL 2.0 license, available at the root
  application directory.
  -->
 <!DOCTYPE beans PUBLIC "-//SPRING//DTD BEAN//EN" "http://www.springframework.org/dtd/spring-beans.dtd">
-
```
aaime, line 8: Several unrelated changes in this file (whitespace)

src/wms/src/main/java/applicationContext.xml
```diff
@@ -1,12 +1,12 @@
-
-<!--
```
aaime, line 2: More whiltespace unrelated changes

src/wms/src/main/java/applicationContext.xml
```diff
@@ -418,6 +418,11 @@
 		class="org.geoserver.ows.OWSHandlerMapping">
 		<constructor-arg ref="catalog"/>
 		<property name="alwaysUseFullPath" value="true"/>
+        <property name="interceptors">
```
aaime, line 421: This block is mis-aligned:

image

aaime (changes requested):
This is definitely going in the right direction. See comments below.

aaime:
Looks good to me. Pay attention the QA build is failing.

src/main/src/main/java/org/geoserver/ows/CiteComplianceHack.java
afabiani: done

src/main/src/main/java/org/geoserver/ows/CiteComplianceHack.java
afabiani: done

src/wfs/src/main/java/applicationContext.xml
afabiani: done

src/wms/src/main/java/applicationContext.xml
afabiani: done

aaime (approved):

PR #7615, opened 2024-05-09
[GEOS-11351] Exact term search in the pages' filters

axl8713:
[GEOS-11351](https://osgeo-org.atlassian.net/browse/GEOS-11351) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

These changes add the ability to refine element searches within paginated results through the search/filter input field, providing users with the option to request an exact-term filter by surrounding the term with quotes or double-quotes.
  

# Checklist


doc/en/user/source/gettingstarted/web-admin-quickstart/index.rst
```diff
@@ -73,6 +73,8 @@ The :ref:`layerpreview` page allows you to quickly view the output of published
    
       Search results for the query "top" on the Workspace page
 
+   .. hint:: Perform an exact term search by enclosing the search term in quotes or double-quotes, e.g. "topp".
```
aaime, line 76:
```suggestion
   .. hint:: Perform an exact term search by enclosing the search term in quotes or double-quotes, e.g. normally ``ads`` would also match ``roads``, but ``"ads"`` wouldn't.
```

src/main/src/main/java/org/geoserver/catalog/Predicates.java
```diff
@@ -198,6 +198,17 @@ public static Filter fullTextSearch(final String subsequence) {
...
 
+    /**
+     * Returns a predicate for exact match search (case-insensitive) against the <code>AnyText
+     * </code> property with the provided term.
+     *
+     * @param term the exact term to search for, ignoring case
+     * @return the filter for exact match search
+     */
+    public static Filter exactTermSearch(final String term) {
+        return factory.like(ANY_TEXT, fixSpecials(term));
```
aaime, line 209: Why use "like", it seems you're looking for simple equality... if so, ``Predicates.equal`` should also work.

src/main/src/test/java/org/geoserver/catalog/impl/CatalogImplTest.java
```diff
@@ -3514,6 +3514,165 @@ public void testFullTextSearchAddedKeyword() {
...
+
+        LayerInfo lproxy = catalog.getLayer(l.getId());
+        FeatureTypeInfo ftproxy = (FeatureTypeInfo) lproxy.getResource();
+
+        ftproxy.getKeywords().add(new Keyword("newKeyword"));
+        catalog.save(ftproxy);
+
+        Filter filter = Predicates.exactTermSearch("newKeyword");
+        assertEquals(newHashSet(ftproxy), asSet(catalog.list(FeatureTypeInfo.class, filter)));
+        assertEquals(newHashSet(lproxy), asSet(catalog.list(LayerInfo.class, filter)));
```
aaime, line 3643: I would add a test before the keyword being added, which should result in an empty result

src/main/src/test/java/org/geoserver/catalog/impl/CatalogImplTest.java
```diff
@@ -3514,6 +3514,165 @@ public void testFullTextSearchAddedKeyword() {
...
+        addLayer();
+        catalog.add(cv);
+
+        LayerInfo l2 = newLayer(cv, s);
+        catalog.add(l2);
+
+        Filter filter = Predicates.exactTermSearch("water");
+        assertEquals(newHashSet(ft), asSet(catalog.list(ResourceInfo.class, filter)));
+        assertEquals(newHashSet(ft), asSet(catalog.list(FeatureTypeInfo.class, filter)));
+        assertTrue(asSet(catalog.list(CoverageInfo.class, filter)).isEmpty());
```
aaime, line 3662: In general, for new code, we try to avoid using assertTrue/assertFalse. Hamcrest provides a way to make the test fail with a more informative message, when that happens (assertTrue/assertFalse only report "I was expecting true but got false", Hamcrest reports "I was expecting an empty collection but got a, b, c")

```suggestion
        assertThat(asSet(catalog.list(CoverageInfo.class, filter)), Matchers.empty());
```

src/web/core/src/main/java/org/geoserver/web/wicket/GeoServerDataProvider.java
```diff
@@ -43,6 +41,8 @@
  * @param <T>
  */
 public abstract class GeoServerDataProvider<T> extends SortableDataProvider<T, Object> {
+    public static final Pattern EXACT_TERM_KEYWORD_PATTERN =
+            Pattern.compile("^([\"'])(?<keyword>.+)\\1$");
```
aaime, line 45: This one needs some explanation, something like:

```
// matches exact terms enclosed by either double or single quotes
// The first matching group matches either " or ', the "keyword" matching group matches any sequence of at least one character, the last group refers to the first one (\\1$) ensuring the start and end quote are balanced
```
A few more notes:
* Does the field need to be public? If not, default to private or protected.
* The documentation does not mention single quote being a possibility, please amend either docs or implementation, or people will wonder why searching for ``un'azione dell'altro mondo`` weirdly only matches ``azione dell`` but not ``azione``. Given the single quote has legit usages, I'd go for removing it from the regex, which will also simplify it
* I don't see a way to escape the double quote character, should it happen to be a valid value in the string. This could also be addressed by documentation, stating explicitly that there is no escaping support.

src/web/core/src/main/java/org/geoserver/web/wicket/GeoServerTablePanel.html
```diff
@@ -12,6 +12,7 @@
 		<div class="float-end">
 			<input id="filter" wicket:id="filter" class="me-1 search" type="text" size="35" title="Search" placeholder="Search" wicket:message="placeholder:Search"/>
 			<input wicket:id="submit" class="d-none" type="submit" value="Search"/>
+			<span class="help-link" title="Enclose the search term in quotes for exact-term search."></span>
```
aaime, line 15: This should be subject to internationalization (you can provide support for english only for starters, in ``GeoServerApplication.properties``). Also, make sure it alignes with actual implementation (single or double quotes, see above).

src/web/core/src/test/java/org/geoserver/web/wicket/GeoServerDataProviderTest.java
```diff
@@ -0,0 +1,149 @@
...
+                }
+            };
+
+    @Test
+    public void fullTextFilter() {
+        geoServerDataProvider.setKeywords(new String[] {"keyword"});
+
+        Filter filter = geoServerDataProvider.getFilter();
+
+        assertEquals("[ AnyText is like *keyword* ]", filter.toString());
```
aaime, line 37: Testing against toString is unreliable, developers are free to change its result without changing the semantics of the method (any string that "textually represents the object" is valid). Instead, use ``ECQL.toCQL(filter)``, it provides a text output with a format and stable grammar.

I'm not commenting the other cases in this class, but please fix them all.

aaime (changes requested):
Overall looks good, see comments below.

src/web/core/src/main/java/org/geoserver/web/wicket/GeoServerDataProvider.java
axl8713: Added a Javadoc for the field and made it private. Thanks for the feedback.

Regarding the search query behavior, if a query such as 'un'azione dell'altro mondo` is entered, the query will be tokenized by splitting on spaces, resulting in the search terms: `["un'azione", "dell'altro", "mondo"]`. This maintains the legitimate use of the single quote within the words.

Additionally, the double quote character could still be part of the string. For example, searching for `un'azione dell'altro "mondo` would produce the keywords: `["un'azione", "dell'altro", ""mondo"]`.

doc/en/user/source/gettingstarted/web-admin-quickstart/index.rst
axl8713: Added the example. Thanks.

src/web/core/src/test/java/org/geoserver/web/wicket/GeoServerDataProviderTest.java
axl8713: Fixed, thanks.
Removed also other `toString()` in assertions.

src/main/src/test/java/org/geoserver/catalog/impl/CatalogImplTest.java
axl8713: Changed. Thanks for the feedback.

src/main/src/test/java/org/geoserver/catalog/impl/CatalogImplTest.java
axl8713: Added a test for a search miss on a different keyword. Thanks.

src/web/core/src/main/java/org/geoserver/web/wicket/GeoServerTablePanel.html
axl8713: Added. Thanks.

src/main/src/main/java/org/geoserver/catalog/Predicates.java
aaime: Equal can also run in case-insensitive mode, but Predicates.equal is forcing case sensitiveness inside. Simple equality is faster, like is implemented as regular expressions, but it should not be a big deal. You can either expose a Predicates.equal variant that allows to tun the matchCase attribute, or keep it this as it is.

aaime (approved):
Guess the exact match performance can be optimized later, if needed. I guess it won't be the number one use case for the search bar anyways

PR #7620, opened 2024-05-10
[GEOS-11040] Fix serviceInfo missing in FeatureService and TileService classes

nikolovasara:
Resolve frequent log messages generated by DisabledServiceCheck class
Referenced JIRA issue: [GEOS-11040](https://osgeo-org.atlassian.net/browse/GEOS-11040)

# Checklist

### DIFF
--- a/src/community/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/FeatureService.java
+++ b/src/community/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/FeatureService.java
@@ -199,6 +199,12 @@ public class FeatureService {
         return geoServer.getService(WFSInfo.class);
     }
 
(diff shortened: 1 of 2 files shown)


src/community/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/FeatureService.java
```diff
@@ -199,6 +199,12 @@ public WFSInfo getService() {
         return geoServer.getService(WFSInfo.class);
     }
 
+    @SuppressWarnings("unused")
+    public WFSInfo getServiceInfo() {
```
aaime, line 203: Oh interesting... yeah, that callback uses reflection to access the info form the service object itself!
One day we should switch to just having an interface for it (ServiceInfoProvider or something similar).

PR #7638, opened 2024-05-15
[GEOS-11399] Use Catalog streaming API in LayerGroupPage

groldan:
[GEOS-11399](https://osgeo-org.atlassian.net/browse/GEOS-11399) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

Make webui's `LayerGroupProvider` use streaming catalog access

  

# Checklist


groldan:
@aaime help

In changing `LayerGroupProvider` to use the catalog streaming API instead of `getLayerGroups():List<>`,
[AbstractAdminPrivilegeTest.testLayerGroupAllPage()](https://github.com/geoserver/geoserver/blob/fb8e2330/src/web/core/src/test/java/org/geoserver/web/admin/AbstractAdminPrivilegeTest.java#L242-L264) is failing.

Turns out `Catalog.getLayerGroups()` and `Catalog.list(LayerGroupInfo.class, Filter.INCLUDE)` return different results.
Logged in as the `cite` test user, the former includes the global layer groups, but the later doesn't, and only includes the `cite` workspace's `cite_local` group.

For instance, one would expect the following to pass:
```
        AdminRequest.start(new Object());
        Set<String> withGet = cat.getLayerGroups().stream().map(LayerGroupInfo::getName).collect(Collectors.toCollection(TreeSet::new));

        AdminRequest.start(new Object());
        Set<String> withList = new TreeSet<>();
        try(CloseableIterator<LayerGroupInfo> it = cat.list(LayerGroupInfo.class, Filter.INCLUDE)){
        	while(it.hasNext()) withList.add(it.next().getName());
        }
        assertEquals(withGet, withList);
```

But we get this instead:
```
java.lang.AssertionError: expected:<[cite_global, cite_local, sf_global]> but was:<[cite_local]>
```

When calling `Catalog.list(LayerGroupInfo.class, Filter.INCLUDE)`, `SecureCatalogImpl` calls `CatalogFacade.list()` methods is called with the following filter:
```
[Filter.INCLUDE AND [ workspace.name = cite ] AND Filter.INCLUDE AND [ true = InternalFunctionImpl() ] AND Filter.INCLUDE]
```
But when calling `Catalog.getLayerGroups()`, such filter is never built. `SecureCatalogImpl` instead gets all layer groups and filters the list with
```
protected LayerGroupInfo checkAccess(
            Authentication user,
            LayerGroupInfo group,
            MixedModeBehavior mixedModeBehavior,
            List<LayerGroupInfo> containers)
```
which includes the global groups.

Now the question is which one is right, first. And whether this exposes a bug or am I missing something?

TIA.

aaime:
Scratch scratch... here are some quick thoughts.

The class is abstract and lets subclasses setup the security restrictions:

-  The AdminPrivilegeTest sets up "cite" as the workspace administrator of the same named workspace, while general read access is granted to all.
- ResourceAccessManagerAdminPrivilegeTest seems to set up a test resource access manager that would turn cite into a workspace admin as well, not sure what happens to general requests.

In terms of what the secured catalog would have to do, the answer is contextual:
* If the current request is an admin request (e.g., Wicket admin GUI or REST API) then it should only return the layer groups the user can administer. So the answer would be, I think, only the groups in the cite workspace.
* If the current request is a general OGC request, then the rules for general data access apply
  - if the request is using a local workspace, then only current workspace groups should be returned, I believe
  - If the request is on the global service instead, then all layer groups would apply.

That's what I believe the server should do... but mind, I've just jotted down some thoughts, could not invest time for a through investigation.

src/web/core/src/main/java/org/geoserver/web/data/layergroup/LayerGroupProvider.java
```diff
@@ -31,17 +57,11 @@ public class LayerGroupProvider extends GeoServerDataProvider<LayerGroupInfo> {
...
-                public Boolean getPropertyValue(LayerGroupInfo item) {
-                    return Boolean.valueOf(item.isEnabled());
-                }
-            };
+    public static Property<LayerGroupInfo> ENABLED = new BeanProperty<>("enabled", "enabled");
 
     static List<Property<LayerGroupInfo>> PROPERTIES = Arrays.asList(NAME, WORKSPACE, ENABLED);
 
+    /** revisit: this seems to be dead code, there're no uses of this interface in the code base */
     protected LayerGroupProviderFilter groupFilter = null;
```
aaime, line 65: This can indeed be removed, it was used by the WMS-EO extension, which was eventually demoted down to community module, and then removed altogether. It can be safely removed, and then you can simplify the code a bit.

src/web/core/src/main/java/org/geoserver/web/data/layergroup/LayerGroupProvider.java
```diff
@@ -51,18 +71,33 @@ public LayerGroupProvider(LayerGroupProviderFilter groupFilter) {
...
+        Filter filter = getFilter();
+        try (Stream<LayerGroupInfo> items = query(filter, (int) first, (int) count, sortOrder)) {
+            return items.collect(Collectors.toList()).iterator();
         }
-        return groups;
+    }
+
+    /**
+     * @throws UnsupportedOperationException, this method shouldn't be called at all due to the
+     *     overloading of {@link #size()}, {@link #fullSize()}, and {@link #iterator(long, long)}
```
aaime, line 94: Odd javadoc comment (as in, the description is not relevant to the exception), I'd simplify down to:

```suggestion
     * This method shouldn't be called at all due to the
     *     overloading of {@link #size()}, {@link #fullSize()}, and {@link #iterator(long, long)}
```

(we have no requirement to document exceptions in javadoc)

groldan: done

src/web/core/src/main/java/org/geoserver/web/data/layergroup/LayerGroupProvider.java
groldan: done

PR #7669, opened 2024-05-21
[GEOS-11408] XmlRequestReader clean generic XML request parsing exceptions

jodygarnett:
[GEOS-11408](https://osgeo-org.atlassian.net/browse/GEOS-11408) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

WCS does a nice job of creating a helpful message, rather than a generic IO exception. This PR makes a similar improvement for the other XmlRequestReaders.

# Checklist

This provides a small API change, an additional utility method in XmlRequestReader that the subclasses can use to process an exception prior to use in SecurityException constructor.

### DIFF
--- a/src/ows/src/main/java/org/geoserver/ows/XmlRequestReader.java
+++ b/src/ows/src/main/java/org/geoserver/ows/XmlRequestReader.java
@@ -5,12 +5,16 @@
  */
 package org.geoserver.ows;
 
+import java.io.IOException;
 import java.io.Reader;
 import java.util.Map;
+import java.util.logging.Logger;
 import javax.xml.namespace.QName;
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 import org.geotools.util.Version;
+import org.geotools.util.logging.Logging;
+import org.xml.sax.SAXException;
 
 /**
  * Creates a request bean from xml.
@@ -24,6 +28,11 @@ import org.geotools.util.Version;
  * @author Justin Deoliveira, The Open Planning Project, jdeolive@openplans.org
  */
 public abstract class XmlRequestReader {
+    /**
+     * Logger for XmlRequestReader subclass established in constructor for logging of parse errors.
+     */
+    final Logger LOGGER;
+
     /** the qualified name of the element this reader can read. */
     final QName element;
 
@@ -63,6 +72,8 @@ public abstract class XmlRequestReader {
         this.element = element;
         this.version = version;
         this.serviceId = serviceId;
+
+        LOGGER = Logging.getLogger(this.getClass());
 
         if (element == null) {
             throw new NullPointerException("element");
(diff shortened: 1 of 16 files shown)


src/ows/src/main/java/org/geoserver/ows/XmlRequestReader.java
```diff
@@ -126,4 +130,73 @@ public int hashCode() {
...
+     * <p>The full stack trace is preserved to support development and debugging.
+     *
+     * @param t Parsing failure
+     * @return Clean SAXException for common parsing failures, or initial throwable
+     */
+    protected Exception cleanException(Exception t) {
+        if (t instanceof IOException) {
+            // FileNotFoundException, ConnectionException, etc...
+            // Log actual failure for debugging and troubleshooting
+            final Logger LOGGER = Logging.getLogger(this.getClass());
```
davidblasby, line 150: (optional) I see LOGGER is only used here, but might make sense to put it in the class....

src/ows/src/main/java/org/geoserver/ows/XmlRequestReader.java
```diff
@@ -126,4 +130,73 @@ public int hashCode() {
...
+        Throwable cause = saxException.getCause();
+        // We only wish to clean the generic sax exception that recycles internal cause message
+        if (saxException != null
+                && saxException.getCause() != null
+                && saxException.getClass() == SAXException.class) {
+            String saxMessage = saxException.getMessage();
+            String causeMessage = cause.getLocalizedMessage();
+            if (causeMessage != null && saxMessage.contains(causeMessage)) {
+                // Log actual failure for debugging and troubleshooting
+                final Logger LOGGER = Logging.getLogger(this.getClass());
```
davidblasby, line 189: looks like you are creating LOGGER in 2 locations - might want to put it in the class.

src/ows/src/main/java/org/geoserver/ows/XmlRequestReader.java
```diff
@@ -126,4 +130,73 @@ public int hashCode() {
...
+    /**
+     * Wrap request parsing failure in a SAX Document to prevent generic FileNotFound and
+     * ConnectionRefused messages.
+     *
+     * <p>This communicates that there is a problem with the XML Document while avoiding providing
+     * internal details.
+     *
+     * <p>The full stack trace is preserved to support development and debugging.
+     *
+     * @param t Parsing failure
```
davidblasby, line 143: io or sax exception?

src/ows/src/main/java/org/geoserver/ows/XmlRequestReader.java
```diff
@@ -126,4 +130,73 @@ public int hashCode() {
...
+     *
+     * @param saxException
+     * @return saxException with nested localized message removed.
+     */
+    protected SAXException cleanSaxException(SAXException saxException) {
+        Throwable cause = saxException.getCause();
+        // We only wish to clean the generic sax exception that recycles internal cause message
+        if (saxException != null
+                && saxException.getCause() != null
+                && saxException.getClass() == SAXException.class) {
```
davidblasby, line 184: just wanted to make sure that you want to use "==" here - I think this will not allow for SAXException sub-classes.

src/wms/src/test/java/org/geoserver/wms/map/GetMapXmlReaderTest.java
```diff
@@ -132,6 +137,32 @@ public void testAllowDynamicStyles() throws Exception {
         }
     }
 
+    @Test
+    public void testCleanServiceException() throws Exception {
```
davidblasby, line 141: the clean exception code has 2 major paths - SAXException and IO Exception.  Should probably have test for both cases.  

Also, the SAXException cleaning might want to have more test cases (see above comment - "if" - in implementation)

src/ows/src/main/java/org/geoserver/ows/XmlRequestReader.java
```diff
@@ -126,4 +130,73 @@ public int hashCode() {
...
+            String saxMessage = saxException.getMessage();
+            String causeMessage = cause.getLocalizedMessage();
+            if (causeMessage != null && saxMessage.contains(causeMessage)) {
```
davidblasby, line 187: Just wanted to make sure this "if" is correct.  It looks a little funny since the exception isn't cleaned unless the saxMessage contains the causeMessage.

davidblasby (approved):

src/ows/src/main/java/org/geoserver/ows/XmlRequestReader.java
jodygarnett: I am actually looking up a specific logger from the subclass, so although it is unusual it is on purpose looking it up this way. Technically I could look it up in the constructor also.

src/ows/src/main/java/org/geoserver/ows/XmlRequestReader.java
jodygarnett: Yes that is exactly what I want, thanks for asking. 

I think my comment communicated that I only wish to check SAXException and not subclasses like SAXParseException?

src/ows/src/main/java/org/geoserver/ows/XmlRequestReader.java
jodygarnett: feedback addressed, created logger during constructor

src/ows/src/main/java/org/geoserver/ows/XmlRequestReader.java
jodygarnett: comment clarified

src/wms/src/test/java/org/geoserver/wms/map/GetMapXmlReaderTest.java
jodygarnett: Added test case for this to DispatcherTest which tests XmlRequestReader using a dummy service.

PR #7668, opened 2024-05-21
[GEOS-11398] Javadoc clarification and test cases for Entity Resolution

jodygarnett:
[GEOS-11398](https://osgeo-org.atlassian.net/browse/GEOS-11398) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

It is great with the 2.25.0 release people are using ENTITY_RESOLUTION_ALLOWLIST. This PR gathers some feedback and clarifications in response to public use.

# Checklist

- [x] I have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/main/CONTRIBUTING.md).
- [x] I have sent a [Contribution Licence Agreement](https://docs.geoserver.org/latest/en/developer/policies/committing.html) (not required for small changes, e.g., fixing typos in documentation).
- [x] First PR targets the `main` branch (backports managed later; ignore for branch specific issues).
- [ ] All the build checks are green ([see automated QA checks](https://docs.geoserver.org/latest/en/developer/qa-guide/index.html)).

For core and extension modules:

- [x] New unit tests have been added covering the changes.
- [x] [Documentation](https://github.com/geoserver/geoserver/tree/main/doc/en/user/source) has been updated (if change is visible to end users).
- [ ] The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers).
- [x] There is an issue in the [GeoServer Jira](https://osgeo-org.atlassian.net/browse/GEOS/summary) (except for changes that do not affect administrators or end users in any way).
- [x] Commit message(s) must be in the form ``[GEOS-XYZWV] Title of the Jira ticket``.
- [x] Bug fixes and small new features are presented as a single commit.
- [x] Each commit has a single objective (if there are multiple commits, each has a separate JIRA ticket describing its goal).

PR #7678, opened 2024-05-22
[Backport 2.24.x] [GEOS-11398] Javadoc clarification and test cases for Entity Resolution

jodygarnett:
[GEOS-11398](https://osgeo-org.atlassian.net/browse/GEOS-11398) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

Backport https://github.com/geoserver/geoserver/pull/7668

However since 2.24.x did not have a test case I have adapted the EntityResolverProviderTest to the expectations of 2.24.x branch (2.24.x it defaults to unrestricted http access).

### DIFF
--- a/src/main/src/test/java/org/geoserver/util/EntityResolverProviderTest.java
+++ b/src/main/src/test/java/org/geoserver/util/EntityResolverProviderTest.java
@@ -5,6 +5,7 @@
 package org.geoserver.util;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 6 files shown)


jodygarnett:
@davidblasby although this is a backport I would like to ask for a review; as the test case has been adapted to reflect 2.24.x

src/main/src/test/java/org/geoserver/util/EntityResolverProviderTest.java
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * Test behaviour of EntityResolveProvider in response to configuration.
+     *
+     * <p>EntityResolver returns {@code null} when the provided URI is *allowed*, Returns an
+     * Inputstream if the content is provided, or throws an Exception if the URI is not allowed.
+     */
+    @Test
+    public void testEntityResolverDefaultBehaviour() throws Exception {
+        EntityResolverProvider provider = new EntityResolverProvider(null);
+        provider.setEntityResolver(new AllowListEntityResolver(null));
```
davidblasby, line 80: This test case looks like its providing null (meaning accept everything) - but the test case is testing a dis-allow.

jodygarnett: sigh, okay! hold on ...

jodygarnett: @davidblasby I have added some additional checks (they were indeed useful and found problems); so it is worth forward porting the additional checks.

GeoServer is mostly test cases...

PR #7692, opened 2024-05-28
[GEOS-11414] Css unique rule names

sweco-se1c8x:
[GEOS-11414](https://osgeo-org.atlassian.net/browse/GEOS-11414) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

The goal of this Pull Request is to update the GeoServer documentation to include the newly added UniqueRuleNames directive for CSS styling. This directive allows users to assign unique names to their styling rules, which is particularly useful for generating accurate legends for complex map symbols. The documentation will provide users with the necessary information on how to implement and benefit from this new feature.

# Checklist


doc/en/user/source/styling/css/directives.rst
```diff
@@ -56,3 +57,8 @@ Supported directives
       * String
       * The generated SLD style abstract/description
       * No
+    - * ``uniqueRuleNames`` 
+      * Boolean
+      * If set to `'true'`, instructs the translator to generate unique SLD Rule Names. This may be desired if individual Rules from the generated SLD is later
```
aaime, line 62: This says nothing about how the names are generated, how they can be used, and what's the context/use case for this flag. Please clarify so that a random CSS user can figure out how to use this new flag.

aaime (changes requested):

doc/en/user/source/styling/css/directives.rst
sweco-se1c8x: I know that is not enough. I just did not know where to write more about the UniqueRoleName. I have never worked on open source earlier and i need little bit support with it,

I think i need to write more description on the cookbook right?

aaime: Here, we need to indicate what the flag actually does. Unique rule names is not useful, unique how? "Instructs the translator to give each generated SLD rule a unique name, as a progressive number starting from zero" is probably enough.

The cookbook focuses on how to achieve a given visual output, not a good place to talk about this directive.

I would add a description in the directives pages itself, as a small separate paragraph, with a description of the use case, how does one leverage the numbering, in which condition is the numbering predictable and so on. If having a predictable numbering is not even necessary, then clarify what's the intended usage, and how is the client interacting with the server to make use of it.

Grammostola: @aaime Hello, I'm part of the organization sponsoring this PR (and the corresponding one for geotools). The directive description is now revised and pretty succinct, would you like for us to flesh it out further? 

`autoRuleNames` or `addRuleNames` may have been slightly more logical. Anyway all we are looking to do is to make CSS styles slightly more equal to SLD styles, by providing them with a bit of functionality only SLD styles currently have. 

Which in itself may not make immediate sense to someone who doesn't employ this functionality and have only written CSS styles. So do advice if the currently proposed documentation appears insufficient : )

aaime: Yes, having the description provided by @steff-o in the [GeoTools PR](https://github.com/geotools/geotools/pull/4786#issuecomment-2149050609), slightly reworded for generality, and exposed as an example, is what I was looking for.

I also like "autoRuleNames" as the name, but it's up to you, if you want to change it, or not.

Grammostola: Understood. The PR has been revised.

PR #7700, opened 2024-05-30
[GEOS-11422] MapML License Metadata Stored With Incorrect Keys

turingtestfail:
[GEOS-11422](https://osgeo-org.atlassian.net/browse/GEOS-11422) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

  

# Checklist

### DIFF
--- a/src/extension/mapml/src/test/java/org/geoserver/mapml/MapMLWMSTest.java
+++ b/src/extension/mapml/src/test/java/org/geoserver/mapml/MapMLWMSTest.java
@@ -1647,6 +1647,47 @@ public class MapMLWMSTest extends MapMLTestSupport {
                 alternateLinks, ProjType.CBMTILE, new Envelope(-8.1E6, 8.3E6, -3.6E6, 1.23E7), 1e5);
     }
 
+    @Test
+    public void testLicenseIsPopulated() throws Exception {
+        // set up mapml layer license data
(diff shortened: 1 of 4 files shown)


src/extension/mapml/src/test/java/org/geoserver/mapml/MapMLWMSTest.java
```diff
@@ -1590,6 +1590,47 @@ public void testLargeBounds() throws Exception {
                 alternateLinks, ProjType.CBMTILE, new Envelope(-8.1E6, 8.3E6, -3.6E6, 1.23E7), 1e5);
     }
 
+    @Test
+    public void testLicenseIsPopulated() throws Exception {
```
aaime, line 1594: Good, but I think this test would have passed even before the fix... (or not?).
What I saw was the MapMLLayerConfig entering the wrong keys in the metadata map, which were then not used by the builder. I believe the actual fix is the change in the constant definitions, although consistent usage of contacts across the code is a very welcomed plus.

turingtestfail: Good point.  I don't want to create a test that validates a specific key value in case there is a future change but I think this is the next best thing.

PR #7704, opened 2024-06-01
[GEOS-11424] Speed up web-ui WorkspaceAdminComponentAuthorizer

groldan:
[GEOS-11424](https://osgeo-org.atlassian.net/browse/GEOS-11424) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

Logging in as a user other than `admin`, and having a large amount of workspaces results in potentlially substantial delays in loading almost every webui wicket page, as `WorkspaceAdminComponentAuthorizer` will be run several times, each of them potentially doing a full scan of workspaces, to determine if the user has admin privileges in at least one workspace.

This patch addresses the long-standing TODO comment in `WorkspaceAdminComponentAuthorizer` by caching the result in the `RequestAttributes`, hence performing the check once per request.

Also, moves the logic to a new default method in the `ResourceAccessManager` interface, giving implementations a chance to provide a more efficient check.

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/ResourceAccessManager.java
+++ b/src/main/src/main/java/org/geoserver/security/ResourceAccessManager.java
@@ -7,12 +7,14 @@ package org.geoserver.security;
 
 import java.util.List;
 import javax.annotation.Nullable;
+import org.geoserver.catalog.Catalog;
 import org.geoserver.catalog.CatalogInfo;
 import org.geoserver.catalog.LayerGroupInfo;
 import org.geoserver.catalog.LayerInfo;
 import org.geoserver.catalog.ResourceInfo;
 import org.geoserver.catalog.StyleInfo;
 import org.geoserver.catalog.WorkspaceInfo;
+import org.geoserver.catalog.util.CloseableIterator;
 import org.geotools.api.filter.Filter;
 import org.springframework.security.core.Authentication;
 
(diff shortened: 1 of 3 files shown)


src/main/src/main/java/org/geoserver/security/ResourceAccessManager.java
```diff
@@ -71,4 +71,7 @@ public default LayerGroupAccessLimits getAccessLimits(
      */
     public @Nullable Filter getSecurityFilter(
             Authentication user, final Class<? extends CatalogInfo> clazz);
+
+    /** Checks if {@code user} has admin privileges on at least one workspace */
+    public boolean isWorkspaceAdmin(Authentication user);
```
aaime, line 76: Can  you please make this a default method instead? 

In my work experience, ResourceAccessManager is one of the most commonly implemented extension points, typically to leverage installation specific rule management and/or to integrate with existing systems.
Implementations typically have no reason to implement the Abtract class, as it's mostly empty methods, and has been used only in tests.

src/web/core/src/test/java/org/geoserver/web/WorkspaceAdminComponentAuthorizerTest.java
```diff
@@ -0,0 +1,132 @@
+package org.geoserver.web;
```
aaime, line 1: Misses copyright header

aaime (changes requested):
Overall looks good, thanks. See feedback below.

src/main/src/main/java/org/geoserver/security/ResourceAccessManager.java
groldan: That was my first instinct as well, but then I thought it'd be weird that the method received the Catalog as parameter.
It'd have to be 
```java
default boolean isWorkspaceAdmin(Authentication user, Catalog catalog){...}
```
where no other method in ResourceAccessManager requires a catalog as argument.
If that's still good enough I can go for it, but had to mention it just in case you didn't notice

aaime: I see what you mean. On the other side, DefaultResourceAccessManager haven't had a reference to catalog until 2016, when container layer group mode was introduced (a group mode that is not used often). The typical implementation of ResourceAccessManager out there won't necessarily have access to a catalog... checking ones I have handy, none of them do... as site-specific implementations, they don't care about group nesting to start with.

Anyhow, personally I'd be ok with the current implementation as well, as long as you [update the upgrade instructions](https://docs.geoserver.org/latest/en/user/installation/upgrade.html#notes-on-upgrading-specific-versions) mentioning the API breakage for existing ResourceAccessManager implementations.

groldan: with that in mind, I actually prefer the default method on the interface. Honestly I thought you were going to oppose to that lol, as that was my first approach

groldan: Okay, moved the method as a default one on the interface, and removed the need to extend AbstractResourceAccessManager altogether.

aaime (approved):
LGTM

PR #7710, opened 2024-06-04
Community assembly rework, split COG module into "per cloud" zip packages

aaime:
The current community module packaging is sometimes failing to deliver working ZIP files due to incompatible dependencies. The packaging is currently done this way:

* A "release" module depends on all the modules that want to be packages
* The release module dumps all these module dependencies into a single directory
* Each module packaging descriptor picks from that pile of jars using globs

This PR reworks the assembly so that each module is responsible for assembling itself, using an assembly descriptor placed in the standard ``src/assembly/assembly.xml" location, based solely on its own dependencies.

When a module is split into sub-modules, and the assemblies need to put together multiple sub-modules without a single "catch-all" module that could be used to gather the dependencies, a dedicated "assembly" module is created.
Examples of this are "backup-restore", but also the OGC API modules, where each service module also needs a shared web-ogcapi module, that cannot be included as a dependency, as that would cause a circularity: for this last case, a dedicated assembly module has been created for each OGC service.

The COG module has also been split into target specific ZIPs, one for plain HTTP reads, and others for AWS, Azure and Google.

As a nice side effect of this new approach, it's now possible to work on the assembly of a single module, or group of modules, e.g., to only assemble the COG modules one can run:

```
mvn clean install -Pcog,assembly -T1C -DskipTests -nsu
```

or even just get into a single COG module, and run assembly, like this:

```
cd cog/cog-azure
mvn clean install -Passembly -T1C -DskipTests -nsu
```

and in general, to assemble everything, the following can be done:

```
mvn clean install -PcommunityRelease,assembly -T1C -DskipTests -nsu
```

The assembly instructions, once part of the "community/release" module, have been moved into the community module itself , ``community/README.md``.

In terms of package testing, I've manually tested all of the COG zips, and run a comparison of the files found in the other zip files, enumerating for each zip which files are inside, and performing a diff. Results are found here:

[comparison.txt](https://github.com/user-attachments/files/15567668/comparison.txt)

As far as I can tell, the differences are due to extra files that were provided by other dependencies and version differences.

# Checklist


aaime:
Phepraps I've found a way to avoid the skipAssembly bit: ``ignoreMissingDescriptor`` set to true. Needs to be tested.

jodygarnett:
I will wait to hear back about ignoreMissingDescriptor - it does sound more elegant.

doc/en/user/source/community/cog/cog.rst
```diff
@@ -13,7 +13,15 @@ As a community module, the package needs to be downloaded from the `nightly buil
...
-To install the module, unpack the zip file contents into the GeoServer ``WEB-INF/lib`` directory and restart GeoServer.
+The module is provided as four different packages, depending on the connection type:
+
+* ``cog-http`` for HTTP connections without a particular cloud target
+* ``cog-aws`` for AWS S3 connections
+* ``cog-google`` for Google Cloud Storage connections
+* ``cog-azure`` for Azure Blob Storage connections
+
+To install the module, download the package you need, unpack the zip file contents into the GeoServer ``WEB-INF/lib`` directory and restart GeoServer.
+Mind, the dependencies of the various cloud providers might interfere with each other, so please never install toghether the AWS, Google and Azure modules.
```
jodygarnett, line 24: Spelling mistake and caution:

```suggestion

.. warning::

   The dependencies of the various cloud providers might interfere with each other, so please never install together the AWS, Google and Azure modules.
```

src/community/backup-restore/assembly/src/assembly/assembly.xml
```diff
@@ -6,7 +6,7 @@
   <includeBaseDirectory>false</includeBaseDirectory>
   <fileSets>
     <fileSet>
-      <directory>release/target/dependency</directory>
+      <directory>target/dependency</directory>
```
jodygarnett, line 9: I love how minimal the change is, a rename and adjusting path.

src/community/cog/cog-core/src/main/java/org/geoserver/cog/CogSettings.java
```diff
@@ -32,31 +28,31 @@ public enum RangeReaderType {
...
+                return "it.geosolutions.imageioimpl.plugins.cog.GSRangeReader";
             }
         },
         /** Reader using the Azure API (again, with security support) */
         Azure {
 
             @Override
             String getRangeReaderClassName() {
-                return AzureRangeReader.class.getName();
+                return "it.geosolutions.imageioimpl.plugins.cog.AzureRangeReader";
```
jodygarnett, line 55: Nice way to avoid forcing azure load.

src/community/cog/cog-google/src/assembly/assembly.xml
```diff
@@ -0,0 +1,31 @@
+<assembly>
```
jodygarnett, line 1: Just a question, pom.xml has header, but assembly.xml does not? I assume it was just that way already.

I personally do not care either way as these files do not form part of the running application..

src/community/pom.xml
```diff
@@ -28,83 +28,21 @@
...
-            <descriptor>release/ext-wps-longitudinal-profile.xml</descriptor>
-            <descriptor>release/ext-monitor-kafka.xml</descriptor>
-            <descriptor>release/ext-graticule.xml</descriptor>
-            <descriptor>release/ext-wfs-freemarker.xml</descriptor>
-          </descriptors>
-          <finalName>geoserver-${project.version}</finalName>
-        </configuration>
+        <groupId>org.commonjava.maven.plugins</groupId>
+        <artifactId>directory-maven-plugin</artifactId>
+        <version>0.3.1</version>
```
jodygarnett, line 33: Move version to root pom plugin management section.

aside: It looks like version 1.0 is available as of Sep 22, 2021 which prevents parallel build warning.

.github/workflows/assembly.yml
```diff
@@ -40,12 +40,9 @@ jobs:
...
       run: |
         mvn -B -ntp -nsu -N -f src/pom.xml assembly:single
-    - name: Build community modules without tests
+    - name: Build and package community modules (without tests)
       run: |
-        mvn -B -ntp -U -T3 -DskipTests -PcommunityRelease -f src/community/pom.xml install
-    - name: Package community modules
-      run:  |
-        mvn -B -ntp -nsu -N -f src/community/pom.xml assembly:single
+        mvn -B -ntp -U -T3 -DskipTests -PcommunityRelease,assembly -f src/community/pom.xml install
```
jodygarnett, line 45: Similar change needed for `build/publish_release.sh`

jodygarnett (changes requested):
Some feedback, you already know from the meeting. I have now reviewed the pull request and when this feedback for `build/publish_release.sh` is addressed this change can be merged and back ported.

I really like the work and would love to see this approach used for the entire project.

.github/workflows/assembly.yml
aaime: Checked, it actually does not need it, as community modules are not assembled for a release.
However, the nightly build jenkins scripts will have to be updated on merge.

src/community/cog/cog-google/src/assembly/assembly.xml
aaime: Yeah, none of the assembly files, neither in community nor in extension, have copyright headers so far.

src/community/pom.xml
aaime: copied over from main... but yes, testing centralizing in dependency management and upgrading

aaime:
Used antrun to check if the assembly files are present, so that we can skip both copy-dependencies and assembly. 
Just using the ignore flag was pushing the overall build time from 43 seconds to 63 seconds, as the copy dependencies was run on every module .

Applied the other reviewer feedback as well.

jodygarnett:
> Phepraps I've found a way to avoid the skipAssembly bit: `ignoreMissingDescriptor` set to true. Needs to be tested.

I do not see ignoreMissingDescriptor used yet; is this an experiment you are still considering or is the PR ready to be merged?

aaime:
Read again my comment. I tried it out, and made the build slower. Went for a better way that does not require the support property to be defined in each module, instead, an antrun will detect the presence of the assembly file in the module, and decide whether to copy resources and assembly, or not.

jodygarnett (approved):
Thanks Andrea, my feedback has been addressed and I love the change.

That was very tricky using ant-run to check for a file :P

aaime:
Ok, let's merge and see what the build server thinks about it

jodygarnett:
> Ok, let's merge and see what the build server thinks about it

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.22.2:test (default-test) on project gs-wms: There are test failures.
```

Huh? Only thing I see is:
```
[ERROR] Tests run: 32, Failures: 1, Errors: 0, Skipped: 1, Time elapsed: 1.09 s <<< FAILURE! - in org.geoserver.wms.wms_1_1_1.DimensionsVectorGetFeatureInfoTest
[ERROR] testSortTimeElevationDescending(org.geoserver.wms.wms_1_1_1.DimensionsVectorGetFeatureInfoTest)  Time elapsed: 0.016 s  <<< FAILURE!
org.junit.ComparisonFailure: expected:<application/vnd.ogc.[gml]> but was:<application/vnd.ogc.[se_xml;charset=UTF-8]>
	at org.geoserver.wms.wms_1_1_1.DimensionsVectorGetFeatureInfoTest.getFeatureAt(DimensionsVectorGetFeatureInfoTest.java:98)
	at org.geoserver.wms.wms_1_1_1.DimensionsVectorGetFeatureInfoTest.testSortTimeElevationDescending(DimensionsVectorGetFeatureInfoTest.java:496)
```

PR #7716, opened 2024-06-11
Update GS Homepage for OGCAPI ServiceDescriptors

davidblasby:
See [GSIP-202](https://github.com/geoserver/geoserver/wiki/GSIP-202) - Welcome Page Layout

This PR added support (GS Home Page) for listing the OGCServices (under the correct heading) as well as links to them (including WS-based links and Layer-based links).

### New Home Page Look

Notice the OGCAPI-* items:

image

### Important Code Changes

The way the current ServiceInfoCapabilitiesProvider infrastructure used to work couldn't quite support this.  A lot of this was because there are now multiple services (i.e. WFS and FeaturesService) to be grouped under the "WFS" heading.

1. There is now ServiceDescription#descriptionPriority.  This is for grouping multiple services under a heading (i.e. "GeoServer Web Map Service".  If there are multiple services (having different names), this allows the user to say which title is used. Right now its set to the main (i.e. WFS/WMS/WCS) title.  
2. There is a also a "specificServiceType" tag - this allows for differentiation with the category (i.e. WFS vs FeaturesService)
3. I added a `OgcApiServiceDescriptionProvider` - you'll see each of the OGC Service Providers just quickly subclass this.
4. I've added a few test cases, however, I cannot run the current test cases on M1 mac and some of them are failing.
  

# Checklist


jodygarnett:
Other than the API messy which is something to work on ... I love the result :)

Comparing with GSIP-202 I was hoping to put the OGCAPI features at the end - and make Open Web Services and OGCAPI services visually distinct. 

image

1) I think we can sort the OGCAPI links to the end as shown easily enough.

2) I see a visual inconsistency - "Open Web Services" is not shown for WFS, WMS, WCS ... but "OGCAPI-" prefix is taking up a lot of visual weight on the screen (especially all in large capitals).

We should talk to @MichelGabriel for ideas here :)
- Can we drop the prefix from "OGCAPI-Features" / "1.0.1" and just have "Features" / "1.0.1" ?
- Icon or colour to differentiate between open web services and ogcapi services?
- Could have three levels "OGCAPI" (small gray) / "Features" / "1.0.1"
- Ideas?

aaime:
Feedback based purely on this screenshot:

> image

Some of the categorization is problematic:
* DGGS is nothing like tiles, it's a mix in between features and coverages that does not fit under the existing organization.
* Images is nothing like coverages, it's like a mosaic REST management API (as such, it should be secured when installed). It's experimental, part of a testbed and does not have its own spec repo, eventually we should just remove it. Should also not go under the existing organization.
* Styles is something completely new in the OGC services landscape, it has an affinity with mapping, but it's meant mostly for client-side mapping, the typical use case is solving the problem of finding a suitable style for a vector tiles dataset, but could be used also to style GeoJSON fetched from OGC API features. It has the thinnest relationship with WMS, although it can also be used to update existing styles and create new ones (when installed, it should always be covered with service security, much like WFS-T). Overall it should be in its own separate category, or thrown in a  catch-all one.

Following @jodygarnett feedback, OGC APIs are versioned, the version number should be made visible.
Visual aspect-wise, I'm wondering if we should indicate the API reliability in terms of:

* Fully stable (every implemented bit has a stable implementation)
* Core stable but with draft extensions and/or vendor bits (maybe a warning icon with a popup holding a list of aspects that are subject to change)
* Core in draft or implementation not  yet matching a stable release

jodygarnett:
For a short-term fix change to "OGCAPI-Features" to "Features". Consider "OGCAPI" a technical detail not needed in the user interface.

davidblasby:
@aaime - I can un-categorize those 3 ogcapi items out of their W* categories and put them in their own categories.
@jodygarnett  - I can remove the "OGCAPI-" part.  I put it in because they are new and people might ask "What's the different between Web Feature Service and Features?"

We can add icons - but I'm not sure which ones to add to which service. I think its a good idea, but it might make things a bit more messy on the screen.

jodygarnett:
> * Images is nothing like coverages, it's like a mosaic REST management API (as such, it should be secured when installed). It's experimental, part of a testbed and does not have its own spec repo, eventually we should just remove it. Should also not go under the existing organization.

Do I understand that this service should be secured with:
```
Images.*.    ADMIN
```
Kind of like we secure WFS.Transaction by default...

aaime:
> Do I understand that this service should be secured with:
> 
> ```
> Images.*.    ADMIN
> ```
> 
> Kind of like we secure WFS.Transaction by default...

Yes, and write operations in Styles as well.

jodygarnett:
Andrea for the services having their own ServiceInfo - do you have plans to make an ogcapi-web module to allow these to be configured? It is something GeoCat could work on if it is part of the plan, and would provide a better location for these service descriptors to avoid adding web dependencies to each module

src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/OgcApiServiceDescriptionProvider.java
```diff
@@ -40,8 +40,8 @@ public class OgcApiServiceDescriptionProvider<SERVICEINFOTYPE extends ServiceInf
      */
     String serviceName;
 
-    Class<ServiceInfo> infoClass;
-    Class serviceClass;
+    Class<SERVICEINFOTYPE> infoClass;
```
jodygarnett, line 43: Oh wild, I have never seen such a long generic identifier before usually they are just a single letter...

jodygarnett:
Looking up headings from OGCAPI website

> Discrete Global Grid Systems
> A DGGS is a spatial reference system that uses a hierarchical tessellation of cells to partition and address the globe.

jodygarnett:
For the experiments how about their heading:

> **OGCAPI  Testbed Experiments**
> The following modules are part of the OGC Testbed experiment

Based on user guide [description](https://docs.geoserver.org/2.25.x/en/user/community/ogc-api/testbed.html).

davidblasby:
I worked with jody and this is the new version;

image

src/community/ogcapi/ogcapi-images/src/main/java/org/geoserver/ogcapi/v1/images/ImagesServiceXStreamLoader.java
```diff
@@ -20,7 +20,10 @@ public ImagesServiceXStreamLoader(GeoServerResourceLoader resourceLoader) {
     protected ImagesServiceInfo createServiceFromScratch(GeoServer gs) {
         ImagesServiceInfoImpl info = new ImagesServiceInfoImpl();
         info.setName("images");
-        info.setTitle("Image mosaicks discovery and management interface");
+        if (info.getTitle() == null) {
+            info.setTitle("Testbed Experiments");
+            info.setAbstract("The following modules are part of the OGCAPI Testbed experiments.");
```
jodygarnett, line 25:
```suggestion
            info.setAbstract("The following services are part of the OGCAPI Testbed experiments.");
```

jodygarnett:
Okay I was looking at why Images service is still listed even when not logged in: [ServicesPanel.populateItem(ListItem<ServiceDescription> listItem)](https://github.com/geoserver/geoserver/blob/main/src/web/core/src/main/java/org/geoserver/web/ServicesPanel.java#L69) checks:

```
boolean enabled = service.isAdmin() ? admin : true;
```

So this is not something to fix in the ServicesDescriptorProvider; it is a bug to fix (in another PR) for [ImagesService](https://github.com/geoserver/geoserver/blob/main/src/community/ogcapi/ogcapi-images/src/main/java/org/geoserver/ogcapi/v1/images/ImagesService.java).

jodygarnett (approved):
All my feedback has been addressed, thank you.

While I am a little uncomfortable with ``SERVICEINFOTYPE`` for generic type it makes it past the style checker so it must be okay :)

src/community/ogcapi/dggs/ogcapi-dggs/src/test/java/org/geoserver/ogcapi/v1/dggs/DGGSServiceDescriptionProviderTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /** tests the links - esp the actual link when its WS-based of WS-and-layer based* */
+    @Test
+    public void testLinks() {
+        var provider = new DGGSServiceDescriptionProvider(this.getGeoServer());
+
+        var links = provider.getServiceLinks(null, null);
+        assertEquals(1, links.size());
+        assertEquals("WMTS", links.get(0).getServiceType());
+        assertEquals("DGGS", links.get(0).getSpecificServiceType());
```
jodygarnett, line 35: Q: Is SpecificServiceType still needed now that Andrea has moved DGGS into its own heading (not part of WMTS?)

src/web/core/src/main/java/org/geoserver/web/ServiceLinkDescription.java
```diff
@@ -83,6 +104,10 @@ public String getProtocol() {
         return protocol;
     }
 
+    public String getSpecificServiceType() {
```
jodygarnett, line 107: Tying to document the new term as it is not clear to me how this differs from protocol:
```suggestion
    /**
      * While serviceType used establish a group with title and description, the specific service type can identify
      * identify the service type used for configuration.
      */
    public String getSpecificServiceType() {
```
```suggestion
    public String getSpecificServiceType() {
```
```suggestion
    public String getSpecificServiceType() {
```
```suggestion
    public String getSpecificServiceType() {
```

src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/OgcApiServiceDescriptionProvider.java
```diff
@@ -0,0 +1,193 @@
...
+     */
+    String serviceName;
+
+    Class<SERVICEINFOTYPE> infoClass;
+    Class<SERVICETYPE> serviceClass;
+
+    GeoServer geoserver;
+    Catalog catalog;
+
+    public OgcApiServiceDescriptionProvider(
```
jodygarnett, line 49:
```suggestion
    /**
     * OGCAPI Service Descriptor with additional information to group with associated Open Web Service heading.
     * @param gs GeoServer configuration
     * @param serviceType Service identifier, example {@code WPS}, used to group for heading and description
     * @param serviceName OGCAPI Name
     * @param specificServiceType OGCAPI specific service type
     */
    public OgcApiServiceDescriptionProvider(
```

davidblasby:
> While I am a little uncomfortable with `SERVICEINFOTYPE` for generic type it makes it past the style checker so it must be okay :)

I think this is ok - I usually use "T" when the type is unimportant and there's just one (like a List<T>).

src/community/ogcapi/dggs/ogcapi-dggs/src/test/java/org/geoserver/ogcapi/v1/dggs/DGGSServiceDescriptionProviderTest.java
davidblasby: I think it is still needed - there's can be more services added later.  I think its needed for the grouped ones (like Features).

src/web/core/src/main/java/org/geoserver/web/ServiceLinkDescription.java
davidblasby: Protocol is what is actually put on the screen - i.e. OGCAPI-Features in my first picture, and just "Features" in my second.

aaime:
@jodygarnett I did had a plan to have separate service infos for services that offer the same functionality.. eventually I would love to see a core module that offers the internal machinery, using APIs that are not EMF based, and then two dependencies, one for classic OGC services, one for OGC API, that one might decide to include, or not.

I imagine the following:
* Short term, less than 2 years hopefully (spec and funding pending), all the OGC APIs equivalent to OGC services have a stable spec and a stable implementation, available as extension
* At some point, OGC APIs are included in core
* Some time later (5-7 years in the future?), classic OGC services are moved to extension, still available and supported, but no longer core
* Further in the future, 10+ years, the classic OGC service implementations are deprecated and moved to community, to be eventually removed

The timeline might seem long to some, but I have doubts we'll be able to make it that quick, considering the speed at which specs are moving, general lack of funding towards them, the simple fact that the modules have a single maintainer so far (a core module should have a bit more distributed knowledge), and the times large organizations need to perform upgrades in general.

src/community/ogcapi/ogcapi-core/pom.xml
```diff
@@ -172,6 +172,10 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
aaime, line 175: Oh dear, this is insanely bad. It means OGC APIs cannot be deployed anymore without a GUI in GeoServer.
Or, in other words, that GeoServer cannot be deployed anymore in secure environments where the GUI and REST APIs are banned (and need to be removed), as well as causing issues in environments were GUI is separate from services (e.g., GeoServer cloud). 

We have a web-ogcapi module fo a reason, please rework the design for this.

aaime (commented):
I was looking and it seemed ok, then noticed the solution is fundamentally at odds with how GeoServer modular system is designed. Hard -1 on merging it as is.

aaime:
Also, please, make sure test pass and QA checks work.

I've been trying to keep these modules clean, to keep them ready for graduation. I've checked and there was one test failing, and 3 deprecated method usages, which I've fixed here: https://github.com/geoserver/geoserver/pull/7727

Please make sure to rebase and keep the modules as clean as you've found them.

jodygarnett:
Sounds good, I think breaking out a single ui module should bed okay, service descriptors can check if their associated class is available. 

The service descriptors were also setup not to have such a direct connection with services. 

Thanks for the thoughts on service info, do you see any use for the title and description information on each ogcapi service?

jodygarnett:
Hey @aaime thanks for some your feedback/discussion above it is helpful to understand some of the background/ roadmap expectations.

aaime:
@jodygarnett you should know by now there is no roadmap, it all depends on what clients ask us to work on.
All the control we have is the power of "no" (AKA "we are not going to do this, or at least, not this way").

src/web/app/pom.xml
```diff
@@ -1625,6 +1625,38 @@
...
+        <dependency>
+          <groupId>org.geoserver.community</groupId>
+          <artifactId>gs-web-styles</artifactId>
+          <version>${project.version}</version>
+        </dependency>
+        <dependency>
+          <groupId>org.geoserver.community</groupId>
+          <artifactId>gs-web-tiles</artifactId>
+          <version>${project.version}</version>
+        </dependency>
```
aaime, line 1658: This is probably a bad idea, considering that OGC API modules are packaged per service, e.g.:

```
> unzip -t geoserver-2.26-SNAPSHOT-ogcapi-features-plugin.zip 
Archive:  geoserver-2.26-SNAPSHOT-ogcapi-features-plugin.zip
    testing: gs-web-ogcapi-2.26-SNAPSHOT.jar   OK
    testing: snakeyaml-2.2.jar        OK
    testing: validation-api-1.1.0.Final.jar   OK
    testing: jackson-dataformat-yaml-2.17.1.jar   OK
    testing: swagger-core-2.0.2.jar   OK
    testing: swagger-models-2.0.2.jar   OK
    testing: gs-ogcapi-features-2.26-SNAPSHOT.jar   OK
    testing: gs-ogcapi-core-2.26-SNAPSHOT.jar   OK
    testing: gt-cql2-text-32-SNAPSHOT.jar   OK
    testing: jackson-dataformat-xml-2.17.1.jar   OK
    testing: GPL.html                 OK
```

This way, installing the features module alone would bring in all the descriptors for all the services. Instead, go and modify [the <service>-assembly modules](https://github.com/geoserver/geoserver/tree/main/src/community/ogcapi) to include the dependency, and eventually the assembly.xml file as well (if needed) to make sure the generated zip file has the base web module, along with the service specific one (instructions on how to use the assembly can be found in the [community README file](https://github.com/geoserver/geoserver/blob/main/src/community/README.md))

aaime (changes requested):
Like the direction, the implementation as-is is likely going to cause installation problems, see below.

src/community/ogcapi/web-tiles/delme
```diff
@@ -0,0 +1,5 @@
+[INFO] Scanning for projects...
```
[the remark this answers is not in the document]
davidblasby: ;)

src/web/app/pom.xml
davidblasby: ok - so installing `features`.zip would also install the `web-features`.
I will look into that today/tomorrow.

aaime:
Had another quick spin, testing the results of the assembly, they look good to me.

aaime (approved):

jodygarnett:
> Each service exposes title and description, so of course, I see a use. However, in my mind, it's a minor detail in the economy of completing the OGC API offering and making it graduate... but if it's something you'd like to see happen, go for it.

Yeah I can see how it would be viewed as a minor technical detail. It does highlight a difference in architectural / user experience vision.

I am trying to set GeoServer to publish "Vector data" and then have "OWS-WFS" and "OGCAPI-Features" being peers - just different protocols that are available. There may be some small settings to configure for each one; just the same way we have some small settings for specific WFS versions...

davidblasby:
I think this is ok to squash-and-merge - I did some testing and its working as expected.
@jodygarnett

PR #7730, opened 2024-06-19
[GEOS-11111] Open search for EO community module: STAC search page has wrong self link

aaime:
See ticket for details

# Checklist

### DIFF
--- a/src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/PaginationLinksBuilder.java
+++ b/src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/PaginationLinksBuilder.java
@@ -7,7 +7,6 @@ package org.geoserver.ogcapi;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Optional;
-import javax.servlet.http.HttpServletRequest;
 import org.geoserver.ows.URLMangler;
 import org.geoserver.ows.util.ResponseUtils;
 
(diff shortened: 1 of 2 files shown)


src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/PaginationLinksBuilder.java
```diff
@@ -82,12 +78,12 @@ public Map<String, Object> getNextMap(boolean includeQueryMap) {
...
     }
 
+    /** Returns a HREF to the current page. */
     public String getSelf() {
-        HttpServletRequest request = APIRequestInfo.get().getRequest();
-        StringBuffer sb = request.getRequestURL();
-        String queryString = request.getQueryString();
-        if (queryString != null && !queryString.isEmpty()) sb.append("?").append(queryString);
-        return sb.toString();
+        APIRequestInfo info = APIRequestInfo.get();
```
turingtestfail, line 83: This never gets used?

turingtestfail (changes requested):

turingtestfail (approved):

src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/PaginationLinksBuilder.java
aaime: Good point, removing

PR #7746, opened 2024-06-21
[GEOS-11443] REST API does not take effect immediately due to 10 minute authentication cache

petersmythe:
[GEOS-11443](https://osgeo-org.atlassian.net/browse/GEOS-11443) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

From my experience with the GeoServer REST API, e.g. adding a user to a group [GeoServer API Docs](https://docs.geoserver.org/stable/en/api/#1.0.0/usergroup.yaml) , the change happens and is written to disk immediately, however there is a 10 minute authentication cache that prevents the change from being **effective** immediately.

If I am impatient, I typically reload the configuration or use the AuthKey synchronise button to clear the 10 min auth cache. I am aware of the reload API endpoint [GeoServer API Docs](https://docs.geoserver.org/stable/en/api/#1.0.0/reload.yaml)  but that is a bit drastic.  
The reset endpoint does not appear to clear the auth cache.

I have found that adding this bean to `src/main/java/applicationSecurityContext.xml` fixes the problem

`<bean id="AuthenticationCacheImpl" class="org.geoserver.security.auth.AuthenticationCacheImpl" />`

but I am unsure of the side effects of this.  Could I please get a review @aaime @jodygarnett 

# Checklist

### DIFF
--- a/src/main/src/main/java/applicationSecurityContext.xml
+++ b/src/main/src/main/java/applicationSecurityContext.xml
@@ -182,4 +182,7 @@
   <bean id="styleUrlChecker" class="org.geoserver.security.urlchecks.StyleURLChecker">
     <constructor-arg ref="dataDirectory"/>
   </bean>
+
+  <bean id="AuthenticationCacheImpl" class="org.geoserver.security.auth.AuthenticationCacheImpl" />
+
 </beans>
--- a/src/main/src/main/java/org/geoserver/security/auth/AuthenticationCache.java
+++ b/src/main/src/main/java/org/geoserver/security/auth/AuthenticationCache.java
@@ -27,7 +27,7 @@ public interface AuthenticationCache {
     /** Clears all cache entries for filterName */
     public void removeAll(String filterName);
 
-    /** Clears a specific chache entry */
+    /** Clears a specific cache entry */
     public void remove(String filterName, String cacheKey);
 
     /** */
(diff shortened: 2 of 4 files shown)


petersmythe:
Sample REST call:

`POST/DELETE http://localhost:8080/geoserver/rest/security/usergroup/service/<my_user_group_service>/user/<username>/group/<GRP_name>`

petersmythe:
Hmm, the `gs-security-tests` are now failing with:

`org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'filterChainProxy' defined in URL [file:src/main/target/classes/applicationSecurityContext.xml]: Invocation of init method failed; nested exception is org.geoserver.platform.GeoServerExtensions$MultipleBeansException: Multiple beans of type org.geoserver.security.auth.AuthenticationCache`

Clearly I do not understand what I am doing!

aaime:
Mind, as said previously, I'm not an expert in this area, that said... see below.

The default cache implementation is the GuavaAuthenticationCache, that has a cleanup time of 10 minutes indeed.
What you're trying to do is to inject a dummy authentication cache, that is, no caching at all... which would be pretty bad for expensive authentication methods, every single request would have to perform again the database queries/LDAP lookups/CAS Interaction/Keycloack interaction needed to re-validate the authentication from scratch.

That said, 10 minutes are a long time...  maybe reduce the default values, and/or try to use the "remove" method when a change to the auth happens through GeoServer itself?

petersmythe:
> What you're trying to do is to inject a dummy authentication cache, that is, no caching at all

Noted, I will try to understand the code better and come up with a better solution.

Withdrawing this PR.

PR #7751, opened 2024-06-26
[Backport 2.25.x] [GEOS-11445] Update GS Homepage for OGCAPI ServiceDescriptors

aaime:
Backport #7716
 **Authored by:** @davidblasby


aaime:
An automated backport is bound to fail here, there are new modules with pom.xml files referencing the wrong version of GeoServer (2.26.x rather than 2.25.x). Not sure this would be the only problem, but it would be one of them.

davidblasby:
ok, i've moved things to `2.25-snapshot`.  However, there are test cases failures - I think i'll need to pull in some of @aaime fixes from `main` (I've seen these failures before).

davidblasby:
ok, I pulled in  https://github.com/geoserver/geoserver/commit/e96cd4ef639d2f936b585b2789c9b878c8fb1df7

which is just commenting out the failed test case (comment says disabling WFS also disables ogcapi-features).

I think pulling this in is ok because its not going to do anything bad and it will make things easier for the next person running tests on 2.25.x.

davidblasby:
@jodygarnett  - please review

jodygarnett (approved):
thanks for fixing backport; glad it was not more serious conflict.

PR #7766, opened 2024-07-11
[GEOS-11462] 500 error thrown when double adding a user to a group via REST with JDBC user/group services

petersmythe:
[GEOS-11462](https://osgeo-org.atlassian.net/browse/GEOS-11462) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

Allow `duplicate key` to pass while throwing other SQL exceptions, to be consistent with XML-based user/group services.

500 Internal Server Error results: 

> org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "group_members_pkey"
>   Detail: Key (groupname, username)=(test_group_name, test_user_name) already exists.

# Checklist

### DIFF
--- a/src/security/jdbc/src/main/java/org/geoserver/security/jdbc/JDBCRoleStore.java
+++ b/src/security/jdbc/src/main/java/org/geoserver/security/jdbc/JDBCRoleStore.java
@@ -11,6 +11,7 @@ import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
 import java.sql.Types;
+import java.util.logging.Logger;
 import org.geoserver.security.GeoServerRoleService;
 import org.geoserver.security.GeoServerRoleStore;
 import org.geoserver.security.impl.GeoServerRole;
@@ -25,6 +26,9 @@ public class JDBCRoleStore extends JDBCRoleService implements GeoServerRoleStore
 
     protected boolean modified;
     protected Connection connection;
+
+    static Logger LOGGER =
+            org.geotools.util.logging.Logging.getLogger("org.geoserver.security.jdbc");
 
     /**
      * The identical connection is used until {@link #store()} or {@link #load()} is called. Within
@@ -43,7 +47,7 @@ public class JDBCRoleStore extends JDBCRoleService implements GeoServerRoleStore
         // do nothing
     }
 
-    /** To be called at the the end of a transaction, frees the current {@link Connection} */
+    /** To be called at the end of a transaction, frees the current {@link Connection} */
     protected void releaseConnection() throws SQLException {
         if (connection != null) {
             connection.close();
@@ -71,7 +75,7 @@ public class JDBCRoleStore extends JDBCRoleService implements GeoServerRoleStore
 
     protected void addRoleProperties(GeoServerRole role, Connection con)
             throws SQLException, IOException {
-        if (role.getProperties().size() == 0) return; // nothing to do
+        if (role.getProperties().isEmpty()) return; // nothing to do
 
         PreparedStatement ps = getDMLStatement("roleprops.insert", con);
         try {
@@ -81,6 +85,12 @@ public class JDBCRoleStore extends JDBCRoleService implements GeoServerRoleStore
                 ps.setString(2, key.toString());
                 ps.setObject(3, propertyVal);
                 ps.execute();
+            }
+        } catch (SQLException ex) {
+            if (ex.getMessage().contains("duplicate key value violates unique constraint")) {
+                LOGGER.fine("Ignoring duplicate key value violates unique constraint " + ex);
+            } else {
+                throw new IOException(ex);
             }
         } finally {
             closeFinally(null, ps, null);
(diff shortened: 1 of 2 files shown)


petersmythe:
No unit test added, because it depends on an external database.

src/security/jdbc/src/main/java/org/geoserver/security/jdbc/JDBCRoleStore.java
```diff
@@ -82,6 +86,12 @@ protected void addRoleProperties(GeoServerRole role, Connection con)
                 ps.setObject(3, propertyVal);
                 ps.execute();
             }
+        } catch (SQLException ex) {
+            if (ex.getMessage().contains("duplicate key value violates unique constraint")) {
```
mprins, line 90: So... these kind of messages are very database and jdbc driver specific and this seems to cater only for PostgreSQL

petersmythe: Thanks for the valuable input @mprins.  Would it be acceptable if I checked for different messages for the following RDBMSs?

| **RDBMS**        | **Error Message**                                                                                                                                                                 |
|------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| **SQL Server**   | `Msg 2627, Level 14, State 1, Line 1 Violation of UNIQUE KEY constraint 'constraint_name'. Cannot insert duplicate key in object 'object_name'. The duplicate key value is (duplicate_value).` |
| **MySQL**        | `ERROR 1062 (23000): Duplicate entry 'duplicate_value' for key 'constraint_name'`                                                                                                |
| **Oracle**       | `ORA-00001: unique constraint (constraint_name) violated`                                                                                                                        |
| **SQLite**       | `SQLITE_CONSTRAINT_UNIQUE: UNIQUE constraint failed: table_name.column_name`                                                                                                      |
| **PostgreSQL**   | `ERROR: duplicate key value violates unique constraint "constraint_name" DETAIL: Key (column_name)=(duplicate_value) already exists.`                                            |
| **DB2**          | `SQL0803N  One or more values in the INSERT statement, UPDATE statement, or foreign key update caused by a DELETE statement are not valid because the primary key, unique constraint or unique index identified by "constraint_name" constrains table "table_name" from having duplicate values for the index key.` |
| **H2**           | `Unique index or primary key violation: "constraint_name index_id"; SQL statement: ...`                                                                                          |

aaime: Pheraps you want to catch a [SQLIntegrityConstraintViolationException](https://docs.oracle.com/javase/8/docs/api/java/sql/SQLIntegrityConstraintViolationException.html) instead, and be done with it?

However, I would take a step back, before getting bogged down into details. What are the XML services doing when there is a duplicate instead? What are the consequences of ignoring the exception?

When I see how this is presented, it seems like a "I want to see this exception gone" but I'd like to see a bit of higher-level discussion instead.  E.g. complaining when inserting a group that already exists does not sound like a bad idea, would it be better to align the XML modules instead? Does the client have a quick way to check for duplicates before entering one?

mprins: Added some comments to the jira ticket and examples of why trying to catch [`SQLIntegrityConstraintViolationException`](https://docs.oracle.com/javase/8/docs/api/java/sql/SQLIntegrityConstraintViolationException.html) will not work in general

petersmythe: Q: Where should discussions take place, in GitHub PRs or in Jira tickets?  Probably not both ;-)

> What are the XML services doing when there is a duplicate instead? What are the consequences of ignoring the exception?

The current (XML service) implementation considers a duplicate a no-op and returns 200.  I personally think that is correct, as the outcome was achieved, but if you want me to follow up this PR with another that changes them all to 409 Conflict, I can look at that.

> Does the client have a quick way to check for duplicates before entering one?

[GET](https://docs.geoserver.org/2.25.x/en/user/rest/api/userrole.html#rest-usergroup-service-servicename-user-user-groups) all users for a specified group (or the other way around) and check that the response contains the provided user is easy enough to do, but it's not as easy as handling a 409 Conflict error.

aaime: Higher-level discussions are best placed on list, it's hard to get everyone's attention on the PR comments.
Gut feeling, if someone is duplicating something existing, a 409 seems like a better choice, but this is exactly the kind of decision where having all consider the issue is best.

petersmythe: Feedback via PSC meeting: 

check first: if association is already there then return success  
Avoids the unmaintainable mess above.

petersmythe:
Superseded by https://github.com/geoserver/geoserver/pull/7783

PR #7768, opened 2024-07-11
[GEOS-11443] REST API does not take effect immediately due to 10 minute authentication cache

petersmythe:
[GEOS-11443](https://osgeo-org.atlassian.net/browse/GEOS-11443) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

From my experience with the GeoServer REST API, e.g. adding a user to a group [GeoServer API Docs](https://docs.geoserver.org/stable/en/api/#1.0.0/usergroup.yaml) , the change happens and is written to disk immediately, however there is a 10 minute authentication cache that prevents the change from being **effective** immediately.

If I am impatient, I typically reload the configuration or use the AuthKey synchronise button to clear the 10 min auth cache. I am aware of the reload API endpoint [GeoServer API Docs](https://docs.geoserver.org/stable/en/api/#1.0.0/reload.yaml)  but that is a bit drastic.  
The reset endpoint does not appear to clear the auth cache.  This now fixes that.

A better solution than my previous attempt: https://github.com/geoserver/geoserver/pull/7746

# Checklist

### DIFF
--- a/src/restconfig/src/main/java/org/geoserver/rest/CatalogReloadController.java
+++ b/src/restconfig/src/main/java/org/geoserver/rest/CatalogReloadController.java
@@ -5,11 +5,13 @@
 package org.geoserver.rest;
 
 import org.geoserver.config.GeoServer;
+import org.geoserver.platform.GeoServerExtensions;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RestController;
+import org.geoserver.security.GeoServerSecurityManager;
 
 /** Catalog reload controller */
 @RestController
@@ -33,5 +35,6 @@ public class CatalogReloadController extends AbstractGeoServerController {
             method = {RequestMethod.POST, RequestMethod.PUT})
     public void reset() {
         geoServer.reset();
+        GeoServerExtensions.bean(GeoServerSecurityManager.class).getAuthenticationCache().removeAll();
     }
 }
--- a/src/restconfig/src/main/java/org/geoserver/rest/CatalogReloadController.java
+++ b/src/restconfig/src/main/java/org/geoserver/rest/CatalogReloadController.java
@@ -5,8 +5,6 @@
 package org.geoserver.rest;
 
 import org.geoserver.config.GeoServer;
-import org.geoserver.platform.GeoServerExtensions;
-import org.geoserver.security.GeoServerSecurityManager;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.web.bind.annotation.RequestMapping;
@@ -35,8 +33,5 @@ public class CatalogReloadController extends AbstractGeoServerController {
             method = {RequestMethod.POST, RequestMethod.PUT})
     public void reset() {
         geoServer.reset();
-        GeoServerExtensions.bean(GeoServerSecurityManager.class)
-                .getAuthenticationCache()
-                .removeAll();
     }
 }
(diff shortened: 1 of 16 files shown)


petersmythe:
https://docs.geoserver.org/latest/en/user/rest/api/reset.html: 

> Resets all store, raster, and schema caches. This operation is used to **force GeoServer to drop all caches** and store connections and reconnect to each of them the next time they are needed by a request. This is useful in case the stores themselves cache some information about the data structures they manage that may have changed in the meantime.

The documentation is still perfectly applicable, no changes needed, unless we add `Resets all authentication, store, raster, and schema caches`?

src/restconfig/src/main/java/org/geoserver/rest/CatalogReloadController.java
```diff
@@ -33,5 +35,8 @@ public void reload() throws Exception {
             method = {RequestMethod.POST, RequestMethod.PUT})
     public void reset() {
         geoServer.reset();
+        GeoServerExtensions.bean(GeoServerSecurityManager.class)
+                .getAuthenticationCache()
+                .removeAll();
```
aaime, line 40: This looks much better indeed, but... it's still backwards.
See how the controller is ust calling the reset method, without any explicit, special action targeting a particular cache? E.g., it's not calling ResourcePool to reset its caches, it's the pool that' listening to the reset events and cleaning itself. I believe you can have the security manager implement GeoServerLifecycleHandler and react on the reset "onReset" event instead. Or have you already tried it, and it did not work?

petersmythe: Implemented, successfully too, as you suggested (I think), but I am getting test errors in AuthenticationCacheFilterTest.java: 

> Caused by: org.geoserver.platform.GeoServerExtensions$MultipleBeansException: Multiple beans of type [org.geoserver.security](http://org.geoserver.security/).auth.AuthenticationCache


Clearly I still do not know how to instantiate the bean correctly.

petersmythe:
@jodygarnett  Here's a summary of the changes, excluding documentation, etc.

`src/main/src/main/java/applicationContext.xml`: create a org.geoserver.security.auth.GuavaAuthenticationCacheImpl, 

why is it necessary to do this, given  https://github.com/geoserver/geoserver/blob/74ddad7db2179de4c9cc1cd266b73dceb938f98a/src/main/src/main/java/org/geoserver/security/GeoServerSecurityManager.java#L698

```
    AuthenticationCache lookupAuthenticationCache() {
        AuthenticationCache authCache = GeoServerExtensions.bean(AuthenticationCache.class);
        return authCache != null ? authCache : new GuavaAuthenticationCacheImpl(1000);
    }
```
	

`src/main/src/main/java/org/geoserver/security/auth/GuavaAuthenticationCacheImpl.java` 	extend  GuavaAuthenticationCacheImpl to also  implements GeoServerLifecycleHandler (onReset calls removeAll)

The rest of the changes are superfluous

petersmythe:
You ask:

> Are there any examples of how to select an alternate cache? 

I don't know if the code comment above helps?

> Or do you have to fix the original …

I'm not sure what you mean here, the original what?  `GuavaAuthenticationCacheImpl` ?

petersmythe:
> https://github.com/geoserver/geoserver/blob/79239c73bc747c3ed529ad0d2e6ee5ea02bc7e60/src/main/src/main/java/applicationContext.xml#L361

If I do not add this bean definition here, and leave the `GuavaAuthenticationCacheImpl` creation to https://github.com/geoserver/geoserver/blob/79239c73bc747c3ed529ad0d2e6ee5ea02bc7e60/src/main/src/main/java/org/geoserver/security/auth/GuavaAuthenticationCacheImpl.java#L31

then I do not see the GuavaAuthenticationCacheImpl listed in `GeoServerLifecycleHandler`s:

image

(and the required fix does not work)

PR #7771, opened 2024-07-12
Geos 11463 improve wms feature dimension validation main

skalesse:
[GEOS-11463](https://osgeo-org.atlassian.net/browse/GEOS-11463) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

# [GEOS-11463](https://osgeo-org.atlassian.net/browse/GEOS-11463) - WMS vector dimension validation should query only one feature and only for dimension attribute 

## Problem description

 [GEOS-9757](https://osgeo-org.atlassian.net/browse/GEOS-9757)  added request validation of raster and vector dimensions.

We observed that the validation code for vector dimensions creates a filter for the particular dimension but wraps it using a ALL_NAMES query, which basically results in a query for all features attributes. This behavior could result in a lot of feature data to be transferred from the backends to GeoServer (probably depending on the store implementation).

On the other hand when querying the data stores for capabilities, the queries used there are limited the particular dimension attribute. This results in much less data to be transferred from the backends, especially if the result is only used to check if it's not empty.

The validation code could be improved in such a way that it functions similar to the way dimension data for capabilities is queried - wrap the filter in a query, but fix the query property to be the particular dimension attribute. An additional improvement could be to limit the query to return only one feature. 

## Solution

The solution is to wrap the before mentioned filer in a query that limits the number of features to 1 and additionally restrict the properties of the query to only the attribute that is required for the particular dimension. 

The fix is in class `org.geoserver.wms.WMS` that adds a new method, which wraps the dimension filter 

```
/**
 * wraps the given filter in a query for dimension requests.
 *
 * @param filter the filter
 * @param dimensionInfo the dimension info
 * @param typeName the feature type name
 * @return a query that wraps the filter and restricts the query to the dimension attribute and
 *     max features = 1.
 */
private static Query getDimensionQuery(
        Filter filter, DimensionInfo dimensionInfo, String typeName) {
    List<String> propertyNames = new ArrayList<>();
    propertyNames.add(dimensionInfo.getAttribute());
    if (dimensionInfo.getEndAttribute() != null
            && dimensionInfo.getPresentation() != DimensionPresentation.LIST) {
        propertyNames.add(dimensionInfo.getEndAttribute());
    }
    Query query = new Query(typeName, filter);
    query.setPropertyNames(propertyNames);
    query.setMaxFeatures(1);
    return query;
}
```

This new method is then used in the three validation methods:

- `org.geoserver.wms.WMS#validateTimes()`
- `org.geoserver.wms.WMS#validateElevations` 
- `org.geoserver.wms.WMS#validateCustomDimensions()`

This should limit significantly the number of data transferred from the backends needed for validation . 

## Testing 

A JUnit suite has been added within the scope of this ticket in order to perform basic validations for the 'time' , 'elevation' and custom dimensions.

The class is located at :  `org.geoserver.wms.WMSVectorDimensionValidationTest`.

# Checklist


aaime:
Please check the [QA Procedures](https://docs.geoserver.org/latest/en/developer/qa-guide/index.html) and run them locally for the wms module, I'm almost sure some of the changes will make QA fail (using star imports should make PMD fail).

skalesse:
Thanks @aaime for the hint. I have corrected the imports with the last commit. I just noticed the I will also have to set the copyright message in the test class. Unfortunately checkstyle keeps crashing with a timeout on my machine, so I didn't notice earlier. Sorry about that.

skalesse:
Alright, copyright notice is there now as well, but I am kind of irritated now. Checkstyle ~13,00 error in gs-wms, 743 alone in WMS.java, which I have modified. So checkstyle will definitely keep failing. It's probably not for me to fix all those, right?

aaime:
I believe you're not following the instructions on how to run the QA tools, or there is some other oddity about your setup.
Checkstyle is configured to only check class headers in our build.

skalesse:
Oh jeez, I am very sorry I bothered you with this. I got running now. It seems checkstyle doesn't work well when running behind a corporate proxy. It doesn't read the proxy settings in `settings.xml`, then fails with a timeout or does other weird things. After applying a workaround while running `mvn`, all QA now reported success. 

Sorry again, and thanks for the patience.

aaime:
Looks like the OSGeo maven repository is having troubles, which is breaking the builds.
Issue is known and has been escalated: https://trac.osgeo.org/osgeo/ticket/3219#ticket

aaime:
Closing and reopening to re-trigger all builds.

src/wms/src/test/java/org/geoserver/wms/WMSVectorDimensionValidationTest.java
```diff
@@ -0,0 +1,443 @@
...
+    }
+
+    /**
+     * for a 'time' dimension, if a matching value is provided in the request the validation should
+     * succeed
+     *
+     * @throws IOException an unexpected IO exception is thrown
+     */
+    @Test
+    public void matching_time_dimension_should_validate() throws IOException {
```
aaime, line 79: We use camel case for method names (snake case is common in other programming languages). Please fix them all.

src/wms/src/main/java/org/geoserver/wms/WMS.java
```diff
@@ -1835,16 +1843,42 @@ private void validateElevations(
...
+                            elevationInfo,
+                            typeInfo.getFeatureType().getName().getLocalPart());
+            if (DataUtilities.first(fs.getFeatures(dimensionQuery)) == null) {
                 throwInvalidDimensionValue(request, ELEVATION, ELEVATION);
             }
         }
     }
 
+    /**
+     * wraps the given filter in a query for dimension requests.
```
aaime, line 1860:
```suggestion
     * Wraps the given filter in a query for dimension requests.
```

Javadoc comments should start with a capital (it's a regular sentence). Can you please also fix the test method javadocs?

aaime (changes requested):
Overall looks good. See minor update requests below.

src/wms/src/test/java/org/geoserver/wms/WMSVectorDimensionValidationTest.java
skalesse: Done.

src/wms/src/main/java/org/geoserver/wms/WMS.java
skalesse: Done.

skalesse:
Hi @aaime. Thanks for pointing out the problems with naming conventions and JavaDoc. I have fixed them now. 

I have to admit that my last commit contains a bit more than just the requested changes. After three nights of thinking about a way to write a test-case that verifies the actual fix (query.properties.size=1 and maxFeatures=1), I finally just now found a way to implement it. I realized too late that I should have put these additional tests in a separate commit. I hope you don't mind. 

Let me know if you require anything else. Thanks!

src/wms/src/test/java/org/geoserver/wms/WMSVectorDimensionValidationTest.java
```diff
@@ -0,0 +1,564 @@
...
+     */
+    @Test
+    public void matchingTimeDimensionShouldValidate() throws IOException {
+        // the dimension and its value
+        String dimension = "time";
+        Date dimensionValue = Date.from(Instant.parse("2012-02-11T10:15:30.00Z"));
+        // create a map request
+        GetMapRequest request = mapRequest(singletonList(dimensionValue), emptyList(), emptyMap());
+        // disable validation
+        setValidationEnabled(true);
```
aaime, line 99:
```suggestion
        // enable validation
        setValidationEnabled(true);
```

The same lines have been copied over, so is the inconsistency between comment and code is found in other test methods as well.

src/wms/src/test/java/org/geoserver/wms/WMSVectorDimensionValidationTest.java
```diff
@@ -0,0 +1,564 @@
...
+     */
+    @Test
+    public void mismatchingTimeDimensionShouldValidateIfValidationDisabled() throws IOException {
+        // the dimension and its value
+        String dimension = "time";
+        Date dimensionValue = Date.from(Instant.parse("2012-02-10T10:15:30.00Z"));
+        // create a map request
+        GetMapRequest request = mapRequest(singletonList(dimensionValue), emptyList(), emptyMap());
+        // disable validation
+        setValidationEnabled(false);
```
aaime, line 169: This one does what it says, instead.

aaime (commented):
Had a last pass. Very thorough testing, nice, this one should not regress in the future :rofl: 

While reading the test I was surprised by some apparent inconsistency between comment text and actual code, can you check?

src/wms/src/test/java/org/geoserver/wms/WMSVectorDimensionValidationTest.java
skalesse: Ups. Thanks for finding those! I fixed two places where there where inconsistent comments.

src/wms/src/test/java/org/geoserver/wms/WMSVectorDimensionValidationTest.java
skalesse: Yeah, I went through all comments again and fixed the ones that were not quite right. Thanks for finding them.

skalesse:
> Had a last pass. Very thorough testing, nice, this one should not regress in the future 🤣

Thanks! I believe though the testing could still be improved. Unfortunately I was not really able to test multiple dimensions at once - especially with respect to custom dimensions. The reason for this is https://osgeo-org.atlassian.net/browse/GEOS-11453. I had test-cases in a draft version where I would also test what happens if I have multiple dimensions enabled, like 'time' and 'reference_time', which is a very common use-case for us. Unfortunately these would usually fail for the problem reported at  GEOS-11453. But yeah, maybe in a later step, testing could even be improved to incorporate multiple dimensions.

skalesse:
Hi @aaime , I was wondering about the status of this request. Is there anything else that is still required for this to be merged? 

Also, I was wondering about the options for back-porting to the 2.23.x releases. These two issues (including https://osgeo-org.atlassian.net/browse/GEOS-11453) are of a great importance for us but unfortunately we currently cannot upgrade to any version that uses GeoTools 30+ (for the breaking changes of gt-opengis). 

We would of course be willing to support the back-migration and whatever is required to create a release based 2.23.x.

Thanks!
Sören

aaime:
This request is just waiting for a last round of review, but I've been either absent or overcommitted in the last few weeks, thus unable to review PRs that are outside of work engagements. I should be back to "normal" PR review times mid September, although I hope to be able to check this one again sooner than that.

skalesse:
@aaime Thanks for merging and backporting. Is there any chance to ask the bot to backport to 2.23.x too? Unfortunately we are currently stuck with that old version, because of the structural changes to GeoTools for the 30.x series. 

Thanks so much again! And please let me know if you require help on the backporting of this (and hopefully also of https://osgeo-org.atlassian.net/browse/GEOS-11453) for the same reasons.

PR #7777, opened 2024-07-16
[GEOS-11453] Regression of adding service exception with GEOS-9757

skalesse:
# [GEOS-11453](https://osgeo-org.atlassian.net/browse/GEOS-11453) - Regression of adding service exception with GEOS-9757

## Problem description

[GEOS-9757](https://osgeo-org.atlassian.net/browse/GEOS-9757) added request validation of raster and vector dimensions.

[GEOS-11453](https://osgeo-org.atlassian.net/browse/GEOS-11453) documents a problem, which breaks the functionality of GeoServer to fallback to the defined default values for custom dimensions in case a particular dimension value is missing from the OGC request. Instead of the expected behaviour a `ServiceException` is thrown, indicating that the particular layer has no support for custom dimensions. 

The problem is in class `org.geoserver.wms.CustomDimensionFilterConverter` that is now used in the validation code flow. With [GEOS-9757](https://osgeo-org.atlassian.net/browse/GEOS-9757) that class got some internal refactoring with respect to how the custom dimension lookup is managed. Internally the custom dimension names are now stored in a map, with the key being the dimension name, but with the "dim_" prefix removed. Later when the filter is assembled and the actual default value is looked up in method `getDefaultCustomDimension()` the prefix is not removed before the lookup in the global map. The leads to the fact that a custom dimension info is never found and instead the above mentioned `ServiceExcpetion` exception is thrown. 

The effect is that validation of custom dimension will always fail, in case the dimension is omitted from the request and default value lookup should be performed. 

## Solution

The solution for this problem is to make sure that before accessing the `customDimesnions` map, the dimension name has the "dim_" prefix removed. Then, the filter class should work as it did before. After the change the method `org.geoserver.wms.CustomDimensionFilterConverter#getDefaultCustomDimension` should look like this:

```
private Object getDefaultCustomDimension(String name, Class<?> binding) {
    // check the time metadata
    final DimensionInfo dimensionInfo = customDimensions.get(unrollDimPrefix(name));
    if (dimensionInfo == null || !dimensionInfo.isEnabled()) {
        throw new ServiceException(
                "Layer "
                        + typeInfo.prefixedName()
                        + " does not have custom dimension support enabled");
    }
    DimensionDefaultValueSelectionStrategy strategy =
            defaultValueStrategyFactory.getDefaultValueStrategy(typeInfo, name, dimensionInfo);
    return strategy.getDefaultValue(typeInfo, name, dimensionInfo, binding);
}
```

Note, how in line three, when the lookup for the dimension info is performed, the "dim_" prefix is stripped from the name, so that the lookup shouldn't fail. 

Please note also, that removing the prefix, before calling  `getDefaultCustomDimension()` is no option, because the second half of the method requires `name` to have the prefix applied. 

## Testing

A few JUnit tests have been written at `org.geoserver.wms.WMSTest` to verify the custom dimension filter is working as expected. 

# Checklist

### DIFF
--- a/src/wms/src/test/java/org/geoserver/wms/WMSTest.java
+++ b/src/wms/src/test/java/org/geoserver/wms/WMSTest.java
@@ -5,17 +5,23 @@
  */
 package org.geoserver.wms;
 
+import static java.time.format.DateTimeFormatter.ISO_INSTANT;
+import static java.util.Collections.emptyList;
+import static java.util.Collections.emptyMap;
+import static org.geotools.factory.CommonFactoryFinder.getFilterFactory;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertTrue;
 
 import java.awt.image.BufferedImage;
+import java.io.IOException;
 import java.lang.reflect.Field;
 import java.net.URL;
 import java.text.SimpleDateFormat;
+import java.time.Instant;
 import java.util.Arrays;
-import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
@@ -24,6 +30,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.TimeZone;
 import javax.xml.namespace.QName;
+import org.geoserver.catalog.DimensionDefaultValueSetting;
 import org.geoserver.catalog.DimensionInfo;
 import org.geoserver.catalog.DimensionPresentation;
 import org.geoserver.catalog.FeatureTypeInfo;
@@ -34,7 +41,6 @@ import org.geoserver.data.test.SystemTestData;
 import org.geoserver.platform.ServiceException;
 import org.geotools.api.data.FeatureSource;
 import org.geotools.api.filter.Filter;
-import org.geotools.factory.CommonFactoryFinder;
 import org.geotools.feature.FeatureCollection;
 import org.geotools.feature.FeatureIterator;
 import org.geotools.renderer.style.DynamicSymbolFactoryFinder;
@@ -57,7 +63,7 @@ public class WMSTest extends WMSTestSupport {
         super.onSetUp(testData);
         testData.addVectorLayer(
                 TIME_WITH_START_END,
-                Collections.emptyMap(),
+                emptyMap(),
                 "TimeElevationWithStartEnd.properties",
                 getClass(),
                 getCatalog());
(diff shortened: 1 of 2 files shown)


src/wms/src/test/java/org/geoserver/wms/WMSTest.java
```diff
@@ -75,9 +81,35 @@ protected void setupStartEndTimeDimension(
         getCatalog().save(info);
     }
 
+    /**
+     * set up a feature dimension, but additionally allow to set a default value setting
```
aaime, line 85: Same as in the other PR, start sentences with a capital

src/wms/src/test/java/org/geoserver/wms/WMSTest.java
```diff
@@ -186,6 +220,150 @@ public void doTimeElevationFilter(Object time, Object elevation, Integer... expe
                 Arrays.asList(expectedIds).containsAll(results));
     }
 
+    private static String CUSTOM_DIMENSION_NAME = WMS.DIM_ + "reference_time";
+
+    @Test
+    public void
```
aaime, line 226: And like in the other, please use camel case and more compact names, with explanation in the javadocs

aaime (changes requested):
Looks good in general, see comments below

src/wms/src/test/java/org/geoserver/wms/WMSTest.java
skalesse: Done.

src/wms/src/test/java/org/geoserver/wms/WMSTest.java
skalesse: Done.

skalesse:
Thanks again for pointing out naming conventions and JavaDoc style. I have made the requested changes in the last commit. 

Let me know if anything else is missing. Thanks!

dromagnoli (approved):

aaime:
```
```
<details>
<summary>stderr</summary>

```
error: could not apply 81b304d35b... [GEOS-11453] Regression of adding service exception with GEOS-9757
hint: Disable this message with "git config advice.mergeConflict false"

```
</details>

<details>
<summary>stdout</summary>

```
Auto-merging src/wms/src/main/java/org/geoserver/wms/CustomDimensionFilterConverter.java
Auto-merging src/wms/src/test/java/org/geoserver/wms/WMSTest.java
CONFLICT (content): Merge conflict in src/wms/src/test/java/org/geoserver/wms/WMSTest.java

```
</details>

```bash
git fetch
# Create a new working tree
cd .worktrees/backport-2.23.x
# Create a new branch
git switch --create backport-7777-to-2.23.x
git cherry-pick 81b304d35bf4c97ee8c06e20b230e284b73e7af7,0de39c44329d307968839d5d14e7f2473875b238
# Push it to GitHub
git push --set-upstream origin backport-7777-to-2.23.x
cd ../..
# Delete the working tree
```
Then, create a pull request where the `base` branch is `2.23.x` and the `compare`/`head` branch is `backport-7777-to-2.23.x`.

aaime:
@skalesse if you want a 2.23.x backport, you'll have to work on it manually, the backport bot cannot cherry-pick cleanly.

PR #7803, opened 2024-07-31
[GEOS-11421] Match webui workspace admin functionality in the REST API

groldan:
[GEOS-11421](https://osgeo-org.atlassian.net/browse/GEOS-11421) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

Extends workspace admin access control to the REST API, matching the permissions workspace administrators already have in the web UI.

The default `rest.properties` restricts the whole REST API to `ROLE_ADMINISTRATOR`. It can be edited to open endpoints to other roles, but it has no concept of workspace scoping.

This change adds a complementary configuration file, `security/rest.workspaceadmin.properties`, that opens selected endpoints to any user the security subsystem identifies as a workspace administrator. Which workspaces and resources are visible is decided by the existing catalog security, based on the admin access rules in `layers.properties`. 

In short: `rest.workspaceadmin.properties` opens the door to endpoints, `layers.properties` determines what's behind the door.

### How it works

- `WorkspaceAdminAuthorizer`: shared authorization logic shared by the web UI and the REST API. The workspace admin check is cached per request.
- `WorkspaceAdminAuthorizationManager`: joins the Spring Security filter chain next to the existing authenticated/role managers. Grants access when a workspace admin rule matches and abstains otherwise.
- `WorkspaceAdminRESTAccessRuleDAO` / `WorkspaceAdminRestfulDefinitionSource`: load and match the URL pattern rules from `rest.workspaceadmin.properties`, created from a template on first startup.
- `RESTfulDefinitionSourceProxy`: combines the standard `rest.properties` rules and the workspace admin rules into a single `SecurityMetadataSource`.
- `SecureResourceStore` / `SecuredResource`: wrap the data directory `ResourceStore` behind `/rest/resource`, restricting workspace admins to their workspaces plus read-only global collections. Hidden resources return 404; visible but read-only ones return 403 on write.

### Default access rules

First match wins; unmatched endpoints fall through to `rest.properties` (admin-only by default).

| Endpoint pattern | Access | Notes |
|---|---|---|
| `/rest/workspaces`, `/rest/namespaces` | read | Listings, filtered to administrable workspaces |
| `/rest/workspaces/{workspace}` | read + PUT | Update allowed, rename and delete denied |
| `/rest/workspaces/{workspace}/**` | read/write | Datastores, styles, layergroups, etc. |
| `/rest/layers/**` | read/write | Filtered to administrable workspaces |
| `/rest/styles/**`, `/rest/templates/**` | read | Global styles/templates are read-only |
| `/rest/resource/workspaces/{workspace}/**` | read/write | Resource browser |
| `/rest/resource/styles/**` | read | Global style resources |
| `/rest/services/*/workspaces/{workspace}/**` | read/write | Per-workspace service config (WFS, WMS, etc.) |
| `/rest/crs/**`, `/rest/fonts/**` | read | CRS lookups and fonts |
| `/rest/services/*/settings` | *(no rule)* | Global service settings stay admin-only |

Noteworthy:

- Workspace admins cannot rename workspaces/namespaces or change the default ones (403).
- The REST index (`/rest`) lists only the endpoints the caller can access.
- `ServiceSettingsController` now returns `404` when the requested workspace does not exist (or is hidden from the caller), instead of falling through to the global service settings. This also affects full admins using a wrong workspace name.

### Documentation

New Workspace Administration section under security: a landing page explaining the three configuration layers, a reference for the REST access rules, and a hands-on tutorial using the default data directory. Cross-links added from the layer security, sandbox, production config, and data admin pages.

### Testing

Unit tests for each new class, plus REST integration tests covering the catalog endpoints (`WorkspaceAdminRestIntegrationTest`), the resource browser (`ResourceControllerWorkspaceAdminTest`), and per-workspace WFS/WMS settings.

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/RESTfulDefinitionSourceProxy.java
+++ b/src/main/src/main/java/org/geoserver/security/RESTfulDefinitionSourceProxy.java
@@ -1,0 +1,84 @@
+/* (c) 2024 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.security;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import org.geoserver.security.workspaceadmin.WorkspaceAdminRestfulDefinitionSource;
+import org.springframework.security.access.ConfigAttribute;
+import org.springframework.security.web.access.intercept.FilterInvocationSecurityMetadataSource;
+
+/**
+ * A proxy implementation of {@link FilterInvocationSecurityMetadataSource} that delegates to a list of other
+ * FilterInvocationSecurityMetadataSource instances.
+ *
+ * <p>This allows combining multiple security metadata sources, such as the standard {@link RESTfulDefinitionSource} and
+ * the {@link WorkspaceAdminRestfulDefinitionSource}, to handle different authorization scenarios.
+ */
+@SuppressWarnings({"deprecation", "removal"})
+public class RESTfulDefinitionSourceProxy implements FilterInvocationSecurityMetadataSource {
+
+    private List<FilterInvocationSecurityMetadataSource> delegates;
+
+    /**
+     * Creates a new proxy with the given list of delegate metadata sources.
+     *
+     * @param delegates the list of metadata sources to delegate to
+     */
+    public RESTfulDefinitionSourceProxy(List<FilterInvocationSecurityMetadataSource> delegates) {
+        this.delegates = delegates;
+    }
+
+    /**
+     * Returns true if any of the delegate metadata sources support the given class.
+     *
+     * @param clazz the class to check for support
+     * @return true if any delegate supports the class, false otherwise
+     */
+    @Override
+    public boolean supports(Class<?> clazz) {
+        for (FilterInvocationSecurityMetadataSource delegate : delegates) {
+            if (delegate.supports(clazz)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
+     * Returns the combined security attributes from all delegate metadata sources. Optimizes for the common case of a
+     * single delegate.
+     *
+     * @param object the object to get attributes for
+     * @return a collection of security config attributes from all delegates
+     * @throws IllegalArgumentException if the object is not supported
+     */
+    @Override
+    public Collection<ConfigAttribute> getAttributes(Object object) throws IllegalArgumentException {
+        if (1 == delegates.size()) return delegates.get(0).getAttributes(object);
+        return delegates.stream()
+                .map(d -> d.getAttributes(object))
+                .filter(Objects::nonNull)
+                .flatMap(Collection::stream)
+                .collect(Collectors.toList());
+    }
+
+    /**
+     * Returns the combined config attributes from all delegate metadata sources. Optimizes for the common case of a
+     * single delegate.
+     *
+     * @return a collection of all config attributes from all delegates
+     */
+    @Override
+    public Collection<ConfigAttribute> getAllConfigAttributes() {
+        if (1 == delegates.size()) return delegates.get(0).getAllConfigAttributes();
+        return delegates.stream()
+                .map(d -> d.getAllConfigAttributes())
+                .flatMap(Collection::stream)
+                .collect(Collectors.toList());
+    }
+}
--- a/src/main/src/main/java/org/geoserver/security/RESTfulDefinitionSourceProxy.java
+++ b/src/main/src/main/java/org/geoserver/security/RESTfulDefinitionSourceProxy.java
@@ -1,0 +1,61 @@
+/* (c) 2024 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.security;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import org.geoserver.security.workspaceadmin.WorkspaceAdminRestfulDefinitionSource;
+import org.springframework.security.access.ConfigAttribute;
+import org.springframework.security.web.access.intercept.FilterInvocationSecurityMetadataSource;
+
+/**
+ * A proxy implementation of {@link FilterInvocationSecurityMetadataSource} that delegates to a list of other
+ * FilterInvocationSecurityMetadataSource instances.
+ *
+ * <p>This allows combining multiple security metadata sources, such as the standard {@link RESTfulDefinitionSource} and
+ * the {@link WorkspaceAdminRestfulDefinitionSource}, to handle different authorization scenarios.
+ */
+@SuppressWarnings({"deprecation", "removal"})
+public class RESTfulDefinitionSourceProxy implements FilterInvocationSecurityMetadataSource {
+
+    private List<FilterInvocationSecurityMetadataSource> delegates;
+
+    public RESTfulDefinitionSourceProxy(List<FilterInvocationSecurityMetadataSource> delegates) {
+        this.delegates = delegates;
+    }
+
+    @Override
+    public boolean supports(Class<?> clazz) {
+        for (FilterInvocationSecurityMetadataSource delegate : delegates) {
+            if (delegate.supports(clazz)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /** Returns the combined security attributes from all delegates. */
+    @Override
+    public Collection<ConfigAttribute> getAttributes(Object object) throws IllegalArgumentException {
+        if (1 == delegates.size()) return delegates.get(0).getAttributes(object);
+        return delegates.stream()
+                .map(d -> d.getAttributes(object))
+                .filter(Objects::nonNull)
+                .flatMap(Collection::stream)
+                .collect(Collectors.toList());
+    }
+
+    /** Returns the combined config attributes from all delegates. */
+    @Override
+    public Collection<ConfigAttribute> getAllConfigAttributes() {
+        if (1 == delegates.size()) return delegates.get(0).getAllConfigAttributes();
+        return delegates.stream()
+                .map(d -> d.getAllConfigAttributes())
+                .flatMap(Collection::stream)
+                .collect(Collectors.toList());
+    }
+}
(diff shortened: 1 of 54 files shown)


aaime:
Just looking at the comment above and thinking out loud... what about the ``/layers`` and/or ``/workspaces/{workspaceName}/layers`` resources? 
A workspace admin can also alter publishing details like styles, HTTP caching, and so on.

aaime:
I see activity after a long hiatus, but the PR is still marked as Draft. So not ready yet for review, yes?

groldan:
> I see activity after a long hiatus, but the PR is still marked as Draft. So not ready yet for review, yes?

Hi Andrea, that was the fix merge conflicts/rebase after the palantir reformatting I guess. Now I'm actively working on getting it ready, mostly in terms of documentation and additional tests

src/main/src/main/java/org/geoserver/security/RESTfulDefinitionSourceProxy.java
```diff
@@ -0,0 +1,84 @@
+/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
```
vuilleumierc, line 1: Wrong year in copyright - please also check other files

src/main/src/test/java/org/geoserver/security/workspaceadmin/WorkspaceAdminAuthorizationManagerTest.java
```diff
@@ -0,0 +1,132 @@
...
+        // But user is not a workspace admin
+        when(authorizer.isWorkspaceAdmin(authentication)).thenReturn(false);
+
+        // Should abstain when matching rule but not workspace admin
+        AuthorizationDecision decision = authManager.check(authSupplier, request);
+        assertEquals(ACCESS_ABSTAIN, decision);
+    }
+
+    @Test
+    public void testCheckWhenMatchingRuleAndWorkspaceAdmin() {
```
vuilleumierc, line 115: Would it make sense to also test the case of a workspace admin but there's no matching rule?

src/restconfig/src/test/java/org/geoserver/rest/resources/SecureResourceStoreTest.java
```diff
@@ -0,0 +1,196 @@
...
+        when(securityManager.checkAuthenticationForAdminRole()).thenReturn(false);
+
+        // Create a spied secureStore so we can mock the canWrite method
+        secureStore = spy(secureStore);
+        doReturn(false).when(secureStore).canWrite(anyString());
+
+        // Should not delegate to underlying store when not allowed
+        assertFalse(secureStore.remove("workspaces/sf"));
+
+        // Note: verify that remove was not called on delegate
```
vuilleumierc, line 150: a TODO is left

vuilleumierc (approved):
LGTM and tested locally, I only have minor comments

jodygarnett:
@vuilleumierc and @groldan This one is open and appears good to go? 

Do you want this in for the 2.27.0 release?

groldan:
Hi @jodygarnett, this is a contribution from the GeoServer Cloud project. We're not in a rush to have it merged and I'd rather have it reviewed by someone else too. Cecile has been using it for months for a customer project. This PR is pretty complete but I'm waiting to finish up a series of integration tests to complement the unit tests in here.
That is to say, I don't want this to delay the 2.27 release, nor rush it up, since it's a pretty critical functionality affecting security.

jodygarnett:
@groldan you have 1 day to go before 🥶

groldan:
It won't happen before the sprint. Got burnt out finishing out other customers work and releasing gs cloud in time 😞

groldan:
This is finally ready for review. Added full test coverage and documentation

groldan:
Rebased on top of the latest main branch.
@aaime @jodygarnett I know it's a large PR, but if someone in the PSC can review that'd be great.
Note again this functionality is being used for years now in GeoServer Cloud production deployments, and we're trying to push it upstream within the ongoing effort to prescind of the geoserver fork we use to apply required patched on top of the geoserver branch.
TIA,
Gabe

aaime:
This is too big to review alongside other PRs, but I made a first pass using an AI trained to prioritize the same aspects I inspect when performing manual reviews. Call it a first pass:

-----

  Overall assessment

  This is a well-structured, security-conscious feature with strong integration-test coverage and genuinely good user documentation. The architecture (abstain-only voter that can only widen access, two-layer
  "door vs behind-door" model, request-scoped caching, deferred resource checks) is sound and the dangerous path — filesystem traversal — is properly defended. The main blemishes are not logic bugs but
  discipline issues: a comparator/equals inconsistency that can silently drop a security rule, one undocumented behavior change, and a heavy layer of AI-style boilerplate javadoc/comments that clashes with the
  repo's terse style and should be trimmed before merge.

  Three to address before merge:
  1. Finding nr. 1 — make WorkspaceAdminRestAccessRule.compareTo consistent with equality so no rule can be silently dropped on a priority collision.
  2. Finding nr. 2 — confirm and document the ServiceSettingsController 404 behavior change (and update REST API docs).
  3. Finding nr. 3,4,5 — strip the restating javadoc/inline comments and fix the stale/invented rule lists, per house style.

  Verdict: Request changes — no blocking defects, but nr 1 (silent rule loss) and the documentation/comment cleanup should land before merge.

----

Given this, I'm not very inclined to get sucked into Shakespearean style docs. Can you clean up the documentation/comments so that it's quicker to understand, both for the poor reviewer and the random developer that wil stumble into the code later? Then I'll do an actual manual pass.

groldan:
Hi @aaime, thanks for the review; addressed the three points: 
- `compareTo` is consistent with equals
- `ServiceSettingsController`'s `404` behavior change documented in `doc/en/user/rest/api/services.md`. Added
```markdown
### Exceptions

| Exception                                              | Status code |
|--------------------------------------------------------|-------------|
| GET, PUT or DELETE for a workspace that does not exist | 404         |
```
which covers both whether the workspace exists or is not visible (for a workspace admin it's the same)

-  agreed on the AI-style boilerplate javadoc/comments. They're now trimmed down with the invented rule lists removed; and the bundled template is now the single source of truth.

jurgispods:
Hi @groldan — we've been running this PR's code in production for several months (it ships in the
geoserver-cloud 2.28.4 Docker images), on a
~34k-layer / ~1k-workspace catalog. It works well. We did hit one issue in
`WorkspaceAdminResourceFilter` that looks unintended, and since the PR is still open it seems better
to report it here.

### `workspaceResourceAccess()` can never return `READ`

```java
private ResourceAccess workspaceResourceAccess(Authentication authentication, final String workspace) {
    WorkspaceAccessLimits wsAccessLimits = authorizer.getWorkspaceAccessLimits(authentication, workspace);
    boolean adminable = wsAccessLimits != null && wsAccessLimits.isAdminable();
    return adminable ? ResourceAccess.WRITE : ResourceAccess.NONE;
}
```

`WorkspaceAccessLimits.isReadable()` is never consulted, so any non-workspace-admin gets `NONE` for
`workspaces/{ws}/**` even when the catalog says the workspace is readable by everyone (`*.*.r=*` in
`layers.properties`). `SecureResourceStore` maps `NONE` to `Resource.Type.UNDEFINED`, and under
`mode=HIDE` that surfaces as a 404 from `/rest/resource/workspaces/{ws}/**`. This is indistinguishable from
"the file isn't there", which made it take a while to track down.

Concretely it broke anonymous reads of style assets - SVG icons referenced from SLDs and served from the resource store - which worked before we migrated to GeoServer Cloud, since standard Geoserver has no
equivalent filter. Note this is independent of `security/rest.properties`: an explicit
`IS_AUTHENTICATED_ANONYMOUSLY` grant for that exact path passes the `AuthorizationFilter` and then
still 404s here.

The sibling method already models what we'd expect for this case:

```java
private ResourceAccess noWorkspaceResourceAccess(Authentication authentication, String path) {
    boolean readable = isAllowedCollectionPath(path) && authorizer.isWorkspaceAdmin(authentication);
    return readable ? ResourceAccess.READ : ResourceAccess.NONE;
}
```

so both `ResourceAccess.READ` and `isReadable()` already exist and are simply unused for the
workspace path.

### Suggested change

```java
 private ResourceAccess workspaceResourceAccess(Authentication authentication, final String workspace) {
     WorkspaceAccessLimits wsAccessLimits = authorizer.getWorkspaceAccessLimits(authentication, workspace);
-    boolean adminable = wsAccessLimits != null && wsAccessLimits.isAdminable();
-    return adminable ? ResourceAccess.WRITE : ResourceAccess.NONE;
+    if (wsAccessLimits == null) {
+        return ResourceAccess.WRITE;
+    }
+    if (wsAccessLimits.isAdminable()) {
+        return ResourceAccess.WRITE;
+    }
+    if (wsAccessLimits.isReadable()) {
+        return ResourceAccess.READ;
+    }
+    return ResourceAccess.NONE;
 }
```

Tested locally against the 2.28.4.0 images with the patch applied, `layers.properties`
granting `*.*.r=*`:

| request                                                     | before | after |
|-------------------------------------------------------------|--------|-------|
| admin `GET /rest/resource/workspaces/{ws}/styles/test.svg`     | 200    | 200   |
| anonymous `GET /rest/resource/workspaces/{ws}/styles/test.svg` | 404    | 200   |
| anonymous `PUT`/`DELETE` on the same resource                  | 401    | 401   |
| anonymous `GET /rest/workspaces/{ws}/styles.json`              | 401    | 401   |
| anonymous `GET /rest/workspaces.json`                          | 401    | 401   |

i.e. only the read-only case changes; write still requires `isAdminable()`, and the catalog endpoints
are untouched.

### One judgement call I'd like your opinion on: the `null` branch

`DefaultResourceAccessManager.getAccessLimits(Authentication, WorkspaceInfo)` returns `null` rather
than a populated `WorkspaceAccessLimits` when the user is readable **and** writable but not adminable
and `AdminRequest.get() == null` - its own comment reads *"It is not an admin request, read+write
means full access"*. The current `wsAccessLimits != null && isAdminable()` test reads that `null` as
"no access", which inverts its meaning. We hit this in a second round: a JWT-authenticated, non-admin
service account with workspace write access still got 404s after the `isReadable()` case alone was
fixed.

Mapping `null` to `WRITE` follows that contract literally, and is what we run. But `READ` would be the
conservative choice and would also have fixed our symptom, so I'd rather you decide.

If it helps, I'm happy to open a PR against `GEOS-11421_admin_workspaces_rest_api` including
tests.

PR #7804, opened 2024-08-02
[GEOS-11368] Allow Freemarker templates to update MapML responses

turingtestfail:
[GEOS-11368](https://osgeo-org.atlassian.net/browse/GEOS-11368) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

started on template page

ftl file names correction

fixed xml code block and added map-property

Created preview header integration test and started extracting template

swapped in better method for empty feature

finished first integration test

switched to dedicated template

head template style

registering servicelink

fixed servicelink

changed serviceLink to base,path,kvp

cleanup

dollar signs breaking code block

started on subfeature span insertion with integration test

Got point xml interpolation working

progressing on polygon

Got polygon unmarshal working

got multipolygon to unmarshal

changed strings to coords and linestring parsing

removed interpolated, created head styles using basic mapml

attributes and point test working and phantom space removed

polygon and multipolygon tests

added feature id check to multipolygon test

tests with space replacement

version that uses lists of coordinates instead of strings

removed some other remainder stuff from the space thing

cleanup

added check for tagged geom

doc update

PR changes, mainly map-span

tests updated without xml escape

updated documentation removing cdata escaping

added template attributes to mapml feature

fixed attribute replacement and added documentation examples

more doc update

line test fix

fixed issue with geometrycollection and updated documentation with example

better output if error

format coordinates, including number of decimals

fixed pmd

PR doc updates and started on a wrapper

fixed underline in template doc

preview header change restored

test

remoe ipr iws

simplified optional

doc updates

  

# Checklist


aaime:
Overall looks good. I've made some editorial changes to the docs, and noticed that MapMLTemplate instances seems to be shared across request threads, but the class was not thread-safe, and I've made it so (the template cache is now a concurrent map, the char writer is using a simple thread local based pool).

PR #7806, opened 2024-08-06
[GEOS-11486]: MapML Adding support for custom dimensions

dromagnoli:
# Checklist

### DIFF
--- a/src/extension/mapml/src/test/java/org/geoserver/mapml/MapMLDimensionsTest.java
+++ b/src/extension/mapml/src/test/java/org/geoserver/mapml/MapMLDimensionsTest.java
@@ -16,6 +16,7 @@ import org.custommonkey.xmlunit.SimpleNamespaceContext;
 import org.custommonkey.xmlunit.XMLUnit;
 import org.custommonkey.xmlunit.XpathEngine;
 import org.geoserver.catalog.Catalog;
+import org.geoserver.catalog.CoverageInfo;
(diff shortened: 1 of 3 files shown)


src/extension/mapml/src/test/java/org/geoserver/mapml/MapMLDimensionsTest.java
```diff
@@ -177,24 +159,192 @@ public void testTimeList() throws Exception {
...
+            Catalog catalog,
+            ResourceInfo layerMeta,
+            String path,
+            String dimension,
+            String expectedCount)
+            throws Exception {
         MetadataMap mm = layerMeta.getMetadata();
-        mm.put("mapml.dimension", "time");
+        mm.put("mapml.dimension", dimension);
         catalog.save(layerMeta);
```
aaime, line 325: I find it confusing that the MapML dimension configuration is part of an assert that checks if the output has the dimension. Should be part of the other setup you have before calling this method, IMHO, eventually in its own method if you want to re-use code.

aaime (changes requested):
Overall looks good, see comment below.

src/extension/mapml/src/test/java/org/geoserver/mapml/MapMLDimensionsTest.java
dromagnoli: Feedback addressed, dedicated method has been created and moved out from within the assertion.

PR #7808, opened 2024-08-06
Declare all GeoTools dependencies in the root pom's dependency management section

groldan:
* Centralize the declaration of geotools dependencies to the root pom's dependencyManagement and remove `<version>${gt.version}</version>` in module poms for consistency.

* Exclude `xml-apis` and apache `fop` in the root pom dependency management section Of especial interes for the `gt-app-schema` module not to carry over the unnecessary (sice Java 1.6) `xml-apis` jar, whose presence also makes the eclipse IDE complain there are multiple sources for classes in the `javax.xml.namespace`, `org.w3c.dom`, and other packages.

  

# Checklist

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -324,6 +324,12 @@
       </dependency>
       <dependency>
         <groupId>org.geotools</groupId>
+        <artifactId>gt-main</artifactId>
+        <version>${gt.version}</version>
+        <classifier>tests</classifier>
+      </dependency>
+      <dependency>
+        <groupId>org.geotools</groupId>
         <artifactId>gt-render</artifactId>
         <version>${gt.version}</version>
       </dependency>
@@ -331,6 +337,16 @@
         <groupId>org.geotools</groupId>
         <artifactId>gt-svg</artifactId>
         <version>${gt.version}</version>
+        <exclusions>
+          <exclusion>
+            <artifactId>fop</artifactId>
+            <groupId>org.apache.xmlgraphics</groupId>
+          </exclusion>
+          <exclusion>
+            <artifactId>xml-apis</artifactId>
+            <groupId>xml-apis</groupId>
+          </exclusion>
+        </exclusions>
       </dependency>
       <dependency>
         <groupId>org.geotools</groupId>
@@ -395,6 +411,12 @@
       </dependency>
       <dependency>
         <groupId>org.geotools.xsd</groupId>
+        <artifactId>gt-xsd-gml3</artifactId>
+        <version>${gt.version}</version>
+        <classifier>tests</classifier>
+      </dependency>
+      <dependency>
+        <groupId>org.geotools.xsd</groupId>
         <artifactId>gt-xsd-filter</artifactId>
         <version>${gt.version}</version>
       </dependency>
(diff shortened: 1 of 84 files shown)


src/pom.xml
```diff
@@ -576,6 +610,181 @@
...
+      <dependency>
+        <groupId>org.geotools.xsd</groupId>
+        <artifactId>gt-xsd-wps</artifactId>
+        <version>${gt.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.geotools</groupId>
+        <artifactId>gt-epsg-wkt</artifactId>
+        <version>${gt.version}</version>
+      </dependency>
```
jodygarnett, line 787: it would be good to set the geotools dependencies as it annoying to search in this file given the sheer number present. 

I note that this specific one happens after a group of org.geotools.xsd package dependencies (rather than grouped with other org.geotools package friends.

jodygarnett (commented):
@groldan I am just commenting for now. I am glad to see this compiles, and think the change will make the codebase easier to maintain.

What do you think about grouping or sorting the geotools dependencies by package in the src/pom.xml dependency management section?

groldan:
Re sorting I think it's up `sortpom:sort`.

More generally speaking, in terms of ease of maintenance, I'd like all the projects in the stack (gt/gs/gwc) to publish [BOM](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#bill-of-materials-bom-poms) (Bill Of Materials) poms, so dependant projects would only need to import the bom in their `dependencyManagement` section.

jodygarnett:
For the bom story, I use the src/pom.xml all the time as a bill of materials. What else did you have in mind?

As for the check-list? Does this need a JIRA, test case, you want to check off the first things and indicate this is a non functional change or something?

Thanks!

aaime:
Test case, not sure how that can be done. A Jira would be a useful way to communicate that the main pom is now intended to be used as a bill of materials.

jodygarnett:
There was an sample application used for dev training, such a thing could be used to test the bom workflow.

1. Use bom to establish version numbers
2. Use war overlay to be a web app
3. depend on a few geotools modules
4. jetty:run

mprins:
> Use war overlay to be a web app

This is not a scenario that benefits from a BOM. 
Typically a war overlay project uses a war artifact of a specified version that should have all transitive dependencies resolved, eg. https://github.com/Tailormap/tailormap-data/blob/31f04909be3e810dd221c7417abe0bbe53646ec0/geoserver/maven/pom.xml#L52-L64


BOM usage shines when a library provides one dependencies, eg. GeoTools, that can be imported in the dependencyManagement section of a dependent project or used as a parent artifact. If GeoTools were to provide a BOM, there would be no reason to repeat all the GT dependencies like: https://github.com/Tailormap/tailormap-data/blob/31f04909be3e810dd221c7417abe0bbe53646ec0/geoserver/maven/pom.xml#L52-L64 (except to ban some transitive deps ofcourse); you'd only need something like
```xml
    <dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>org.geotools</groupId>
                <artifactId>geotools-bom</artifactId>
                <version>32.0</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>

```
The good thing about that is if a library appears or disappears in the upstream, there is no need to update that long list of artifacts in the downstream.

The typical case for having a BOM provided by GeoServer is of course the geoserver-cloud project that uses some GeoServer libs.

groldan:
In the end, the discussion about whether to provide a bom is tangential. This PR is about homogeneous treatment of geotools dependencies, as we currently have a mix of version declarations in the root pom and on each module.

Also, excluding the xml-apis transitive dep will let eclipse stop complaining org.w3c.dom.Document and other classes is not found, where the problem is it's declared both in the jre library and in the xml-apis dependency

jodygarnett (approved):
Okay I like this PR, if there was some sort pom policy to order dependence management it would be better than managing that ourself.

I am re-running a failed test, and then this should be good to go.

PR #7826, opened 2024-08-12
[Backport 2.23.x] [GEOS 11463] improve wms feature dimension validation main

aaime:
Backport #7771
 **Authored by:** @skalesse


aaime:
@skalesse this one is API incompatible, I'm closing it, you'll have to create a manual backport PR.
See https://docs.geotools.org/stable/userguide/welcome/upgrade.html for instructions (read the "GeoTools 30.x" section).

aaime:
@skalesse same problem here, closing this one as well

skalesse:
> @skalesse this one is API incompatible, I'm closing it, you'll have to create a manual backport PR. See https://docs.geotools.org/stable/userguide/welcome/upgrade.html for instructions (read the "GeoTools 30.x" section).

Hi @aaime. yeah, I kind of anticipated that this change will not be merged automatically because of gt-opengis. I have branches ready in my fork that work for 2.23.x 'cos I needed to verify that the fixes also work in our environment.  So, PRs should be ready. 

Thanks for the link to the upgrade manual. Unfortunately the problem are the backend libraries we use in our own raster / vector data store implementations (currently using GT 29.x). Those have been migrated to GT 30.x but due to release schedules we cannot use them yet and are thus stuck with 2.23.x. 

Also, unfortunately , our nightlies are automated and rely on the libs in the mvn repos - therefore we cannot build on or own.... So we'll have to think about how to deal with that. 

Anyway, still providing the PRs for others to use. Thanks!

aaime:
@skalesse I don't know about your particular situation, but from experience, trying to maintain an extension to GeoTools/GeoServer outside of the main tree comes with quite a bit of extra work that would be spared if they were in mainline (most refactors done against core classes/interfaces would come for free through IDE refactoring tools).

So, if your raster sources are general enough (even in the specific niche of weather forecasting for example), I would recommend migrating them into GeoTools instead.

PR #7834, opened 2024-08-15
Remove unsupported GeoMesa community module

elahrvivaz:
The GeoMesa community module has not been updated in years. GeoMesa integration is fully described in the separate GeoMesa [documentation](https://www.geomesa.org/documentation/stable/user/geoserver.html).
  

# Checklist


elahrvivaz:
Would there be any place in the documentation to reference related projects that aren't community modules? I.e. a place that would be appropriate to link to GeoMesa? Thanks,

aaime:
We actually don't have a place for that... maybe we should create one. Thinking out loud, a section for "other modules" in our guide, @jodygarnett what do you think? 

We could stick the GeoStyler external plugin there too (right now they have [updated the old community page to point to it](https://docs.geoserver.org/latest/en/user/community/geostyler/index.html), but I believe a dedicated  section would be better).

jodygarnett:
I don’t know; I am frustrated by the lack of participation. The community project approach was one of the initiatives we used to be welcoming to new contributors; it has a mixed track record with some modules graduating quickly; and others in a holding pattern (used but not attracting investment enough to be completed).

I think the PSC has to negotiate a bit more strongly (we only have 2 sponsors presently).

But let me try and be positive; the user documentation may not be the best spot; but the website “Downloads” page may be appropriate, and extra tab linking to compatible stuff. Some thought would be needed so these could clearly indicate compatibility.

aaime:
@jodygarnett I think the community area is working fine, we get lots of action, it's normal that some stuff graduates, other do not, and some just die or leave. I believe the community section has more modules than we have in supported: it's a game of numbers, if only 10% of them manage to become supported it's worth it IMHO.

The download page is an idea, but when a module is controlled outside, it's not easy to find out which module is compatible with which version of GeoServer... I'd have it somewhere in a more general section, don't know, "Friends of GeoServer". We do clearly state that we don't have a say in what happens over there, we don't test them, and we cannot provide support, but users are free to follow the link and see if they find something useful.

jodygarnett:
What about on the download page having:

* Production: active stable and maintenance 
* Development: remove this it was only useful for milestones and RCs, nightly builds are available from “Production” already
* Archive: generated 
* External: let GeoMesa and others be added here

Optional: Make a page for GeoMesa with metadata indicating what versions are supported. Use that to add a “geomesa” link to releases of GeoServer that they are compatible with. Pros: encourage and promote up to date external plugins, no effort to depreciate us maintained external modules. Cons: would provide a better experience over community modules

aaime:
@jodygarnett I would go for the download links change, but I'd also keep Development there, in my circles reaching out to the development series is common for a subset of partners (research institutes mainly). Pheraps the "main" bit should be articulated a bit more, with indication that it's not meant for production, but great to try the latest innovations.

And then yes, I would add an "external" for GeoMesa and GeoStyler, with also a note that anyone maintaining a module in an external repository can push a PR to add theirs, but also that the GeoServer community is not involved in their development, and any question should go directly to the external repository.

aaime:
Merging the removal in the meantime, it's a clean PR and the geomesa plugin was not part of the community nightly build anyways.

elahrvivaz:
Thanks guys - is there any follow up you'd like to see from me/GeoMesa regarding the other changes you've been discussing here?

jodygarnett:
I guess I would like your opinion @elahrvivaz - for the GeoMesa project, and your employer, what is a good fit?

PR #7841, opened 2024-08-19
[GEOS-11503] Update mongo schemaless DWITHIN to support non-point geometry

Mike7311:
[GEOS-11503](https://osgeo-org.atlassian.net/browse/GEOS-11503) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

https://osgeo-org.atlassian.net/browse/GEOS-11503
  
Current implementation of DWITHIN for Mongo is being delegated to Mongo's $near operator, however, this operator accepts only points. The goal of this task is to refactor delegation to instead make use of $geoIntersects, which accepts other geometry types.
The buffer is being applied to input geometry, and resulting geometry is being used in $geoIntersects.

Depends on https://github.com/geotools/geotools/pull/4880 (build will fail without it)

# Checklist

### DIFF
--- a/src/community/schemaless-features/mongodb-schemaless/src/main/java/org/geoserver/schemalessfeatures/mongodb/data/MongoSchemalessFeatureSource.java
+++ b/src/community/schemaless-features/mongodb-schemaless/src/main/java/org/geoserver/schemalessfeatures/mongodb/data/MongoSchemalessFeatureSource.java
@@ -21,7 +21,6 @@ import org.geoserver.schemalessfeatures.data.SchemalessFeatureSource;
 import org.geoserver.schemalessfeatures.mongodb.MongoSchemalessUtils;
 import org.geoserver.schemalessfeatures.mongodb.filter.MongoTypeFinder;
 import org.geoserver.schemalessfeatures.mongodb.filter.SchemalessFilterToMongo;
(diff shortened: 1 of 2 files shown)


src/community/schemaless-features/mongodb-schemaless/src/main/java/org/geoserver/schemalessfeatures/mongodb/data/MongoSchemalessFeatureSource.java
```diff
@@ -219,19 +218,6 @@ protected void visitBinaryComparisonOperator(BinaryComparisonOperator filter) {
         return new Filter[] {splitter.getFilterPre(), splitter.getFilterPost()};
     }
 
-    private Map<String, String> getIndexesInfoMap() {
```
Mike7311, line 222: Removed because $near operator required spatial index, but $geoIntersects does not have such requirement

Mike7311: Reverted this

jodygarnett:
I am resolving the associated JIRA issues GEOS-11503 - I trust that is okay?

PR #7847, opened 2024-08-20
[GEOS-11507] Add layerGroup in StyleParameterFilter

f-necas:
Fixes Add Styles in Tile Caching for Layer groups
image
  

# Checklist

- [x] ~~[Documentation](https://github.com/geoserver/geoserver/tree/main/doc/en/user/source) has been updated (if change is visible to end users).~~
- [x] ~~The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers).~~

### DIFF
--- a/src/gwc/src/main/java/org/geoserver/gwc/layer/StyleParameterFilter.java
+++ b/src/gwc/src/main/java/org/geoserver/gwc/layer/StyleParameterFilter.java
@@ -13,6 +13,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.Sets;
 import com.thoughtworks.xstream.annotations.XStreamAlias;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
@@ -20,8 +21,11 @@ import java.util.Set;
 import java.util.TreeSet;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
+import org.geoserver.catalog.LayerGroupInfo;
 import org.geoserver.catalog.LayerInfo;
+import org.geoserver.catalog.PublishedInfo;
 import org.geoserver.catalog.StyleInfo;
 import org.geotools.util.logging.Logging;
 import org.geowebcache.filter.parameters.ParameterException;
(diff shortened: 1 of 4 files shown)


src/gwc/src/main/java/org/geoserver/gwc/layer/StyleParameterFilter.java
```diff
@@ -190,20 +192,31 @@ public List<String> getLegalValues() {
...
+                newStyles.add(style.prefixedName());
+            }
+            defaultStyle =
+                    layerInfo.getDefaultStyle() != null
+                            ? layerInfo.getDefaultStyle().prefixedName()
+                            : null;
         } else {
-            defaultStyle = null;
+            LayerGroupInfo layerGroupInfo = (LayerGroupInfo) layer;
+            for (StyleInfo style : layerGroupInfo.getStyles()) {
```
groldan, line 211: beware `LayerGroupInfo.getStyles():List<StyleInfo>` can have `null` elements.
I'd like a little bit more concise impl like
```
    public void setLayer(PublishedInfo layer) {
    	checkArgument(layer instanceof LayerInfo || layer instanceof LayerGroupInfo);
    	
        StyleInfo defStyle;
        Collection<StyleInfo> styles;
        if (layer instanceof LayerInfo) {
            defStyle = ((LayerInfo) layer).getDefaultStyle();
            styles = ((LayerInfo) layer).getStyles();
        } else {
            defStyle = ((LayerGroupInfo) layer).getRootLayerStyle();
            styles = ((LayerGroupInfo) layer).getStyles();
        }

        this.defaultStyle = defStyle == null ? null : defStyle.prefixedName();
        this.availableStyles = styles.stream()
                        .filter(Objects::nonNull)
                        .map(StyleInfo::prefixedName)
                        .collect(Collectors.toCollection(TreeSet::new));
    }
```

f-necas:
Thanks @groldan, I edited the method and have done the persistance BUT the comment tells that I shouldn't do it. Do you know why or is it deprecated (2013)? 
```diff
        // Remove the filter as groups shouldn't have auto-updating styles
        if (filter != null && filter instanceof StyleParameterFilter) {
-            layerInfo.removeParameterFilter("STYLES");
+            ((StyleParameterFilter) filter).setLayer(layer);
        }
```

src/gwc/src/main/java/org/geoserver/gwc/layer/StyleParameterFilter.java
```diff
@@ -12,16 +12,15 @@
...
 import com.google.common.collect.Sets;
 import com.thoughtworks.xstream.annotations.XStreamAlias;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Objects;
-import java.util.Set;
-import java.util.TreeSet;
+
+import java.util.*;
```
aaime, line 16: PMD checks (part of the QA build) will fail, wildcard imports are disallowed

src/gwc/src/main/java/org/geoserver/gwc/layer/StyleParameterFilter.java
```diff
@@ -190,20 +194,25 @@ public List<String> getLegalValues() {
...
+        checkArgument(layer instanceof LayerInfo || layer instanceof LayerGroupInfo);
+
+        StyleInfo defStyle;
+        Collection<StyleInfo> styles;
+        if (layer instanceof LayerInfo) {
+            defStyle = ((LayerInfo) layer).getDefaultStyle();
+            styles = ((LayerInfo) layer).getStyles();
         } else {
-            defaultStyle = null;
+            defStyle = ((LayerGroupInfo) layer).getRootLayerStyle();
```
aaime, line 206: I don't believe this is correct... the "root layer" is a special layer used in WMS for EO (you're picking its style), while the list of styles you're retrieving in the next line is the set of styles associated to each layer in the group.

In both cases, these styles cannot be associated to the group as a whole
Layer group have "styles" which are full alternative definitions of the layer group, either with the same layers but different per layer styles, or with different layers altogether. Each one of these is a "LayerGroupStyle".

What you need to do for layer groups, is to call getLayerGrouStyles and collect their names instead (careful, it could be empty, in that case no alternative style is supported).

PR #7870, opened 2024-08-27
removed 'recommended' from the datadir set up wording

ale-cristofori:
This PR applies a change in the Geoserver docs, for the part that documents how to externalize the data-directory.
The wording contained the word 'recommended' for the set up of the context-parameter in the web.xml file.
The word 'recommended' was removed since it could be mistaken by the readers, there is not a 'recommended'  method, but is explained as an alternative.
  

# Checklist

### DIFF
--- a/doc/en/user/source/datadirectory/setting.rst
+++ b/doc/en/user/source/datadirectory/setting.rst
@@ -47,7 +47,7 @@ For the Mac OS X install, set the ``GEOSERVER_DATA_DIR`` environment variable to
 Web archive
 -----------
 
-When running a GeoServer WAR inside a servlet container, the data directory can be specified in a number of ways. The recommended method is to set a **servlet context parameter**. An alternative is to set a **Java system property**.
+When running a GeoServer WAR inside a servlet container, the data directory can be specified in a number of ways. One alternative is to set a **servlet context parameter**, another alternative is to set a **Java system property**.
 
 Context parameter
 ^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


aaime:
Looks good to me, the configuration in web.xml is actually annoying when upgrading, as one has to patch the file again.

jodygarnett:
why remove? setting up a geoserver.xml file to define data directory location is exactly what I recommend. 

The other techniques, system property and environmental variable, may accidentally be picked up by another web app.

doc/en/user/source/datadirectory/setting.rst
```diff
@@ -47,7 +47,7 @@ For the Mac OS X install, set the ``GEOSERVER_DATA_DIR`` environment variable to
 Web archive
 -----------
 
-When running a GeoServer WAR inside a servlet container, the data directory can be specified in a number of ways. The recommended method is to set a **servlet context parameter**. An alternative is to set a **Java system property**.
+When running a GeoServer WAR inside a servlet container, the data directory can be specified in a number of ways. One alternative is to set a **servlet context parameter**, another alternative is to set a **Java system property**.
```
jodygarnett, line 50:
```suggestion
When running a GeoServer WAR inside a servlet container, the data directory can be specified in a number of ways. 

See :ref:`application_properties` section for details on setting this location using java system property, web context parameter, or environmental variable.
```

aaime:
Nobody said to "remove" it, just to remove its recommendation. The system variable is unlikely to be picked from another application, because it would have to be another GeoServer to make sense of it... two GeoServer in the same Tomcat seems to be pretty dumb (high memory usage, does not undeploy cleanly...). The environment variables are the default for docker images instead (normally Tomcat is literally the main process in the docker image).

Your alternative suggestion is good though, I've merged it.

petersmythe:
@aaime this PR wasn't merged, so I have gone ahead and done that.

PR #7896, opened 2024-09-16
[GEOS-11543] WFS Unable to use propertyName to filter properties in a GetFeature request

turingtestfail:
[GEOS-11543](https://osgeo-org.atlassian.net/browse/GEOS-11543) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

fixed to ignore lists of lists and added tests

  

# Checklist

### DIFF
--- a/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
+++ b/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
@@ -714,8 +714,10 @@ public class Dispatcher extends AbstractController {
                 boolean kvpParsed = false;
                 boolean xmlParsed = false;
 
-                if (req.getKvp() != null && req.getKvp().size() > 0) {
-                    // use the kvp reader mechanism
+                if (req.getKvp() != null
+                        && req.getKvp().size() > 0
+                        && (!(req.getKvp().size() == 1 && req.getKvp().containsKey("service")))) {
+                    // use the kvp reader mechanism, but not if the only kvp is the service
                     try {
                         requestBean = parseRequestKVP(parameterType, req);
                         kvpParsed = true;
@@ -1530,6 +1532,19 @@ public class Dispatcher extends AbstractController {
     }
 
     Map<String, Object> parseKVP(Request req, Map<String, Object> kvp) {
+        // if the service is not explicitly set, some parsers will not be found
+        if (!kvp.containsKey("service") && !kvp.containsKey("SERVICE") && !citeCompliant) {
+            String service = null;
+            try {
+                service = getServiceFromRequest(req);
+                if (!"OWS".equalsIgnoreCase(service)) { // OWS is too generic for the parser lookup
+                    kvp.put("service", service);
+                }
+                logger.log(Level.FINER, "service kvp parameter not found, setting to " + service);
+            } catch (Exception e) {
+                logger.log(Level.WARNING, "Unable to determine service from kvp or context");
+            }
+        }
         List<Throwable> errors = KvpUtils.parse(kvp);
         if (!errors.isEmpty()) {
             req.setError(errors.get(0));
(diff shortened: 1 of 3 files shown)


jratike80:
Do you know when this issue has come? We have even an example in the standard demo requests "WFS_getFeatureBetweenCQL.url"
http://localhost:8080/geoserver/wfs?request=GetFeature&typeName=topp:states&propertyName=STATE_NAME,LAND_KM,the_geom&version=1.0.0

Maybe there could be also a JIRA issue that explains which versions are/were affected. I feel that &propertyName used to work.

turingtestfail:
> Do you know when this issue has come? We have even an example in the standard demo requests "WFS_getFeatureBetweenCQL.url" http://localhost:8080/geoserver/wfs?request=GetFeature&typeName=topp:states&propertyName=STATE_NAME,LAND_KM,the_geom&version=1.0.0
> 
> Maybe there could be also a JIRA issue that explains which versions are/were affected. I feel that &propertyName used to work.

Not sure the history of this but the issue is not directly with propertyName.  propertyName works but the "service" KVP has to be set, otherwise the parser for propertyName is not found.  The fix I just added to this PR will insert the service KVP if servlet path starts with "wfs/" and we are not in strict CITE mode.

src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
```diff
@@ -1530,6 +1530,14 @@ void parseKVP(Request req) throws ServiceException {
     }
 
     Map<String, Object> parseKVP(Request req, Map<String, Object> kvp) {
+        // if the service is not explicitly set, some WFS parsers will not be found
+        if (!kvp.containsKey("service")
+                && !kvp.containsKey("SERVICE")
```
aaime, line 1535: The KVP map shoud be a CaseInsensitivemap. Little known fact, all keys in OGC services are case insensitive, one can write a valid request with "sErViCe=WFS". However, values are case sensitive, so "service=wfs" is actually not valid, strictly speaking)

src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
```diff
@@ -1530,6 +1530,14 @@ void parseKVP(Request req) throws ServiceException {
...
+                && req.getHttpRequest().getServletPath().toLowerCase().startsWith("/wfs")
```
aaime, line 1536: Good start but I'd generalize it a bit, to make it valid across all OGC services.

See what the Dispatcher is already doing [here](https://github.com/geoserver/geoserver/blob/ce22d0d99c89bc265c64e61dda1340ce1ebf63ef/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java#L587) and [here].(https://github.com/geoserver/geoserver/blob/main/src/ows/src/main/java/org/geoserver/ows/Dispatcher.java#L1597). 
It's basically assuming the request path is the service, it if could not be determined in any other way. 
I would do the same, and ignore the path if it happens to be the generic "ows".

This way you have a fix that generalizes for wms, wfs, wcs and so on (take the path, make it uppercase, use as service).

aaime (changes requested):

src/ows/src/main/java/org/geoserver/ows/Dispatcher.java
turingtestfail: Ok, I piggybacked off of the method you pointed me too and expanded it a bit with a check on the HttpRequest if the path is not set (which was the case with KVP parse)

jratike80:
Ok, I did not notice that this happens only when the GetFeature request is missing `service=WFS`. It is wrong by the WFS standard because "service" has always been mandatory, but unfortunately GeoServer has always been lenient with the missing parameter.

PR #7915, opened 2024-09-27
Adding CycloneDX to generate a Software Bill Of Materials

aaime:
Experimenting with the CycloneDX plugin to generate a Software Bill Of Materials in said format.
Apparently this is becoming a common request to validate software contents.

Here is an example output tested on GeoServer 2.24.3 with a given set of plugins enabled via profiles, so that you can get an idea:
[CycloneDX-Sbom.json](https://github.com/user-attachments/files/17164424/CycloneDX-Sbom.json)

The command to generate it was:

```
> cd web/app
> mvn clean install -nsu -fae -PsldService,printing,monitor,control-flow,wps,kmlppio,wps-download,excel,querylayer,gdal,authkey,css,ysld,importer,wmts-multi-dimensional,backup-restore,oauth2-geonode,oauth2-openid-connect,geofence-server,geofence-wps
```

and then grab the results in web/app/target.

This would help supporting the EU [Cyber Resilence Act](https://digital-strategy.ec.europa.eu/en/library/cyber-resilience-act), which among the other things, requires a SBOM structure (I've been told this is already active in Germany). See also [this article by OWASP](https://owasp.org/blog/2024/03/07/OWASP-CycloneDX-is-ready-to-support-your-CRA-compliance-journey).

I've also noticed software, like PMD, that is adding these SBOM files as part of their release process:
https://github.com/pmd/pmd/releases/tag/pmd_releases%2F7.6.0

# Checklist

### DIFF
--- a/src/web/app/pom.xml
+++ b/src/web/app/pom.xml
@@ -237,6 +237,33 @@
   </dependencies>
   <build>
     <plugins>
+      <plugin>
+        <groupId>org.cyclonedx</groupId>
+        <artifactId>cyclonedx-maven-plugin</artifactId>
+        <version>2.8.2</version>
+        <executions>
+          <execution>
+            <phase>package</phase>
+            <goals>
+              <goal>makeAggregateBom</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <projectType>library</projectType>
+          <schemaVersion>1.5</schemaVersion>
+          <includeBomSerialNumber>true</includeBomSerialNumber>
+          <includeCompileScope>true</includeCompileScope>
+          <includeProvidedScope>true</includeProvidedScope>
+          <includeRuntimeScope>true</includeRuntimeScope>
+          <includeSystemScope>true</includeSystemScope>
+          <includeTestScope>false</includeTestScope>
+          <includeLicenseText>false</includeLicenseText>
+          <outputReactorProjects>true</outputReactorProjects>
+          <outputFormat>all</outputFormat>
+          <outputName>CycloneDX-Sbom</outputName>
+        </configuration>
+      </plugin>
       <!-- Builds a valid data directory into the web app -->
       <plugin>
         <artifactId>maven-antrun-plugin</artifactId>


mprins:
I usually do: `mvn clean package org.cyclonedx:cyclonedx-maven-plugin:2.8.2:makeBom -DprojectType=application -U -f src/pom.xml -pl :gs-web-app` main difference to your config is that I've chosen to use "application" for the geoserver war.

our use-case involves submitting the sbom to [Dependency-Track](https://docs.dependencytrack.org/) to monitor dependencies.

Alternatively GH provides SBOM download/export for GeoServer at https://github.com/geoserver/geoserver/network/dependencies though that also includes lots of irrelevant things like python and maven deps for building

src/web/app/pom.xml
```diff
@@ -237,6 +237,33 @@
   </dependencies>
   <build>
     <plugins>
+      <plugin>
+        <groupId>org.cyclonedx</groupId>
+        <artifactId>cyclonedx-maven-plugin</artifactId>
+        <version>2.8.2</version>
```
jodygarnett, line 243: Usual feedback of pushing version up to root pom.xml

jodygarnett:
I really like this activity @aaime and @mprins - thank you both for sharing.

aaime:
@jodygarnett all I needed was a "one off" sbom generation, but thought to open the PR to gather feedback. I have no block of time allocated to push this further, but happy to keep discussing.

In my mind the main question remaining is how to deliver this, should we decide to include it in releases (as said, at the moment, I only needed a one-off SBOM for a GeoServer with a particular set of plugins). 
Should one have a sbom for GeoServer, as a release artifact, and one in each extension zip? Or maybe a zip with all the sboms for the main war and for the various extensions, to be then processed by online tools like https://cyclonedx.github.io/cyclonedx-web-tool/ (see the "merge" option).

The idea to use the GeoServer own REST API (if I understood correctly) sounds weird, although I guess it may be convenient to go and ask a given GeoServer installation for its own SBOM... but we'd have to compute the SBOMS during every build, and add them into the classpath as resources, to pull that off. The calculation seemed to take a while, this PR is just a draft, for something "real" I would have it as part of the release procedure only.

jodygarnett:
Now that I understand what this is, I think it would be easiest to:
- include this in the build process packaging step
- include the file in the zip download (not sure if their is a folder convention to follow?)

Anyone wishing to do an audit can process the bill of materials for core geoserver, and any extensions they choose to install also.

mprins:
if/when Maven is used to publish/push artifacts the sbom is attached to the artifact including signature files for integrity eg. 

```xml
<dependency>
  <groupId>org.tailormap</groupId>
  <artifactId>tailormap-api</artifactId>
  <version>11.4.0</version>
  <classifier>cyclonedx</classifier>
  <type>json</type>
</dependency>
```

see: https://repo.b3p.nl/nexus/#browse/browse:public:org%2Ftailormap%2Ftailormap-api%2F11.4.0%2Ftailormap-api-11.4.0-cyclonedx.json

PR #7930, opened 2024-10-03
[GEOS-11561]: Support client-delegating MapML-Proxy

dromagnoli:
[GEOS-11561](https://osgeo-org.atlassian.net/browse/GEOS-11561) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

  

# Checklist


src/extension/mapml/src/test/java/org/geoserver/mapml/MapMLBaseProxyTest.java
```diff
@@ -0,0 +1,153 @@
...
+        }
+    }
+
+    /**
+     * Executes a request using the GET method and returns the result as an MapML document.
+     *
+     * @param path The portion of the request after the context, example:
+     * @return A result of the request parsed into a dom.
+     */
+    protected org.w3c.dom.Document getMapML(final String path) throws Exception {
```
aaime, line 125: Unnecessary class name qualification?

```suggestion
    protected Document getMapML(final String path) throws Exception {
```

src/extension/mapml/src/test/java/org/geoserver/mapml/MapMLBaseProxyTest.java
```diff
@@ -0,0 +1,153 @@
...
+        MockHttpServletRequest request = createRequest(path, false);
+        request.addHeader("Accept", "text/mapml");
+        request.setMethod("GET");
+        request.setContent(new byte[] {});
+        String resp = dispatch(request, "UTF-8").getContentAsString();
+        return dom(new ByteArrayInputStream(resp.getBytes()), true);
+    }
+
+    /** For debugging purposes */
+    public static void printDocument(Document doc) throws TransformerException {
```
aaime, line 135: Why not just use the existing "GeoServerSystemTestSupport.print(Document)"  method?
It's already set up to write on System.out when running tests from IDE, but skip prints when running them over a maven build.

src/extension/mapml/src/test/java/org/geoserver/mapml/MapMLWMSDataAccessProxyTest.java
```diff
@@ -0,0 +1,108 @@
...
+    @Override
+    protected void onSetUp(SystemTestData testData) throws Exception {
+        super.onSetUp(testData);
+        Catalog catalog = getCatalog();
+
+        WMSStoreInfo wmsStore = catalog.getFactory().createWebMapServer();
+        wmsStore.setName("wmsStore");
+        wmsStore.setWorkspace(catalog.getDefaultWorkspace());
+        wmsStore.setCapabilitiesURL(
+                "http://localhost:" + mockService.port() + getCapabilitiesURL());
```
aaime, line 58: Every single usage of getCapabilitiesURL is a copy/paste of this concatenation. I'd recommend merging everything inside getCapabilitiesURL then.

src/extension/mapml/src/test/java/org/geoserver/mapml/MapMLWMSDataAccessProxyTest.java
```diff
@@ -0,0 +1,108 @@
...
+        tam.putLimits(
+                username, layer2, new DataAccessLimits(CatalogMode.CHALLENGE, intersectsFilter));
+    }
+
+    @Test
+    public void testMapMLNotCascadingWithAccessLimits() throws Exception {
+        // get the mapml doc for the layer
+        String path = BASE_REQUEST.replace("cascadedLayer", "cascadedLayerAccessLimits");
+
+        checkCascading(path, false, "image");
```
aaime, line 106: "image" and "tile" should probably be constants in the base class

src/extension/mapml/src/test/java/org/geoserver/mapml/MapMLWMSDataAccessProxyTest.java
```diff
@@ -0,0 +1,108 @@
...
+        catalog.add(wmsStore);
+
+        // Create A layer with access limits
+        WMSLayerInfo wmsLayer2 = catalog.getFactory().createWMSLayer();
+        wmsLayer2.setName("cascadedLayerAccessLimits");
+        wmsLayer2.setNativeName("topp:states");
+        wmsLayer2.setStore(wmsStore);
+        wmsLayer2.setAdvertised(true);
+        wmsLayer2.setEnabled(true);
+        wmsLayer2.getMetadata().put("mapml.useRemote", true);
```
aaime, line 69: There is already a constant for this one, I would use it in all tests

src/extension/mapml/src/test/java/org/geoserver/mapml/MapMLWMTSProxyTest.java
```diff
@@ -0,0 +1,160 @@
...
+                    + "&format_options="
+                    + MapMLConstants.MAPML_WMS_MIME_TYPE_OPTION
+                    + ":image/png";
+
+    @Override
+    protected void registerNamespaces(Map<String, String> namespaces) {
+        super.registerNamespaces(namespaces);
+        namespaces.put("wmts", "http://www.opengis.net/wmts/1.0");
+        namespaces.put("gml", "http://www.opengis.net/gml");
+        namespaces.put("xlink=", "http://www.w3.org/1999/xlink");
```
aaime, line 49: Typo?

```suggestion
        namespaces.put("xlink", "http://www.w3.org/1999/xlink");
```

src/extension/mapml/src/test/java/org/geoserver/mapml/MapMLWMTSProxyTest.java
```diff
@@ -0,0 +1,160 @@
...
+                    + "&SRS=MapML:OSMTILE"
+                    + "&BBOX=-1.3885038382960921E7,2870337.130793682,-7455049.489182421,6338174.0557576185"
+                    + "&WIDTH=768"
+                    + "&HEIGHT=414"
+                    + "&format_options="
+                    + MapMLConstants.MAPML_WMS_MIME_TYPE_OPTION
+                    + ":image/png";
+
+    @Override
+    protected void registerNamespaces(Map<String, String> namespaces) {
```
aaime, line 45: Wondering if all the namespaces can just be declared in the base class... or are there conflicts?

src/extension/mapml/src/test/java/org/geoserver/mapml/MapMLWMTSProxyTest.java
```diff
@@ -0,0 +1,160 @@
...
+    public void testRemoteVsNotRemote() throws Exception {
+        Catalog cat = getCatalog();
+        // Verify the layer was added
+        LayerInfo li = cat.getLayerByName("cascadedLayer");
+        assertNotNull(li);
+        assertEquals("cascadedLayer", li.getName());
+
+        ResourceInfo layerMeta = li.getResource();
+        layerMeta.getMetadata().put("mapml.useRemote", false);
+        layerMeta.getMetadata().put("mapml.useTiles", true);
```
aaime, line 116: This one should have a constant already available too I believe (not verified though).

aaime (changes requested):
Overall looks good, see comments below

src/extension/mapml/src/main/java/org/geoserver/mapml/MapMLRequestMangler.java
```diff
@@ -0,0 +1,675 @@
...
+                        "EPSG:3857"));
+        crsMappers.add(
+                new CRSMapper(
+                        Set.of("EPSG:5936", "urn:ogc:def:crs:EPSG::5936", "MapML:APSTILE"),
+                        "EPSG:5936"));
+        crsMappers.add(
+                new CRSMapper(
+                        Set.of("EPSG:3978", "urn:ogc:def:crs:EPSG::3978", "MapML:CBMTILE"),
+                        "EPSG:3978"));
+    }
```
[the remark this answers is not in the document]
dromagnoli: Yep Indeed. This has been setup at the very beginning of the development, even before I was aware of the task of custom TCRS support. I will check that indeed.

src/extension/mapml/src/test/java/org/geoserver/mapml/MapMLBaseProxyTest.java
dromagnoli: Right. No idea where did that come from :)

src/extension/mapml/src/test/java/org/geoserver/mapml/MapMLBaseProxyTest.java
dromagnoli: Ah. I didn't know about that. Nice. Updating it accordingly.

src/extension/mapml/src/main/java/org/geoserver/mapml/MapMLRequestMangler.java
```diff
@@ -0,0 +1,675 @@
...
+                            }
+                        }
+                    } catch (IOException e) {
+                        reason =
+                                "Unable to extract the WMS remote capabilities. Cascading won't be performed";
+                        LOGGER.warning(reason + "due to:" + e);
+                        cascadeToRemote = false;
+                    }
+                } else if (storeInfo instanceof WMTSStoreInfo) {
+                    WMTSStoreInfo wmtsStoreInfo = (WMTSStoreInfo) storeInfo;
```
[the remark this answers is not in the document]
dromagnoli: There was a stashed commit, applying the same logic. I have restored it.

prushforth:
Hi Daniele,  Sorry for this late comment.  I noticed that when I set up this wms https://agriculture.canada.ca/atlas/services/imageservices/annual_crop_inventory_2023/ImageServer/WMSServer?request=GetCapabilities&service=WMS as a proxied layer, that the GetFeatureInfo requests go to localhost not to the original domain.  Also, the localhost getfeatureinfo raises an exception because the request is missing the QUERY_LAYERS parameter:  Either no layer was queryable, or no layers were specified using QUERY_LAYERS

aaime:
@prushforth that server does not even declare GetFeatureInfo support, in the capabilities, which also implies we don't know which formats are supported. Proxying is thus impossible. Having a fallback to the local server might not be the best option (in this case it will fail), but I also don't want to lose the option of having a working GetFeatureInfo when the remote server can do GML, but not MapML: the local GeoServer can turn that GML into whatever format available, but the MAPML client cannot use it directly, no?

prushforth:
I thought it should not generate a map-link rel=query when the remote layer is not queryable.

aaime:
I've discussed it with Daniele, the situation is as follows:

image

1) If the MapML client can talk directly to the remote server (queriable and supported format) it will
2) If not, but GeoServer can do a GetFeatureInfo translating the eventual remote GML format support, the client will ask GFI to the local GeoServer
3) Otherwise, no link is going to be implemented

prushforth:
Sounds good, thanks

src/extension/mapml/src/main/java/org/geoserver/mapml/MapMLDocumentBuilder.java
```diff
@@ -1695,17 +1661,33 @@ private void generateWMTSQueryClientLinks(MapMLLayerMetadata mapMLLayerMetadata)
...
-                                    baseUrlPattern, path, params, URLMangler.URLType.SERVICE),
-                            "UTF-8");
-        } catch (UnsupportedEncodingException uee) {
+        MapMLRequestMangler mangler =
+                new MapMLRequestMangler(
+                        mapContent, mapMLLayerMetadata, baseUrlPattern, path, params, proj);
+        String urlTemplate = mangler.getUrlTemplate();
+        // It may be that the mangler decided to not generate any query URL due
+        // to unsupported info formats from the remote layer. So we are not
+        // generating the query link.
```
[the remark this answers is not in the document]
dromagnoli: Fixed as result of the discussion about different support types of GetFeatureInfo

prushforth:
Thank you

aaime (approved):
LGTM

PR #7942, opened 2024-10-08
Add missing JARs to assembly.xml file for cog-s3 plugin

mhn17:
The dependency `AWS SDK` of the `cog-s3` plugin requires additional libraries after it was upgraded from version 2.24.13 to 2.27.23.

This fixes the issue discussed in https://discourse.osgeo.org/t/missing-jars-for-cog-s3-plugin-in-assembly-xml/110332

# Checklist


aaime:
@groldan could you have a look?

groldan:
yes I will asap in case the azure one is also missing something

groldan (approved):
Tested on main and on 2.26.0 (re-building, the same package won't work)

PR #7959, opened 2024-10-15
Fix Eclipse IDE compilation errors

sikeoka:
This PR fixes some Eclipse IDE compilation errors in some web modules after the Wicket 9 upgrade. This PR only fixes the core web modules and the WPS extension. Other extension and community modules may still have errors in Eclipse.

  

# Checklist


aaime:
Thanks for this PR. 

I'm wondering about makig sure we're not breaking Eclipse users again... which could be done if we had a Github that would build with the Eclipse compiler. I've found this post which could be of inspiration:

https://stackoverflow.com/questions/33164976/using-eclipse-java-compiler-ecj-in-maven-builds

Wondering if we could limit this build to just run compile, without tests... or do you think even the compiled code behavior might be different when using the Eclipse compiler?

sikeoka:
Can this PR be merged? I think these Eclipse specific errors are rare enough that you don't need to add another check.

aaime:
Go ahead

PR #7965, opened 2024-10-18
[GEOS-11624] Split geopackage extension into three (sub)modules

woutergd:
As I am working on implementing the Geopackage Output extension into the Geoserver Cloud project, I found out that this was not easily possible as it was a single module that imported the GWC, WMS and WFS modules. To make it a better fit, hereby I propose the changes to split this extension into three separate modules: a wms and wfs module, and a core module that support both of them.

# Checklist


aaime:
Sure, I don't see a problem here. By looking at ext-geopkg-output.xml it seems the assembly will still catch the right jar files. @jodygarnett what do you think?=

jratike80:
Will the WFS submodule take care of the OGC API Features GeoPackage output if that will be supported in the future?

aaime:
It should... in theory even today, when installing both the OGC API features package and the Geopackage output format, the GeoPackage format should become available in OGC API features. But I haven't tried.

jodygarnett:
It should be fine, a bit of a pain to have more jars; but it is maven ...

woutergd:
Sorry, had not enough time last weeks to continue on this PR. Good to hear that these changes are not expected to be an issue.

I noticed that I have for some strange reason failures on the JDK 17 and 21 builds, is there someone who can point me in the right direction to resolve these build errors? I will try also to fix this week the merge conflicts, so this PR can move forward.

aaime:
You can safely ignore those failures, they are happening sometimes on the Github actions but we cannot reproduce them on developer machines. But please make sure you rebase this PR on top of the latest main branch, as we try to mitigate those failures over time by giving tests more leniency (some are time based, others will retry on lock failures, and so on).

woutergd:
@aaime thanks for the info. I rebased this PR, and also created a JIRA-issue as was suggested to me in Discourse as something that was required to provide a PR.

I noticed however that after the rebase I have different / new tests that are failing. Are these also known? Otherwise I will make some time this week to investigate what the issue is for these build errors.

aaime:
Had a look, everything seems to check out. Merging.

PR #7968, opened 2024-10-21
GEOS-11272: spring-security-oauth replacement, with spring-security 5.8

awaterme:
[GEOS-11272](https://osgeo-org.atlassian.net/browse/GEOS-11272) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

Hi community,

this is a pull request for https://osgeo-org.atlassian.net/browse/GEOS-11272.

# Checklist

- [x ] Bug fixes and small new features are presented as a single commit.

# GeoServer OpenID Connect Rewrite

**Status:** January 2025

## Current Status

- The implementation is ready for review and public testing.
- Existing unit tests have been ported and supplemented, except for functionality that has been dropped (see details below). Test coverage (excluding "Resource Server" UCS, see below)
  - 82% of instruction for gs-sec-oidc-core
  - 93% of instruction for gs-sec-oidc-web

## Features

- Working with Google, GitHub, Microsoft Azure, and one custom OIDC provider.
- The "Resource Server" functionality (i.e., "OpenID Connect With Attached Access Bearer Tokens") was originally available. However, it was decided not to support this feature for the time being, as a separate extension (~ `gs-sec-jwt`) already provides similar functionality.
  - The current extension still contains code for the "Resource Server" use case, but it is in an initial status.
  - A first test of the feature was successful, but the functionality is currently disabled (commented out in `application.xml`).
  - Consider removing this code if it remains unused (though this might be unfortunate).

## Design & Goals

- Leverage Spring’s public API to configure Spring filters and associated classes for a future-proof solution.
  - Filters are created by Spring and integrated into GeoServer, minimizing custom setup code.
- While Spring is not inherently designed for the dynamic behavior required by GeoServer, the approach still appears reasonable.
- Avoid "fishing" for request parameters and headers.

### References

- For further information, refer to JavaDocs:
  - `org.geoserver.security.oauth2.login.GeoServerOAuth2LoginAuthenticationProvider`
  - `org.geoserver.security.oauth2.login.GeoServerOAuth2LoginAuthenticationFilter`
- See the class diagram in `gs-sec-oidc/doc/diagrams` for orientation.

## Installation

- The project includes an assembly module. Place the JAR files in `WEB-INF/lib`.

## Configuration

1. Create a filter named **"openid-connect"** using the Admin Web UI.
2. Add the **"openid-connect"** filter to the `web` filter chain.
3. Create a new filter chain named **"oauth-callback"** with the following configuration:
   - **Ant Patterns:** `/oauth2/authorization/**,/login/oauth2/code/**` (Attention: no space between)
   - Assign the **"openid-connect"** filter to this chain.
   - Use default settings for everything else.
   - Ensure the chain is positioned before `webLogin`.
4. If using a user group service as the role source, create users with roles that correspond to the identity provider's users.
5. Set up the identity provider (Google, GitHub, Microsoft, or custom OIDC provider) as usual.

### High Relevance

- Update the user guide.
  - generally
  - specifics:
    - special cases for `tokenRolesClaim` (`scope` for source `access token` and `authorities` for source `userInfo`, see `GeoServerOAuth2RoleResolver`)
- Implement GEOS-11635: _"Add support for opaque auth tokens in OpenID Connect"_.
- Verify compatibility with Keycloak and GeoNode.

### Medium Relevance

- Validation: Previously `checkTokenEndpointUrl` or `jwkURI` was required, see prior OpenIdConnectFilterConfigValidator. Situation is a little confusing.

  - What I understand:
    - `checkTokenEndpointUrl` refers to `userInfoUri` (according to Spring naming). API purpose: Exchange an access token into userInfo data.
    - `jwkURI` refers to `jwkSetUri` (according to Spring naming). API purpose: Load IDP public keys for signature validation.
    - I suppose the prior code is using one of those methods to validate the access token.
  - The current Spring approach is different:
    - Spring uses the `jwkSetUri` for validation by default, which is determined by the selected JwsAlgorithm, which defaults to RS256. This algorithm belongs to the family of signature algorithms using asymetric encryption and public/private keys. Therefor the public keys are loaded from the `jwkSetUri`.
      - If an algorithm of the MAC family is configured instead (regarding configuration: see below) the signature is validated using a pre-shared secret instead (here: the client secret).
    - so in neither case the `userInfoUri` is used for token validation. Even the `jwkSetUri` is not necessarily used - it depends on the algorithm.
    - however the `userInfoUri` is used by GeoServer if `UserInfo` is the selected role source.
    - I think this is only affecting the validations which currently do not reflect this. Also, this should be explained in the user guide. I suppose:
      - the config UI should be extended to select an algorithm, RS256 by default (see below)
      - if the algorithm belongs to the signature family the `jwkSetUri` is mandatory, otherwise it is optional
      - if `UserInfo` is selected as role source the `userInfoUri` is mandatory, otherwise it is optional

- Create integration tests:
  - Consider using [Spring Authorization Server](https://spring.io/projects/spring-authorization-server) as an OIDC provider. While simple to use, it requires newer Spring versions, making the setup currently more complex.
  - An example setup is available in the [gs-sec-oidc-integration-tests repository](https://github.com/awaterme/gs-sec-oidc-integration-tests). This setup is also useful as a lightweight development OIDC provider.

### Lower Relevance

- Improve unit test coverage
- Some validators have been removed:
  - `AudienceAccessTokenValidator`: Likely replaced by `OidcIdTokenValidator`.
  - `SubjectTokenValidator`: Previously used only for the Resource Server use case?
- Consider implementing Andrea’s suggestion for a dropdown in the UI to "add provider xy." However, in the current implementation deactivated providers are hidden, reducing the relevance of this feature.

## Compatibility with Prior Implementation

- **Role Sources:**
  - Previous implementations may have supported nested JSON paths. It’s unclear if this was intentional or necessary.
- **Token Validation:**
  - Previously, invalid signatures may have been accepted if "enforce token validation" was set to `false`.
  - Now, invalid signatures are always rejected (a reasonable change).
  - "Enforce token validation" now only tolerates invalid claims.
- Other potential differences in behavior may exist.

## Next Steps

1. Conduct a thorough review.
   - Note: As mentioned earlier, I am not a trained security specialist.

## Future Ideas

- Introduce additional OIDC protocol configurations, such as:
  - Allowing users to specify the `JwsAlgorithm` (this has been newly introduced and is now part of the configuration but currently without a UI counterpart).
  - I wonder if using the wrong algorithm in the past led to the requirement to make "force token validation" optional.
- Improve parsing of the `.wellknown-operations` endpoint:
  - The `JwsAlgorithm` and other details might be automatically detectable.
- Provide an "Apply Preset" UI action for certain identity providers (e.g., ADFS), which pre-fills settings for those providers.
- Document available claims and their semantics for each identity provider as thoroughly as possible.


aaime:
Hi, I've tried to take this one for a spin. Some feedback.

The GUI is organized now in panels that can be enabled and disabled... this can be confusing, one would not enable more than one:

image

I'd suggest to have a drop-down selection at the top, which allows to choose the one configuration desired. And then the associated panel would appear.

Working on the custom connection, I've noticed that the discovery does not seem to work... actually it works code wise, but it does not appear inside the browser. I could not figure out why thought.
I've manually filled the entries, and saved without issues... but then when trying to put the filter in a chain, it failed... turns out the configuration is not validated when pressing save, but when trying to use it: it should be done when hitting save on the configuration panel instead, so that one can act directly, and not just get a stack trace in the UI.

Also, the misconfiguration in my case turned out the scopes, which are supposed to be comma separate... but the OIDC world separates scopes by spaces, why change that? The old UI also had them space-separated.

Finally, I've managed to configure the filter in the web chain, and pressing on the OIDC button, I'm redirected to "http://localhost:8080/geoserver/oauth2/authorization/oidc?"... but this path has no handler, so nothing happens. 
Maybe this part has not been fleshed out yet?

awaterme:
Hello Andrea,

thank you for your feedback.

I still have a lot of work to do on the Config UI. I'll be happy to look at your points in this context.

I was planning to only display the input per provider if the “enabled” checkbox is activated. Then the UI should be much clearer. Personally, I find the drop-down method rather indirect and therefore less intuitive. But I'm dispassionate about it. However, I would take care of the functional things first, knowing that an intuitive UI is also crucial.

Regarding the filter chain: I have added a missing, new configuration instruction above, which was certainly the cause of your error:

* add a filter chain "oauth-callback":
   * Ant-Patterns: /oauth2/authorization/\*\*, /login/oauth2/code/\*\*
   * assign OAuth2-Filter instance to this chain
   * everything else: default

Maybe try it again with that. If it works then we will all have a better feeling.

Best regards,
Andreas

Translated with DeepL.com (free version)

aaime:
This is likely going to be of interest: https://github.com/geoserver/geoserver/pull/8104
TBH, the amount of options and possible paths in OAuth2 + Openid connect is overwhelming...

Syndlex:
Hey,

Any ideas when this is finished?

jodygarnett:
Hello @awaterme - we are in position to work on GS3 now (there will be some blog posts tomorrow).  Rough idea is to use the test cases on the old implementation to identify any additional features that need to be developed.  Individual team members will introduce themselves here.

Thanks so much for your work; glad to join you on making this happen!

jodygarnett:
> Any ideas when this is finished?

@Syndlex if you have interest and capacity reviewing this pull-request and testing the implementation in your environment would be very helpful.

waterman:
Hi Jody,

> The rough idea is to use the test cases from the old implementation to identify any additional features that need to be developed. Individual team members will introduce themselves here.

There’s one important point I’ve mentioned a couple of times that I’d like to highlight again.

I suppose the old implementation was used for two main purposes:
a) Login
b) Running GeoServer as an OAuth2 Resource Server.

The new implementation currently supports a).

Initial support for b) exists in the new codebase but remains unfinished, as we decided early on to focus primarily on login functionality.
So the code related to b) may either be completed or removed. Additionally, there’s the GeoServer "JWT-something" extension, which (as far as I understand) provides support for b).

As a result, the old test cases are only partially relevant.
I have already ported the automated tests that existed at the time I started the new branch. I believe I dropped some tests related to b).

By the way: When you refer to "test cases", do you mean the automated JUnit tests? Or are you referring to something else?

aaime:
Support for b) is quite important and (AFAIK) only the vendor contributing the jwt module is using it. So we'll have to complete it and make sure all existing test pass. Thanks for reminding us, we'll have to plan accordingly.

waterman:
One additional thought: I recently came across [Matt Raible](https://github.com/mraible). I don’t know him personally, but he’s an open source contributor to JHipster and, as far as I know, works at Okta. He appears to have deep knowledge of OAuth2, and likely Spring Security OAuth2 as well.

Perhaps he would be available for a review under whatever conditions might apply — which could be a valuable step from a security perspective.

pmauduit:
Hi, we attempted to test the PR with @vuilleumierc, and it looks great, we were able to:
* rebase onto the main branch
* compile
* launch the jetty maven plugin to runtime test it

For now on, the 2 related modules (oidc-core, oidc-web) announce themselves onto the admin UI, and we can configure a OIDC provider. Adding a OIDC filter to the filter chain allows us to see the openid button next to the login form:

Screenshot from 2025-05-15 11-33-52

Unfortunately clicking on it has no effect and leads to the following error being displayed in the developer console:

```
Refused to send form data to 'http://localhost:8080/geoserver/oauth2/authorization/oidc?' because it violates the following Content Security Policy directive: "form-action 'self'".
```

We believe that this is somehow due to a recent upgrade of Wicket and Content-Security-Policy enforcement. Deactivating CSP into the web UI made it. But we could not find a sample OIDC provider which would allow us to have redirect URLs in plain http (http://localhost:8080, as we were using the jetty maven plugin).

But so far, the configuration looks quite straightforward. The rebase onto main was quite trivial, but if needed our work branch based on this PR is here: https://github.com/camptocamp/geoserver/tree/GEOS-11272-OIDC-Mig-c2ctest

pmauduit:
> But so far, the configuration looks quite straightforward.

We were able to launch a local keycloak and configure GS to authenticate onto it.

vuilleumierc:
Thanks @awaterme / @waterman for this huge work. I have been doing some further testing (using a GeoServer role, role from the access token, role from the user info) and it works well.
The next step I think should be to merge this before porting the existing tests and missing functionalities. Since it's a new community module at this stage it's pretty low risk but I am wondering about the changes in the main module (classes `GeoServerRoleResolvers` and `GeoServerPreAuthenticatedUserNameFilter`). Are these changes well tested? I will try to figure this out but if you have some more info it's welcome.

waterman:
> Thanks @awaterme / @waterman for this huge work. I have been doing some further testing (using a GeoServer role, role from the access token, role from the user info) and it works well. The next step I think should be to merge this before porting the existing tests and missing functionalities. Since it's a new community module at this stage it's pretty low risk but I am wondering about the changes in the main module (classes `GeoServerRoleResolvers` and `GeoServerPreAuthenticatedUserNameFilter`). Are these changes well tested? I will try to figure this out but if you have some more info it's welcome.

Hi @vuilleumierc, thank you, good to hear it is working for you.
I have no special remarks regarding the changes in gs-main. Although I have to admit I am not entirely happy with the structures in this area as a whole. I think the inheritance based design has not evolved into the right direction. But I found the changes relatively straight forward and I think the risk of causing unexpected side effects is low because the actually methods are still in place as far as I remember.

jodygarnett:
Replaced by https://github.com/geoserver/geoserver/pull/8569 which has been refactored.

PR #7972, opened 2024-10-22
[GEOS-11580] Improve embedded GWC meta-tiling performance

aaime:
[GEOS-11580](https://osgeo-org.atlassian.net/browse/GEOS-11580) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

Replacement for:

https://github.com/geoserver/geoserver/pull/7457

Compared to the original PR:
* Dusted off and aligned to GeoServer's current code
* QA run (formatting, PMD, error-prone)
* Improved UI, documentation
* Confirmed the performance improvements compared to the current GeoServer code (which already has two of the improvements mentioned in #7457), it's faster to return tiles, and under load (many concurrent requests for un-cached tiles) it's still 10% faster.

# Checklist

- [x] First PR targets the `main` branch (backports managed later; ignore for branch-specific issues).


groldan:
this looks very good at a first glance, I'd like to give it a deeper look. Wondering if locking at the tile level could be too much for a clustered set up where the gwc lock provider is cluster-aware.

groldan:
> But also.... are you using embedded GWC in GeoServer cloud to start with? The change here is for GeoServerTileLayer

yes indeed. The gwc-service in GeoServer Cloud are stripped down versions of GeoServer, with just the gwc components. They use their own instance of the wms service just as usual, as a rendering engine, not exposing the WMS OWS api. This is so gwc pods can function autonomously without calling the wms over http.

Also, in gscloud, the geoserver LockProvider oughta be always cluster aware, depending on the catalog backend, it'll be file-based or in the database (both for pgconfig and jdbcconfig), and the gwc LockProvider is always GeoServerLockProvider delegating to the ResourceStore's lock provider.

So locking at the tile level is certainly concerning, especially thinking of production deployments with 10x10 meta-tiling factors.

But as said, I'd like to give it a deeper look and maybe test the integration with gscloud

groldan:
One option it just occurred to me, would be to keep the meta-tile locking at the LockProvider level but only use in-process (single JVM) locking for the single tile locks. Makes sense or am I missing something?

aaime:
Let's see. In a normal situation, tiles of the same meta are likely requested at the same time, ideally, we'd like to compute the meta tile, encode the PNGs and save them on disk just once. 

The current code tries to return each tile to the client as soon as possible operating as follows:

* Get the metatile lock
* Compute the meta tile as an in-memory image
* Prepare a clip/encode/save task for each tile, but then:
   - for the requested tile, execute clip and encode in memory, and push save to the background pool with a tile lock around it
   - for all the other tiles, clip, encode and save are all pushed in the background pool with a tile lock around them
* Release the metatile lock and return the PNG (possibly  only available in memory at that point in time) to the client

The other tiles of the same meta at that point will be released from the meta lock, and will try to get their primary tile, blocking on the tile lock until the background pool has managed that one tile on disk (so they wait for their own tile, not for the entire meta to be done).

This minimizes the client wait time, at the expense of having more locks.

Now, as you say, we could have in-memory locks for the single tiles. That works perfectly as long as there is only one GWC+WMS in the cluster. When you have more than one, the load balancer has likely spread the tiles from the same meta on multiple nodes, causing the metatile computation and all the background encoding and saving to happen again.

So it looks a matter of balance... what's more expensive, duplicating the same work across nodes, or managing a large number of locks?

src/gwc/src/main/java/org/geoserver/gwc/layer/GeoServerTileLayer.java
```diff
@@ -579,57 +587,303 @@ protected final void sendTileRequestedEvent(ConveyorTile tile) {
...
+            return finalizeTile(conveyorTile);
         }
 
-        final GeoServerMetaTile metaTile = createMetaTile(tile, metaX, metaY);
-        Lock lock = null;
+        final GeoServerMetaTile metaTile = createMetaTile(conveyorTile, metaX, metaY);
+
+        /* ****************** Acquire lock on metatile ******************* */
+        final Lock metaTileLock =
+                GWC.get().getLockProvider().getLock(buildMetaTileLockKey(conveyorTile, metaTile));
```
groldan, line 602: since `GWC.get().getLockProvider().getLock(...)` is a repeating pattern, can we extract it out to a `getLock(String key)` method? (it'd improve readability as I don't need to know how it's acquired everytime while glimpsing over the code)

src/gwc/src/main/java/org/geoserver/gwc/layer/GeoServerTileLayer.java
```diff
@@ -579,57 +587,303 @@ protected final void sendTileRequestedEvent(ConveyorTile tile) {
...
-                                + Arrays.toString(metaTile.getMetaGridPos())
-                                + " on "
-                                + metaTile);
-                WebMap map;
+            boolean foundInCache = false;
+            if (tryCache) {
+                /* ****************** Acquire lock on individual tile ******************* */
+                // Will block here if there is an async thread currently saving this tile
+                String lockKey = buildTileLockKey(conveyorTile, conveyorTile.getTileIndex());
+                final Lock tileLock = GWC.get().getLockProvider().getLock(lockKey);
```
groldan, line 610: ok so for the on-demand seeding this looks good. At this point we'll hold both the metatile and tile locks.
For the seeding jobds though it looks like we'll be acquiring the tile locks anyway, as called by `SeedTask`: 
```java
boolean tryCache = !reseed
...
tl.seedTile(tile, tryCache);
```
I think we should prevent acquiring the tile lock as in `computeMetatile()` below (`if (Dispatcher.REQUEST.get() == null)` then don't lock the tile)

groldan (changes requested):
Hi Andrea, just some minor comments.
Given this change is only for on-demand metatile seeding, provided we also don't lock on the tile during explicit seeding I think it'll be good enough also for the distributed application scenario.

src/gwc/src/main/java/org/geoserver/gwc/GWC.java
```diff
@@ -1235,6 +1261,12 @@ public void saveConfig(GWCConfig gwcConfig) throws IOException {
 
         // make sure we switch to the lock provider just configured
         updateLockProvider(gwcConfig.getLockProviderName());
+
+        // Reconfigure the metatiling executor because the thread count might have changed
+        if (this.metaTilingExecutor != null) {
+            this.metaTilingExecutor.shutdown();
+        }
+        this.metaTilingExecutor = buildMetaTilingExecutor(gwcConfig.getMetaTilingThreads());
```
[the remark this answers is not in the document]
aaime: Yes, good idea

src/gwc/src/main/java/org/geoserver/gwc/layer/GeoServerTileLayer.java
```diff
@@ -579,57 +587,303 @@ protected final void sendTileRequestedEvent(ConveyorTile tile) {
...
+                                            resource,
+                                            requestTime));
+                    if (executor != null) {
+                        CompletableFuture<Void> completableFuture =
+                                CompletableFuture.runAsync(saveTileTask, executor);
+                        completableFutures.add(completableFuture);
+                    } else {
+                        // Save in cache on main thread if there's no executor
+                        saveTileTask.run();
+                    }
```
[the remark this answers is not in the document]
aaime: Sure

src/gwc/src/main/java/org/geoserver/gwc/layer/GeoServerTileLayer.java
```diff
@@ -579,57 +587,303 @@ protected final void sendTileRequestedEvent(ConveyorTile tile) {
...
+
+                    if (executor != null) {
+                        // Fully asynchronous
+                        CompletableFuture<Void> completableFuture =
+                                CompletableFuture.runAsync(encodeAndSaveTask, executor);
+                        completableFutures.add(completableFuture);
+                    } else {
+                        // Run on main thread if there's no executor
+                        encodeAndSaveTask.run();
+                    }
```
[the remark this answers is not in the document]
aaime: roger

src/gwc/src/main/java/org/geoserver/gwc/layer/GeoServerTileLayer.java
aaime: Yes

src/gwc/src/main/java/org/geoserver/gwc/layer/GeoServerTileLayer.java
aaime: Yep, I've reshuffled code around so that the tile lock is not taken in that situation.

aaime:
@groldan any more feedback, or do you think it's good to go?

groldan (approved):
LGTM

aaime:
Merged. I would also like to backport it to 2.26.x. I'll leave it for a month on the main branch and then ask for a backport.

PR #7976, opened 2024-10-23
Fixed some Wicket CSP violations

sikeoka:
This PR fixes some CSP issues with after the Wicket 9 upgrade. The big one is with the node info since that has an API change.

The org.geoserver.web.csp.strict property isn't working correctly and this PR fixes it so that it actually disables blocking CSP violations by default. My separate CSP PR that Jody is reviewing allows an administrator to switch between blocking or reporting headers or completely disabling CSP through the UI.

The period in front of CSS class names actually matters and isn't just a code style issue. An easy way to to see this is that the Style Editor is missing a black border around the main text editor (at least in Chrome).

I'm only reviewing a few extensions and no community modules but besides the modal dialog and the Tile Layers page, the remaining CSP violations I've encountered are with the CodeMirror library used by the Style Editor and Demo Requests page (specifically the Find, Find Next, Replace buttons) and the Spectrum.js library used by the Choose Color button in the Style Editor. I'm not sure what can be done about them since they are third-party libraries and upgrading the libraries wouldn't fix the CSP issues.

A separate PR was created to fix the Tile Layers page since it already had a JIRA ticket: https://github.com/geoserver/geoserver/pull/7977

  

# Checklist

### DIFF
--- a/src/extension/monitor/core/src/main/java/org/geoserver/monitor/web/ActivityChartBasePanel.html
+++ b/src/extension/monitor/core/src/main/java/org/geoserver/monitor/web/ActivityChartBasePanel.html
@@ -1,22 +1,14 @@
 <html xmlns:wicket="http://wicket.apache.org/">
 <head>
 <title><wicket:message key="title">Activity</wicket:message></title>
-<wicket:head>
-  <style type="text/css">
-   form label {
-      display: inline;
-   }
-  </style>
-</wicket:head>
-
 </head>
 <body>
 <wicket:panel>
      <form wicket:id="form" class="pb-1 ps-5">
-        <label for="from">From</label>
+        <label for="from" class="inline">From</label>
         <input type="text" wicket:id="from" id="start" class="pe-4 d-inline"></input>
         
-        <label for="to">To</label>
+        <label for="to" class="inline">To</label>
         <input type="text" wicket:id="to" id="end" class="d-inline"></input>
         
         <a href="#" wicket:id="refresh" class="refresh-link">Refresh</a>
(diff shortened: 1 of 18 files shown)


sikeoka:
Yes, the CSP violations were fixed by moving the styles into the CSS file but they weren't properly declared so the styles weren't actually being applied.

CodeMirror and Spectrum both use inline styles which will cause CSP violations if you click on one of the buttons I mentioned. The CodeMirror issue you linked to does seem like it can solve the problem there but someone would have to upgrade the library to a current version. The CodeMirror.fromTextArea function was removed which would complicate the upgrade. It would also take some investigating to determine if a Wicket page can even get access to the CSP nonce. I don't see any equivalent functionality in 
Spectrum.

aaime:
Yeah [Spectrum appears to be pretty dead](https://github.com/bgrins/spectrum), last commit is from a few years ago... thought it seems to have [a fork that's more alive](https://seballot.github.io/spectrum/) I could not confirm CSP support in it.

There are maintained color pickers that do have support for CSP, found for example [vanilla-picker](https://github.com/Sphinxxxx/vanilla-picker) that had [a ticket solving CSP support](https://github.com/Sphinxxxx/vanilla-picker/issues/45). Just a quick search, there could be better pickers out there that also support CSP, this is the first one I've found.

sikeoka:
Spectrum is a single, non-minified javascript file and all of the CSP violations actually come from a single function so it is possible for me to patch the file to work without CSP violations if you didn't have a better option.

sikeoka:
I can create separate PRs to patch Spectrum and CodeMirror so that they work with the CSP. CodeMirror should probably be updated instead of patched but that would be a major upgrade and I don't have time to look into that.

With all of those fixes, everything in the base GeoServer WAR and the few extensions that I'm testing should be working fine with a blocking CSP except for the Wicket dialog. I don't know what the plan to deal with that is.

aaime:
Separate PRs for those two are good.

About the dialog... the plan is to rewrite it using Wicket 9+ dialog base classes support, but it's a big effort, the support in the new Wicket is very low-level. So that will likely happen along with the Wicket 10 upgrade, as part of the GeoServer 3 "big bang" code sprint.

This particular part doe snot need to wait till then, but unless someone shows up to create a GeoServerDialog replacement class based on the new Wicket support, well... we'll have to put up with the CSP warnings for the time being.

aaime:
Gave it a spin, looks good, nice work. Merging.

PR #8014, opened 2024-11-07
Introduce Conformance API to configure OGCAPI Features

jodygarnett:
Looking at how to add conformance to the ogcapi-core as something we can configure on and off.

Registers an AdminPagePanel with WFSInfo to be shown on the WFS page:

image

Feedback welcome.

# Checklist


src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/APIConformance.java
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * OGC Open Web Services can be extended using application profiles with additional, optional, functionality.
+ */
+public class APIConformance {
+
+    /**
+     * Standards approval status.
+     */
+    public enum Status {
+        APPROVED, DRAFT, INFORMAL
```
aaime, line 19: The three statuses I know of are Approved, Draft and Proposal. However there is a lot of action happening outside of OGC and we are used to add our own vendor extensions too
With basically no-one in our community having time to also participate in working groups I would probably address the representation of the status as two enums:

Stability: finalized, draft (whether the conformance is yet subject to change)
Source: OGC, Vendor (whether it's an official standard or coming from outside)

That said... I believe in terms of defaults and user concerns, only the first one is relevant... can one rely on the behavior of a given conformance class long term, or not?  That basically reduces what we need to communicate to a boolean. The source can be an interesting informative notion, but that's it.

src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/APIConformance.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public enum Type { CORE, EXTENSION }
+
+    final APIConformance parent;
```
aaime, line 24: Do conformances have a "parent"? Don't think they extend in terms of inheritance, but surely some make sense only when a given set of other conformance classes have been implemented (eventually just one, typically the "Part 1, core" of the relevant service).

Some however exist on their own, typically the ones in OGC API commons. E.g., the JSON conformance class for OGC services is recommended but not required.

That said, wondering if it's something we need to communicate. I see the field, but no getter, and no usage.

src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/ConformanceInfo.java
```diff
@@ -0,0 +1,91 @@
+package org.geoserver.ogcapi;
+
+import org.geoserver.config.ServiceInfo;
+
+/**
+ * Configuration for a service capability or feature, identified by conformance class.
```
aaime, line 6: I'd clarify that this is indicate the presence and configuration of a given conformance class in a specific service. E.g., one will have the sortby conformance class in both STAC, Features and Records.

The HTML and JSON conformance classes are likely going to be found everywhere, but it's not something we likely want to expose as configurable. Same goes for the various Part 1, they are there because the service is there, if one does not want it, the service is not enabled (that's however normally controlled by the enabled flag in the ServiceInfo)

src/community/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/FeatureService.java
```diff
@@ -106,19 +96,38 @@ public class FeatureService {
...
+    public static final APIConformance CQL2_ARITHMETIC = new APIConformance(ConformanceClass.CQL2_ADVANCED);
+    public static final APIConformance CQL2_BASIC = new APIConformance(ConformanceClass.CQL2_BASIC);
+    public static final APIConformance CQL2_BASIC_SPATIAL = new APIConformance(ConformanceClass.CQL2_BASIC_SPATIAL);
+    public static final APIConformance CQL2_FUNCTIONS = new APIConformance(ConformanceClass.CQL2_ADVANCED);
+    public static final APIConformance CQL2_PROPERTY_PROPERTY = new APIConformance(ConformanceClass.CQL2_PROPERTY_PROPERTY);
+    public static final APIConformance CQL2_SPATIAL = new APIConformance(ConformanceClass.CQL2_SPATIAL);
+    public static final APIConformance CQL2_TEXT = new APIConformance(ConformanceClass.CQL2_TEXT);
+
+    // informal
+    public static final APIConformance ECQL = new APIConformance(ConformanceClass.ECQL, INFORMAL);
```
aaime, line 121: Vendor, but stable. I believe that's what's important in user eyes, can they rely on the implementation long term, or not? "Informal" does not say, it can be both a vendor experiment (e.g. ogcapi-changesets) or a long term stable feature of the product (e.g., ECQL)

src/web/core/src/test/java/org/geoserver/web/admin/StatusPageTest.java
```diff
@@ -178,7 +178,7 @@ public void testModuleStatusPanelVersion() {
             index++;
         }
         if (found == -1) {
-            fail("Module jvm not found, required for version check");
+            fail("ServiceConformance jvm not found, required for version check");
```
aaime, line 181: I believe this is not intended, probably result of search and replace.
The status page reports the presence of GeoServer own modules, core, extensions and community modules, which might have, or not, a relationship to conformance classes.

aaime (commented):
I like the initial idea, looking forward to see some GUI to configure, how it's stored, and how it's eventually used to control the service behavior.

Left a bunch of comments below.

src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/APIConformance.java
jodygarnett: I also wish to have an "ogc only" mode if *strict* is in play? I guess given the building block approach it should not technically matter but ...

src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/APIConformance.java
jodygarnett: There is a factory method, you can see its use here:

```
    public static final APIConformance CORE = new APIConformance("http://www.opengis.net/spec/ogcapi-features-1/1.0/conf/core");
    public static final APIConformance FEATURES_FILTER = CORE.extend(ConformanceClass.FEATURES_FILTER);
```

In reading the standards they often say they are an extension of a specific core.

src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/ConformanceInfo.java
jodygarnett: I need to mark some of the APIConformance as *required*, not subject to enabled/disabled. It would still be nice to list them. 

Somehow I need to turn these things into beans, and I am not quite sure where to begin.

src/community/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/FeatureService.java
jodygarnett: Yeah I see your logic; and vendor is consistent with our other extensions.

src/community/ogcapi/web-features/src/main/java/org/geoserver/ogcapi/web/FeatureServiceAdminPanel.html
```diff
@@ -0,0 +1,254 @@
+<html xmlns:wicket="http://wicket.apache.org/">
+<head></head>
+<body>
+<wicket:panel>
+    <h3><wicket:message key="defaultTitle">Feature Service</wicket:message></h3>
+    <p><wicket:message key="configureInformation">Feature Service is a modular OGC API service with "building blocks" that can be enabled and configured for use.</wicket:message></p>
+    <ul>
+        <li>
+            <fieldset>
+                <legend><span><wicket:message key="features">OGC API - Features</wicket:message></span></legend>
```
aaime, line 10: I did not expect to see this page hand-written... actually it might be a problem.
What if a given conformance class is provided through a plugin?

Probably better to have enough configuration to auto-generate this from the conformance classes that might be found in the spring context.

aaime (commented):
Had a look and provided some comments. Will be interesting to see how you want to address turning off some behavior. Or are you settling down to just stop advertising it, but then leave it there?

src/community/ogcapi/web-features/src/main/java/org/geoserver/ogcapi/web/FeatureServiceAdminPanel.html
jodygarnett: I am walking before I can run on this one Andrea. I think the next step is to make a conformance panel that can be used multiple times. 

Want to get something that functions at all here to control the module.

src/community/ogcapi/ogcapi-features/src/test/java/org/geoserver/ogcapi/v1/features/ConformanceTest.java
```diff
@@ -49,15 +46,17 @@ private void checkConformance(DocumentContext json) {
...
+            FeatureConformance.OAS30.getId(),
+            FeatureConformance.HTML.getId(),
+            FeatureConformance.GEOJSON.getId(),
+            FeatureConformance.CRS_BY_REFERENCE.getId(),
+            FEATURES_FILTER,
             FILTER,
             QUERYABLES,
-            FEATURES_FILTER,
-            SEARCH,
+            // IDS,
```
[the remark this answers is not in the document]
jodygarnett: That is correct

src/community/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/CQL2Conformance.java
```diff
@@ -0,0 +1,168 @@
...
+    public static final APIConformance CQL2_TEXT = new APIConformance(ConformanceClass.CQL2_TEXT);
+
+    /**
+     * CQL JSON conformance - not implemented yet (very different from the binding we have)
+     */
+    public static final APIConformance CQL2_JSON = new APIConformance(ConformanceClass.CQL2_JSON, DRAFT_STANDARD);
+
+    // CQL is optional
+    public static final APIConformance CQL2_ADVANCED = new APIConformance(ConformanceClass.CQL2_ADVANCED);
+    public static final APIConformance CQL2_ARITHMETIC = new APIConformance(ConformanceClass.CQL2_ARITHMETIC);
```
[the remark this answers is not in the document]
jodygarnett: The user interface has examples of ones that are hard coded as enabled, shows up as grayed out.

I have not checked how configurable the CQL2 parser is, if the functionality is not configurable these conformances will be hard coded to true. You can see this in the screen snap above where HTML and JSON are built-in.

src/community/ogcapi/web-features/src/main/java/org/geoserver/ogcapi/web/FeatureServiceAdminPanel.html
jodygarnett: If a conformance class is provided through a plugin it provides its own `AdminPagePanel`, as that is the existing UI infrastructure for contributing.

aaime: That is one possibility, but it's work-intensive and clumsy. Alsor remember admin panels are not following an order, you can have panels out of any logical order, and other unrelated extensions have their eventual panel  show up in the middle.

src/community/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/CQL2Conformance.java
```diff
@@ -0,0 +1,244 @@
...
+    /**
+     * Obtain CQL2Conformance configuration for WFSInfo.
+     *
+     * <p>Uses configuration stored in metadata map, or creates default if needed.
+     *
+     * @param wfsInfo WFSService configuration
+     * @return CQL2 configuration
+     */
+    public static CQL2Conformance configuration(WFSInfo wfsInfo) {
+        synchronized (wfsInfo) {
```
aaime, line 70: Not sure this is going to do anything... the WFSInfo you have here should be a modification proxy created when the WFSInfo was requested (which must happen on a per request basis, these objects cannot be stored). In other words, synchronization should not be needed at all, if WFSInfo has been obtained properly (as said, dynamically in the current request, never stored as a field).

src/community/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/FeatureService.java
```diff
@@ -407,26 +417,92 @@ public FeaturesResponse items(
...
+
+            if (featureServiceInfo.filter(wfsInfo)) {
+                if (filterLanguage == APIFilterParser.ECQL_TEXT && !ecqlInfo.text(wfsInfo)) {
+                    LOGGER.warning(
+                            () ->
+                                    "The filter language '"
+                                            + APIFilterParser.ECQL_TEXT
+                                            + "' is not supported by the service, requires "
+                                            + ECQLConformance.ECQL_TEXT.getId()
+                                            + " conformance to be enabled.");
```
aaime, line 448: This bit is duplicated over and over and over... the pattern to report a disabled conformance class seems the same... suggest factoring out a common method. Should also make the code more readable, right now the logging is obfuscating what's going on in the method.

src/community/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/FeatureConformance.java
```diff
@@ -0,0 +1,491 @@
...
+            CORE.extend("http://www.opengis.net/spec/ogcapi-features-2/1.0/conf/crs");
+    public static final APIConformance FEATURES_FILTER =
+            CORE.extend(ConformanceClass.FEATURES_FILTER);
+    public static final APIConformance FILTER = CORE.extend(ConformanceClass.FILTER);
+    public static final APIConformance QUERYABLES = CORE.extend(ConformanceClass.QUERYABLES);
+
+    // draft
+    public static final APIConformance IDS =
+            new APIConformance(ConformanceClass.IDS, DRAFT_STANDARD);
+    public static final APIConformance SEARCH =
```
aaime, line 50: So... when this one is not enabled, the search method in FEatureService should throw a 404?

src/community/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/FeatureServiceXStreamPersisterInitializer.java
```diff
@@ -0,0 +1,20 @@
+package org.geoserver.ogcapi.v1.features;
```
aaime, line 1: Lacks copyright header (QA must be run manually, the module is still community here)

src/community/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/FeaturesAPIBuilder.java
```diff
@@ -68,7 +86,17 @@ public OpenAPI build(WFSInfo wfs) throws IOException {
 
         // list of valid filter-lang values
         Parameter filterLang = parameters.get("filter-lang");
-        filterLang.getSchema().setEnum(new ArrayList<>(APIFilterParser.SUPPORTED_ENCODINGS));
+        ArrayList<String> filterLangValues = new ArrayList<>(APIFilterParser.SUPPORTED_ENCODINGS);
```
aaime, line 89: Based on configuration, the sortby, filter, filter-lang parameters should also be removed (and so on)

src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/OGCAPIXStreamPersisterInitializer.java
```diff
@@ -20,6 +21,7 @@ public void init(XStreamPersister persister) {
         xs.alias("link", LinkInfo.class);
         xs.addDefaultImplementation(LinkInfoImpl.class, LinkInfo.class);
         persister.registerBreifMapComplexType("list", List.class);
+        persister.registerBreifMapComplexType("map", HashMap.class);
```
aaime, line 24: Is this still needed?

src/community/ogcapi/web-features/src/main/java/org/geoserver/ogcapi/web/FeatureServiceAdminPanel.java
```diff
@@ -0,0 +1,292 @@
...
+            String key,
+            final APIConformance conformance,
+            IModel<Boolean> booleanModel,
+            final IModel<Boolean> enabled) {
+        WFSInfo info = (WFSInfo) getDefaultModel().getObject();
+        boolean stable = conformance.getLevel().isStable();
+        boolean endorsed = conformance.getLevel().isEndorsed();
+
+        final Label label = new Label(key + "Label", conformance.getId());
+        //        {
```
aaime, line 126: Avoid commented out block

aaime (commented):
Had another go looking at the diff (but could not try it out interactively yet)

src/community/ogcapi/web-features/src/main/java/org/geoserver/ogcapi/web/FeatureServiceAdminPanel.html
jodygarnett: I am thinking how to do that, and we have an extension that provides:
- APIConformance being configured
- property expression `metadata/cql2/text` from WFSInfo

Then FeatureServiceAdminPanel could do the very simple story we have now. But not additional settings.

May be better to have an extension like ConformancePanel, with a default implementation for the simple case.

src/community/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/FeatureConformance.java
jodygarnett: Done

src/community/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/FeatureService.java
jodygarnett: resolved

src/community/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/FeaturesAPIBuilder.java
jodygarnett: path: search, queryables and functions now removable

parameters: filter, filter-lang, sortBy now removable

src/community/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/OGCAPIXStreamPersisterInitializer.java
jodygarnett: Removed

jodygarnett:
Rebased to get the cite fixes from main branch. Addressed formatting and QA failures also.

PR #8027, opened 2024-11-15
[GEOS-11605]: MapML Supporting custom TCRS/gridsets

dromagnoli:
[GEOS-11605](https://osgeo-org.atlassian.net/browse/GEOS-11605)
MapML Support custom TCRS projections from existing GridSets
  

# Checklist

### NOTE:
This PR depends on https://github.com/geotools/geotools/pull/4986.
Check Failures are expected until the corresponding GT PR get merged


doc/en/user/source/extensions/mapml/installation.rst
```diff
@@ -47,6 +47,33 @@ If the ``Represent multi-layer requests as multiple elements`` is checked (and t
...
+It provides a selector containing available GridSets. Administrator can select GridSets from the left list that will be converted to TiledCRSs.
+
+
+.. figure:: images/mapml_tcrs_selector.png
+
+
+Notes:
+
+- Gridsets containing ":" character in the name won't be listed
+- Gridsets with not-numeric levels or without a common prefix won't be listed
```
aaime, line 74:
```suggestion
- Gridsets with non-numeric levels or without a common prefix won't be listed
```

aaime: I'd also add an example of a gridset that would work (pick one of the built-in UTMs for example), e.g. UTM32WGS84Quad

src/extension/mapml/src/main/java/org/geoserver/mapml/MapMLHTMLOutput.java
```diff
@@ -127,6 +122,7 @@ public String toHTML() {
                 .append("mapml-viewer:not(:defined) > * { display: none; } n")
                 .append("map-layer { display: none; }\n")
                 .append("</style>\n")
+                .append(buildProjectionScript(projType))
```
aaime, line 125: Better to pass in the stringbuilder and allow appending stuff as needed inside, no?

src/extension/mapml/src/main/java/org/geoserver/mapml/MapMLURLBuilder.java
```diff
@@ -613,56 +616,27 @@ private String getSupportedWMTSGridSet(
             }
 
             // check same scales
-            for (int i = 0; i < tileMatrices.size(); i++) {
-                if (Math.abs(tileMatrices.get(i).getDenominator() - tiledCRSScales[i])
+            boolean sameScale = true;
+            for (int i = 0; i < tiledCRSResolutions.length; i++) {
+                if (Math.abs(tileMatrices.get(i).getResolution() - tiledCRSResolutions[i])
```
aaime, line 621: Everything was renamed to "resolution" (is it actually) but it's still compared to a SCALE_DELTA. So it is a pixel resolution, a scale denominator, a scale... what?

src/extension/mapml/src/main/java/org/geoserver/mapml/gwc/gridset/MapMLGridsets.java
```diff
@@ -281,4 +329,127 @@ public String getIdentifier() {
...
+     *     </ul>
+     *     Returns {@code null} if the input list is {@code null} or empty.
+     */
+    public static GridSetLevelType getLevelType(List<String> levels) {
+        if (levels == null || levels.isEmpty()) {
+            return null;
+        }
+        // First check: levels are simple numbers:
+        GridSetLevelType levelType = new GridSetLevelType();
+        String numericPattern = "-?\\d+(\\.\\d+)?";
```
aaime, line 412: Ouch, this way the pattern is re-compiled over and over. Create a Pattern as a static variable instead.

src/extension/mapml/src/main/java/org/geoserver/mapml/gwc/gridset/MapMLGridsets.java
```diff
@@ -281,4 +329,127 @@ public String getIdentifier() {
...
+                prefix = prefix.substring(0, prefix.length() - 1);
+                if (prefix.isEmpty()) {
+                    levelType.prefixed = false;
+                    return levelType; // No common prefix found
+                }
+            }
+        }
+
+        // Check if the remaining prefix is actually a valid common prefix (not just a number)
+        if (prefix.matches("\\d+")) {
```
aaime, line 448: Same as above, avoid dynamic pattern compile.

src/extension/mapml/src/main/java/org/geoserver/mapml/tcrs/TiledCRSParams.java
```diff
@@ -30,6 +34,13 @@ public TiledCRSParams(
         this.TILE_SIZE = tileSize;
         this.origin = origin;
         this.scales = scales;
+        // although they are redundant since one is the inverse of the other
+        // we store both of them for quick access (e.g. debug) without need
```
aaime, line 38: Looks like they are being used for more than just debug?

src/extension/mapml/src/main/java/org/geoserver/mapml/tcrs/WrappingProjType.java
```diff
@@ -0,0 +1,90 @@
...
+ */
+package org.geoserver.mapml.tcrs;
+
+import org.geoserver.mapml.xml.ProjType;
+import org.geotools.api.referencing.FactoryException;
+import org.geotools.api.referencing.crs.CoordinateReferenceSystem;
+import org.geotools.referencing.CRS;
+
+/** Wrap the Projection. In case of a Built-in Projection, it will contain the ProjType enum. */
+public class WrappingProjType {
```
aaime, line 13: I would call it based on what it is, rather than how it's implemented. MapMLProjection maybe?

aaime (changes requested):
Looks good overall, see comments below

src/extension/mapml/src/main/java/org/geoserver/mapml/tcrs/TiledCRSParams.java
dromagnoli: Correct. they were not needed during the first phases of the development. Removing that comment.

doc/en/user/source/extensions/mapml/installation.rst
dromagnoli: Done

dromagnoli:
Thx @aaime for the review. I have addressed all the feedbacks.

aaime (approved):

PR #8063, opened 2024-11-26
Workaround for 'could not acquire lock', see https://issues.apache.org/jira/browse/MNG-7868

aaime:
[MNG-7868](https://osgeo-org.atlassian.net/browse/MNG-7868) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

We have a fair amount of builds randomly failing with "could not acquire lock" in Maven itself.
Tracked it back to this, looks like it might be a networking issue, it's not solved but they propose workarounds to make it less likely:

https://issues.apache.org/jira/browse/MNG-7868

Those extra settings have been added to the maven options in this PR.

# Checklist

### DIFF
--- a/.github/workflows/assembly.yml
+++ b/.github/workflows/assembly.yml
@@ -8,7 +8,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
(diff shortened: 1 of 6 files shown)


groldan (approved):
Ff the issue is related to `mvn -T` one option would be to run a `go-offline` phase first and then build with `-o` but that won't work unfortunately, or at least it didn't every time I've tried in the past.  `mvn dependency:go-offline` won't actually pull everything it needs to build with `-o`. There is other plugin to "actually go offline" that I can't remember right now, but in any case I'd say lets try these maven params and see if we get less of these lock errors.

PR #8066, opened 2024-11-27
GEOS-11625: Add "Challenge Anonymous Sessions" Option to AuthKey Filter

afabiani:
[GEOS-11625](https://osgeo-org.atlassian.net/browse/GEOS-11625) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This pull request introduces the following enhancements to the GeoServer AuthKey extension to address stricter session handling and stateless authentication use cases:

1. Challenge Anonymous Sessions:

   - Adds a new "Challenge Anonymous Sessions" option to the AuthKey filter configuration.
   - When enabled, the filter will:
      - Ignore the current SecurityContext principal (if any), ensuring that the user linked to the authkey is always authenticated independently.
      - Force authorization even when a session is linked to an anonymous user, ensuring stateless behavior for requests using the authkey.
   - By default, this option is disabled to preserve the existing behavior.

2. Stateless Mode with Internal User Cache:

   - Introduces a configurable internal user cache for the AuthKey filter.
   - The cache optimizes performance in stateless mode by avoiding repeated backend service calls to retrieve the user associated with the authkey for every request.
   - Administrators can configure the cache expiration policy to balance performance and data freshness.

**Motivation:** 

These enhancements address scenarios where:

 - A specific path (e.g., WMS requests) is secured by the AuthKey filter, and stateless behavior is required to ensure stricter validation of the authkey query parameter.
 - Existing sessions, especially those linked to anonymous users, should not interfere with the security challenge enforced by the AuthKey filter.

**Key Changes:**

- Updates to the AuthKey filter logic to support the new "Challenge Anonymous Sessions" option.
- Addition of an internal user cache to reduce backend load and improve performance in stateless mode.
- Updates to filter configuration to expose these options via the GeoServer UI.

**Backward Compatibility:** 
- The "Challenge Anonymous Sessions" option is disabled by default, and the internal user cache is optional and configurable, ensuring no changes to existing setups unless explicitly enabled.

Please see the updated documentation for more details on configuration and usage.

# Checklist


groldan:
@afabiani please don't mind the ogcapi-features10 CITE failure. It should be good if you rebase on top of main.

doc/en/user/source/extensions/authkey/index.rst
```diff
@@ -237,6 +237,31 @@ After configuring the filter it is necessary to put this filter on the authentic
    saves the current configuration and triggers a synchronize. If users are added/removed from 
    the backing user/group service, the synchronize logic should be triggered.
 
+Challenge Anonymous Session
+---------------------------
+
+The AuthKey filter now includes a new option, "Challenge Anonymous Sessions," designed to enforce stricter authorization and stateless behavior.
```
aaime, line 243: This is not an announcement... in a few months there will be nothing new about this functionality any more. Best to just state what it does and use present tense

```suggestion
"Challenge Anonymous Sessions" is designed to enforce stricter authorization and stateless behavior.
```

doc/en/user/source/extensions/authkey/index.rst
```diff
@@ -237,6 +237,31 @@ After configuring the filter it is necessary to put this filter on the authentic
...
+
+When enabled, the filter will ignore any existing session or SecurityContext principal (e.g., anonymous or previously authenticated sessions created by other filters) and will always authenticate the user linked to the provided authkey. This ensures that the authkey is validated independently for every request, regardless of the current session state.
```
aaime, line 245:
```suggestion
When enabled, the filter ignores any existing session or SecurityContext principal (e.g., anonymous or previously authenticated sessions created by other filters) and always authenticates the user linked to the provided authkey. This ensures that the authkey is validated independently for every request, regardless of the current session state.
```

doc/en/user/source/extensions/authkey/index.rst
```diff
@@ -237,6 +237,31 @@ After configuring the filter it is necessary to put this filter on the authentic
...
+
+This feature is particularly useful for deployments that require stateless authentication or need to ensure that requests are processed strictly based on the authkey query parameter. By default, this option is disabled to preserve existing behavior.
+
+**Internal User Cache**
+
+To enhance performance in stateless mode, an internal user cache has been added to the AuthKey filter. The cache prevents unnecessary backend authentication requests for the same authkey by storing the user information locally for a configurable period.
```
aaime, line 251:
```suggestion
To enhance performance in stateless mode, the AuthKey filter can use an internal user cache. The cache prevents unnecessary backend authentication requests for the same authkey by storing the user information locally for a configurable period.
```

src/extension/authkey/src/main/java/org/geoserver/security/GeoServerAuthenticationKeyFilter.java
```diff
@@ -68,26 +71,27 @@ public void initializeFromConfig(SecurityNamedServiceConfig config) throws IOExc
...
-        String cacheKey = authenticateFromCache(this, (HttpServletRequest) request);
+        String cacheKey =
+                authenticateFromCache(
+                        this, (HttpServletRequest) request, allowChallengeAnonymousSessions);
 
-        if (SecurityContextHolder.getContext().getAuthentication() == null) {
+        Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+        if (allowChallengeAnonymousSessions
+                || authentication == null
+                || authentication instanceof AnonymousAuthenticationToken) {
```
aaime, line 88: It seems we get into this branch even if it's just an anonymous auth, even whenallowChallengeAnonymousSessions is false... was that intended?

src/extension/authkey/src/test/java/org/geoserver/security/AuthKeyAuthenticationTest.java
```diff
@@ -354,19 +365,19 @@ public void testFileBasedWithSession() throws Exception {
...
+
+            insertAnonymousFilter();
+            request = createRequest("foo/bar");
+            response = new MockHttpServletResponse();
+            chain = new MockFilterChain();
+            getProxy().doFilter(request, response, chain);
+            assertEquals(HttpServletResponse.SC_OK, response.getStatus());
+            // Anonymous context is not stored in http session, no further testing
+            removeAnonymousFilter();
+        }
```
aaime, line 380: Logic wise this is quite the change...  shouldn't you always get "forbidden"?
I saw something about windows in the commit list and wondering if it may be due to a late file system change notification or about granularity of the last modified date... I was bitten by it recently on windows. To avoid the issue, I've forced a reload from disk during the test, bypassing the "last modified" check.

src/extension/authkey/src/test/java/org/geoserver/security/AuthKeyAuthenticationTest.java
```diff
@@ -418,56 +436,58 @@ public void testUserPropertyWithCache() throws Exception {
...
-        insertAnonymousFilter();
-        request = createRequest("foo/bar");
-        response = new MockHttpServletResponse();
-        chain = new MockFilterChain();
-        getProxy().doFilter(request, response, chain);
-        assertEquals(HttpServletResponse.SC_OK, response.getStatus());
-        // Anonymous context is not stored in http session, no further testing
-        removeAnonymousFilter();
+        auth = auth != null ? auth : SecurityContextHolder.getContext().getAuthentication();
+        if (auth != null) {
```
aaime, line 444: Same as above

aaime (changes requested):
Overall looks good, I have some suggestions for documentation and a question about the filter itself

src/extension/authkey/src/main/java/org/geoserver/security/GeoServerAuthenticationKeyFilter.java
afabiani: Yes, it was intended. If the filter chain hits the default one, the latter injects an `AnonymousAuthenticationToken` into the `SecurityContext` and we must threat this one the same as it was not authenticated (`authentication == null`).

src/extension/authkey/src/test/java/org/geoserver/security/AuthKeyAuthenticationTest.java
afabiani: Done

aaime (approved):

jodygarnett:
@afabiani this change is getting some feedback from the user forum https://discourse.osgeo.org/t/authkey-extension-throws-nosuchmethoderror-2-26-1-after-dec-4-2024/112199/3

PR #8067, opened 2024-11-27
ogcapi-features CITE tests: build from official git repository

groldan:
Revert to build the `ogccite/ets-ogcapi-features10:1.8-SNAPSHOT-teamengine-5.4.1` docker image for ogcapi-features CITE testing from the official repository's `master` branch now that the [requried fix](https://github.com/opengeospatial/ets-ogcapi-features10/pull/255) is merged.

  

# Checklist

### DIFF
--- a/build/cite/ogcapi-features10/build-ets.sh
+++ b/build/cite/ogcapi-features10/build-ets.sh
@@ -1,13 +1,20 @@
 #!/bin/sh
 
 # build the ets-ogcapi-features10 Docker image from the
-# (at the time of writing, ogccite/ets-ogcapi-features10:1.8-SNAPSHOT-teamengine-5.4.1)
-# from the geoserver's fork `geoserver/integration` branch, which applies
+# https://github.com/opengeospatial/ets-ogcapi-features10 main repo's master branch,
+# since the required pull request has been merged (https://github.com/opengeospatial/ets-ogcapi-features10/pull/255)
+# but there's no new beta release so far.
+# This will build ogccite/ets-ogcapi-features10:1.8-SNAPSHOT-teamengine-5.4.1
+#
+# May other patches be needed in the future, we'll use the
+# geoserver's fork `geoserver/integration` branch again, which is intended to apply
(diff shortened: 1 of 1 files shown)


aaime:
Nice!

build/cite/ogcapi-features10/build-ets.sh
```diff
@@ -1,13 +1,20 @@
...
+# May other patches be needed in the future, we'll use the
+# geoserver's fork `geoserver/integration` branch again, which is intended to apply
 # patches to the ets that have not yet being released upstream
 
-echo "Building the ogccite/ets-ogcapi-features10 docker image from the geoserver/integration branch"
+echo "Building the ogccite/ets-ogcapi-features10 docker image from opengeospatial/ets-ogcapi-features10 master branch"
 rm -rf ets-ogcapi-features10
-git clone https://github.com/geoserver/ets-ogcapi-features10.git
+#git clone https://github.com/geoserver/ets-ogcapi-features10.git
+git clone https://github.com/opengeospatial/ets-ogcapi-features10.git
```
aaime, line 16: In general, we'll eventually want to use a fixed tag.  With pros and cons:

* Pros: the random PR author will not get caught in test engine changes that have nothing to do with their own GeoServer changes
* Cons: we won't be notified of the eventual failures, which might be due to regressions in the test engine or new stricter/more comprehensive tests being implemented

In an ideal world, I guess we'd want both. Maybe a fixed tag in the PRs, and the "latest and greatest" on the jenkins build server?

groldan: sounds good. My preference would be to run it also on github actions. We can see when and why. Like as a daily cron job, or on new tags/releases, pushes to main, not sure.

aaime: The github action works fine too, sure, the important bit is not causing confusion for PR submitters. Daily seems like a good idea. How would we be notified in case of failure?

PR #8053, opened 2024-11-21
[GEOS-11623] Refactor GSModalWindow to use ModalDialog instead of deprecated ModalWindow

sikeoka:
[GEOS-11623](https://osgeo-org.atlassian.net/browse/GEOS-11623) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This PR refactors GSModalWindow to use ModalDialog instead of ModalWindow which is deprecated in Wicket 9 and removed from Wicket 10 which is required for the Jakarta EE migration. The new modals should be visually very similar to the old modals. Two capabilities of the old modals that have not been ported into the new modals are dragging modals around the screen and resizing modals. They can be future enhancements if anyone really needs that.

The PageCreator interface was removed and everything now sets a Component as the modal content instead of a Page. The only case where this caused issues besides styling issues was with the WPS subprocess builder which required a workaround to get the select2 dropdown working properly. Unused ModalWindow API was removed and resizing-related API was also removed since that isn't currently supported.

This PR includes the changes in https://github.com/geoserver/geoserver/pull/7988 and replaces that PR. If this PR were merged along with https://github.com/geoserver/geoserver/pull/7985 and https://github.com/geoserver/geoserver/pull/7986, the only three remaining CSP violations are inline script usage in the importer extension and elasticsearch and taskmanager community modules. The core GeoServer webapp should be fully functional with the blocking CSP enabled.

  

# Checklist

- [x] I have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/main/CONTRIBUTING.md).
- [x] I have sent a [Contribution Licence Agreement](https://docs.geoserver.org/latest/en/developer/policies/committing.html) (not required for small changes, e.g., fixing typos in documentation).
- [x] First PR targets the `main` branch (backports managed later; ignore for branch specific issues).
- [x] All the build checks are green ([see automated QA checks](https://docs.geoserver.org/latest/en/developer/qa-guide/index.html)).

For core and extension modules:

- [x] New unit tests have been added covering the changes.
- [ ] [Documentation](https://github.com/geoserver/geoserver/tree/main/doc/en/user/source) has been updated (if change is visible to end users).
- [ ] The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers).
[...]

PR #8079, opened 2024-12-02
Fixed some Wicket CSP violations

sikeoka:
This PR should fix all of the remaining CSP violations from the Wicket 9 upgrade after https://github.com/geoserver/geoserver/pull/8053 is merged. I do not use any of the extensions or community modules so those changes are completely untested by me but I'm putting it here if anyone is interested.

  

# Checklist

### DIFF
--- a/src/community/elasticsearch/src/main/java/org/geoserver/elasticsearch/ElasticConfigurationPage.java
+++ b/src/community/elasticsearch/src/main/java/org/geoserver/elasticsearch/ElasticConfigurationPage.java
@@ -18,7 +18,9 @@ import org.apache.wicket.Component;
 import org.apache.wicket.ajax.AjaxRequestTarget;
 import org.apache.wicket.ajax.markup.html.form.AjaxButton;
 import org.apache.wicket.ajax.markup.html.form.AjaxCheckBox;
-import org.apache.wicket.behavior.AttributeAppender;
+import org.apache.wicket.behavior.Behavior;
+import org.apache.wicket.markup.head.IHeaderResponse;
+import org.apache.wicket.markup.head.OnLoadHeaderItem;
 import org.apache.wicket.markup.html.basic.Label;
 import org.apache.wicket.markup.html.form.CheckBox;
 import org.apache.wicket.markup.html.form.DropDownChoice;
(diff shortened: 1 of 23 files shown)


mprins (approved):
lgtm (untested)

src/extension/importer/web/src/main/java/org/geoserver/importer/web/ImportTaskTable.java
```diff
@@ -449,11 +451,21 @@ public LayerPreviewPanel(String id, IModel<ImportTask> model) {
...
-            String javascript =
-                    "go(document.getElementById('" + get("links").getMarkupId() + "'));";
-            add(
-                    new ExternalLink("go", "#")
-                            .add(new AttributeModifier("onclick", new Model<>(javascript))));
+            add(new ExternalLink("go", "#").setOutputMarkupId(true));
+        }
+
+        @Override
+        public void renderHead(IHeaderResponse response) {
```
aaime, line 458: This one is not working. When importing multiple layers, the "go" links all make the first layer preview appear. E.g., I've imported the tasmania shapefiles in another workspace, after a successful import, all "go" link preview the cities layer:

image

src/community/elasticsearch/src/main/java/org/geoserver/elasticsearch/ElasticConfigurationPage.java
```diff
@@ -293,12 +295,25 @@ protected Component getComponentForProperty(
                         } else if (property == ElasticAttributeProvider.USE) {
                             CheckBox checkBox =
                                     new CheckBox("use", new PropertyModel<>(itemModel, "use"));
-                            final String onclick =
-                                    "document.getElementById(\""
+                            checkBox.setOutputMarkupId(true);
```
aaime, line 298: Manually tested, it works.

aaime (changes requested):

src/extension/importer/web/src/main/java/org/geoserver/importer/web/ImportTaskTable.java
sikeoka: I looked at the code and I'm wondering if this behavior you're describing is actually the current behavior without this PR?

sikeoka: There's also still more inline script on this page that I need to remove.

aaime: > I looked at the code and I'm wondering if this behavior you're describing is actually the current behavior without this PR?

That may well be true, I haven't used the importer GUI in a long while (and did not test it before this change).

sikeoka:
It turns out that that original commit missed a lot of inline styles that used style tags in the HTML page. I updated this PR to fix them all too. The CAS and OAUTH2 panels will inherit the correct style from `org.geoserver.security.web.auth.PreAuthenticatedUserNameFilterPanel`.
An inline script issue in GeoServerLoginPage was fixed. GeoServerBasePage already adds the `placeholder()` call so it was being called twice on the login page.

src/extension/importer/web/src/main/java/org/geoserver/importer/web/ImportTaskTable.java
aaime: Tested again and found it working... duh, not sure what happened here.

PR #8101, opened 2024-12-09
Upgrade wiremock from 2.35.1 to 3.10.0

sikeoka:
Upgrade the wiremock dependency to the current version. GeoServer uses the standalone JAR which includes all of its dependencies under a separate package so it isn't impacted by wiremock's Tomcat 11 and Jakarta EE dependencies.

The assembly workflow was updated to ensure that community modules are built using the root GeoServer POM from the local PR branch and not what is published in the public Maven repository which would cause invalid POM errors when a PR adds or changes a dependency's groupId/artifactId in the root POM.

One of the WPS test classes was updated to properly use the class rule which improves performance when the test classes contain a lot of tests. Some tests that programmatically handle the wiremock server weren't shutting down the server when the tests were finished and some other tests were changed to call stop() instead of shutdown() which is the documented way to stop the server.

  

# Checklist

### DIFF
--- a/.github/workflows/assembly.yml
+++ b/.github/workflows/assembly.yml
@@ -46,7 +46,7 @@ jobs:
         mvn -B -ntp -nsu -N -f src/pom.xml assembly:single
     - name: Build and package community modules (without tests)
       run: |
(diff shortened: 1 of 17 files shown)


mprins (approved):
lgtm

PR #8102, opened 2024-12-09
JWT Headers - better handling of JWK Set URL reading

davidblasby:
For some Geonetwork 5 development, it would be nice to load JWK Sets (i.e. public keys) from `"file://..."` URLs.  The underlying library doesn't allow this.

This PR uses the URL#open() stream instead of having the library read the URL.

# Checklist

### DIFF
--- a/src/community/jwt-headers/jwt-headers-util/src/main/java/org/geoserver/security/jwtheaders/token/TokenSignatureValidator.java
+++ b/src/community/jwt-headers/jwt-headers-util/src/main/java/org/geoserver/security/jwtheaders/token/TokenSignatureValidator.java
@@ -50,7 +50,9 @@ public class TokenSignatureValidator {
 
     public static JWKSet loadJWKSet(String urlStr) throws Exception {
         URL url = new URL(urlStr);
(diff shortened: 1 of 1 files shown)


davidblasby:
This is a trivial change to a community module - will merge tomorrow (tuesday)

src/community/jwt-headers/jwt-headers-util/src/main/java/org/geoserver/security/jwtheaders/token/TokenSignatureValidator.java
```diff
@@ -50,7 +50,9 @@ public JWKSet load(String urlStr) throws Exception {
 
     public static JWKSet loadJWKSet(String urlStr) throws Exception {
         URL url = new URL(urlStr);
-        return JWKSet.load(url);
+        try (var stream = url.openStream()) {
```
jodygarnett, line 53: Huh, it does not use the http client event.

@davidblasby (or @david-blasby) this is your community module so you do not need to wait on PRs while under development.

jodygarnett (approved):

PR #8120, opened 2024-12-17
[GEOS-11647] Restore "quiet on not found" configuration for REST in global settings

groldan:
[GEOS-11647](https://osgeo-org.atlassian.net/browse/GEOS-11647) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

In [this commit](e57616971c4c11f6efdd7d12b855b509fbbe9ae1) from 2014, each resource lookup in the REST API would check for the configuration setting, and avoid throwing `RestletException` (which we use `ResourceNotFoundException` for those cases).

For example:

```
boolean quietOnNotFound=quietOnNotFoundEnabled(request)
if(quietOnNotFound){
    return null;
}
throw new RestletException( "No such layer: " + layer, Status.CLIENT_ERROR_NOT_FOUND );
```

Now the checkbox to enable it persists in `org.geoserver.web.RESTSettingsPanel` but it has no effect, and the only way to silence the log flooding is by adding a `?quietOnNotFound=true` parameter to every REST API request.

Make `org.geserver.rest.RestControllerAdvice` account for either the request parameter or the global settings configuration when deciding if it'll log the stack trace or not.

  

# Checklist

### DIFF
--- a/src/rest/pom.xml
+++ b/src/rest/pom.xml
@@ -61,4 +61,16 @@
     </dependency>
   </dependencies>
 
+  <build>
+    <plugins>
+      <plugin>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <!-- required for logging related tests in ExceptionHandlingTest not to interfere with each other -->
+          <forkCount>1</forkCount>
+          <reuseForks>false</reuseForks>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
 </project>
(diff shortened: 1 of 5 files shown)


groldan:
Hi @aaime, I'm trying but couldn't realize why the log handler wouldn't be called, then I tried running `SecuredFeatureSourceTest` and noticed it's not being called in there either, it's just that the test only checks no events have been captured:

image

I'll dig into it a little bit more

groldan:
yeah it's because the logger is a `Log4j2Logger`

aaime:
If I recall properly, @jodygarnett did develop support for logging tests when switching to log4j2... @jodygarnett do you remember where that support is?

groldan:
@aaime managed to create a strategy object to handle logs capturing, check out 604f77a91e76f610fda52d28131068e1ab4df16c

jodygarnett:
Here you go: https://github.com/geoserver/geoserver/blob/main/src/main/src/test/java/org/geoserver/logging/TestAppender.java

PR #8137, opened 2024-12-20
Add proper JSON array representation to GeoServer

Arxa:
# Checklist

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
@@ -17,6 +17,7 @@ import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import net.sf.json.JSONException;
+import net.sf.json.JSONObject;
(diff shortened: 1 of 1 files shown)


ianturton:
That's a real fail for the QA build, and it could really use a test.

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
```diff
@@ -461,7 +462,12 @@ protected FeaturesInfo encodeSimpleFeatures(
                                     || value instanceof Float
                                             && ((Float) value) == Float.NEGATIVE_INFINITY) {
                                 jsonWriter.value("-Infinity");
-                            } else {
+                            }
+                            else if (value instanceof String && isJsonString((String) value)) {
```
aaime, line 466: Oh no, that's going to slow down general JSON output generation too much (for all normal strings).
Instead, look into the attribute descriptor and look for the native type name, if available:

https://github.com/geotools/geotools/blob/e58d016c76e375653267c63eddbd17b5cbbf1a28/modules/library/jdbc/src/main/java/org/geotools/jdbc/JDBCFeatureSource.java#L289

PR #8167, opened 2024-12-26
[GEOS-11346] Enable strict Content-Security-Policy by default

sikeoka:
Using the `Content-Security-Policy-Report-Only` header was necessary to support the Wicket 9 upgrade but the `Content-Security-Policy` header which blocks CSP violations can be used now that most of the remaining CSP violations, particularly the modal window, have been resolved. This will provide a more secure default mode, make any remaining CSP violations (if there are any left) more obvious and ensure that developers do not add new CSP violations.

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/csp/CSPConfiguration.java
+++ b/src/main/src/main/java/org/geoserver/security/csp/CSPConfiguration.java
@@ -20,8 +20,7 @@ public class CSPConfiguration implements Serializable {
 
     // default values
     private static final Boolean DEFAULT_ENABLED = true;
-    // TODO: can switch to false after all Wicket CSP issues are fixed
-    private static final Boolean DEFAULT_REPORT_ONLY = true;
+    private static final Boolean DEFAULT_REPORT_ONLY = false;
     private static final Boolean DEFAULT_ALLOW_OVERRIDE = false;
     private static final Boolean DEFAULT_INJECT_PROXY_BASE = false;
     private static final String DEFAULT_REMOTE_RESOURCES = "";
--- a/src/web/core/src/main/java/org/geoserver/web/GeoServerApplication.java
+++ b/src/web/core/src/main/java/org/geoserver/web/GeoServerApplication.java
@@ -102,7 +102,7 @@ public class GeoServerApplication extends WebApplication
      *
      * <p>Sample use: {@code org.geoserver.web.csp.strict=true}
      */
-    public static boolean CSP_STRICT = Boolean.valueOf(System.getProperty("org.geoserver.web.csp.strict", "false"));
+    public static boolean CSP_STRICT = Boolean.valueOf(System.getProperty("org.geoserver.web.csp.strict", "true"));
 
     public static final String GEOSERVER_CSRF_DISABLED = "GEOSERVER_CSRF_DISABLED";
     public static final String GEOSERVER_CSRF_WHITELIST = "GEOSERVER_CSRF_WHITELIST";


sikeoka:
@jodygarnett do you have thoughts about this?

jodygarnett:
I would like to see this enabled on main branch, and we can tell folks about the setting in the release notes.

aaime:
Merging: let's put this under more eyeys and see how this goes, we still have some time to course-correct before the 2.27.0 release.

jodygarnett:
yep, just do not want us to forget the update notes

aaime:
Some update notes regarding CSP were already in there before this PR got merged, due to the services related CSP work:
https://docs.geoserver.org/latest/en/user/installation/upgrade.html#content-security-policy-geoserver-2-27-and-newer

I don't see org.geoserver.web.csp.strict mentioned anywhere, could be useful for those that have customized GeoServer with extra panels (for those, a mention of the Wicket 9 upgrade may also be useful).

aaime:
Something seems to have gone really wrong with this PR being merged... on [gs-main](https://gs-main.geosolutionsgroup.com/geoserver/web/), a GeoSolutions demo server updated 3 times a day with the latest build from main, I can see all sorts of misbehavior:

* Loading the home page returns a number of CSP errors

image

* Logging in does not seem to work, no messages, I'm just staying in the same page as if the login did not occur.  But if I click on the GeoServer log, then I find myself logged in (the redirect seems to fail for some reason)

* Trying to select anything in table pages (layers, stores, workspaces) does nothing

* Switching tabs does not seem to work in any page (layers, workspaces, server status) with errors like "Refused to send form data to 'https://gs-main.geosolutionsgroup.com/geoserver/web/wicket/bookmarkable/org.geoserver.web.data.layergroup.LayerGroupEditPage?14-1.-publishedinfo&group=dark_labels&workspace=eumetsat' because it violates the following Content Security Policy directive: "form-action 'self'".

* Going to authentication and then clicking on any filter chain for edit works, but once done, the "close" button refuses to work, again with a "Refused to send form data to 'https://gs-main.geosolutionsgroup.com/geoserver/web/wicket/page?22-1.-form' because it violates the following Content Security Policy directive: "form-action 'self'"."

Now... some of this seems to be related to caching. I can force-reload the home page (CTRL-F5) and the CSP failures are gone, same goes for the checkboxes in the table pages, they start working. But I cannot get the tabs and the "close" buttons to work by just reloading, or even using an incognito tab.

This experience will hit anyone upgrading, if it's cache-related, we need some way to break the existing caches, but it seems there's more at play.

@sikeoka do you think it's something that can be fixed very quickly? Otherwise I'm suggesting to revert this PR and merge again when the obvious issues

sikeoka:
I am starting to look at this CSP stuff now. Figuring out the REST API unit test took way too long and it ended up being a bug with the base GeoServer test class setting incorrect paths in a specific case that I had to figure out how to fix without breaking anything.

I don't know if it's a proxy configuration issue or something in the GeoServer code but the "form-action 'self'" CSP violations are actually because the server is redirecting the HTTPS request to HTTP so they are considered to be different origins (https://gs-main.geosolutionsgroup.com to http://gs-main.geosolutionsgroup.com). The HTTP URL returns a 404 Not Found response (this is not relevant to the CSP violation).

```
HTTP/1.1 302
Server: nginx
Date: Tue, 28 Jan 2025 16:01:36 GMT
Content-Length: 0
Connection: keep-alive
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
Content-Security-Policy: base-uri 'self'; form-action 'self'; default-src 'none'; child-src 'self'; connect-src 'self'; font-src 'self'; img-src 'self' data:; style-src 'self' 'unsafe-inline'; script-src 'self';, frame-ancestors 'self';
Set-Cookie: remember-me=; Max-Age=0; Expires=Thu, 01-Jan-1970 00:00:10 GMT; Path=/geoserver
Location: http://gs-main.geosolutionsgroup.com/geoserver/web/wicket/bookmarkable/org.geoserver.web.GeoServerLoginPage?error=true
Access-Control-Allow-Origin: https://gs-main.geosolutionsgroup.com
Access-Control-Allow-Credentials: true
Access-Control-Allow-Methods: GET, POST, OPTIONS, PUT, DELETE
Access-Control-Allow-Headers: DNT,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range
Access-Control-Expose-Headers: Content-Length,Content-Range
```

groldan:
Here's a `compose.yml` and `nginx.conf` file to try to reproduce:

```yaml
services:
  geoserver:
    image: docker.osgeo.org/geoserver:2.26.1
    container_name: geoserver
    ports:
      - "8090:8080" # GeoServer will be accessible internally at port 8080
    environment:
      - GEOSERVER_LOG_LOCATION=/opt/geoserver/data_dir/logs/geoserver.log
      - CORS_ENABLED=true
      - CORS_ALLOWED_ORIGINS=*
      - INSTALL_EXTENSIONS=true
      - STABLE_EXTENSIONS=sldservice

  nginx:
    image: nginx:latest
    container_name: nginx
    depends_on:
      - geoserver
    ports:
      - "8080:8080"
    volumes:
      - ./nginx.conf:/etc/nginx/nginx.conf
```

`nginx.conf`:

```
worker_processes 1;

events { worker_connections 1024; }

http {
    upstream geoserver {
        server geoserver:8080;
    }

    server {
        listen 8080;

        location / {
            proxy_pass http://geoserver;
            proxy_set_header Host $host:$server_port;
            proxy_set_header X-Real-IP $remote_addr;
            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
            proxy_set_header X-Forwarded-Host $host:$server_port;
            proxy_set_header X-Forwarded-Proto $scheme;
            proxy_set_header X-Forwarded-Port $server_port;
        }
    }
}
```

sikeoka:
@aaime , I feel like the CSP violations in your first bullet are actually a weird false positive from the browser. When I clear my browser cache and reproduce the issue, I can see that everything appears to be loading correctly even though those CSP violations still appear in the logs. From what I am seeing, the CSP violations will appear for each stylesheet link on any GeoServer Wicket page only if the page is loaded while the developer tools is closed and they won't appear if you load the page while the developer tools is open.

Your remaining issues may all be because of a bad server configuration that is causing HTTPS URLs to have an HTTP URL in the Location response header. I don't use nginx so I don't know what might be going wrong but I would recommend fixing that and then re-testing.

jodygarnett:
> yep, just do not want us to forget the update notes

Moving to https://osgeo-org.atlassian.net/browse/GEOS-11698

petersmythe:
Does anyone know if GeoServer respects the `X-Forwarded-Proto https` HTTP header set by the proxy (nginx) which should tell GeoServer that the original request was HTTPS?  Or does it always use the protocol included in system properties `-DPROXY_BASE_URL=https://cite.geoserver.org/geoserver` (note, not set in GUI > Settings > Global > Proxy Base URL - haven't tried that yet) because that seems to be `http` always.

sikeoka:
Also, I created a simple test case using wiremock to prove that this is a false positive and that nothing is actually being blocked by the CSP.

jodygarnett:
There must be a way to upgrade, if we change the names of URLs alongside GS3, or add a CLA cache buster, or...

sikeoka:
[csp-mock.zip](https://github.com/user-attachments/files/18591224/csp-mock.zip)
I attached a wiremock example to prove that this is false positive from the browser. Download and unzip the file, run `java -jar wiremock-standalone-3.10.0.jar`, open Edge or Chrome and navigate to `http://localhost:8080/test.html` and then open the Developer Tools and you will see the CSP violation even though nothing was actually blocked. If you load the page with the Developer Tools already open, then you won't see any CSP violation. I can't test if other browsers have a similar problem

jodygarnett:
@sikeoka can you think of anything we can do to mitigate this issue?

> From what I am seeing, the CSP violations will appear for each stylesheet link on any GeoServer Wicket page only if the page is loaded while the developer tools is closed and they won't appear if you load the page while the developer tools is open.

This could be addressed by renaming the stylesheets alongside GS3 theme change?

>  bad server configuration that is causing HTTPS URLs to have an HTTP URL in the Location response header.

Can we detect this somehow and offer administrator a warning?

sikeoka:
The stylesheet CSP violations are false positives because of a bug in the browser and have no impact on the rendered page. Just ignore them. Users won't even notice them unless the open the developer tools.

Redirecting from HTTPS to HTTP is a bad thing to do even before the CSP issue and that would be unacceptable behavior for any environment that cares about security. This would likely be detected by security scanning software against the server and I don't think GeoServer should attempt to do anything about this insecure case.

jodygarnett:
@sikeoka the http / https thing I do not understand it, and Andrea indicated he could not reproduce it?

aaime:
I've been digging a bit into this, **if** what I've found is troublesome.

So all CSP errors I get on the console talk about https, no trace of HTTP, however, checking the actual HTTP response headers, I see an HTTP (why the console does not report it, it's beyond my understanding).

Now, on login, there is code redirecting to "/web" as a relative location, literally response.sendRedirect("/web"). The web container (both Tomcat and Jetty) does not trust relative redirects, and ends up building an absolute path instead.

nginx is correctly setting the "x-forwarded-proto: https" header, but Tomcat is ignoring it by default, as there is no way in general to know if that was sent from a trustworthy proxy, or directly from the client (as a means to mount some sort of attack). In order for Tomcat to trust the header, a configuration must be added in its contex.xml:

```
<Context>
    <Valve className="org.apache.catalina.valves.RemoteIpValve"
           protocolHeader="X-Forwarded-Proto"
           protocolHeaderHttpsValue="https"/>
</Context>
```

Now, in my experience almost all deploys have a proxy in front handling HTTPs, very few ones configure HTTPS directly in Tomcat. So now all deploys will have to go and setup that valve? Put simply, that's not going to fly.

The redirect target is set in the authentication filters, here is one example:
https://github.com/geoserver/geoserver/blob/99ae245402fca51e306eaf9817faaa891bf89e73/src/main/src/main/java/org/geoserver/security/filter/GeoServerUserNamePasswordAuthenticationFilter.java#L96

We could build a full URL there using ResponseUtils.buildURL, which would account for the configured proxy base URL... however that code is called on startup, when there is no HTTP request in context, which may break the ProxifyingURLMangler (it can use some of the headers). 
I guess it's probably easier/better to write a new AbstractAuthenticationTargetUrlRequestHandler subclass that can compute an absolute redirect URL on the fly?

That said... I'm not 100% sure of that I'm seeing, I'd have to setup some local proxying with nginx and HTTPs handling to be sure (not something I've ever done before).

sikeoka:
> Now, on login, there is code redirecting to "/web" as a relative location, literally response.sendRedirect("/web"). The web container (both Tomcat and Jetty) does not trust relative redirects, and ends up building an absolute URL instead, inclusive of protocol.

That is actually the behavior required by the Servlet API (https://javaee.github.io/javaee-spec/javadocs/javax/servlet/http/HttpServletResponse.html#sendRedirect-java.lang.String-):

Sends a temporary redirect response to the client using the specified redirect location URL and clears the buffer. The buffer will be replaced with the data set by this method. Calling this method sets the status code to [SC_FOUND](https://docs.oracle.com/javaee/6/api/javax/servlet/http/HttpServletResponse.html#SC_FOUND) 302 (Found). **This method can accept relative URLs;the servlet container must convert the relative URL to an absolute URL before sending the response to the client.** If the location is relative without a leading '/' the container interprets it as relative to the current request URI. If the location is relative with a leading '/' the container interprets it as relative to the servlet container root.

aaime:
Good to know, but also configurable otherwise, see [useRelativeRedirect](https://tomcat.apache.org/tomcat-9.0-doc/config/context.html) in the Tomcat configuration (debugging I've noticed a setting also in Jetty, code wise at least).

But that's beyond the point. GeoServer popularity is due, among other things, by its relatively good success at hiding "that Java thing" which many users/admin are not familiar with (more often than not, GeoServer is the only Java deploy they have). Docker installations helps going in that direction too, people do not need to setup java and configure Tomcat.

Asking people to tweak their existing setup in a different way based on the web container they are using is a nonstarter, especially in a situation where the UI is just not working, with errors only visible through the developer tools.

sikeoka:
It looks like someone created an issue about the stylesheet nonce error messages a few years ago but it doesn't look like it's ever going to be fixed in the browser so you'll just have to learn to ignore them: https://issues.chromium.org/issues/40858453

There is a more recent issue about providing clearer error messages when the CSP blocks something because of a redirect response but that doesn't seem to be getting much activity either: https://issues.chromium.org/issues/384523574

PR #8169, opened 2024-12-30
Update CITE tests to the production images where possible

aaime:
This PR updates a number of CITE tests and switches them to the production docker image of the teamengine, starting them up with the REST API. Exceptions:

* **WCS 1.0** has no way to be started with REST, [I would not hold my breath waiting for it](https://github.com/opengeospatial/ets-wcs10/issues/43) and the process is not immediate, here is a [WCS 1.1 example for reference](https://github.com/opengeospatial/ets-wcs11/pull/54)  (and they did not have to split the CTL files into per-request sub-files). I also could not get to pass the test suite due to some oddball issue invoking a XSTL function... however, running the WCS 1.0 CITE tests from the web interface works (so it could be used for certification, even if we cannot keep tabs on it).
* **WFS 1.1** fail to work when invoked through the REST API, apparently [due to a Teamengine packaging issue](https://github.com/opengeospatial/ets-wfs11/issues/109). When that is fixed, we can switch this one to latest as well.

In order to get things passing with all the suites, I had to centralized the version negotiation, and pull some hacks to identify the version of ServiceException the WCS test demand (they don't provide the version number, and assume one still responds with the ServiceException from the WCS being tested).

# Checklist


groldan (approved):
Amazing

PR #8173, opened 2025-01-02
Running WFS 2.0 CITE tests

aaime:
Sets up and runs the WFS 2.0 CITE tests using the existing data directory (moved onto build/cite, like the others).
The dataset being used is the WFS 1.1 one. I have see one locking failures during tests, but could not reproduce it afterwards... let's keep an eye on it.

# Checklist


build/cite/wfs20/compose.override.yml
```diff
@@ -0,0 +1,32 @@
+services:
+  geoserver:
+    image: ogccite/geoserver:wfs11
```
mprins, line 3: should this not be `ogccite/geoserver:wfs20`?

looking at https://github.com/geoserver/geoserver/blob/352c0a8d7121107a4e1d5cab18b08c4169bacae6/.github/workflows/cite.yml#L58 
and 
https://github.com/geoserver/geoserver/blob/352c0a8d7121107a4e1d5cab18b08c4169bacae6/.github/workflows/cite.yml#L83C43-L85
that image seems to be build as part of the workflow

aaime:
Merging to get one more test in. No GS code changes were needed for this one, just CITE machinery stuff.

PR #8177, opened 2025-01-07
[GEOS-11669] Patch jscolor to work with Wicket's CSP

sikeoka:
[GEOS-11669](https://osgeo-org.atlassian.net/browse/GEOS-11669) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This PR patches the jscolor library to parse the JSON using JSON.parse instead of eval which is blocked by Wicket's CSP and updates the ColorPickerField class to be valid JSON. I looked into upgrading jscolor instead but it has other CSP violations that would need to be patched anyways. The indentation only looks weird because the jscolor file uses tabs but the code I added uses spaces.
  

# Checklist

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/wicket/ColorPickerField.java
+++ b/src/web/core/src/main/java/org/geoserver/web/wicket/ColorPickerField.java
@@ -5,14 +5,11 @@
  */
 package org.geoserver.web.wicket;
 
-import org.apache.wicket.Component;
-import org.apache.wicket.behavior.AttributeAppender;
-import org.apache.wicket.behavior.Behavior;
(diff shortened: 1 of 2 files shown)


mprins (approved):
LGTM, just wondering if we could do away with the jscolor.js and use a native picker with `<input type="color" .../>` instead

aaime:
@mprins good point, all major browsers seem to support the color input type:
https://caniuse.com/input-color

Even if someone is using an outdated browser, the effect would be minimal, one can still just type hex colors.

That said, merging this one, and looking forward to an eventual PR switching to use the color input type.

PR #8184, opened 2025-01-09
Make tutorial easier to understand by using full path to plugin folder in vectortiles tutorial.

Severino:
I think you should simply provide the actual path to the folder. For me installing a plugin first time I needed to do an 'extensive search' to get this simple information.

# Checklist

### DIFF
--- a/doc/en/user/source/extensions/vectortiles/install.rst
+++ b/doc/en/user/source/extensions/vectortiles/install.rst
@@ -7,7 +7,11 @@ Installing the Vector Tiles Extension
 
    .. warning:: Make sure to match the version of the extension to the version of GeoServer.
 
-#. Extract the archive and copy the contents into the GeoServer :file:`webapps/geoserver/WEB-INF/lib` directory.
+#. Extract the archive and copy the contents into the GeoServer library :file:`WEB-INF/lib` directory located in:
+   
+  * GeoServer binary Jetty: :file:`<GEOSERVER_ROOT>/webapps/geoserver/WEB-INF/lib`
+  * Default Tomcat deployment: :file:`<CATALINA_BASE>/webapps/geoserver/WEB-INF/lib`
+
 
 #. Restart GeoServer.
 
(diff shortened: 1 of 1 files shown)


doc/en/user/source/extensions/vectortiles/install.rst
```diff
@@ -7,7 +7,7 @@ Installing the Vector Tiles Extension
 
    .. warning:: Make sure to match the version of the extension to the version of GeoServer.
 
-#. Extract the archive and copy the contents into the GeoServer :file:`WEB-INF/lib` directory.
+#. Extract the archive and copy the contents into the GeoServer :file:`webapps/geoserver/WEB-INF/lib` directory.
```
aaime, line 10: The "webapps" directory is commonly found in Tomcat/Jetty, and in our "bin" package, but it's not guaranteed to be found in other web containers, e.g., Wildfly and WebSphere do not have it.

Rewording the instruction might be useful, but it should be done by adding an extra sentence, something like:

> For example, if you are using the GeoServer bin package, or the Windows installer, look for `webapps/geoserver/WEB-INF/lib`

Severino: Afterwards I found that the information is contained inside the README.html of the plugin archieve. If this is consistent for all platforms, it could just reference that?

aaime: The READE has correctly contextualized information:

> Unzip this archive into the GeoServer library directory. In the binary
   install this is under ``<GEOSERVER_ROOT>/webapps/geoserver/WEB-INF/lib``.

The first part is generic (even more so than the existing doc for vector tiles), the second one refers specifically to the binary (bin) package, where we know where GeoServer ends up being.  It's so small that I'd rather copy it instead of having people look first at the docs and then also the zip.

jodygarnett, line 10:
```suggestion
#. Extract the archive and copy the contents into the GeoServer library :file:`WEB-INF/lib` directory located in:
   
  * GeoServer binary Jetty: :file:`<GEOSERVER_ROOT>/webapps/geoserver/WEB-INF/lib`
  * Default Tomcat deployment: :file:`<CATALINA_BASE>/webapps/geoserver/WEB-INF/lib`

```

jodygarnett:
Folks rather than have this discussion open I have revised the text based on feedback. I note that the text is kind of boilerplate and reused for the install instructions for several different extensions. Perhaps in a future version of the docs we can address this consistently.

PR #8219, opened 2025-01-23
[GEOS-11672] WMTS layer service enablement

jodygarnett:
[GEOS-11672](https://osgeo-org.atlassian.net/browse/GEOS-11672) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

Allows WMTS to be selectable as a serivce to disable; and will only list WMTS servivces if TileLayer is defined for the layer or LayerGroup.

Noticed reuse the GWC "voter" to avoid duplicating logic (such as checking for a tile layer).

# Checklist

### DIFF
--- a/src/web/gwc/src/main/java/org/geoserver/gwc/web/GWCServiceDescriptionProvider.java
+++ b/src/web/gwc/src/main/java/org/geoserver/gwc/web/GWCServiceDescriptionProvider.java
@@ -7,12 +7,14 @@ package org.geoserver.gwc.web;
 import java.util.ArrayList;
 import java.util.List;
 import org.geoserver.catalog.Catalog;
+import org.geoserver.catalog.LayerInfo;
 import org.geoserver.catalog.PublishedInfo;
 import org.geoserver.catalog.WorkspaceInfo;
 import org.geoserver.config.GeoServer;
 import org.geoserver.config.ServiceInfo;
 import org.geoserver.gwc.GWC;
 import org.geoserver.gwc.config.GWCConfig;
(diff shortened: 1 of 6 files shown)


src/web/gwc/src/main/java/org/geoserver/gwc/web/GWCServiceDescriptionProvider.java
```diff
@@ -57,11 +59,17 @@ protected WMTSInfo info(WorkspaceInfo workspaceInfo, PublishedInfo layerInfo) {
 
     /** GWC-bases services don't have layer-specific enabling... */
     @Override
-    protected boolean isAvailable(String serviceType, ServiceInfo info, PublishedInfo layerInfo) {
-        if (layerInfo != null && !layerInfo.isEnabled()) {
+    protected boolean isAvailable(String serviceType, ServiceInfo serviceInfo, PublishedInfo layerInfo) {
+        if (layerInfo != null && layerInfo instanceof LayerInfo) {
+            GWCResourceServiceVoter voter = new GWCResourceServiceVoter();
```
jodygarnett, line 64: Making use of voter to avoid duplicate logic

jodygarnett:
> @jodygarnett - looks good. However, there seems to be quite a bit changed and only one new test case. Is that ok?

Yeah the others were covered by an existing test case; indeed it caught a mistake during development. I was starting to duplicate logic before noticing I could reuse the existing "service voters" that perform a similar check.

PR #8227, opened 2025-01-24
[GEOS-11693] Replace jscolor.js with native colour picker

mprins:
[GEOS-11693](https://osgeo-org.atlassian.net/browse/GEOS-11693) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

Remove jscolor and use built-in (html 5) `color` type input

see also: https://github.com/geoserver/geoserver/pull/8177

# Checklist


sikeoka:
I

sikeoka:
I tested this out and a major problem with this is that the HTML5 color picker does not support setting a null value which GeoServer needs. There are workarounds available that all require custom JavaScript code but I don't know how easily they could be integrated with Wicket.

Another more minor problem is that there is some conversion code adding an extra `#` to the color value that is saved in the catalog (`##000000` instead of `#000000`).

aaime:
Inability to unset the color is a show stopper indeed.

mprins:
I missed that the colour is a dual-state field (`null`/empty string or a valid hex code); that is indeed a showstopper at least until https://github.com/whatwg/html/issues/9572 is resolved. Replacing the current javascript with other javascript defeats the purpose of this enhancement

PR #8232, opened 2025-01-27
[GEOS-11694] OpenID connect: allow caching authentication when an expiration is declared in the access token

aaime:
[GEOS-11694](https://osgeo-org.atlassian.net/browse/GEOS-11694) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

See ticket and new docs for details

# Checklist

### DIFF
--- a/src/community/security/oauth2/oauth2-core/src/main/java/org/geoserver/security/oauth2/GeoServerOAuthAuthenticationFilter.java
+++ b/src/community/security/oauth2/oauth2-core/src/main/java/org/geoserver/security/oauth2/GeoServerOAuthAuthenticationFilter.java
@@ -110,11 +110,15 @@ public abstract class GeoServerOAuthAuthenticationFilter extends GeoServerPreAut
     @Override
     public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
             throws IOException, ServletException {
+        // try authenticating from cache (token caching mechanism). If the authentication is found in
+        // the cache, the cache key will be null and the authentication set in the security context,
+        // otherwise, the cache key will be returned and the authentication in the context will remain unset
+        String cacheKey = authenticateFromCache(this, (HttpServletRequest) request);
         Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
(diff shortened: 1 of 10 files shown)


src/community/security/oauth2/oauth2-core/src/main/java/org/geoserver/security/oauth2/GeoServerOAuthAuthenticationFilter.java
```diff
@@ -145,22 +147,30 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
 
                 Authentication postAuthentication =
                         SecurityContextHolder.getContext().getAuthentication();
                 if (postAuthentication != null) {
-                    if (cacheAuthentication(postAuthentication, (HttpServletRequest) request)) {
-                        getSecurityManager()
-                                .getAuthenticationCache()
-                                .put(getName(), getCacheKey((HttpServletRequest) request), postAuthentication);
-                    }
+                    tryCacheAuthentication((HttpServletRequest) request, cacheKey, postAuthentication);
```
afabiani, line 156: Heads up... the `authenticateFromCache` form the `GeoServerSecurityFilter` class seems to be returning a `NULL` `cacheKey` whenever the `getSecurityManager().getAuthenticationCache().get(getName(), cacheKey)` is `HIT`... could this have an impact on the code? The `tryCacheAuthentication` could throw an exception if the `cacheKey` is null?

I don't fully understand the `authenticateFromCache` logic, but we should problably also check for the `cacheKey` validity here.

aaime: It's actually going to work out. If the cache is hit we don't need to authenticate again, and the authentication is set inside the SecurityContextHolder, so the cacheKey is not used any more. 
That said, just for extra safety, I've guarded the "tryCache" with a non-null check and clarified with a comment what is going on.

PR #8254, opened 2025-02-04
[GEOS-11702] GeoFence: major libs update

etj:
Update many libs according to [GeoFence 3.8](https://github.com/geoserver/geofence/pull/266)

Major points:
- update of hibernate / hibernate spatial
- use of an H2 version not compatible with other plugins and previous geofence version. This part has been documented in the revised doc pages (installation/migrations).

GeoFence 3.8 SNAPSHOT is not merged into geofence/main yet, but has been already built and deployed in the osgeo maven repo.

# Checklist


doc/en/user/source/extensions/geofence-server/installing.rst
```diff
@@ -3,18 +3,83 @@
...
    
-   The download link will be in the :guilabel:`Extensions` section under :guilabel:`Other`.
+    The download link will be in the :guilabel:`Extensions` section under :guilabel:`Other`.
    
-   .. warning:: Ensure to match plugin (example |release| above) version to the version of the GeoServer instance.
+    Make sure to match the plugin version (e.g. |release| above) to the version of the GeoServer instance.
 
  #. Extract the files in this archive to the :file:`WEB-INF/lib` directory of your GeoServer installation.
+ 
+    .. warning:: this plugin will install a version of the `H2 <http://www.h2database.com>`__  library that **is not compatible**
```
aaime, line 18: I would indicate that it's also safe to remove (it is, right?), in other words, that it's possible to use GeoFence with NetCDF/Grib provided the H2 v2 database is removed and the rule storage is moved to an external database.

I see it's written down in the same page, but some careless user might not read to the end, better to inform them now and indicate there are instructions later in the page.

etj:
There's a single error when compiling with JDK21:
```
Java 21 (65) is not supported by the current version of Byte Buddy which officially supports Java 20 (64) - update Byte Buddy or set net.bytebuddy.experimental as a VM property
```

Byte Buddy is used by hibernate instead of javaassist since version 5.3CR2 (where long ago javassist replaced cglib)

src/web/app/pom.xml
```diff
@@ -1073,6 +1073,14 @@
           <artifactId>gs-geofence-server</artifactId>
           <version>${project.version}</version>
         </dependency>
+        <!-- Older versions of h2 fail in drop/create tables   -->
```
aaime, line 1076: Is it needed? The dependency towards gs-geofence-server should be enough, it's already declared as a dependency there.

src/extension/geofence/geofence-server/pom.xml
```diff
@@ -45,19 +30,30 @@
...
-      <artifactId>hibernate-spatial</artifactId>
+      <groupId>org.opengeo</groupId>
+      <artifactId>geodb</artifactId>
+      <version>${opengeo-geodb.version}</version>
     </dependency>
+
+    <!-- Older versions of h2 fails in drop/create tables   -->
+    <!-- This version may break netcdf/grib modules         -->
+    <!-- Please do consider using a postgis backend instead -->
     <dependency>
```
aaime, line 43: Wondering if we shoud have a profile to enable/disable this dependency.
Use case, starting GeoServer during development with both geofence and netcdf, with this setup it would be impossible...

aaime (changes requested):

src/web/app/pom.xml
etj: Without it, mvn will usee the 1.1 version :|

aaime:
Looks like we have a legit failure under Java 21:

```
2025-02-04T15:09:34.6500839Z WARNING: Dynamic loading of agents will be disallowed by default in a future release
2025-02-04T15:09:35.1389246Z 15:09:35,138 [ERROR] Tests run: 4, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 0.992 s <<< FAILURE! - in org.geoserver.geofence.server.internal.SecurityContextUserResolverTest
2025-02-04T15:09:35.1394297Z 15:09:35,139 [ERROR] getRoles(org.geoserver.geofence.server.internal.SecurityContextUserResolverTest)  Time elapsed: 0.838 s  <<< ERROR!
2025-02-04T15:09:35.1395473Z org.mockito.exceptions.base.MockitoException: 
2025-02-04T15:09:35.1396042Z 
```

while not yet officially supported, we're keeping the entire GeoServer working under Java 21 too.

src/extension/geofence/geofence-server/pom.xml
etj: I guess that during development we need to work on the `src/web/app/pom.xml`.
We may add the profiles
 - `geofence-server-pg` --> includes postgres + postgis jars
 - `geofence-server-h2` --> includes h2 + geodb jars

Only specifying the profile `geofence-server` will not include any jdbc driver (feedback needed here @aaime)

I guess that in the same way we could create 2 different assembly files, one containing H2 related libs and the other postgres+postgis.

etj:
@aaime about profiles:

I guess that during development we need to work on the `src/web/app/pom.xml`.
We may add the profiles
 - `geofence-server-pg` --> includes postgres + postgis jars
 - `geofence-server-h2` --> includes h2 + geodb jars
 - while, only specifying the profile `geofence-server`, no jdbc driver will be included.

@aaime:  feedback needed here:
the `geofence-server` profile would only be valid on top directory: it would build the module, but would not create a valid war (if not specified together with either `geofence-server-pg` or `geofence-server-h2` profile). I guess that this behaviour may break some convention, doent it?

As a separate configuration, I guess that in the same way we could create 2 different assembly files, one containing H2 related libs and the other postgres+postgis.

aaime:
Thanks for fixing the build.

Regarding the profiles for development, that sounds great. Development wise, there is no actual convention, each extension is looked after by one or more people, as long as they understand what it going on, we're good. I suggest to remove the geofence-server profile and leave the other two, forcing a choice when someone is about to choose which profile to enable. @afabiani opinions?

About the packaging, that sounds like a good idea too, forcing a choice on download. That can potentially break some users with scripts to download extensions and assemble GeoServer. If it's limited to the main branch and [there is upgrade instructions](https://docs.geoserver.org/latest/en/user/installation/upgrade.html#notes-on-upgrading-specific-versions)  there is no issue. Occasonaly we did upgrade instructions for a specific stable version too, but since they are breaking changes in a stable series, we normally do so only for security issues.

etj:
Recap on profiles and packaging:
- `-P geofence-server-pg` --> will build the `geofence-server` module and include geofence-server + pg deps in webapp
- `-P geofence-server-h2` --> will build the `geofence-server` module and include geofence-server + h2 deps in webapp
- `-P geofence-server` --> will only build the `geofence-server` module, nothing will be included in webapp

About packaging:
- `geofence-server-h2-plugin.zip` --> gefence server with h2 deps
- `geofence-server-postgres-plugin.zip` --> gefence server and pg deps

I had an idea about also having `geofence-server-plugin.zip` (geofence server with h2+pg deps)  as it is now, anyway the deploy will differ from previous versions (there was a hell of hibernatespatial deps to handle), so we may forget about any previous deployment procedure compatibility.

aaime (approved):

PR #8247, opened 2025-01-30
[GEOS-11657] New GeoServer modal dialog is not resizeable

davidblasby:
This PR is for adding Drag/Resize to the (wicket) GSModalWindow.  It also fixes a bug in the CRSPanel.

https://github.com/user-attachments/assets/bc3c02de-2c1a-44ec-8e97-021f95d111ff

1. Move/resize.  This is mostly done in Javascript - see `GSModalWindow.js`.
   * Drag was relatively easy to add, however, resize was not.  I found the layout of the GSModalWindow to be a bit complex - and couldnt get the resize control to be in the correct location.  I have simplified it (both html and css).  See the video, `modal.css` and `GSModalWindow$ContentsPanel.html`.
   * There are a few, minor, changes between the old (i.e. years ago in the earlier wicket), the newer GSModalWindow (doesn't have resize/drag).  Please see the video.  

3. On the WCSRequest page (demos).  There are two CRSs - the `EPSG:..` is a link, but it wasn't doing anything.  This was a side effect of changing the coverage - the changes were not recognized by the two CRSPanels.

image

 For those interested, the original GSModalWindow PR (now closed and superseded by other work) is here - https://github.com/geoserver/geoserver/pull/7871

# Checklist

- [x] I have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/main/CONTRIBUTING.md).
- [x] I have sent a [Contribution Licence Agreement](https://docs.geoserver.org/latest/en/developer/policies/committing.html) (not required for small changes, e.g., fixing typos in documentation).
- [x] First PR targets the `main` branch (backports managed later; ignore for branch specific issues).
- [x] All the build checks are green ([see automated QA checks](https://docs.geoserver.org/latest/en/developer/qa-guide/index.html)).

For core and extension modules:

- [n/a] New unit tests have been added covering the changes.
- [n/a] [Documentation](https://github.com/geoserver/geoserver/tree/main/doc/en/user/source) has been updated (if change is visible to end users).
- [n/a] The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers).
- [x] There is an issue in the [GeoServer Jira](https://osgeo-org.atlassian.net/browse/GEOS/summary) (except for changes that do not affect administrators or end users in any way).
- [x] Commit message(s) must be in the form ``[GEOS-XYZWV] Title of the Jira ticket``.
[...]

PR #8259, opened 2025-02-05
[GEOS-11706] fix WCS Request panel - get CRS details (both)

jodygarnett:
[GEOS-11706](https://osgeo-org.atlassian.net/browse/GEOS-11706) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

Created from @davidblasby https://github.com/geoserver/geoserver/pull/8247, breakout out as a distinct PR so it can be back ported to 2.26:
  
On the WCSRequest page (demos).  There are two CRSs - the `EPSG:..` is a link, but it wasn't doing anything.  This was a side effect of changing the coverage - the changes were not recognized by the two CRSPanels.

image

# Checklist

### DIFF
--- a/src/web/core/src/main/java/org/geoserver/web/wicket/CRSPanel.java
+++ b/src/web/core/src/main/java/org/geoserver/web/wicket/CRSPanel.java
@@ -143,6 +143,11 @@ public class CRSPanel extends FormComponentPanel<CoordinateReferenceSystem> {
         initComponents();
     }
 
+    @Override
+    protected void onModelChanged() {
+        wktLink.setEnabled(getModelObject() != null);
+    }
+
     /*
      * helper for internally creating the panel.
      */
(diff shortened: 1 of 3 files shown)


jodygarnett (commented):
Because I split this out a distinct PR I cannot offer a review. I performed manual testing of this change on https://github.com/geoserver/geoserver/pull/8247 

WCSRequestBuilderPanel does not already have a test case and the effort required to add one for this small change is not warranted.

aaime:
" the effort required to add one for this small change is not warranted."... I strongly disagree on this approach, everyone has a different feeling about what a small change that does not need tests is, and what a hard test is. We go down this slope, and the requirement of having a test will soon mean nothing.

Did you even trying writing a test before declaring you don't need one?
I understand when something uses javascript or CSS, we'd need to use Selenium to write tests, but this one should have been possible to write using WicketTester.

As a PSC we need to be more clear about the few exceptions that would allow merging without a test. I'll add a discussion topic at the next meeting to work this out.

PR #8301, opened 2025-02-10
[GEOS-11716] Added resource error handling check while initializing WFS configuration.

axl8713:
[GEOS-11716](https://osgeo-org.atlassian.net/browse/GEOS-11716) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

If a layer is currently in an erroneous state (e.g. the backing source table has been deleted), POST requests to WFS GetFeature fail with an error, even when the request does not query the problematic layer.

These changes add the possibility to avoid failures in these cases, checking the value of global's `ResourceErrorHandling`.

  

# Checklist

### DIFF
--- a/src/wfs/src/main/java/org/geoserver/wfs/xml/WFSXmlUtils.java
+++ b/src/wfs/src/main/java/org/geoserver/wfs/xml/WFSXmlUtils.java
@@ -9,10 +9,13 @@ import java.io.Reader;
 import java.util.Collection;
 import java.util.Map;
 import java.util.Optional;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import javax.xml.namespace.QName;
 import org.geoserver.catalog.Catalog;
 import org.geoserver.catalog.FeatureTypeInfo;
 import org.geoserver.config.GeoServer;
+import org.geoserver.config.ResourceErrorHandling;
 import org.geoserver.ows.XmlRequestReader;
 import org.geoserver.platform.GeoServerExtensions;
 import org.geoserver.wfs.CatalogNamespaceSupport;
@@ -24,6 +27,7 @@ import org.geotools.api.referencing.crs.CoordinateReferenceSystem;
 import org.geotools.gml2.FeatureTypeCache;
 import org.geotools.gml2.SrsSyntax;
 import org.geotools.util.Converters;
+import org.geotools.util.logging.Logging;
 import org.geotools.xsd.Configuration;
 import org.geotools.xsd.OptionalComponentParameter;
 import org.geotools.xsd.Parser;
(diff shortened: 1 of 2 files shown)


groldan (approved):
Looks good to me. But wait for another reviewer before merging.

On a side note, while manually testing this, noticed it takes too long the first time if there are several FT's in the catalog. Added some tracing and got this:

```
-------- 5,069 FTs obtained from the Catalog in 87.76 ms -------
-------- 5,069 FTs added to cache in 1.712 min -------
```

wit this call:

```shell
time curl -s "http://localhost:8080/geoserver/wfs" -H "Content-Type: application/xml" -d '<?xml version="1.0" encoding="UTF-8"?>
<wfs:GetFeature
    service="WFS"
    version="2.0.0"
    xmlns:tiger="http://www.census.gov"
    xmlns:wfs="http://www.opengis.net/wfs/2.0"
>
    <wfs:Query typeNames="tiger:poi"/>
</wfs:GetFeature>' > /dev/null

0.01s user 0.01s system 0% cpu 2:23.87 total
```
That is, 2 minutes 23 secs for a first response, of which 1.7 minutes is populating the cache.
The datadirectory has 1,010 copies of the `ne` workspace, and the rest of the time comes from `FeatureTypeInfo.getFeatureType()` calling `ResourcePool.getFeatureType(FeatureTypeInfo)` which populates the resource pool cache and incurs in data I/O while doing so.

But this also seems to work:

```java
    public static void initWfsConfiguration(
            Configuration config, GeoServer gs, FeatureTypeSchemaBuilder schemaBuilder) {

        MutablePicoContainer context = config.getContext();

        FeatureTypeCache featureTypeCache = new CatalogFeatureTypeCache(gs.getCatalog());

        // add the wfs handler factory to handle feature elements
        context.registerComponentInstance(featureTypeCache);
        context.registerComponentInstance(new WFSHandlerFactory(gs.getCatalog(), schemaBuilder));
    }
```

and the response time is now:

```shell
time curl -s "http://localhost:8080/geoserver/wfs" -H "Content-Type: application/xml" -d '<?xml version="1.0" encoding="UTF-8"?>
<wfs:GetFeature
    service="WFS"
    version="2.0.0"
    xmlns:tiger="http://www.census.gov"
    xmlns:wfs="http://www.opengis.net/wfs/2.0"
>
    <wfs:Query typeNames="tiger:poi"/>
</wfs:GetFeature>' > /dev/null

0.01s user 0.01s system 0% cpu 41.319 total
```

@aaime do you foresee any trouble? we could create a separate issue to further pursue it

aaime:
@groldan I did not know that class existed to start with, but it's already used elsewhere so... maybe trouble-free? The javadoc talks about simple features only, not sure if it might have consequences on complex ones, something to test out. 
Anyhow, not directly related to this PR, I suggest we move the discussion in a new PR, to avoid confusing new contributors like @axl8713

PR #8325, opened 2025-02-13
[GEOS-11720] Fix quoting issues in AttributeTypeInfoImpl

olsen232:
At least two issues found here:
1. Simply being able to parse a string as ECQL doesn't tell us if the resulting ECQL would do what we want - what we need to know, before deciding to apply quotes, is if the ECQL parser would currently parse this field name as a PropertyName (what we want) or something else.
2. Not all ECQL parsing failures are CQLException - there is also EmptyStackException and there may be more.

I found this issue by trying to create a feature type with a field called "POINT". This failed with the following stack trace:

java.util.EmptyStackException
    at java.base/java.util.Stack.peek(Unknown Source)
    at java.base/java.util.Stack.pop(Unknown Source)
    -- lots of ECQL stack trace removed for brevity --
    at org.geotools.filter.text.ecql.ECQL.toExpression(ECQL.java:125)
    at org.geoserver.catalog.impl.AttributeTypeInfoImpl.getSource(AttributeTypeInfoImpl.java:147)

# Checklist

- [ ] `N/A` [Documentation](https://github.com/geoserver/geoserver/tree/main/doc/en/user/source) has been updated (if change is visible to end users).
- [ ] `N/A` The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers). 

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/impl/AttributeTypeInfoImpl.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/AttributeTypeInfoImpl.java
@@ -11,8 +11,8 @@ import org.geoserver.catalog.FeatureTypeInfo;
 import org.geoserver.catalog.MetadataMap;
 import org.geoserver.util.InternationalStringUtils;
 import org.geotools.api.feature.type.AttributeDescriptor;
+import org.geotools.api.filter.expression.PropertyName;
 import org.geotools.api.util.InternationalString;
-import org.geotools.filter.text.cql2.CQLException;
 import org.geotools.filter.text.ecql.ECQL;
 import org.geotools.util.GrowableInternationalString;
 
@@ -142,18 +142,23 @@ public class AttributeTypeInfoImpl implements AttributeTypeInfo {
     @Override
     public String getSource() {
         if (source == null && name != null) {
-            try {
-                // is it usable as is?
-                ECQL.toExpression(name);
-                // even if parseable, dots should be escaped
-                if (name.contains(".")) return "\"" + name + "\"";
-                return name;
-            } catch (CQLException e) {
-                // quoting to avoid reserved keyword issues
-                return "\"" + name + "\"";
-            }
+            return needsQuoting(name) ? ("\"" + name + "\"") : name;
         }
         return source;
+    }
+
+    private static boolean needsQuoting(String name) {
+        if (name.contains(".")) {
+            return true;
+        }
+        try {
+            if (!(ECQL.toExpression(name) instanceof PropertyName)) {
+                return true;
+            }
+        } catch (Exception e) {
+            return true;
+        }
+        return false;
     }
 
     @Override
--- a/src/main/src/test/java/org/geoserver/catalog/impl/AttributeTypeInfoImplTest.java
+++ b/src/main/src/test/java/org/geoserver/catalog/impl/AttributeTypeInfoImplTest.java
@@ -1,0 +1,40 @@
+/* (c) 2025 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.catalog.impl;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
+public class AttributeTypeInfoImplTest {
+    @Test
+    public void testQuoting() {
+        assertNotQuoted("foo");
+        assertNotQuoted("foo123");
+        assertNotQuoted("FOO");
+        assertNotQuoted("FOO123");
+        assertNotQuoted("foo_bar");
+        assertNotQuoted("Foo_Bar");
+
+        assertQuoted("123foo");
+        assertQuoted("foo.bar");
+        assertQuoted("point");
+        assertQuoted("POINT");
+        assertQuoted("BBOX");
+        assertQuoted("intersects");
+    }
+
+    private void assertNotQuoted(String name) {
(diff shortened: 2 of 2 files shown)


src/main/src/main/java/org/geoserver/catalog/impl/AttributeTypeInfoImpl.java
```diff
@@ -142,20 +142,25 @@ public String getRawSource() {
...
         }
         return source;
     }
 
+    private static boolean needsQuoting(String name) {
+        if (name.contains(".")) {
+            return true;
+        }
+        try {
+            if (!(ECQL.toExpression(name) instanceof PropertyName)) {
```
aaime, line 155: So how does one create a new attribute based on an expression now?
Say I have two attributes A and B, and want to create a new one, called C = A + B?

olsen232: I believe in that case, it is up to the calling code to set the source explicitly.
This code is only used for an AttributeTypeInfo where the name has been set but the source has not been set.
In this case, if the name is N, then the source is simply the column called N - unless the column called N is actually called N.N or something else that needs quoting, in which case the source is the quoted string "N.N". 

I'm just trying to fix the bug in the detection of what kind of names quoting. If you believe this was not the original intended behaviour, or if this is the intended behaviour but it should be redesigned in some way, let me know what you propose.

It is currently possible to trigger this issue using the REST API in something like the following manner:
```
POST http://geoserver/rest/workspaces/WORKSPACE/datastores/DATASTORE/featuretypes.json
...
"attributes": {
      "attribute": [
        {
          "name": "day-of-the-week",
          "minOccurs": 0,
          "maxOccurs": 1,
          "nillable": true,
          "binding": "java.lang.Integer"
        },
      }
```
I don't think the expected behaviour in this case would be that "day-of-the-week" would be evaluated as an expression with 4 variables and 3 minus operators, but I leave it up to you to decide.

olsen232: I've thought about this a bit more and I want to try and make a case that the fix I've made takes the current slightly broken behaviour, and fixes it to something more consistent. As before, you can, if you want, redesign the behaviour to something totally different.

In all cases, the behaviour happens when you create an AttributeTypeInfo with a name but no source, and a source needs to be auto-generated.
Current behaviour: there are 4 cases.

Case 1: The name is already a valid property name. Eg: `day_of_week`
Behaviour: The auto-generated source is the same as the name: -> `day_of_week` so it looks for a source field of the same name.

Case 2: The name is not a valid property name, but is valid ECQL. Eg: `day-of-week`
Behaviour: The auto-generated source is the same as the name: -> `day-of-week` which is equivalent to `day - of - week` (ie, day minus of minus week)

Case 3: The name is not a valid property name and is also not valid ECQL. Eg `day-of-week-`
Behaviour: The auto-generated souce adds quotes around the name: `"day-of-week-"` so it looks for a source field of the same name.

Case 4: The name is not a valid property name, and is also not valid ECQL, but doesn't result in CQLException. Eg `POINT`
Behavior: Exception is thrown.

I think the differences between case 2, 3, and 4 are not sensible - it is surprising that supplying a name but no source sometimes causes the name to be treated as a PropertyName, even adding quotes as needed (case 3) but sometimes it chooses not to add those quotes (2) and sometimes it just crashes instead of adding quotes (4).
 
 I think there are 3 sensible options for what to in the case where the caller supplies a name but no source.
Option 1: assume that the name is intended as a PropertyName and *always* add quotes (when they are needed). This is what my PR does. Callers who want to add or subtract properties will have to set the source explicitly.

Option 2: assume the name is valid ECQL and *never* add quotes. Callers who want to add or subtract properties can do so just by setting a name like `field1+field2`, callers who want to reference fields containing special characters which need quoting will need to set the source explicitly.

Option 3: always require callers to set the source explicitly.
This one would be pretty simple - this little bit of code would go away and instead we could have something like 
```
public String getSource() {
  if (source == null) {
    throw new NullPointerException("source is not set");
  }
  return source;
}  
```

olsen232: What's the verdict? I just looked it up and I see you're the original author of this code so you have a lot of authority on what the intended behaviour is or should be in this piece of code. What should it mean to try and create an attribute with a tricky name like `day-of-week`, `day-of-week-` or `point` without setting the attribute source explicitly?

olsen232: PS I was previously using the REST API for such attributes in the manner shown above, it used to work in older versions of Geoserver but at some point it stopped working, so I started by treating this bug as a regression and trying to restore something like the original behaviour, but you may see it differently

aaime: Had a look, your fix makes sense. That said, you're probably one of the REST users that posted the full set of attributes in the payload, while that hasn't been needed for ages. If you don't need to customize the attributes, make yourself a favour and just don't include the attributes when creating a new feature type, GeoServer will look for them in the data source by itself (as a bonus, if any source attribute changes, GeoServer will just need a configuration "reset" to drop the feature type cache).

GeoServer will still return the list of attributes when reading the feature type, but will do so for your information, it really does not need to have it amongts its configuration

PR #8332, opened 2025-02-14
[GEOS-11722] Coverage view reader partially ignores multithreaded loading

aaime:
[GEOS-11722](https://osgeo-org.atlassian.net/browse/GEOS-11722) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

See ticket for details.
Testing wise, I've verified that enabling MT does not change the result, and checked with debugging that full MT loading is happening. I don't see a reasonable and clean way to actually test full MT loading is happening.

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/CoverageViewReader.java
+++ b/src/main/src/main/java/org/geoserver/catalog/CoverageViewReader.java
@@ -5,6 +5,8 @@
  */
 package org.geoserver.catalog;
 
+import static org.geoserver.catalog.CoverageView.EnvelopeCompositionType.INTERSECTION;
+
 import com.sun.media.imageioimpl.common.BogusColorSpace;
 import it.geosolutions.imageio.maskband.DatasetLayout;
 import it.geosolutions.imageio.utilities.ImageIOUtilities;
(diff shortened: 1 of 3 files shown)


dromagnoli (approved):
LGTM

src/main/src/main/java/org/geoserver/catalog/CoverageViewReader.java
```diff
@@ -99,6 +103,15 @@ public class CoverageViewReader implements GridCoverage2DReader {
 
     private static final Logger LOGGER = Logging.getLogger(CoverageViewReader.class);
 
+    /**
+     * Executor for parallel loading of coverages that do support multithreaded loading to start with (e.g., image
+     * msoaic). It's not limiting the number of threads, because when MT loading is one, that's already limited in the
```
dromagnoli, line 108:
```suggestion
     * mosaic). It's not limiting the number of threads, because when MT loading is one, that's already limited in the
```

src/main/src/main/java/org/geoserver/catalog/CoverageViewReader.java
```diff
@@ -99,6 +103,15 @@ public class CoverageViewReader implements GridCoverage2DReader {
...
+     * readers. It cannot be the same executor used by the image mosaic (ResourcePool#getCoverageExecutor) because that
+     * would cause deadlock (each read would use at least 2 threads, one inside the image mosaic and one inside the
+     * coverage view). Should we switch to allow parlale
```
dromagnoli, line 111:
```suggestion
     * coverage view). Should we switch to allow parallel
```

Is the statement complete?

dromagnoli (approved):

PR #8344, opened 2025-02-19
[GEOS-11724] DisabledServiceCheck check resource disabled service list

jodygarnett:
[GEOS-11724](https://osgeo-org.atlassian.net/browse/GEOS-11724) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This prevents dispatcher calling a disabled per-layer service.

# Checklist

### DIFF
--- a/src/extension/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/APIRequestInfo.java
+++ b/src/extension/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/APIRequestInfo.java
@@ -215,8 +215,9 @@ public class APIRequestInfo {
      * @return
      */
     public String getRequestPath() {
-        String pathInfo = request.getPathInfo();
         String servletPath = request.getServletPath();
+        String pathInfo = request.getPathInfo();
+
         return ResponseUtils.appendPath(servletPath, pathInfo);
     }
 
--- a/src/main/src/main/java/org/geoserver/ows/DisabledServiceCheck.java
+++ b/src/main/src/main/java/org/geoserver/ows/DisabledServiceCheck.java
@@ -7,12 +7,19 @@ package org.geoserver.ows;
 
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
+import java.util.List;
 import java.util.logging.Logger;
+import org.apache.commons.lang3.StringUtils;
+import org.geoserver.catalog.Catalog;
+import org.geoserver.catalog.LayerInfo;
+import org.geoserver.config.GeoServer;
 import org.geoserver.config.ServiceInfo;
 import org.geoserver.ows.util.OwsUtils;
 import org.geoserver.platform.Operation;
 import org.geoserver.platform.Service;
 import org.geoserver.platform.ServiceException;
+import org.geoserver.security.DisabledServiceResourceFilter;
+import org.geotools.util.decorate.Wrapper;
 import org.geotools.util.logging.Logging;
 
 /**
@@ -23,6 +30,12 @@ import org.geotools.util.logging.Logging;
 public class DisabledServiceCheck implements DispatcherCallback {
 
     static final Logger LOGGER = Logging.getLogger(DisabledServiceCheck.class);
+    /** GeoServer configuration */
+    private GeoServer geoServer;
+
+    public DisabledServiceCheck(GeoServer geoServer) {
+        this.geoServer = geoServer;
+    }
 
     @Override
     public Request init(Request request) {
@@ -63,6 +76,24 @@ public class DisabledServiceCheck implements DispatcherCallback {
                         throw new ServiceException(
                                 "Service " + info.getName() + " is disabled", ServiceException.SERVICE_UNAVAILABLE);
                     }
+                    // check if service is disabled for layer
+                    String context = context(request);
+                    if (context != null && context.contains("/")) {
+                        String layerName = context.replace("/", ":");
+                        LayerInfo layerInfo = getLayerByName(layerName);
+                        if (layerInfo != null) {
+                            List<String> disabledServices =
+                                    DisabledServiceResourceFilter.disabledServices(layerInfo.getResource());
+                            boolean disabled = disabledServices.stream()
+                                    .anyMatch(
+                                            serviceType -> StringUtils.equalsIgnoreCase(service.getId(), serviceType));
+                            if (disabled) {
+                                throw new ServiceException(
+                                        "Service " + info.getName() + " is disabled for layer " + layerName,
+                                        ServiceException.SERVICE_UNAVAILABLE);
+                            }
+                        }
+                    }
                 }
             } catch (Exception e) {
                 // TODO: log this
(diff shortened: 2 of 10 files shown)


jodygarnett:
@aaime I got an interesting challenge with APIDispatcher ...

``DisabledServiceCheck`` is seeing a request:

* Request.getContext(): ``ogc/features`` (should be empty for global service)
* Request.getHttpRequest().getServletPath(): ``ogc/``
* Request.getHttpRequest().getPath(): ``ogc/features/v1
* Request.getHttpRequest().getContextPath(): ``/geoserver/``

I am used to checking Request.getContext() to determine workspace and layer context for virtual services.

jodygarnett:
Finding out that Request.getContext() has different assumptions in different parts of the code:

* ``<workspace>/``
* ``<workspace>/<layer>``
* Or as documented in ``Dispatcher.initRequestContext(request)`` .. everything up until the last ``/``

The results in GWC dispatches failing as context ends up being ``gwc/service`` with code designed to check for that.

It looks like the way forward is to duplicate the work of ``OWSHandlerMapping`` and always look up workspace and layer by name and not rely on how request has parsed anything.

src/main/src/main/java/org/geoserver/ows/DisabledServiceCheck.java
```diff
@@ -63,6 +74,37 @@ public Service serviceDispatched(Request request, Service service) {
...
+                        } else {
+                            context = context.substring(0, context.indexOf("gwc/service"));
+                        }
+                    }
+                    if (context != null && context.contains("/")) {
+                        String layerName = context.replace("/", ":");
+                        LayerInfo layerInfo = geoServer.getCatalog().getLayerByName(layerName);
+                        if (layerInfo == null) {
+                            // DisabledServiceResourceFilter has checked getDisabledServices() list and hidden layer
+                            throw new ServiceException(
```
jodygarnett, line 92: Can anyone think of a way of preventing DisabledServiceResourceFilter from being applied here? If so I could look up the layerInfo and know specifically that it is disabled (rather than not found). 

This would be much better and prevent the need to do a special case for ``/gwc/service`` above.

jodygarnett:
Okay I am happy with this PR, while it fixes a bug it cannot be easily backport due to the refactoring of APIDispatcher between 2.26.x and main.

aaime:
I won't review a PR with 4 commits for the same Jira ticket, though I was clear in the last meeting.

jodygarnett:
> I won't review a PR with 4 commits for the same Jira ticket, though I was clear in the last meeting.

Happy to rebase, is Squash and Merge not an option?

aaime:
We discussed the option of having separate commits to handle review feedback, that is the one option to use squash-merge.

jodygarnett:
@aaime I am concerned about the failed WCS CITE test, can rebase when it is sorted out.

aaime:
The WFS 2.0 cite test failure happens because WFS 2.0 has time based locking tests, using short lock times. That interacts poorly with the github actions (loose clocks maybe?) and sometimes fails. You can restart the cite tests, it should run fine.
Unfortunately, I'm not aware of a simple way to skip that test.

jodygarnett:
Ok thanks I was worried about that failure after writing and rewriting the fix this week.

src/main/src/main/java/org/geoserver/ows/DisabledServiceCheck.java
```diff
@@ -81,6 +112,52 @@ public Service serviceDispatched(Request request, Service service) {
...
+     * @return The context.
+     */
+    String context(Request request) {
+        String context = request.getContext();
+        if (context != null && context.contains("gwc/service")) {
+            // Account for context provided to GwcServiceDispatcherCallback
+            if (context.contains("/gwc/service")) {
+                context = context.substring(0, context.indexOf("/gwc/service"));
+            } else {
+                context = context.substring(0, context.indexOf("gwc/service"));
```
jodygarnett, line 128: I guess with subsequent removal of leading and trailing `/` only this line is needed.

jodygarnett, line 128: I guess with subsequent removal of leading and trailing / only this line is needed.

```suggestion
            if (context.contains("gwc/service")) {
                context = context.substring(0, context.indexOf("gwc/service"));
```

src/main/src/main/java/org/geoserver/ows/DisabledServiceCheck.java
```diff
@@ -81,6 +112,52 @@ public Service serviceDispatched(Request request, Service service) {
...
+        // We need access to actual catalog, not filtered for current service
+        Catalog catalog = geoServer.getCatalog();
+        while (catalog instanceof Wrapper && ((Wrapper) catalog).isWrapperFor(Catalog.class)) {
+            Catalog unwrapped = ((Wrapper) catalog).unwrap(Catalog.class);
+            if (unwrapped == catalog || unwrapped == null) {
+                break;
+            }
+            catalog = unwrapped;
+        }
+        return catalog.getLayerByName(layerName);
```
aaime, line 158: Confused... when is catalog filtered "by service"'?

src/extension/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/APIRequestInfo.java
```diff
@@ -215,11 +215,37 @@ public Service getService() {
...
+     * @return request context
+     */
+    public String getContext() {
+        String landingPage = getServiceLandingPage();
+        String context = getRequestPath().substring(1);
+
+        if (context.contains(landingPage)) {
+            int idx = context.lastIndexOf(landingPage);
+            context = context.substring(0, idx);
+        } else if (landingPage.contains("/")) {
```
aaime, line 236: The landing page is typically "ogc/<myService</v<N>" so this check is basically always true, however, this code smees to be ecluding the "v1" from the context... are you sure it's the right call?

For reference, checking the equivalent OWS code:

```java
        String ctxPath = request.httpRequest.getContextPath();
        String reqPath = request.httpRequest.getRequestURI();
        reqPath = reqPath.substring(ctxPath.length());

        // strip off leading and trailing slashes
        if (reqPath.startsWith("/")) {
            reqPath = reqPath.substring(1, reqPath.length());
        }

        if (reqPath.endsWith("/")) {
            reqPath = reqPath.substring(0, reqPath.length() - 1);
        }

        String context = reqPath;
        String path = null;
        int index = context.lastIndexOf('/');
        if (index != -1) {
            path = context.substring(index + 1);
            context = context.substring(0, index);
        } else {
            path = reqPath;
            context = null;
        }
```

When I run it against ``http://localhost:8080/geoserver/ows?service=WFS&acceptversions=2.0.0&request=GetCapabilities`` I end up with 

* context: null
* path: ows

If I use a workspace-specific URL instead I get:
``http://localhost:8080/geoserver/cite/ows?service=WFS&acceptversions=2.0.0&request=GetCapabilities``

* context: cite
* path: ows

And finally, with a layer specific URL I get:
``http://localhost:8080/geoserver/cite/Lakes/ows?service=WFS&acceptversions=2.0.0&request=GetCapabilities``

* context: cite/Lakes
* path: ows

I would expect to get something similar when using OGC APIs, where the context is basically the "workspace and layer" and the path is what follows.

jodygarnett: I assumed that as well, that the content would be "workspace and layer" (which would be useful) ... and then found out that it was documented in the dispatcher as "everything before the servlet". Which is why ``v1`` was being excluded. 

I considered a more extensive change to make context be ``workspace and layer`` but geowebcache was depending on context "/gwc/sevice" for ``/gwc/service/wmts``.

So the answer is the code is working as intended, which is surprising to both of us.

See ``Dispatcher.initServiceContext`` javadocs for definition:

> The 'context' is the part of the URI before the last '/' and the 'path' is the part after the last '/'.

jodygarnett: The Dispatcher.initServiceContext also had some examples that were not visible in the javadocs. I moved them up and formatted the to be visible in javadocs, and included the GWC example as seeing that in the debugger helped make it clear to me.

src/main/src/main/java/org/geoserver/ows/DisabledServiceCheck.java
jodygarnett: When providing WMS then all the layers that have been disabled for WMS need to be filtered. 

This is the issue that is being addressed, producing a layer WMS that has no content. Rather than a 404.

src/extension/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/APIRequestInfo.java
jodygarnett: action plan:

- Ignore gwc example as something terrible
- Make APIRequestInfo context "workspace and layer" and consider the servlet `ogcapi/features/v1` for APIRequestInfo
- Make a note in Dispatcher javadoc

jodygarnett: This method was no longer required due to improvements in DisabledServiceCheck. I did add some checks to APIDispatcher to ensure the generated test HTTPRequest was produced as expected.

PR #8345, opened 2025-02-19
[GEOS-11725] Environment parameters resolving is not working on Smart data loader

fernandor777:
[GEOS-11725](https://osgeo-org.atlassian.net/browse/GEOS-11725) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

JIRA issue:
https://osgeo-org.atlassian.net/browse/GEOS-11725

When a GeoServer catalog have a Postgis datastore working with environment parametrization enabled and parametrizaed parameters including user and password, selecting this store from smart data loader throws an error trying to use the parametrizaed strings without performing the environment resolving.

This PR targets to fix the root problem, allowing smart data loader to reuse parametrized stores.
  

# Checklist


src/community/smart-data-loader/src/main/java/org/geoserver/smartdataloader/data/SmartDataLoaderDataAccessFactory.java
```diff
@@ -10,13 +10,7 @@
...
 import java.net.URL;
 import java.sql.SQLException;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
mprins, line 13: I'm pretty sure wildcard imports are not allowed, but somehow the QA / PMD workflow has not picked up on this... 

see also: https://github.com/geoserver/geoserver/blob/6c2ffdf81676c624c3216b680df0963a1968bcd8/build/qa/pmd-ruleset.xml#L105-L117

mprins (changes requested):
It seems the PMD profile is not triggered during the QA, while I think it should be?... (thus not failing build on wild-card imports)
Apart from that LGTM.

/cc @aaime

src/community/smart-data-loader/src/main/java/org/geoserver/smartdataloader/data/SmartDataLoaderDataAccessFactory.java
fernandor777: Hi @mprins thanks for your feedback, fixing it.

aaime:
Wondering if it's related to the recent PMD update?

src/community/smart-data-loader/src/main/java/org/geoserver/smartdataloader/data/SmartDataLoaderDataAccessFactory.java
```diff
@@ -284,9 +286,25 @@ private File saveMappingDocument(String pathname, String filename, Document mapp
...
     private DataStoreInfo getDataStoreInfo(Map<String, Serializable> params) throws IOException {
         String jdbcDataStoreId = lookup(DATASTORE_METADATA, params, String.class);
-        Catalog c = getGeoServer().getCatalog();
-        DataStoreInfo ds = c.getDataStore(jdbcDataStoreId);
-        return ds;
+        Catalog catalog = getGeoServer().getCatalog();
+        return catalog.getDataStore(jdbcDataStoreId);
+    }
+
+    static Map<String, Serializable> resolveParams(Map<String, Serializable> params) {
```
aaime, line 293: Good intention but done in the wrong place IMHO, stores should not be aware of environment parameterization (or we'd have to make the all aware).

The GeoServer resource pool is [already resolving environment parameters on its own here](https://github.com/geoserver/geoserver/blob/99ae245402fca51e306eaf9817faaa891bf89e73/src/main/src/main/java/org/geoserver/catalog/ResourcePool.java#L806).
Please try to see why it's not working as intended, and fix it instead.

I'd keep the tests class instead.

aaime (changes requested):

aaime:
Thinking out loud, since this is a community module it can also live with some tech debt, but I'd spend a bit of time at least figuring out why the core env substitution is not working.

fernandor777:
Thanks @aaime indeed I investigated and found the way to make resourcepool perform the env parameter resolving, I refactored the implementation and added a test integrating the env parameter on the database host.

src/community/smart-data-loader/src/main/java/org/geoserver/smartdataloader/data/SmartDataLoaderDataAccessFactory.java
```diff
@@ -284,9 +285,15 @@ private File saveMappingDocument(String pathname, String filename, Document mapp
...
     private DataStoreInfo getDataStoreInfo(Map<String, Serializable> params) throws IOException {
         String jdbcDataStoreId = lookup(DATASTORE_METADATA, params, String.class);
-        Catalog c = getGeoServer().getCatalog();
-        DataStoreInfo ds = c.getDataStore(jdbcDataStoreId);
-        return ds;
+        Catalog catalog = getGeoServer().getCatalog();
+        return catalog.getResourcePool().clone(catalog.getDataStore(jdbcDataStoreId), true);
+    }
+
+    private static Serializable resolveValue(Serializable value, GeoServerEnvironment env) {
```
aaime, line 292: Does not seem to be used anywhere?

src/community/smart-data-loader/src/test/java/org/geoserver/smartdataloader/data/store/EnvParametrizationStoreTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+    public EnvParametrizationStoreTest(JDBCFixtureHelper fixtureHelper) {
+        super(fixtureHelper);
+        NAMESPACE_PREFIX = "mt";
+        TARGET_NAMESPACE = "http://www.geo-solutions.it/smartappschema/1.0";
+    }
+
+    @BeforeClass
+    public static void testClassSetup() {
+        System.setProperty(ALLOW_ENV_PARAMETRIZATION, "true");
```
aaime, line 52: I would also have a test where env parametrization is not enabled, to check there is no substitution

aaime (commented):
Overall looking good, see comments below

src/community/smart-data-loader/src/main/java/org/geoserver/smartdataloader/data/SmartDataLoaderDataAccessFactory.java
fernandor777: Right, removed it.  Thanks.

fernandor777: Holy leftover, batman!

Thanks, removed.

aaime (approved):
Looks good , thanks!

PR #8347, opened 2025-02-24
DRAFT: Add health check endpoint to GeoServer

petersmythe:
Add an HTTP endpoint to perform a health check http://localhost:8080/geoserver/health-check and return 503 Service Unavailable when the catalog is being loaded or reloaded, otherwise 200 Success.

* **HealthCheckEndpoint.java**: Create a new class `HealthCheckEndpoint` to handle the HTTP endpoint for health checks. Implement a method `isCatalogLoading` to return the catalog loading status. Use `@RestController` and `@GetMapping` annotations to define the endpoint.
* **GeoServer.java**: Add a new method `isCatalogLoading` to check the catalog loading status.
* **GeoServerConfigPersister.java**: Update the `handleAddEvent`, `handleModifyEvent`, and `handleRemoveEvent` methods to set the catalog loading status. Add a field to track the catalog loading status.

### DIFF
--- a/src/main/src/main/java/org/geoserver/config/CatalogTimeStampUpdater.java
+++ b/src/main/src/main/java/org/geoserver/config/CatalogTimeStampUpdater.java
@@ -31,6 +31,7 @@ public class CatalogTimeStampUpdater implements CatalogListener {
     static final Logger LOGGER = Logging.getLogger(CatalogTimeStampUpdater.class);
 
     Catalog catalog;
+    private boolean catalogLoading;
 
     public CatalogTimeStampUpdater(Catalog catalog) {
 
@@ -56,10 +57,14 @@ public class CatalogTimeStampUpdater implements CatalogListener {
     }
 
     @Override
-    public void handleAddEvent(CatalogAddEvent event) throws CatalogException {}
+    public void handleAddEvent(CatalogAddEvent event) throws CatalogException {
+        catalogLoading = true;
+    }
 
     @Override
-    public void handleRemoveEvent(CatalogRemoveEvent event) throws CatalogException {}
+    public void handleRemoveEvent(CatalogRemoveEvent event) throws CatalogException {
+        catalogLoading = true;
+    }
 
     @Override
     public void handleModifyEvent(CatalogModifyEvent event) throws CatalogException {
@@ -74,12 +79,21 @@ public class CatalogTimeStampUpdater implements CatalogListener {
             Date dateModified = new Date();
             info.setDateModified(dateModified);
         }
+        catalogLoading = true;
         LOGGER.finest(event.toString() + " :handleModifyEvent");
     }
 
     @Override
-    public void handlePostModifyEvent(CatalogPostModifyEvent event) throws CatalogException {}
+    public void handlePostModifyEvent(CatalogPostModifyEvent event) throws CatalogException {
+        catalogLoading = false;
+    }
 
     @Override
-    public void reloaded() {}
+    public void reloaded() {
+        catalogLoading = false;
+    }
+
+    public boolean isCatalogLoading() {
+        return catalogLoading;
+    }
 }
--- a/src/main/src/main/java/org/geoserver/config/GeoServer.java
+++ b/src/main/src/main/java/org/geoserver/config/GeoServer.java
@@ -250,4 +250,7 @@ public interface GeoServer {
      * new one
      */
     void reload(Catalog catalog) throws Exception;
+
+    /** Checks if the catalog is currently loading or reloading. */
+    boolean isCatalogLoading();
 }
--- a/src/main/src/main/java/org/geoserver/config/GeoServerConfigPersister.java
+++ b/src/main/src/main/java/org/geoserver/config/GeoServerConfigPersister.java
@@ -50,6 +50,8 @@ public class GeoServerConfigPersister implements CatalogListener, ConfigurationL
     GeoServerDataDirectory dd;
     XStreamPersister xp;
 
+    private boolean catalogLoading;
+
     public GeoServerConfigPersister(GeoServerResourceLoader rl, XStreamPersister xp) {
         this.rl = rl;
         this.dd = new GeoServerDataDirectory(rl);
@@ -58,6 +60,7 @@ public class GeoServerConfigPersister implements CatalogListener, ConfigurationL
 
     @Override
     public void handleAddEvent(CatalogAddEvent event) {
+        catalogLoading = true;
         Object source = event.getSource();
         try {
             if (source instanceof WorkspaceInfo) {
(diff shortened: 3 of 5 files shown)


groldan:
Hi Pete, this is interesting, though I don't understand what `CatalogTimeStampUpdater` has to do with health checks, or why the app would be deemed unavailable whenever a CatalogInfo object is added/modified/removed.

Maybe it's just an exploratory prototype, but in any case, it reminds me @aaime asked some time ago if we could integrate spring-boot-actuator in GeoServer despite not being a Spring Boot app.

I just did a quick check and indeed it'd be possible without carrying over a lot of spring-boot transitive dependencies.

It would give us a framework to aggregate health checks (including liveness/readiness probes, as used in Kubernetes).

If you're interested I'd like us to pair up and follow up on this approach: module

Integrating Spring Boot Actuator into GeoServer as a community:

- **Enhanced Monitoring with Health Probes:** Actuator’s `/actuator/health` endpoint aggregates health checks (e.g., database, services), reporting a status like `{"status": "UP", "components": {"catalog": {"status": "UP"}}}`. This provides immediate visibility into GeoServer’s health.
- **Cloud-Native Readiness with Liveness and Readiness Probes:** Actuator offers `/actuator/health/liveness` (ensures GeoServer is running, e.g., `{"status": "UP", "details": {"core": "Alive"}}`) and `/actuator/health/readiness` (confirms it’s ready to serve, e.g., `{"status": "UP", "components": {"catalog": {"status": "UP", "details": "Catalog initialized"}}}`). In Kubernetes, liveness probes restart GeoServer on failure, while readiness probes manage traffic—boosting reliability and scalability.
- **How Probes Work in Actuator:** Health probes compile data from `HealthIndicator` beans, customizable for GeoServer needs. Liveness and readiness are health groups (e.g., `management.endpoint.health.group.readiness.include=catalog`), splitting checks into targeted endpoints. For example, a readiness probe might fail if the Catalog isn’t ready: `{"status": "DOWN", "components": {"catalog": {"status": "DOWN", "details": "Catalog reloading"}}}`.
- **Extensibility:** Developers can craft indicators for GeoServer’s geospatial components (e.g., Catalog, WMS), offering precise diagnostics without altering core code, enhancing adaptability.
- **Minimal Overhead:** Leveraging GeoServer’s Spring framework, Actuator integrates as a lightweight JAR in `WEB-INF/lib`, avoiding heavy dependencies while maintaining efficiency.
- **Additional Endpoints for Insight:** Beyond health, Actuator enables endpoints like `/actuator/metrics` (performance stats, e.g., JVM memory usage), `/actuator/info` (app metadata, e.g., version details), `/actuator/env` (environment properties), and more. We could check which ones to include in the future. A good practice is to let `/actuator/health/**` be accessed anonymously, while other endpoints like `/actuator/metrics` require authentication.

petersmythe:
> ... or why the app would be deemed unavailable whenever a CatalogInfo object is added/modified/removed.

Hi @groldan , I have discovered that with a very large catalog, it can take our GeoServer over 1 minute extra after starting up to be ready to serve layers, and the same delay when reloading config (GUI or REST API), and during this time, some layers are available and some layers are not, depending on where they fall within the config files, thus the overall readiness status should reflect negative.

I would love to work with you on the full spring framework, but I don't think it'll get any attention with the current GeoServer 3 focus for a long while still. 

In the meanwhile, this simple addition (while imperfect) would assist AfriGIS rolling deployments, and I would like to get it into 2.28, if possible, rather than building our own fork.

Can I resurrect this PR now, and then we work on your plan next year?

groldan:
Hi @petersmythe, definitely, I never meant to slow you down.

> I have discovered that with a very large catalog, it can take our GeoServer over 1 minute extra after starting up to be ready to serve layers

That might be more related to loading the gwc layers and/or `LayerGroupContainmentCache` double initialization (at its constructor and onApplicationEvent()) than the catalog load process itself. If so, the flag won't help you much I'm afraid.

So, while I sympathize with you moving this forward, I still have some concerns on the approach:
- Ideally we wouldn't need to touch core code to implement this
- The controller wouldn't belong to the gs-main module. I'd rather make it a community module. If we add it directly to a core module we're committing to the service interface
- With GeoServerTimestampUpdater being a CatalogListener and setting `catalogLoading = true` on each method it means during normal operations geoserver will be deemed unhealthy/healthy constantly, which isn't true. When a catalog object is changed geoserver is not unhealthy until the object is saved.

Unless I'm missing something, I think the `GeoServerTimestampUpdater` bean should instead implement `GeoServerReinitializer`, so that during startup and at `beforeReinitialize()`, its `catalogLoading` flag is `true`, and is set to `false` (i.e. ready) on `initialize(GeoServer)`. Or maybe  implement `GeoServerLifecycleHandler.beforeReload()/onReload()` instead? @aaime ? I'm getting a bit lost.

Finally I'd advise to be cautious on the semantics of healthy/unhealthy vs ready. If you'll be using the health endpoint in a containerized environment, marking the app un unhealthy can lead to the pod being shutdown or requests not being propagated to.

My understanding is you want to use the health endpoint as a synonym of "readiness" instead of "liveness".
A liveness (health) check determines if an application is running and healthy, triggering a restart if it's not. A readiness check verifies if the application is ready to serve requests, pausing traffic until it can handle requests without errors.

In my mind it means the app is healthy when it's re/loading the catalog, it just may not be ready to process requests, but that's no reason to kill it.

To summarize, my advise would be:
- community module
- one endpoint for liveness/health checks (e.g. `/health/liveness`), this one returns 200 `{"status": "UP"}` for as long as the app can be reached, by lack of better checks, but it'd give us room for improvement in the future.
- one endpoint for readiness checks (e.g. `/health/readiness`), this one will return 503 `{"status": "DOWN"}` while geoserver is re/loading, and 200 `{"status": "UP"}` when it's ready.

If we want to prepare even further for spring actuator, we can set the endpoints to `/management/health/liveness|readiness` instead, which is another common URI pattern used for the actuator and may even make more sense semantically.

Does that make any sense? comments?

groldan:
btw, I thought `CatalogTimeStampUpdater` was a new bean but it's not. You should create a new bean to track for this nonetheless

aaime:
The CatalogTimeStampUpdater should not be touched yes, a new bean is a better choice, also having a new community module means that you can move with basically no restrictions. Catalog events happen all the time during normal operation, so indeed a health check should not be based on them.

Catalog reload does not stop operation either, if memory serves me right GeoServer is running on top of the previous catalog while the re-loading happens (for OGC requests). Long story short, I would suggest ignoring loading/resetting events, and only focusing on startup complete events.

It seems to me what you want is instead to implement an ApplicationListener<ContextRefreshedEvent>:

```
public class StartupListener implements ApplicationListener<ContextRefreshedEvent> {
    @Override
    public void onApplicationEvent(ContextRefreshedEvent event) {
        System.out.println("Spring context is ready, web app can serve requests.");
    }
}
```

That should trigger when the application context is fully loaded, rather than when the catalog is loaded.

@groldan for the double initializations feel free to send over a pull request to amend them, the faster GeoServer starts up, the better for everyone.

petersmythe:
Thank you for the input Gabe & Andrea, I will do that.

PR #8357, opened 2025-02-27
[GEOS-11738] Prevent error when oidc provider sends empty "&state="

pazepaze:
This fixes the bug I described in GEOS-11738.

When using Dex IdP as OpenId Connect provider with the oauth2-openid-connect, an Exception is thrown in 
AuthorizationCodeAccessTokenProvider  from spring-security-oauth2 when logging in via the OIDC login button in the GeoServer web interface.

This occurs because the GeoServer oidc plugin does not set the (optional, protects against CSRF) state parameter in the request to the oidc provider. When using Dex IdP, it responds with an empty "&state=" parameter when redirecting the browser back to GeoServer with the "code". I am not sure whether this behavior by Dex violates some oidc spec or not.

The root cause is in [this line](https://github.com/spring-attic/spring-security-oauth/blob/0344de8cc75055742f638b1e476a646a6f5891d9/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/client/token/grant/code/AuthorizationCodeAccessTokenProvider.java#L255) in AuthorizationCodeAccessTokenProvider, which throws an Exception when the state parameter is present, but empty.

Everything works if we just remove the state parameter from the request back to GeoServer, if it is empty anyways. The cases "state parameter is present but empty" and "no state parameter present at all" don't need to be distinguished as far as I understand oidc.

Admittedly, this "fix" would not be needed when fully implementing actually setting the state parameter instead. But if noone has the time to do that, this is an improvement.

# Checklist

### DIFF
--- a/src/community/security/oauth2-openid-connect/oauth2-openid-connect-core/src/main/java/org/geoserver/security/oauth2/ValidatingOAuth2RestTemplate.java
+++ b/src/community/security/oauth2-openid-connect/oauth2-openid-connect-core/src/main/java/org/geoserver/security/oauth2/ValidatingOAuth2RestTemplate.java
@@ -45,6 +45,12 @@ class ValidatingOAuth2RestTemplate extends OAuth2RestTemplate {
 
         OAuth2AccessToken result = null;
         try {
+            // prevent error in AuthorizationCodeAccessTokenProvider from spring-security-oauth2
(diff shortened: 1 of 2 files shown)


aaime:
Thanks, looks good. The OSX github action failure is unrelated (random error).

PR #8366, opened 2025-03-04
[GEOS-11747] GeoServer does not throw JAI runtime exceptions

sikeoka:
[GEOS-11747](https://osgeo-org.atlassian.net/browse/GEOS-11747) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This PR updates GeoServer's ImagingListener to throw runtime exceptions that don't come from an OperationRegistry and don't meet the other conditions to hide an exception. It was also moved into a separate method that is called in GeoServerSystemTestSupport so that all GeoServer unit tests now use the GeoServer listener instead of the GeoTools one. The GeoTools listener has the same if statement to throw JAI runtime exceptions so there was different behavior between the behavior in unit tests and when GeoServer was deployed in a server.

This PR is related to https://github.com/geosolutions-it/jai-ext/pull/307
  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/GeoserverInitStartupListener.java
+++ b/src/main/src/main/java/org/geoserver/GeoserverInitStartupListener.java
@@ -91,38 +91,7 @@ public class GeoserverInitStartupListener implements ServletContextListener {
         LOGGER.config("Logging policy: " + policy);
         GeoTools.init((Hints) null);
 
-        // Custom GeoTools ImagingListener used to ignore common warnings
-        JAI.getDefaultInstance().setImagingListener(new ImagingListener() {
-            final Logger LOGGER = Logging.getLogger("javax.media.jai");
-
-            @Override
-            public boolean errorOccurred(String message, Throwable thrown, Object where, boolean isRetryable)
-                    throws RuntimeException {
-                if (isSerializableRenderedImageFinalization(where, thrown)) {
-                    LOGGER.log(Level.FINEST, message, thrown);
-                } else if (message.contains("Continuing in pure Java mode")) {
-                    LOGGER.log(Level.FINE, message, thrown);
-                } else {
-                    LOGGER.log(Level.INFO, message, thrown);
-                }
-                return false; // we are not trying to recover
-            }
-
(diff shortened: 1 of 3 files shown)


src/main/src/main/java/org/geoserver/GeoserverInitStartupListener.java
```diff
@@ -220,6 +189,49 @@ private boolean isSerializableRenderedImageFinalization(Object where, Throwable
...
 
+    /** Initializes the default JAI ImagingListener */
+    public static void initJAIDefaultInstance() {
+        JAI jaiDef = JAI.getDefaultInstance();
+        // Custom ImagingListener used to ignore common warnings
+        if (!(jaiDef.getImagingListener() instanceof GeoServerImagingListener)) {
+            jaiDef.setImagingListener(new GeoServerImagingListener());
+        }
+    }
+
```
jodygarnett, line 200:
```suggestion
   /**
    * Custom GeoTools ImagingListener used to ignore common warnings.
    */
```

jodygarnett (approved):

src/main/src/main/java/org/geoserver/GeoserverInitStartupListener.java
```diff
@@ -220,6 +189,49 @@ private boolean isSerializableRenderedImageFinalization(Object where, Throwable
...
+    private static final class GeoServerImagingListener implements ImagingListener {
+        private static final Logger LOGGER = Logging.getLogger("org.geoserver.jai");
```
jodygarnett, line 202: I am not sure how I feel about changing the logger in this manner, if you wish to do so please note the change in the update instrutions also.

aside: Looking ahead to GeoServer 3 we will be changing to ImageN shortly, can you think of a different logger that would not need to get renamed?

sikeoka: I'll just drop the logger change. The main focus of the PR should just be to throw the exceptions.

jodygarnett:
Do you need a back port?

sikeoka:
Yes

PR #8368, opened 2025-03-05
GEOS-11726: slow post requests: caching improvements

NielsCharlier:
I have made three separate changes (one commit for each) to address the issue(s).

(1) As by @groldan's suggestion, replace the wfs XML parser eager featuretype cache with a lazy one.

(2) Introduce a cache for security filters in the security system

(3) Improve the jdbcconfig caching for layer/resource links (which we may assume are one-to-one)

While either (1) or (2) by itself resolves this specific issue with WFS post requests; and (3) goes a long way, I decided to submit all three fixes because they will  each improve performance elsewhere.

[GEOS-11726](https://osgeo-org.atlassian.net/browse/GEOS-11726) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
+++ b/src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
@@ -16,6 +16,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -335,6 +336,7 @@ public class DefaultResourceAccessManager implements ResourceAccessManager {
         if (lastLoaded < daoLastModified || force) {
             root = buildAuthorizationTree(dao);
             lastLoaded = daoLastModified;
+            // The filter cache must be invalidated, since the security rules have changed!!!
             filterCache.invalidateAll();
         }
     }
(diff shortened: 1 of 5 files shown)


NielsCharlier:
Note on the removal of the unit test "testPostWithMisconfiguredTypeAndExceptionReport":

This test no longer makes any sense. The test performs a GetFeature on another layer then the misconfigured one. There is no reason any longer to fail on this test, since there is no need any more to load all feature types for each POST request.Applying the GetFeature to the misconfigured test will always result in failure, even before the feature type is requested, because it fails to get the data.

src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
```diff
@@ -493,12 +506,26 @@ private String[] getLayerGroupPath(LayerGroupInfo layerGroup) {
...
+            return filter;
+        }
+    }
+
+    public Filter getSecurityFilterInternal(Authentication user, Class<? extends CatalogInfo> clazz) {
         if (getMode() == CatalogMode.CHALLENGE) {
             // If we're in CHALLENGE mode, we cannot pre-filter
             // for the other types we have no clue, use the in memory filtering
-            return InMemorySecurityFilter.buildUserAccessFilter(this, user);
+            return null;
```
aaime, line 527: The change makes the comment above harder to understand... are we actually moving to "in memory filtering" here. Not really, it's returning null because in the case of "challenge" mode we don't want to cache the filter, right? But I'm not sure why.... just because this filter is not expensive to build?

aaime (changes requested):
Had a look, the change to the WFS test makes sense, the other changes looked  weird  at a first sight, I recommend documenting them better.

Also @groldan could you have a look at this one too? (since it's related to changes you made in the past?)

src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
```diff
@@ -493,12 +506,26 @@ private String[] getLayerGroupPath(LayerGroupInfo layerGroup) {
 
     @Override
     public Filter getSecurityFilter(Authentication user, Class<? extends CatalogInfo> clazz) {
+        Filter filter = filterCache.getIfPresent(Pair.of(user, clazz));
+        if (filter == null) {
+            filter = getSecurityFilterInternal(user, clazz);
+            if (filter != null) {
+                filterCache.put(Pair.of(user, clazz), filter);
+            }
+        }
```
groldan, line 515: Please use `V get(K key, Callable<? extends V> loader)` instead, it'll handle concurrency for you.
e.g.
```java
Filter filter = filterCache.get(Pair.of(user, clazz)), ()-> getSecurityFilterInternal(user, clazz))
```

src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
groldan: I honestly find it confusing. IMO it calls for a couple test cases too, besides improving the comment for accuracy

src/main/src/test/java/org/geoserver/catalog/impl/CatalogImplTest.java
```diff
@@ -3564,6 +3564,7 @@ public void testExclusiveExactTermSearch() {
         cv.setName("waterways");
 
         addLayer();
+        catalog.add(cs);
         catalog.add(cv);
```
groldan, line 3568: If I'm not mistaken, this is ok but the prescribed way of doing so would be `addCoverage()` which already adds bod cs and cv.

groldan (changes requested):

src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
NielsCharlier: This will fail if getSecurityFilterInternal returns null. I didn't want to store post filters in the cache, that is why I create them only if getSecurityFilterInternal returns null

src/main/src/test/java/org/geoserver/catalog/impl/CatalogImplTest.java
NielsCharlier: That will fail if you have already called addLayer(). That is why they didn't do it when they added these new testing methods. But they forgot this one line, so they created a layer without store. Jdbcconfig fails on that.

But this change shouldn't be here, that was an accident. It needs to be in a separate PR, because it has a different purpose, it's to make those new exact term tests work in jdbcconfig

src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
```diff
@@ -493,12 +506,26 @@ private String[] getLayerGroupPath(LayerGroupInfo layerGroup) {
...
+        if (filter == null) {
+            return InMemorySecurityFilter.buildUserAccessFilter(this, user);
+        } else {
+            return filter;
+        }
+    }
+
+    public Filter getSecurityFilterInternal(Authentication user, Class<? extends CatalogInfo> clazz) {
```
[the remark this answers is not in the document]
NielsCharlier: indeed no

src/main/src/main/java/org/geoserver/security/impl/DefaultResourceAccessManager.java
NielsCharlier: see my other comment, I wanted to avoid putting the InMemorySecurityFilter in the cache. I will make improvements

groldan (approved):

PR #8374, opened 2025-03-07
[GEOS-11750] Community module for Micrometer metrics.

axl8713:
[GEOS-11750](https://osgeo-org.atlassian.net/browse/GEOS-11750) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

### GeoServer Monitor Micrometer community module

This module provides a specialized visualization of requests made to a GeoServer instance, integrating  a Micrometer registry to expose a Prometheus-compatible metrics endpoint. 
This will allow Prometheus instances to scrape and collect usage metrics for GeoServer, enabling better insight into request patterns and system performance.

  

# Checklist


doc/en/user/source/community/monitor-micrometer/installation.rst
```diff
@@ -0,0 +1,13 @@
+.. _monitor_micrometer_installation:
+
+Installing the Monitor Micrometer Extension
+===========================================
+
+As a community module, the package needs to be downloaded from the `nightly builds <https://build.geoserver.org/geoserver/>`_,
```
mprins, line 6: or built from source code

src/community/monitor-micrometer/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.geoserver.extension</groupId>
+      <artifactId>gs-monitor-core</artifactId>
+      <version>${project.version}</version>
+      <scope>compile</scope>
```
mprins, line 32: since you stated in the docs the monitor extension must be installed; I think this should be a `runtime` scope (in which case the scope can be omitted, that being the default)

doc/en/user/source/community/monitor-micrometer/installation.rst
axl8713: Instruction updated. Thanks!

src/community/monitor-micrometer/pom.xml
axl8713: Thanks for the suggestion! 
Since `compile` is the default scope in Maven, I can simply omit it. 
However, I cannot use `runtime` because the code has direct imports from `gs-monitor-core`, which are needed at compile-time. 
Updated the PR accordingly.

mprins (approved):
lgtm

src/community/pom.xml
```diff
@@ -447,6 +448,12 @@
         <module>monitor-kafka</module>
       </modules>
     </profile>
+    <profile>
+      <id>monitor-micrometer</id>
```
jodygarnett, line 452:
```suggestion
      <id>monitor-micrometer</id>
      <activation>
        <property>
          <name>monitoring</name>
        </property>
      </activation>
```

src/web/app/pom.xml
```diff
@@ -634,6 +634,21 @@
         </dependency>
       </dependencies>
     </profile>
+    <profile>
+      <id>monitor-micrometer</id>
+      <activation>
+        <property>
+          <name>monitoring</name>
+        </property>
+      </activation>
```
jodygarnett, line 643: See above - If you are going to have an activation property here, you should have it in community/pom.xml also :)

doc/en/user/source/community/monitor-micrometer/installation.rst
```diff
@@ -0,0 +1,13 @@
...
+===========================================
+
+As a community module, the package needs to be built from source code or downloaded from the `nightly builds <https://build.geoserver.org/geoserver/>`_,
+picking the community folder of the corresponding GeoServer series (e.g. if working on GeoServer main development branch nightly
+builds, pick the zip file form ``main/community-latest``).
+
+To install the module, unpack the zip file contents into GeoServer own ``WEB-INF/lib`` directory and
+restart GeoServer.
+
+For the module to work, the :ref:`monitor_extension` extensions must also be installed.
```
jodygarnett, line 13: 1. Requested change: We have an external reference ``download_community`` defined for community modules so you can provide a direct URL download.

   ```
   To install the JDBCConfig module:

   #. Visit the :website:`website download <download>` page and download :download_community:`jdbcconfig`.

   #. Extract this file and place the JARs in ``WEB-INF/lib``.
   ```
   
   This is very handy as it provides the person with the exact name to look for filled in with the correct version number.
       
   See example: https://docs.geoserver.org/latest/en/user/community/jdbcconfig/installing.html
   
   
2: Suggestion: I have been experimenting with offering docker instructions alongside module testing.

https://docs.geoserver.org/latest/en/user/services/features/install.html

jodygarnett (changes requested):

src/web/app/pom.xml
axl8713: Thanks for the comment.
None of the monitoring extensions have an activation property in the `community/pom.xml`. Would you like me to add it specifically for this extension?

aaime: The monitor extension is supported, so you'll find its activation property under extensions/pom.xml
The monitor-hibernate and monitor-kafka both have profiles in community/pom.xml:

```
    <profile>
      <id>monitor-hibernate</id>
      <modules>
        <module>monitor-hibernate</module>
      </modules>
    </profile>
    <profile>
      <id>monitor-kafka</id>
      <modules>
        <module>monitor-kafka</module>
      </modules>
    </profile>
```

aaime: Ok I see now... the monitoring property is really a bad idea, when used it brings in all sorts of monitoring storages, but only one should be included instead. Better to remove the property altogether and just rely on direct profile activation with -P instead.

doc/en/user/source/community/monitor-micrometer/installation.rst
axl8713: > Requested change: We have an external reference download_community defined for community modules so you can provide a direct URL download.

Thanks! I've revised the installation instruction to add this.

> 2: Suggestion: I have been experimenting with offering docker instructions alongside module testing.

Thanks for the suggestion! That sounds like a great approach. However, since the module is not yet in the build, I’m unable to test it at this stage. Do you have any suggestions on how to approach this without it being part of the build yet?

src/web/app/pom.xml
axl8713: Removed the activation properties to avoid load all the extensions at once.
Thanks for the comments.

doc/en/user/source/community/monitor-micrometer/installation.rst
jodygarnett: No good suggestion (like you could build the docker image locally, there is a way it picks up locally provided community modules).

You can either add the section after the PR is done (low risk and doc changes can be done with out a Jira). Or write it now and test with the understand you may need to fix it based on public feedback and testing.

Cheers!

axl8713: Thanks for the clarification! Since I can update the docs later without a Jira, I’ll proceed with the PR as-is and revisit this section once the module is in the build and testable. Appreciate the guidance!

jodygarnett:
All feedback addressed, merging! @axl8713 just to confirm you have ability to commit and take care of this module (or do I need to update your permissions).

PR #8376, opened 2025-03-07
[GEOS-11760] Fix a potential OOM in the KML transformation

sikeoka:
[GEOS-11760](https://osgeo-org.atlassian.net/browse/GEOS-11760) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This PR replaces using the old XSLT transformation to fix invalid KML to using a custom SAX ContentHandler that programmatically filters out the invalid KML elements because the XSLT requires loading the entire document into memory which can cause OutOfMemoryErrors.
  

# Checklist


sikeoka:
@aaime, if you're interested in fixing the OOM error you mentioned in https://osgeo-org.atlassian.net/browse/GEOS-7963 and can confirm that this fix works, I can clean up the code and create a JIRA ticket. I only tested the issue using the unit test in this PR and not in a running GeoServer and the test is ignored because it can take five minutes to run.

aaime:
I've tested it as it is with a very large shapefile layer (the .shp alone is 1.3GB)... it's behaving in a strange way:

- Nothing comes out in output, which made me think it's accumulating stuff in memory
- However, after a long wait (some 40 seconds), no OOM occurs, and it starts streaming out KML and success in generating tens of GB of KML in output. 
- For reference, the WFS output does not need to be adjusted and it streams out immediately.

I've checked, the initial wait is due to first pass over the data due to LookAtDecorator calling ReprojectFeatureResults.getBounds. So not related to the change. Looks good IMHO.

By the way, last week I actually tested out a JAXB upgrade to JakartaEE. As far as I can see, it does not depend on other JakartaEE libraries and it could be done now, but the changes are very much spread out, I could not figure out a migration in RESTConfig and eventually run out of spare time. For reference, here are the branches:
* [ImageIO-EXT](https://github.com/aaime/imageio-ext/tree/jaxb_jakarta) (also requires the java version to be upgraded from 8 to 11)
* [GeoTools](https://github.com/aaime/geotools/tree/jaxb-jakarta)
* [GeoServer](https://github.com/aaime/geoserver/tree/kml_jakarta) (this one is not working, some issue in JAXB usage in restconfig, run out of time before I could fix it).

aaime:
@sikeoka you good with the plan?
- Merge this one now
- Complete the upgrade to Jakarta EE when main becomes 2.28.x?

sikeoka:
I created a JIRA ticket and updated the PR.

Besides restconfig, the MapML extension also depends on a Spring class that is using the old JAXB API until Spring 6.
https://github.com/geoserver/geoserver/blob/main/src/extension/mapml/src/main/resources/applicationContext.xml#L49

PR #8382, opened 2025-03-08
[GEOS-11755] AbstractCatalogFacade leaves dangling references to temporary Catalog

groldan:
[GEOS-11755](https://osgeo-org.atlassian.net/browse/GEOS-11755) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

The data directory loading process involves loading all catalog objects to a temporary catalog and then synchronizing the real catalog with the temporary one.

In doing so, AbstractCatalogFacade.resolve() does not replace the catalog property of `StyleInfoImpl`, `ResourceInfoImpl`, and `StoreInfoImpl`, with the real catalog reference, hence leaving all those objects with dangling references to the temporary Catalog instance. With the default implementation, things keep on working by accident due to `CatalogImpl.sync(Catalog)` taking ownership over the temporary catalog’s internals.

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/impl/AbstractCatalogFacade.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/AbstractCatalogFacade.java
@@ -160,14 +160,16 @@ public abstract class AbstractCatalogFacade implements CatalogFacade {
 
     protected void resolve(StyleInfo style) {
         setId(style);
+        StyleInfoImpl s = (StyleInfoImpl) style;
+        s.setCatalog(getCatalog());
 
         // resolve the workspace
-        WorkspaceInfo ws = style.getWorkspace();
+        WorkspaceInfo ws = s.getWorkspace();
         if (ws != null) {
             WorkspaceInfo resolved = ResolvingProxy.resolve(getCatalog(), ws);
             if (resolved != null) {
                 resolved = unwrap(resolved);
-                style.setWorkspace(resolved);
+                s.setWorkspace(resolved);
             } else {
                 LOGGER.log(
                         Level.INFO,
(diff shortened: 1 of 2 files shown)


src/main/src/main/java/org/geoserver/catalog/impl/AbstractCatalogFacade.java
```diff
@@ -160,14 +160,16 @@ private void resolveLayerGroupLayers(List<PublishedInfo> layers) {
 
     protected void resolve(StyleInfo style) {
         setId(style);
+        StyleInfoImpl s = (StyleInfoImpl) style;
+        s.setCatalog(getCatalog());
```
aaime, line 164: Makes some sense but I'm wondering... the catalog field is not in the StyleInfo interface, and I don't see anything using it. So two questions:

* How is this causing a problem?
* Should it be just removed instead?

aaime (commented):
Makes sense for resource info and catalog info, but the style case seems odd, can you double check? See my comment below.

src/main/src/main/java/org/geoserver/catalog/impl/AbstractCatalogFacade.java
groldan: the catalog field is not in the StyleInfo interface, but it's required by `getSLD()` and `getStyle()`. Using a dangling reference to a defuncted Catalog instance is a problem because it's working by accident as explained in the description. I don't know what problems could happen when calling `getSLD()` or `getStyle()` on the temporary catalog that survived the load process but shares the ResourcePool and the CatalogInfoLookUps with the final/real Catalog but a different CatalogFacade, but in any case I think it'd be better to get rid of this inconsistency.

groldan: Oh by the way IIRC there was a functional defect caused by this when the datadir-catalog-loader created a second temporary copy of the catalog and transferred the contents to the final catalog, but now I've changed it to follow the same workflow as DefaultGeoServerLoader.

PR #8384, opened 2025-03-09
[GEOS-11757] Optimize ConfigurationPasswordEncryptionHelper to cache encrypted fields by store type

groldan:
[GEOS-11757](https://osgeo-org.atlassian.net/browse/GEOS-11757) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

The `ConfigurationPasswordEncryptionHelper` class has been updated to improve performance by caching encrypted fields based on the StoreInfo type when the class cache is hit.

Previously, when a DataStoreInfo object lacked a type property, the method would repeatedly invoke `getCatalog().getResourcePool().getDataStoreFactory(info)`, causing unnecessary factory lookups and expensive serialization/deserialization operations within `ResourcePool`.

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/password/ConfigurationPasswordEncryptionHelper.java
+++ b/src/main/src/main/java/org/geoserver/security/password/ConfigurationPasswordEncryptionHelper.java
@@ -120,6 +120,9 @@ public class ConfigurationPasswordEncryptionHelper {
 
         toEncrypt = CACHE.get(factory.getClass());
         if (toEncrypt != null) {
+            if (storeType != null && null == STORE_INFO_TYPE_CACHE.putIfAbsent(storeType, toEncrypt)) {
+                LOGGER.fine("Cached encryption fields for storeType " + storeType);
+            }
             return toEncrypt;
         }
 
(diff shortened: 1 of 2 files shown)


src/main/src/main/java/org/geoserver/security/password/ConfigurationPasswordEncryptionHelper.java
```diff
@@ -129,7 +132,7 @@ public Set<String> getEncryptedFields(StoreInfo info) {
         }
 
         toEncrypt = Collections.emptySet();
-        if (info != null && info.getConnectionParameters() != null) {
+        if (info.getConnectionParameters() != null) {
```
groldan, line 135: info can never be null here, it's being

src/main/src/main/java/org/geoserver/security/password/ConfigurationPasswordEncryptionHelper.java
```diff
@@ -170,10 +173,9 @@ public String encode(String value) {
 
     /** Decrypts previously encrypted store connection parameters. */
     public void decode(StoreInfo info) {
-        List<GeoServerPasswordEncoder> encoders = securityManager.loadPasswordEncoders(null, true, null);
-
         Set<String> encryptedFields = getEncryptedFields(info);
-        if (info.getConnectionParameters() != null) {
+        if (!encryptedFields.isEmpty() && info.getConnectionParameters() != null) {
```
groldan, line 177: `securityManager.loadPasswordEncoders()` comes with an overhead and is not necessary to call it if there are no encryptable fields to check

aaime:
Doh, tricky, but makes sense, merging.

PR #8406, opened 2025-03-15
Allow to forward options to jetty using the JETTY_OPTS env var

rolfschr:
This PR allows to modify the jetty configuration via the environment variable `JETTY_OPTS`. This eases configuration testing. My main motivation here is to be able to more easily use geoserver on NixOS where modifying files in `GEOSERVER_HOME` is usually not possible (if installed as a package).

I did not test the `startup.bat` file as I do not have Windows at hand. I changed that file for consistency. I do not now how batch files behave if an env var is not set (in Linux, I refer to `${JETTY_OPTS}` vs `${JETTY_OPTS:-}`).

I did not find any test involving an invocation of `startup.sh` and thus refrained changing any test. I tested with `JETTY_OPTS=jetty.http.port=1234 path/to/startup.sh`

# Checklist

   - Please let me know if this is necessary

### DIFF
--- a/doc/en/user/source/installation/linux.rst
+++ b/doc/en/user/source/installation/linux.rst
@@ -41,6 +41,13 @@ Installation
       echo "export GEOSERVER_HOME=/usr/share/geoserver" >> ~/.profile
       . ~/.profile
 
+#. Optionally, set the environment variable ``JETTY_OPTS`` to tweak the jetty configuration upfront:
+
+   .. code-block:: bash
+
+      echo "export JETTY_OPTS='jetty.http.port=1234'" >> ~/.profile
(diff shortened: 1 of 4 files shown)


aaime:
Looks legit to me, but it should be documented in these pages, otherwise users won't know about this possibility:
* https://github.com/geoserver/geoserver/blob/main/doc/en/user/source/installation/linux.rst
* https://github.com/geoserver/geoserver/blob/main/doc/en/user/source/installation/win_binary.rst

aaime:
Lost track of this one! Merged now.

rolfschr:
Cool, thanks!

PR #8409, opened 2025-03-15
QA: Add type parameters and @SuppressWarnings to reduce Java compiler warnings in core modules

groldan:
The current commit is adding appropriate Java generics type parameters and @SuppressWarnings annotations to reduce compiler warnings across multiple core modules. The changes are strictly related to code quality improvement and don't change any functionality.

  

# Checklist

### DIFF
--- a/src/gwc/src/test/java/org/geoserver/gwc/GWCSeedingSecurityIntegrationTest.java
+++ b/src/gwc/src/test/java/org/geoserver/gwc/GWCSeedingSecurityIntegrationTest.java
@@ -120,7 +120,7 @@ public class GWCSeedingSecurityIntegrationTest extends GeoServerSystemTestSuppor
         assertNull("Cache should be empty at test start", tileObject.getBlob());
 
         SeedRequest sr = createSeedRequest(PUB_LAKES);
-        @SuppressWarnings("PMD.UnusedLocalVariable")
+        @SuppressWarnings({"PMD.UnusedLocalVariable", "unused"})
         GWCTask[] unUsedTasks = executeSeedRequest(sr);
 
         // verify the tasks have been queued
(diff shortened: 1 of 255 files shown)


aaime:
If possible, I'd make it part of the QA profile and enforce it, otherwise it's going to bit rot again

src/gwc/src/test/java/org/geoserver/gwc/GWCSeedingSecurityIntegrationTest.java
```diff
@@ -120,7 +120,7 @@ public void testSeedPublic() throws GeoWebCacheException, InterruptedException,
         assertNull("Cache should be empty at test start", tileObject.getBlob());
 
         SeedRequest sr = createSeedRequest(PUB_LAKES);
-        @SuppressWarnings("PMD.UnusedLocalVariable")
+        @SuppressWarnings({"PMD.UnusedLocalVariable", "unused"})
```
aaime, line 123: I don't see an "unused" warning in the Java compiler lint settings... is this Eclipse specific?

src/main/src/main/java/org/geoserver/catalog/impl/LocalWorkspaceCatalog.java
```diff
@@ -532,11 +533,13 @@ public static <T> T create(T object, Class<T> clazz) {
 
         public static <T> List<T> createList(List<T> object, Class<T> clazz) {
             return new ProxyList<>(object, clazz) {
+                @SuppressWarnings("hiding")
```
aaime, line 536: This also does not seem part of the javac lint warnings

src/main/src/main/java/org/geoserver/catalog/impl/MapInfoImpl.java
```diff
@@ -11,6 +11,7 @@
 import org.geoserver.catalog.LayerInfo;
 import org.geoserver.catalog.MapInfo;
 
+@SuppressWarnings("serial")
```
aaime, line 14: I would tell the IDE to ignore serial version UID warnings... none of this classes is meant to be serialized and re-used long term, and very often we change the structure of the class without changing the serialVersionUID anyways, which would break its intended usage anyways.

PR #8415, opened 2025-03-17
Refactor catalog proxy resolution into dedicated utility class

groldan:
Extract proxy resolution logic from AbstractCatalogFacade into ResolvingProxyResolver to centralize this functionality and make it reusable across the codebase. This change:

- Creates a new `ResolvingProxyResolver` utility with comprehensive documentation
- `ResolvingProxyResolver` is only dedicated to resolve proxies, while `AbstractCatalogFacade`'s `resolve()` methods call `setId()` and `ResolvingProxyResolver.resolve()`.
- Avoids code duplication by moving proxy resolution to an utility class
- Simplifies `AbstractCatalogFacade` by delegating to the specialized resolver
- Adds test suite for `ResolvingProxyResolver` with 97.3% coverage

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/impl/ResolvingProxyResolver.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/ResolvingProxyResolver.java
@@ -1,0 +1,283 @@
+/* (c) 2025 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.catalog.impl;
+
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import org.geoserver.catalog.Catalog;
+import org.geoserver.catalog.CatalogInfo;
+import org.geoserver.catalog.LayerGroupInfo;
+import org.geoserver.catalog.LayerInfo;
+import org.geoserver.catalog.MapInfo;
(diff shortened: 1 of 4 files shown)


src/main/src/main/java/org/geoserver/catalog/impl/ResolvingProxyResolver.java
```diff
@@ -0,0 +1,284 @@
...
+     *
+     * <p>This helper method delegates to {@link ModificationProxy#unwrap(Object)} to ensure we're working with the
+     * actual object instances rather than their proxy wrappers.
+     *
+     * @param <T> the type of the object being unwrapped
+     * @param obj the potentially proxied object
+     * @return the unwrapped object, or the original object if not a proxy
+     */
+    private static <T> T unwrap(T obj) {
+        return ModificationProxy.unwrap(obj);
```
jodygarnett, line 70: not sure I get the value of having this method here rather than using static method ModificationProxy.unwrap(obj) directly?

jodygarnett:
Does this need a Jira? Not against the code clean up ...

src/main/src/main/java/org/geoserver/catalog/impl/ResolvingProxyResolver.java
groldan: It's only a refactoring of what was in AbstractCatalogFacade, no judgements.

groldan:
Closing as the merge of #8370 includes this

PR #8420, opened 2025-03-18
JDBCConfig: Fix failing Catalog tests

NielsCharlier:
Three recently added catalog tests are failing for jdbcconfig

* Two "exact term" tests actually had errors in the order of saving items to the catalog, that are forgiven by the default catalog since everything lives in memory.
* The sync test failed because the jdbccatalogfacade sync method was apparently faulty. Because it is quite similar to the "manual" sync of the defaultcatalogfacade I made a new method in AbstractCatalogFacade to avoid code duplication.

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/impl/AbstractCatalogFacade.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/AbstractCatalogFacade.java
@@ -5,13 +5,21 @@
  */
 package org.geoserver.catalog.impl;
 
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
 import java.lang.reflect.Proxy;
 import java.rmi.server.UID;
 import java.util.List;
+import org.geoserver.catalog.Catalog;
 import org.geoserver.catalog.CatalogFacade;
 import org.geoserver.catalog.CatalogInfo;
+import org.geoserver.catalog.DataStoreInfo;
 import org.geoserver.catalog.LayerGroupInfo;
 import org.geoserver.catalog.LayerInfo;
+import org.geoserver.catalog.LockingCatalogFacade;
 import org.geoserver.catalog.MapInfo;
 import org.geoserver.catalog.NamespaceInfo;
 import org.geoserver.catalog.ResourceInfo;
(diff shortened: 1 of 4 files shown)


aaime:
Tests are passing but there is a conflict to be resolved

src/main/src/test/java/org/geoserver/catalog/impl/CatalogImplTest.java
```diff
@@ -3434,11 +3434,12 @@ public void testExactTermSearch() {
...
         addLayer();
-        catalog.add(cs);
-        catalog.add(cv);
 
         LayerInfo l2 = newLayer(cv, s);
         l2.setTitle("Global");
+
+        catalog.add(cs);
+        catalog.add(cv);
         catalog.add(l2);
```
groldan, line 3443: I think this is what you mean by *Two "exact term" tests actually had errors in the order of saving items to the catalog, that are forgiven by the default catalog since everything lives in memory.*

But I see no error in there:
- `addLayer()` implies adding the feature type and the datastore, so it doesn't affect that follow from there
- `add(cs)` and `add(cv)` add the coverage store and the coverage itself, which `l2` is built for
- `add(l2)` adds the layer once its resource is added

Can you explain what the issue is?

> The one thing I know is that all those tests depend on an implementation detail: that `Catalog.add(CataogInfo)` sets the id and commits the `ModificationProxy`. So sometimes on other implementations you need to do a `Catalog.getXXX(id)` in order to assign the just added object as a reference. It'd make more sense if `void Catalog.add(xxx)` returned the "added" object, hence ensuring any internal consistency it requires (for example, it may return a different instance than the argument one, if the store is based on an RDBMS that will automatically update some properties like `lastUpdated`), but most importantly I guess is once you create the layer out of the coverage/store, you're not using a property managed instance for the coverage/store. In the most common case it just means it won't be a `ModificationProxy`.

In any case, this has always been like that and changing it would require a proposal.

groldan (changes requested):
doing `facade.add(deserialize(serialize(info))` is non-sense to me. Please read the comments below.

src/main/src/main/java/org/geoserver/catalog/impl/AbstractCatalogFacade.java
```diff
@@ -105,4 +113,88 @@ protected void setId(Object o) {
...
+            target.add(syncInfo(l, deep, null));
+        }
+
+        for (LayerGroupInfo lg : source.getLayerGroups()) {
+            target.add(syncInfo(lg, deep, null));
+        }
+
+        for (MapInfo m : source.getMaps()) {
+            target.add(syncInfo(m, deep, null));
+        }
```
groldan, line 181: This is unacceptable IMHO
- I don't think a community module should have this big of an impact in core without discussion
- This is replacing `facade.add(info)` by `facade.add(deserialize(serialize(info)))`, which is going to be catastrophic in terms of performance
- It seems to me there might be some implementation detail in `jdbconfig` that might require this, or for which this would be a work-around, but definitely can't go in core.
- The rationale is not clear nor explained

groldan (changes requested):

src/main/src/main/java/org/geoserver/catalog/impl/AbstractCatalogFacade.java
NielsCharlier: @groldan

1. there is no impact on core behaviour, because the serialize/deserialize only happens if deep=true, but the default catalogue passes deep=false, so that the original behaviour is entirely preserved.

2. Regarding the rationale, the only thing that is a work-around to me is that I have to call serialize/deserialize for a deep copy because no other cloning method seems to work. I think it is inherently wrong that the catalog info object are reference copied instead of deeply copied. From one catalog to another. The result is that two catalog refer to the same object, while the object has a reference back to only one of the catalogs. That IMHO is dangerous and badly designed. Nonetheless , I didn't touch this, I only avoided code duplication.

src/main/src/test/java/org/geoserver/catalog/impl/CatalogImplTest.java
NielsCharlier: l2.setTitle("Global") does not modify the layer, it modifies the resource, because title is a property of resource, not layer. The method directly modifies the linked resource, and leaves the layer unchanged. The resource needs to be saved into the catalog to ensure the change is permanent. This is so by design.

src/main/src/test/java/org/geoserver/catalog/impl/CatalogImplTest.java
```diff
@@ -3567,6 +3568,7 @@ public void testExclusiveExactTermSearch() {
         cv.setName("waterways");
 
         addLayer();
+        catalog.add(cs);
```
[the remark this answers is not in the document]
NielsCharlier: The store was nowhere being added to the catalogue.

src/main/src/test/java/org/geoserver/catalog/impl/CatalogImplTest.java
NielsCharlier: Note that these two issues in the "exact term" tests are in fact errors, regardless of jdbcconfig. If data was entered like this in a live running geoserver instance with default catalog, resource title and store would be gone after a geoserver restart / reload. Jdbcconfig only servers as the canary in the coal mine here.

groldan (approved):
Hi @NielsCharlier,
you're right, I missed the `deep` parameter, shouldn't do code reviews at 1:30am

Regarding
> Regarding the rationale, the only thing that is a work-around to me is that I have to call serialize/deserialize for a deep copy because no other cloning method seems to work. I think it is inherently wrong that the catalog info object are reference copied instead of deeply copied from one catalog to another. The result is that two catalog refer to the same object, while the object has a reference back to only one of the catalogs. That IMHO is dangerous and an anti-pattern. There is in fact now an undocumented assumption that the source catalog will be immediately discarded, since it has become inconsistent and defunct, returning objects that believe they are (only) part of another catalog.

That is also true and a historical defect of the default catalog facade and CatalogImpl.sync(). BUT, CatalogImpl.sync() is not API, it's always assumed the facade is a DefaultCatalogFacade, and the only purpose for sync() is to transfer the contents and discard once the catalog is loaded.

Conversely, `CatalogFacade.syncTo()` is API, for it's defined in the interface. And hence a can of worms if you're extending `DefaultCatalogFacade`, but there's nothing in the contract that says how objects are to be stored in the backend. I'm with you in that what DefaultCatalogFacade does is weird. At the same time CatalogFacade shouldn't be considered API like in nobody in its same mind should be calling `Catalog.getFacade().syncTo()`.

PR #8443, opened 2025-03-25
[GEOS-11788] Apply feedback from CSP testing

sikeoka:
[GEOS-11788](https://osgeo-org.atlassian.net/browse/GEOS-11788) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This PR adds a new field to the web interface to allow administrators to more easily configure the `form-action` directive, which seems to be source of most, if not all, CSP-related issues coming from GeoServer's CSP (and not Wicket's).

This PR also removes all of the `frame-ancestors` rules from the default CSP configuration to reduce confusion when administrators are troubleshooting CSP-related issues. The default behavior is moved into the Java code and is identical to the previous behavior.

A special keyword `HIDE` was added to allow administrators to remove the `form-action` or `frame-ancestors` from the CSP header. The `'self'` source is automatically added if a value is provided that does not contain `'self'` or `'none'` to prevent accidentally causing a denial of service. Also, the web interface will prevent setting a `form-action` value containing `'none'` since that will break the UI.

  

# Checklist


sikeoka:
This can wait until the next release.

groldan:
@sikeoka cool, I don't feel qualified to properly review this one, so you're positive we can go with 2.27.0 without it?

sikeoka:
I was under the impression that 2.27.0 was already built when I made that comment. If you guys are still merging PRs for 2.27.0, then it would be good for this to be in there too.

aaime:
Merging on main for the time being, don't feel comfortable having this in 2.27.1 right away (mostly because I still don't understand what's going on with all this CSP stuff).

PR #8446, opened 2025-03-26
Added note to clarify that the header value should match a username

KroniK907:
In going through the process of using the http header auth for the first time, it was very unclear in the documentation what the header value was supposed to match. After doing a bunch of testing it became clear that it needed to match the username of a user, but that was not included anywhere in the documentation. This update adds a note at the end that makes it very clear what each part of the header ``name: value`` pair is associated with in the system.
  

# Checklist


KroniK907:
Is there anything I can do to help move this forward?

PR #8455, opened 2025-03-29
[GEOS-11792] Fix welcome page listing all services via sevice caps

jodygarnett:
[GEOS-11792](https://osgeo-org.atlassian.net/browse/GEOS-11792) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This only occurred when no workspaces were available; but was still annoying. 

  

# Checklist

### DIFF
--- a/src/extension/csw/web-csw/src/main/java/org/geoserver/csw/web/CSWServiceDescriptionProvider.java
+++ b/src/extension/csw/web-csw/src/main/java/org/geoserver/csw/web/CSWServiceDescriptionProvider.java
@@ -32,6 +32,7 @@ public class CSWServiceDescriptionProvider extends ServiceDescriptionProvider {
     Catalog catalog;
 
     public CSWServiceDescriptionProvider(GeoServer gs) {
+        super(SERVICE_TYPE);
         this.geoserver = gs;
         catalog = gs.getCatalog();
     }
@@ -61,7 +62,7 @@ public class CSWServiceDescriptionProvider extends ServiceDescriptionProvider {
         CSWInfo info = info(workspaceInfo, layerInfo);
 
         if (workspaceInfo != null || geoserver.getGlobal().isGlobalServices()) {
-            descriptions.add(description(SERVICE_TYPE, info, workspaceInfo, layerInfo));
+            descriptions.add(description(serviceType, info, workspaceInfo, layerInfo));
         }
         return descriptions;
     }
@@ -78,16 +79,18 @@ public class CSWServiceDescriptionProvider extends ServiceDescriptionProvider {
 
         for (Service service : extensions) {
             if (service.getService() instanceof WebCatalogService) {
-                String link = null;
+                String link;
                 if (service.getOperations().contains("GetCapabilities")) {
                     link = getCapabilitiesURL(workspaceInfo, layerInfo, service);
                 } else if (service.getCustomCapabilitiesLink() != null) {
                     link = service.getCustomCapabilitiesLink();
+                } else {
+                    link = null;
                 }
 
                 if (link != null) {
                     links.add(new ServiceLinkDescription(
-                            SERVICE_TYPE,
+                            serviceType,
                             service.getVersion(),
                             link,
                             workspaceInfo != null ? workspaceInfo.getName() : null,
--- a/src/extension/ogcapi/web-ogcapi/src/test/java/org/geoserver/web/ogcapi/OgcApiServiceDescriptionProviderTest.java
+++ b/src/extension/ogcapi/web-ogcapi/src/test/java/org/geoserver/web/ogcapi/OgcApiServiceDescriptionProviderTest.java
@@ -7,6 +7,7 @@ package org.geoserver.web.ogcapi;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
+import java.util.Arrays;
 import org.geoserver.data.test.SystemTestData;
 import org.geoserver.platform.GeoServerExtensionsHelper;
 import org.geoserver.test.GeoServerSystemTestSupport;
@@ -28,6 +29,10 @@ public class OgcApiServiceDescriptionProviderTest extends GeoServerSystemTestSup
         assertEquals("TestCaseServiceType", provider.getServiceType());
         assertEquals("OGCAPI-TestCase", provider.getServiceName());
         assertEquals("TestCase", provider.getSpecificServiceType());
+        assertTrue(
+                "skip service capabilities",
+                provider.getServiceTypes()
+                        .containsAll(Arrays.asList(provider.getServiceType(), provider.getSpecificServiceType())));
 
         assertEquals(TestCaseInfo.class, provider.getInfoClass());
         assertEquals(TestCaseService.class, provider.getServiceClass());
--- a/src/web/core/src/main/java/org/geoserver/web/ServiceInfoCapabilitiesProvider.java
+++ b/src/web/core/src/main/java/org/geoserver/web/ServiceInfoCapabilitiesProvider.java
@@ -9,10 +9,10 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.logging.Logger;
 import org.apache.wicket.Component;
-import org.geoserver.catalog.WorkspaceInfo;
-import org.geoserver.config.GeoServer;
 import org.geoserver.config.ServiceInfo;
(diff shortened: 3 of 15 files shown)


src/web/core/src/main/java/org/geoserver/web/ServiceInfoCapabilitiesProvider.java
```diff
@@ -67,34 +79,13 @@ public Component getCapabilitiesComponent(final String id) {
...
-     * willing to participate. In this case we check each workspace to see what services are available for the
-     * application as a whole.
-     *
      * @return list of services to skip, in lowercase (for case-insensitive matching)
      */
     protected Set<String> skipServiceDescriptionProviders() {
-        GeoServer geoServer = GeoServerApplication.get().getGeoServer();
-
         Set<String> skip = new HashSet<>();
-        List<WorkspaceInfo> workspaceList = new ArrayList<>();
```
aaime, line 80: Whoa, what happened here... I see the service descriptor provider had the service types buried in the ServiceDescriptor and and ServiceLinkDescription it was returning, and you basically exposed them in a new method, instead of having them be hunted down piece by piece.

However how does that relate to the exception you were fixing here? I guess it was happening in the code that got removed. If so, I'd have a test for ServiceInfoCapabilitiesProvider replicating the context of the exception and making sure it's not happening any more.
The kind of test that's easier to write against the un-modified code first.

src/extension/ogcapi/web-ogcapi/src/test/java/org/geoserver/web/ogcapi/OgcApiServiceDescriptionProviderTest.java
```diff
@@ -28,6 +29,10 @@ public void testBasic() {
         assertEquals("TestCaseServiceType", provider.getServiceType());
         assertEquals("OGCAPI-TestCase", provider.getServiceName());
         assertEquals("TestCase", provider.getSpecificServiceType());
+        assertTrue(
```
aaime, line 32: I'm confused as to what this test is checking. What's the point of having a list with the same value twice?
And how does it relate to the issue reported?
A test for the Pr should replicate the failure in the home page and check it's gone after the fix...

aaime: Also, shouldn't the test check for serviceType and specificServiceType?
It seems to be passing by accident.

src/extension/csw/web-csw/src/main/java/org/geoserver/csw/web/CSWServiceDescriptionProvider.java
```diff
@@ -36,6 +37,11 @@ public CSWServiceDescriptionProvider(GeoServer gs) {
         catalog = gs.getCatalog();
     }
 
+    @Override
+    public List<String> getServiceTypes() {
+        return Collections.singletonList(SERVICE_TYPE);
```
aaime, line 42: Two bits that I don't consider in actual need of fixing, but give me pause so I'll share a comment anyways:

* I see this code repeated over and over... feels like this should be a field of the class, provided via constructor, and then have a getter in the base class
* Using ``List.of(SERVICE_TYPE)`` would be the Java11+ way to achieve the same goad (all types of immutable lists have basically been replaced by ``<MyCollectionType>.of(xyz)``

aaime (changes requested):
Overall I think the change is good, but testing is not actually checking the issue being fixed, that needs to be addressed before merging.

jodygarnett:
Fair enough, I was testing I can get a skip list without a workspace. With a new method that does not take a workspace parameter I tested the hardest example.

I am in transit, may be in a position to write a specific test case today or Tuesday.

Alternative: We could remove the service capabilities section completely (as it was made available as a fall back for dispatcher services that did not have a service descriptor provider).

src/extension/csw/web-csw/src/main/java/org/geoserver/csw/web/CSWServiceDescriptionProvider.java
jodygarnett: thanks for feedback, resolved.

src/extension/ogcapi/web-ogcapi/src/test/java/org/geoserver/web/ogcapi/OgcApiServiceDescriptionProviderTest.java
jodygarnett: Yes test was not correct, added an explicit "FakeServiceDescriptorProvider" to use with CapabilitiesHomePagePanelTest to ensure the skipped service capabilities are not displayed.

landryb:
fwiw, this PR caused test issues in our [downstream project](georchestra/geoserver#4452) because we add bean definitions that apparently conflict with the constructor for `FakeServiceDescriptionProvider`
```
2025-04-07T06:30:36.9936827Z [ERROR] testNodeInfoInvisible(org.geoserver.web.CustomGeoServerNodeIdTest)  Time elapsed: 0.616 s  <<< ERROR!
2025-04-07T06:30:36.9940519Z org.apache.wicket.WicketRuntimeException: Can't instantiate page using constructor 'public org.geoserver.web.GeoServerHomePage()'. An exception has been thrown during construction!
2025-04-07T06:30:36.9942177Z 	at org.geoserver.web.CustomGeoServerNodeIdTest.testNodeInfoInvisible(CustomGeoServerNodeIdTest.java:34)
2025-04-07T06:30:36.9943373Z Caused by: java.lang.reflect.InvocationTargetException
2025-04-07T06:30:36.9944273Z 	at org.geoserver.web.CustomGeoServerNodeIdTest.testNodeInfoInvisible(CustomGeoServerNodeIdTest.java:34)
2025-04-07T06:30:36.9950298Z Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'org.geoserver.web.FakeServiceDescriptionProvider#0' defined in URL [file:/home/runner/work/georchestra/georchestra/geoserver/geoserver-submodule/src/web/core/target/test-classes/applicationContext.xml]: Unsatisfied dependency expressed through constructor parameter 0; nested exception is org.springframework.beans.factory.NoUniqueBeanDefinitionException: No qualifying bean of type 'java.lang.String' available: expected single matching bean but found 8: webDispatcherMappingSystem,georchestraHeaderHeight,georchestraHeaderUrl,georchestraHeaderScript,georchestraLegacyHeader,georchestraLogoUrl,georchestraStylesheet,georchestraHeaderConfigFile
2025-04-07T06:30:36.9956985Z 	at org.geoserver.web.CustomGeoServerNodeIdTest.testNodeInfoInvisible(CustomGeoServerNodeIdTest.java:34)
2025-04-07T06:30:36.9960668Z Caused by: org.springframework.beans.factory.NoUniqueBeanDefinitionException: No qualifying bean of type 'java.lang.String' available: expected single matching bean but found 8: webDispatcherMappingSystem,georchestraHeaderHeight,georchestraHeaderUrl,georchestraHeaderScript,georchestraLegacyHeader,georchestraLogoUrl,georchestraStylesheet,georchestraHeaderConfigFile
2025-04-07T06:30:36.9965032Z 	at org.geoserver.web.CustomGeoServerNodeIdTest.testNodeInfoInvisible(CustomGeoServerNodeIdTest.java:34)
``` 

which im locally fixing with georchestra/geoserver@ebbe027faa, @jodygarnett you might want to pick that, because if another bean of type `java.lang.String` is added someday tests will mysteriously fail.

landryb:
oh and this commit also seems to break the homepage in .. mysterious ways
```
Caused by: java.lang.NullPointerException: Cannot invoke "org.geoserver.config.ServiceInfo.getInternationalTitle()" because "info" is null
	at org.geoserver.web.ServiceDescriptionProvider.description(ServiceDescriptionProvider.java:141)
	at org.geoserver.wfs.web.WFSServiceDescriptionProvider.getServices(WFSServiceDescriptionProvider.java:79)
	at org.geoserver.web.GeoServerHomePage.homeInit(GeoServerHomePage.java:184)
	at org.geoserver.web.GeoServerHomePage.<init>(GeoServerHomePage.java:118)
	... 128 more
```
we had a working homepage before rebasing on the 2.27.0 release, so i'm suspecting this commit as it touched those codepaths...

any idea @jodygarnett ?

landryb:
i've filed https://osgeo-org.atlassian.net/browse/GEOS-11798 as a followup

PR #8459, opened 2025-04-01
[GEOS-11274,GEOS-11751] Generate URL for graphic inside workspace

vuilleumierc:
[GEOS-11274](https://osgeo-org.atlassian.net/browse/GEOS-11274) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

This PR proposes a couple of improvements in the JSON legend graphic builder for styles that are within a workspace.

[GEOS-11274](https://osgeo-org.atlassian.net/browse/GEOS-11274): In case of a single external graphic, use the workspace style directory to build the URL, since it is also published (similarly to the global style directory). This was causing a NPE ("iconUrl" is null).

[GEOS-11751](https://osgeo-org.atlassian.net/browse/GEOS-11751): When calling the icon service, use either the workspace-prefixed style name or add the workspace name to the path.

  

# Checklist

### DIFF
--- a/src/wms/src/main/java/org/geoserver/wms/icons/IconProperties.java
+++ b/src/wms/src/main/java/org/geoserver/wms/icons/IconProperties.java
@@ -12,6 +12,7 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.Collections;
 import java.util.Map;
+import org.geoserver.catalog.WorkspaceInfo;
 import org.geoserver.config.GeoServerDataDirectory;
 import org.geoserver.ows.URLMangler.URLType;
 import org.geoserver.ows.util.ResponseUtils;
@@ -34,7 +35,7 @@ public abstract class IconProperties {
 
     public abstract Double getHeading();
 
-    public abstract String href(String baseURL, String workspace, String styleName);
+    public abstract String href(String baseURL, WorkspaceInfo workspace, String styleName);
 
     public abstract Style inject(Style base);
 
@@ -68,10 +69,11 @@ public abstract class IconProperties {
             }
 
             @Override
-            public String href(String baseURL, String workspace, String styleName) {
+            public String href(String baseURL, WorkspaceInfo workspace, String styleName) {
                 String stylePathFragment;
                 if (workspace != null) {
-                    stylePathFragment = ResponseUtils.urlEncode(workspace) + "/" + ResponseUtils.urlEncode(styleName);
+                    stylePathFragment =
+                            ResponseUtils.urlEncode(workspace.getName()) + "/" + ResponseUtils.urlEncode(styleName);
                 } else {
                     stylePathFragment = ResponseUtils.urlEncode(styleName);
                 }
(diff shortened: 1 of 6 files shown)


src/wms/src/main/java/org/geoserver/wms/icons/IconProperties.java
```diff
@@ -130,54 +132,69 @@ public Double getHeading() {
...
+                            } else if (workspaceStyles != null
+                                    && file.getAbsolutePath().startsWith(workspaceStyles.getAbsolutePath())) {
+                                // part of the workspace styles directory
+                                String relativePath = file.getAbsolutePath()
+                                        .substring(workspaceStyles
+                                                        .getAbsolutePath()
+                                                        .length()
+                                                + 1);
+                                file = new File(relativePath);
+                                if (file != null) {
```
mprins, line 179: I don't think `file` can ever be `null` here

src/wms/src/main/java/org/geoserver/wms/icons/IconProperties.java
```diff
@@ -130,54 +132,69 @@ public Double getHeading() {
...
-                            if (file != null && styles != null) {
-                                return ResponseUtils.buildURL(
-                                        baseURL,
-                                        "styles/" + styles.toURI().relativize(graphicFile.toURI()),
-                                        null,
-                                        URLType.RESOURCE);
-                            } else {
-                                // we don't know how to handle this then...
-                                return null;
+                                if (file != null) {
```
mprins, line 162: And here: I don't think file can ever be null here

mprins: it does not hurt

src/wms/src/main/java/org/geoserver/wms/icons/IconProperties.java
```diff
@@ -130,54 +132,69 @@ public Double getHeading() {
...
+                                        .substring(workspaceStyles
+                                                        .getAbsolutePath()
+                                                        .length()
+                                                + 1);
+                                file = new File(relativePath);
+                                if (file != null) {
+                                    return buildGraphicURL(
+                                            baseURL,
+                                            "styles/" + workspace.getName(),
+                                            workspaceStyles
```
aaime, line 183: Since you already rolled out a method for this, what about giving it both the styles and graphicfile reference, and handle the following business in there, rather than duplicating it twice?

```java
workspaceStyles
                                                    .toURI()
                                                    .relativize(graphicFile.toURI())
                                                    .toString()
```

aaime (commented):
Looks good to me, see one suggestion below.

src/wms/src/main/java/org/geoserver/wms/icons/IconProperties.java
vuilleumierc: Good point

src/wms/src/main/java/org/geoserver/wms/icons/IconProperties.java
vuilleumierc: Looking closer you are right, the constructor will never return null. Furthermore the `relativePath` is not useful since it's recomputed later with `relativize()`, so I have removed these couple of lines

mprins (approved):

PR #8491, opened 2025-04-17
[GEOS-11800]: Spatiotemporal Zonal Statistics WPS process

dromagnoli:
[GEOS-11800](https://osgeo-org.atlassian.net/browse/GEOS-11800) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

SpatioTemporal zonal stats WPS process.
The PR depends on this GeoTools PR: https://github.com/geotools/geotools/pull/5153
where common/shared code has been refactored into CoverageUtilities methods.

Build Failures are expected before merging the above PR.
  

# Checklist


src/extension/wps/wps-core/src/main/java/org/geoserver/wps/gs/SpatioTemporalZonalStatistics.java
```diff
@@ -0,0 +1,583 @@
...
+            }
+            return aggregator;
+        }
+    }
+
+    private static double getStatsValue(ZonalStats zonalStats, Statistic statistic) {
+        return zonalStats.statistic(statistic).results().get(0).getValue();
+    }
+
+    private static void addAttributes(SimpleFeatureTypeBuilder tb, Set<Statistic> requestedStats) {
```
aaime, line 472: I'd suggest keeping this method close to its only usage point.

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/gs/SpatioTemporalZonalStatistics.java
```diff
@@ -0,0 +1,583 @@
...
+
+        @Override
+        public boolean hasNext() {
+            return !features.isEmpty() || zones.hasNext();
+        }
+
+        @Override
+        public SimpleFeature next() throws NoSuchElementException {
+            // build the next set of features if necessary
+            if (features.isEmpty()) {
```
aaime, line 323: How do ever get more than one feature in this buffer? It would be simpler to understand if it was a ``SimpleFeature nextFeature``, null or not.

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/gs/SpatioTemporalZonalStatistics.java
```diff
@@ -0,0 +1,583 @@
...
+            }
+        }
+
+        private StatisticsAggregator processStatistics(Geometry geometry) throws TransformException, IOException {
+            GridCoverage2D refCoverage = null;
+            GridCoverage2D dataCoverage = null;
+            GridCoverage2D cropped = null;
+            ReferencedEnvelope geometryEnvelope = null;
+            List<Range<Double>> noDataValueRangeList = null;
+            ROI roi = null;
```
aaime, line 391: This is not a C program, are you sure all these variables cannot be declared closer to their first usage point? (e..g, the coverages could be declared just before the try for example)

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/gs/SpatioTemporalZonalStatistics.java
```diff
@@ -0,0 +1,583 @@
...
+            if (requestedStats.contains(Statistic.MEAN)) {
+                builder.add(stats.getAggregatedMean());
+            }
+            if (requestedStats.contains(Statistic.MEDIAN)) {
+                builder.add(stats.getAggregatedMedian());
+            }
+        }
+
+        private StatisticsAggregator processStatistics(Geometry geometry) throws TransformException, IOException {
+            GridCoverage2D refCoverage = null;
```
aaime, line 386: The refCoverage is only used once to figure out the roi/nodata, but it's kept around for the whole computation. I would advise using a boolean flag instead, and avoid using the refCoverage.

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/gs/SpatioTemporalZonalStatistics.java
```diff
@@ -0,0 +1,583 @@
...
+                    timeParam.setValue(Collections.singletonList(time));
+                } else if (temporalItem instanceof DateRange) {
+                    DateRange range = (DateRange) temporalItem;
+                    timeParam.setValue(Collections.singletonList(range));
+                } else {
+                    throw new IllegalArgumentException("Unsupported temporal item: " + temporalItem);
+                }
+                GeneralParameterValue[] gpv = {timeParam};
+                try {
+                    dataCoverage = reader.read(gpv);
```
aaime, line 408: This might be dangerous, if the reader can only do immediate read, I would suggest trying to read only the portion we need (get the bbox of the current zone, make sure partial pixels are covered just in case by increasing it by the pixel resolutions in all directions)

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/gs/SpatioTemporalZonalStatistics.java
```diff
@@ -0,0 +1,583 @@
...
+
+    private Catalog catalog;
+
+    public SpatioTemporalZonalStatistics(Catalog catalog) {
+        this.catalog = catalog;
+    }
+
+    @DescribeResult(
+            name = "result",
+            description = "A feature collection containing aggregated statistics on each zone",
```
aaime, line 92: I would expand a bit, this is most of what they user will see. If  it does not fit well here in the code, you can probably create a separate constant and refer it from here.

aaime (changes requested):
Looks good in general, a few issues noted below.

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/gs/SpatioTemporalZonalStatistics.java
dromagnoli: Right. beside the 2 coverages, all the other variables are initialized once and reused within the for loop.
I'm moving the 2 coverages closer to their usage.

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/gs/SpatioTemporalZonalStatistics.java
```diff
@@ -0,0 +1,583 @@
...
+                requestedStats.add(stat);
+            } catch (IllegalArgumentException e) {
+                throw new IOException("Unknown statistic: " + statName, e);
+            }
+        }
+
+        return requestedStats;
+    }
+
+    private static List<Object> parseTimes(GridCoverage2DReader reader, String timeValues)
```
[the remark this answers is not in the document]
dromagnoli: Fixed

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/gs/SpatioTemporalZonalStatistics.java
dromagnoli: Fixed

src/extension/wps/wps-core/src/main/java/org/geoserver/wps/gs/SpatioTemporalZonalStatistics.java
dromagnoli: Fixed

dromagnoli:
@aaime 
All feedbacks have been addressed.

PR #8512, opened 2025-04-28
[GEOS-11805] Option to disable stored queries

Mike7311:
https://osgeo-org.atlassian.net/browse/GEOS-11805

The main goal of this PR is to add an option to WFS configuation page that would allow the user to disable the stored queries.
This includes:
- removing them from GetCapabilities response when option is enabled.
- throwing an exception if an attempt to invoke operation related to stored queries is made when option is enabled.
  
# Checklist


src/wfs/src/main/java/org/geoserver/wfs/WFSInfoImpl.java
```diff
@@ -213,6 +224,7 @@ public int hashCode() {
         result = prime * result + (getFeatureOutputTypeCheckingEnabled ? 1231 : 1237);
         result = prime * result + ((getFeatureOutputTypes == null) ? 0 : getFeatureOutputTypes.hashCode());
         result = prime * result + ((csvDateFormat == null) ? 0 : csvDateFormat.hashCode());
+        result = prime * result + (disableStoredQueriesManagement ? 1231 : 1237);
```
aaime, line 227: Thanks for updating hashCode, equals should be updated too

aaime (changes requested):
Overall looks good, please check the documentation build failure and see comments below

src/wfs/src/main/java/org/geoserver/wfs/WFSInfoImpl.java
Mike7311: was updated too

doc/en/user/source/services/wfs/webadmin.rst
```diff
@@ -227,6 +227,10 @@ Stored Queries
 
 Selecting the :guilabel:`Allow Global Stored Queries` checkbox determines if global stored queries will included for usage in workspace virtual services, or not. When disabled, only stored queries created inside the workspace will be visible.
 
+.. figure:: img/disable-stored-queries-management.png
+
+Selecting the :guilabel:`Disable stored queries management` checkbox will disable stored queries write operations from being present in GetCapabilities response. Attempting to triggering one of operations related to stored queries will result in exception.
```
aaime, line 232:
```suggestion
Selecting the :guilabel:`Disable stored queries management` checkbox will disable stored queries write operations from being present in GetCapabilities response. Attempting to trigger one of the operations related to stored queries will result in an exception.
```

src/wfs/src/test/java/org/geoserver/wfs/v2_0/StoredQueryTest.java
```diff
@@ -679,6 +679,35 @@ public void testDisallowPerWorkspaceQueries() throws Exception {
         }
     }
 
+    @Test
+    public void testDisabledStoredQueriesManagement() {
```
aaime, line 683: We also need a test checking the capabilities documents contents

aaime (changes requested):

src/wfs/src/test/java/org/geoserver/wfs/v2_0/StoredQueryTest.java
Mike7311: Added a GetCapabilitiesTest#testOperationsMetadataWithDisabledStoredQueryManagement

aaime (approved):
LGTM

PR #8547, opened 2025-05-14
[GEOS-11820] - WCS spatial sub-setting does not work when native CRS != declared CRS

skalesse:
[GEOS-11820](https://osgeo-org.atlassian.net/browse/GEOS-11820) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

The goal of this PR is to fix https://osgeo-org.atlassian.net/browse/GEOS-11820. 

When performing WCS requests with spatial sub-setting, the request will fail, in cases where the native CRS (CRS of the data) of a `StructuredGridCoverage2DReader` is not equal the declared CRS in GeoServer coverage catalogue and REPROJECT_NATIVE_TO_DECLARED projection policy is enabled.

**Example**: 

- the native CRS (of the data) has EPSG:4087
- the declared CRS was set to EPSG:4326

Then, when spatial sub-setting WCS  requests are performed, like 

```http://localhost:8080/geoserver/ows?service=WCS&version=2.0.1&request=GetCoverage&coverageId=dwd.modeldata__ICON-D2.Temperature.height_level_above_ground&format=application/x-netcdf&subsettingCRS=http://www.opengis.net/def/crs/EPSG/0/4326&subset=Long(7,13)&subset=Lat(45,55)...```

 the request will fail, because the granules filtering in `org.geoserver.wcs2_0.response.WCSDimensionsSubsetHelper#filterSpatial()` is performed on the declared CRS, not the native CRS at which the granules of the reader are defined. 

Note, that what is the request's sub-setting CRS or output CRS shouldn't matter in this case.  The problem in `org.geoserver.wcs2_0.response.WCSDimensionsSubsetHelper#filterSpatial()`  seems to be that when building the `nativeCRSPolygon` the method uses the declared CRS of the coverage info, not the native:

```
private Filter filterSpatial(GridCoverageRequest gcr, GranuleSource source)
        throws IOException, MismatchedDimensionException, TransformException, FactoryException {

	// extract the envelope from the request
    WCSEnvelope envelope = gcr.getSpatialSubset();
	
	// convert to polygon geometry 
    Polygon llPolygon = JTS.toGeometry(new ReferencedEnvelope(envelope));
    GeometryDescriptor geom = source.getSchema().getGeometryDescriptor();
    PropertyName geometryProperty = ff.property(geom.getLocalName());
	
	// convert the polygon to the native CRS .... Ooops ! wrong CRS used here?!
    Geometry nativeCRSPolygon = JTS.transform(
            llPolygon, CRS.findMathTransform(envelope.getCoordinateReferenceSystem(), coverageInfo.getCRS()));
			
    // build and return the filter 			
    Literal polygonLiteral = ff.literal(nativeCRSPolygon);
    return ff.intersects(geometryProperty, polygonLiteral);
}
```

The fix should be to use `coverageInfo.getNativeCRS()` instead. 

This PR proposes to change `WCSDimensionsSubsetHelper` accordingly. Two test-cases have been created in module netcdf-out to verify the fix. The reason why the test-cases where put in that module is that we need a `StructuredGridCoverage2DReader` as well as support for multi-dimensional output format (such as netcdf). 

# Checklist

### CODE REVIEW

**Comment by skalesse** [comment]
Hm, the two builds were failing with test-case errors for components `gs-sec-jdbc` and `gs-app-schema-test`.  I am not aware that the changes in `wcs2_0` should influence those  tests in any kind.

File: src/wcs2_0/src/main/java/org/geoserver/wcs2_0/response/WCSDimensionsSubsetHelper.java [hunk 1]
```diff
@@ -1033,7 +1033,7 @@ private Filter filterSpatial(GridCoverageRequest gcr, GranuleSource source)
         GeometryDescriptor geom = source.getSchema().getGeometryDescriptor();
         PropertyName geometryProperty = ff.property(geom.getLocalName());
         Geometry nativeCRSPolygon = JTS.transform(
-                llPolygon, CRS.findMathTransform(envelope.getCoordinateReferenceSystem(), coverageInfo.getCRS()));
```
aaime, line 1036: This fix, as is, introduces a different kind of regression. When the srs handling is "force declared" the native CRS must be ignored and the declared used in its place.

Please:
* Check the srs handling to decide which CRS to use
* In case of "force declared" use the declared CRS, otherwise use the nativ CRS
* Add a test for the force-declared case too

aaime (changes requested):
The change introduces a regression for the force-declared case, see comment below.

The test data must be reduced in size too, 1.8MB worth of test data is too large. The test and behavior should not be resolution dependent, the easiest way to reduce the size of the test image by both clipping it and reducing the pixel resolution (both can be achieved using gdal_translate for example).
The test data should be a handful of kilobytes.

aaime:
Ah, regarding the random failures, you might want to rebase on main, action has been taken to avoid at least a few of them

src/wcs2_0/src/main/java/org/geoserver/wcs2_0/response/WCSDimensionsSubsetHelper.java
skalesse: Ah, thanks for the tip. You're right that case needs to be handled as well. I will go and see if I can update the PR and provide a test-case for fore-declared

skalesse: hi @aaime , 

I have a question regarding the SRS handling. To me it's not quite clear how this is supposed to work: Imagine I have the data as described in the test-case: 

- native CRS : EPSG:4087
- declared CRS: EPSG:4326
- SRS handling FORCE_DECLARED

Now I am using declared CRS for the query, because FORCE_DECLARED is used: This will not work properly, right? Because still, the data is on EPSG:4087, so my query based on EPSG:4326 will return nothing useful. 

What am I missing here?
Thanks!

skalesse:
Hi @jratike80 ,

the plan is that the change will 100% backward compatible. In fact it should only fix the bug native CRS != declared CRS. 

Please note, that "native vs. declared CRS" has nothing to do with the subSettingCRS or the outputCRS. In your example, you are requesting a coverage on EPSG:3857 (outputCRS) and you are performing spatial sub-setting also on EPSG:3857. That's totally fine, but now imagine that the GeoTiff is defined on any other CRS (e.g. EPSG:4087) - that's the native CRS, and imagine also that the GeoServer layer has the declared CRS configured to be EPSG:4326. The latter two CRS's are what we are dealing with here in this PR. That's talking about the GeoServer's internal management of the coverage's spatial representation. 

So, of course the request you mentioned above should still work.

jratike80:
In my example the native CRS is EPSG:4326 and the declared CRS is EPSG:3857. Isn't it exactly `native CRS != declared CRS`?

skalesse:
> In my test the native CRS is EPSG:4326 and the declared CRS is EPSG:3857. Isn't it exactly `native CRS != declared CRS`?

If that's how you have configured your layer, then Yes - that would be native != declared. 

And yes to your other question, plain GeoTiff doesn't seem to come across this path, because they are not `StructuredGridCoverage2DReader`. Not sure though about your data, because if it's processed using the image_mosaic plugin, then that would be `StructuredGridCoverage2DReader`.

jratike80:
> If that's how you have configured your layer, then Yes - that would be native != declared.

Sorry, I forgot to add a link to my comment on the ticket. I made a test with a single image "Image_Sample" and with a mosaic "mosaic". Subsetting works in both cases.

https://osgeo-org.atlassian.net/browse/GEOS-11820?focusedCommentId=86639

> I re-published the “worldImageSample” store, previously published as “nurc:Img_Sample”, but this with name “nurc:usa and with settings:
> native SRS = EPSG:4326
> declared SRS = EPSG:3857
> SRS handling = Reproject native to declared
> 

I did the same thing also with "mosaic".

skalesse:
> > If that's how you have configured your layer, then Yes - that would be native != declared.
> 
> Sorry, I forgot to add a link to my comment on the ticket. I made a test with a single image "Image_Sample" and with a mosaic "mosaic". Subsetting works in both cases.
> 
> https://osgeo-org.atlassian.net/browse/GEOS-11820?focusedCommentId=86639
> 
> > I re-published the “worldImageSample” store, previously published as “nurc:Img_Sample”, but this with name “nurc:usa and with settings:
> > native SRS = EPSG:4326
> > declared SRS = EPSG:3857
> > SRS handling = Reproject native to declared
> 
> I did the same thing also with "mosaic". Could it be that because the data in "mosaic" is from Italy, the wrong EPSG:3857 bounding box still finds the right EPSG:4326 granules? Or maybe all granules.

Alright, I have finally taken the time to check your setup. The reason why it works for you, is that 

- yes, your reader is `StructuredGridCoverage2DReader` (actually a wrapped `WorldImageReader` ), **but**
- you are not specifying an output format, so the default "image/tiff" is used, which is not a multi-dimensional output format.

That's why in all your examples https://github.com/geoserver/geoserver/blob/main/src/wcs2_0/src/main/java/org/geoserver/wcs2_0/GetCoverage.java#L265-L336 is skipped and https://github.com/geoserver/geoserver/blob/main/src/wcs2_0/src/main/java/org/geoserver/wcs2_0/GetCoverage.java#L339 is performed instead. So we are not passing the faulty code path.

jratike80:
All right, you did mention that the bug requires multidim output format but I did not notice that so I tested with the default installation that lacks NetCDF. NetCDF output requires this extension https://docs.geoserver.org/2.27.x/en/user/extensions/netcdf-out/index.html, is that correct?
Of the default formats https://docs.geoserver.org/main/en/user/services/wcs/outputformats.html GML may support multidimensions. Does the bug affect also GML output?

skalesse:
> All right, you did mention that the bug requires multidim output format but I did not notice that so I tested with the default installation that lacks NetCDF. NetCDF output requires this extension https://docs.geoserver.org/2.27.x/en/user/extensions/netcdf-out/index.html, is that correct? Of the default formats https://docs.geoserver.org/main/en/user/services/wcs/outputformats.html GML may support multidimensions. Does the bug affect also GML output?

Yes, that should be the plugin, About GML, I don't know if it's mutli-dimensional. If it was, then it'd be affected, if not then no. I guess now that you have a reproducible case that fails, you can try yourself. :-)

jratike80:
> I guess now that you have a reproducible case that fails, you can try yourself. :-)

Can I try it with the "mosaic" layer if I install the NetCDF plugin? Do you think it is easier for me to test than for you to run your test with `&format=application/gml+xml`?

I have never done anything with NetCDF and I am only a GeoServer user, not a developer. I have some experience on WCS and I am willing to help with bug reports by trying to reproduce the issues but I prefer to use brand new bin-zip installations and the default data directory, if possible. And unfortunately I do not know how to judge if some multidimensional GML coverage is correct or not.

I can read from the WMS 2.0.1 standard

> WCS 2.0 uses the coverage model of the GML Application Schema for Coverages [OGC 09-
> 146r1] which has been developed with the goal that coverages handled by a WCS can be 
> more easily interchanged with other OGC services. WCS 2.0 supports all coverage types supported by said Application Schema; it is not constrained to quadrilateral grid coverages like 
> previous WCS versions

I suppose that the mentioned GML schema covers multidimensional coverages, but the standard does not require GML as supported format. But maybe it would be worth having a test in the CI also for GML because GeoServer by default supports it.

aaime:
Looks good, merging.

skalesse:
@aaime , for the failed backports,do you need help to resolve?

aaime:
Ah yes, if you want the fix to be backported, you'll need to create backport PRs.

skalesse:
at least 2.27.x would be nice, 'cos it's the current release branch

aaime:
A fix can be backported on both stable and maintenance.

skalesse:
I have created the two PRs:

- 2.26.x: https://github.com/geoserver/geoserver/pull/8621
- 2.27.x: https://github.com/geoserver/geoserver/pull/8620

Thanks for looking into them. Both are really only cherry-picks from the original.

skalesse:
> I have created the two PRs:
> 
>     * 2.26.x: [[Backport 2.26.x] [GEOS-11820] - WCS spatial sub-setting does not work when native CRS … #8621](https://github.com/geoserver/geoserver/pull/8621)
> 
>     * 2.27.x: [[Backport 2.27.x] [GEOS-11820] - WCS spatial sub-setting does not work when native CRS … #8620](https://github.com/geoserver/geoserver/pull/8620)
> 
> 
> Thanks for looking into them. Both are really only cherry-picks from the original.

I have no idea why these two are failing now. Both errors look unrelated.

PR #8566, opened 2025-05-22
[GEOS-11839]: WPS download NetCDF output format

dromagnoli:
[GEOS-11839](https://osgeo-org.atlassian.net/browse/GEOS-11839) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

  

# Checklist


src/community/wps-download-netcdf/src/main/java/org/geoserver/wps/ppio/BaseNetCDFPPIO.java
```diff
@@ -0,0 +1,100 @@
...
+
+        public NetCDF4PPIO() {
+            super(NetCDFUtilities.NETCDF4_MIMETYPE);
+        }
+    }
+
+    private NetCDFCoverageResponseDelegate delegate;
+
+    protected BaseNetCDFPPIO(String mimeType) {
+        super(GridCoverage2D.class, GridCoverage2D.class, mimeType);
```
aaime, line 67: This has me wondering... the netcdf output format will appear as a possibility for any process that generates a GridCoverage2D,e.g., coverage clipping for example. This could generate confusing:
* Is the NetCDF output format able to encode something that's not a GranuleStack (a plain jane gridcoverage coming  from a tiff file?)
* Even if it did, one might then assume it's possible to take a NetCDF in input, run a clip process, and grab a ND output via this PPIO... but it's not the case because the process in the middle cannot preserve dimensions.

Can you bind the PPIO to GranuleStack instead? Would it work anyways?

src/community/wps-download-netcdf/src/main/java/org/geoserver/wps/ppio/BaseNetCDFPPIO.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+    }
+
+    @Override
+    public String getFileExtension() {
+        return "nc";
+    }
+
+    @Override
+    public Object decode(InputStream input) throws Exception {
```
aaime, line 97: In addition to this, can you implement getDirection and specify it's PPIODirection.ENCODING only?

src/community/wps-download-netcdf/src/main/java/org/geoserver/wps/ppio/BaseNetCDFPPIO.java
```diff
@@ -0,0 +1,100 @@
...
+    @Override
+    public void encode(Object value, OutputStream os) throws Exception {
+        encode(value, Collections.emptyMap(), os);
+    }
+
+    @Override
+    public void encode(Object value, Map<String, Object> encodingParameters, OutputStream os) throws Exception {
+        GridCoverage2D coverage = (GridCoverage2D) value;
+        try {
+            Map<String, String> params = encodingParameters.entrySet().stream()
```
aaime, line 80: What about validating the parameters and throwing an error instead? Or at least writing a warning? Otherwise the user making a typo may not even realize the parameter provided was ignored.

src/community/wps-download-netcdf/src/test/java/org/geoserver/wps/ppio/NetCDFDownloadProcessTest.java
```diff
@@ -0,0 +1,222 @@
...
+        super.onSetUp(testData);
+        testData.addRasterLayer(TIMESERIES, "timeseries.zip", null, getCatalog());
+        setupRasterDimension(TIMESERIES, ResourceInfo.TIME, DimensionPresentation.LIST, null, "ISO8601", null);
+    }
+
+    protected WPSResourceManager getResourceManager() {
+        return GeoServerExtensions.bean(WPSResourceManager.class);
+    }
+
+    public static class AutoDisposableNetCDFReader extends NetCDFReader implements AutoCloseable {
```
aaime, line 64: One day we should make  all readers AutoCloseable. Something to consider as a stretch goal for GeoServer 3 I guess.

src/community/wps-download-netcdf/src/test/java/org/geoserver/wps/ppio/NetCDFDownloadProcessTest.java
```diff
@@ -0,0 +1,222 @@
...
+import ucar.nc2.Attribute;
+import ucar.nc2.Dimension;
+import ucar.nc2.Variable;
+import ucar.nc2.dataset.DatasetUrl;
+import ucar.nc2.dataset.NetcdfDataset;
+import ucar.nc2.dataset.NetcdfDatasets;
+
+public class NetCDFDownloadProcessTest extends WPSTestSupport {
+
+    private static QName TIMESERIES = new QName(WCS_URI, "timeseries", WCS_PREFIX);
```
aaime, line 50: That's good testing but I'm wondering about a second dimension, e.g., an elevation, and then checking about filtering only one dimension, or the other, or both.

src/extension/wps-download/src/main/java/org/geoserver/wps/gs/download/RasterDownload.java
```diff
@@ -207,188 +214,246 @@ public Resource execute(
...
+                    useTargetCrsAsNative,
+                    ctx.requestedGridGeometry,
+                    parameterDescriptors);
+
+            ctx.computeBackgroundValues(readParameters);
+            slicer = SpatioTemporalCoverageSlicer.SpatioTemporalCoverageSlicerHolder.get();
+            if (slicer == null) {
+                slicer = new SpatioTemporalCoverageSlicer(reader, coverageInfo, filter);
+                slicer.initialize();
+            }
```
aaime, line 243: I get the intention (avoid querying the same granule source twice) but it has a potential to go sideways. Clients like mapstore call the estimator directly before calling the download process... is this thread local cleared also in that case?

Generally speaking, we need some mechanism to ensure that the  thread local gets removed no matter what, regardless of whether the process runs in foreground or in an async thread, like org.geoserver.wps.ProcessListener, registered as a bean in the application context.

Also, I'd recommend moving all the setup machinery in the part of code asking to grab the slicer, to aviod duplications.

aaime (changes requested):
Overall looks good, thanks! See comments below.

doc/en/user/source/community/wps-download-netcdf/index.rst
```diff
@@ -0,0 +1,132 @@
...
+ #. Restart GeoServer
+
+
+Module description
+------------------
+This module adds further capabilities to the :ref:`community_wpsrawdownload` Raw Data Download process of the :ref:`wpsdownload` extension, by supporting dimension-aware filtering and NetCDF output.
+It enables users to download spatio-temporal grid coverages (such as climate or remote sensing data, for example) in NetCDF format, 
+while filtering the data based on specific dimension values—like time, elevation, or other custom-defined dimensions.
+The filter is applied directly to the underlying data source (GranuleSource), and the matching data entries (Granules) 
+are grouped into a GranuleStack—a collection of GridCoverage2D instances, each representing a unique combination of dimension values.
```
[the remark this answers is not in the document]
dromagnoli: Fixing...

src/community/wps-download-netcdf/src/main/java/org/geoserver/wps/ppio/BaseNetCDFPPIO.java
dromagnoli: Sure. I didin't know about that method. Fixing.

src/community/wps-download-netcdf/src/main/java/org/geoserver/wps/ppio/BaseNetCDFPPIO.java
dromagnoli: Fixing...

src/community/wps-download-netcdf/src/test/java/org/geoserver/wps/ppio/NetCDFDownloadProcessTest.java
dromagnoli: Fixed.

src/extension/wps-download/src/main/java/org/geoserver/wps/gs/download/RasterDownload.java
dromagnoli: Added a SlicerCleanupListener that takes care of removing the ThreadLocal.
Initialization also moved in the slicer creation.

dromagnoli:
All the feedbacks have been addressed

PR #8567, opened 2025-05-22
Sort rules in layers.properties alphabetically when updating file

petersmythe:
By applying a defined sort order, it becomes easier to find and manage changes in .properties (and soon all .xml) files.


src/main/src/main/java/org/geoserver/security/impl/ServiceAccessRuleDAO.java
```diff
@@ -113,7 +113,8 @@ ServiceAccessRule parseServiceAccessRule(String ruleKey, String ruleValue) {
     @Override
     protected Properties toProperties() {
         Properties props = new Properties();
-        for (ServiceAccessRule rule : rules) {
+        TreeSet<ServiceAccessRule> sortedRules = new TreeSet<>(rules);
```
mprins, line 116: You could use a sorted stream instead

aaime: Regardless, does it actually work? Properties extends HashTable, putting data in it may scramble the order anyways?

petersmythe:
Superseded by https://github.com/geoserver/geoserver/pull/8665

PR #8570, opened 2025-05-28
[GEOS-11688] After restart unable to load External Style SLD when req…

Geodan-RoemerR:
[GEOS-11688](https://osgeo-org.atlassian.net/browse/GEOS-11688) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

When requesting multiple WMS tiles with an remote SLD at the same time after a Geoserver restart lead to an error. The tiles could not be generated. As explained in:
https://osgeo-org.atlassian.net/browse/GEOS-11688
This PR fixed the logic which caused this bug, and added a test.

# Checklist

### DIFF
--- a/src/wms/src/test/java/org/geoserver/wms/map/GetMapKvpRequestReaderTest.java
+++ b/src/wms/src/test/java/org/geoserver/wms/map/GetMapKvpRequestReaderTest.java
@@ -14,9 +14,9 @@ import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.spy;
-import static org.mockito.Mockito.when;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.*;
+import static org.mockito.Mockito.times;
 
 import com.sun.net.httpserver.HttpHandler;
 import com.sun.net.httpserver.HttpServer;
@@ -39,9 +39,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.TimeZone;
-import java.util.concurrent.Executors;
-import java.util.concurrent.ThreadPoolExecutor;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.media.jai.InterpolationBicubic;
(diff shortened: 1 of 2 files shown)


sebastianfrey:
~Backporting this fix to 2.27.x and 2.26.x would be appreciated.~

Edit: commented the wrong issue

Geodan-RoemerR:
> Backporting this fix to 2.27.x and 2.26.x would be appreciated.

@sebastianfrey not very experienced with this. 
I am not able to add the backport label, not here and not in a new pr. Could you add it? Or should i do something else

sebastianfrey:
> > Backporting this fix to 2.27.x and 2.26.x would be appreciated.
> 
> @sebastianfrey not very experienced with this. I am not able to add the backport label, not here and not in a new pr. Could you add it? Or should i do something else

 @Geodan-RoemerR Sorry for the confusion, I did comment on the wrong issue. So you can ignore my comment.

src/wms/src/test/java/org/geoserver/wms/map/GetMapKvpRequestReaderTest.java
```diff
@@ -771,6 +772,65 @@ protected CloseableHttpResponse executeRequest(HttpCacheContext cacheContext, Ht
...
+                try {
+                    URL url = new URL("http://localhost/geoserver/rest/sldurl/sample.sld");
+                    String urlDecoded = URLDecoder.decode(url.toExternalForm(), "UTF-8");
+                    HashMap kvp = new HashMap<>();
+                    kvp.put("layers", getLayerId(BASIC_POLYGONS));
+                    kvp.put("sld", urlDecoded);
+
+                    reqReader.read(spyRequest, parseKvp(kvp), caseInsensitiveKvp(kvp));
+                    assertSameHeader(headers, expectedBasicHeader);
+                } catch (Exception ignored) {
```
aaime: I don't think the exception should be ignored, the test report states this exception happening due to race conditions:

```
21 Jan 12:21:02 WARN   [geoserver.ows] - Exception while getting SLD.
java.lang.NullPointerException
	at org.geoserver.wms.map.GetMapKvpRequestReader.executeRequest(GetMapKvpRequestReader.java:1001)
	at org.geoserver.wms.map.GetMapKvpRequestReader.getHttpInputStream(GetMapKvpRequestReader.java:933)
```

If you're ignoring the exception because the request is not complete enough to pass a full reqReader.read, add some extra bits instead, but make the test fail in case any exception is thrown during reading.

Geodan-RoemerR: good remark. i addressed you comments, made sure the requests succeeded and assert that the same httpclient instance is used. the test fails if any exception occurs during reading

aaime:
All the builds are red, due to some changes happened in GeoTools. Could you rebase this PR on top of main? Should make builds go back to green

Geodan-RoemerR:
can be retested

mprins:
@Geodan-RoemerR your branch is missing (amongst other things) the Java 17 upgrade commits, hence all builds are failing. Please [`git fetch upstream;git rebase upstream/main`](https://docs.github.com/en/get-started/using-git/about-git-rebase) your branch and `git push -f`, assuming https://github.com/geoserver/geoserver is set up as upstream repository

aaime:
I apologize for these continous rebuild requests... it's not the best period to have a PR reviewed, we had FOSS4G EU last week (and the week before, necessary preparations for the conference), and we are working on GeoServer 3, which has several very deep changes landing on the code base, one after the other.

PR #8571, opened 2025-05-28
[GEOS-11844] Edit Layer page freezes when displayed in fr

prushforth:
[GEOS-11844](https://osgeo-org.atlassian.net/browse/GEOS-11844) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

Fix for [GEOS-11844] Edit Layer page freezes when displayed in fr

When displaying the Edit Layer page in french only (afaict), the wicket javascript throws an exception and the page freezes.  It probably affects all released versions, because the problem is an unescaped apostrophe in the reset attribute grid changes confirmation dialog, which has been around for a while, I think.  The solution is to double escape the ' in the string because it’s getting parsed a couple of times, I think. Double escape in this context means like this: \\\\' 
  
# Checklist

- [ ] ~New unit tests have been added covering the changes.~
- [ ] ~[Documentation](https://github.com/geoserver/geoserver/tree/main/doc/en/user/source) has been updated (if change is visible to end users).~
- [ ] ~The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers).~


prushforth:
whoops closing for a moment

PR #8580, opened 2025-05-30
[GEOS-11814] Create a REST API for Filter Chains

alan-geosolutions:
[GEOS-11814](https://osgeo-org.atlassian.net/browse/GEOS-11814) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

Add the following ‘REST’ endpoints and a set of integration tests The integration tests will
Filter Chains
Body: JSON payloads (omit id; server will assign).
GET /geoserver/rest/security/filterChains → Lists all filter chains (the “web”, “webLogin”, “webLogout”, “rest”, “gwc”, “default”, etc.).
GET /geoserver/rest/security/filterChains/{chainName} → Retrieve a specific chain’s config (e.g., path, class, filters, disabled, etc.).
POST /geoserver/rest/security/filterChains → Create a new chain or update an existing chain.
PUT /geoserver/rest/security/filterChains/{chainName} → Possibly separate “update” operation.
DELETE /geoserver/rest/security/filterChains/{chainName} → Remove a chain

# Checklist

- [✅] I have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/main/CONTRIBUTING.md).
- [✅] I have sent a [Contribution Licence Agreement](https://docs.geoserver.org/latest/en/developer/policies/committing.html) (not required for small changes, e.g., fixing typos in documentation).
- [✅] First PR targets the `main` branch (backports managed later; ignore for branch specific issues).

- [✅] New unit tests have been added covering the changes.
- [✅] [Documentation](https://github.com/geoserver/geoserver/tree/main/doc/en/user/source) has been updated (if change is visible to end users).
- [✅] The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers).
- [✅] There is an issue in the [GeoServer Jira](https://osgeo-org.atlassian.net/browse/GEOS/summary) (except for changes that do not affect administrators or end users in any way).
- [✅] Commit message(s) must be in the form ``[GEOS-XYZWV] Title of the Jira ticket``.
- [✅] Bug fixes and small new features are presented as a single commit.
- [✅] Each commit has a single objective (if there are multiple commits, each has a separate JIRA ticket describing its goal).


doc/en/user/source/rest/filterchains.rst
```diff
@@ -0,0 +1,182 @@
...
+
+Delete an Authentication Filter
+-------------------------------
+
+*Response*
+
+.. admonition:: curl
+
+    curl --location --request DELETE 'http://localhost:9002/geoserver/rest/security/filterChains/web-test-2' \
+    --header 'ACCept: application/json' \
```
aaime: A DELETE comes with no return payload, why have an accept header to start with? (amended to clarify I meant the return payload)

doc/en/user/source/rest/filterchains.rst
```diff
@@ -0,0 +1,182 @@
...
+    --header 'Authorization: XXXXXX'
+
+*Response*
+
+200 OK
+
+.. code-block:: xml
+
+    <filterChain>
```
aaime: Does a deletion actually respond with an XML representation?

doc/en/user/source/rest/filterchains.rst
```diff
@@ -0,0 +1,182 @@
...
+    </filterChain>
+
+Create an Authentication Filter
+-------------------------------
+
+*Response*
+
+.. admonition:: curl
+
+    curl --location 'http://localhost:9002/geoserver/rest/security/filterChains' \
```
aaime: There is no method... I supposed you'd need a -XPOST here?

src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java
```diff
@@ -1251,6 +1252,18 @@ protected MockHttpServletResponse postAsServletResponse(
         return dispatch(request, charset);
     }
 
+    protected MockHttpServletResponse postAsServletResponse(
```
aaime: Does not seem to be used anywhere, if so, please remove.
Also, the body parameter is not used anywhere this way, don't think it can work? (which seems to match the idea that this is probably a leftover).

src/restconfig/src/main/java/org/geoserver/rest/security/AuthenticationFilterChainRestController.java
```diff
@@ -0,0 +1,476 @@
...
+    @GetMapping(
+            produces = {
+                MediaType.APPLICATION_JSON_VALUE,
+                MediaType.APPLICATION_XML_VALUE,
+            })
+    public ResponseEntity<RestWrapper<AuthFilterChainList>> list() {
+        checkAuthorisation();
+        List<AuthFilterChain> filterChains = listFilterChains();
+        AuthFilterChainList authFilterChainList = new AuthFilterChainList(filterChains);
+        return ResponseEntity.ok(wrapObject(authFilterChainList, AuthFilterChainList.class));
```
aaime: ResponseEntity is normally used in GeoServer code when we need to return a specific state along with custom headers (e.g., CREATED along with the location of the new enityt after a POST).
Otherwise the RestWrapper is returned directly.
Was there anything making it necessary to wrap the response object in a ResponseEntity here?

src/restconfig/src/main/java/org/geoserver/rest/security/AuthenticationFilterChainRestController.java
```diff
@@ -0,0 +1,476 @@
...
+            value = "/{chainName}",
+            produces = {
+                MediaType.APPLICATION_JSON_VALUE,
+                MediaType.APPLICATION_XML_VALUE,
+            })
+    public ResponseEntity<RestWrapper<AuthFilterChain>> view(@PathVariable("chainName") String chainName) {
+        checkAuthorisation();
+
+        AuthFilterChain filterChain = viewFilterChain(chainName);
+        return ResponseEntity.ok(wrapObject(filterChain, AuthFilterChain.class));
```
aaime: Same as above

src/restconfig/src/main/java/org/geoserver/rest/security/AuthenticationFilterChainRestController.java
```diff
@@ -0,0 +1,476 @@
...
+            consumes = {
+                MediaType.APPLICATION_JSON_VALUE,
+                MediaType.APPLICATION_XML_VALUE,
+            })
+    public ResponseEntity<RestWrapper<AuthFilterChain>> create(@RequestBody AuthFilterChain authFilterChain) {
+        checkAuthorisation();
+
+        RequestFilterChain filterChain = authFilterChain.toRequestFilterChain();
+        AuthFilterChain savedFilterChain = saveFilterChain(filterChain, authFilterChain.getPosition());
+        return new ResponseEntity<>(wrapObject(savedFilterChain, AuthFilterChain.class), HttpStatus.CREATED);
```
aaime: GeoServer does not normally return the full representation of a resource just created, but an empty response with a Location header, e..g, taking from WorkpaceController:

```java
        // build the new path
        UriComponents uriComponents = getUriComponents(name, builder);
        HttpHeaders headers = new HttpHeaders();
        headers.setLocation(uriComponents.toUri());
        headers.setContentType(MediaType.TEXT_PLAIN);
        return new ResponseEntity<>(name, headers, HttpStatus.CREATED);
```

(where name is the name of the workplace here, other controllers just use the empty string, e.g., FeatureTypeController)

src/restconfig/src/main/java/org/geoserver/rest/security/AuthenticationFilterChainRestController.java
```diff
@@ -0,0 +1,476 @@
...
+                MediaType.APPLICATION_JSON_VALUE,
+                MediaType.APPLICATION_XML_VALUE,
+            })
+    public ResponseEntity<RestWrapper<AuthFilterChain>> update(
+            @PathVariable("chainName") String chainName, @RequestBody AuthFilterChain authFilterChain) {
+        checkAuthorisation();
+
+        RequestFilterChain filterChain = authFilterChain.toRequestFilterChain();
+        AuthFilterChain updatedFilterChain = updateFilterChain(chainName, filterChain, authFilterChain.getPosition());
+        return ResponseEntity.ok(wrapObject(updatedFilterChain, AuthFilterChain.class));
```
aaime: A put in the GeoServer REST API normally returns only a 200 on success (the controller method has typically  a void return type)

src/restconfig/src/main/java/org/geoserver/rest/security/AuthenticationFilterChainRestController.java
```diff
@@ -0,0 +1,476 @@
...
+            value = "/{chainName}",
+            produces = {
+                MediaType.APPLICATION_JSON_VALUE,
+                MediaType.APPLICATION_XML_VALUE,
+            })
+    public ResponseEntity<RestWrapper<AuthFilterChain>> delete(@PathVariable("chainName") String chainName) {
+        checkAuthorisation();
+
+        AuthFilterChain deleted = deleteFilterChain(chainName);
+        return ResponseEntity.ok(wrapObject(deleted, AuthFilterChain.class));
```
aaime: A put in the GeoServer REST API normally returns only a 200 on success (the controller method has typically  a void return type)

src/restconfig/src/main/java/org/geoserver/rest/security/AuthenticationFilterChainRestController.java
```diff
@@ -0,0 +1,476 @@
...
+
+        public void setMessage(String message) {
+            this.message = message;
+        }
+    }
+
+    // ///////////////////////////////////////////////////////////////////////
+    // Helper methods
+
+    // It appears  these two methods require @ControllerAdvice annotation to be present
```
aaime: That is indeed what the "supports" method javadoc says:

```
    /**
     * Any subclass that implements {@link #configurePersister(XStreamPersister, XStreamMessageConverter)} and require
     * this configuration for reading objects from incoming requests should override this method to return true when
     * called from the appropriate controller, and should also be annotated with
     * {@link org.springframework.web.bind.annotation.ControllerAdvice}
     */
```

So why bother adding a comment about it here?

src/restconfig/src/main/java/org/geoserver/rest/security/xml/AuthFilterChain.java
```diff
@@ -0,0 +1,183 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Predicate;
+import org.geoserver.rest.security.AuthenticationFilterChainRestController.CannotMakeChain;
+import org.geoserver.security.HTTPMethod;
+import org.geoserver.security.RequestFilterChain;
+
+@XStreamAlias("filterChain")
+public class AuthFilterChain {
```
aaime, line 21: Looks like a DTO. The GeoServer REST API uses the configuration object directly whenever possible. Looking at it, it's not evident to me why this DTO it was needed? Maybe the HTTP methods list? However XStream does support enumeration mapping directly.

src/restconfig/src/test/java/org/geoserver/rest/security/AuthenticationFilterChainRestControllerMarshallingTest.java
```diff
@@ -0,0 +1,124 @@
+package org.geoserver.rest.security;
```
aaime, line 5: Copyright header missing

src/restconfig/src/test/java/org/geoserver/rest/security/AuthenticationFilterChainRestControllerTest.java
```diff
@@ -0,0 +1,270 @@
+package org.geoserver.rest.security;
```
aaime, line 5: Copyright header missing

src/restconfig/src/main/java/org/geoserver/rest/security/xml/AuthFilterChainList.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package org.geoserver.rest.security.xml;
+
+import com.thoughtworks.xstream.annotations.XStreamAlias;
+import com.thoughtworks.xstream.annotations.XStreamImplicit;
+import java.util.ArrayList;
+import java.util.List;
+
+@XStreamAlias("filterChainList")
+public class AuthFilterChainList {
```
aaime: Same as above, why is it needed? See for example WorspaceController.workspacesGet().

aaime (changes requested):
See comments below.

Also, please, have a functional test like most of the existing rest API test, one that simulates sending a xml/json document to the API and check the response is the expected one, by actually testing its contents (using either the DOM or XMLAssert).

doc/en/user/source/rest/filterchains.rst
mprins: the response can have a payload, eg in the case it is 200 OK or 202 Accepted (and not eg. 204 No Content) it can return a document, adding the accept header to the client request signals the server that this is the accepted format for the response from the server

aaime: Difference between theory and practice. GeoServer delete requests do not have a response payload (there might be some exception I'm not aware of, but I'm sure the vast majority do not, and unless there is a good reason to go off script, this one should not either).

alan-geosolutions: I will just change it, this is what our users expect then that is fine.

src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java
alan-geosolutions: Done

src/restconfig/src/main/java/org/geoserver/rest/security/AuthenticationFilterChainRestController.java
alan-geosolutions: Done

src/restconfig/src/main/java/org/geoserver/rest/security/AuthenticationFilterChainRestController.java
alan-geosolutions: Removed

src/restconfig/src/main/java/org/geoserver/rest/security/xml/AuthFilterChain.java
alan-geosolutions: It has a position property

aaime: Please document it clearly in the class javadoc and also show an example in the documentation on how the position can be modified.

doc/en/user/source/rest/api/filterchains.rst
```diff
@@ -0,0 +1,450 @@
...
+            <string>form</string>
+            <string>anonymous</string>
+        </filters>
+        <disabled>false</disabled>
+        <allowSessionCreation>true</allowSessionCreation>
+        <requireSSL>false</requireSSL>
+        <matchHTTPMethod>false</matchHTTPMethod>
+        <position>1</position>
+    </filterChain>
+
```
aaime, line 273: I may be misunderstanding here but... is this example showing a response from the PUT? (which normally does not return a payload in the GeoServer REST API?).

aaime (changes requested):
Testing still does not conform to the GeoServer practice. 
Quoting what I said before:

> Also, please, have a functional test like most of the existing rest API test, one that simulates sending a xml/json document to the API and check the response is the expected one, by actually testing its contents (using either the DOM or XMLAssert).

The test should be end-to-end, like one performed by an external tester, just using the family of methods like getAsDom, getAsJson, and so on, provided by GeoServerSystemTestSupport.

You can use most core REST API tests as a reference, here is a simple (even if pretty old) one:
https://github.com/geoserver/geoserver/blob/99ae245402fca51e306eaf9817faaa891bf89e73/src/restconfig/src/test/java/org/geoserver/rest/catalog/WorkspaceTest.java#L42

A method that returns a payload should have the response parsed and examined, you can use the DOM, XMLAssert, JSONLib, Jayway JSONPath, up to you.
A method that modifies something shoud be run and the actual changes expected should be verified either by doing a GET afterwards, or by reaching out to the configuration programmatically.

Please make sure the feedback above is applied to the other PRs related to the REST API as well, and let me know when they have all been updated.

PR #8584, opened 2025-06-02
[GEOS-11852] Remove master password info page

sikeoka:
# Checklist


jodygarnett:
I fixed some links; and updated UI for consistency

aaime:
Overall it looks good, but I'm missing something, quoting from the description of the issue:

> The PR replaces the master password info page that dumps the master password into a file with a link to the REST API to get the master password

and also:

> [x] The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers).

but checking the diff I'm not seeing anything related to REST API documentation nor the link to how to get the keystore password using the REST API.
Also related, while the master password has been renamed everywhere to keystore password, the "MasterPasswordController" is at "security/masterpw" and its XML representation also uses "master password" naming.
Kind of wondering if that was an oversight or not... I see there is a problem here, as it would create a backwards incompatbility API wise, but at the same time, users of the REST API also read the rest of the documentation and use the UI, where the term master password was carefully removed.

jodygarnett:
Thanks for feedback @aaime, the rename is left over from the BLM master -> main / master -> keystore switch. I did not see a good reason to break backwards compatibility (especially since this is in someway a fix to be backported).

I did change the description in the REST API docs, just not the endpoint.

action:
- Add an example in the rest docs, for the other pages to link to
- Make a note in the update instructions
- Update the description in the REST API docs

aaime:
@sikeoka are you still around? We have made comments on all your open PRs but saw no feedback for a long while

jodygarnett:
@aaime I think I have mostly taken this over, if I address the three "Actions" listed above can we consider feedback addressed and merge?

jodygarnett:
All feedback has now been addressed.

PR #8594, opened 2025-06-05
[GEOS-11813] Create REST API For Security Providers

alan-geosolutions:
[GEOS-11813](https://osgeo-org.atlassian.net/browse/GEOS-11813) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

Added a new REST API to list, create, update, and delete authentication providers. Includes detailed endpoint documentation and Swagger definitions for ease of integration.

  

# Checklist


src/main/src/test/java/org/geoserver/wfsng/WfsCompatibilityTest.java
```diff
@@ -31,6 +32,7 @@
  *
  * @author Niels Charlier
  */
+@Ignore
```
mprins: if all of this test is to be ignored it should just be removed as dead code. Otherwise at least some comment as to why it is to be ignored... the commit message "Fixed security tests regressions" sugegst something is fixed, but it is actually disabled and not fixed...

aaime: This test should not have been disabled to start with. It's however known to fail if one has a running GeoServer without the default data directory. In the long term it should be migrated to use test containers and a ad-hoc GeoServer instance, for now one just needs to pay attention not to have a GeoServer running on 8080 during the build.

alan-geosolutions: Sorry my bad, I have to disable this test locally

src/restconfig/src/test/java/org/geoserver/rest/security/AuthenticationProviderHelper.java
```diff
@@ -0,0 +1,43 @@
+/* (c) 2025 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.rest.security;
+
+import static org.jsoup.helper.Validate.fail;
```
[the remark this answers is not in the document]
alan-geosolutions: done

src/restconfig/src/test/java/org/geoserver/rest/security/AuthenticationProviderRestControllerTest.java
```diff
@@ -0,0 +1,356 @@
...
+                    authenticationProviderHelper.createUsernamePasswordAuthenticationProviderConfig(name, true);
+            AuthProvider authProvider = new AuthProvider(provider);
+            ResponseEntity<RestWrapper<AuthProvider>> result = controller.update(authProvider.getName(), authProvider);
+            int status = result.getStatusCode().value();
+            assertEquals("Expected an Created response", HttpStatus.SC_OK, status);
+            assertNotNull(Objects.requireNonNull(result.getBody()).getObject());
+            AuthProvider createdAuthProvider = (AuthProvider) result.getBody().getObject();
+            assertNotNull(createdAuthProvider.getId());
+            assertEquals(provider.getName(), createdAuthProvider.getName());
+            assertEquals(provider, authProvider.getConfig());
```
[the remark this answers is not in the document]
alan-geosolutions: done

src/restconfig/src/test/java/org/geoserver/rest/security/AuthenticationProviderRestControllerTest.java
```diff
@@ -0,0 +1,356 @@
...
+                    authenticationProviderHelper.createUsernamePasswordAuthenticationProviderConfig(name, false);
+            AuthProvider authProvider = new AuthProvider(provider);
+            ResponseEntity<RestWrapper<AuthProvider>> result = controller.create(authProvider);
+            int status = result.getStatusCode().value();
+            assertEquals("Expected an Created response", HttpStatus.SC_CREATED, status);
+            assertNotNull(Objects.requireNonNull(result.getBody()).getObject());
+            AuthProvider createdAuthProvider = (AuthProvider) result.getBody().getObject();
+            assertNotNull(createdAuthProvider.getId());
+            assertEquals(provider.getName(), createdAuthProvider.getName());
+            assertEquals(provider, authProvider.getConfig());
```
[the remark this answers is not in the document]
alan-geosolutions: Added call to check provider is valid
checkProvideUsernamePasswordAuthenticationProvider(provider, createdAuthProvider.getConfig());


done

src/restconfig/src/main/java/org/geoserver/rest/security/xml/AuthProviderList.java
```diff
@@ -0,0 +1,26 @@
...
+import com.thoughtworks.xstream.annotations.XStreamImplicit;
+import java.util.Collections;
+import java.util.List;
+
+@XStreamAlias("authProviderList")
+public class AuthProviderList {
+    @XStreamImplicit
+    protected List<AuthProvider> providers;
+
+    public AuthProviderList() {}
```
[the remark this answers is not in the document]
alan-geosolutions: done

src/restconfig/src/test/java/org/geoserver/rest/security/AuthenticationProviderRestControllerTest.java
```diff
@@ -0,0 +1,356 @@
...
+                });
+    }
+
+    @Test
+    public void testList() throws IOException {
+        try {
+            setUser();
+            ResponseEntity<RestWrapper<AuthProviderList>> result = controller.list();
+            int status = result.getStatusCode().value();
+            assertEquals("Expected an OK response", HttpStatus.SC_OK, status);
```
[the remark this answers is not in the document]
alan-geosolutions: This is probably clearer


            HttpStatus status = result.getStatusCode();
            assertEquals("Expected an OK response", HttpStatus.OK, status);

src/restconfig/src/test/java/org/geoserver/rest/security/AuthenticationProviderHelper.java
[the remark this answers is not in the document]
alan-geosolutions: Done

src/main/src/test/java/org/geoserver/wfsng/WfsCompatibilityTest.java
alan-geosolutions: Restored

doc/en/api/1.0.0/authenticationprovider.yaml
```diff
@@ -0,0 +1,317 @@
+
+swagger: '2.0'
```
aaime: Pasting this into [the online swagger editor](https://editor.swagger.io/) it reports the following issues:

image

src/restconfig/src/main/java/org/geoserver/rest/security/AuthenticationProviderRestController.java
```diff
@@ -0,0 +1,425 @@
...
+import org.springframework.web.bind.annotation.PutMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.ResponseStatus;
+import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.util.UriComponents;
+import org.springframework.web.util.UriComponentsBuilder;
+
+/**
+ * Rest Controller for AuthenticationProvider provider resource Extends SequentialExecutionController as we do not want
```
aaime: This is a good point , but then it does not implement SequentialExecutionController and the base class does n't either.

src/restconfig/src/main/java/org/geoserver/rest/security/xml/AuthProvider.java
```diff
@@ -0,0 +1,87 @@
+/* (c) 2025 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.rest.security.xml;
+
+import com.thoughtworks.xstream.annotations.XStreamAlias;
+import org.geoserver.security.config.SecurityAuthProviderConfig;
+
+@XStreamAlias("authProvider")
```
aaime: Already programmatically configured in the XStream setup, this annotation seems redundant.

src/restconfig/src/main/java/org/geoserver/rest/security/xml/AuthProvider.java
```diff
@@ -0,0 +1,87 @@
...
+    private boolean disabled;
+
+    public AuthProvider() {}
+
+    public AuthProvider(SecurityAuthProviderConfig provider) {
+        this.id = provider.getId();
+        this.name = provider.getName();
+        this.className = provider.getClassName();
+        this.userGroupServiceName = "default";
+        this.config = provider;
```
aaime: This looks confusing... I understand the need of wrapping to handle the position, but all the other fields duplicate what's already in the config, which is also serialized, leading to a duplication. E.g.: this is what comes out of the default data directory:

image

With the exception of position, the other fields are duplicated (and the userGroupServiceName is hard coded, too, but no matter, I believe it should be removed)

I've also included the XML of the configuration from the data directory, see how the class name has a clear name rather than being generic? I was expecting to see ``<config type="usernamePassword">``.

The authentication providers are pluggable and various optional modules provide one, e.g., GeoFence, LDAP, and more.
Each one declares a GeoServerSecurityProvider bean that can configure a XStreamProvider, e.g., see the [WebSErviceSecurityProvider](https://github.com/geoserver/geoserver/blob/99ae245402fca51e306eaf9817faaa891bf89e73/src/community/web-service-auth/src/main/java/org/geoserver/security/auth/web/WebServiceSecurityProvider.java#L17) for example.

Normally these are called by [the GeoServerSecurityManager only](https://github.com/geoserver/geoserver/blob/4615b41e6ca359a242c769ca7587278e56ccb37b/src/main/src/main/java/org/geoserver/security/GeoServerSecurityManager.java#L2570) (there was no other need so far) but the rest API should do the same. Besides nicer aliases, the providers might be configuring custom converters and the like (they don't in the current codebase, but we have no control over user provided plugins).

src/restconfig/src/test/java/org/geoserver/rest/security/AuthenticationProviderRestControllerTest.java
```diff
@@ -0,0 +1,330 @@
+package org.geoserver.rest.security;
```
aaime, line 5: Copyright header missing

src/restconfig/src/main/java/org/geoserver/rest/security/xml/AuthProvider.java
```diff
@@ -0,0 +1,87 @@
...
+
+    public String getClassName() {
+        return className;
+    }
+
+    public void setClassName(String className) {
+        this.className = className;
+    }
+
+    public boolean isDisabled() {
```
aaime: How does one get to have a disabled provider in the GeoServer auth providers configuration?

image

The main configuration only has a list of names:

https://github.com/geoserver/geoserver/blob/main/data/release/security/config.xml#L3

and the single provider configuration does not have a notion of enabled/disabled:

https://github.com/geoserver/geoserver/blob/main/data/release/security/auth/default/config.xml

Seems to me that a provider is there, or not, but cannot be there and be disabled?

src/restconfig/src/main/java/org/geoserver/rest/security/AuthenticationProviderRestController.java
```diff
@@ -0,0 +1,425 @@
...
+            if (Strings.isNullOrEmpty(authProvider.getName())) {
+                throw new IllegalArgumentException("Cannot create a new AuthProvider with without a name");
+            }
+
+            if (Strings.isNullOrEmpty(authProvider.getClassName())) {
+                throw new IllegalArgumentException("Cannot create a new AuthProvider with without a className");
+            }
+
+            // Ensure name
+            authProvider.getConfig().setName(authProvider.getName());
```
aaime: I see a lot of authProvider.getConfig() around... why not allocate a local variable for it, when the code is using it heavily, and in the end, only intends to save it?
The AuthProvider wrapper is there just to carry the position (and eventually the enabled flag, assuming it has some use, see my other comments).

aaime (changes requested):

afabiani:
@aaime this one ready for review too. Thank you very much!

doc/en/api/1.0.0/authenticationproviders.yaml
```diff
@@ -0,0 +1,332 @@
...
+      are enabled. Presence = enabled; absence = disabled.
+
+host: localhost:8080
+basePath: /geoserver/rest
+schemes: [http]
+security:
+  - basicAuth: []
+
+paths:
+  /security/authProviders:
```
mprins: As a general practice is it best to try and keep uri's lowercase, so this would become `/security/authproviders`

mprins (commented):
lgtm; I'd prefer having lower-case uri's/paths

doc/en/api/1.0.0/authenticationproviders.yaml
afabiani: @mprins done thanks.

@aaime @mprins I guess I will need to change also the api for the auth filters accordingly... I will open a new PR for that

`authFilters` --> `authfilters`

PR #8607, opened 2025-06-09
[GEOS-11841] Use GeoTools Bill of Materials (BOM) to manage geotools dependency versions

groldan:
[GEOS-11841](https://osgeo-org.atlassian.net/browse/GEOS-11841) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

Integrates the GeoTools Bill of Materials (gt-bom) into GeoServer, centralizing management of GeoTools dependency versions.

* Import `org.geotools:gt-bom` in the root `pom.xml`'s `<dependencyManagement>` section. This makes GeoServer a direct consumer of GeoTools' curated dependency set.
* Remove explicit `<version>${gt.version}</version>` declarations for individual GeoTools artifacts from all module POMs. Versions are now inherited from the imported `gt-bom`.
* Standardize GeoTools test JAR dependencies from `<type>test-jar</type>` to `<classifier>tests</classifier>`, aligning with the conventions established in the `gt-bom`.
* Delegate the following dependency versions to the GeoTools BOM:
  * apache commons (io, logging, commons-lang3, pool, collections, cli, text, dbcp, exec)
  * guava
  * h2
  * hsqldb
  * httpclient
  * imageio-ext
  * jackson-annotations
  * jackson-core
  * jackson-databind
  * jai_core/codec
  * jai_imageio
  * jaiext
  * java-vector-tile
  * jaxb-api
  * jaxb-runtime
  * jaxen
  * jdom2
  * json-sKema
  * jts
  * log4j-1.2-api
  * log4j-api
  * log4j-core
  * log4j-jcl
  * log4j-jul
  * log4j-slf4j2-impl
  * mysql-connector-j
  * ojdbc11
  * oro
  * postgresql
  * slf4j-api
  * solr-solrj
  * unit-api
  * wkb4j

  

# Checklist

### DIFF
--- a/doc/en/developer/source/maven-guide/index.rst
+++ b/doc/en/developer/source/maven-guide/index.rst
@@ -57,7 +57,7 @@ Maven automatically downloads dependencies declared by
 modules being built. In the case of SNAPSHOT dependencies,
 Maven downloads updates each time it performs the first build of the day.
 
-GeoServer depends on SNAPSHOT versions of the GeoTools library.  
+GeoServer depends on SNAPSHOT versions of the GeoTools library managed through a Bill of Materials (BOM).
 The automatic download can result in lengthy build time
 while Maven downloads updated GeoTools modules. 
 If GeoTools was built locally, these downloads are not necessary.
@@ -203,17 +203,37 @@ You may also use an absolute path, if you have a custom data directory you would
 Dependency Management
 ---------------------
 
-GeoServer uses Maven's Bill of Materials (BOM) pattern to manage Spring Framework and Spring Security dependencies centrally. This ensures consistent versions across all modules and simplifies dependency declarations.
+GeoServer uses Maven's Bill of Materials (BOM) pattern to centrally manage dependency versions for GeoTools, Spring Framework, and Spring Security.
+This ensures consistent versions across all modules and simplifies dependency declarations.
 
-Spring Framework and Spring Security BOMs
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Bill of Materials (BOMs)
+^^^^^^^^^^^^^^^^^^^^^^^^
 
-The root ``src/pom.xml`` imports both Spring Framework and Spring Security BOMs in the ``<dependencyManagement>`` section:
+The root ``src/pom.xml`` imports multiple BOMs in the ``<dependencyManagement>`` section:
 
 .. code-block:: xml
 
    <dependencyManagement>
      <dependencies>
+       <!-- GeoTools platform dependencies BOM - provides managed versions for third-party libraries -->
+       <dependency>
+         <groupId>org.geotools</groupId>
+         <artifactId>gt-platform-dependencies</artifactId>
+         <version>${gt.version}</version>
+         <type>pom</type>
+         <scope>import</scope>
+       </dependency>
+
+       <!-- GeoTools BOM - provides access to all GeoTools artifacts -->
+       <dependency>
+         <groupId>org.geotools</groupId>
+         <artifactId>gt-bom</artifactId>
+         <version>${gt.version}</version>
+         <type>pom</type>
+         <scope>import</scope>
+       </dependency>
+       
+       <!-- Spring Framework BOM -->
        <dependency>
          <groupId>org.springframework</groupId>
          <artifactId>spring-framework-bom</artifactId>
@@ -221,6 +241,8 @@ The root ``src/pom.xml`` imports both Spring Framework and Spring Security BOMs 
          <type>pom</type>
          <scope>import</scope>
        </dependency>
+       
+       <!-- Spring Security BOM -->
        <dependency>
          <groupId>org.springframework.security</groupId>
          <artifactId>spring-security-bom</artifactId>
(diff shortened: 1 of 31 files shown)


sikeoka:
I had a question here that is also relevant to https://github.com/GeoWebCache/geowebcache/pull/1415 but the primary advantage of having each project manage its own third-party dependencies was to deal with newly deprecated API causing QA build failures when upgrading dependencies since the QA build is very strict about that. With all dependencies managed from GeoTools, would it be possible that these issues won't be found until the next GS/GWC PR to use the GT snapshot with the upgrade?

aaime:
@sikeoka we have a downstream build, but indeed not a QA downstream build. Given the build times involved, it would be probably best to separate the QA Build from the normal build, have them run in parallel (with QA not running tests).

groldan:
I can see the timing issue re upgrading shared dependencies in GeoTools not immediately complaining on the gs/gwc builds.
How about a downstream QA build in GeoTools that triggers only when gt-platform-dependencies is changed?

groldan:
I would like to merge this PR and tackle the concern raised in [this comment](https://github.com/geoserver/geoserver/pull/8607#issuecomment-2956942313) as a separate issue, for which I've created this ticket: https://osgeo-org.atlassian.net/browse/GEOT-7777

groldan:
Downstream QA Build PR ready: https://github.com/geotools/geotools/pull/5246

jodygarnett:
Downstream QA build is merged, @sikeoka feedback above is addressed

doc/en/developer/source/maven-guide/index.rst
```diff
@@ -203,24 +203,46 @@ You may also use an absolute path, if you have a custom data directory you would
...
+       <!-- GeoTools BOM - provides access to all GeoTools artifacts -->
+       <dependency>
+         <groupId>org.geotools</groupId>
+         <artifactId>gt-bom</artifactId>
+         <version>${gt.version}</version>
+         <type>pom</type>
+         <scope>import</scope>
+       </dependency>
+       
+       <!-- Spring Framework BOM -->
```
jodygarnett, line 236: Minor feedback, can use sphinx to inline the actual example as has been done for QA section of the documentation. This ensure the docs do not fall out of date.

doc/en/developer/source/maven-guide/index.rst
```diff
@@ -231,33 +253,74 @@ The root ``src/pom.xml`` imports both Spring Framework and Spring Security BOMs
...
 
-     <!-- Correct - no version specified -->
+     <!-- Spring Security dependencies - no version specified -->
      <dependency>
        <groupId>org.springframework.security</groupId>
        <artifactId>spring-security-core</artifactId>
      </dependency>
    </dependencies>
 
+For GeoTools test JARs, use the ``<classifier>tests</classifier>`` approach instead of ``<type>test-jar</type>``:
```
jodygarnett, line 283: Q: Is `test-jar` classifier normal for maven projects now? This would not be a bad time to change.

doc/en/developer/source/programming-guide/wps-services/implementing.rst
```diff
@@ -98,19 +111,30 @@ For this example the project will be called "hello_wps".
         </plugins>
        </build>
 
-        <repositories>
-            <repository>
-                <id>boundless</id>
-                <name>Boundless Maven Repository</name>
-                <url>https://repo.boundlessgeo.com/main</url>
-            </repository>
```
jodygarnett, line 106: Ah what is this still doing here! We should inline from the actual code.

doc/en/developer/source/maven-guide/index.rst
groldan: I wasn't sure which one to choose, but `<classifier>tests</classifier>` was the most used one across all poms. Problem is in the gt bom I had to choose one, cause if I used `<classifier>tests</classifier>` in the bom and `<type>test-jar</type>` downstream it wouldn't work.

Hopefully the maven documentation says both ways of specifying the test dependency are completely equivalent, but `<classifier>tests</classifier>` is preferred. See https://maven.apache.org/repositories/dependencies.html. Quoting:

*Consider the artifact org.project:reusable-test-support:1.0:tests:jar. With the type handlers above, maybe surprisingly, a dependency on this very same artifact can be described in two ways:*

```xml
<dependency>
  <groupId>org.project</groupId>
  <artifactId>reusable-test-support</artifactId>
  <version>1.0</version>
</dependency>
```

**and a completely equivalent** dependency is:

```xml
<dependency>
  <groupId>org.project</groupId>
  <artifactId>reusable-test-support</artifactId>
  <version>1.0</version>
  <type>test-jar</type>
</dependency>
```

The obvious difference is the presence of classifier in the first case, and the absence of it in the second. However, in the second case, the type “test-jar” implies a classifier of “tests”. In both cases, the extension is “jar”. The first uses the default value for this property, while the second infers it from the type.

**The first way is more explicit and therefore preferred**. In this case the type “test-jar” serves as an alias for ordinary JARs with the “tests” classifier. If the type handler carries important extra information such as custom packaging, using type is more appropriate.

PR #8616, opened 2025-06-16
[GEOS-11866] Prevent requests setting variables that should only be set by GeoServer

sikeoka:
# Checklist

### DIFF
--- a/src/main/src/test/java/org/geoserver/ows/EnvironmentInjectionCallbackTest.java
+++ b/src/main/src/test/java/org/geoserver/ows/EnvironmentInjectionCallbackTest.java
@@ -1,0 +1,98 @@
+/* (c) 2017 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.ows;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+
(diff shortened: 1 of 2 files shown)


src/main/src/test/java/org/geoserver/ows/EnvironmentInjectionCallbackTest.java
```diff
@@ -0,0 +1,98 @@
...
+        testData.setUpSecurity();
+    }
+
+    @Before
+    public void init() {
+        this.request = new Request();
+        this.kvp = new HashMap<>();
+        this.request.setKvp(this.kvp);
+        this.env = new HashMap<>();
+    }
```
aaime: This arrangement has me confused... the kvp map is the request kvp in a stable way, the env is a field that occasionally becomes the env map in the kvp, but one has to carefully look at the tests to figure that out.
I believe it would be easier to understand if the code was working directly against the request object itself? Or make methods to setup the request rather than keeping everything as a field, or as another possibility, just document clearly the role of each bit using javadoc?

sikeoka: I updated the tests so it might be clearer now?

aaime: Yes, the test intention is clearer now, thank you!

PR #8643, opened 2025-07-04
[GEOS-11879] Xalan causes a java.lang.NoClassDefFoundError

etj:
[GEOS-11879](https://osgeo-org.atlassian.net/browse/GEOS-11879) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

See https://osgeo-org.atlassian.net/browse/GEOS-11879  

Added a dep to xalan:serializer to avoid random errors.

# Checklist

### DIFF
--- a/src/extension/printing/pom.xml
+++ b/src/extension/printing/pom.xml
@@ -48,9 +48,12 @@
           <artifactId>xml-apis</artifactId>
           <groupId>xml-apis</groupId>
         </exclusion>
+        <exclusion>
+          <artifactId>xalan</artifactId>
+          <groupId>xalan</groupId>
+        </exclusion>
       </exclusions>
     </dependency>
-
     <dependency>
       <groupId>org.geoserver.web</groupId>
       <artifactId>gs-web-core</artifactId>


aaime:
The dependency seems to have been removed by @sikeoka in this ticket: https://osgeo-org.atlassian.net/browse/GEOS-11242 and this PR: https://github.com/geoserver/geoserver/pull/7331

The printing module was not modified back then, but I guess the jar got pulled away from the main war and this caused the issue?

etj:
@aaime 

> The dependency seems to have been removed by @sikeoka in this ticket: https://osgeo-org.atlassian.net/browse/GEOS-11242 and this PR: #7331

Yes, all the related issues have been already documented in the issue https://osgeo-org.atlassian.net/browse/GEOS-11879

> The printing module was not modified back then, but I guess the jar got pulled away from the main war and this caused the issue?

I guess that the xalan:xalan in the printing module was using the xalan:serializer jar imported from somewhere else (in the wms module probably, but where it was from is not relevant). Once the serializer was explicitly removed from the geoserver deps, the xalan:xalan in the printing module started misbehaving.

etj:
If xalan is not included in the plugin's zip file, it won't break GoeServer since other registered xml engines will be used, the the serializer jar will be needed.
Our use cases is about a geoserver.war generated with all the required plugins using maven profiles, so the serializer.jar in that casa is needed to avoid the error.

Now, only including the serializer,jar in the war where xalan is included by the mvn dependencies will make geoserver work, and not including the serializer in the plugin where xalan is not present will make geoserver run without errors, **but** the set of jar files in these two cases will differ. Is it acceptable?   
Otherwise, in order to have the plugin and the built geoserver.war aligned, the options are:
- include xalan+serializer in the plugin 
- exclude xalan from the depenendencies, sto that the war won't have neither xalan nor serializer .

etj:
It turns out that when removing xalan, everything still works perfectly, and the print module do actually print.

Adding an `<exclusion>` in the printing module.

The web/app module seems to properly catch the change, and xalan is not in the `dependency:tree`

PR #8648, opened 2025-07-06
More compact, easier to maintain conformance configuration UI (DRAFT)

aaime:
Hi @jodygarnett,
I've started some work to make the OGC API conformance configuration easier to maintain, and more in line with the rest of the GeoServer GUI.

Problems I see right now:
* The GUI for conformance configuration takes a lot of space
* It lists non-configurable items (that's the job of documentation, if anything, nothing else amongst the GeoServer admin panels tries to list options... it's as if the admin GUI was listing that you can do GetFeature on WFS)
* One has to write a lot GUI code for any new conformance class being added
* The configured items are not clearly identifiable (you try to keep the "null" state, but it does not show visually).

This branch contains a **draft** compact GUI code that does not need to be modified when adding a new API conformance, and that's reusable across services. It ends up looking like this:

image

I still dont' like a few things, for example, I had to use reflection to make editing conformance properties generic, and that the code is still pretty maintenance heavy.

Pheraps making those classes basically maps from APIConformace to Boolean value, and having generic methods taking APIConformance (which would be a constant) to decide if it's to be enabled or not would simplify things further.
A custom XStreamSerializer would keep backwards compatibility with the current storage.
Then adding a new APIConformance would literally be just adding a constant, and adding it into the map. Just thinking out loud here.
Also, there is a things that are not used, which could just be removed... like the parent of a APIConformance.

GUI wise, a bit of work is still needed:
* Move the ConformanceTable out to web-ogcapi (generic)
* i18n labels for level, type, table headers

# Checklist


src/extension/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/ConformanceClass.java
```diff
@@ -31,12 +31,6 @@ public class ConformanceClass {
     /** Ids filter conformance class for "OGC API - Features - Part n: Query by IDs" */
     public static final String IDS = "http://www.opengis.net/spec/ogcapi-features-1/1.1/req/ids";
 
-    /**
-     * A custom conformance class for GeoServer own ECQL, not further split into parts (as only GeoServer implements it
-     * anyways
-     */
-    public static final String ECQL = "http://geoserver.org/spec/ecql/1.0/req/gs-ecql";
-
     public static final String ECQL_TEXT = "http://geoserver.org/spec/ecql/1.0/req/ecql-text";
```
aaime, line 34: ECQL was removed as it does not identify a conformance per see, just like CQL2 does not have a CQL2 conformance, but has cql2-text and cql2-json.

src/extension/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/FeatureConformance.java
```diff
@@ -31,8 +31,8 @@ public class FeatureConformance extends ConformanceInfo<WFSInfo> {
             CORE.extend("http://www.opengis.net/spec/ogcapi-features-1/1.0/conf/oas30");
 
     // optional output format from WFS
-    public static final APIConformance GML321 =
-            new APIConformance("http://schemas.opengis.net/gml/3.2.1/gml.xsd", STANDARD);
+    public static final APIConformance GML321 = // this is not a recognized conformance class though??
+            new APIConformance("http://schemas.opengis.net/gml/3.2.1/gml.xsd", STANDARD, "gml321");
```
aaime, line 35: Had to modify a bunch since the end of the URI does not match the field/property name, as assumed by the constructor.

jodygarnett:
Thanks for working on this; I am not sure we need the `id` (it could be a tooltip to save space.

This entire thing is not shaping out how I had envisioned, since the only setting we care about for the most part of enable/disable.

Your PR makes things better and is appreciated!

My expectations are not shaping up:

> I had thought that some detailed settings would be associated with a given conformance.
> 
> We have examples, where we use a checkbox to enable functionality, and then options are shown to manage. 
> 
> 
> 
> Many of the formats have their own options, kind of like the GML and CSV settings:
> 
> 
> 
>

aaime:
> Thanks for working on this; I am not sure we need the `id` (it could be a tooltip to save space).

Sure, why not.

> 
> This entire thing is not shaping out how I had envisioned, since the only setting we care about for the most part of enable/disable.
> 
> Your PR makes things better and is appreciated!

All right, so it seems you like the new GUI.
 
> My expectations are not shaping up:
> 
> > I had thought that some detailed settings would be associated with a given conformance.
> > We have examples, where we use a checkbox to enable functionality, and then options are shown to manage.
> > 
> > Many of the formats have their own options, kind of like the GML and CSV settings:
> > 
> > 

I'm a bit lost here... you mean, like having a checkbox to enable a class like "CQL2" and then having a selector to choose which palette to enable? Yeah, it does not fully work that way. For example, one cannot disable the entire features group, but disabling "filtering" inside it makes the CQL2 conformance irrelevant, but so is keeping filtering enabled and disabling both cql2-text and cql2-json, because then you don't have an actual language syntax to use, then whether we use functions or not makes no difference.
 It's  more like a complex dependency graph, where each conformance class can depend on other conformance classes. E.g. "functions" should be enabled if "features/filtering" is enabled, and at least one of "cql2-text" or "cql2-json", which is difficult to describe. 
A simple listing works better, and the admin has to add the "soft knowledge" about how the various classes relate to each other.

jodygarnett:
Andrea I am going to take larger discussion off to a document for when we have time.

The part I cannot understand for your existing table is how "default" works. I expected some experimental functionality (DRAFT_STANDRDS) to be disabled, but available for folks to try out if they enable it.

src/extension/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/FeatureConformance.java
jodygarnett: ``/ this is not a recognized conformance class though??``

That is correct, it is more about the output format showing up in OGCAPI -Features and WFS, and would be used to enable/disable the format, and provide a heading to configure the GML3 options show presently known the WFS Settings page.

image

aaime: When creating custom conformance class you'll have to create a new unique URI that references to GeoServer.

aaime:
The GUI shows nulls as "undefined", it's a well known checkbox HTML state:

image

What it does not show, is what would happen as a consequence of a lack of choice, we could have a separate column for it.
Why is this important? Well for example CQL_JSON is currently marked as draft (by mistake) and so if not chosen, it would be off by default.
However, CQL_JSON is actually a full standard and fully supported, so it should switch from DRAFT_STANDARD to STANDARD... meaning it should be enabled by default.

If someone had made a choice for it in the past, or saved the checkboxes with unchecked state, they would have to go and enable it manually. If they left it untouched (undefined) then upgrading CQL_JSON would automatically enable itself.
Which is what I would expect for any extension that switches from draft to final no?

One thing that is missing, in this regard, is that clicking on the checkbox switches it from undefined to checked/unchecked, but one cannot go back to undefined... which is something we probably want.

aaime:
@jodygarnett GUI updated based on your feedback:

image

Docs updated accordingly

PR #8665, opened 2025-07-08
Add SortedPropertiesWriter and switch all properties writing to this

petersmythe:
# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/impl/AbstractAccessRuleDAO.java
+++ b/src/main/src/main/java/org/geoserver/security/impl/AbstractAccessRuleDAO.java
@@ -27,6 +27,7 @@ import org.geoserver.platform.resource.Resource.Lock;
 import org.geoserver.platform.resource.Resource.Type;
 import org.geoserver.security.PropertyFileWatcher;
 import org.geoserver.util.IOUtils;
+import org.geoserver.util.SortedPropertiesWriter;
 import org.geotools.util.logging.Logging;
 
 /**
@@ -132,13 +133,12 @@ public abstract class AbstractAccessRuleDAO<R extends Comparable<R>> {
 
     /** Writes the rules back to file system */
     public synchronized void storeRules() throws IOException {
-        // turn back the users into a users map
         Properties p = toProperties();
 
         // write out to the data dir
         Resource propFile = securityDir.get(propertyFileName);
         try (OutputStream os = propFile.out()) {
-            p.store(os, null);
+            SortedPropertiesWriter.storeSorted(p, os, null);
             lastModified = System.currentTimeMillis();
             // avoid unnecessary reloads, the file just got fully written
             if (watcher != null) watcher.setKnownLastModified(lastModified);
--- a/src/main/src/main/java/org/geoserver/util/SortedPropertiesWriter.java
+++ b/src/main/src/main/java/org/geoserver/util/SortedPropertiesWriter.java
@@ -1,0 +1,104 @@
+/* (c) 2025 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2013 OpenPlans
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.util;
+
+import java.io.BufferedWriter;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.nio.charset.StandardCharsets;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Locale;
+import java.util.Properties;
+import java.util.TimeZone;
+import java.util.TreeSet;
+
+public class SortedPropertiesWriter {
+
+    /**
+     * Stores the given properties to the output stream in sorted key order, with a properties-style timestamp header
+     * and Unicode escaping, using UTF-8 encoding. The output will be in the format: #Mon Aug 26 10:16:17 SAST 2024
+     *
+     * @param props the Properties to write
+     * @param out the OutputStream to write to (not closed)
+     * @param comment an optional single-line comment to add at the top (without #)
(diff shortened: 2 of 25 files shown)


src/main/src/main/java/org/geoserver/security/impl/AbstractAccessRuleDAO.java
```diff
@@ -138,7 +139,7 @@ public synchronized void storeRules() throws IOException {
         // write out to the data dir
         Resource propFile = securityDir.get(propertyFileName);
         try (OutputStream os = propFile.out()) {
-            p.store(os, null);
+            SortedPropertiesWriter.storeSorted(p, os, null);
```
groldan, line 142: I'm afraid for `AbstractAccessRuleDAO` we can't assume storing the properties in sorted order is inocuous.
For instance, `RESTAcessRulesDAO` is key'ed by URI paths, and hence the order matters. If you change the order then the access rules will be applied in a different order than intended, potentially breaking the spring security access rules. At load time it uses `PropertyFileWatcher.LinkedProperties` to preserve the order defined in the .properties file.

src/main/src/main/java/org/geoserver/util/SortedPropertiesWriter.java
```diff
@@ -0,0 +1,104 @@
...
+import java.io.OutputStreamWriter;
+import java.nio.charset.StandardCharsets;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Locale;
+import java.util.Properties;
+import java.util.TimeZone;
+import java.util.TreeSet;
+
+public class SortedPropertiesWriter {
```
groldan, line 20: I wonder how can we reduce the dissonance between `SortedPropertiesWriter` and `PropertyFileWatcer.LinkedProperties` 

The latter **Is a** `java.util.Properties` that is uses to preserve the order of properties on `load()` and `save()`. Also, `Properties` already handles escaping of special/unicode characters.
Maybe we could "pull up" `PropertyFileWatcer.LinkedProperties`  to `org.geoserver.util.LinkedProperties`, add a `sorted():LinkedProperties` method that returns a sorted copy (Or a `static LinkedProperties sorted(Properties)` factory method), and let the normal workflow take place?

Still we need to let each `AbstractAcessRuleDAO` provide the `Properties` instance that works better for each implementation. They implement a `protected Properties toProperties()` method, so instead of having `AbstractAcessRuleDAO` decide the properties should be stored in sorted order, the concrete implementations could return a `LinkedProperties.sorted()` when possible, but not in the case of `RESTAccessRulesDAO`

groldan (changes requested):
I

src/main/src/main/java/org/geoserver/security/impl/AbstractAccessRuleDAO.java
jodygarnett: @groldan I think you mean to reject this PR correct?

groldan: not reject but requested changes for AbstractAccessRuleDAO not to assume it's ok to store the properties sorted,  not duplicate/replicate the escaping of values, and have synergy with PropertyFileWatcher.LinkedProperties

petersmythe: Hi @groldan you would know better than I, but I can't see how the property order matters for this instance https://github.com/geoserver/geoserver/blob/574a566de9fcf1d21c01b014e6e9096f026df06f/src/main/src/main/java/org/geoserver/security/impl/RESTAccessRuleDAO.java#L60, nor does the documentation https://docs.geoserver.org/main/en/user/security/rest.html mention anything about it being order dependent.  Another problem that I have is that with the current `p.store` implementation, the properties are [not guaranteed](https://www.google.com/search?q=java+properties+store+order) to be written to file in any order, so I don't understand how it could ever work, if it was order dependent.
Could you please guide my understanding, here?

groldan: well it's subtle, RESTAccessRuleDAO has
```java
    @Override
    protected void loadRules(Properties props) {
        // no concurrent version of LinkedHashMap, making it synchronized to preserve rule order
...
```
called by `AbstractAccessRuleDAO.checkPropertyFile()`, which passes it a `PropertyFileWatcher.LinkedProperties`, which preserves order.

Now, `RESTAccessRulesDATO.toProperties()` indeed returns a normal Properties, so that might be a bug. 

Truth is, in `rest.properties` it is not the same to have

```
/rest/workspaces/**;GET,HEAD,OPTIONS=ROLE_WS
/**;POST,DELETE,PUT=ROLE_ADMINISTRATOR
/**;GET,HEAD,OPTIONS=ROLE_ADMINISTRATOR
```

Than

```
/**;POST,DELETE,PUT=ROLE_ADMINISTRATOR
/**;GET,HEAD,OPTIONS=ROLE_ADMINISTRATOR
/rest/workspaces/**;GET,HEAD,OPTIONS=ROLE_WS
```

In the latter, the `ROLE_WS` rule will never be applied, the `ROLE_ADMINISTRATOR` rule is applied first. In the end this list will contribute to `GeoServerSecurityInterceptorFilter.RoleAuthorizationManager.check(Supplier<Authentication> authentication, HttpServletRequest request) -> vote()`, that will short circuit to `ACCESS_DENIED` when it had to be `ACCESS_GRANTED`

groldan: Aside, we have the same constraints for this PR which I intend to get ready asap:
- https://github.com/geoserver/geoserver/pull/7803

petersmythe:
Feedback from @groldan incorporated into new attempt: https://github.com/geoserver/geoserver/pull/9076

PR #8676, opened 2025-07-10
[GEOS-11889] Add prefix to smart data loader entities generation

fernandor777:
JIRA ticket:
https://osgeo-org.atlassian.net/browse/GEOS-11889

# Checklist


petersmythe:
Should 2.27.2 incorporate this change?  Ref: https://discourse.osgeo.org/t/geoserver-2-27-2-release-train/148298

src/community/smart-data-loader/src/test/java/org/geoserver/smartdataloader/postgis/PostGisPrefixAppSchemaVisitorTest.java
```diff
@@ -0,0 +1,16 @@
+package org.geoserver.smartdataloader.postgis;
```
aaime, line 5: Copyright header is missing in new files (reporting here, please check them all)

doc/en/user/source/community/smart-data-loader/data-store.rst
```diff
@@ -454,3 +454,8 @@ The Smart Override rules are defined using the following parameters on the ``Add
 .. figure:: images/smart-loader-overrides.png
 
   Smart Overrides configuration panel
+
+Entities prefix
+---------------
+
+The ``Entities prefix`` parameter allows to generate all the mapped feature types adding the configured prefix text.  This is useful to generate an appschema mapping structure with shared feature types from other smart data loader store, but using a different root type and override rules.
```
aaime, line 461: I'm not sure I understand here.... if memory serves me right, in app-schema one should match the target namespace with the workspace URI (or something like that). Could you maybe clarify with an example?

aaime (changes requested):

doc/en/user/source/community/smart-data-loader/data-store.rst
fernandor777: Right, and this remains uinchanged.  We add an option to generate the app-schema entity names from databsae table names with a fixed prefix, example: ```stations``` table with a ```St2``` prefix will generate the ```St2Stations``` feature type.  Theses prefixes are not related to workspace namespaces.

src/community/smart-data-loader/src/test/java/org/geoserver/smartdataloader/postgis/PostGisPrefixAppSchemaVisitorTest.java
fernandor777: Right, added to the new added files.  I will create a new PR adding the missing headers for all the already existing files.

aaime:
Merged. I don't see backport labels, you don't need any?

PR #8693, opened 2025-07-24
Fix jwt-headers and basic auth when both are using the Authorization HTTP header

jurgispods:
When using both basic auth and the jwt-headers community extension (to allow for authentication with a previously obtained JWT token) in the filter chain, currently jwt-header throws an exception when both auth filters use the `Authorization` header and the user tries to authenticate with basic auth, because it fails to parse the `Authorization: Basic <b64encoded user:password>` header. This PR catches the exception and lets the filter chain authenticate the user.

# Checklist


jurgispods:
I consider this a small change without the need for signing the CLA. Let me know if you think otherwise and I will happily sign it.
Unfortunately, I don't seem to be able to log into the Geoserver JIRA instance. I've tried multiple times in the last few days, created a new account and tried Google SSO, but neither Chrome nor Firefox lets me log in (not even in an incognito window), there seems to be something broken with the site. If any of the maintainers could create a ticket on my behalf, that would be much appreciated.

davidblasby (approved):

PR #8697, opened 2025-08-01
[GEOS-11894] Propagate predefined WMS env variables to query hints

Mike7311:
Update RenderingVariablesCallback to propagate [WMS env variables](https://docs.geoserver.org/main/en/user/styling/sld/extensions/substitution.html?utm_source=chatgpt.com#predefined-variables) to query hints to be used with views.

# Checklist


src/wms/src/main/java/org/geoserver/wms/RenderingVariablesCallback.java
```diff
@@ -0,0 +1,108 @@
...
+    @Override
+    public Layer beforeLayer(WMSMapContent mapContent, Layer layer) {
+        if (layer instanceof FeatureLayer) {
+            @SuppressWarnings("unchecked")
+            Map<String, Object> virtualTableParamsMap =
+                    (Map<String, Object>) layer.getQuery().getHints().get(VIRTUAL_TABLE_PARAMETERS);
+            if (virtualTableParamsMap == null) {
+                virtualTableParamsMap = new HashMap<>();
+                layer.getQuery().getHints().put(VIRTUAL_TABLE_PARAMETERS, virtualTableParamsMap);
+            }
```
aaime: Problematic code, it should be using layer.setQuery instead.

Rationale, the code in Layer just returns Query.ALL, which is immutable (returns a new hints object every time), while the version in FeatureLayer does this:

```java
    /**
     * Returns the definition query (filter) for this layer. If no definition query has been defined {@link Query.ALL}
     * is returned.
     *
     * @return Query used to process content prior to display, or Query.ALL to indicate all content is used
     */
    @Override
    public Query getQuery() {
        if (query == null) {
            return Query.ALL;
        } else {
            return query;
        }
    }
```

This one will work, but only if the layer has a Query object already set.
Instead, please grab the Query, use the Query(Query other) copy constructor, alter the hints and then set the new query back into the Layer.

src/wms/src/test/java/org/geoserver/wms/map/RenderingVariablesCallbackTest.java
```diff
@@ -0,0 +1,106 @@
...
+import org.geotools.feature.DefaultFeatureCollection;
+import org.geotools.filter.function.EnvFunction;
+import org.geotools.map.FeatureLayer;
+import org.geotools.styling.StyleBuilder;
+import org.geotools.util.factory.Hints;
+import org.junit.Before;
+import org.junit.Test;
+
+@SuppressWarnings("unchecked")
+public class RenderingVariablesCallbackTest extends WMSTestSupport {
```
aaime: Does this really need to use the WMSTestSupport machinery? Looks like it's mostly unit testing.

aaime: I would also like to see a functional test. It seems hard, but it does not need to be.

Take this example setting up a sql view on H2, from the wfs module:
https://github.com/geoserver/geoserver/blob/f79b907118907728d9604be2d02ddf75bf5a9230/src/wfs/src/test/java/org/geoserver/wfs/SQLViewTest.java#L45

Suggested changes:
* Use CiteTestData.LAKES, or whatever other polygon layer already available
* Use a SQL view that just returns the injected parameters in the select, in addition to the table attributes
* Run a GetFeatureInfo on it, it should contain the values from the GetMap

This will also ensure that the whole machinery is working as expected, end to end.

src/wms/src/main/java/org/geoserver/wms/RenderingVariablesCallback.java
```diff
@@ -0,0 +1,108 @@
...
+            setValueIfNotPresent(virtualTableParamsMap, localValues, WMS_SCALE_DENOMINATOR);
+        }
+
+        return layer;
+    }
+
+    private static void setValueIfNotPresent(
+            Map<String, Object> virtualTableParamsMap, Map<String, Object> localValues, String key) {
+        if (!virtualTableParamsMap.containsKey(key) && localValues.containsKey(key.toUpperCase())) {
+            virtualTableParamsMap.put(key, localValues.get(key.toUpperCase()));
```
aaime: It seems you haven't actually tested this change interactively... unless I'm missing something.
See the code expanding the variables down in GeoTools:

https://github.com/geotools/geotools/blob/afdeb4f33495eab8bbbc2ee0785222cfda881b4e/modules/library/jdbc/src/main/java/org/geotools/jdbc/VirtualTable.java#L190

The code expects a Map<String, String>, virtual table parameters just does a string expansion. With this code, I would expect to see a class cast exception when it does:

```java
            String value = values.get(param.getName());
```

aaime (changes requested):
I believe it needs some changes to work, see below.

src/wms/src/test/java/org/geoserver/wms/map/RenderingVariablesCallbackTest.java
Mike7311: So to clarify:
1) Create a virtual table with such parameters for Lakes layer
2) Run the WMS GetMap with env params provided and ensure the response contains feature (that is env params were propagated).
3) Run the GetFeatureInfo, but with usual parameters and ensure it contains feature.
Is my understanding correct?

aaime: Not really need to be unusual... just have the virtual table do something like:

select name, geom, %wms_width% as width, %wms_height% as height, ...

and ensure they are getting replaced with the values in the request. Please perform both the functional test, and an interactive test in the geoserver gui + wms request.

src/wms/src/main/java/org/geoserver/wms/RenderingVariablesCallback.java
Mike7311: Changed

src/wms/src/test/java/org/geoserver/wms/map/RenderingVariablesCallbackTest.java
Mike7311: Changed

src/wms/src/main/java/org/geoserver/wms/RenderingVariablesCallback.java
Mike7311: Yeah. Got a bit confiused with map being used with integers, but turns out it is in dggs module.

src/wms/src/main/java/org/geoserver/wms/RenderingVariables.java
```diff
@@ -21,18 +30,34 @@ public class RenderingVariables {
...
     /** The GetMap height, as an {@link Integer} */
-    private static final String WMS_HEIGHT = "wms_height";
+    public static final String WMS_HEIGHT = "wms_height";
     /** The GetMap scale denominator, as an {@link Double} */
-    private static final String WMS_SCALE_DENOMINATOR = "wms_scale_denominator";
+    public static final String WMS_SCALE_DENOMINATOR = "wms_scale_denominator";
+
+    public static final String WMS_CRS = "wms_crs";
+
+    public static void setupEnvironmentVariables(FeatureInfoRequestParameters params) {
```
aaime, line 47: Javadoc for public methods please

src/wms/src/main/java/org/geoserver/wms/RenderingVariables.java
```diff
@@ -21,18 +30,34 @@ public class RenderingVariables {
...
+        EnvFunction.setLocalValue(WMS_BBOX, params.getRequestedBounds());
+        EnvFunction.setLocalValue(WMS_SRS, params.getGetMapRequest().getSRS());
+        EnvFunction.setLocalValue(WMS_WIDTH, params.getWidth());
+        EnvFunction.setLocalValue(WMS_HEIGHT, params.getHeight());
+        try {
+            double scaleDenominator = params.getScaleDenominator();
+            EnvFunction.setLocalValue(WMS_SCALE_DENOMINATOR, scaleDenominator);
+        } catch (Exception e) {
+            LOGGER.log(
```
aaime: How can this one fail? In the getter I see:

```
    /** The scale denominator of the requested map */
    public double getScaleDenominator() {
        return scaleDenominator;
    }
```

src/wms/src/main/java/org/geoserver/wms/RenderingVariables.java
```diff
@@ -21,18 +30,34 @@ public class RenderingVariables {
...
+                    Level.SEVERE,
+                    "Failed to compute the scale denominator, wms_scale_denominator env variable is unset",
+                    e);
+        }
+    }
 
     public static void setupEnvironmentVariables(WMSMapContent mapContent) {
-        // setup some SLD variable substitution environment used by rendering transformations
         EnvFunction.setLocalValue(WMS_BBOX, mapContent.getRenderingArea());
-        EnvFunction.setLocalValue("wms_crs", mapContent.getRenderingArea().getCoordinateReferenceSystem());
```
aaime, line 35: Why is this gone? It's actually used by some rendering transformations.

src/wms/src/main/java/org/geoserver/wms/RenderingVariablesCallback.java
```diff
@@ -0,0 +1,51 @@
+/* (c) 2014 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2014 OpenPlans
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
```
aaime:
```suggestion
/* (c) 2025 Open Source Geospatial Foundation - all rights reserved
 * This code is licensed under the GPL 2.0 license, available at the root
 * application directory.
 */
```

src/wms/src/main/java/org/geoserver/wms/RenderingVariablesCallback.java
```diff
@@ -0,0 +1,51 @@
...
+import static org.geoserver.wms.RenderingVariables.setQueryHintsFromEnv;
+
+import org.geotools.filter.function.EnvFunction;
+import org.geotools.map.FeatureLayer;
+import org.geotools.map.Layer;
+
+/**
+ * Helper that injects enviroment variables in the {@link EnvFunction} given a map context
+ *
+ * @author Andrea Aime - GeoSolutions
```
aaime: I don't remember writing this code? :rofl:

src/wms/src/test/java/org/geoserver/wms/map/RenderingVariablesCallbackTest.java
```diff
@@ -0,0 +1,108 @@
+/*
+ *    GeoTools - The Open Source Java GIS Toolkit
+ *    http://geotools.org
+ *
```
aaime: This is GeoServer, please use a GeoServer copyright header, e.g.:

```
/* (c) 2025 Open Source Geospatial Foundation - all rights reserved
 * application directory.
 */
```

src/wms/src/test/java/org/geoserver/wms/map/WMSEnvParamsToViewPropagationTest.java
```diff
@@ -0,0 +1,138 @@
+/*
```
aaime: Still GeoServer....

src/wms/src/test/java/org/geoserver/wms/map/RenderingVariablesCallbackTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+@SuppressWarnings("unchecked")
+public class RenderingVariablesCallbackTest {
+
+    private static final RenderingVariablesCallback callback = new RenderingVariablesCallback();
+
+    @Before
+    public void init() {
+        EnvFunction.setLocalValue(WMS_BBOX, ReferencedEnvelope.rect(-180, -90, 10, 5));
+        EnvFunction.setLocalValue(WMS_CRS, "EPSG:4836");
```
aaime: WMS_CRS is supposed to be a CoordinateReferenceSystem object.

src/wms/src/test/java/org/geoserver/wms/map/RenderingVariablesCallbackTest.java
```diff
@@ -0,0 +1,108 @@
...
+        EnvFunction.setLocalValue(WMS_SRS, "EPSG:4836");
+        EnvFunction.setLocalValue(WMS_WIDTH, "5");
+        EnvFunction.setLocalValue(WMS_HEIGHT, "4");
+        EnvFunction.setLocalValue(WMS_SCALE_DENOMINATOR, "1_000_000");
```
aaime: Why are these strings? The env function takes object. Please do not confuse the capabiliites of EnvFunction with the VirtualTable parameters. It should work with env using the right type of object, and the code injecting in the virtual table turning them into strings.

src/wms/src/test/java/org/geoserver/wms/map/RenderingVariablesCallbackTest.java
```diff
@@ -0,0 +1,108 @@
...
+        wmsMapContent.addLayer(layer);
+        callback.beforeRender(wmsMapContent);
+
+        Map<String, String> virtualParams =
+                (Map<String, String>) layer.getQuery().getHints().get(Hints.VIRTUAL_TABLE_PARAMETERS);
+        assertEquals("-180.000000,-90.000000,-170.000000,-85.000000", virtualParams.get(WMS_BBOX));
+        assertEquals("EPSG:4836", virtualParams.get(WMS_SRS));
+        assertEquals("5", virtualParams.get(WMS_WIDTH));
+        assertEquals("4", virtualParams.get(WMS_HEIGHT));
+        assertEquals("1_000_000", virtualParams.get(WMS_SCALE_DENOMINATOR));
```
aaime: It would be pretty weird to find underscores in what is supposed to be a number (Java can parse them as separators, I cannot guarantee for databases).

src/wms/src/main/java/org/geoserver/wms/RenderingVariables.java
```diff
@@ -46,4 +71,53 @@ public static void setupEnvironmentVariables(WMSMapContent mapContent) {
                     e);
         }
     }
+
+    public static void setQueryHintsFromEnv(Layer layer) {
+        Query query = setHintsToQuery(layer.getQuery());
+        ((FeatureLayer) layer).setQuery(query);
+    }
```
aaime: Please have the method require a FeatureLayer, the signature seems to suggest it would work with any layer type. Please add javadoc.

src/wms/src/main/java/org/geoserver/wms/RenderingVariables.java
```diff
@@ -46,4 +71,53 @@ public static void setupEnvironmentVariables(WMSMapContent mapContent) {
...
+
+    public static Query setHintsToQuery(Query query) {
```
aaime, line 88: Add javadoc please

aaime (changes requested):

src/wms/src/main/java/org/geoserver/wms/RenderingVariables.java
Mike7311: added

src/wms/src/main/java/org/geoserver/wms/RenderingVariables.java
Mike7311: Just copy/paste issue. Removed try/catch from here.

src/wms/src/main/java/org/geoserver/wms/RenderingVariables.java
Mike7311: restored

src/wms/src/main/java/org/geoserver/wms/RenderingVariablesCallback.java
Mike7311: Added

src/wms/src/main/java/org/geoserver/wms/RenderingVariablesCallback.java
Mike7311: Yep, removed.

src/wms/src/test/java/org/geoserver/wms/map/RenderingVariablesCallbackTest.java
Mike7311: Updated.

src/wms/src/test/java/org/geoserver/wms/map/WMSEnvParamsToViewPropagationTest.java
Mike7311: Also updated

src/wms/src/test/java/org/geoserver/wms/map/RenderingVariablesCallbackTest.java
Mike7311: Changed to CRS.decode("EPSG:4326")

src/wms/src/test/java/org/geoserver/wms/map/RenderingVariablesCallbackTest.java
Mike7311: Changed to integers.

src/wms/src/test/java/org/geoserver/wms/map/RenderingVariablesCallbackTest.java
Mike7311: Removed underscores.

src/wms/src/main/java/org/geoserver/wms/RenderingVariables.java
Mike7311: Changed from Layer to FeatureLayer

src/wms/src/main/java/org/geoserver/wms/RenderingVariables.java
Mike7311: Added

doc/en/user/source/styling/sld/extensions/substitution.rst
```diff
@@ -70,7 +70,7 @@ The predefined variables are:
      - Either ``vector`` or empty
      - this variable gets set to ``vector`` when the kml generator is writing out vector features as placemarks, as opposed to ground overlays
 
-     
+Mentioned variables (with exception of kmlOutputMode) are also propagated to query hints for feature layers, in order to be used by SQL views.
```
aaime: wms_crs is not being propagated either

src/wms/src/main/java/org/geoserver/wms/RenderingVariables.java
```diff
@@ -46,4 +68,64 @@ public static void setupEnvironmentVariables(WMSMapContent mapContent) {
...
+            virtualTableParamsMap.put(key, (String) localValues.get(key.toUpperCase()));
+        }
+    }
+
+    private static void setValueIfNotPresent(
+            Map<String, String> virtualTableParamsMap,
+            Map<String, Object> localValues,
+            String key,
+            Function<Object, String> transformer) {
+        if (!virtualTableParamsMap.containsKey(key) && localValues.containsKey(key.toUpperCase())) {
```
aaime: Why touppsercase here? The local values have been set with the constant, and the map is case sensitive, how can they be modified in the meantime?

src/wms/src/main/java/org/geoserver/wms/RenderingVariables.java
```diff
@@ -46,4 +68,64 @@ public static void setupEnvironmentVariables(WMSMapContent mapContent) {
...
+            virtualTableParamsMap.put(key, transformer.apply(localValues.get(key.toUpperCase())));
```
aaime: Same here

aaime (changes requested):
Almost there

src/wms/src/main/java/org/geoserver/wms/RenderingVariables.java
Mike7311: Yes, they are set via constants. However in [EnvFunction.setLocalValues](https://github.com/geotools/geotools/blob/9a4ab7253c83397c70ab894ee6cd6a0eaf9c8f8d/modules/library/main/src/main/java/org/geotools/filter/function/EnvFunction.java#L178) the incoming keys are uppercased.

doc/en/user/source/styling/sld/extensions/substitution.rst
Mike7311: Mentioned it too

src/wms/src/main/java/org/geoserver/wms/RenderingVariables.java
aaime: Then add a clarifying comment for it. Put yourself in the place of who's going to read the code later, and does not have all the behavior of GT/GS classes memorized

Mike7311: Added comment explaining that.

src/wms/src/main/java/org/geoserver/wms/RenderingVariables.java
Mike7311: Also added comment regarding uppercase.

aaime:

PR #8698, opened 2025-08-03
[GEOS-11814] Create a REST API for Filter Chains - refactor

afabiani:
This PR refactors the [GEOS-11814] - REST API for Filter Chains one.

1. `security/filterChains` renamed to `security/filterChain` according to the current GeoServer model
2. Removed wrappers around `filterChains` by exposing the real GeoServer Security model
    e.g.: http://localhost:8080/geoserver/rest/security/filterChain/web
    ```json
    <filters name="web" class="org.geoserver.security.HtmlLoginFilterChain" path="/web/**,/gwc/rest/web/**,/" disabled="false"     allowSessionCreation="true" ssl="false" matchHTTPMethod="false" interceptorName="interceptor" exceptionTranslationName="exception">
        <filter>rememberme</filter>
        <filter>form</filter>
        <filter>oidc-test</filter>
        <filter>anonymous</filter>
    </filters>
```
3. Introduced a protected path `security/filterChain/order` allowing to programmatically update the filter chain order.

image

# Checklist

### DIFF
--- a/src/restconfig/src/main/java/org/geoserver/rest/security/AuthenticationFilterChainRestController.java
+++ b/src/restconfig/src/main/java/org/geoserver/rest/security/AuthenticationFilterChainRestController.java
@@ -4,122 +4,634 @@
  */
 package org.geoserver.rest.security;
 
-import static com.google.common.base.Preconditions.checkArgument;
-import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.base.Preconditions.*;
 
 import com.google.common.base.Strings;
 import com.thoughtworks.xstream.XStream;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.lang.reflect.Type;
-import java.util.List;
-import java.util.Objects;
+import java.lang.reflect.Method;
+import java.nio.charset.StandardCharsets;
+import java.util.*;
 import java.util.stream.Collectors;
+import javax.servlet.http.HttpServletRequest;
 import org.geoserver.config.util.XStreamPersister;
+import org.geoserver.config.util.XStreamPersisterFactory;
 import org.geoserver.rest.RestBaseController;
-import org.geoserver.rest.converters.XStreamMessageConverter;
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/restconfig/src/main/java/org/geoserver/rest/security/AuthenticationFilterChainRestController.java [hunk 1]
```diff
@@ -5,121 +5,647 @@
...
     @GetMapping(
-            value = "/{chainName}",
-            produces = {
-                MediaType.APPLICATION_JSON_VALUE,
-                MediaType.APPLICATION_XML_VALUE,
-            })
-    public RestWrapper<AuthFilterChain> view(@PathVariable("chainName") String chainName) {
-        checkAuthorisation();
+            path = "",
+            produces = {MediaType.APPLICATION_XML_VALUE, MediaType.APPLICATION_JSON_VALUE})
```
aaime: Hum... still stating it supports JSON too. And it should be able to, like the rest of the REST API (JSON is not the best format, but having a REST API that does not support JSON is some parts is even more confusing IMHO).

aaime (commented):
The rest of the GeoServer REST API supports JSON too, this one should as well.
(and btw, I do agree that the RestWrappers are ugly to use, but they allow for on-the-fly configuration of the xstream overrides and HTML representations)

afabiani:
> The rest of the GeoServer REST API supports JSON too, this one should as well. (and btw, I do agree that the RestWrappers are ugly to use, but they allow for on-the-fly configuration of the xstream overrides and HTML representations)

I see, that makes sense. In that case I will refactor this one a bit in order to support JSON too.
Most probably I'll need to wrap the filter-chains and the mapping won't exactly match the original one.

aaime:
In order to support json all you need to do is to wrap the configuration objects in the appropriate RestWrapper. The configuration is stored as XML using xstream so you can check how the security manager is doing that. Let's talk about it if something is not clear.

afabiani:
@aaime thanks for your hints. I have updated the PR accordingly by adding JSON support too.

aaime:
Looks good to me.

PR #8747, opened 2025-08-28
[GEOS-11914] Using namespaces parameter with virtual services does not work

Mike7311:
[GEOS-11914](https://osgeo-org.atlassian.net/browse/GEOS-11914) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

Adding support for schema override using "namespaces" parameter to virtual services.

For example:
```
curl --request GET \
  --url 'http://localhost:8080/geoserver/topp/ows?service=WFS&version=2.0.0&request=GetFeature&typeName=vopp%3Astates&maxFeatures=50&outputFormat=application%2Fjson&namespace=xmlns(vopp%3Dhttp%3A%2F%2Fwww.openplans.org%2Ftopp)' \
  --header 'Authorization: Basic YWRtaW46Z2Vvc2VydmVy' \
  --header 'User-Agent: insomnia/11.0.0'
```

that is we can provide namespace=xmlns(vopp=http://www.openplans.org/topp)
and use typeName=vopp:states while issuing requests against /topp virtual service.
  

# Checklist

### DIFF
--- a/src/web/wfs/src/main/java/org/geoserver/wfs/web/WFSAdminPage.html
+++ b/src/web/wfs/src/main/java/org/geoserver/wfs/web/WFSAdminPage.html
@@ -136,10 +136,6 @@
           <input wicket:id="disableStoredQueriesManagement" type="checkbox"></input>
           <label for="disableStoredQueriesManagement"><wicket:message key="disableStoredQueriesManagement">Disable stored queries management</wicket:message></label>
         </li>
-        <li>
-          <input wicket:id="enabledSchemaOverrideForVirtualAPI" type="checkbox"></input>
-          <label for="enabledSchemaOverrideForVirtualAPI"><wicket:message key="enabledSchemaOverrideForVirtualAPI">Schema override for virtual service</wicket:message></label>
-        </li>
       </ul>
     </fieldset>
     <fieldset>
--- a/src/web/wfs/src/main/java/org/geoserver/wfs/web/WFSAdminPage.java
+++ b/src/web/wfs/src/main/java/org/geoserver/wfs/web/WFSAdminPage.java
@@ -187,7 +187,6 @@ public class WFSAdminPage extends BaseServiceAdminPage<WFSInfo> {
         // allowGlobalQueries checkbox
         form.add(new CheckBox("allowGlobalQueries"));
         form.add(new CheckBox("disableStoredQueriesManagement"));
-        form.add(new CheckBox("enabledSchemaOverrideForVirtualAPI"));
         form.add(new LocalesDropdown("defaultLocale", new PropertyModel<>(info, "defaultLocale")));
     }
 
--- a/src/web/wfs/src/main/resources/GeoServerApplication.properties
+++ b/src/web/wfs/src/main/resources/GeoServerApplication.properties
@@ -3,7 +3,6 @@ wfs.title=WFS
 
 WFSAdminPage.allowGlobalQueries = Allow Global Stored Queries
 WFSAdminPage.disableStoredQueriesManagement = Disable stored queries management
-WFSAdminPage.enabledSchemaOverrideForVirtualAPI = Enable schema override for virtual service
 WFSAdminPage.basic=Basic
 WFSAdminPage.canonicalSchemaLocation=Encode canonical WFS schema location
 WFSAdminPage.complete=Complete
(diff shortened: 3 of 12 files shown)


jratike80:
Would some XML data be read from the given URL after this PR, or would it offer some other possibility to bypass the security settings https://docs.geoserver.org/main/en/user/production/config.html#external-entities-resolution and https://docs.geoserver.org/main/en/user/security/urlchecks.html?

aaime:
This is not a "schema override"...
To understand better, what is a "schema" to start with? It is the description of the structure of a feature type, e.g. this is a schema (described using the XML schema language):

```
<?xml version="1.0" encoding="UTF-8"?><xsd:schema xmlns:gml="http://www.opengis.net/gml" xmlns:sf="http://www.openplans.org/spearfish" xmlns:xsd="http://www.w3.org/2001/XMLSchema" elementFormDefault="qualified" targetNamespace="http://www.openplans.org/spearfish">
  <xsd:import namespace="http://www.opengis.net/gml" schemaLocation="https://gs-main.geosolutionsgroup.com/geoserver/schemas/gml/2.1.2/feature.xsd"/>
  <xsd:complexType name="bugsitesType">
    <xsd:complexContent>
      <xsd:extension base="gml:AbstractFeatureType">
        <xsd:sequence>
          <xsd:element maxOccurs="1" minOccurs="0" name="the_geom" nillable="true" type="gml:PointPropertyType"/>
          <xsd:element maxOccurs="1" minOccurs="0" name="cat" nillable="true" type="xsd:long"/>
          <xsd:element maxOccurs="1" minOccurs="0" name="str1" nillable="true" type="xsd:string"/>
        </xsd:sequence>
      </xsd:extension>
    </xsd:complexContent>
  </xsd:complexType>
  <xsd:element name="bugsites" substitutionGroup="gml:_Feature" type="sf:bugsitesType"/>
</xsd:schema>

```

A schema override would change the structure of the schema, that's not what you're doing here.

So what's the deal with the namespaces parameter in the request?
Elements in a GML schema have a namespace. The namespace is the URI, it's ``http://www.openplans.org/spearfish``.
In the rest of the document, the namespace is referred to via a "prefix", which is normally declared at the beginning of the document, e.g., ``xmlns:sf=http://www.openplans.org/spearfish``.

The prefix is local to the document and can be almost anythying. e.g. ``xmlns:mickeyMouse=http://www.openplans.org/spearfish`` is totally valid, as long as the rest of the document uses it, e.g., ``mickeyMouse:the_geom``.

Outside of that one GML document, where the declaration takes place, the prefix has no meaning. Unfortunately, it's common convention to use well known prefixes, and GeoServer enforces it by using the workspace name as the prefix, making it look as if it was "important" and changing it in a XML context was "an override". It's not.

Specifically for your case, the WFS specification allows one to use the namespace prefixes declared in capabiliites document (the typename contains the prefix), but one can follow XML rules and use whatever prefix they want, by adding the namespaces KVP parameter.

Long story short, what you did is just a bugfix for the namespaces parameter handling in virtual service request:
* it's not a schema override
* it does not need a dedicated configuration

Mike7311:
> Long story short, what you did is just a bugfix for the namespaces parameter handling in virtual service request:
> 
> * it's not a schema override
> * it does not need a dedicated configuration

ok. Going to rename it.

Mike7311:
> Would some XML data be read from the given URL after this PR, or would it offer some other possibility to bypass the security settings https://docs.geoserver.org/main/en/user/production/config.html#external-entities-resolution and https://docs.geoserver.org/main/en/user/security/urlchecks.html?

As was mentioned by @aaime, this is just a fix for namespaces parameter not working on virtual services (name of the ticket already changed). It only allows to override prefix name, but not URL. Thus it is unlikely this will cause any security issues.

src/web/wfs/src/main/java/org/geoserver/wfs/web/WFSAdminPage.html
```diff
@@ -136,6 +136,10 @@
           <input wicket:id="disableStoredQueriesManagement" type="checkbox"></input>
           <label for="disableStoredQueriesManagement"><wicket:message key="disableStoredQueriesManagement">Disable stored queries management</wicket:message></label>
         </li>
+        <li>
```
aaime: Since it's just a bugfix, it does not need a dedicated configuration to enable it...

Mike7311: Configuration removed.
Initially it was treated as an functionality extension (that is support was namespaces parameter was not expected for virtual services).
Jira ticket description was also updated a bit.

aaime:

PR #8757, opened 2025-09-01
[GEOS-11917] INSPIRE configuration does not get properly saved when OGC API module is included

Mike7311:
[GEOS-11917](https://osgeo-org.atlassian.net/browse/GEOS-11917) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

During serialization of WFS config to XML GeoServer attempts to use converters from XStream, and if not found from GeoTools.
Matching for suitable converters in XStream is done by class, and by its superclasses.
This way generic converter from XStream may be used instead of dedicated converter from GeoTools.

In this case when both INSPIRE and OGCAPI modules are enabled there is an issue when serializing [UniqueResourceIdentifiers](https://github.com/geoserver/geoserver/blob/081911229c75d5d977b6e7647ef5e9624b4a916f/src/extension/inspire/src/main/java/org/geoserver/inspire/UniqueResourceIdentifiers.java#L15). This class extends ArrayList, and OGCAPI registers generic list converter. Thus it is being matched, and used instead of dedicated [SpatialDatasetIdentifiersConverter](https://github.com/geoserver/geoserver/blob/99ae245402fca51e306eaf9817faaa891bf89e73/src/extension/inspire/src/main/java/org/geoserver/inspire/InspireConverterFactory.java#L38).
A "backwardBriefIgnored" list was added to XStreamPersister. Classes from this list will not be matched against XStream converters.

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
+++ b/src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
@@ -262,16 +262,16 @@ public class XStreamPersister {
     /** Flag controlling how references to objects are encoded. */
     boolean referenceByName = false;
 
-    /** The type map used in {@link BreifMapConverter} to handle complex objects */
-    Map<String, Class<?>> forwardBreifMap = new HashMap<>();
+    /** The type map used in {@link BriefMapConverter} to handle complex objects */
+    Map<String, Class<?>> forwardBriefMap = new HashMap<>();
 
-    Map<Class<?>, String> backwardBreifMap = new HashMap<>();
+    Map<Class<?>, String> backwardBriefMap = new HashMap<>();
 
     /**
      * List containing classes that should be ignored by XStream. This is required because sometimes generic converter
      * may be registered in XStream, which will be used instead of dedicated converter from GeoTools
      */
-    List<Class<?>> backwardBreifIgnored = new ArrayList<>();
+    List<Class<?>> backwardBriefIgnored = new ArrayList<>();
 
     private Level forceLevel = LOGGER.getLevel() == null ? Level.INFO : LOGGER.getLevel();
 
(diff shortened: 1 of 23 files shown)


src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
```diff
@@ -267,6 +267,12 @@ protected void postEncodeWMTSStore(
 
     Map<Class<?>, String> backwardBreifMap = new HashMap<>();
 
+    /**
+     * List containing classes that should be ignored by XStream. This is required because sometimes generic converter
+     * may be registered in XStream, which will be used instead of dedicated converter from GeoTools
+     */
+    List<Class<?>> backwardBreifIgnored = new ArrayList<>();
```
mprins: this seems to be a typo
```suggestion
    List<Class<?>> backwardBriefIgnored = new ArrayList<>();
```

aaime:
@mprins unfortunately the whole class uses "breif" instead of "brief", including it its public methods. I'd rather keep things consistent. Plan B would be to fix all the typos, and keep the old methods as deprecated.

mprins:
@aaime Yes, but it is a new class member/variable added in this PR, so now is a good time to do it right

src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
```diff
@@ -541,15 +547,19 @@ protected void init(XStream xs) {
...
     /**
      * Use this method to register complex types that cannot be simply represented as a string in a
-     * {@link BreifMapConverter}. The {@code typeId} will be used as a type discriminator in the brief map, as well as
+     * {@link BriefMapConverter}. The {@code typeId} will be used as a type discriminator in the brief map, as well as
      * the element root for the complex object to be converted.
      */
-    public void registerBreifMapComplexType(String typeId, Class<?> clazz) {
-        forwardBreifMap.put(typeId, clazz);
-        backwardBreifMap.put(clazz, typeId);
+    public void registerBriefMapComplexType(String typeId, Class<?> clazz) {
```
aaime, line 564: Nice, but this is an API change, registering brief map converters is a common occurrence for plugin modules outside of the GeoServer codebase (as they have to add their own specific configuration in the metadata map).

Please add back also the old method name, deprecated, and delegating to the new method.

Mike7311: Added back

Mike7311:
Jira ticket description was updated a bit.

aaime:

PR #8774, opened 2025-09-10
OIDC: fix problems with content security policy

davidblasby:
This PR fixes problems with content security policy - it removes inline style and removse inline JS.

While trying to run the new OIDC, I noticed quite a bit of the web functionality wasn't working and I noticed content-security-policy errors in the console.

This is caused because there are inline-styles in the OIDC web module as well as inline-JS.

This PR removes these ;

1. `<style>` - replaced with the "signed" (nonce) version. 
2. `style=` - replaced with classes
3. JS was attached as a change listener to the checkbox elements.  this is replaced with  "signed" (nonce) version that runs when the page is loaded.

  

# Checklist

### DIFF
--- a/src/community/security/oidc/oidc-web/src/main/java/org/geoserver/web/security/oauth2/login/ToggleDisplayCheckboxBehavior.java
+++ b/src/community/security/oidc/oidc-web/src/main/java/org/geoserver/web/security/oauth2/login/ToggleDisplayCheckboxBehavior.java
@@ -36,7 +36,7 @@ public class ToggleDisplayCheckboxBehavior extends Behavior {
     public void bind(Component component) {
         super.bind(component);
 
-        //this is used during ajax calls (also to set it up)
+        // this is used during ajax calls (also to set it up)
         targetComponent.add(new ClassAttributeModifier() {
             @Override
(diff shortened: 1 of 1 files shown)


src/community/security/oidc/oidc-web/src/main/java/org/geoserver/web/security/oauth2/login/OAuth2LoginAuthProviderPanel.java
```diff
@@ -61,6 +67,20 @@ public class OAuth2LoginAuthProviderPanel
     /** Prefix of custom OIDC specific attributes */
     private static final String PREFIX_OIDC = "oidc";
 
+    public static String oidcPanelCSS;
+
+    static {
+        try {
+            oidcPanelCSS = CharStreams.toString(new InputStreamReader(
+                    OAuth2LoginAuthProviderPanel.class.getResourceAsStream(
+                            "/org/geoserver/web/security/oauth2/login/css/oidc.css"),
```
pmauduit, line 76: since you already are in the `org.geoserver.web.security.oauth2.login` package, maybe just the relative path `css/oidc.css` should do ?

PR #8784, opened 2025-09-15
oidc: dont require oidc-specific filter chain (use "web/")

davidblasby:
During testing, I noticed that the login/logout wasn't working.
This is because it was using a URL like `http://localhost:8080/geoserver/oauth2/authorization/oidc`.
The user would have to setup their own (new) filter chain so that this would be processed.

I've change the login/logout urls so they are under `web/` ie. `http://localhost:8080/geoserver/web/oauth2/authorization/oidc`.  I had to also modify the redirect-from-idp (i.e. with the code) and the logout.

I also noticed that it was taking the `service_account` scope from keycloak by default - this causes problems so I remove it in the Discovery Client.  This shouldn't cause any issues.

1. used "web/" endpoints so you don't need to create an entirely new filter chain.  This would have caused no end of problems for users.
2. remove  `service_account` scope
3. update intgration test case so it doesn't add the new filter chain

Next (not this PR), I am looking at the login-button.  Its not `Content-security-policy` compliant because the form response will redirect to the IDP - which isn't allowed.  I'll see if that's fixable by - but it buried deep inside wicket.

# Checklist

### DIFF
--- a/src/community/security/oidc/oidc-core/src/main/java/org/geoserver/security/oauth2/login/GeoServerOAuth2LoginFilterConfig.java
+++ b/src/community/security/oidc/oidc-core/src/main/java/org/geoserver/security/oauth2/login/GeoServerOAuth2LoginFilterConfig.java
@@ -121,7 +121,7 @@ public class GeoServerOAuth2LoginFilterConfig extends PreAuthenticatedUserNameFi
 
     private String redirectUri(String pRegId) {
         String lBase = baseRedirectUriNormalized();
-        return lBase + "login/oauth2/code/" + pRegId;
+        return lBase + "web/login/oauth2/code/" + pRegId;
     }
 
     private String createPostLogoutRedirectUri() {
@@ -157,7 +157,7 @@ public class GeoServerOAuth2LoginFilterConfig extends PreAuthenticatedUserNameFi
             return null;
         }
         String lBase = baseRedirectUriNormalized();
-        return lBase + "oauth2/authorization/" + lRegIds.get(0);
+        return lBase + "web/oauth2/authorization/" + lRegIds.get(0);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/community/security/oidc/oidc-core/src/main/java/org/geoserver/security/oauth2/login/GeoServerOAuth2LoginFilterConfig.java
```diff
@@ -121,7 +121,7 @@ public void calculateRedirectUris() {
 
     private String redirectUri(String pRegId) {
         String lBase = baseRedirectUriNormalized();
-        return lBase + "login/oauth2/code/" + pRegId;
+        return lBase + "web/login/oauth2/code/" + pRegId;
```
ianturton: can we pull `"web/login/oauth2/code/"` out as a constant?

src/community/security/oidc/oidc-core/src/main/java/org/geoserver/security/oauth2/login/GeoServerOAuth2LoginFilterConfig.java
```diff
@@ -157,7 +157,7 @@ public String getAuthenticationEntryPointRedirectUri() {
             return null;
         }
         String lBase = baseRedirectUriNormalized();
-        return lBase + "oauth2/authorization/" + lRegIds.get(0);
+        return lBase + "web/oauth2/authorization/" + lRegIds.get(0);
```
ianturton: and this one is a constant already `DEFAULT_AUTHORIZATION_REQUEST_BASE_URI`

ianturton (approved):
looks good to me

davidblasby:
@ianturton  - I updated based on your two comments.  However, I noticed that the oidc test cases were not working.  I am surprised that the build server's `-PcommunityRelease` wasn't running them.  At least it didn't seem to be running them.

jodygarnett:
> @ianturton - I updated based on your two comments. However, I noticed that the oidc test cases were not working. I am surprised that the build server's `-PcommunityRelease` wasn't running them. At least it didn't seem to be running them.

The build server does not run tests on community modules, they are not considered stable / finished. I have a proposal to change that, but no time to implement quite yet.

jodygarnett:
@ianturton's feedback is addressed

jodygarnett (approved):
Feedback has been addressed

PR #8792, opened 2025-09-17
[GEOS-11937] Advertise and Enforce Attribute Restrictions.

axl8713:
[GEOS-11937](https://osgeo-org.atlassian.net/browse/GEOS-11937) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

These changes introduce support for defining, advertising, and enforcing restrictions on feature type attributes in GeoServer. The goal is to improve data integrity and support external clients by:
 - Allowing administrators to configure restrictions on feature attributes (e.g., enumerations, min/max ranges).
 - Advertising these restrictions to clients via DescribeFeatureType using standard XML Schema restrictions.
 - Enforcing the restrictions during write operations (e.g., WFS-T, REST API).

# Checklist


axl8713:
Builds are failing due to missing GeoTools changes.

* https://github.com/geotools/geotools/pull/5359

jodygarnett:
close/reopen to rerun actions after https://github.com/geotools/geotools/pull/5359 merge

doc/en/user/source/data/webadmin/layers.rst
```diff
@@ -195,30 +195,71 @@ Vector layers have a list of the :guilabel:`Feature Type Details`. These include
...
+* The type of the new attribute using the :guilabel:`Type` field. The most common types are available in a drop-down on editing,
+  but it's possible to indicate any valid Java class, as long as GeoServer has a converter that goes from the value produced by the
+  :guilabel:`Source` expression to the target type (new converters can be plugged in with some Java programming).
+* The source for the new attribute, which will be computed based on the :guilabel:`Source` CQL expression in the respective field.
+* A description for the new attribute, which will be visible wherever the feature type is described.
+* The `nillability` of the new attribute, for example, making the attribute mandatory even if it's not in the data source, and vice-versa.
+
+.. figure:: img/data_layers_feature_customize_add_attribute.png
+
+   Add attribute dialog
```
jodygarnett, line 227: I *really* like this change, thank you for surfacing this FeatureType model functionality that was made so long ago for app-schema.

src/main/src/main/java/org/vfny/geoserver/global/GeoServerFeatureStore.java
```diff
@@ -61,6 +70,28 @@ public List<FeatureId> addFeatures(FeatureCollection<SimpleFeatureType, SimpleFe
         return store().addFeatures(fc);
     }
 
+    private void enforceFeatureRestrictions(FeatureCollection<SimpleFeatureType, SimpleFeature> fc) {
+        try (FeatureIterator<SimpleFeature> features = fc.features()) {
+            while (features.hasNext()) {
+                SimpleFeature feature = features.next();
+
+                for (PropertyDescriptor propertyDescriptor : schema.getDescriptors()) {
```
jodygarnett: There should be some GeoTools support class for performing these kind of checks also, are you making your own in order to throw a clear IllegalArgumentException below?

See documentation example: https://docs.geotools.org/latest/userguide/library/main/feature.html

```
for (PropertyDescriptor property : feature.getType().getAttributes() )) {
   Object value = feature.getAttribute( property.getName() );

   Types.validate( property, value );
}
```

It also has an example of checking by hand that covers the AttributeType, not only the AttributeType restrictions...
```
// check restrictions for this propertyType and all super types
  for(PropertyType type=propertyType; type !=null; type=propertyType.getSuper() ){
     for( Filter valid : type.getRestrictions() ){
          if( !valid.evaulate( value ) ){
              throw new Exception(
                  "Not a valid "+type.getName()+" values must be:"+valid
               );
          }
     }
  }
```

src/main/src/main/java/org/vfny/geoserver/global/GeoServerFeatureStore.java
```diff
@@ -96,29 +127,80 @@ public Transaction getTransaction() {
...
+
+        for (PropertyDescriptor propertyDescriptor : schema.getDescriptors()) {
+            Object value = updates.get(propertyDescriptor.getName().toString());
+            Class<?> binding = propertyDescriptor.getType().getBinding();
+
+            Object convertedValue = Converters.convert(value, binding);
+
+            if (isValueViolatingRestrictions(
+                    convertedValue, propertyDescriptor.getType().getRestrictions())) {
+                throw new IllegalArgumentException(String.format(
```
jodygarnett: As above, ``Types.validate( convertedValue, propertyDescriptor )`` should offer to check restrictions from super types and so on.

src/web/core/src/main/java/org/geoserver/web/data/resource/AttributeTypeInfoEditPanel.java
```diff
@@ -0,0 +1,473 @@
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
```
jodygarnett: I expect you were just copying an example file, or if this is new I can be 2025

src/web/core/src/main/java/org/geoserver/web/data/resource/AttributeTypeInfoEditPanel.java
```diff
@@ -0,0 +1,473 @@
...
+ * @author Alessandro Ricchiuti
+ */
+public class AttributeTypeInfoEditPanel extends Panel {
+
+    private static final long serialVersionUID = -4226325094883373205L;
+
+    enum RESTRICTION_TYPE {
+        NONE,
+        OPTIONS,
+        RANGE
```
jodygarnett, line 54: Idea for later: ``CQL`` for ad-hoc restrictions, that is how we restrict string length for example.

jodygarnett (approved):
I am approving this for milestone release, but have some questions about possible code duplication below.

My educated guess is you only wish to check against he restrictions that the user can edit, not any that have come in from a AttributeType super. The only place I could see that happening is if we run into one of the "built-in" GML classes like "pointProperty" where additional care is taken in GeoTools.

jodygarnett:
Rebase, to see if this passes now

jodygarnett:
Compile warnings after rebase:
```
[INFO] Compiling 268 source files with javac [debug parameters release 17] to target/classes
[WARNING] /Users/jgarnett/dev/geoserver/geoserver/src/wfs/src/main/java/org/geoserver/wfs/xml/FeatureTypeSchemaBuilder.java:[870,41] unchecked cast
  required: java.util.List<org.eclipse.xsd.XSDConstrainingFacet>
  found:    java.lang.Object
[WARNING] /Users/jgarnett/dev/geoserver/geoserver/src/wfs/src/main/java/org/geoserver/wfs/xml/RestrictionToXSDConstrainingFacetVisitor.java:[95,70] unchecked cast
  required: java.util.Collection<? extends org.eclipse.xsd.XSDConstrainingFacet>
  found:    java.lang.Object
[WARNING] /Users/jgarnett/dev/geoserver/geoserver/src/wfs/src/main/java/org/geoserver/wfs/xml/RestrictionToXSDConstrainingFacetVisitor.java:[94,25] unchecked cast
  required: java.util.ArrayList<org.eclipse.xsd.XSDConstrainingFacet>
  found:    java.lang.Object
[WARNING] /Users/jgarnett/dev/geoserver/geoserver/src/wfs/src/main/java/org/geoserver/wfs/xml/RestrictionToXSDConstrainingFacetVisitor.java:[113,44] unchecked cast
  required: java.util.ArrayList<org.eclipse.xsd.XSDConstrainingFacet>
  found:    java.lang.Object
[WARNING] /Users/jgarnett/dev/geoserver/geoserver/src/wfs/src/main/java/org/geoserver/wfs/xml/RestrictionToXSDConstrainingFacetVisitor.java:[123,44] unchecked cast
  required: java.util.ArrayList<org.eclipse.xsd.XSDConstrainingFacet>
  found:    java.lang.Object
[WARNING] /Users/jgarnett/dev/geoserver/geoserver/src/wfs/src/main/java/org/geoserver/wfs/xml/RestrictionToXSDConstrainingFacetVisitor.java:[143,44] unchecked cast
  required: java.util.ArrayList<org.eclipse.xsd.XSDConstrainingFacet>
  found:    java.lang.Object
[WARNING] /Users/jgarnett/dev/geoserver/geoserver/src/wfs/src/main/java/org/geoserver/wfs/xml/RestrictionToXSDConstrainingFacetVisitor.java:[158,44] unchecked cast
  required: java.util.ArrayList<org.eclipse.xsd.XSDConstrainingFacet>
  found:    java.lang.Object
````

File: src/wfs/src/main/java/org/geoserver/wfs/xml/FeatureTypeSchemaBuilder.java [hunk 6]
```diff
@@ -848,6 +857,23 @@ private XSDComplexTypeDefinition buildComplexSchemaContent(
...
+        restrictionsCache.forEach((element, restrictions) -> {
+            if (!restrictions.isEmpty()) {
+
+                XSDSimpleTypeDefinition simple = factory.createXSDSimpleTypeDefinition();
+                simple.setBaseTypeDefinition(element.getTypeDefinition().getSimpleType());
+                EList<XSDConstrainingFacet> facets = simple.getFacetContents();
+                restrictions.stream()
+                        .map(r -> (List<XSDConstrainingFacet>)
+                                r.accept(restrictionsVisitor, new ArrayList<XSDConstrainingFacet>()))
+                        .forEach(facets::addAll);
```
jodygarnett, line 870: This code is producing a warning, and failing the build QA build with ``-Werror``flag:

```
[WARNING] /Users/jgarnett/dev/geoserver/geoserver/src/wfs/src/main/java/org/geoserver/wfs/xml/FeatureTypeSchemaBuilder.java:[870,41] unchecked cast
  required: java.util.List<org.eclipse.xsd.XSDConstrainingFacet>
  found:    java.lang.Object
```

I am not sure how to avoid this can keep your nice lambda expression

src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
```diff
@@ -446,7 +444,12 @@ protected void init(XStream xs) {
         // AttributeTypeInfo
         xs.omitField(impl(AttributeTypeInfo.class), "featureType");
         xs.omitField(impl(AttributeTypeInfo.class), "attribute");
+        xs.registerLocalConverter(impl(AttributeTypeInfo.class), "binding", (SingleValueConverter)
```
aaime, line 447: This change has me worried, changing how we serialize existing fields may cause backwards incompatibilities. Why was this needed, and does it alter the XML being produced for configuration storage and REST API representation, compared to the stable series? (2.27.x?)

src/main/src/main/java/org/vfny/geoserver/global/GeoServerFeatureStore.java
aaime: Types.validate can be an avenue, but should not be called against Attribute/Property:

* It is inefficient, SimpleFeatureImpl does not keep these for efficiency, generates a wrapper on the fly instead, promoting heap garbage generation
* It does not try to perform a conversion (while the getters for the values do).

Types.validate(descriptor, value) seems like it might work instead.

src/main/src/main/java/org/vfny/geoserver/global/GeoServerFeatureStore.java
aaime: WFS-T only works for simple features, the super type they have is basically empty. But it's nice to reuse code, as long as it's doing the right thing (I believe that Types validation methods were written almost 20 years ago and haven't been used much since).

src/wfs/src/main/java/org/geoserver/wfs/xml/RestrictionToXSDConstrainingFacetVisitor.java
```diff
@@ -0,0 +1,301 @@
+package org.geoserver.wfs.xml;
```
aaime, line 5: Copyright header missing

src/wfs/src/main/java/org/geoserver/wfs/xml/RestrictionToXSDConstrainingFacetVisitor.java
```diff
@@ -0,0 +1,301 @@
...
+import org.geotools.api.filter.temporal.EndedBy;
+import org.geotools.api.filter.temporal.Ends;
+import org.geotools.api.filter.temporal.Meets;
+import org.geotools.api.filter.temporal.MetBy;
+import org.geotools.api.filter.temporal.OverlappedBy;
+import org.geotools.api.filter.temporal.TContains;
+import org.geotools.api.filter.temporal.TEquals;
+import org.geotools.api.filter.temporal.TOverlaps;
+
+public class RestrictionToXSDConstrainingFacetVisitor implements FilterVisitor {
```
aaime: Consider exteding AbstractFilterVisitor instead, it should result in much less boilerplate code in this class.

src/web/core/src/main/java/org/geoserver/web/data/resource/ClassTextField.java
```diff
@@ -4,73 +4,22 @@
...
-import org.locationtech.jts.geom.MultiPoint;
-import org.locationtech.jts.geom.MultiPolygon;
-import org.locationtech.jts.geom.Point;
-import org.locationtech.jts.geom.Polygon;
 
 /** An {@link AutoCompleteTextField} dealing with common attribute type bindings */
 class ClassTextField extends AutoCompleteTextField<Class> {
 
-    /** List of well known class names this component will auto-complete to */
-    private static final List<Class> BINDINGS = Arrays.asList(
```
aaime, line 35: Looks like an unnecessary refactor... is there any actual behavioral change?

Generally speaking, while we should avoid God classes (1000+ lines), we should also avoid having many tiny classes spread out, which hinders navigation. 

I've also asked a straight question to ChatGPT, to get an third point of view:

"Should this class be split? <source code here>":

> ### Industry practice
> * Small UI components with 1–2 tiny helper classes: keep nested.
> * “God classes” over ~200 LOC with multiple unrelated helpers: split into top-level classes in the same package.
>
> Here, the inner classes are clearly supportive, so the consensus would lean toward keeping them nested. Splitting would only add boilerplate without improving clarity.
```

File: src/wfs/src/main/java/org/geoserver/wfs/xml/FeatureTypeSchemaBuilder.java (hunk 6 above)
Comment (aaime) [reply src/wfs/src/main/java/org/geoserver/wfs/xml/FeatureTypeSchemaBuilder.java:870]: I guess a suppress warnings is in order

File: src/web/core/src/test/java/org/geoserver/web/data/resource/AttributeTypeInfoEditPanelTest.java [hunk 11]
```diff
@@ -0,0 +1,245 @@
+/* (c) 2014 - 2016 Open Source Geospatial Foundation - all rights reserved
+ * (c) 2001 - 2013 OpenPlans
```
aaime: Should be only 2025:

/* (c) 2025 Open Source Geospatial Foundation - all rights reserved
 * application directory.
 */

src/wfs/src/test/java/org/geoserver/wfs/v1_1/TransactionTest.java
```diff
@@ -1011,6 +995,175 @@ public void testInsertUnknownFeatureType() throws Exception {
...
+    public void testInsertWithRestrictions() throws Exception {
+
+        TestFeatureStoreContext testFeatureStoreContext = createTestFeatureStoreContext();
+        FeatureTypeInfo featureTypeInfo = testFeatureStoreContext.featureTypeInfo;
+
+        AttributeTypeInfo restrictedAttribute = new AttributeTypeInfoImpl();
+        restrictedAttribute.setName("radius");
+        restrictedAttribute.setRange(NumberRange.create(Math.E, Math.PI));
+
+        featureTypeInfo.getAttributes().add(restrictedAttribute);
```
aaime, line 1008: Is there anything cleaning the feature type across tests? Otherwise cross-test pollution is possible here (the same data dir is used for all test methods in the class, otherwise the tests would be too slow).

src/main/src/main/java/org/geoserver/catalog/AttributeTypeInfo.java
```diff
@@ -79,8 +81,16 @@ public interface AttributeTypeInfo extends Serializable {
...
+
+    void setRange(NumberRange<? extends Number> range);
+
+    List<Object> getOptions();
+
+    void setOptions(List<Object> options);
+
     /** The same as {@link #equals(Object)}, except doesn't compare {@link FeatureTypeInfo}s, to avoid recursion. */
-    boolean equalsIngnoreFeatureType(Object obj);
+    boolean equalsIgnoreFeatureType(Object obj);
```
aaime, line 99: This is an API change and will break backwards compatibility (e.g., plugins external to the codebase that rely on this method will stop working). 
Fixing the typo is not a problem if you don't need the changes on 2.27.x, but if you do, then this change will have to be removed in the 2.27.x backport (it's fine to break backwards compatibilty in the main branch instead, it already contains many of those due to the ImageN switch).

src/restconfig/src/main/java/org/geoserver/rest/catalog/FeatureTypeController.java
```diff
@@ -214,10 +214,16 @@ public ResponseEntity featureTypePost(
             boolean virtual = mdm != null && mdm.containsKey(FeatureTypeInfo.JDBC_VIRTUAL_TABLE);
 
             if (!virtual && !typeExists) {
```
aaime, line 216: While unrelated to these changes, I would add a test for restrictions, checking that they can be set up and read via the REST API, and validating the generated XML while we're at it.

doc/en/user/source/data/webadmin/layers.rst
axl8713: Glad to help!

src/main/src/main/java/org/vfny/geoserver/global/GeoServerFeatureStore.java
axl8713: Please, have a look at the changes in f14556a6c76c54b2216fba34b1f314d6f5cebbe8, where I switched to `Types.validate`.

src/main/src/main/java/org/vfny/geoserver/global/GeoServerFeatureStore.java
axl8713: Resolving this conversation; continuing in the discussion above about Types.validate and super-type restrictions.

src/wfs/src/main/java/org/geoserver/wfs/xml/RestrictionToXSDConstrainingFacetVisitor.java
axl8713: Fixed. Thanks!

src/web/core/src/main/java/org/geoserver/web/data/resource/AttributeTypeInfoEditPanel.java
axl8713: Fixed. Thanks!

src/web/core/src/main/java/org/geoserver/web/data/resource/AttributeTypeInfoEditPanel.java
axl8713: Good idea; this seems out of scope for this PR, so I suggest addressing it in a follow‑up.

src/main/src/main/java/org/geoserver/catalog/AttributeTypeInfo.java
axl8713: Noted, thanks.

src/wfs/src/test/java/org/geoserver/wfs/v1_1/TransactionTest.java
axl8713: Would this be sufficient? 

https://github.com/axl8713/geoserver/blob/69cdd33098270692aba2d50ca9038e827ab087e0/src/wfs/src/test/java/org/geoserver/wfs/v1_1/TransactionTest.java#L57

Recursively deleting the store and everything under it.

src/wfs/src/main/java/org/geoserver/wfs/xml/RestrictionToXSDConstrainingFacetVisitor.java
axl8713: Brilliant. Thanks!

src/web/core/src/test/java/org/geoserver/web/data/resource/AttributeTypeInfoEditPanelTest.java
axl8713: Fixed. Thanks!

src/wfs/src/test/java/org/geoserver/wfs/v1_1/TransactionTest.java
aaime: Yes

src/wfs/src/main/java/org/geoserver/wfs/xml/FeatureTypeSchemaBuilder.java
axl8713: Added suppress warnings annotation.

src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
axl8713: That line was just moved from several lines above to keep  it together with the AttributeTypeInfo serialization configuration.
The XML should not be affected.

src/web/core/src/main/java/org/geoserver/web/data/resource/ClassTextField.java
axl8713: I needed to create a new Wicket `ClassLabel` component to show the attribute binding in the table "description" column.
Since both components need the `ClassNameConverter` and the latter was an inner class of `ClassTextField`, it made good sense to extract it to a package-private class,  bringing also the bindings for which it will convert.

Please, let me know if you have suggestion on how to refactor this structure differently.
Thanks!

aaime: Ah, so it is reused. Fine then, you don't need to change anything else.

aaime:
I see that the CITE tests are not passing, and it's specifically failing WFS-T tests.
First of all, is the branch up to date compared to main? If not, I would recommend rebasing and force-pushing.

If that is not enough, How would the system react to a configuration where the the attribute customization is there, but the restrictions are not specified? You can find it here:
https://github.com/geoserver/geoserver/tree/main/build/cite/wfs11/citewfs-1.1/workspaces/sf/sf

Running the CITE tests is also possible locally, if you want to reproduce the failure, ping me and I'll give you a quick intro.

src/restconfig/src/main/java/org/geoserver/rest/catalog/FeatureTypeController.java
axl8713: Tests added. Thanks!

jodygarnett (commented):
Thanks, I do not have time to do another detailed review. If you have addressed feedback this should be good to go for next week 2.28.0 release.

src/main/src/main/java/org/vfny/geoserver/global/GeoServerFeatureStore.java
```diff
@@ -95,29 +128,71 @@ public Transaction getTransaction() {
...
+
+    private void enforceFeatureUpdateValidation(String[] attributeNames, Object[] attributeValues) {
+        for (int i = 0; i < attributeNames.length; i++) {
+            String name = attributeNames[i];
+            Object value = attributeValues[i];
+
+            AttributeDescriptor attributeDescriptor = schema.getDescriptor(name);
+
+            Class<?> binding = attributeDescriptor.getType().getBinding();
+            Object convertedValue = Converters.convert(value, binding);
```
aaime, line 181: One bit here. If one tries to update a numeric value to the value "abc", the conversion will fail, return null, which might even be acceptable for the attribute, and if not, it will result in an odd error message, "attribute xyz cannot be null").

Suggestion: if the value is not null, and the convertedValue is, throw a validation error right away?

src/restconfig/src/test/java/org/geoserver/rest/catalog/FeatureTypeControllerTest.java
```diff
@@ -943,4 +945,110 @@ public void testPostWithAttributes() throws Exception {
...
+    }
+
+    @Test
+    public void testPutWithRestrictedAttributes() throws Exception {
+        String xml = getAsString(BASEPATH + "/workspaces/sf/featuretypes/PrimitiveGeoFeature.xml");
+
+        // update feature type with restricted attributes
+        String xmlWithRestrictedAttributes = xml.replace(
+                        "<name>intProperty</name>",
+                        """
```
aaime, line 1000: Just an observation, when you backport mind that code will have to be java 11 compatible and text block will have to be converted to string concatenation.

src/wfs/src/main/java/org/geoserver/wfs/xml/RestrictionToXSDConstrainingFacetVisitor.java
```diff
@@ -0,0 +1,75 @@
...
+import org.geotools.api.filter.And;
+import org.geotools.api.filter.Filter;
+import org.geotools.api.filter.Not;
+import org.geotools.api.filter.Or;
+import org.geotools.api.filter.PropertyIsBetween;
+import org.geotools.api.filter.PropertyIsEqualTo;
+import org.geotools.filter.visitor.AbstractFilterVisitor;
+
+@SuppressWarnings("unchecked")
+public class RestrictionToXSDConstrainingFacetVisitor extends AbstractFilterVisitor {
```
aaime, line 36: Maybe add a comment that this visitor is expected to work only with the well known flter patterns (explain which ones) and won't to do right thing if a manual validation filter is added, e.g.. "att < 10 or att > 20 and not (add = 15 or add = 30)" will not be translated correctly

src/wfs/src/test/java/org/geoserver/wfs/v1_1/TransactionTest.java
```diff
@@ -1056,4 +1209,52 @@ private String xmlEntityExpansionLimitBody() {
...
+        } catch (IllegalArgumentException ex) {
+            /* ignore exception if schema does not exist */
+        }
+
+        SimpleFeatureTypeBuilder featureTypeBuilder = new SimpleFeatureTypeBuilder();
+        featureTypeBuilder.setName("bar");
+        featureTypeBuilder.add("name", String.class);
+        featureTypeBuilder.add("geom", Point.class);
+        //        featureTypeBuilder.nillable(false);
+        //        featureTypeBuilder.minOccurs(1);
```
aaime: Please remove commeted out code

aaime (changes requested):
Looking good, I've found a couple of small bits to be amended, then it's good for merge IMHO.

src/main/src/main/java/org/vfny/geoserver/global/GeoServerFeatureStore.java
axl8713: Thanks for pointing this out.
While I was trying to test this scenario though, I've noticed that this kind of check is carried out here https://github.com/geoserver/geoserver/blob/99ae245402fca51e306eaf9817faaa891bf89e73/src/wfs/src/main/java/org/geoserver/wfs/UpdateElementHandler.java#L185-L194

I've also added a test in TransactionTests to verify this.

A validation error is then thrown before even reaching that code. Will this happen with every datastore? Is it still worth to add the suggested check?

src/restconfig/src/test/java/org/geoserver/rest/catalog/FeatureTypeControllerTest.java
axl8713: Noted. Thanks!

src/wfs/src/test/java/org/geoserver/wfs/v1_1/TransactionTest.java
axl8713: Done. Thanks!

src/wfs/src/main/java/org/geoserver/wfs/xml/RestrictionToXSDConstrainingFacetVisitor.java
axl8713: Done. Thanks!

aaime:
Looking good, and merging. I'm not running the backport bot as you know you already have a 2.27.x branch for this, please open a PR manually for that branch

axl8713:
Backport to 2.27.x: https://github.com/geoserver/geoserver/pull/8826

jodygarnett:
I notice this new feature available for the 2.27.3 release, but was not announced in the 2.28.x release notes :P

Turns out it was only marked as resolved for 2.27.3 in Jira ... we should update the 2.28.0 blog post :)

PR #8788, opened 2025-09-17
DO NOT MERGE - Not working

davidblasby:
See comments in  elements.

# Checklist

- [ ] I have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/main/CONTRIBUTING.md).
- [ ] I have sent a [Contribution Licence Agreement](https://docs.geoserver.org/latest/en/developer/policies/committing.html) (not required for small changes, e.g., fixing typos in documentation).
- [ ] First PR targets the `main` branch (backports managed later; ignore for branch specific issues).

For core and extension modules:

- [ ] New unit tests have been added covering the changes.
- [ ] [Documentation](https://github.com/geoserver/geoserver/tree/main/doc/en/user/source) has been updated (if change is visible to end users).
- [ ] The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers).
- [ ] There is an issue in the [GeoServer Jira](https://osgeo-org.atlassian.net/browse/GEOS/summary) (except for changes that do not affect administrators or end users in any way).
- [ ] Commit message(s) must be in the form ``[GEOS-XYZWV] Title of the Jira ticket``.
- [ ] Bug fixes and small new features are presented as a single commit.
- [ ] Each commit has a single objective (if there are multiple commits, each has a separate JIRA ticket describing its goal).

The PR will be merged when all the build checks are green ([see automated QA checks](https://docs.geoserver.org/latest/en/developer/qa-guide/index.html)), there is a code committer review, and the checklist has been fulfilled.

PR #8798, opened 2025-09-23
OIDC roles

davidblasby:
This needs:

- https://github.com/geoserver/geoserver/pull/8788

See:

- https://github.com/geoserver/gs3/issues/45

This adds:
1. a new UX for ID, Access, and UserInfo role sources
2. adds role translation (from the role in IDP to a Geoserver role)
3. updates testcontainer test cases  

NOTE: this ux is very similar to the one in JWT-Header community module.

image

# Checklist

### DIFF
--- a/src/community/security/oidc/oidc-core/src/main/java/org/geoserver/security/oauth2/common/GeoServerOAuth2RoleResolver.java
+++ b/src/community/security/oidc/oidc-core/src/main/java/org/geoserver/security/oauth2/common/GeoServerOAuth2RoleResolver.java
@@ -5,7 +5,6 @@
 package org.geoserver.security.oauth2.common;
 
 import static java.lang.String.format;
-import static java.util.logging.Level.INFO;
 import static java.util.logging.Level.SEVERE;
 import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.toList;
@@ -18,7 +17,13 @@ import static org.geoserver.security.oauth2.login.GeoServerOAuth2ClientRegistrat
 import com.nimbusds.jose.JWSObject;
 import java.io.IOException;
 import java.text.ParseException;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.function.Supplier;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -184,7 +189,9 @@ public class GeoServerOAuth2RoleResolver implements RoleResolver {
                         lRoles = asStringList(JwtHeaderUserNameExtractor.getClaim(claims, lClaimName));
                     }
                 } catch (ParseException e) {
-                    LOGGER.log(INFO, "could not parse Access Token as JWT!"); // likely not a JWT
+                    LOGGER.log(
+                            SEVERE,
+                            "could not parse Access Token as JWT - Is the remote IDP responding with non-JWT Access Tokens!"); // likely not a JWT
                 }
             }
         }


src/community/security/oidc/oidc-core/src/main/java/org/geoserver/security/oauth2/common/GeoServerOAuth2RoleResolver.java
```diff
@@ -5,21 +5,20 @@
...
+import com.nimbusds.jose.JWSObject;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
-import java.util.Set;
+import java.text.ParseException;
+import java.util.*;
```
ianturton: we don't usually use * imports, in fact the QA checks should fail this.

src/community/security/oidc/oidc-core/src/main/java/org/geoserver/security/oauth2/common/GeoServerOAuth2RoleResolver.java
```diff
@@ -154,19 +162,37 @@ private Collection<GeoServerRole> getRolesFromAccessToken(ResolverParam pParam)
...
+                // we assume access token is a JWT and get the internal claims
+                JWSObject jwsObject = null;
+                try {
+                    jwsObject = JWSObject.parse(lAccessToken.getTokenValue());
+                    if (jwsObject != null) {
+                        Map<String, Object> claims = jwsObject.getPayload().toJSONObject();
+                        lRoles = asStringList(JwtHeaderUserNameExtractor.getClaim(claims, lClaimName));
+                    }
+                } catch (ParseException e) {
+                    LOGGER.log(INFO, "could not parse Access Token as JWT!"); // likely not a JWT
```
ianturton: Is this ok at INFO or should it be SEVERE?

ianturton (changes requested):

src/community/security/oidc/oidc-core/pom.xml
```diff
@@ -120,6 +120,12 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
```
afabiani, line 123: Uhm... we are introducing a dependency from another community module. I don't think this is a good idea. Can we envisage extracting a common set of utilities and maybe have the jwt-headers under the same module?

src/community/security/oidc/oidc-core/src/main/java/org/geoserver/security/oauth2/common/GeoServerOAuth2RoleResolver.java
davidblasby: intellij automatically does this - I will fix it.

src/community/security/oidc/oidc-core/src/main/java/org/geoserver/security/oauth2/common/GeoServerOAuth2RoleResolver.java
davidblasby: I will change it to severe - it means that they missconfigured.  I make a more obvious error message.

src/community/security/oidc/oidc-core/pom.xml
davidblasby: yes - thats how its already setup.

This is the `-util` jwt-headers module that is shared with the Geoserver and GeoNetwork implementations - both GN4 (spring-based) and GN5 (spring-boot based).  The Geoserver-specific code is in another module (same with the Geonetwork-specific code - but its in the GN4/GN5 projects).

davidblasby:
Thanks for review @ianturton and @afabiani - see you in the meeting tomorrow (thursday) evening.

src/community/security/oidc/oidc-core/pom.xml
jodygarnett: @afabiani with the goal of sharing with geonetwork also this common code cannot be moved into `gs-main` for example.

However we should note that this utility `jwt-headers-util` module should graduate alongside gs-sec-oidc-plugin or gs-sec-jwt-headers-plugin as required.

petersmythe: When testing 2.28.0 I get:

image

PR #8815, opened 2025-09-26
[GEOS-11947] Add the ability to skip numberMatched in STAC/OpenSearch for EO responses

aaime:
[GEOS-11947](https://osgeo-org.atlassian.net/browse/GEOS-11947) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

Optionally avoid computing number matched, just like in WFS 2.0

# Checklist

### DIFF
--- a/src/community/oseo/oseo-service/src/test/java/org/geoserver/opensearch/eo/SearchTest.java
+++ b/src/community/oseo/oseo-service/src/test/java/org/geoserver/opensearch/eo/SearchTest.java
@@ -177,7 +177,6 @@ public class SearchTest extends OSEOTestSupport {
     @Test
     public void testAllCollectionCountZero() throws Exception {
         MockHttpServletResponse response = getAsServletResponse("oseo/search?count=0&httpAccept=" + ENCODED_ATOM_MIME);
-        System.out.println(response.getContentAsString());
         assertEquals(AtomSearchResponse.MIME, response.getContentType());
         assertEquals(200, response.getStatus());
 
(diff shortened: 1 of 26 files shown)


src/community/oseo/oseo-service/src/test/java/org/geoserver/opensearch/eo/SearchTest.java
```diff
@@ -177,6 +177,7 @@ public void testAllCollection() throws Exception {
     @Test
     public void testAllCollectionCountZero() throws Exception {
         MockHttpServletResponse response = getAsServletResponse("oseo/search?count=0&httpAccept=" + ENCODED_ATOM_MIME);
+        System.out.println(response.getContentAsString());
```
dromagnoli: System.out here. Is it ok? or is it a leftover from some debug?

src/community/oseo/oseo-stac/src/main/java/org/geoserver/ogcapi/v1/stac/AbstractQueryResult.java
```diff
@@ -45,4 +47,12 @@ public Map<String, RootBuilder> getTemplateMap() {
     public void setTemplateMap(Map<String, RootBuilder> templateMap) {
         this.templateMap = templateMap;
     }
+
+    /**
+     * Indicates whether there is a next page of results, or <code>null</code> if unknown (check #numberMatched in that
```
dromagnoli: not sure to have understood correctly. Will that return false if unknown or will it return null?
In the latter case, this getter is using boolean primitiveType so it may throw an NPE.

dromagnoli (approved):
Looks good to me beside one small doubt about the hasNextPage return type.

src/community/oseo/oseo-service/src/test/java/org/geoserver/opensearch/eo/SearchTest.java
aaime: Whoops yes, removing

src/community/oseo/oseo-stac/src/main/java/org/geoserver/ogcapi/v1/stac/AbstractQueryResult.java
aaime: Right sorry, used to be fed a Boolean but now it's just primitive boolean (see the constructor), amending description and field.

PR #8821, opened 2025-09-29
[GEOS-11948] GeoServer update for JAI to ImageN rename of classes and constants

jodygarnett:
# Checklist


aaime:
Compared to the other PRs for the same topic, this one is pretty simple.

Let's have a look at the remaining JAI references in the docs:

```
> git grep -i JAI -- "**/*.rst" | wc -l
56
```

Still quite a number. The two things popping at my eyes are references to JAI-EXT, and the image mosaic usage of "USE_JAI_IMAGEREAD".... hum... that thing is configured, if we were to turn it into "USE_IMAGEN_IMAGEREAD" we'd need to have a backwards compatiblity layer.

When it comes to code instead:

```
> git grep -i JAI -- "**/*.java" | wc -l
313
```

* JAIInfo is ... probably here to stay? Changing it should be possibile if we want to , it would require backwards compatibility layers for configuration (xstream persistence and rest api). Topic for 3.0?
I guess we should at least document what JAI even is, since it's in there. 
* USE_JAI_IMAGEREAD pops up in very many places 
* The usual ``ImageN jaiVariableHere`` pattern and ParameterBlockJAI showing up here too

jodygarnett:
Refactoring ``JAIInfo`` seems to be fine with:
```
xs.alias("jai", ImageProcessingInfo.class);
```

aaime:
Added automated migration of the USE_JAI_IMAGEREAD and verified it works both reading and in the UI. The migration happens in memory and will be persisted the first time they save. This way old REST scripts should keep on working too.
That said, we might want to add migration instructions in the update page, as users might get annoyed by XML config files being modified piecemeal (mostly those that keep the data dir in version control and will notice this changes little by little)

PR #8827, opened 2025-10-06
[GEOS-11952] GeoServer 2.28.0 Release cleanup

jodygarnett:
[GEOS-11952](https://osgeo-org.atlassian.net/browse/GEOS-11952) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

- [x] clean up broken javadoc links and unused tags
- [x] windows installer java 11 -> java 17
- [ ] assembly definitions move to src/assembly folder structure
- [x] java 17 download link
  

# Checklist

### DIFF
--- a/src/release/installer/win/build/geoserver_winsetup.nsi
+++ b/src/release/installer/win/build/geoserver_winsetup.nsi
@@ -16,26 +16,26 @@
 !pragma warning disable 6020 ; disable warning about unused uninstaller script code
 
 ; Constants (VERSION should be updated by a script!)
-!define APPNAME "GeoServer"                                     ; application name
+!define APPNAME "GeoServer"                                       ; application name
 !ifndef VERSION
   !searchparse /file ..\source\VERSION.txt `version = ` VERSION   ; Read version from VERSION.txt
 !endif
-!define FULLNAME "${APPNAME} ${VERSION}"                        ; app name and version combined
-!define FULLKEY "${APPNAME}-${VERSION}"                         ; app name and version combined (delimited)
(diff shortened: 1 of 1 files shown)


jodygarnett:
note: We will need to check the windows installer during the release process

doc/en/developer/source/policies/community-modules.rst
```diff
@@ -287,24 +287,25 @@ Process
 
    *Extensions*
 
-   #. Create a new directory under ``release/extensions`` which matches the
+   #. Create a new directory under ``release/src/extensions`` which matches the
```
aaime, line 290: /src/ seems to be showing up from a previous commit in ths file?
Looks like it's causing most of the changes

src/README.md
```diff
@@ -2,45 +2,73 @@
...
+1. Recommended: https://sdkman.io
+   
+   ```
+   # list to determine latest Temurin JDK 17
+   sdk list java | grep "17.*-tem"
+
+   # Installing latest Temurin JDK 17 shown above
+   sdk install java 17.0.16-tem
+   sdk install maven
+   ```
```
aaime: Performance-wise, it's better to get the distribution own JDK, see for example:
https://github.com/adoptium/containers/issues/641

3x difference is pretty significant, I cannot stay behind this reccomendation. I would provide a more nuanced view:
* Using sdkman one can probably get the latest release even when the distribution goes cold
* Using the distribution (Ubuntu) own JDK provides significanly better JPEG encoding performance.

jodygarnett: That is an interesting thing to know, let me update the text. I just wanted it to not be out of date.

PR #8905, opened 2025-10-28
[GEOS-11634] gwc: possibility to invalidate cache through timestamp

NielsCharlier:
[GEOS-11634](https://osgeo-org.atlassian.net/browse/GEOS-11634) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)

  

# Checklist

### DIFF
--- a/doc/en/user/source/geowebcache/webadmin/defaults.rst
+++ b/doc/en/user/source/geowebcache/webadmin/defaults.rst
@@ -77,6 +77,11 @@ Enable Data Security
 
 Enables the :ref:`gwc_data_security` in the embedded GeoWebCache.
 
+Cache Validation Timestamp Property
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ 
+Specify a property inside the layer metadata that contains a timestamp for the data. If the timestamp indicates that the data has been updated since it was last cached, the cache is invalidated. 
+
 Metatiling threads count
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/src/gwc/src/main/java/org/geoserver/gwc/config/GWCConfig.java
+++ b/src/gwc/src/main/java/org/geoserver/gwc/config/GWCConfig.java
@@ -86,6 +86,8 @@ public class GWCConfig implements Cloneable, Serializable {
 
     private String lockProviderName;
 
+    private String cacheValidationProperty;
+
     // Set of cache warnings that would cause caching being skipped
     Set<DimensionWarning.WarningType> cacheWarningSkips;
 
@@ -541,5 +543,13 @@ public class GWCConfig implements Cloneable, Serializable {
 
     public void setCacheWarningSkips(Set<DimensionWarning.WarningType> cacheWarningSkips) {
         this.cacheWarningSkips = new LinkedHashSet<>(cacheWarningSkips);
+    }
+
+    public String getCacheValidationProperty() {
+        return cacheValidationProperty;
+    }
+
+    public void setCacheValidationProperty(String cacheValidationProperty) {
+        this.cacheValidationProperty = cacheValidationProperty;
     }
 }
--- a/src/gwc/src/main/java/org/geoserver/gwc/layer/GeoServerTileLayer.java
+++ b/src/gwc/src/main/java/org/geoserver/gwc/layer/GeoServerTileLayer.java
@@ -24,6 +24,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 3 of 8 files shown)


groldan:
In the jira ticket you say:

> we already keep a timestamp in the layer’s metadata that  contains the last update of the layer’s data. If a cached tile is older than this timestamp, it should then be invalidated.

who's "we"? I don't remember GeoServer storing such info in the layer's medatada, and if it did, wouldn't there be a prescribed property name already?

I see there's a global GWC config form to indicate the name of the metadata config property. It makes me think, though I might be wrong, that this is completely new as "we" don't currently set a metadata property with the last update of the cache?

Also, in trying to understand the requirement "If a cached tile is older than this timestamp, it should then be invalidated.", I think a tile may be older than the "last update" due to seeding vs re-seeding, and why would that draw the tile invalid?

If the requirement is "Set an absolute expiration time for tile cache" I get it, but I think the solution proposed has some problems:
From a user's point of view, configuring the name of the metadata property used to establish the cache deadline seems too odd, why would a regular user/admin care about which metadata property name is used?. For a core feature (it's a core feature since you're adding it to a core module), it just looks odd and confusing. How is the timestamp supposed to be configured on a per layer basis? by setting this metadata property through the rest api?
 
Being a core feature proposal I think it would make more sense to add a property to GeoServerTileLayerInfo and the corresponding per-layer configuration form component in the tile layer config page.

As a side note I wonder why's the config proeprty being read from the LayerInfo.getResourceInfo metadata and not the LayerInfo metadata itself (that is, generically from the PublishedInfo metadata map)

NielsCharlier:
@groldan I apologize for the lack of clarity, I failed to mention in the ticket that this feature particularly makes sense in combination with either the geoserver Metadata Extension or Taskmanager community module. The metadata module (a supported extension) allows you to add unlimited metadata fields to resources. You can define a schema of (nested) fields of various types and the extension builds a GUI tab  with text fields, drop boxes, etc. for the user to fill in. On top of this, it also has a function to keep a timestamp that is updated automatically. The taskmanager community module also has an automated task that timestamps layers automatically when you sync the data from an exterior source, and can synchronize this timestamp across different geoservers.

This kind of explains all of your questions. I did consider that the feature could be useful without those modules but in combination with other software that communicates with geoserver, since one can indeed manipulate metadata fields using the REST service. I leave the metadata as a configurable field in order to provide maximal flexibility and compatibility.

NielsCharlier:
@groldan with the explanation given, is there any chance of having this merged, perhaps with adjustments?

groldan:
Thanks @NielsCharlier, that context helps a lot, and I agree the underlying need is legitimate. My remaining concern is where the feature lives rather than what it does.

The PR puts a configuration component on the core GWC Settings page that is only meaningful if you've also installed the Metadata extension, the Taskmanager community module, or you're driving ResourceInfo.metadata via REST from some external system. For an admin who has none of those, the "Cache Validation Timestamp Property" field has no obvious meaning and silently does nothing if filled in, which I think is the "misleading and unrelated" aspect we should avoid in core.

The implementation also fixes to one specific invalidation semantic (absolute timestamp compared against `stObj.getCreated()`) into core GWC's fetch path.

Proposal: a small extension point in core, the feature in the extensions. Rather than teaching core about a configurable metadata property name, add a neutral "is this cached tile still good?" hook that any module can contribute to:

```java
  public interface CachedTileValidator {
      /** @return false to invalidate (force MISS), true to accept */
      boolean isValid(GeoServerTileLayer layer, ConveyorTile tile);
  }
```

`GeoServerTileLayer.tryCacheFetch` keeps owning the existing retrieve logic (so `expireCache`, meta-tile cache-only, and `CacheResult` bookkeeping don't leak out), and only adds a post-retrieve veto chain:

```java
  private boolean tryCacheFetch(ConveyorTile tile) {
      int expireCache = getExpireCache((int) tile.getTileIndex()[2]);
      if (expireCache == GWCVars.CACHE_DISABLE_CACHE) return false;
      try {
          if (!tile.retrieve(expireCache * 1000L)) return false;
          for (CachedTileValidator v : GeoServerExtensions.extensions(CachedTileValidator.class)) {
              if (!v.isValid(this, tile)) {
                  tile.setCacheResult(CacheResult.MISS);
                  return false;
              }
          }
          return true;
      } catch (GeoWebCacheException gwce) {
          LOGGER.info(gwce.getMessage());
          tile.setErrorMsg(gwce.getMessage());
          return false;
      }
  }
```

This lets us with:

- Composition is free. If both the Metadata extension and Taskmanager are installed, each contributes a validator and any one can invalidate. No policy debate about which "wins."
- Semantic-neutral. A validator can compare timestamps today, compare hashes/ETags tomorrow, call an external invalidation bus the day after. We don't tie core to the "last data update timestamp" model.
- Extensions don't touch GWC internals. They answer a yes/no question; they don't re-implement retrieve semantics, meta-tile handling, or `Conveyor.CacheResult` side-effects.
- `ConveyorTile` is in the signature, so a future validator can be per-tile / per-parameter without widening the SPI.

Then you can provide a `MetadataTimestampCachedTileValidator` or such in the extension(s) where its relevant, and the same for the GUI component.

PR #9037, opened 2026-01-14
Fix/validate layers properties after catalog load

petersmythe:
[GEOS-12032] Spurious "Namespace/Workspace ... is unknown in rule ..." warnings during startup

See [above issue](https://osgeo-org.atlassian.net/browse/GEOS-12032) for the background to the fix below.

**Fix implemented**

I implemented the safer/less-invasive option: defer validation warnings until the Catalog is known to be populated while keeping early parsing available for access control.
**Changes (file and behavior):**
  - **File edited:** `DataAccessRuleDAO.java`
  - **Key changes:**
    - Downgraded immediate startup "namespace/group unknown" logs in `parseDataAccessRule(...)` from WARNING to FINE and append "(validation deferred)".
    - Introduced `rulesValidated` flag, reset when rules are reloaded so validation runs again on change.
    - Added `registerCatalogListener()` that registers a CatalogListener on `rawCatalog`. On `reloaded()` the listener calls `validateRulesAgainstCatalog()`.
    - Added `validateRulesAgainstCatalog()` that iterates current rules and emits WARNING only when the catalog is actually missing referenced workspace / layer group.
    - Call `registerCatalogListener()` from the DAO constructors.
**Rationale:**
    Keeps early parsing (so ResourceAccessManager can be created and used), avoids spurious WARNINGs, and performs authoritative validation once the Catalog is ready.

**Code snippets / pointers**
  - *Original (conceptual):*
	  parse time: if workspace missing → `LOGGER.warning(...)`
- *New approach:*
	parse time: if workspace missing → `LOGGER.fine(... + " (validation deferred)")`
	later on catalog `reloaded()` → `validateRulesAgainstCatalog()`:
		if workspace missing → `LOGGER.warning(...)`

**Tests added**
- *Integration test:* simulate `rawCatalog.reloaded()` event while a referenced workspace is absent
Assert that the WARNING is emitted by `validateRulesAgainstCatalog()`.
- *Regression test:* ensure the ResourceAccessManager still builds a rule tree and enforces rules the same as before.

**Alternatives considered (and reasons rejected)**
- *Delay parsing entirely until Catalog is loaded:*
  Pros: eliminate any chance of startup spurious logs.
  Cons: ResourceAccessManager (and other services) may require parsed rules early — changing initialization order is invasive and may break components that expect security rules to be available during startup.
- *Build authorization tree lazily on first use:*
  Pros: avoids startup ordering issues.
  Cons: may delay security decisions or cause subtle differences in behavior; requires more changes and tests.
- *Rely on Spring bean ordering or lifecycle events:*
  Pros: could coordinate initialization.
  Cons: brittle, hard to guarantee consistent across environments and tests.

I chose the validation-defer approach because it's minimal, safe, preserves current behaviour, removes the noise, and keeps a clear validation step when catalog data is known to be available.

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/impl/DataAccessRuleDAO.java
+++ b/src/main/src/main/java/org/geoserver/security/impl/DataAccessRuleDAO.java
@@ -54,7 +54,7 @@ public class DataAccessRuleDAO extends AbstractAccessRuleDAO<DataAccessRule> {
     String filesystemSandbox;
 
     // set to true once rules have been validated against a populated catalog
-    private volatile boolean rulesValidated = false;
+    private volatile boolean rulesValidated = false; 
 
     /** Returns the instanced contained in the Spring context for the UI to use */
     public static DataAccessRuleDAO get() {
@@ -123,7 +123,7 @@ public class DataAccessRuleDAO extends AbstractAccessRuleDAO<DataAccessRule> {
         this.catalogMode = catalogMode;
         this.rules = result;
         // rules changed, validation needs to be re-run against the catalog
-        this.rulesValidated = false;
+        this.rulesValidated = false; 
     }
 
     /** Parses a single layer.properties line into a {@link DataAccessRule}, returns false if the rule is not valid */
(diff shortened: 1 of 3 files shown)


petersmythe:
@groldan this bug was perhaps introduced in the Faster Catalog Loading in 2.27.0, is that possible?  Is this a reasonable solution to the problem?

petersmythe:
I'm not sure what the 4 failing tests are about, unrelated, I believe.

mprins:
> I'm not sure what the 4 failing tests are about, unrelated, I believe.

yes; these seem unrelated
```
Error: 2,133 [ERROR] Failures: 
Error: 2,133 [ERROR]   CollectionLayerTest.testBandsFlagsGrayFlag:414 Images are visibly different, found 1519 different pixels, against a threshold of 1000
You can add -Dorg.geotools.image.test.interactive=true to show a dialog comparing them (requires GUI support)
Error: 2,133 [ERROR]   CollectionLayerTest.testCreateCollectionSimpleLayer:189 Images are visibly different, found 1601 different pixels, against a threshold of 1000
You can add -Dorg.geotools.image.test.interactive=true to show a dialog comparing them (requires GUI support)
Error: 2,134 [ERROR]   CollectionLayerTest.testCreateCollectionSimpleLayerTestWorkspace:247 Images are visibly different, found 1601 different pixels, against a threshold of 1000
You can add -Dorg.geotools.image.test.interactive=true to show a dialog comparing them (requires GUI support)
Error: 2,134 [ERROR]   CollectionLayerTest.testCreateCollectionSimpleLayerUTM:759 Images are visibly different, found 1115 different pixels, against a threshold of 1000
You can add -Dorg.geotools.image.test.interactive=true to show a dialog comparing them (requires GUI support)
Error: 2,134 [ERROR]   CollectionLayerTest.testModifyConfigurationSingleBand:331->testCreateCollectionSimpleLayer:189 Images are visibly different, found 1601 different pixels, against a threshold of 1000
You can add -Dorg.geotools.image.test.interactive=true to show a dialog comparing them (requires GUI support)
```

aaime:
Needs to be rebased on top of the current main, then the tests should pass

petersmythe:
> Needs to be rebased on top of the current main, then the tests should pass

Yup, checks now pass, thanks

groldan:
@petersmythe the previous rebase attempt ended up with 13 commits and 56 files changed (0b4467296cf52cd650861d30c3e8e5dab4f63510), I just performed the rebase locally and force-pushed prior verification the changes are the same.

groldan:
> @groldan this bug was perhaps introduced in the Faster Catalog Loading in 2.27.0, is that possible? Is this a reasonable solution to the problem?

Peter and I verified it's not related to whether the new or old catalog loader is used.

Also, I've spent a lot of time trying to replicate the issue with no success:

- The reported issue is that log warnings are emitted during startup for a data access rule where the layer does exist
- DataAccessRuleDAO depends on rawCatalog as a constructor argument. This means when Spring instantiates it, rawCatalog has already being wired up by Spring
- The rawCatalog is populated when GeoServerLoaderProxy, acting as a BeanPostProcessor, processes it. Hence, by the time DataAcessRuleDAO is instantiated, the Catalog is populated.
- I've tried hitting several URLs (rest/getcapabilities) repeteadly during startup (while the catalog is yet loading) to try to reproduce, with added data access rules to layers.properties

Yet, Pete says it happens to him nonetheless.

As it stands, the patch seems pretty safe, but I'm weary of including it in 2.28.2 "last minute" and without having identified the root cause.

@aaime comments? suggestions?

aaime:
I'd like to see a data directory allowing to reproduce the problem, at least on some systems. 
The issue Peter is seeing in the validation might be rooted somewhere else, and eventualy cause other problems

PR #9064, opened 2026-01-23
[GEOS 11977] Broken LegendURL in WMTS GetCapabilities for styles with custom legend

asedgmen:
[GEOS-11977](https://osgeo-org.atlassian.net/browse/GEOS-11977) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

Missing "styles" path segment has been included in the legend graphic URL for styles that have a custom legend, as presented in the WMTS GetCapabilities statement.  The workspace name is also included in the URL path if the style is in a workspace.

Existing testGetLegendsLayer unit test in the GeoServerTileLayerTest class updated to test change.

# Checklist


src/gwc/src/main/java/org/geoserver/gwc/layer/GeoServerTileLayer.java
```diff
@@ -1637,17 +1638,55 @@ public Map<String, org.geowebcache.config.legends.LegendInfo> getLayerLegendsInf
...
+                    // The base URL should be composed using the context path of the web application, obtained from
+                    // the ServletContext object. The base URL provided by the org.geoserver.ows.util.baseURL utility
+                    // cannot be reliably used as it obtains the context path from the HttpServletRequest object
+                    // which may include path segments in addition to the servlet context path, resulting in a
+                    // malformed custom legend URL.
+                    HttpServletRequest request = Dispatcher.REQUEST.get().getHttpRequest();
+                    ServletContext servletContext = request.getServletContext();
+                    String baseUrl = request.getScheme() + "://" + request.getServerName() + ":"
+                            + request.getServerPort() + servletContext.getContextPath();
+
```
aaime: The bug report only talks about "styles" path being missing, and that's indeed a problem, but here you're targeting something else... 
The calculation of the base URL is done almost everywhere using ReponseUtils.baseURL, what you describe in the comments should be a common occurrence in all capabilities backlinks for example, which would make the services unusable. 
Maybe there is something else going on? Some issue with the proxy base?

I need to see some examples of how this came to be, before approving this part of the PR.

aaime (commented):
Thanks for your contribution. The QA build failure is legit, to fix it:

* cd geoserver/src/gwc
* mvn clean install -Dqa

See also the comment about baseURL below.

src/gwc/src/main/java/org/geoserver/gwc/layer/GeoServerTileLayer.java
asedgmen: Thanks for speedy review.

I came across an issue when a workspace is included in the WMTS GetCapabilities request, for example the "ne" workspace with the demo data_dir:

	http://localhost:8080/geoserver/ne/gwc/service/wmts?service=WMTS&request=GetCapabilities
	
The base URL provided by the org.geoserver.ows.util.baseURL utility includes the servlet context path with the workspace name:

	http://localhost:8080/geoserver/ne

aaime:
Many github actions failing... probably because the branch is outdated. Can you rebase on top of the current main?

src/gwc/src/main/java/org/geoserver/gwc/layer/GeoServerTileLayer.java
aaime: The ResponseUtils.baseURL does other things, like port elimination for common ports (80, 443), ability to add authentication keys back in the URL and other stuff that might be needed, dependingon the requirements of the deploy.

Rather than stripping away baseURL, can you do something like this (mind, not tested, and the comments are just to support explaining the code in this comment):

```
  String wsSuffix = LocalWorkspace.get() != null ? "/" + LocalWorkspace.get().getName() : "";
  String baseUrl = baseUrl();  // may end with .../ne/
  if (!wsSuffix.isEmpty() && baseUrl != null) {
      baseUrl = baseUrl.replaceFirst(Pattern.quote(wsSuffix) + "/?$", "/");
  }
```

That reuses baseUrl() (default-port handling, REST fallback, null guard, security keys and removes only the GWC-injected segment.

asedgmen: Code updated pretty much as suggested, i.e. use the ResponseUtils.baseURL to obtain the base URL and strip the workspace suffix using regex.  Note that the last occurrence of the workspace suffix string is removed to avoid issues if the same string happens to occur elsewhere in URL.

ianturton:
You need to run spotless across your changes to pass the QA checks, `mvn -Dqa install` should do it

PR #9072, opened 2026-01-28
Fix aboutcontroller advice v2

AjvadZ:
# Checklist

# Description

This PR fixes an issue in the REST /about/version.json endpoint where JSON output could be duplicated due to AboutController applying @ControllerAdvice to itself.

The self-applied advice caused XStream serialization to run twice, resulting in two consecutive JSON objects being written to the response stream during REST JSON serialization.

### DIFF
--- a/src/restconfig/src/test/java/org/geoserver/rest/catalog/AboutControllerTest.java
+++ b/src/restconfig/src/test/java/org/geoserver/rest/catalog/AboutControllerTest.java
@@ -80,6 +80,7 @@ public class AboutControllerTest extends GeoServerSystemTestSupport {
         JSONObject json = (JSONObject) getAsJSON(BASEPATH + "/about/version.json");
         // print(json);
         checkJSONModel(json);
+        assertTrue(json.has("about"));
     }
 
     @Test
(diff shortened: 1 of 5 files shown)


jodygarnett:
Approving the workflow so we can see the tests run - do you have access to Jira to create a bug for the release notes?

aaime:
Also, did anyone check that the test fails if run on un-modified code?

src/restconfig/src/test/java/org/geoserver/rest/catalog/AboutControllerTest.java
```diff
@@ -80,6 +80,7 @@ public void testGetAsVersionsJSON() throws Exception {
         JSONObject json = (JSONObject) getAsJSON(BASEPATH + "/about/version.json");
         // print(json);
         checkJSONModel(json);
+        assertTrue(json.has("about"));
```
mprins, line 83: this is validated in the method above

mprins (changes requested):
I don't see how these changes to the dependencies relato to the PR title

PR #9125, opened 2026-03-02
[GEOS-12056] OGC API CQL conformance URIs are wrong

RaineEkman:
[GEOS-12056](https://osgeo-org.atlassian.net/browse/GEOS-12056) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

OGC API conformance URIs should point to the correct resource, as defined in the relevant standards.

# Checklist

### DIFF
--- a/src/extension/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/ConformanceClass.java
+++ b/src/extension/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/ConformanceClass.java
@@ -9,7 +9,7 @@ public class ConformanceClass {
     public static final String CORE = "http://www.opengis.net/spec/ogcapi-common-1/1.0/conf/core";
     public static final String HTML = "http://www.opengis.net/spec/ogcapi-common-1/1.0/conf/html";
     public static final String JSON = "http://www.opengis.net/spec/ogcapi-common-1/1.0/conf/json";
-    public static final String OAS3 = "http://www.opengis.net/spec/ogcapi-common/1.0/req/oas30";
+    public static final String OAS3 = "http://www.opengis.net/spec/ogcapi-common/1.0/conf/oas30";
 
     public static final String COLLECTIONS = "http://www.opengis.net/spec/ogcapi-common-2/1.0/conf/collections";
     public static final String GEODATA = "http://www.opengis.net/spec/ogcapi-common-2/1.0/conf/geodata";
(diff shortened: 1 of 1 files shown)


src/extension/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/ConformanceClass.java
```diff
@@ -26,37 +26,37 @@ public class ConformanceClass {
...
-    public static final String CQL2_ADVANCED = "http://www.opengis.net/spec/cql2/1.0/req/advanced-comparison-operators";
+    public static final String CQL2_ADVANCED = "http://www.opengis.net/spec/cql2/1.0/conf/advanced-comparison-operators";
     // right now includes also accent insensitive under this clause
     public static final String CQL2_CASE_INSENSITIVE =
-            "http://www.opengis.net/spec/cql2/1.0/req/case-insensitive-comparison";
-    public static final String CQL2_BASIC_SPATIAL = "http://www.opengis.net/spec/cql2/1.0/req/basic-spatial-operators";
-    public static final String CQL2_SPATIAL = "http://www.opengis.net/spec/cql2/1.0/req/spatial-operators";
+            "http://www.opengis.net/spec/cql2/1.0/conf/case-insensitive-comparison";
+    public static final String CQL2_BASIC_SPATIAL = "http://www.opengis.net/spec/cql2/1.0/conf/basic-spatial-operators";
+    public static final String CQL2_SPATIAL = "http://www.opengis.net/spec/cql2/1.0/conf/spatial-operators";
```
jratike80, line 53: These seem to be changed into
http://www.opengis.net/spec/cql2/1.0/conf/basic-spatial-functions
http://www.opengis.net/spec/cql2/1.0/conf/spatial-functions

src/extension/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/ConformanceClass.java
```diff
@@ -26,37 +26,37 @@ public class ConformanceClass {
...
 
     // requires implementation of all the operators, basic comparisons do work against temporal data
     // too
-    public static final String CQL2_TEMPORAL = "http://www.opengis.net/spec/cql2/1.0/req/temporal-operators";
-    public static final String CQL2_PROPERTY_PROPERTY = "http://www.opengis.net/spec/cql2/1.0/req/property-property";
-    public static final String CQL2_FUNCTIONS = "http://www.opengis.net/spec/cql2/1.0/req/functions";
-    public static final String CQL2_ARITHMETIC = "http://www.opengis.net/spec/cql2/1.0/req/arithmetic";
+    public static final String CQL2_TEMPORAL = "http://www.opengis.net/spec/cql2/1.0/conf/temporal-operators";
```
jratike80: http://www.opengis.net/spec/cql2/1.0/conf/temporal-functions

jratike80:
Have you just changed req->conf, or have you checked that the corrected forms really appear that way in corresponding standards? I checked the CQL2 changes against https://docs.ogc.org/is/21-065r2/21-065r2.html and found some differences.

It would be easier for reviewers to split the PR and add links to the standards. For example, I have no idea how to check if this corrected URI is right or wrong:
https://geoserver.org/spec/ecql/1.0/conf/ecql-text

src/extension/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/ConformanceClass.java
RaineEkman: Yes, looks like those have changed at some point before the standard was final.

src/extension/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/ConformanceClass.java
RaineEkman: Noted and corrected.

RaineEkman:
> Have you just changed req->conf, or have you checked that the corrected forms really appear that way in corresponding standards? I checked the CQL2 changes against https://docs.ogc.org/is/21-065r2/21-065r2.html and found some differences.

I've been mostly looking at secondary sources, like this Hakunapi issue: https://github.com/nlsfi/hakunapi/issues/27, so my understanding of the details here is a bit thin. And I indeed did just a search & replace, didn't expect those ends of URIs to have changed.


> It would be easier for reviewers to split the PR and add links to the standards. For example, I have no idea how to check if this corrected URI is right or wrong: https://geoserver.org/spec/ecql/1.0/conf/ecql-text

I reverted that change after a closer look, can't figure out the status of ECQL. Is it anywhere close to a standard or just something GeoServer invented? Probably an entirely separate battle from this one.

jodygarnett:
Q: We do run CITE tests, is that not checking the conformance URIs?

ECQL is a community standard, implemented by GeoTools. The URI is "fine", although I note for many of the OGC URIs the resolve to a web page (indeed I thought I had done that).

Options:
- `https://docs.geotools.org/latest/userguide/library/cql/ecql.html`
- `https://docs.geoserver.org/latest/en/user/filter/ecql_reference.html`

src/extension/ogcapi/ogcapi-core/src/main/java/org/geoserver/ogcapi/ConformanceClass.java
```diff
@@ -26,37 +26,37 @@ public class ConformanceClass {
...
 
     public static final String ECQL_TEXT = "http://geoserver.org/spec/ecql/1.0/req/ecql-text";
 
     /** Search conformance class for "OGC API - Features - Part 5: Search (PROPOSAL DRAFT)" */
-    public static final String SEARCH = "http://www.opengis.net/spec/ogcapi-features-5/1.0/req/search";
+    public static final String SEARCH = "http://www.opengis.net/spec/ogcapi-features-5/1.0/conf/search";
 
     /** CQL2 encoding conformance classes */
-    public static final String CQL2_TEXT = "http://www.opengis.net/spec/cql2/1.0/req/cql2-text";
+    public static final String CQL2_TEXT = "http://www.opengis.net/spec/cql2/1.0/conf/cql2-text";
```
jodygarnett, line 40: See https://github.com/OGC-tracker/OGC-Standards-Tracker/issues/726 the documentation is inconsistent. 

We have to stick to whatever the CITE tests expect.

aaime:
Quick note: the CITE tests only check conformance for the parts they are intended to test, that is, core and CRS by reference. There is no test for filtering, so that portion just gets ignored, it can have wrong URIs and the test would not notice.

PR #9316, opened 2026-04-12
Fix fixed-timezone handling for WFS GeoJSON `timeStamp`

Tian-geer:
## Background
  In the current WFS GeoJSON response, the top-level `timeStamp` and the time fields inside `properties` use different timezone handling strategies.

  Before this fix:
  - the top-level `timeStamp` was always serialized with a fixed UTC/GMT-based formatter;
  - time fields inside `properties` followed the existing date/time serialization path, which is affected by the `TemporalUtils` / `GeoJSONBuilder` handling
  strategy.

  As a result, different time fields in the same response were serialized differently. This is confusing for clients because `timeStamp` and the temporal
  fields in `properties` all represent time values, but they do not follow the same timezone rules.

  ## What is changed
  This change aligns the serialization strategy of the top-level `timeStamp` with the one already used for temporal fields in `properties`.

  Specifically:
  - the semantic meaning of `timeStamp` remains unchanged and still represents the response generation time;
  - `timeStamp` no longer uses the fixed-UTC `ISO8601Formatter` directly;
  - when date/time formatting is enabled, `timeStamp` is now serialized via `TemporalUtils.printDate(...)`;
  - when date/time formatting is disabled, `timeStamp` now goes through the default `GeoJSONBuilder` date serialization path.

  ## Result
  After this change, the top-level `timeStamp` and the temporal fields inside `properties` follow the same timezone handling strategy in WFS GeoJSON
  responses. This removes the inconsistency within the response and makes the behavior easier for clients to understand and use.

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/json/GeoJSONFeatureWriter.java
+++ b/src/main/src/main/java/org/geoserver/json/GeoJSONFeatureWriter.java
@@ -30,7 +30,6 @@ import org.geoserver.ows.Dispatcher;
 import org.geoserver.ows.Request;
 import org.geoserver.platform.GeoServerExtensions;
 import org.geoserver.platform.ServiceException;
-import org.geoserver.util.ISO8601Formatter;
 import org.geotools.api.feature.Feature;
 import org.geotools.api.feature.simple.SimpleFeature;
 import org.geotools.api.feature.simple.SimpleFeatureType;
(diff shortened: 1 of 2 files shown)


src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
aaime: Please don't change the signature of a protected method, this is an API change. Just create the Date before the if, inside the method, instead.

aaime (changes requested):
The change looks good, thank you, but you should not change API, see comment below.

Also, the PR will not be merged without a test checking it's working correctly.
The new test must fail if the code has not been patched, and pass with the patch.

src/wfs/src/main/java/org/geoserver/wfs/json/GeoJSONGetFeatureResponse.java
Tian-geer: Thanks for the review. I updated the PR to keep the protected writeCollectionTimeStamp(GeoJSONBuilder) method signature unchanged and moved the Date
  creation back inside the method. I also added a unit test covering the local timezone handling for the GeoJSON collection timeStamp output.

mprins:
Please rebase to the main branch to resolve the merge conflict

aaime:
@ianturton I believe this can go under the "small change", there are only a few lines of non-test code. Thoughts?

PR #9364, opened 2026-04-14
Re-enable acl community module in the assembly smoke test

groldan:
The acl plugin was excluded from communityRelease in d05aa2756c because it refused to start under `build/assembly_test`: its client probes the remote ACL API at boot and its access manager intercepts even anonymous GetCapabilities calls, both of which fail without a reachable ACL server. Fixing it needed two small generic hooks on the smoke tester that other plugins can reuse.

- `AbstractPluginTester.systemProperties()` — returns extra `-D` flags that get appended to `JAVA_OPTS` before the GeoServer process is spawned. Default empty.
- `AbstractPluginTester.basicAuthCredentials()` — returns `"user:password"` for an `Authorization: Basic` header on every HTTP call the tester makes (readiness probe + `getAsDom`/`getAsString`/`get`). Default `null`, no header sent.
- `GeoServerAclTester` uses both: `startupCheck=false` plus placeholder basePath/username/password to let the Spring beans load, and admin:geoserver credentials so GetCapabilities bypasses the ACL access manager instead of hitting the unreachable API.
- src/community/pom.xml: re-add <module>acl</module> to communityRelease.
- src/community/acl/README.md: document `geoserver.acl.client.startupCheck` and `geoserver.acl.client.initTimeout`, the knobs that make offline smoke testing (and slow-cold-start deployments) possible.

  

# Checklist

### DIFF
--- a/build/assembly_test/src/test/java/org/geoserver/AbstractPluginTester.java
+++ b/build/assembly_test/src/test/java/org/geoserver/AbstractPluginTester.java
@@ -13,6 +13,8 @@ import java.net.HttpURLConnection;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
+import java.util.Base64;
+import java.util.Map;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
 import org.w3c.dom.Document;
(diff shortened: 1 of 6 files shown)


aaime:
Looks good, thanks! Merging.

PR #9403, opened 2026-04-17
Fix for parallel extension assemblies

petersmythe:
**Problem**: PR #8910 refactored extension assemblies from a centralized run (at the root pom level, after all modules including `release` were built) to per-module runs. But the assembly descriptors still reference HTML files generated by the `release` module at `${geoserverBaseDir}/release/target/html/extensions/LICENSE.html` and `${geoserverBaseDir}/release/target/html/licenses/*.html`. Since the `release` module builds last in the reactor (step 141/141) and extensions build much earlier (e.g., KML at step 44/141), those files don't exist yet when the assembly plugin runs.

**Fix**: Added two new `markdown-page-generator-plugin` executions to the `assembly-docs` profile in `src/extension/pom.xml`:

1. `shared-extension-docs` — renders the shared `LICENSE.md`, `LICENSE_GS.md`, `LICENSE_GT.md`, and `README.md` from `src/release/extensions/` into `src/release/target/html/extensions/`, the exact path all assembly descriptors reference.

2. `shared-licenses` — renders the top-level license files (`GPL.md`, `LGPL.md`, `NOTICE.md`, `GEOTOOLS_NOTICE.txt`, etc.) from the `licenses/` directory into `src/release/target/html/licenses/`, the path referenced by the assembly descriptors' `licenses` fileSet.

The first extension module to build generates these shared files; subsequent extensions find them already in place. No changes to any assembly descriptors are needed, keeping the diff minimal and the fix safe.

# Checklist

### DIFF
--- a/src/extension/pom.xml
+++ b/src/extension/pom.xml
@@ -728,6 +728,53 @@
                   <inputFileExtensions>md</inputFileExtensions>
                 </configuration>
               </execution>
+              <!--
+                Render the shared LICENSE.md / README.md from src/release/extensions/ into the
+                same path the per-module assembly descriptors expect.  The release module builds
+                AFTER extensions in the reactor, so extensions cannot rely on it having run first.
+                The first extension to build will generate these files; subsequent ones find them
+                already in place (the markdown plugin is a no-op when outputs are up-to-date).
+              -->
+              <execution>
+                <id>shared-extension-docs</id>
+                <phase>process-resources</phase>
+                <goals>
+                  <goal>generate</goal>
+                </goals>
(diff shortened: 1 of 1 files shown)


groldan:
Hi Pete, I can confirm this fix works.
Difference between github workflow and jenkins build job is parallelism, workflow runs with `-T1C`, Jenkins doesn't, causing a race condition in the build order for release/pom.xml.
With this fix the build works even without `-T<greater than one>`

Feel free to update the pr title and commit message to reflect it's not just a "possible fix"

PR #9411, opened 2026-04-17
[GEOS-12093] GSIP 226 release script updates for outstanding GSIPs

jodygarnett:
[GEOS-12093](https://osgeo-org.atlassian.net/browse/GEOS-12093) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

Updating release scripts to make commit change visible in git history:

* [x] GSIP 226: Change to `3.0.0-SNAPSHOT` ahead of RC  
  This requires some fixes for cite and acceptance Makefile using skipTests
* [x] GSIP 226: Include [release branch in commit history](https://github.com/geoserver/geoserver/wiki/GSIP-226) via merge commit
* [x] GSIP 226: Change over to `3.0.1-SNAPSHOT` after merge commit
* [x] GSIP 233: Including [pending](https://github.com/geoserver/geoserver/compare/GSIP-233) in release process
* [x] GSIP 221: Update release scripts for [mkdocs](https://github.com/geoserver/geoserver/wiki/GSIP-221)

This PR is used to confirm that change to `3.0.0-SNAPSHOT`.

Alternate:

* https://github.com/geoserver/geoserver/pull/9418

# Checklist

### DIFF
--- a/build/acceptance/Makefile
+++ b/build/acceptance/Makefile
@@ -17,9 +17,9 @@ war: ## Build the geoserver.war file to use for testing and place it in ./geoser
 .PHONY: build-war
 build-war:
 	# a single mvn package -pl :gs-web-app -am should work but it seems it doesn't take into account test dependencies
-	# so it fails when the version number changes (e.g. 2.28-SNAPSHOT -> 3.0-SNAPSHOT)
+	# so it fails when the version number changes (e.g. 3.0.0-SNAPSHOT -> 3.0.1-SNAPSHOT).
 	@mvn clean install -f ../../src/pom.xml \
-	-DskipTests -Dspotless.apply.skip=true -Pogcapi-features -B -ntp -U -T2C
+	-DskipTests -Dspotless.apply.skip=true -Papp-schema,ogcapi-features -B -ntp -U -T2C
 	@mvn package -f ../../src/pom.xml \
 	-DskipTests -Dspotless.apply.skip=true -Pogcapi-features -B -nsu -ntp -T2C -pl :gs-web-app -am
 
--- a/build/cite/Makefile
+++ b/build/cite/Makefile
@@ -272,13 +272,13 @@ pull: $(suite)
 .PHONY: build-war
 build-war:
 	# a single mvn package -pl :gs-web-app -am should work but it seems it doesn't take into account test dependencies
-	# so it fails when the version number changes (e.g. 2.28-SNAPSHOT -> 3.0-SNAPSHOT)
+	# so it fails when the version number changes (e.g. 3.0.0-SNAPSHOT -> 3.0.1-SNAPSHOT)
 	@mvn clean install -f ../../src/pom.xml \
-	-Pogcapi-features,ogcapi-tiles,geopkg-output,app-schema,wcs10,wcs11 \
-	-DskipTests  -B -ntp -U -T1C
+	-Pogcapi-features,vectortiles,ogcapi-tiles,geopkg-output,app-schema,wcs10,wcs11 \
+	-DskipTests -Dspotless.apply.skip=true -B -ntp -U -T1C
 	@mvn package -f ../../src/pom.xml \
 	-Pogcapi-features,ogcapi-tiles,geopkg-output,wcs10,wcs11 \
-	-DskipTests -B -nsu -ntp -T1C -pl :gs-web-app -am
+	-DskipTests -Dspotless.apply.skip=true -B -nsu -ntp -T1C -pl :gs-web-app -am
 
 .PHONY: copy-war
 copy-war:
--- a/src/extension/ogcapi/ogcapi-core/pom.xml
+++ b/src/extension/ogcapi/ogcapi-core/pom.xml
@@ -10,7 +10,7 @@
   <parent>
     <groupId>org.geoserver.extension</groupId>
     <artifactId>gs-ogcapi</artifactId>
-    <version>3.0-SNAPSHOT</version>
+    <version>3.0.0-SNAPSHOT</version>
   </parent>
 
   <groupId>org.geoserver.extension</groupId>
--- a/src/extension/ogcapi/ogcapi-core/pom.xml
+++ b/src/extension/ogcapi/ogcapi-core/pom.xml
@@ -113,18 +113,6 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.geoserver.extension</groupId>
-      <artifactId>gs-css</artifactId>
-      <version>${project.version}</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.geoserver.extension</groupId>
-      <artifactId>gs-mbstyle</artifactId>
-      <version>${project.version}</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
       <groupId>org.easymock</groupId>
       <artifactId>easymock</artifactId>
       <scope>test</scope>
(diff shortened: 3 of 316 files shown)


src/web/app/pom.xml
```diff
@@ -1539,11 +1539,6 @@
           <artifactId>gs-web-processes</artifactId>
           <version>${project.version}</version>
         </dependency>
-        <dependency>
```
aaime, line 1542: Why did you remove it? The module is part of the build, seems to be properly wired too (I see it in ogcapi/pom.xml, has dependencies and so on):

```
git grep 3d-geovolumes -- "**/pom.xml"
src/community/ogcapi/ogcapi-3d-geovolumes-assembly/pom.xml:      <artifactId>gs-ogcapi-3d-geovolumes</artifactId>
src/community/ogcapi/ogcapi-3d-geovolumes-assembly/pom.xml:      <artifactId>gs-web-3d-geovolumes</artifactId>
src/community/ogcapi/ogcapi-3d-geovolumes/pom.xml:  <artifactId>gs-ogcapi-3d-geovolumes</artifactId>
src/community/ogcapi/pom.xml:    <module>ogcapi-3d-geovolumes</module>
src/community/ogcapi/pom.xml:    <module>ogcapi-3d-geovolumes-assembly</module>
src/community/ogcapi/pom.xml:    <module>web-3d-geovolumes</module>
src/community/ogcapi/web-3d-geovolumes/pom.xml:  <artifactId>gs-web-3d-geovolumes</artifactId>
src/community/ogcapi/web-3d-geovolumes/pom.xml:      <artifactId>gs-ogcapi-3d-geovolumes</artifactId>
src/web/app/pom.xml:          <artifactId>gs-web-3d-geovolumes</artifactId>
src/web/app/pom.xml:          <artifactId>gs-web-3d-geovolumes</artifactId>
```

What problem were you trying to solve here?

build/acceptance/Makefile
```diff
@@ -17,11 +17,14 @@ war: ## Build the geoserver.war file to use for testing and place it in ./geoser
 .PHONY: build-war
 build-war:
 	# a single mvn package -pl :gs-web-app -am should work but it seems it doesn't take into account test dependencies
-	# so it fails when the version number changes (e.g. 2.28-SNAPSHOT -> 3.0-SNAPSHOT)
+	# so it fails when the version number changes (e.g. 3.0.0-SNAPSHOT -> 3.0.1-SNAPSHOT).
+	#
+	# mvn clean install -f ../../src/pom.xml -DskipTests -Dspotless.apply.skip=true -Papp-schema,ogcapi-features -B -ntp -U -T2C
 	@mvn clean install -f ../../src/pom.xml \
-	-DskipTests -Dspotless.apply.skip=true -Pogcapi-features,wcs10,wcs11 -B -ntp -U -T2C
+	-DskipTests -Dspotless.apply.skip=true -Papp-schema,ogcapi-features -B -ntp -U -T2C
```
aaime, line 24: Hum interesting removal of wcs10 and wcs11 ... the acceptance tests do not seem to mind though, so they are probably not using them.

src/extension/ogcapi/ogcapi-core/pom.xml
```diff
@@ -147,4 +135,30 @@
     </dependency>
   </dependencies>
 
+  <profiles>
```
aaime, line 138: I think I understand where you're going here, and it's probably  working fine for mind, but this is a danger zone.

-DskipTests skips running the tests, but does not skip compiling them 
Now, this works because the tests rely on the modules being there at run time, but if a test dependency was used also at compile time (import, direct code usage) this approach would fail.

If you need to skip both compiling and running the tests, -Dmaven.test.skip=true should do the trick.

src/extension/ogcapi/ogcapi-features/pom.xml
```diff
@@ -137,6 +130,13 @@
       <classifier>tests</classifier>
       <scope>test</scope>
     </dependency>
-
+    <dependency>
```
aaime, line 133: It was part of the block of test dependencies before, and it is now too. I don't see what moving the block buys us?

build/cite/Makefile
```diff
@@ -272,13 +272,16 @@ pull: $(suite)
 .PHONY: build-war
 build-war:
 	# a single mvn package -pl :gs-web-app -am should work but it seems it doesn't take into account test dependencies
-	# so it fails when the version number changes (e.g. 2.28-SNAPSHOT -> 3.0-SNAPSHOT)
+	# so it fails when the version number changes (e.g. 3.0.0-SNAPSHOT -> 3.0.1-SNAPSHOT)
+	#
+	# mvn clean install -f ../../src/pom.xml -Pogcapi-features,ogcapi-tiles,geopkg-output,app-schema,wcs10,wcs11  -DskipTests  -B -ntp -U -T1C
```
aaime, line 277: Why do we have a commented out maven command here, just to have an actual version running below? If it was for a test, clean up after you're done.

aaime (commented):
The PR is so big that doing a full review is hard, I tried to do it commit by commit but it looks like you went back and forth with some changes, one commit looks like a "backing off a previous change", I would recommend squashing it.

petersmythe:
Note to self, once the versioning is sorted out, this file needs attention to produce the correct links: https://github.com/geoserver/geoserver/blob/main/doc/version.py

build/cite/Makefile
jodygarnett: So I could tell in the build output what had actually been run 😆 

Yes I could comment it out, but it does provide some value for troubleshooting. Maybe `make` has a flag that can share what it is running.

build/acceptance/Makefile
jodygarnett: It was more that I made the dependency optional based on skipTests (so they are no longer needed here when `-DskipTests` is used.

src/extension/ogcapi/ogcapi-core/pom.xml
jodygarnett: It all comes back to that `Makefile`, it needs to build tests to ensure the test modules are available for subsequent steps.

We can try asking that `Makefile` to `maven.test.skip=true` and see if that is an easier way to resolve the build order issues.

src/extension/ogcapi/ogcapi-features/pom.xml
jodygarnett: This was a case where I tried the profile trick, but since it was needed at compile time I had to put it back.

Let me put it back in the previous location to avoid noise

jodygarnett:
> The PR is so big that doing a full review is hard, I tried to do it commit by commit but it looks like you went back and forth with some changes, one commit looks like a "backing off a previous change", I would recommend squashing it.

Thanks for review, I was indeed troubleshooting in order to see what was required to these Makefile based workflows to pass. I will leave the version number change distinct and squash the other changes

src/web/app/pom.xml
jodygarnett: It is already included above on line 1499 so maven produces a warning

src/extension/ogcapi/ogcapi-core/pom.xml
jodygarnett: @aaime here is a PR with the use of `maven.test.skip` approach:
* https://github.com/geoserver/geoserver/pull/9418

I did a minimal change pulling `gs-*` test classification into profile, rather than all test dependences.

build/cite/Makefile
jodygarnett: removed

jodygarnett:
Superceeded by https://github.com/geoserver/geoserver/pull/9418 based on feedback about `maven.test.skip=true` being preferred.

PR #9419, opened 2026-04-19
control-flow: stuck-slot watchdog

groldan:
## TL;DR

Adds a diagnostic to `ControlFlowCallback`: on an HTTP 503 queue-wait timeout, log a WARN per peer that has been holding a slot past `timeout`, including the peer thread's top stack frame. Also warn on the release path if a slot was held beyond `timeout`. No new threads, no scheduled tasks, observability only.

Also adds `src/extension/control-flow/README.md` — developer onboarding docs for the extension (first commit of this branch).

## Backstory

During stress/load testing the 3.x artifacts an incident looked like a control-flow deadlock: under concurrent WMS GetMap load on a GeoPackage-backed layer group, the server would freeze, the "running requests" counter would peg at the configured concurrency cap, and every subsequent request would return the usual `Requested timeout out while waiting to be executed` 503. Smelled like control-flow misbehaving, so started investigating inside this extension.

After reading through `ControlFlowCallback`, `SimpleThreadBlocker`, `SingleQueueFlowController`, and the nesting/release paths, nothing in control-flow itself matched the symptoms: the `SimpleThreadBlocker` queue was doing exactly what it advertises, `finished()` was not being called simply because the operation it was supposed to wrap never returned, and the `doFilter` safety net couldn't fire either for the same reason.

Reproducing with `ab -c 8` against the `ne:world` layer group, then `ab -c 32` against a single GeoPackage layer, and capturing two `jstack` dumps a few seconds apart showed the real story: the threads holding slots were all wedged in native SQLite calls (`org.sqlite.core.NativeDB._open_utf8`, `NativeDB._close`) with zero CPU time advancing between dumps. The server also hangs without the control-flow extension installed. Root cause is in `gt-geopkg` / xerial `sqlite-jdbc`, not in GeoServer. Filed as **[GEOS-12094](https://osgeo-org.atlassian.net/browse/GEOS-12094)**.

So the diagnosis cost a `jstack` cycle that could be avoided. Control-flow has all the information needed to point the finger at the right layer.

## What's in the PR

1. `ControlFlowCallback` tracks each running request's start time and owning thread in a `ConcurrentHashMap<Long, CallbackContext> ACTIVE_RUNNING`, populated after controller acquisition completes, cleared on release.
2. When a request is about to throw 503 because its queue-wait expired, the callback walks `ACTIVE_RUNNING` and logs one WARN per peer whose hold duration exceeds `timeout`. The WARN includes the peer thread name, its `Request`, and its top stack frame.
3. When a request releases after holding its slot longer than `timeout`, a WARN is logged naming the hold duration. For the truly-stuck case this line never fires (the peer-dump from (2) is the primary signal).
4. README section explaining the distinction between the queue-wait window (bounded by `timeout`) and the running window (not bounded by control-flow at all), and what the watchdog can and cannot tell you. The sample log output is the real GEOS-12094 trace.

The heuristic is "a hold longer than the queue-wait budget is suspicious, not necessarily stuck".  A legit slow render will also trigger the WARN, but its top frame will read as progress (`StreamingRenderer.drawFeature` and similar) rather than as a wedge in native I/O. The watchdog is a pointer, not a verdict; the stack frame is the evidence.

## Why it belongs in control-flow even though the bug isn't here

Control-flow is where the 503 is emitted, and it is the component with the process-wide view of who is currently holding slots. Any other place in GeoServer that tried to provide this signal would have to reimplement the same bookkeeping. Keeping the watchdog inline with the 503 path also means it has zero cost when no one is timing out.

## Tests

`ControlFlowCallbackTest` adds two tests, modelled on the existing `testTimeout`:

- `testStuckSlotWatchdogLogsPeerOnQueueTimeout`: a peer thread fills the single-capacity controller slot and parks on a `CountDownLatch`; the test thread issues a second request that 503s; a JUL `Handler` captures the emitted WARN and asserts it names the peer thread, its request, and carries a `top frame:`.
- `testHeldSlotWarnsOnSlowRelease`: single request held past `timeout` then releases normally; assert the release-path WARN fires.

## Reproduction / verification

Against the Natural Earth dataset from the release datadir:

```
ab -s 15 -t 5 -c 8 "<WMS GetMap on ne:world>"
```

Once wedged, a follow-up `curl` returns 503 and the log now reads:

```
WARN  [geoserver.flow] - Request [WMS 1.1.0 GetMap] timed out waiting on BasicOWSController(wms.getmap,SimpleBlocker(4));
      peer thread qtp…-81 has been holding a slot for 769328ms running [WMS 1.1.1 GetMap]
      - top frame: org.sqlite.core.NativeDB._open_utf8(Native Method)
```

…which is the information GEOS-12094 needed.

  

# Checklist

### DIFF
--- a/src/extension/control-flow/README.md
+++ b/src/extension/control-flow/README.md
@@ -1,0 +1,256 @@
+# control-flow
+
+> **Note:** this document was prepared in April 2026 against the module behavior at that time. It is intended as
+> onboarding material and is not actively maintained alongside the code. Treat it as a starting point: the module may
+> have evolved since, and individual statements below may be out of sync with the current source. When in doubt,
+> trust the code.
+
+Throttles concurrent OWS requests and optionally rate-limits or prioritises them by user, IP or HTTP header.
+
+The extension also ships a small IP blacklist filter that rejects blacklisted addresses before the dispatcher runs.
+
+This README covers the internals - for configuration reference, see the GeoServer user manual.
+
+## How it works
+
+Every HTTP request that reaches GeoServer goes through a Spring-delegated servlet filter. For OWS requests the
+dispatcher also fires callbacks once the request has been parsed into a service+operation. `ControlFlowCallback` is
+registered as both a filter and a dispatcher callback, hooking both points from a single class. When the dispatcher
+callback fires it asks a `FlowControllerProvider` for the list of controllers that apply to the request, walks that list
+in priority order calling `requestIncoming(request, timeout)` on each, and only lets the operation run after every controller
+has granted a slot. When the operation finishes the list is walked in reverse and each slot is released. Everything
+else in this module is configuration surface around that core flow.
+
+Rules live in `$GEOSERVER_DATA_DIR/controlflow.properties`. Example:
+
+```properties
+timeout=10
+ows.global=16
+ows.wms.getmap=4
+user.ows.wms.getmap=6/s;2s
+ip=8
+ip.blacklist=192.168.1.8, 192.168.1.10
+ows.priority.http=gs-priority,0
+```
+
+## Where the code lives
+
+```
+org.geoserver.flow              ControlFlowCallback (filter + DispatcherCallback in one class),
+                                ControlFlowConfigurator, FlowController, FlowControllerProvider,
+                                DefaultFlowControllerProvider, ControllerPriorityComparator, ControlModuleStatus
+org.geoserver.flow.config       DefaultControlFlowConfigurator - the controlflow.properties parser
+org.geoserver.flow.controller   FlowController implementations + ThreadBlocker implementations +
+                                key/priority providers + IpBlacklistFilter + request matchers
+```
+
+Start with `ControlFlowCallback.java` - the rest of the module is easier to follow afterwards.
+
+## Spring wiring (`applicationContext.xml`)
+
+Three beans:
+
+| Bean                   | Class                 | Role |
+| ---------------------- | --------------------- | ---- |
+| `controlFlowCallback`  | `ControlFlowCallback` | Implements **both** `AbstractDispatcherCallback` **and** `GeoServerFilter`. Acquires/releases controllers per request. |
+| `ipBlacklistFilter`    | `IpBlacklistFilter`   | Rejects requests from blacklisted IPs with HTTP 403. Reads the same `controlflow.properties`. |
+| `controlExtension`     | `ControlModuleStatus` | Shows up in the module status panel on the admin UI. |
+
+Both `ControlFlowCallback` and `IpBlacklistFilter` are auto-discovered by GeoServer's `SpringDelegatingFilter`, which wraps every incoming
+request in all `GeoServerFilter` beans it finds in the context. No `web.xml` changes, no explicit filter mapping.
+
+If the data dir doesn't ship a custom `ControlFlowConfigurator` or `FlowControllerProvider`, `ControlFlowCallback.setApplicationContext` registers
+the defaults (`DefaultControlFlowConfigurator`, `DefaultFlowControllerProvider`) as singletons on first access. See `ControlFlowCallback.registDefaultBeansIfNeeded`.
+
+## High-level architecture
+
+```mermaid
+graph TD
+    A[HTTP request] --> B[GeoServer filter chain]
+    B --> C[IpBlacklistFilter]
+    C -->|403| Z1[rejected]
+    C --> D[ControlFlowCallback.doFilter]
+    D --> E[Spring Dispatcher]
+    E --> F[Dispatcher.operationDispatched]
+    F --> G[ControlFlowCallback.operationDispatched]
+    G --> H[FlowControllerProvider.getFlowControllers]
+    H --> I[FlowController list<br/>sorted by priority]
+    I --> J[for each controller:<br/>requestIncoming within remaining timeout]
+    J -->|timeout| Z2[HTTP 503<br/>Requested timeout out...]
+    J --> K[operation runs<br/>GetMap, GetCapabilities, ...]
+    K --> L[Dispatcher.finished]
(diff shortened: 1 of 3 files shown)


src/extension/control-flow/README.md
```diff
@@ -0,0 +1,330 @@
+# control-flow
```
aaime, line 1: This document is very detailed and likely useful for someone jumping on the module without previous experience. 
However, we don't have a pratice (not the time) to keep this document up to date in the future. See the developer guide as an example, besides the getting started section, it's abandonware. It's not a matter of where it is, or the format, it's just that we don't have the cycles to maintain it.
I would make it explicit at the top of the document, stating "This document has been prepared on April 2026 based on the module behavior. The module code migth have changed and be out of synch with this description".

src/extension/control-flow/src/main/java/org/geoserver/flow/ControlFlowCallback.java
```diff
@@ -205,6 +234,32 @@ private static String getControllerExitMessage(FlowController controller, Reques
...
+     * been holding its slot for longer than the timeout. Each line includes the top stack frame of the owner thread so
+     * operators can tell queue-wait timeouts (harmless, benchmarking) apart from stuck-slot timeouts (the peer is stuck
+     * in the I/O layer, the datastore, rendering, etc.) without needing a separate jstack.
+     */
+    private static void logStuckRunningRequests(FlowController blockedOn, Request timedOutRequest, long timeout) {
+        if (!LOGGER.isLoggable(Level.WARNING) || ACTIVE_RUNNING.isEmpty()) {
+            return;
+        }
+        final long now = System.currentTimeMillis();
+        for (CallbackContext peer : ACTIVE_RUNNING.values()) {
```
aaime, line 249: The logging looks useful, and it's well guarded for threads that only took the spot for a very long time, so it won't report cases where there is just too much traffic, and the current request as hundreds of relatively requests in front.
I'm however wondering about cases that are still just legitimate, like lots of traffic of relatively short lived WMS requests chewing 7 out of 8 slots, and a single WFS request that is taking a long time because it's doing a large download. This log would point the finger just at the WFS request, while the actual reason is more the excess traffic side. Maybe add a note in the logs to warn against this case, clearifying that the threads that held the lock for a long time are not necessarily to blame for the timeout?

aaime (changes requested):

groldan:
Thanks Andrea, both comments are fair, thanks for the review.

On the README: split it out into its own PR as #9443, with the "prepared April 2026, may drift" disclaimer at the top. From here on this PR is code-only. I'd like to keep the README because I had to understand the module when I thought it was the culprit of the timeouts. Nowadays with the help of AI it'd be easier to keep it up to date from time to time :}

About the watchdog false-positives: you're right that a long-held slot is not the same as the cause. I'm genuinely torn on how far to take this, so happy to go with whatever you prefer. Any combination of the below works for me:

1. Ditch it. The GEOS-12094 wedge was diagnosable with one jstack cycle; maybe that's enough for how rarely this class of bug shows up.
2. One WARN per 503 instead of one per peer, with a trailing `Note:` caveat in the message itself reminding operators that a long-held peer may be a victim, not the cause. Ready on branch [https://github.com/groldan/geoserver/tree/control-flow_stuck-slot_watchdog_single_log_message_per_503](control-flow_stuck-slot_watchdog_single_log_message_per_503]. Could fold it into this PR. The log entry would look like follows, not sure a multi-line log entry is such a good idea though:
```
'qtp1439007204-148' WARN  [geoserver.flow] - Request [WMS 1.1.0 GetMap] timed out waiting on BasicOWSController(wms.getmap,SimpleBlocker(4)); the following peer(s)
have been holding a slot longer than the 10000ms timeout:
 - peer thread qtp1439007204-81 has been holding a slot for 769328ms running [WMS 1.1.1 GetMap] - top frame: org.sqlite.core.NativeDB._open_utf8(Native Method)
 - peer thread qtp1439007204-149 has been holding a slot for 769348ms running [WMS 1.1.1 GetMap] - top frame: org.sqlite.core.NativeDB._close(Native Method)
 - peer thread qtp1439007204-154 has been holding a slot for 769349ms running [WMS 1.1.1 GetMap] - top frame: org.sqlite.core.NativeDB._open_utf8(Native Method)
 - peer thread qtp1439007204-155 has been holding a slot for 769349ms running [WMS 1.1.1 GetMap] - top frame: org.sqlite.core.NativeDB._open_utf8(Native Method)
Note: a long-held slot is not necessarily the cause of this timeout. Under heavy concurrent traffic the listed peer(s) may simply be legitimately slow requests
```
(e.g. a large WFS download) holding a slot when many shorter requests piled up behind them. Use the top stack frame to tell a wedged peer (stuck in native I/O, a lock, etc.) from a peer that is making progress.
3. Raise the bar and only fire on holds above N * timeout (2x or 3x), so that ordinary legit-slow requests don't cross the line as easily.
4. Wrap the response. Add an HttpServletResponseWrapper that flips a flag on first write to `getOutputStream()`/`getWriter()`; the watchdog skips peers that are already streaming bytes. But more invasive than (3).
5. (2) + (3) + (4). Multiplier handles "busy but not yet streaming" (heavy query/parse, buffered capabilities doc); "wrote-byte" flag handles active downloads (maybe not slow renders); single-WARN keeps log volume bounded.

I'd lean (5) if it's worth the code, (2) alone if you want minimal, and I'm equally fine closing this in favour of just the README. Your call.

aaime:
Thinking out loud... I like the output of option 2, it has the right level of info. Maybe just one think, if you are experiencing stuck threads you're probably already looking to perform some debugging, what about lowering the logging level to INFO and maybe add a one line in the docs pointing to this possibility?

aaime:
@groldan do you think the logging could be amended? the description part already landed

PR #9448, opened 2026-04-23
[GEOS-12096] GeoFence: getSecurityFilter is too permissive

etj:
[GEOS-12096](https://osgeo-org.atlassian.net/browse/GEOS-12096) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

`GeoFenceAccessManager::getSecurityFilter` used to return an `INCLUDE` filter.
Now it relies on a permission filter (https://github.com/geoserver/geofence/issues/286) based on the actual GeoFence rules.

This PR should be tested by building locally the code in PR https://github.com/geoserver/geofence/pull/287, since both GeoServer and GeoFence changes should be tested at the same time.
For this very reason, since the GeoFence jars are not in the maven repos yet, the github tests will fail.

# Checklist

### DIFF
--- a/src/extension/geofence/geofence-server/src/main/resources/applicationContext.xml
+++ b/src/extension/geofence/geofence-server/src/main/resources/applicationContext.xml
@@ -55,11 +55,11 @@
         <property name="locations">
             <list>
                 <!-- Default values for an embedded file-based H2 db -->
-                <value>classpath*:geofence-default-override.properties</value>
+                <!--<value>classpath*:geofence-default-override.properties</value>-->
                 <!-- Optional overriding values for custom db. Path relative to data dir -->
                 <value>file:geofence/geofence-datasource-ovr.properties</value>
                 <!-- Optional, use this property if you want to keep your ovr file in a different location than the data dir -->
-                <value>${geofence-ovr}</value>
+                <!--<value>${geofence-ovr}</value>-->
             </list>
         </property>
         <property name="properties">
--- a/src/extension/geofence/geofence/src/main/java/org/geoserver/geofence/GeofenceAccessManager.java
+++ b/src/extension/geofence/geofence/src/main/java/org/geoserver/geofence/GeofenceAccessManager.java
@@ -977,7 +977,15 @@ public class GeofenceAccessManager implements ResourceAccessManager, DispatcherC
         }
 
         RuleFilter ruleFilter = buildPermissionRuleFilter(user);
-        PermsResult permsResult = rulesService.getPermissionFilter(ruleFilter);
+
+        PermsResult permsResult;
+        try {
+            permsResult = rulesService.getPermissionFilter(ruleFilter);
+        } catch (Exception e) {
+            LOGGER.log(Level.SEVERE, "Error retrieving permissions, skipping access to " + ruleFilter, e);
+            return null;
+        }
+
         if (permsResult == null) {
             LOGGER.log(Level.SEVERE, "GeoFence returned null PermsResult for filter {0}, denying access", ruleFilter);
             return Filter.EXCLUDE;
--- a/src/extension/geofence/geofence/src/main/java/org/geoserver/geofence/GeofenceAccessManager.java
+++ b/src/extension/geofence/geofence/src/main/java/org/geoserver/geofence/GeofenceAccessManager.java
@@ -29,7 +29,6 @@ import org.geoserver.catalog.CoverageInfo;
 import org.geoserver.catalog.FeatureTypeInfo;
 import org.geoserver.catalog.LayerGroupInfo;
 import org.geoserver.catalog.LayerInfo;
-import org.geoserver.catalog.Predicates;
 import org.geoserver.catalog.ResourceInfo;
 import org.geoserver.catalog.StyleInfo;
 import org.geoserver.catalog.WMSLayerInfo;
@@ -46,9 +45,11 @@ import org.geoserver.geofence.core.model.enums.GrantType;
 import org.geoserver.geofence.services.RuleReaderService;
 import org.geoserver.geofence.services.dto.AccessInfo;
 import org.geoserver.geofence.services.dto.CatalogModeDTO;
+import org.geoserver.geofence.services.dto.PermsResult;
 import org.geoserver.geofence.services.dto.RuleFilter;
 import org.geoserver.geofence.util.AccessInfoUtils;
 import org.geoserver.geofence.util.GeomHelper;
+import org.geoserver.geofence.util.PermissionCatalogFilterHelper;
 import org.geoserver.geofence.wpscommon.WPSHelper;
 import org.geoserver.ows.Dispatcher;
 import org.geoserver.ows.DispatcherCallback;
@@ -970,7 +971,36 @@ public class GeofenceAccessManager implements ResourceAccessManager, DispatcherC
 
     @Override
     public Filter getSecurityFilter(Authentication user, Class<? extends CatalogInfo> clazz) {
-        return Predicates.acceptAll();
+        // Admin users can access everything
+        if (user != null && !(user instanceof AnonymousAuthenticationToken) && isAdmin(user)) {
+            return Filter.INCLUDE;
+        }
+
+        RuleFilter ruleFilter = buildPermissionRuleFilter(user);
+        PermsResult permsResult = rulesService.getPermissionFilter(ruleFilter);
+        if (permsResult == null) {
+            LOGGER.log(Level.SEVERE, "GeoFence returned null PermsResult for filter {0}, denying access", ruleFilter);
+            return Filter.EXCLUDE;
+        }
+        return new PermissionCatalogFilterHelper().buildCatalogFilter(permsResult, clazz);
+    }
+
+    /**
+     * Builds the {@link RuleFilter} to be used with {@link RuleReaderService#getPermissionFilter}.
+     *
+     * <p>All catalog-scoping fields (service/request/workspace/layer/subfield) are set to ANY so that the call acts as
+     * a discovery query. User, role, source address, and date are populated from the current authentication context.
+     */
+    private RuleFilter buildPermissionRuleFilter(Authentication user) {
+        RuleFilter ruleFilter = new RuleFilter(RuleFilter.SpecialFilterType.ANY);
+        ruleFilter.setInstance(configurationManager.getConfiguration().getInstanceName());
+        setRuleFilterUserAndRole(user, ruleFilter);
+        String sourceAddress = retrieveCallerIpAddress();
+        if (sourceAddress != null) {
+            ruleFilter.setSourceAddress(sourceAddress);
+        }
+        ruleFilter.setDate(DateTimeFormatter.ISO_LOCAL_DATE.format(LocalDate.now()));
+        return ruleFilter;
     }
 
     @Override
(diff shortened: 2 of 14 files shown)


etj:
@copilot review

afabiani (commented):
Read this end-to-end against #287, support#5847, support#5815, and GEOS-12096 — the security-filter implementation itself is small, focused, and has solid test coverage (15 helper-unit tests + 10 access-manager tests). Architecturally I think the shape is right: admin shortcut + GeoFence-backed `PermsResult` + property-name remapping per `CatalogInfo` subtype.

Most of what I'd flag is around release discipline rather than the engineering itself.

**Three things I'd want addressed before this lands in 2.28.4:**

**1. Two `<value>` entries in `applicationContext.xml` are commented out, not removed.** Looks like a debug-config artifact ("comment out the H2 default fallback so my own ovr file takes effect"). It changes deployment behavior (silently disables the `geofence-default-override.properties` fallback and the `${geofence-ovr}` system-property override path) and is unrelated to GEOS-12096. CONTRIBUTING is explicit about "include only relevant changes". I'd revert (or split into a separate PR with its own justification). Inline below.

**2. SNAPSHOT GeoFence dependency on a release branch** — `gf.version: 3.8.2 → 3.8-SNAPSHOT`. The PR body acknowledges "github tests will fail" because the GeoFence jars aren't in the maven repos. 2.28.x is a stable line — this needs to wait for a GeoFence release (3.8.3 or whatever contains #287) and a concrete version pin before it can merge. Inline below with a coordination question.

**3. PR targets `2.28.x` not `main`** — the PR's own checkbox is unchecked. Jira fix-versions are 2.28.4 AND 3.0, so `main` needs the change too; standard policy is merge-to-main-then-backport. As-is the 3.0 line could ship without the fix.

**One genuine correctness concern:**

**4. Inconsistent fail-safe in `getSecurityFilter`** — exception path returns `null` (caller treats as "no filter applied" → fail-open), null-result path returns `Filter.EXCLUDE` (fail-closed). Since the PR is closing a "layers leaking" bug, the safer default for both error paths is `Filter.EXCLUDE`. The exception case (GeoFence restart, JDBC timeout, network blip) is the more common one in production and is the one with the worse failure mode. Inline below.

**Worth eyeballing before final pass:**

- `new PermissionCatalogFilterHelper()` is allocated on every call to `getSecurityFilter` — the helper is stateless (only static `FF` and `LOGGER`). Single Spring bean / static method / `private static final` instance.
- `PermissionCatalogFilterHelper.buildCatalogFilter` returns `Filter.EXCLUDE` for `CatalogInfo` subtypes outside the explicit list (`StoreInfo`, `NamespaceInfo`, `MapInfo`, …). `SecureCatalogImpl` calls `getSecurityFilter` for many of those — worth tracing to confirm nothing breaks. The `WARNING` log in that branch will also spam on busy systems.
- Naming inside `PermissionCatalogFilterHelper`: outer class declares `FF` uppercase; the inner `PropertyNameRemapper` uses `ff` (inherited from `DuplicatingFilterVisitor`). GeoServer/GeoTools idiom is `ff`. Static fields (`FF`, `LOGGER`, `GF_WORKSPACE`, `GF_LAYER`) are package-private — `private static final` would be more idiomatic.
- User docs aren't updated (checkbox unchecked). The behavior change is end-user-visible (Layer Preview, admin UI dropdowns, workspace listings) — at minimum a "behavior change in 2.28.4 / 3.0" note in `doc/en/user/source/extensions/geofence/configuration.rst`.
- `GeofenceAccessManagerVirtualServiceTest` change moves the adapter init from `@Before` to a field — small refactor unrelated to GEOS-12096, same "include only relevant changes" rule.
- Test gap: no test for the exception path on `getPermissionFilter` (would have caught #4). Easy to add — stub that throws.
- The if-chain in `buildCatalogFilter` relies on `LayerInfo` matching its branch *before* `PublishedInfo` (since `LayerInfo extends PublishedInfo`). Reordering would silently change `LayerInfo`'s `wsProp`. A comment in the production code (or a test that locks both branches' wsProp values explicitly) would prevent regressions.

All in all the engineering is fine — the blockers are about release discipline (release-branch SNAPSHOT, scope creep in `applicationContext.xml`, target branch) rather than the security-filter code itself. Once those are sorted I'll do the end-to-end test pass on a local stack with the actual Layer Preview / Catalog UI flows from #5815.

src/extension/geofence/geofence-server/src/main/resources/applicationContext.xml
```diff
@@ -55,11 +55,11 @@
         <property name="locations">
             <list>
                 <!-- Default values for an embedded file-based H2 db -->
-                <value>classpath*:geofence-default-override.properties</value>
+                <!--<value>classpath*:geofence-default-override.properties</value>-->
```
afabiani, line 58: These two `<value>` entries (this one for `geofence-default-override.properties` and the `${geofence-ovr}` one a few lines below) are commented out, not removed. Net effect:
- The embedded H2 default-fallback is gone, so deployments without an explicit `geofence/geofence-datasource-ovr.properties` in the data dir will behave differently or fail.
- The `${geofence-ovr}` system-property override path is also gone, removing a documented way to keep the ovr file outside the data dir.

The change is unrelated to GEOS-12096 and looks like a leftover from a local debugging setup (matches the testing recipe in geosolutions-it/support#5847 where the user is told to put their own ovr file in `<datadir>/geofence/`, which only works cleanly if the H2 fallback isn't shadowing it).

CONTRIBUTING explicitly says "include only relevant changes" — and a release-branch PR amplifies the point. Suggest reverting this hunk in this PR. If the change is intentional behavior, it deserves its own PR with its own Jira ticket and release-notes entry.

src/pom.xml
```diff
@@ -107,7 +107,7 @@
     <compress-lzf.version>1.0.3</compress-lzf.version>
     <eclipse.emf.version>2.15.0</eclipse.emf.version>
     <errorProne.version>2.31.0</errorProne.version>
-    <gf.version>3.8.2</gf.version>
+    <gf.version>3.8-SNAPSHOT</gf.version>
```
afabiani, line 110: SNAPSHOT dep on a release branch — the PR body acknowledges the build fails because `org.geoserver.geofence:* 3.8-SNAPSHOT` isn't in any maven repo yet. 2.28.x is a stable release line and can't ship like this.

Coordination question: do we have a target GeoFence release date for the new `getPermissionFilter` API in geoserver/geofence#287? Once it's tagged (3.8.3?), this should pin to that concrete version, not a SNAPSHOT, and the PR body should link the release. Until that release lands I'd mark the PR Draft (or block on the release in the description), so it doesn't get accidentally merged in a green-build window.

src/extension/geofence/geofence/src/main/java/org/geoserver/geofence/GeofenceAccessManager.java
```diff
@@ -970,7 +971,44 @@ private void checkStyleAllowed(AccessInfo accessInfo, String styleName) {
...
+        }
+
+        RuleFilter ruleFilter = buildPermissionRuleFilter(user);
+
+        PermsResult permsResult;
+        try {
+            permsResult = rulesService.getPermissionFilter(ruleFilter);
+        } catch (Exception e) {
+            LOGGER.log(Level.SEVERE, "Error retrieving permissions, skipping access to " + ruleFilter, e);
+            return null;
```
afabiani: The two error paths here have opposite security semantics:

- exception path → `return null` — `SecureCatalogImpl` interprets a null filter as "no security filter to apply" → fail-OPEN (everything visible)
- a few lines down, null `permsResult` → `return Filter.EXCLUDE` → fail-CLOSED (nothing visible)

Given we're closing a "layers leaking" bug, the safer default for both paths is `Filter.EXCLUDE`. The exception case (GeoFence restart, JDBC timeout, network blip) is the more common one in production and is the one with the worse failure mode if it fails open — silently reverting to the old buggy behavior.

Suggest changing this `return null` to `return Filter.EXCLUDE` for consistency.

Also worth a unit test for this path — easy with a stubbed `RuleReaderService` that throws. The current `GeofenceAccessManagerSecurityFilterTest` covers the null-`PermsResult` path but not the exception path.

afabiani:
Adding one observation from end-to-end testing of this PR against a fresh GeoServer 2.28 + PostGIS stack.

**The new `permCache` inherits the existing pattern of "no automatic invalidation on rule writes" — and the empirical impact is more visible than for the old `ruleCache`.**

Empirical timing on this branch (default cache config):

| Step | Anon Layer Preview rows |
|---|---|
| Baseline — DB empty | 0 |
| `POST /rest/geofence/rules` — `ALLOW workspace=topp` (HTTP 201) | — |
| Query Layer Preview ~1s after insert | **0** (still stale) |
| `PUT /rest/geofence/ruleCache/invalidate` (HTTP 200) | — |
| Query Layer Preview ~1s after invalidate | **5** (correct — only `topp:*`) |

And separately, with the rule already in place but a stale cache, TTL-driven expiry kicks in at **~30s** (`expireAfterWrite`) — so a sysadmin who doesn't know about the `/ruleCache/invalidate` endpoint would see rule edits "take effect after ~30s" with no warning that the lag is cache-induced.

This is **pre-existing behavior** — the older `ruleCache` (for `getAccessInfo`) has the same property — so it's not a regression introduced by this PR. But two things are worth doing in this PR specifically:

1. **`CacheController.getCacheInfo()`** ([`CacheController.java:46-77`](https://github.com/geoserver/geoserver/blob/2.28.x/src/extension/geofence/geofence/src/main/java/org/geoserver/geofence/rest/CacheController.java#L46)) currently only emits `RuleStats`, `AdminAuthStats`, `UserStats`, `ContStats`. With `permCache` added in this PR, the new cache should be visible in the same `/info` output — otherwise operators using the REST cache stats endpoint (or whatever scrapes it) won't see hit/miss/eviction counts for the new path. Trivial: add an `appendStats(sb, "PermStats", cacheManager.getPermCache());` line.

2. **Document the cache TTL ↔ rule edit lag** in the user-facing docs entry that's already needed for this PR (the unchecked "Documentation has been updated" checkbox). One sentence in `doc/en/user/source/extensions/geofence/configuration.rst` along the lines of "rule edits propagate to the catalog filter via `expireAfterWrite` (default ~30s) or by calling `PUT /rest/geofence/ruleCache/invalidate` for immediate effect" would save operators a debugging hour.

**Optional, larger:** wire `RuleAdminService.{insert,update,delete}` (or the underlying Hibernate `@PostUpdate`/`@PostPersist`/`@PostRemove` events on `Rule`/`AdminRule`) to call `cacheManager.invalidateAll()` so the plugin auto-invalidates on rule writes. This would be a clear UX improvement over the current "wait or click invalidate" pattern, but it's a behavioral change in the existing `ruleCache` flow too, so probably worth its own PR with broader discussion.

afabiani (approved):

etj:
WIth the latest commit the problematic `null` response has been removed:

https://github.com/geoserver/geoserver/pull/9448/changes/ed0e091564553e5bdaeeb76ebe00664f66771828#diff-f33784bf4aa943e6ca20bd3835142c6032b43406398e466c9ef2bdf16286ada0R984-R986

@afabiani If the currently running tests pass, I guess this PR can be merged.

We'll be releasing GeoFence 3.8.3 shortly after, and update the `gf.version` in GeoServer accordingly

petersmythe:
@etj this PR targeted 2.28, should it not have also targeted main?

https://osgeo-org.atlassian.net/browse/GEOS-12096 is still open, presumably for 3.0?

etj:
@petersmythe GeoFence is not working in GS3 yet. Porting is currently WIP, and when ready it will also contain the feature/fix in this PR.

PR #9457, opened 2026-04-27
Preview links improvement

aaime:
Hi @jodygarnett and @allyoucanmap , while I was travelling I've followed up on the ideas about improving the preview links in layer/group home pages.

Here is how it looks for a layer group:

image

and for a layer:

image

Highlights:
 - Refactors home page preview generation around a new HomePagePreviewSectionProvider extension point, so preview links can be contributed per protocol/module instead of being hardwired in one place (will eventually allow a WCS sample download too)
  - Adds tiled preview integration to the home page, reusing GWC preview target generation
  - Introduces layout choice for preview sections, allowing providers to render either classic link lists or compact dropdowns depending on the use case (explicit list for common links, dropdown for the other types)

Thoughts?

# Checklist


jodygarnett:
Hey nice work - I personally do not like the drop down, not sure why quite yet. Thinking about it. 50% It does change the visual appearance so this no longer appears like a document and 50% cannot click on link to copy URL. Stil that is just me; it was not the look I was going for.

I really like the change per protocol, and **HomePagePreviewSectionProvider**. I would of added a default method to the exsiting endpoint, but a new interface works fine. Visually I would like to move these under their respective service.

src/web/core/src/main/java/org/geoserver/web/HomePagePreviewSectionProvider.java
```diff
@@ -0,0 +1,33 @@
+/* (c) 2026 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.web;
+
+import java.util.List;
+import org.geoserver.catalog.PublishedInfo;
+
+/** Contributes a preview section to the GeoServer home page. */
```
jodygarnett, line 10: Can we have a method indicating which service heading this belongs to as part of this interface?

src/web/gwc/src/main/java/org/geoserver/gwc/web/layer/TiledPreviewSectionProvider.java
```diff
@@ -0,0 +1,54 @@
...
+    public List<PreviewLink> getLinks(PublishedInfo published) {
+        TileLayer layer;
+        try {
+            layer = GWC.get().getTileLayer(published);
+        } catch (RuntimeException e) {
+            Logging.getLogger(getClass()).fine(e.getMessage());
+            return List.of();
+        }
+        String baseURL = ResponseUtils.baseURL(GeoServerApplication.get().servletRequest());
+        return new ArrayList<>(CachedLayersPage.computePreviewTargets(layer, baseURL));
```
jodygarnett, line 42: Nice, just to confirm this is going to be the workspace (or layer) preview link - we do not want trouble if global services are off.

src/web/wms/src/main/java/org/geoserver/wms/web/WMSPreviewSectionProvider.java
```diff
@@ -0,0 +1,189 @@
...
+        return null;
+    }
+
+    private String baseURL() {
+        return ResponseUtils.baseURL(GeoServerApplication.get().servletRequest());
+    }
+
+    private String translate(String prefix, String format) {
+        Localizer localizer = GeoServerApplication.get().getResourceSettings().getLocalizer();
+        return localizer.getString(prefix + format, null, format);
```
jodygarnett, line 187: I find we are duplicating the format names in each service. I considered providing these once in core to avoid duplication.

Could be done with something like:
```
String genericFormatName = super.translate("",format);
Localizer localizer = GeoServerApplication.get().getResourceSettings().getLocalizer();
return localizer.getString(prefix + format, null, genericFormatName);
```

prushforth:
Hi @aaime  I tried this out with the MapML extension and it works well; I liked that the different structures available (Map, Feature, Tile) are separate previews.  I think this highlights the fact that data can be used in different  ways by different maps.  I agree with Jody that the two different approaches (preview link vs dropdown), although consistent with the past, is not clickable, but I understand that there are UI challenges in there.

Having the Features available led me to find [a bug in the MapML (WFS) module](https://github.com/Maps4HTML/geoserver/issues/95). I will try to address when I can.   

It would be great for a module to be able to assign a different client viewer to any one of the formats e.g. in the tile cached formats I would like to handle things that end in "mapml".  Some of these things are currently handled by a checkbox in the Edit Layer MapML settings e.g. "Use Tiles" will make the preview link try to use a tile cache if available.  

Anyway, long story short, looks good, thanks.

aaime:
Eh, the main reason I started this is that the long list of links that we have right now is messy, looks like a bad HTML page from the early 90s. 

Having tiled layers list just makes things worse, to give you an idea (mind, I just flipped the presentation mode in this PR from drodown to list, the HTML could be improved a bit, but would still be a large list of formats, see later in the comment for a better description of the issues with the current setup):

image

Mind, the idea of having the preview links in the layer home page is great, I'm just having issue with how it's implemented.

Maybe we could have the cake and eat it too by having a popover instead? Something like this (AI-generated mockup, not actual code, and ignore the CRS sections, I did not want them but haven't spent enough time arguing with the AI to remove it):

image

Or we could have a foldable panel that is normally folded, and when it unfolds, it takes enough space to have a proper display, rather than just a list of links, maybe the same contents as the popover above, but organized in a bit more compact way, I don't know, 2-4 columns, with the buttons for copy and open in new window?

Thinking more, another possibility could be a simpler list of links with 2-3 visible and then a "..." that would open a full view, with all the links. I think they should be organized in a table-ish manner though, one of the aspects that make the current list visually sloppy is that it folds on multiple lines as a text, with longer and shorter links taking different spacing:

image

What bothers me the most is that I basically have to CTRL-F to find a format in there, despite the alphabetical order, because it looks like prose, rather than a set of separate links. Also consider stuff like "GeoServer Vector Tiles". Is that one link? Or is it three links? What separates the space in between the words and the space in between links? 
Even if each link had a "tag" like a background behind it, the list would just be too long to display by default.

prushforth:
Just to be clear, I do think the original PR is an improvement.  The list of links is messy, as you say.  The distinction between viewer/preview link, format, data structure and associated CRS of the preview is complicated to manage, and I think, without going into big architectural changes, you have done well already. 

> Maybe we could have the cake and eat it too by having a popover instead? Something like this (AI-generated mockup, not actual code, and ignore the CRS sections, I did not want them but haven't spent enough time arguing with the AI to remove it):

I would not add to the challenges by saying yes to the AI design!

prushforth:
Did this get merged into main by mistake?

image

prushforth:
Not sure what's going on but I just built from main and I get this:

image

All the Vector Format links lead to a GML `<FeatureCollection>` -rooted document.  I built with this command line:

```bash
cd geoserver
git checkout main
git pull upstream main
mvn clean install -Dall -Pmapml,vectortiles,image,wcs,feature-pregeneralized,css,arcgrid,kml-out -DskipTests=true --file ./src/pom.xml -T4
```

aaime:
I don't see it in the main history:
https://github.com/geoserver/geoserver/commits/main/

aaime:
@prushforth I just tried on a version without these changes, and indeed every link in the home page produces GML instead of the format we advertise... funny thing, you're the first to notice. I guess it's confirmation that the vector download links are not used that much.

prushforth:
> I just tried on a version without these changes,

Ah, I guess that's what I meant by this being in main.  I actually hadn't noticed this UI until checking out your proposed changes.  I do prefer your changes, btw.

> funny thing, you're the first to notice. 

It's the kind of thing that could be covered by playwright tests.

> I guess it's confirmation that the vector download links are not used that much.

Perhaps; it's hard to click all the buttons and links oneself!

aaime:
Closing, superceded by https://github.com/geoserver/geoserver/pull/9494

PR #9475, opened 2026-05-01
[GEOS-12104] fix: ClassCastException escaping catch block in CatalogLoader.describe()

alexkuretz:
[GEOS-12104](https://osgeo-org.atlassian.net/browse/GEOS-12104) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

Note: I created a Jira account to file a bug report but am blocked with an "You are not authorized to perform this operation. Please log in." error message, similar to https://discourse.osgeo.org/t/cannot-create-a-jira-account-to-log-a-geoserver-geotools-issue/148684/4

## Problem

Upgrading from 2.27.3 to 2.28.3 causes GeoServer to fail startup entirely when WMS cascade layers or Coverage layers exist in the data directory, with the following error:

  RuntimeException: IllegalStateException: ExecutionException: ClassCastException:
  $ProxyN cannot be cast to WMSStoreInfo (or CoverageStoreInfo)

This is a regression introduced by the developer logging added in GEOS-12023.

Disabling Parallel Loading with `GEOSERVER_DATA_DIR_LOADER_ENABLED` = false is a workaround.

## Root Cause

Two bugs interact:

**1.** `CatalogLoader.describe()` (added in GEOS-12023) calls `resource.getStore()` to  build a log message. `WMSLayerInfoImpl.getStore()` and `CoverageInfoImpl.getStore()` override `ResourceInfo.getStore()` with a type-specific cast:

    return (WMSStoreInfo) super.getStore();

If the store field is still an unresolved `ResolvingProxy` — which happens during parallel catalog loading when the store ID cannot be matched in the catalog — that cast throws `ClassCastException`. This is the same pitfall already documented
in `ResolvingProxyResolver` (line ~270): *"gotta use rawStore() cause the subclasses will override getStore() with a cast to their concrete store types that will fail when the store is a ResolvingProxy".*

**2.** The `ClassCastException` thrown by the CONFIG-level success-log lambda inside `doAddToCatalog()` is caught correctly. However, the SEVERE failure-log lambda *also* calls `describe()`, throwing a *second* `ClassCastException` from inside the catch
block. This second exception is uncaught and propagates through `ForkJoinTask.get()` as an `ExecutionException`, wrapped in `IllegalStateException`, wrapped in `RuntimeException`, preventing the `rawCatalog` Spring bean from initializing.

## Fix

Wrap `resource.getStore()` in `describe()` with a `ClassCastException` catch. `describe()` is a logging helper only and must never propagate exceptions.

## Testing

Added regression test `wmsLayerWithUnresolvableStoreDoesNotCrashLoader` in `DataDirectoryGeoServerLoaderTest`. The test creates a WMS store and layer, then overwrites the resource XML with a bogus store ID to reliably reproduce the
unresolved-proxy condition. Verified the test fails without the fix and passes with it.

Fixes GEOS-12104
  
# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/config/datadir/CatalogLoader.java
+++ b/src/main/src/main/java/org/geoserver/config/datadir/CatalogLoader.java
@@ -415,7 +415,16 @@ class CatalogLoader {
         if (info instanceof LayerInfo layer) {
             parent = layer.getResource() != null ? layer.getResource().prefixedName() : null;
         } else if (info instanceof ResourceInfo resource) {
-            StoreInfo store = resource.getStore();
+            // WMSLayerInfoImpl.getStore() and CoverageInfoImpl.getStore() override ResourceInfo.getStore()
+            // with a cast to their concrete store type. If the store reference is still an unresolved
+            // ResolvingProxy (e.g. due to a parallel loading race), that cast throws ClassCastException.
+            // This is a logging helper only, so we swallow the exception rather than propagating it.
+            StoreInfo store = null;
+            try {
+                store = resource.getStore();
+            } catch (ClassCastException e) {
+                LOGGER.log(Level.FINE, "Could not determine parent store for {0} during catalog loading", baseName);
+            }
             parent = store != null ? store.getWorkspace().getName() + "/" + store.getName() : null;
         } else if (info instanceof StoreInfo store) {
             parent = store.getWorkspace() != null ? store.getWorkspace().getName() : null;
--- a/src/main/src/main/java/org/geoserver/config/datadir/CatalogLoader.java
+++ b/src/main/src/main/java/org/geoserver/config/datadir/CatalogLoader.java
@@ -30,6 +30,7 @@ import org.geoserver.catalog.StyleInfo;
 import org.geoserver.catalog.WorkspaceInfo;
 import org.geoserver.catalog.impl.CatalogImpl;
 import org.geoserver.catalog.impl.ResolvingProxy;
+import org.geoserver.catalog.impl.ResourceInfoImpl;
 import org.geoserver.config.GeoServer;
 import org.geoserver.config.GeoServerDataDirectory;
 import org.geoserver.config.GeoServerInfo;
@@ -414,18 +415,13 @@ class CatalogLoader {
         String parent = null;
         if (info instanceof LayerInfo layer) {
             parent = layer.getResource() != null ? layer.getResource().prefixedName() : null;
-        } else if (info instanceof ResourceInfo resource) {
-            // WMSLayerInfoImpl.getStore() and CoverageInfoImpl.getStore() override ResourceInfo.getStore()
-            // with a cast to their concrete store type. If the store reference is still an unresolved
-            // ResolvingProxy (e.g. due to a parallel loading race), that cast throws ClassCastException.
-            // This is a logging helper only, so we swallow the exception rather than propagating it.
-            StoreInfo store = null;
-            try {
-                store = resource.getStore();
-            } catch (ClassCastException e) {
-                LOGGER.log(Level.FINE, "Could not determine parent store for {0} during catalog loading", baseName);
+        } else if (info instanceof ResourceInfoImpl impl) {
+            StoreInfo store = impl.rawStore();
+            if (store != null && ResolvingProxy.getRef(store) == null) {
+                parent = store.getWorkspace().getName() + "/" + store.getName();
+            } else {
+                parent = "unresolved proxy";
             }
-            parent = store != null ? store.getWorkspace().getName() + "/" + store.getName() : null;
         } else if (info instanceof StoreInfo store) {
             parent = store.getWorkspace() != null ? store.getWorkspace().getName() : null;
         } else if (info instanceof StyleInfo style) {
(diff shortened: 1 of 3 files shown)


aaime:
@groldan and @fernandor777  thoughts?

src/main/src/main/java/org/geoserver/config/datadir/CatalogLoader.java
```diff
@@ -415,7 +415,16 @@ private <I extends CatalogInfo> String describe(I info, Function<I, String> name
...
         } else if (info instanceof ResourceInfo resource) {
-            StoreInfo store = resource.getStore();
+            // WMSLayerInfoImpl.getStore() and CoverageInfoImpl.getStore() override ResourceInfo.getStore()
+            // with a cast to their concrete store type. If the store reference is still an unresolved
+            // ResolvingProxy (e.g. due to a parallel loading race), that cast throws ClassCastException.
+            // This is a logging helper only, so we swallow the exception rather than propagating it.
+            StoreInfo store = null;
+            try {
+                store = resource.getStore();
+            } catch (ClassCastException e) {
```
fernandor777: Instead of catching the triggered exception (slower) we can detect it is an unresolved proxy and skip parent resolving, something like(just a raw dirty example):
```java
if (info instanceof ResourceInfoImpl impl) {
                StoreInfo store = impl.rawStore();
                if (store != null && ResolvingProxy.getRef(store) == null) {
                    parent = store.getWorkspace().getName() + "/" + store.getName();
                } else {
                    parent = "unresolved proxy";
                }
            }
```

We may need to solve the fact that the ```rawStore()``` method is package private, maybe with a helper method on the same package.

src/main/src/test/java/org/geoserver/config/datadir/DataDirectoryGeoServerLoaderTest.java
```diff
@@ -670,6 +672,54 @@ public void initializeDefaultStyle() throws Exception {
...
+    /**
+     * Regression test for GEOS-12023: CatalogLoader.describe() calls resource.getStore(), which subclasses like
+     * WMSLayerInfoImpl and CoverageInfoImpl override with a type-specific cast. If the store reference is an unresolved
+     * ResolvingProxy (e.g. when the store ID can't be matched during parallel catalog loading), that cast throws
+     * ClassCastException. This exception escapes the try-catch in doAddToCatalog because the SEVERE failure-log lambda
+     * also calls describe(), causing a second throw that is uncaught, propagating fatally through ForkJoin and
+     * preventing GeoServer startup.
+     */
+    @Test
+    public void wmsLayerWithUnresolvableStoreDoesNotCrashLoader() throws Exception {
```
fernandor777, line 684: Test looks able to catch the exception issue in general way, the only concern is it may be too generic, catching any form of Exception and from any other catalog resource since this catalog already has other loaded stores and layers.  I'm wondering if we can refine it by setup only the target store and layer, and being specific of what Exception we are tracking.

fernandor777 (changes requested):
Fix and test works, indeed the issue is reproducible.  I'm submitting some improvements requests where the important one is to make the test more specific catching a regression.

src/main/src/main/java/org/geoserver/config/datadir/CatalogLoader.java
alexkuretz: CatalogLoader.describe() — replaced try-catch with proxy detection

Instead of catching the ClassCastException reactively, describe() now casts to ResourceInfoImpl and calls rawStore() to get the unproxied store reference, then uses ResolvingProxy.getRef() to check whether it's still an unresolved proxy before accessing it. This matches the same pattern already used in ResolvingProxyResolver. rawStore() was also widened to public so it's accessible from CatalogLoader (different package).

src/main/src/test/java/org/geoserver/config/datadir/DataDirectoryGeoServerLoaderTest.java
alexkuretz: Test — isolated loader and explicit exception documentation

The test now uses newLoader() + a fresh CatalogImpl rather than getLoader().reload(), so it only exercises the affected WMS layer path instead of reloading the full test catalog. The comment on the assertion specifically names IllegalStateException caused by ClassCastException as what propagates without the fix:

```
// Use an isolated loader so we only exercise the WMS layer path, not the full catalog.
// Without the fix this throws IllegalStateException caused by ClassCastException.
DataDirectoryGeoServerLoader loader = newLoader();
CatalogImpl newCatalog = new CatalogImpl();
loader.postProcessBeforeInitialization(newCatalog, "catalog");
```

alexkuretz:
Thanks for the feedback @fernandor777! I'm a DevOps engineer not a Java developer, just trying to keep our Geoserver instance running well for my GIS devs. Claude helped me with this one. Let me know if there's anything else needed.

aaime:
I have created a jira ticket: https://osgeo-org.atlassian.net/browse/GEOS-12104

groldan (approved):
Thanks @alexkuretz and sorry for the late reply. The fix is appropriate and in-line with the rest of the test suite. @fernandor777 's comment was on-point and well addressed. This LGTM.

groldan:
btw, verified the patch applies cleanly to 2.28.x and added the backport label.
Rebased and amended the commit message to follow the title convention to mention the jira ticket.

PR #9476, opened 2026-05-04
[GEOS-12118] ReprojectingFeatureCollection can fail with ClassCastException while inserting CompoundCurve via WFS-T

ludotosk:
[GEOS-12118](https://osgeo-org.atlassian.net/browse/GEOS-12118) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

This fix addres a bug that happen when trying to insert a compound curve (with the column type set as org.geotools.geometry.jts.CompoundCurve) with the WFS, this issue was causing a error leading to the failure of the request. Geoserver was throwing this error:

> Insert error: class java.util.HashMap cannot be cast to class org.geotools.api.referencing.crs.CoordinateReferenceSystem (java.util.HashMap is in module java.base of loader 'bootstrap'; org.geotools.api.referencing.crs.CoordinateReferenceSystem is in unnamed module of loader org.eclipse.jetty.webapp.WebAppClassLoader @2d2fe68a)

And in the stack trace this line was present:

> at org.geoserver.feature.ReprojectingFeatureCollection.reproject(ReprojectingFeatureCollection.java:226)

So I have copied from _src/wfs-core/src/main/java/org/geoserver/wfs/UpdateElementHandler.java_  line 286 where there is a if statement for checking if the crs was a instance of CoordinateReferenceSystem, and it worked. Leaving the crs null, which was a case already handled by a if condition right after the line I have changed.

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/feature/ReprojectingFeatureCollection.java
+++ b/src/main/src/main/java/org/geoserver/feature/ReprojectingFeatureCollection.java
@@ -223,7 +223,10 @@ public class ReprojectingFeatureCollection extends DecoratingSimpleFeatureCollec
 
             if (object instanceof Geometry geometry) {
                 // check for crs
-                CoordinateReferenceSystem crs = (CoordinateReferenceSystem) geometry.getUserData();
+                CoordinateReferenceSystem crs = null;
+                if (geometry.getUserData() instanceof CoordinateReferenceSystem) {
+                    crs = (CoordinateReferenceSystem) geometry.getUserData();
+                }
 
                 if (crs == null) {
                     // no crs specified on geometry, check default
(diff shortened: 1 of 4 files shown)


aaime:
To consider the change for merge we'll need a test showing the issue, preferably a functional test inside AbstractTransactionCurveTest. The test must fail when the patch is not there, and succeed once the patch is applied.

ludotosk:
I have added a test which is similar to the existing test of the update but force a reprojection which trigger the bug, so the test crash when the bug is there and pass when the patch is applied.

aaime:
Merged. Please don't remove the checklist next time you open a PR, and make sure to go through it.

PR #9494, opened 2026-05-12
[GEOS-12063] Update formats layout in homepage, add filtering

subashtiwari1010:
[GEOS-12063](https://osgeo-org.atlassian.net/browse/GEOS-12063) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

This PR updates the layout for the formats in the home page. It also updates broken url for some formats. The filtering functionality is also introduced for the formats.
image

Expanded Map Formats:
image

Filtering the formats:
image

After copying:
image

# Checklist


aaime:
Looks lovely to me:
* All links are present in the HTML, can be scanned by crawlers
* Links open by default in new page, but can also be copied
* Usage of page real estate is controlled, but one can quickly open/fold
* The whole experience is client side based (open/fold, filter)
* Matches well the style of the rest of the page

Here is a quick video of it in action:

https://github.com/user-attachments/assets/74dbf564-1377-4d5f-ae44-a22a601c9d7a

@jodygarnett @petersmythe @groldan @jratike80, what do you think? Better than what we have now?

petersmythe:
I think it's beautiful - superb job @subashtiwari1010 !

If anything, I am not so keen on the `...less` (once expanded, I think you can leave it expanded) and I would add a space between `...` and `more`, but that is really just nit-picking.

However, for consistency (ref [Git issue about links and buttons](https://github.com/geoserver/gs3/issues/120)), it should not look like a link, but rather an action button, shouldn't it?

And how does it look on a narrow mobile screen?

nmco:
+1 from me looks really nice. Only a couple of comments:
* I would expect GeoJSON as  one of te common formats - quite useful when dealing with vector data.
* I would remove the "more formats" label \ text:
image
* I don't have a strong opinion about the "more" and "less" - I like the "more" but the "less" a bit less.

jratike80:
> Better than what we have now?

Absolutely.

jodygarnett:
Wow that is fun! And thanks for taking the time to make this searchable etc...
I especially appreciate the change for the Tiled Preview 💯 

I would like to consider how the metadata and download links, keywords, and other non-preview things play into this mix? Like for example on the `ne:countries` layer?

Feedback:
- Copy button is lovely, but the service links would like them also?  
  The service links have the same function as these format links; perhaps they could visually use the same rounding angle 
- I did not expect a new heading "Preview maps and data". I see "Feature sample" is used to try and address the expectation of full-data-download vs sample-data-download.
- Not sure the horizontal line is necessary as each section has its own label?  
- "More" / "Less" is on a different right hand side location each section.  
  Idea: Use left hand side ▶️ / 🔽 for consistent click location?  
- The text `...more` is awkward not agreeing with `...` use. If you stick with text: `more` and `less` is common enough AI recommends it, although prefers icons.

aaime:
@jodygarnett without the lines on a wide screen it got hard to associate the links to their label. Pheraphs they can be made fainter, but I would not remove them completely.

jodygarnett:
So the metadata and download links shown on this page:
- https://docs.geoserver.org/main/en/user/webadmin/welcome/#workspace-web-services

This should also be updated to use the same look and feel? Presently they are just provided as links.

The layers in `ne` have been filled in with `application/zip` download links, linking to website download, and metadata links in `text/html` linking to website description of the content.

The order was established so that metadata was at the  top of the contributions, and the download links were last.

image

aaime:
@subashtiwari1010 in other words, we need these two links to be part of the "preview maps and data" by rolling a link provider for them too:

image

This is a intersting spin, in that the links might just be missing, in that case I'd skip their section entirely (so that no "Data links" would show, rather than an empty data links section).

src/web/core/src/main/java/org/geoserver/web/PreviewSectionLayout.java
```diff
@@ -0,0 +1,11 @@
+/* (c) 2026 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.web;
+
+/** Supported layouts for a home page preview section. */
+public enum PreviewSectionLayout {
```
aaime: This enum and its usage is effectively dead, there is no way to get link dropdowns anymore. Please remove.

src/web/demo/src/main/java/org/geoserver/web/demo/PreviewHomePageContentProvider.java
```diff
@@ -55,241 +54,116 @@ public Component getPageBodyComponent(String id) {
...
-            LoadableDetachableModel<List<ExternalLink>> previewLinks = new LoadableDetachableModel<>() {
-                @Override
-                protected List<ExternalLink> load() {
-                    GeoServerHomePage homePage = (GeoServerHomePage) PreviewPanel.this.getPage();
-                    PublishedInfo layerInfo = homePage.getPublishedInfo();
-
-                    List<ExternalLink> links = vectorFormatLinks(new PreviewLayer(layerInfo));
-                    vectorPreview.setVisible(!links.isEmpty());
-                    return links;
+                    addCatalogLinkSections(result, published);
```
aaime: Hum no, they should have been added as their own HomePagePreviewSectionProvider. 
Suggestion: turn PreviewCatalogLinkSupport into a HomePagePreviewSectionProvider.
For the priority, don't hard-code the numbers in a specific place, have the PreviewSectionProvider implement ExtensionPriority instead and GeoServerExtensions.extensions will return them already sorted.
You can then remove the two sorts in this method, too.

src/web/gwc/src/main/java/org/geoserver/gwc/web/layer/TiledPreviewSectionProvider.java
```diff
@@ -0,0 +1,54 @@
...
+    public List<PreviewLink> getLinks(PublishedInfo published) {
+        TileLayer layer;
+        try {
+            layer = GWC.get().getTileLayer(published);
+        } catch (RuntimeException e) {
+            Logging.getLogger(getClass()).fine(e.getMessage());
+            return List.of();
+        }
+        String baseURL = ResponseUtils.baseURL(GeoServerApplication.get().servletRequest());
+        return new ArrayList<>(CachedLayersPage.computePreviewTargets(layer, baseURL));
```
aaime: Is there any need for a defensive ArrayList copy here?

aaime (changes requested):
Tested, looks and works good, thanks!

For long term maintenance sake, please address the feedback below:

petersmythe:
Nudge @subashtiwari1010 , can you update and then merge, please?

aaime:
Looks good, merging.

PR #9495, opened 2026-05-12
[GEOS-12063] Fix/preserve context when navigating

subashtiwari1010:
[GEOS-12063](https://osgeo-org.atlassian.net/browse/GEOS-12063) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

This PR updates context for some cases when navigating.
When cancel, save buttons are clicked, it brings back to the list/table page previously, but now it will take you to the previous page.

# Checklist


aaime:
It does not seem to be working, for me? I go editing a layer group and a layer from their respective home pages, and I don't get back to them:

https://github.com/user-attachments/assets/f46de941-47ed-4667-8969-57378dcefcc3

src/web/core/src/main/java/org/geoserver/web/GeoServerBasePage.java
```diff
@@ -803,6 +815,22 @@ protected void doReturn(Class<? extends Page> defaultPageClass) {
         setResponsePage(defaultPageClass, currentParams);
     }
 
+    public static void setReturnDestination(
+            Session session, Class<? extends Page> pageClass, PageParameters pageParameters) {
+        session.setMetaData(RETURN_DESTINATION_KEY, new ReturnDestination(pageClass, pageParameters));
```
aaime: This is problematic... it's not uncommon for a user to have multiple tabs open against the same geoserver UI, they would share the same session and get confused on their return targets.

Here is a sketch of a different approach, using page parameters instead:

```java

  // HomeReturnPageLink.onClick()
  @Override
  public void onClick() {
      PageParameters params = new PageParameters(targetParams);
      params.add("_returnClass", GeoServerHomePage.class.getName());
      if (homeReturnParams != null) {
          String ws = homeReturnParams.get("workspace").toOptionalString();
          if (ws != null) params.add("_returnWs", ws);
      }
      setResponsePage(pageClass, params);
  }

  // GeoServerBasePage.doReturn() - check BEFORE session metadata
  String returnClass = getPageParameters().get("_returnClass").toOptionalString();
  if (returnClass != null) {
      PageParameters retParams = null;
      String ws = getPageParameters().get("_returnWs").toOptionalString();
      if (ws != null) retParams = new PageParameters().add("workspace", ws);
      setResponsePage(GeoServerHomePage.class, retParams);
      return;
  }

```

It's hard-coded on the home page for now, but it's the target that we were looking for... although if you think you can generalize, it would be best.

aaime (changes requested):
Tested and was happy it worked... until I found it does not.
Try this:
* Open two tabs
* In the first tab, go to the topp:states home page and clik edit layer
* In the second tab, go to the nurc:Arc_Sample2 home page and click edit layer
* Now go back to the first tab and save the layer.... you'll find yourself in the nurc:Arc_Sample2 home page.

This is due to tabs sharing the session, see below.

src/web/core/src/main/java/org/geoserver/web/GeoServerBasePage.java
```diff
@@ -803,6 +845,60 @@ protected void doReturn(Class<? extends Page> defaultPageClass) {
...
+                String paramName = key.substring(RETURN_PARAM_PREFIX.length());
+                String value = source.get(key).toOptionalString();
+                if (value != null) {
+                    returnParams.add(paramName, value);
+                }
+            }
+        }
+        try {
+            @SuppressWarnings("unchecked")
+            Class<? extends Page> pageClass = (Class<? extends Page>) Class.forName(returnClass);
```
aaime: This is dangerous and can to remote code execution, as the return class value is decided by whoever composes the URL. Use an enumerated whitelist instead.

src/web/core/src/main/java/org/geoserver/web/GeoServerBasePage.java
```diff
@@ -116,6 +129,26 @@ protected GeoServerBasePage() {
         commonBaseInit();
     }
 
+    /**
```
aaime, line 133: I'd suggest having GeoServerBasePage(PageParameters) automatically call this method, ensures the method is called without having to remind each subclass to do so. Still needs to be called by subclasses taking PageParameters but then calling onto the GeoServerBasePage() empty parameter constructor (so please adapt the javadoc accordingly)

aaime (commented):
Works fine thanks. There code still contains a serious security issue, as well as a smaller maintainability issue, see below.

aaime:
Looks good, merging

PR #9501, opened 2026-05-13
[GEOS-12136] IOTestUtils.createRandomDirectory() replacing mkdir call to more recent java.nio.files API

pmauduit:
[GEOS-12136](https://osgeo-org.atlassian.net/browse/GEOS-12136) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

File.mkdir is known not to be thread-safe. Replacing the call by the `java.nio.files`  Java api which is more recent. The impact should be quite limited though, as grepping the codebase reveals that the method is only called in the testsuite.

Removing the suffix parameter from the `IOUtils` static method, as the `createTempDirectory` does not support passing a suffix.
  

# Checklist

### DIFF
--- a/src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java
+++ b/src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java
@@ -149,7 +149,7 @@ public abstract class AbstractAppSchemaMockData extends SystemTestData implement
 
     static File newRandomDirectory() {
         try {
-            return IOUtils.createRandomDirectory("target", "app-schema-mock", "data");
+            return org.geoserver.test.IOUtils.createRandomDirectory("target", "app-schema-mock");
         } catch (IOException e) {
             throw new RuntimeException(e);
         }
--- a/src/extension/kml/src/test/java/org/geoserver/kml/KMLReflectorTest.java
+++ b/src/extension/kml/src/test/java/org/geoserver/kml/KMLReflectorTest.java
@@ -488,7 +488,7 @@ public class KMLReflectorTest extends WMSTestSupport {
         assertEquals(KMZMapOutputFormat.MIME_TYPE, response.getContentType());
 
         // create the kmz
-        File tempDir = org.geoserver.util.IOUtils.createRandomDirectory("./target", "kmplacemark", "test");
+        File tempDir = org.geoserver.test.IOUtils.createRandomDirectory("./target", "kmplacemark");
         tempDir.deleteOnExit();
 
         File zip = new File(tempDir, "kmz.zip");
--- a/src/main/src/test/java/org/geoserver/test/IOUtils.java
+++ b/src/main/src/test/java/org/geoserver/test/IOUtils.java
@@ -1,0 +1,22 @@
+/* (c) 2026 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.test;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public class IOUtils {
+
+    /**
+     * Creates a directory as a child of baseDir. The directory name will be preceded by prefix and followed by suffix
+     */
+    public static File createRandomDirectory(String baseDir, String prefix) throws IOException {
+        Path tmpPath = java.nio.file.Files.createTempDirectory(Paths.get(baseDir), prefix);
+
+        return tmpPath.toFile();
+    }
+}
--- a/src/platform/src/main/java/org/geoserver/util/IOUtils.java
+++ b/src/platform/src/main/java/org/geoserver/util/IOUtils.java
@@ -20,10 +20,10 @@ import java.io.FilenameFilter;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.nio.file.Path;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.concurrent.ThreadLocalRandom;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.zip.ZipEntry;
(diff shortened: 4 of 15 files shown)


src/platform/src/main/java/org/geoserver/util/IOUtils.java
```diff
@@ -157,14 +159,12 @@ public static void deepCopy(File fromDir, File toDir) throws IOException {
...
-     */
-    public static File createRandomDirectory(String baseDir, String prefix, String suffix) throws IOException {
-        File tempDir = File.createTempFile(prefix, suffix, new File(baseDir));
-        tempDir.delete();
-        if (!tempDir.mkdir()) throw new IOException("Could not create the temp directory " + tempDir.getPath());
-        return tempDir;
+    /** Creates a directory as a child of baseDir. The directory name will be preceded by prefix */
+    public static File createRandomDirectory(String baseDir, String prefix) throws IOException {
+        Path basePath = Paths.get(baseDir);
+        Path tempDirPath = java.nio.file.Files.createTempDirectory(basePath, prefix);
```
jodygarnett: This is only a partial solution, from the javadocs:

> As with the createTempFile methods, this method is only part of a temporary-file facility. A [shutdown-hook](https://docs.oracle.com/javase/8/docs/api/java/lang/Runtime.html#addShutdownHook-java.lang.Thread-), or the [File.deleteOnExit()](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#deleteOnExit--) mechanism may be used to delete the directory automatically.

jodygarnett:
```suggestion
        Path tempDirPath = java.nio.file.Files.createTempDirectory(basePath, prefix);
        tempDirPath().deleteOnExit();
```

jodygarnett (changes requested):
Files only creates a good random path, need to call `File.deleteOnExit()` also (or make your own shutdown hook).

src/platform/src/main/java/org/geoserver/util/IOUtils.java
pmauduit: I thought it was the caller's responsability to call `deleteOnExit()`, as I saw it in the KML testsuite:
https://github.com/geoserver/geoserver/blob/main/src/extension/kml/src/test/java/org/geoserver/kml/KMLReflectorTest.java#L489

But indeed, the `deleteOnExit` is not called everywhere on the other files I modified, I don't see any issue of doing the call into the utility class, I'll adapt consequently :+1: .

pmauduit: @jodygarnett Revisiting my modifications, we can add a `deleteOnExit` call, but I suspect that it will delete the directory only if it is totally empty, and won't recurse on the actual content. The Javadoc does not mention this case though: https://docs.oracle.com/javase/8/docs/api/java/io/File.html#deleteOnExit--

Checking the sources again, it still looks like the caller is responsible for the deletion, e.g.:
https://github.com/geoserver/geoserver/blob/main/src/extension/app-schema/sample-data-access-test/src/test/java/org/geoserver/test/SampleDataAccessMockData.java#L115-L124

Maybe we should be more explicit about it ?

bradh: Looks like this deprecated method is going to cause the CI checks to fail.

aaime: There are various problems with using deleteOnExit():

* The JDK code register all the paths in a DeleteOnExitHook. Currently this method is only called by tests and it's thus not too problematic, but this is a public method. If any hot path starts invoking it, in the future, we'll have a slow but unbounded memory leak
* The  DeleteOnExitHook will just call ``new File(path).delete()``, which does not work on directories unless they are empty --> in a number of cases, the deletion will just not happen

Suggested approach:
* Kill this method entirely
* Switch the tests using it towards TemporaryFolder instead

pmauduit: > * Switch the tests using it towards TemporaryFolder instead

This is also what @groldan suggested

pmauduit: > Kill this method entirely

I created another working branch where I totally removed the method, but I was not expecting that it would lead me to rework the entire app-schema-test module: one of the issue I encountered is that some of the classes which are making use of the former createTempDirectory method are not managed by Junit, so my approach was to provide to each `*MockData` classes a File object from a Junit `TempFolder` object. But it looks like the createTestData method is called before Junit creates an actual temporary directory, leading to the following error while running the testsuite:

```
IllegalState the temporary folder has not yet been created
```
the current diff is here:
https://github.com/geoserver/geoserver/compare/main...pmauduit:use-temfolder-junit-constructs?expand=1#diff-4e39d170241f66c18c3d22c2809e4c8acb8f8ffb9874181e2a15aad0c64b00d8

but I am asking myself if I am on the right track or if I am out of scope already.

aaime: Question: does Sonar check test code too? If not, the easy approach is to move the method we want to remove into a app-schema test specific helper.

pmauduit: why not having a IOUtils class into the gs-main-test module with the method being moved there ? In my opinion, getting through junit helpers (TemporaryFolder) could be a topic for another PR.

petersmythe:
@pmauduit , I believe [you are waiting](https://discourse.osgeo.org/t/geotools-geoserver-pmc-meeting-notes-2026-06-02/153958#p-445983-pr-review-catch-up-8) for @jodygarnett to review this.  Is it intended for the GeoServer 3.0 release, or can it be delayed to give us all a bit of breathing room?  If it's urgent, I think you should ask someone else for a review.

src/extension/app-schema/app-schema-test/src/test/java/org/geoserver/test/AbstractAppSchemaMockData.java
petersmythe, line 1: These should be imports rather, I believe, especially because of the change from

import org.geoserver.util.IOUtils;
to
_______ org.geoserver.test.IOUtils

And a comment explaining the difference would probably be helpful to our future selves.

src/extension/kml/src/test/java/org/geoserver/kml/KMLReflectorTest.java
```diff
@@ -488,7 +488,7 @@ protected void doTestRasterPlacemark(boolean doPlacemarks) throws Exception {
         assertEquals(KMZMapOutputFormat.MIME_TYPE, response.getContentType());
 
         // create the kmz
-        File tempDir = org.geoserver.util.IOUtils.createRandomDirectory("./target", "kmplacemark", "test");
+        File tempDir = org.geoserver.test.IOUtils.createRandomDirectory("./target", "kmplacemark");
```
petersmythe, line 491: Ditto - import

src/main/src/test/java/org/geoserver/test/IOUtils.java
```diff
@@ -0,0 +1,22 @@
...
+ * application directory.
+ */
+package org.geoserver.test;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public class IOUtils {
```
jodygarnett, line 12: Perhaps call this IOTestUtils to ease confusion when we see this in code.

jodygarnett (approved):
Minor feedback, going to see if I can modify the branch

jodygarnett:
Manually merged in order to provide GEOS-12136 commit message and change name to IOTestUtils

PR #9502, opened 2026-05-13
[GEOS-12110] Make use of XMLUtils for better integration with GeoTools Hints

jodygarnett:
[GEOS-12110](https://osgeo-org.atlassian.net/browse/GEOS-12110) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

Makes use of the new `XMLUtils` utility methods introduced by the following PR:

* https://github.com/geotools/geotools/pull/5639

This will need to be merged at the same time. I have named the branches the same, perhaps we can adjust the "downstream integration workflow" to look for same branch to check out.

Two internal API changes to note:

- EntityResolverProvider now uses `NullEntityResolver` when `ENTITY_RESOLUTION_UNRESTRICTED` is configured.
- `EntityResolverProvider.RESOLVE_DISABLED_PROVIDER` now provides an EntityResolver throwing an IllegalStateException, comments at that line indicate tests are required to call `setEntityProvider` prior to parsing.

# Checklist


src/main/src/test/java/org/geoserver/test/DevModeEntityResolver.java
```diff
@@ -0,0 +1,76 @@
+package org.geoserver.test;
```
aaime, line 5: Lacks copyright header

src/main/src/main/java/org/geoserver/util/EntityResolverProvider.java
```diff
@@ -50,12 +52,15 @@ public EntityResolverProvider(GeoServer geoServer) {
...
+     * <p>This method will also stage the provided EntityResolver to {@code GeoTools.init(hints)}.
+     *
      * <p>Primarily used to stage an EntityResolver for test cases or local development.
      *
-     * @param resolver Entity resolver
+     * @param resolver Entity resolver, if {@code} then {@code NullEntityResolver.INSTANCE} is used.
      */
     public static void setEntityResolver(EntityResolver resolver) {
-        entityResolver = resolver;
+        entityResolver = resolver != null ? resolver : NullEntityResolver.INSTANCE;
```
aaime, line 70: This seems the only non-test code initializing the geotools system hints, but the method is called only by test code. I was expecting to see some initialization of this hint as early as possible during the startup, since XMLUtils ends up using the GeoTools hints. And so any call to XMLUtils is probably ignoring the GeoServer entity resolver too? 

Btw, I see it was attempted in GeoServerInitStartupListener, but then commented out.

src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java
```diff
@@ -294,16 +239,20 @@ protected final void setUp(SystemTestData testData) throws Exception {
 
         // setup quiet logging (we need to to this here because Data
         // is loaded before GeoServer has a chance to setup logging for good)
-        try {
-            Logging.ALL.setLoggerFactory(Log4J2LoggerFactory.getInstance());
-        } catch (Exception e) {
-            LOGGER.log(Level.SEVERE, "Could not configure log4j2 logging redirection", e);
-        }
+        //        try {
```
aaime: Please don't leave commented out code...

src/main/src/test/java/org/geoserver/test/DevModeEntityResolver.java
```diff
@@ -0,0 +1,76 @@
...
+        path = path.toLowerCase();
+        // Match the GeoTools locations having schemas we resolve against
+        return path.matches(".*modules[\\\\/]extension[\\\\/]xsd[\\\\/].*\\.xsd")
+                || path.matches(".*modules[\\\\/]ogc[\\\\/].*\\.xsd");
+    }
+
+    private boolean isDataDirectory(String systemId) {
+        if (GeoServerSystemTestSupport.applicationContext != null) {
+            GeoServerDataDirectory dd =
+                    GeoServerSystemTestSupport.applicationContext.getBean(GeoServerDataDirectory.class);
```
aaime, line 89: This seems like a bit too much coupling, it would be interesting to use it for example in Start.java too (running GeoServer from sources). Maybe use GeoServerExtensions.getBean instead? (I know it's not great, but for a dev support class it's not so bad either).

src/main/src/main/java/org/geoserver/GeoserverInitStartupListener.java
```diff
@@ -88,6 +88,15 @@ public void contextInitialized(ServletContextEvent sce) {
 
         LOGGER = Logging.getLogger("org.geoserver.logging");
         LOGGER.config("Logging policy: " + policy);
+
+        // GeoTools Library initialization
```
aaime, line 94: This seems to be a problem, see other comments.

aaime (commented):
Most of the changes look good, but please check carefully the comments below, I might have spotted a significant problem.

src/main/src/test/java/org/geoserver/test/DevModeEntityResolver.java
jodygarnett: I just pulled this out of another class to give it a name when stepping through in the debugger, I did not really look at this implementation. 


It was previously an anonymous class GeoServerSystemTestSupport.RESOLVE_DISABLED_PROVIDER_DEVMODE

Adjusting so this is more standalone.

src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java
jodygarnett: Code restored, was troubleshoonintg GeoTools init startup

src/main/src/main/java/org/geoserver/util/EntityResolverProvider.java
jodygarnett: Yes I will restore the one in GeoServerInitStartup, both approaches can now work together I believe.

src/main/src/main/java/org/geoserver/GeoserverInitStartupListener.java
jodygarnett: This code is now cleaned up and restored

jodygarnett:
Thanks for review @aaime feedback addressed, also added a bunch of PMD rules and cleaned up the resulting mess.

@pmauduit I would appreciate you review and feedback from automated tools on this branch, and have requested your review.

I have modified https://github.com/geotools/geotools/pull/5639 so its integration tests also test this geoserver branch so we can get automated assurance that this builds cleanly prior to merging.

build/qa/pmd-ruleset.xml
pmauduit, line 1: checking automated tools reports, there could also be the following to catch:
* `javax.xml.stream.XMLInputFactory.newFactory()`
* `org.jdom2.input.SAXBuilder()`

These have been reported on GT (I haven't run an analysis GS-side on this branch), but grepping for jdom into the codebase seems to reveal it can be used by some modules

pmauduit (approved):

src/main/src/test/java/org/geoserver/test/DevModeEntityResolver.java
jodygarnett: @aaime I tried making a setter and having GeoServerSystemTestSupport inject the support context, sadly it did not work. Since this is test classes I am leaving it well enough alone and focusing on another section which is inconsistent.

jodygarnett:
Okay I have found an inconsistency, but I am not sure how long it has been present.

Some parsing code takes a `EntityResolverProvider.RESOLVE_DISABLED_PROVIDER` which is configured with a partially setup AllowListEntityResolver (that does not have a GeoServer to determine data directory).

However this is being ignored in favour of DevModeEntityResolver configured at the GeoTools level, and things pass. When I actually smash in the intended AllowListEntityResolver I get inconsistent failures between maven and IDE making me grumpy - strangely it cannot access `jar:file` URLs which I would expect to work.

Okay - this is used to wire up during `applicationContext.xml` processing; may have to break this cycle if the setting is being ignored.

jodygarnett:
> Some parsing code takes a `EntityResolverProvider.RESOLVE_DISABLED_PROVIDER` which is configured with an partially setup AllowListEntityResolver (that does not have a GeoServer to determine data directory).

I updated the code so `EntityResolverProvider.RESOLVE_DISABLED_PROVIDER` throws IllegalStateException if actually used, with a comment in the code that use of setEntityResolver is required.

aaime:
All green and the PR looks good, merging. There is a typo but I'll make a separate PR to address it, right now we need to get the build back on green before it triggers more confusion.

jodygarnett:
wow thanks @aaime I am so happy this is merged, let's not do this again.

PR #9506, opened 2026-05-16
[GEOS-11581] Set up leaner attribute transformations when attribute customization is enabled

aaime:
[GEOS-11581](https://osgeo-org.atlassian.net/browse/GEOS-11581) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

Needs https://github.com/geotools/geotools/pull/5647

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/TransformFeatureTypeCallback.java
+++ b/src/main/src/main/java/org/geoserver/catalog/TransformFeatureTypeCallback.java
@@ -18,10 +18,12 @@ import org.geotools.api.data.SimpleFeatureSource;
 import org.geotools.api.feature.Feature;
 import org.geotools.api.feature.type.FeatureType;
 import org.geotools.api.filter.Filter;
+import org.geotools.api.filter.FilterFactory;
 import org.geotools.api.filter.expression.Expression;
 import org.geotools.api.util.InternationalString;
 import org.geotools.data.transform.Definition;
 import org.geotools.data.transform.TransformFactory;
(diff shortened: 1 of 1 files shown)


src/main/src/main/java/org/geoserver/catalog/TransformFeatureTypeCallback.java
```diff
@@ -75,7 +79,12 @@ private DataStore getDataStore(FeatureTypeInfo fti) throws IOException {
     private Definition toDefinition(AttributeTypeInfo ati) {
         try {
             String name = ati.getName();
-            Expression source = ECQL.toExpression(ati.getSource());
+            Expression source;
+            if (ati.getRawSource() == null) {
+                source = FF.property(name);
+            } else {
+                source = ECQL.toExpression(ati.getSource());
+            }
```
dromagnoli, line 87: I'm not familiar with this. 
Is it ok to check on the getRawSource and then use the getSource?

Beside my doubt, it looks good to me.

dromagnoli (approved):
Looks good to me.
I have made a question but it's only to clarify my doubts.

src/main/src/main/java/org/geoserver/catalog/TransformFeatureTypeCallback.java
aaime: Ah yes, "source" is a potentially quoted version of rawSource, if rawSource is null so is source.

PR #9528, opened 2026-05-27
[GEOS-12123] Login failure in FIPS mode (Oracle Linux 8): SHA1PRNG

mortenlud:
[GEOS-12123](https://osgeo-org.atlassian.net/browse/GEOS-12123) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

When running GeoServer on Oracle Linux 8 with FIPS mode enabled, login attempts fail with an exception originating from Jasypt password handling.

To fix the bug i have changed the GeoServerDigestPasswordEncoder to use a new LegacyPasswordByteDigester and not use  Jasypt lib, since the bug originate in Jasypt.
  

# Checklist

- [ x] I have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/main/CONTRIBUTING.md).
- [ x] I have sent a [Contribution Licence Agreement](https://docs.geoserver.org/latest/en/developer/policies/committing.html) (not required for small changes, e.g., fixing typos in documentation).

- [ x] New unit tests have been added covering the changes.
- [x ] There is an issue in the [GeoServer Jira](https://osgeo-org.atlassian.net/browse/GEOS/summary) (except for changes that do not affect administrators or end users in any way).
- [ x] Commit message(s) must be in the form ``[GEOS-XYZWV] Title of the Jira ticket``.
- [ x] Bug fixes and small new features are presented as a single commit.


aaime:
I haven't looked at the code changes yet, but can you please explain how this affects existing data directories?
Does it affect only stores, or does it also extend to other parts (e.g., user password handling)?
The project tries to avoid backward-incompatible changes, especially when they are made to address specific use cases at the expense of others.
As a project, we also try to ensure that an older GeoServer can open a data directory modified by a newer one whenever possible, because upgrading and then rolling back due to an accidental incompatibility is a common occurrence (e.g., users leveraging an OGC compliance bug that we fixed afterwards).

If the change is not backwards-compatible, it's not a deal-breaker, and it can still be managed. A preferable approach would be to create an extension point in the core modules and then create a FIPS extension (a new optional module) that people interested in running in FIPS mode can install, with a clear indication that they will need to start from an empty data directory.

mortenlud:
The changes I’ve made are limited to replacing the use of the Jasypt
library in one class, while preserving the same encryption format and
behavior. The goal has been specifically to avoid any backward
compatibility issues.

In short, this is intended to be a drop-in replacement of the hash
implementation rather than a behavioral or format change.

I will remove tis pull request, and add some more test to ensure the backward compapility.

PR #9537, opened 2026-05-30
[GEOS-12126] Use LinkedHashSet for deterministic XML config output

petersmythe:
[GEOS-12126](https://osgeo-org.atlassian.net/browse/GEOS-12126) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

https://osgeo-org.atlassian.net/browse/GEOS-12126

Use LinkedHashSet instead of HashSet for persisted Set fields in config model
classes to ensure deterministic XML output. Previously, HashSet fields (e.g.
layer styles, WFS output types, WMS MIME types) were serialized in
non-deterministic iteration order, causing spurious diffs when config files are
re-written.

The fix changes field initializers in the config model classes from
`new HashSet<>()` to `new LinkedHashSet<>()`, which preserves insertion order.
Since XStream deserializes elements in document order, a round-trip
(read → write) produces identical output — no spurious diffs.

A source-scanning lint test (`LinkedHashSetLintTest`) is included to catch
future regressions: it walks all `*InfoImpl.java` files and fails if any
non-excluded file contains `new HashSet<>`.

## Checklist


petersmythe:
Waiting on https://github.com/geoserver/geoserver/pull/9538 to fix the broken tests

src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
```diff
@@ -803,8 +803,27 @@ public void visit(String name, Class type, Class definedIn, Object value) {
 
                 // unwrap any proxies
                 value = unwrapProxies(value);
+
+                // sort Set fields for deterministic XML output
+                // Skip SortedSet (e.g. TreeSet) — those already have a defined order and
+                // replacing them with LinkedHashSet would drop the comparator from XML
+                if (value instanceof java.util.Set<?> set
```
aaime: Please avoid fully qualified references, add an import instead... it's verbose and a common sign of AI-generated code (commented here, but please fix in the whole diff)

src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
```diff
@@ -803,8 +803,27 @@ public void visit(String name, Class type, Class definedIn, Object value) {
...
                 wrapped.visit(name, type, definedIn, value);
             }
+
+            private static java.util.Set<?> sortSet(java.util.Set<?> set) {
+                java.util.List<Object> sorted = new java.util.ArrayList<>(set);
+                sorted.sort(
+                        java.util.Comparator.<Object, String>comparing(
+                                        o -> o == null ? "" : o.toString())
+                                .thenComparingInt(System::identityHashCode));
+                return new java.util.LinkedHashSet<>(sorted);
```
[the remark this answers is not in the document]
aaime: Agreed

src/main/src/test/java/org/geoserver/config/util/XStreamPersisterSetOrderTest.java
```diff
@@ -0,0 +1,168 @@
+/* (c) 2024 Open Source Geospatial Foundation - all rights reserved
```
aaime: Since it's a new file, it needs the year of creation (while there is no need to bother changing the year in existing files)

```suggestion
/* (c) 2026 Open Source Geospatial Foundation - all rights reserved
```

aaime (changes requested):
@peter-afrigis this seems another case of "I have one case in front of my eyes and I'm fixing it", like in https://github.com/GeoWebCache/geowebcache/pull/1536

I'm sure it works in the case you're specifically trying to address, but what if the set already was a LinkedHashSet with a manually configured order? You're changing it here. Rather than adding one more exclusion, concentrate on HashSet, it's the common Set implementation that does not have a predictable iteration order.

Also, the approach is brittle, toString is not guaranteed to have a sensible implementation in all objects.  Default Object.toString() is ClassName@hexHashCode — hash code is JVM-run-specific, so sort order is non-deterministic across restarts for any object that doesn't override toString(). The fix only  works for types that have meaningful toString() implementations.

Rather than this, please go hunt the places that are using a non-stable set, like LayerInfoImpl.styles, and make them use a LinkedHashSet intead. 

If someone goes and changes the order manually, then the GetCapabilites will also return a different response, so it's a legit config change.

src/main/src/test/java/org/geoserver/config/util/LinkedHashSetLintTest.java
```diff
@@ -0,0 +1,140 @@
...
+            // Runtime utility/fallback returns, not field initializers
+            "GeoServerTileLayerInfoImpl.java",
+            // Error fallback returning mutable empty set, not a persisted field
+            "WMSLayerInfoImpl.java");
+
+    /** Pattern matching {@code new HashSet<>()} or {@code new HashSet<>(someArg)} in source. */
+    private static final Pattern HASHSET_PATTERN = Pattern.compile("new\\s+HashSet<");
+
+    @Test
+    public void testNoHashSetInInfoImplClasses() throws IOException {
```
jodygarnett: Huh, I would expect to see this as a maven plugin rather than as a unit test.

For examples see: https://github.com/geotools/geotools/tree/main/build/maven/enforce-managed-dependencies

Actually you can do it with a PMD rule - see: https://github.com/geoserver/geoserver/blob/main/build/qa/pmd-ruleset.xml

```xml

  <rule name="AvoidHashMap"
        language="java"
        message="Do not use HashMap for factory config objects, prefer LinkedHashSet in order to preserve order."
        class="net.sourceforge.pmd.lang.rule.xpath.XPathRule">
    <description>
      Notice use of new HashMap() in factory config objects, prefer LinkedHashSet in order to preserve order
    </description>
    <priority>3</priority>
    <properties>
      <property name="xpath">
        <value>
          <![CDATA[
              //MethodCall[pmd-java:matchesSig("java.util.HashMap#HashMap()")]
              ]]>
        </value>
      </property>
    </properties>
  </rule>
```

Have to look up how to apply it only to the specific list of files.

jodygarnett (changes requested):
Hi @petersmythe - the change to LinkedHashSet seems fine,

Can you remove LinkedHashSetLintTest please, you can do a seperate PR to add a PMD rule covering the same effect. That is easier than maintaining a unit test, or writing a new maven plugin.

src/main/src/test/java/org/geoserver/config/util/LinkedHashSetLintTest.java
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * Source-scanning lint test that finds {@code new HashSet<>()} usage in persisted config model classes
+ * ({@code *InfoImpl.java}). These fields should use {@code LinkedHashSet} to preserve insertion order and produce
+ * deterministic XML output when serialized by XStreamPersister.
+ *
+ * <p>If this test fails, change the field initializer from {@code new HashSet<>()} to {@code new LinkedHashSet<>()}.
+ *
+ * <p>Fields that are genuinely transient/runtime (not serialized) can be added to {@link #EXCLUDED_PATTERNS}.
+ */
+public class LinkedHashSetLintTest {
```
jodygarnett: Now that https://github.com/geoserver/geoserver/pull/9565 is providing a PMD check can this LinkedHashSetLintTest be removed from the PR?

jodygarnett: Going ahead and removing this so the change can be merged, will let build complete first.

src/main/src/main/java/org/geoserver/config/util/XStreamPersister.java
```diff
@@ -803,6 +803,7 @@ public void visit(String name, Class type, Class definedIn, Object value) {
 
                 // unwrap any proxies
                 value = unwrapProxies(value);
+
```
jodygarnett:
```suggestion 
```

PR #9542, opened 2026-05-31
[GEOS-12128] Compute from SRS bounds returns valid bbox for EPSG:404000

petersmythe:
[GEOS-12128](https://osgeo-org.atlassian.net/browse/GEOS-12128) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

When clicking 'Compute from SRS bounds' for a layer with EPSG:404000 (the generic wildcard CRS), the bounding box was left as the JTS null envelope (0, 0, -1, -1) because CRS.getEnvelope() returns null for an engineering CRS with no domain of validity.

Fix: detect wildcard engineering CRS in getBoundsFromCRS() and ~return the full WGS84 extent (-180, -90, 180, 90) as a practical fallback~. Add info message in the UI ~when fallback bounds are used,~ suggesting 'Compute from data' for accurate bounds. Document the behaviour.

  

# Checklist


aaime:
Thanks for the contribution. Unfortunately, this approach isn't the right direction.

EPSG:404000 is a wildcard CRS precisely because the coordinate space is unknown — it's used for floor plans, engineering grids, and other non-geographic data just as much as for unidentified geographic data. Substituting WGS84 world bounds produces a plausible-looking wrong answer, which is worse than the null envelope it replaces.

The correct fix is either to disable or block 'Compute from SRS bounds' for wildcard/engineering CRSs, or provide a clear error message that SRS bounds are unknown,  and tell the user to use 'Compute from data' instead.

petersmythe:
Thank you for the guidance, I have now done just that:

image

And the X/Y extents remain blank.

petersmythe:

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
```diff
@@ -98,6 +98,20 @@ public BasicResourceConfig(String id, IModel<ResourceInfo> model) {
         final EnvelopePanel nativeBBox = new EnvelopePanel("nativeBoundingBox", nativeBBoxModel);
         nativeBBox.setOutputMarkupId(true);
         refForm.add(nativeBBox);
+
+        // Warning label shown when CRS has no defined bounds (e.g. EPSG:404000)
+        final Label noBoundsWarning = new Label(
```
aaime: This is user-visible, so it should be internationalized. Something like this:

```
 new StringResourceModel("noBoundsWarning", this)
      .setParameters(model.getObject().getSRS())
```

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.html
```diff
@@ -107,6 +107,7 @@
                     <wicket:message key="nativeBbox">Native BBox</wicket:message>
                   </label>
                   <div>
+                    <div wicket:id="noBoundsWarning" class="warning-link" style="display:none"></div>
```
aaime: I think the rest of the codebase uses span for warnings... can you check?

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
aaime: But at the same time... don't do it here. This captures the SRS at construction time, but users can modify it, one could add 404000 or another engineering CRS when editing. Make it consider the current value instead.

src/web/core/src/main/java/org/geoserver/web/data/resource/BasicResourceConfig.java
```diff
@@ -98,6 +98,20 @@ public BasicResourceConfig(String id, IModel<ResourceInfo> model) {
...
+                "noBoundsWarning",
+                "CRS "
+                        + model.getObject().getSRS()
+                        + " has no defined spatial extent."
+                        + " Use 'Compute from data' to set bounds, or enter them manually.");
+        noBoundsWarning.setOutputMarkupId(true);
+        noBoundsWarning.setOutputMarkupPlaceholderTag(true);
+        noBoundsWarning.setVisible(
```
aaime: I suggest not fixating on 404000, pretty much any CRS provided by WKT definition lacks a valid area, only CRS coming from the EPSG database have one.

aaime (changes requested):
Good, the direction is good now. There are a few issues I noted below, and the changes lack testing.

petersmythe:
Changes made as requested.

aaime:
Code and doc changes look good now, we have a leftover communication issue: the users reading the jira ticket and description will get a description that does not match what is actually done in the PR. I suggest amending it, and update the commit message to match.

PR #9545, opened 2026-06-01
[GEOS-12123] Login failure in FIPS mode for GeoServerDigestPasswordEncoder

mortenlud:
[GEOS-12123](https://osgeo-org.atlassian.net/browse/GEOS-12123) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

Fix GeoServerDigestPasswordEncoder to not throw NoSuchAlgorithmException: SHA1PRNG SecureRandom in fips enabled systems.

The changes I’ve made are limited to replacing the use of the Jasypt library in one class, while preserving the same hashing format and behavior. The goal has been specifically to avoid any backward compatibility issues.

I have added some tests to ensure backward compability with the old Jasypt library

# Checklist


src/main/src/main/java/org/geoserver/security/password/GeoServerDigestPasswordEncoder.java
```diff
@@ -23,39 +21,59 @@
...
     @Override
     protected PasswordEncoder createStringEncoder() {
-        PasswordEncoder encoder = new JasyptPasswordEncoderWrapper();
-        ((JasyptPasswordEncoderWrapper) encoder).setPasswordEncryptor(new StrongPasswordEncryptor());
-        ((JasyptPasswordEncoderWrapper) encoder).setPrefix(getPrefix());
-        return encoder;
+        return new PasswordEncoder() {
+            @Override
+            public String encode(CharSequence rawPassword) {
+                byte[] bytes = toBytes(rawPassword);
```
aaime: The old code using jasypt ensured the UTF-8 usage, the new one does not. This can result in a backwards incompatible behavior change if the passoword is written using non ASCII chars.
Suggestion: change the toBytes utility below to call:

tils.toBytes(chars, StandardCharsets.UTF

Also add tests for exotic UTF-8 chars

src/main/src/test/java/org/geoserver/security/password/GeoServerDigestPasswordEncoderTest.java
```diff
@@ -0,0 +1,90 @@
+/* (c) 2026 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.security.password;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
```
aaime: JUnit 5 is indeed available in GeoServer, but only because WicketTester requires it
The rest of the codebase is still using JUnit 4, please keep uniformity

aaime (changes requested):
Overall looks good, but there is one regression and a project testing style issue

src/main/src/main/java/org/geoserver/security/password/GeoServerDigestPasswordEncoder.java
mortenlud: I have now changed the toBytes(CharSecuence) method in this class to use UTF8

aaime:
I think it's still leaving some classes of UTF situations out. I cooked up a patch, care to have a look at it? 
If you like it, add it to this PR:
[normalize.diff.txt](https://github.com/user-attachments/files/29386433/normalize.diff.txt)

PR #9546, opened 2026-06-01
[GEOS-12127] Fix flaky VectorTileMetatilingTest cache assertion

petersmythe:
[GEOS-12127](https://osgeo-org.atlassian.net/browse/GEOS-12127) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

https://osgeo-org.atlassian.net/browse/GEOS-12127

`testMetatilingEnabled` fails intermittently because it asserts HIT on sibling tiles immediately after the first metatile request, but the cache write for sibling tiles completes asynchronously. On slow CI runners the write has not finished by the time the next request arrives, producing MISS instead of the expected HIT.

**Fix**: Use Awaitility to poll for the expected HIT on the first sibling tile before asserting the remaining tiles. This tolerates the async nature of metatile cache population (up to 5 seconds) without watering down the assertions. The non-metatiling path (`testMetatilingDisabled`) is unaffected — it expects all MISS and does not wait.

## Checklist

- [x] I have sent a Contribution Licence Agreement (not required for small changes, e.g., fixing typos in documentation).
- [x] First PR targets the main branch (backports managed later; ignore for branch specific issues).

- [ ] Documentation has been updated (if change is visible to end users).
- [ ] The REST API docs have been updated (when changing configuration objects or the REST controllers).
- [x] There is an issue in the GeoServer Jira (except for changes that do not affect administrators or end users in any way).

### DIFF
--- a/src/extension/vectortiles/src/test/java/org/geoserver/wms/vector/VectorTileMetatilingTest.java
+++ b/src/extension/vectortiles/src/test/java/org/geoserver/wms/vector/VectorTileMetatilingTest.java
@@ -4,6 +4,9 @@
  */
 package org.geoserver.wms.vector;
 
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.awaitility.Awaitility.await;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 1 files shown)


petersmythe:

src/extension/vectortiles/src/test/java/org/geoserver/wms/vector/VectorTileMetatilingTest.java
```diff
@@ -124,8 +127,28 @@ private void assertTiles(CacheResult[][] cacheResults) throws Exception {
...
+        MockHttpServletResponse firstResp = getTile(z, x, y);
+        assertEquals(200, firstResp.getStatus());
+        byte[] firstData = firstResp.getContentAsByteArray();
+        assertCacheResult(firstResp, cacheResults[0][0]);
+        assertTrue(firstData.length > 0);
+        assertFeature(decoder, firstData, 0, 0);
+
+        // When metatiling is enabled, sibling tiles are stored asynchronously after the first
+        // request completes. Wait for the cache to be populated before asserting HIT on them.
+        if (cacheResults[0][1] == CacheResult.HIT) {
```
aaime, line 140: Hum... it's going in the right direction but it's weak. Saving happens in a secondary thread pool, so we cannot assume all tiles in the metatile have been saved by checking just one of them.
I would await for all tiles in the meta instead (this might be doable in the testing loop below directly,  not 100% sure)... something like this? (warning, untested)

```java
  for (int j = 0; j < 2; j++) {
      for (int i = 0; i < 2; i++) {
          if (j == 0 && i == 0) continue;
          final int fi = i, fj = j;
          if (cacheResults[j][i] == CacheResult.HIT) {
              await().atMost(5, SECONDS).pollInterval(100, MILLISECONDS).untilAsserted(() -> {
                  MockHttpServletResponse resp = getTile(z, x + fi, y + fj);
                  byte[] data = resp.getContentAsByteArray();
                  assertCacheResult(resp, CacheResult.HIT);
                  assertTrue(data.length > 0);
                  assertFeature(decoder, data, fi, fj);
              });
          } else {
              MockHttpServletResponse resp = getTile(z, x + i, y + j);
              byte[] data = resp.getContentAsByteArray();
              assertCacheResult(resp, cacheResults[j][i]);
              assertTrue(data.length > 0);
              assertFeature(decoder, data, i, j);
          }
      }
  }
```

aaime (changes requested):

PR #9547, opened 2026-06-01
[GEOS-12127] Add diagnostics to flaky PipeliningTaskQueueTest

petersmythe:
[GEOS-12127](https://osgeo-org.atlassian.net/browse/GEOS-12127) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

https://osgeo-org.atlassian.net/browse/GEOS-12127

`PipeliningTaskQueueTest` fails intermittently (~once per 2 months on CI) but we have no logs from previous failures — GitHub Actions retains logs for only 90 days. Without knowing whether the failure is a `ConditionTimeoutException` (Awaitility 60s timeout) or an `AssertionError` (ordering violation), we cannot determine the root cause or apply a targeted fix.

**What this change does**: Instruments the test to capture diagnostic state on failure, without changing the test logic:

- **On timeout**: reports which tasks completed, which never started vs started but never finished, actual sleep durations, and elapsed time. This will distinguish scheduler thread starvation (likely given observed GitHub Actions runner starvation) from a deadlock/livelock in `PipeliningTaskQueue`.

- **On ordering violation**: reports the full completion order for the affected group with timestamps, enabling root-cause analysis of the pipeline's ordering guarantee.

The original test logic (Random sleep, 60s timeout, ordering assertion) is preserved unchanged — only the failure reporting is enhanced.

**Context**: Andrea Aime has already increased the timeout twice (20s → 60s in [49390d678d8](https://github.com/geoserver/geoserver/commit/49390d678d8)). A further increase to 600s was tested locally but never merged. The current hypothesis is runner starvation causing the `ScheduledExecutorService` (which polls every 10ms) to not fire for extended periods, but this needs confirmation from actual failure diagnostics.

## Checklist

- [x] I have sent a Contribution Licence Agreement (not required for small changes, e.g., fixing typos in documentation).
- [x] First PR targets the main branch (backports managed later; ignore for branch specific issues).

- [ ] Documentation has been updated (if change is visible to end users).
- [ ] The REST API docs have been updated (when changing configuration objects or the REST controllers).
- [x] There is an issue in the GeoServer Jira (except for changes that do not affect administrators or end users in any way).

### DIFF
--- a/src/extension/monitor/core/src/test/java/org/geoserver/monitor/PipeliningTaskQueueTest.java
+++ b/src/extension/monitor/core/src/test/java/org/geoserver/monitor/PipeliningTaskQueueTest.java
@@ -7,15 +7,17 @@ package org.geoserver.monitor;
 
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.awaitility.Awaitility.await;
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Queue;
 import java.util.Random;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 import org.geotools.util.logging.Logging;
 import org.junit.After;
 import org.junit.Before;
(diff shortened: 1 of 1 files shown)


petersmythe:

aaime:
I run the build with a script bombarding the build with lots of threads for a few minutes... here is my machine under load in technicolor:

image

Eventually it failed and produced enough diagnostic locally to go hunt for a fix... this should be it, at line 104 of pipeliningTaskQueue:

```diff
diff --git a/src/extension/monitor/core/src/main/java/org/geoserver/monitor/PipeliningTaskQueue.java b/src/extension/monitor/core/src/main/java/org/geoserver/monitor/PipeliningTaskQueue.java
index 6a661402833..d926c13c965 100644
--- a/src/extension/monitor/core/src/main/java/org/geoserver/monitor/PipeliningTaskQueue.java
+++ b/src/extension/monitor/core/src/main/java/org/geoserver/monitor/PipeliningTaskQueue.java
@@ -101,7 +101,9 @@ public void run() {
                         if (LOGGER.isLoggable(Level.FINEST)) {
                             LOGGER.finest("Removing task from queue " + job.key);
                         }
-                        pipeline.remove();
+                        // remove this specific job, not the head: a concurrent thread may have
+                        // already advanced the head, and a blind remove() would drop an unrun task
+                        pipeline.remove(job);
                     }
                 } else {
                     // start the job
```

The stress tester is here, assumes Linux and a 12-core machine (24 threads), but you can adjust it to fit your environment:
[bombard-pipelining.sh](https://github.com/user-attachments/files/28945144/bombard-pipelining.sh)

It would be useful if you could run it too, before and after the fix, to confirm.

PR #9552, opened 2026-06-01
Throw HttpErrorCodeException for unsupported tile format

petersmythe:
When a client requests a tile in a format not supported by the layer (e.g. `image/gif` for a layer that only caches `image/png`), the code previously threw `IllegalArgumentException`. This was caught by the generic exception handler in `GeoWebCacheDispatcher` which logged at SEVERE with a full stack trace — excessive noise for a routine client error.

**Fix**: Change the single `throw` in `GeoServerTileLayer.getTile()` from `IllegalArgumentException` to `HttpErrorCodeException(400, message)`. The dispatcher already handles `HttpErrorCodeException` cleanly: returns 400 with the error message to the client, no stack trace logged.

This is a targeted fix at the source — only the specific "unsupported format" validation is changed. All other `IllegalArgumentException` throws elsewhere remain untouched.

**Tested**: Docker deployment confirms:
- WMTS path: client gets `400 text/plain` with message `image/gif is not a supported format for sf:archsites`, no server log entry
- WMS `tiled=true` path: format mismatch detected before `getTile()` is called, graceful fallback to standard WMS renderer

# Checklist

### DIFF
--- a/src/gwc/src/main/java/org/geoserver/gwc/layer/GeoServerTileLayer.java
+++ b/src/gwc/src/main/java/org/geoserver/gwc/layer/GeoServerTileLayer.java
@@ -116,6 +116,7 @@ import org.geowebcache.locks.LockProvider.Lock;
 import org.geowebcache.mime.FormatModifier;
 import org.geowebcache.mime.MimeException;
 import org.geowebcache.mime.MimeType;
+import org.geowebcache.service.HttpErrorCodeException;
 import org.geowebcache.storage.StorageBroker;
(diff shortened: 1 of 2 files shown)


jratike80:
I think this is also in line with the WMTS standard. Fro Table 24 — HTTP exception codes and meanings on GetTile operation:

```
exceptionCode             value     Code Message
InvalidParameterValue      400     Bad request  
```

mprins (approved):

PR #9554, opened 2026-06-01
[GEOS-12130] Tolerate security filters/providers from uninstalled plugins instead of failing startup (legacy OIDC/Keycloak/OAuth2 data dirs)

afabiani:
[GEOS-12130](https://osgeo-org.atlassian.net/browse/GEOS-12130) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

  

A data directory created by a removed or uninstalled security plugin (the Spring
Security 5 OAuth2/OpenID Connect, Keycloak, Google, GitHub or GeoNode plugins replaced
by the unified OIDC connector) can no longer have its persisted authentication filter
and role service configurations loaded by GeoServer 3.x. Rather than aborting startup,
or silently dropping an authentication filter from a chain, GeoServer now degrades
gracefully:

 * the unloadable component is disabled and removed from every request filter chain
   that referenced it; the on-disk configuration is left untouched (non-destructive)
   until the administrator migrates it;
 * it fails closed: if removing the filter would leave a chain with no authenticator
   and no security interceptor, a 403 deny filter is injected, so a previously
   protected chain can never silently become open;
 * the problem is recorded (SecurityConfigDiagnostics) and reported to administrators
   on the GeoServer home page.

The mechanism is generic and lives in core (gs-main): a GeoServerSecurityProvider
contributes LegacySecurityAlias entries via getLegacyAliases(), which buildPersister()
maps to generic DisabledFilterConfig / DisabledRoleServiceConfig placeholders. The OIDC
community module contributes the legacy OAuth2/OIDC aliases as data
(OidcLegacyAliasProvider); a vanilla GeoServer without it still tolerates the data
directory through the generic fallback.

Adds core and oidc-core integration tests and an OIDC migration guide.

image

# Checklist


jodygarnett:
Thanks this is what we were we are waiting on to take OIDC to extension status.
After this is merged can we prep a PR to change to extension please?

doc/en/user/source/community/oidc/images/home-page-warning.png
jodygarnett, line 1: Two things:
- These screen snaps are very wide, our doc guide asks for 800 width
- Please crop to only show the warnings if you can take the time

doc/en/user/source/community/oidc/images/oidc-login-filter.png
jodygarnett, line 1: Similar feedback to above, crop to show only the form area, you may wish to break up by heading to make the documentation easier to follow (this also results in less changes as options change).

doc/en/user/source/community/oidc/migrating.rst
```diff
@@ -0,0 +1,302 @@
...
+   * *The following authentication filter chains were updated* clears as soon as no chain still references the
+     unknown filter — saving the *Authentication* page already rewrites the chains without it.
+   * *The following security components could not be loaded* clears only once the obsolete ``security/filter/<name>``
+     (and, for Keycloak, ``security/role/<name>``) directory is removed from the data directory.
+
+   A disabled component is **not** listed among the *Authentication* filters, so it cannot be removed with the
+   *Remove selected* button — delete its directory on disk, then restart GeoServer, which is the reliable way to
+   re-evaluate both parts.
+
+.. figure:: images/oidc-login-filter.png
```
jodygarnett: We now use markdown, what is this `rst` file doing?

jodygarnett (changes requested):
Please convert docs to markdown for publication.

src/main/src/main/java/org/geoserver/security/GeoServerSecurityProvider.java
```diff
@@ -108,6 +110,42 @@ public void configure(XStreamPersister xp) {
...
+        return Collections.emptyList();
+    }
+
+    /** Finds the {@link LegacySecurityAlias} contributed for the given persisted class name, across all providers. */
+    public static Optional<LegacySecurityAlias> findLegacyAlias(String className) {
+        if (className == null) {
+            return Optional.empty();
+        }
+        for (GeoServerSecurityProvider provider : GeoServerExtensions.extensions(GeoServerSecurityProvider.class)) {
+            for (LegacySecurityAlias alias : provider.getLegacyAliases()) {
```
jodygarnett, line 128: Is there some CONFIG level logging when migration is applied? That is what happens when migrating data directory; it also offers some solace for those operating headless who are not going to see UI notifications.

afabiani: I placed it in `SecurityConfigDiagnostics`
  (where the migration is recorded) rather than at `findLegacyAlias` (a pure lookup), so
  it fires once per applied migration — consistent with the existing `GeoServerSecurityManager`
  "security migration" CONFIG logging.

doc/en/user/community/oidc/index.md
```diff
@@ -13,6 +13,7 @@ The `oidc` module is a security module that accepts users (and roles) from exter
 - [Configuring with a Generic OIDC IDP](oauth2/generic.md)
 - [Configuring the roles source](role-config.md)
 - [Advanced Information](advanced.md)
+- [Migrating from the legacy OAuth2/OIDC plugins](migrating.md)
```
aaime, line 16: While I'm doing a review, I checked the docs end to end. In the installing.md file I see a couple of things that are odd and should be fixed, if the objective is to push the module to extension (and are wrong anyways for a pending module):

* "Community modules are not yet ready for distribution with GeoServer release." Not really, no need to build from sources if the module is extension, pending, or even just in the community release profile
* "Using with the GeoServer Docker Container" points at 2.28.x, and the instructions would be wrong for an extension module (and for a pending module too as well, since it's part of releases, but I don't know what the right way to start geoserver would be... anyways moot point if you're trying to go to extension).

aaime: In addition, the module has extensive documentation for Google/Github/Azure/Keycloak setups. If the module is going to extension status, the instructions really need to work. Have they been tested/updated?

doc/en/user/community/oidc/migrating.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+![](images/home-page-warning.png)
+
+*Administrator home page listing the disabled legacy security components and the affected filter chains.*
+
+!!! note
+    Tolerance is generic and lives in GeoServer core, so even a vanilla GeoServer without the OpenID Connect module will start and disable the unknown filters --- it just reports them with a generic message instead of naming the originating plugin.
+
+### Fail-closed behaviour
```
aaime: While fail-closed is a well-known term among security experts, I would not assume the generic GeoServer/OIDC user knows what "fail-closed" means. Maybe 

### Removing the filter never exposes protected resources

or 

### Fail-closed behavior (removing the filter never exposes protected resources)

aaime: Also, while this is the right behavior, there's a real risk that the admin will lock himself out of the GeoServer, if OAuth was the only authentication mechanism left in the web/REST chains.

I would either add a description of how to manually setup the plugins before migration, or just tell them to set up an XML user/group store and basic/form auth enabled in the filter chains before attempting a migration to GeoServer 3.

src/main/src/main/java/org/geoserver/security/GeoServerSecurityManager.java
```diff
@@ -635,10 +663,26 @@ synchronized void init(SecurityManagerConfig config) throws Exception {
         // add the custom/configured ones
```
aaime, line 672: While not in the diff, a few lines above this we have:

```java
        if (roleService == null) {
            try {
                roleService = loadRoleService("default");
            } catch (Exception e) {
                throw new RuntimeException("Fatal error occurred loading default role service", e);
            }
        }
```

The old keycloak plugin was providing a role service, which might now fall back to the default role service, attaching different roles to users. This can have different consequences, like being locked out from the server due to insufficient roles, or getting different roles, potentially with more power than before (not fail-closed). Should this be handled the same way as filters?

src/main/src/main/java/org/geoserver/security/GeoServerSecurityManager.java
```diff
@@ -662,13 +706,233 @@ synchronized void init(SecurityManagerConfig config) throws Exception {
...
+    private SecurityConfigDiagnostics.DisabledComponent classifyFilter(String name) {
+        SecurityNamedServiceConfig config;
+        try {
+            // load as the base config type: a filter referenced by a chain may be persisted with a generic
+            // SecurityNamedServiceConfig rather than a SecurityFilterConfig, and narrowing to the latter here would
+            // throw ClassCastException and spuriously disable a perfectly loadable filter
+            config = filterHelper.loadConfigAsBase(name, true);
+        } catch (Exception e) {
+            // log the full cause so an administrator can diagnose; keep the reported reason generic so a malformed
+            // configuration file cannot surface raw (possibly sensitive) field content on the home page
```
aaime, line 835: Ok, but then at line 863 the class name coming from the XML is actually added in the disabled component and eventually moved to the home page?
Maybe the name should be sanitized against HTML and log manipulation.

src/community/security/oidc/oidc-core/src/test/java/org/geoserver/security/oauth2/legacy/LegacyOAuth2MigrationTest.java
```diff
@@ -0,0 +1,113 @@
...
+                .filter(c -> name.equals(c.name()))
+                .findFirst()
+                .orElse(null);
+    }
+
+    @Test
+    public void testLegacyOAuth2FiltersRecognizedDisabledAndStripped() throws Exception {
+        GeoServerSecurityManager secMgr = getSecurityManager();
+
+        // a 2.28.x data directory produced by the now-removed community plugins
```
aaime, line 53: Nice to see all components loaded. It's however not covering the keycloak role service migration:

  - install a keycloak-role-service.xml fixture under security/role/keycloak-roles/config.xml, configure it as the active role service in SecurityManagerConfig, call secMgr.reload(), then assert:
  - getConfigDiagnostics().getDisabledComponents() contains an entry with ComponentType.ROLE_SERVICE and alias keycloakRoleService
  - active role service fell back to default, or if you go for fail-closed, with something that does not provide roles at all
  - a warning is visible that role assignments from Keycloak are lost

src/main/src/test/java/org/geoserver/security/LegacySecurityConfigTest.java
```diff
@@ -0,0 +1,131 @@
...
+        custom.setName("customNoInterceptor");
+        custom.setFilterNames("legacyFilter");
+        custom.setInterceptorName(null);
+        config.getFilterChain().getRequestChains().add(custom);
+
+        secMgr.getConfigDiagnostics().clear();
+        secMgr.stripDisabledFiltersFromChains(config, new TreeSet<>(List.of("legacyFilter")));
+
+        RequestFilterChain after = config.getFilterChain().getRequestChainByName("customNoInterceptor");
+        assertTrue(
```
aaime, line 108: Good for the injection. I would actually go one step further and try a HTTP request, making sure we are getting a 403 with the  DisabledSecurityFilter.DENY_MESSAGE

aaime (changes requested):
Nice work @afabiani 
See below for a few comments worth considering.

afabiani:
Thanks @aaime — all addressed and pushed:

- **Role service:** good catch. A legacy/unresolvable role service is now recorded as a disabled `ROLE_SERVICE` in the diagnostics (home page + logs) *before* falling back to `default`, with a message spelling out that its role assignments are lost. Kept the default fallback so the root account stays as the recovery path. Added a Keycloak role-service migration test.
- **Sanitization:** the persisted name/class is control-char- and length-sanitized at the diagnostics boundary (log injection). The home-page label already HTML-escapes — added a `<script>`-payload render test to lock that in.
- **Fail-closed test:** `LegacySecurityConfigTest` now drives a real request through a chain that lost its authenticator and asserts `403` + `DisabledSecurityFilter.DENY_MESSAGE`.
- **Docs:** reworded the fail-closed heading; added a "Before you migrate: keep a non-OAuth admin login" section; fixed `installing.md` (dropped build-from-source, switched to the extension download/`STABLE_EXTENSIONS`); and re-audited the provider guides — Keycloak re-tested end-to-end against 26.1, fixed a couple of stale fields in `advanced.md`. Google/GitHub/Azure are audited against the current connector; a live pass there needs real IdP accounts.

aaime:
I'd say good to go. Will you do a separate PR for the graduation?

afabiani:
> I'd say good to go. Will you do a separate PR for the graduation?

Yes, and thanks for your time!

PR #9578, opened 2026-06-05
[GEOS-12132] Promote the OIDC (OAuth2/OpenID Connect) module to an extension

afabiani:
[GEOS-12132](https://osgeo-org.atlassian.net/browse/GEOS-12132) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

Implements **GSIP-239** (accepted: 7 +1, 1 abstention) — promotes the OAuth2/OpenID Connect security module from community to extension.

- JIRA: https://osgeo-org.atlassian.net/browse/GEOS-12132
- GSIP: https://github.com/geoserver/geoserver/wiki/GSIP-239

## What this does
Moves `gs-sec-oidc` (`oidc-core`, `oidc-web`, `oidc-assembly`) from `src/community/security/oidc/` to `src/extension/sec-oidc/` (history preserved via `git mv`). The plugin artifact name (`geoserver-<ver>-sec-oidc-plugin.zip`) is unchanged.

### Build / wiring
- Reparent to `org.geoserver:extension`; groupId `org.geoserver.community(.security)` -> `org.geoserver.extension`; artifactIds unchanged.
- Wired `sec-oidc` into `src/extension/pom.xml` (individual + `allExtensions` + `release` profiles); removed the `oidc` profile + the `communityRelease` entry from `src/community/pom.xml`; deleted the now-empty `src/community/security` grouping pom; fixed `src/web/app/pom.xml`'s `oidc` profile dependency groupId.

### Dependencies
- Dropped the unused `spring-security-jwt:1.0.11.RELEASE` (vestigial from the legacy plugin -- zero `org.springframework.security.jwt` imports).
- `nimbus-jose-jwt` is now managed at **10.4** in `src/pom.xml` `<dependencyManagement>` (the version `spring-security-oauth2-jose:7.0.5` declares) and inherited in `oidc-core`. The previous `9.37.2` pin was a stale leftover from the spring-security 5.8 rewrite (GEOS-11272), never bumped when GeoServer moved to spring-security 7. Two malformed JWK test fixtures were refreshed for nimbus 10's stricter JSON parser (an unquoted-JSON `jkws.json`, and a `stubJwks()` that emitted a non-JSON `Map.toString()`).

### Fixes surfaced by running the OIDC tests in CI
Community modules build with `-DskipTests`, so `GeoServerOAuth2LoginIntegrationTest` never ran in CI before. As an extension it runs in the main build matrix, which surfaced genuine, pre-existing bugs (now fixed):

- **Scoped-registration-id matching (3 components).** The scoped-registration-id refactor makes the runtime registration id `openidconnect__oidc`, but several components matched it with an exact `equals(REG_ID_OIDC)` (true only for the bare `oidc`), silently taking the wrong branch. All now use the `isRegIdOfType(...)` helper:
    - `GeoServerJwsAlgorithmResolver` — dropped the configured `HS256` and fell back to `RS256`, so Spring built a JWKS (asymmetric) decoder and nimbus 10 rejected the HS256-signed id_token with *"no matching key(s) found"*. (Looked like a nimbus-10 HMAC regression; it isn't — the symmetric decoder works fine on 10.4.) Added `GeoServerJwsAlgorithmResolverHmacTest`.
    - `GeoServerAuthorizationRequestCustomizer` — **user-visible**: wrongly forced PKCE on and suppressed `response_mode` for the scoped id. Added scoped-id test cases (re-validated against the live Keycloak e2e suite).
    - `GeoServerOAuth2AccessTokenResponseClient` — mislabeled the provider in debug logs.
- **Integration-test fixes:** assert the auth-server redirect against the IdP base (not the GeoServer base); refresh the id_token fixture's hardcoded **2020** `exp`/`iat` (expired since 2020) to dynamic values; update stale unscoped `redirect_uri` in the token stubs to the scoped form.
- Removed a duplicate `gs-sec-oidc-core` declaration in `oidc-web/pom.xml`.

### QA
Extension `-Dqa` gate passes -- Checkstyle / PMD / SpotBugs / Errorprone / `-Werror` all clean. Tests green: `oidc-core` 331, `oidc-web` 73 (incl. the 27 Keycloak Testcontainers ITs).

### Docs
Moved `doc/en/user/community/oidc/` -> `doc/en/user/extensions/oidc/`; relocated the mkdocs nav block from *Community modules* to *Extensions*; fixed the `upgrade3.md` cross-references.


aaime (changes requested):
Hi @afabiani overall looks pretty good, thanks!
I found a few leftovers. 

1) The documentation still mentions "community" in anchors, while that won't change the output, it would be nice to clean it up:

```
> cd doc/en/user/extensions/oidc; git grep community
advanced.md:## Troubleshooting {: #community_oidc_troubleshooting }
advanced.md:- Not setting `PROXY_BASE_URL` correctly in Docker or container deployments, causing the Redirect URI to resolve to an internal hostname that the user's browser cannot reach. See [Redirect Base URI](configuring.md#community_oidc_redirect_base_uri).
advanced.md:- After logout, being unexpectedly signed out of all applications that share the same IDP --- this is the expected default behavior. See [Logout Behavior](configuring.md#community_oidc_logout_behavior).
configuring.md:## Common Login Settings {: #community_oidc_common_settings }
configuring.md:| Redirect Base URI | The public base URL of your GeoServer instance, ending with the context path (e.g. `https://geoserver.example.com/geoserver`). | Automatically resolved --- see [Redirect Base URI](#community_oidc_redirect_base_uri) below. |
configuring.md:| After-Logout Redirect URI | The URI the user is sent to after the IDP completes a logout. | Must be registered with the IDP as a permitted redirect URI. See [Logout Behavior](#community_oidc_logout_behavior). |
configuring.md:### Redirect Base URI {: #community_oidc_redirect_base_uri }
configuring.md:### Multiple OIDC filters {: #community_oidc_multiple_filters }
configuring.md:### Logout Behavior {: #community_oidc_logout_behavior }
installing.md:    Setting `PROXY_BASE_URL` ensures that the OIDC [Redirect Base URI](configuring.md#community_oidc_redirect_base_uri) is correctly resolved to the external URL that users access. Without it, GeoServer may use an internal Docker hostname for the redirect URI, causing authentication to fail.
migrating.md:by a community plugin that is not installed in this GeoServer; reconfigure or remove them and restart GeoServer:
oauth2/azure.md:        The exact redirect URI that GeoServer will use is shown as the read-only **Redirect URI** field in the filter configuration form --- copy it verbatim. In production, use that value instead of `localhost`. The filter-name prefix lets several OIDC filters share an IDP without colliding on their redirect URIs. See [Redirect Base URI](../configuring.md#community_oidc_redirect_base_uri).
oauth2/azure.md:See [troubleshooting](../advanced.md#community_oidc_troubleshooting).
oauth2/generic.md:- **Post-Logout Redirect URI:** If your IDP supports single logout, also register the **After-Logout Redirect URI** shown in the GeoServer configuration. See [Logout Behavior](../configuring.md#community_oidc_logout_behavior).
oauth2/generic.md:### Manual Configuration (without Discovery) {: #community_oidc_manual_config }
oauth2/generic.md::   The most common configuration error is a mismatch between the Redirect URI that GeoServer sends and the one registered with your IDP. Copy the exact value from the read-only **Redirect URI** field in the GeoServer filter form into your IDP's client configuration. See [Redirect Base URI](../configuring.md#community_oidc_redirect_base_uri) for how this value is calculated.
oauth2/generic.md:See [troubleshooting](../advanced.md#community_oidc_troubleshooting).
oauth2/github.md:        The exact callback URL that GeoServer will use is shown as the read-only **Redirect URI** field in the filter configuration form --- copy it verbatim. In production, use that value instead of `localhost`. The filter-name prefix lets several OIDC filters share an IDP without colliding on their redirect URIs. See [Redirect Base URI](../configuring.md#community_oidc_redirect_base_uri).
oauth2/github.md:See [troubleshooting](../advanced.md#community_oidc_troubleshooting).
oauth2/google.md:        The exact redirect URI that GeoServer will use is shown as the read-only **Redirect URI** field in the filter configuration form --- copy it verbatim. In production, use that value instead of `localhost`. The filter-name prefix lets several OIDC filters share an IDP without colliding on their redirect URIs. See [Redirect Base URI](../configuring.md#community_oidc_redirect_base_uri).
oauth2/google.md:See [troubleshooting](../advanced.md#community_oidc_troubleshooting).
oauth2/keycloak.md:        The exact redirect URI that GeoServer will use is shown as the read-only **Redirect URI** field in the filter configuration form (e.g. `http://localhost:8080/geoserver/web/login/oauth2/code/<filterName>__oidc`, where `<filterName>` is the name of the GeoServer OIDC filter). The filter-name prefix lets several OIDC filters share an IDP without colliding on their redirect URIs. In production, you may want to register only the specific URIs you actually use rather than a wildcard; i→
oauth2/keycloak.md:### Configure Role Source — Keycloak Admin API (server-side fetch) {: #community_oidc_keycloak_admin_api }
oauth2/keycloak.md:See [troubleshooting](../advanced.md#community_oidc_troubleshooting).
role-config.md:| Keycloak Admin API | Server-side fetch of the logged-in user's role assignments via Keycloak's [Admin REST API](https://www.keycloak.org/docs-api/latest/rest-api/index.html). Use this when you need to surface roles assigned to the user **through groups** (not just direct realm-role assignments) or when composite-role expansion needs to happen server-side. Only valid against a Keycloak IDP. See [Configuring with Keycloak](oauth2/keycloak.md#community_oidc_keycloak_admin_api) for the prerequ→
role-config.md:## Configuring Roles from the Keycloak Admin REST API {: #community_oidc_keycloak_admin_api_role_source }
role-config.md:See [Configuring Keycloak Admin API as the role source](oauth2/keycloak.md#community_oidc_keycloak_admin_api) for the step-by-step Keycloak setup.
role-config.md:## User Profile Properties {: #community_oidc_user_profile_properties }
```

2) The migration-readme.md in the sec-oidc folder looks outdated... but I may be wrong. If you agree on its status, I would remove it.

3) The module landed in extension/sec-oidc, but the security modules have a practice of landing in a "security" parent also in core and extension. If you check you'll find "extension/security/cas" is already using this patter, would be nice if you could do the same for oidc, fixing the parent poms and the module references accordingly.

4) While not feedback for this PR specificatlly, you'll also have to update the [3.0 release template in the site repo](https://github.com/geoserver/geoserver.github.io/blob/main/_layouts/release_30.html) so that OIDC is not an experimental module any longer, but a supported one instead.

afabiani:
@aaime thanks for the review — all four addressed:

- Doc anchors — renamed the community_oidc* anchors and their cross-references to oidc* across the 9 pages (left the legitimate "community plugin that is not installed" line in migrating.md).
- migration-readme.md — removed (stale rewrite note).
- Module layout — moved under extension/security/ following the cas pattern: oidc-core / oidc-web / oidc-assembly now parent gs-security, activated via a -Poidc profile (consistent with the existing web/app oidc profile) and added to allExtensions/release; dropped the standalone sec-oidc profile + module refs from extension/pom.xml. Also removed the gs-web-sec-core compile dependency that gs-security was force-injecting into every child (so oidc-core stays web-free), declaring it explicitly in web-cas where it's actually used.
- release_30.html — that's the site repo; I'll move OIDC from Community to Extensions there.

Separately, while validating the move I built the plugin and deployed it into a clean GeoServer `WEB-INF/lib` end-to-end. The assembly whitelist had gone stale: missing several required runtime jars (`oauth2-oidc-sdk`, `content-type`, `lang-tag`, `jcip-annotations`, `asm`) and bundling ~10 it shouldn't (`spring-security-core/web/config`, `jackson`, … already provided by core). 

Rewrote `assembly.xml` to the exact OIDC-specific set — the `-Poidc` dependency delta over a vanilla build (15 jars), cas-style. 

Verified with `jdeps` (no unresolved references from the OIDC classes) and by booting GeoServer with the plugin (the OAuth2 login provider + button manager wire cleanly).

petersmythe:
Congrats, can this be merged now?

aaime:
Done

PR #9582, opened 2026-06-08
[GEOS-12125] Add missing excel-wfs profile to web app WAR build

petersmythe:
[GEOS-12125](https://osgeo-org.atlassian.net/browse/GEOS-12125) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

The [GEOS-11949](https://osgeo-org.atlassian.net/browse/GEOS-11949) refactor (commit 0fa23c0f318) split the monolithic `excel` extension into three sibling modules (`excel-core`, `excel-wfs`, `excel-wps`) but only added a `web/app` profile for the new WPS module (`xlsxppio`). The old `excel` profile that bundled the WFS output format jar into the WAR was removed without a replacement, so `gs-excel-wfs` is never included and the Excel/Excel2007 WFS download options are unavailable in Layer Preview.

This commit adds a coherent set of profiles across both POMs:

**`src/extension/pom.xml`** (controls which modules are compiled):
| Profile | Modules built |
|---------|---------------|
| `excel` | excel-core, excel-wfs, excel-wps |
| `excel-wfs` | excel-core, excel-wfs |
| `excel-wps` | excel-core, excel-wps |

**`src/web/app/pom.xml`** (controls which jars are included in the WAR):
| Profile | Artifacts included |
|---------|-------------------|
| `excel` | gs-excel-wfs, gs-excel-wps |
| `excel-wfs` | gs-excel-wfs |
| `xlsxppio` | gs-excel-wps (unchanged) |

A developer can now say `-Pexcel` and get both compilation and WAR inclusion of all Excel modules, or use the granular profiles for selective inclusion.

Reported at: https://discourse.osgeo.org/t/excel-wfs-extension-profile-missing-2-27-3/151120
Patch credit: Fincer (https://fjordtek.com/git/Fincer/geoserver_excel_patch)
Jira: https://osgeo-org.atlassian.net/browse/GEOS-12125

# Checklist

- [x] The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers). N/A - no REST changes


src/extension/pom.xml
```diff
@@ -154,6 +154,8 @@
       <id>excel</id>
       <modules>
         <module>excel-core</module>
+        <module>excel-wfs</module>
```
aaime, line 157: A bug for sure, but would affect only someone that uses -Pexcel. Uncommon. Not wrong mind.

src/web/app/pom.xml
```diff
@@ -1027,6 +1027,16 @@
         </dependency>
       </dependencies>
     </profile>
+    <profile>
+      <id>excel-wfs</id>
```
aaime, line 1046: Oh my what a mess. Situation after this PR:
* If one just says -Pexcel, the modules will be compiled, but not included in the war
* If one just says -Pexcel-wfs, the modules will not be compiled, and the war will include whatever wfs-excel jar is found in the local repository
* If one just says -Pxmlsppio, this brings in transitively wps too, and the jars will still be picked from the local repository rather than being included in the build

Confusion much?

Maybe we need 3 separate profiles for both extension/pom.xml and web/pom.xml:
* excel: brings in everything. Useful for the developer that works on it and wants to do interactive testing from Start.java
* excel-wfs, wfs only
* xmlsppio, wps only

aaime (changes requested):
This "improvement" is quite confusing, see comments below.

src/extension/pom.xml
```diff
@@ -154,6 +154,24 @@
...
+    <profile>
+      <id>excel-wfs</id>
+      <modules>
+        <module>excel-core</module>
+        <module>excel-wfs</module>
+      </modules>
+    </profile>
+
+    <profile>
+      <id>excel-wps</id>
```
aaime, line 171: Should have been xmlsppio, or this will load the modules in the build, but don't include them in the war. As an alternative, rename xmlsppio in web/app instead (probably better, excel-wps is easier to understand). But they need to be consistent across the two poms.

PR #9583, opened 2026-06-08
Improve CI workflow consistency and add surefire report artifacts

petersmythe:
## Summary

Improves consistency across all GitHub Actions workflows and adds surefire test report upload on failure for easier debugging.

### Surefire report artifacts on failure (7 workflows)
- Added `upload-artifact` step with `if: failure()` to upload surefire reports (`.xml` and `.txt`) when tests fail
- Applies to: linux, windows, macos, linux_locales, linux_timezone, postgis_appschema_online, assembly
- All use `retention-days: 1` to minimize storage cost

### Retention-days consistency (2 workflows)
- Added `retention-days: 1` to existing artifact uploads in acceptance and cite workflows (previously defaulted to 90 days)

### MAVEN_OPTS deduplication (5 workflows: linux, windows, macos, linux_locales, linux_timezone)
- Removed duplicate `-Daether.syncContext.named.time=120` and `-Daether.syncContext.named.factory=` entries
- Fixed double-space before `-Daether.syncContext.named.factory=noop`
- Standardized property order

### fetch-depth: 500 removed (3 workflows: linux, linux_locales, linux_timezone)
- None of these workflows use git history, so shallow clone (default depth 1) is sufficient and faster

### macOS SNAPSHOT cleanup path fixed (macos.yml)
- Changed `find .m2/repository` to `find ~/.m2/repository` to match the actual cache location

### Checkout/Java ordering fixed (postgis_appschema_online.yml)
- Moved `actions/checkout` before `actions/setup-java` to match all other workflows

### paths-ignore standardized (3 workflows: qa, assembly, postgis_appschema_online)
- Added the missing `GeoServerApplication_*.properties` exclusion (with `!*_fr.properties` re-inclusion) to match other workflows

## Testing
- No functional changes to build or test logic
- Changes are limited to workflow orchestration and artifact handling

# Checklist

- [ ] The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers). N/A


.github/workflows/linux.yml
```diff
@@ -15,7 +15,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  MAVEN_OPTS: -Xmx2g -Daether.connector.basic.threads=8 -Daether.metadataResolver.threads=8 -Daether.syncContext.named.time=120 -Daether.syncContext.named.factory=file-lock -Daether.syncContext.named.nameMapper=file-gav -Dmaven.wagon.httpconnectionManager.ttlSeconds=25 -Dmaven.wagon.http.retryHandler.count=3 -Dorg.slf4j.simpleLogger.showDateTime=true -Dorg.slf4j.simpleLogger.dateTimeFormat=HH:mm:ss,SSS -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn -Dspotless.apply.skip=true -Daether.syncContext.named.time=120 -Daether.syncContext.named.time.unit=SECONDS -Daether.syncContext.named.factory=noop
```
aaime, line 18: So I get the deduplication of the repeated strings, but here we also have a change in locking?
Also note, assembly.yml and qa.yaml did not have the duplication, but they are still using file-lock.

.github/workflows/acceptance.yml
```diff
@@ -96,6 +96,7 @@ jobs:
         with:
           name: failed-test-files-java-${{ matrix.java-version }}
           path: ${{ env.GEOSERVER_ACCEPTANCE_FAILED_TESTS_DIR }}
+          retention-days: 1
```
aaime, line 99: Just... no? You're assuming someone with a day time job to look at the PR failures here.
The day to day reality I'm living is that the the PRs migth be looked at after days, depending on what other emergencies pop up. 90 days might be too much, but 1 is definitely too short. 
I'd go for at least one week. Btw, what problem is this change trying to address? Are we running out of some quota on github?

.github/workflows/postgis_appschema_online.yml
```diff
@@ -7,6 +7,8 @@ on:
       - 'mkdocs.yml'
       - 'requirements.txt'
       - '.github/workflows/docs.yml'
+      - '**/src/main/resources/GeoServerApplication_*.properties'
+      - '!**/src/main/resources/GeoServerApplication_fr.properties'
```
aaime, line 11: Why the "!" at the beginning? Also, are you sure it's supported? (sounds like a negation)

.github/workflows/macos.yml
```diff
@@ -43,7 +43,16 @@ jobs:
...
+        name: surefire-reports-macos
+        path: |
+          **/target/surefire-reports/*.xml
+          **/target/surefire-reports/*.txt
+        retention-days: 1
     - name: Remove SNAPSHOT jars from repository
       if: always()
       run: |
-        find .m2/repository -name "*SNAPSHOT*" -type d | xargs rm -rf {}
+        find ~/.m2/repository -name "*SNAPSHOT*" -type d | xargs rm -rf {}
```
aaime, line 58: Not a bug in this PR, but xargs has an extra {} here (that's used only in combination with -I).

.github/workflows/macos.yml
```diff
@@ -43,7 +43,16 @@ jobs:
     - name: Build with Maven
       run: |
         mvn -B -ntp -U -T1C -Prelease,pending,macos-github-build -f src/pom.xml clean install
+    - name: Upload surefire reports on failure
```
aaime, line 46: Hum... I normally just check build failures in the "logs as text" and get the information needed. Not sure we need the .xml/.txt files. I believe the advantage is not getting interleaved output like in the parallel build logs, althought normally I can read it anyways, and move to try to reproduce locally.
Long story short, is this upload worth it? (and all the other similar ones?).

aaime (commented):
The string deduplication is good, but leaves inconsistencies.
I normally don't have a need for surefire xml/txt reports, I would ask other java devs if they normally use them/look for them. See below for more notes.

PR #9584, opened 2026-06-08
[GEOS-12134] Fix N+1 query in JDBC role hierarchy resolution at login

petersmythe:
[GEOS-12134](https://osgeo-org.atlassian.net/browse/GEOS-12134) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

When a user with many roles logs in via the JDBC Role Service, `RoleCalculator.addInheritedRoles()` calls `JDBCRoleService.getParentRole()` individually for each role. Each call opens a separate database connection and executes individual queries, resulting in O(N) round-trips. For users with 2000+ roles over SSL connections to remote databases, login degrades to 100-200+ seconds.

This fix introduces a `BulkLoadableRoleService` interface. When the role service implements it, `addInheritedRoles()` bulk-loads all role-to-parent mappings and role properties in 2 queries (using existing `roles.all` and `roleprops.all` SQL), then resolves the hierarchy in-memory. Non-JDBC services continue using the existing per-role path unchanged.

**Changes:**
- New `BulkLoadableRoleService` interface in `org.geoserver.security` (gs-main)
- `JDBCRoleService` implements `BulkLoadableRoleService` with `getAllRoleProperties()`
- `RoleCalculator.addInheritedRoles()` uses bulk path when service is `BulkLoadableRoleService`
- Legacy path preserved for in-memory services (AbstractRoleService)

**Tested:** Verified with 1,247 roles resolving in ~500ms (2 bulk queries) vs. previous 1,247+ individual DB connections.

# Checklist

- [x] There is an issue in the [GeoServer Jira](https://osgeo-org.atlassian.net/browse/GEOS-12134) (except for changes that do not affect administrators or end users in any way).


petersmythe:
Last call for GeoServer 3.0

aaime:
Too central to security to change last minute. No proficient maintainer for this class can be found, the last substantial changes to it date 12 years ago:

```
git log --pretty=format:'%ad %an %s' --date=short -- src/main/src/main/java/org/geoserver/security/impl/RoleCalculator.java
2024-12-23 Andrea Aime [GEOS-11659] Apply Palantir Java format on GeoServer
2020-12-19 Andrea Aime [GEOS-9833] Remove explicit type when diamond operator can be used - add QA rule to enforce it
2020-02-23 Andrea Aime Cleaning up empty parameters, returns, throws.
2018-06-09 Andrea Aime The big bad reformat
2016-02-26 Jody Garnett first round of java doc errors identified
2014-09-04 Jody Garnett c) 2014 Open Source Geospatial Foundation (GSIP-118)
2013-01-22 jdeolive updated openplans copyright headers, GSIP-38
2012-11-14 Christian Mueller Fix for GEOS-5402 and GEOS-5388
2012-06-19 mcr Security role rework
2012-04-04 jdeolive GEOS-4554, GSIP 71 - new security subsystem
```

We need to revisit after 3.0 when there is some time to evaluate the consequences in all security subsystems (role calculator is used everywhere, not just in JDBC).

aaime:
I investigated this one; it works well in one use case, but regresses in other use cases:

  - Bulk loading grabs all roles in the system (not related to the user) in memory, works fine if there is relatively few roles, but will start causing memory pressure and OOMs under a combination of a large role
  system and many authenticated requests happening at the same time (e.g., 100000 roles, 200+ requests, makes for the materialization of a 20 million strings)
  - It assumes there are more roles than groups; in case there is actually relatively few roles but several groups, it causes more queries, not less

  Ideally, we'd want to load all the roles related to a given user with a single query, which is feasible in many databases using a recursive CTE query, e.g. something like this:

```sql
  WITH RECURSIVE ancestors(name, parent) AS (
      SELECT name, parent FROM roles WHERE name IN (?, ?, ?)  -- role list here, list of placeholders is dynamic
    UNION
      SELECT r.name, r.parent
      FROM roles r JOIN ancestors a ON r.name = a.parent
  )
  SELECT a.name, a.parent, p.propname, p.propvalue
  FROM ancestors a
  LEFT JOIN role_props p ON p.rolename = a.name;
```

Which would be great, but has a significant downside... the CTE syntax is not stable across databases, and Oracle has a limit of 1000 placeholders in a query (in practical terms it should work, but you'd have to code a fallback anyway). To go there, and handle the database syntax differences, we'd need more robust dialect machinery in the JDBCRoleService implementation.

An alternative that is likely going to work, with a few more queries, is loading one "wave" of parent roles at a time, based on all the currently known roles, without any database-specific queries. Loop until it returns nothing new (to avoid infinite recursion in case of loops); at each call past the first, feed in the newly discovered roles:

```sql
  select distinct parent from roles where name in (?, ...) and parent is not null
```

  Then at the end, load the role properties for all discovered roles:

```sql
  select rolename, propname, propvalue from role_props where rolename in (?, ?, ...)
```

This is still a loop mind, but it's proportional to the depth of the hierarchy, rather than all roles or all groups. Normally, a hierarchy is a pyramid, not a list, and the depth is normally small (it normally matches some organizational structure). Worth noting: even in the degenerate deep-chain case, this is no worse than the current per-role behavior, which already does one query per level.

Both these approaches would require a bulk-loading interface, but are faster and less OOM-prone on systems with many roles/much traffic, while still providing a significant benefit over the current code.

PR #3108, opened 2018-09-19
[GEOS-8929] Alias old com.vividsolutions.jts.geom.* to new org.locationtech.jts.geom.* in XStream reader

tbarsballe:
For https://osgeo-org.atlassian.net/browse/GEOS-8929

This change only affects reading; writing output is unchanged.
It is possible to alter this so that writing is likewise aliased in order to address [GEOS-8886](https://osgeo-org.atlassian.net/browse/GEOS-8886); see discussion in that ticket.

PR #9589, opened 2026-06-09
Skip KeycloakAdminApiIntegrationTest when the Keycloak container fails to start

afabiani:
Fixes the `gs-sec-oidc-web` failure breaking `main` on the OSGeo build server (e.g. [geoserver-main #3108](https://build.geoserver.org/job/geoserver-main/3108/)).

**Root cause:** the build agent's CPU lacks `x86-64-v2`, so the Keycloak 26.x container images cannot start there:
```
Fatal glibc error: CPU does not support x86-64-v2
... Container startup failed for image quay.io/keycloak/keycloak:26.0
```
The module's tests only began running in CI now that OIDC is a supported extension (community modules build with `-DskipTests`), so this path was never exercised before.

`KeyCloakIntegrationTestSupport` (used by the other three Keycloak tests) already catches a container-launch failure and `Assume`-skips — those report `Tests run: 0`. `KeycloakAdminApiIntegrationTest` called `keycloak.start()` unguarded in `@BeforeClass`, so it **errored** instead of skipping → `BUILD FAILURE`.

**Fix:** wrap the container construction/start in a `try/catch` and `Assume`-skip on launch failure, matching the sibling tests. Realm provisioning stays outside the `try`, so genuine provisioning regressions are still reported. On an agent where the container starts (e.g. GitHub Actions PR CI, which has x86-64-v2) the test runs exactly as before.

### DIFF
--- a/src/extension/security/oidc-web/src/test/java/org/geoserver/web/security/oauth2/intgration/keycloak/KeycloakAdminApiIntegrationTest.java
+++ b/src/extension/security/oidc-web/src/test/java/org/geoserver/web/security/oauth2/intgration/keycloak/KeycloakAdminApiIntegrationTest.java
@@ -69,8 +69,16 @@ public class KeycloakAdminApiIntegrationTest {
         Assume.assumeTrue("Docker not available — skipping", dockerAvailable);
 
         // Start a vanilla Keycloak — no realm-import, no fixtures. dasniko's container sets admin/admin via env vars.
-        keycloak = new KeycloakContainer("quay.io/keycloak/keycloak:26.0").withCustomCommand("--log-level=INFO");
-        keycloak.start();
+        try {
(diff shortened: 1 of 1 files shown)


aaime:
QA fails on app-schema but the module was not touched by this PR.
@jodygarnett I guess the issue was with the other PR:

```
Error: 5,865 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-pmd-plugin:3.24.0:check (default) on project gs-app-schema-test: PMD 7.23.0 has found 1 violation. For more details see: /home/runner/work/geoserver/geoserver/src/extension/app-schema/app-schema-test/target/pmd.xml -> [Help 1]
```

aaime:
Merging to get things moving. As far as I can see there are a few more fixes to be made, but they are not related to OIDC.

PR #9593, opened 2026-06-09
XMLUtils italian locale followup

aaime:
There are some test failures in italian locale, I've tried to adapt the test with these principles:
* Check the output is a service exception
* Parse the message out of it
* Check it contains what looks like the stable portion of the error message
 
# Checklist

### DIFF
--- a/src/wfs1_x/src/test/java/org/geoserver/wfs/ExternalEntitiesTest.java
+++ b/src/wfs1_x/src/test/java/org/geoserver/wfs/ExternalEntitiesTest.java
@@ -71,50 +71,58 @@ public class ExternalEntitiesTest extends WFSTestSupport {
 
     @Test
     public void testWfs1_0() throws Exception {
-        String output;
-        // enable entity parsing
+        Document dom;
         System.setProperty(EntityResolverProvider.ENTITY_RESOLUTION_UNRESTRICTED, "true");
         try {
-            output = string(post("wfs", WFS_1_0_0_REQUEST));
-            // the server tried to read a file on local file system
+            // enable entity parsing: server tries to read file on local file system
+            dom = postAsDOM("wfs", WFS_1_0_0_REQUEST);
             assertTrue(
                     "SAXException",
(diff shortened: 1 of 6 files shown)


jodygarnett:
The macOS test failed on gs-metadata, on a UI test `testMultiTabField` where wicket was looking at page output. Re-running in case the test is not stable or time dependent.

aaime:
Yes, that module is well known to have unstable tests on github (but they work 99.9% of the time locally)

src/wfs1_x/src/test/java/org/geoserver/wfs/ExternalEntitiesTest.java
```diff
@@ -71,50 +71,58 @@ public class ExternalEntitiesTest extends WFSTestSupport {
...
     public void testWfs1_0() throws Exception {
-        String output;
-        // enable entity parsing
+        Document dom;
         System.setProperty(EntityResolverProvider.ENTITY_RESOLUTION_UNRESTRICTED, "true");
         try {
-            output = string(post("wfs", WFS_1_0_0_REQUEST));
-            // the server tried to read a file on local file system
+            // enable entity parsing: server tries to read file on local file system
+            dom = postAsDOM("wfs", WFS_1_0_0_REQUEST);
```
jodygarnett, line 78: There is also a flat, postAsDOM("wfs", WFS_1_0_0_REQUEST,true) - where `true` is provided to skipDTD, providing an EmptyEntityResolver when parsing output back into the document (avoid geoserver trying access itself for the DTD) content.

Splitting the post and the parsing of output may easier to follow then the skipDTD flag.

jodygarnett (approved):
Thanks, so many ways to check exceptions.

src/wfs1_x/src/test/java/org/geoserver/wfs/ExternalEntitiesTest.java
aaime: Update as you see fit, the initial version is really just a quick fix.

PR #9598, opened 2026-06-10
[GEOS-12136] Use emptyDTD when testing html content

jodygarnett:
[GEOS-12136](https://osgeo-org.atlassian.net/browse/GEOS-12136) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

This avoid overloading w3c servers requesting http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd

  

# Checklist

### DIFF
--- a/src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java
+++ b/src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java
@@ -1248,7 +1248,7 @@ public class GeoServerSystemTestSupport extends GeoServerBaseTestSupport<SystemT
      * @return A result of the request parsed into a dom.
      */
     protected Document getAsDOM(final String path) throws Exception {
-        return getAsDOM(path, !path.endsWith("html"));
+        return getAsDOM(path, true);
     }
 
     /**
@@ -1259,20 +1259,20 @@ public class GeoServerSystemTestSupport extends GeoServerBaseTestSupport<SystemT
      * @return A result of the request parsed into a dom.
      */
     protected Document getAsDOM(final String path, int statusCode) throws Exception {
-        return getAsDOM(path, !path.endsWith("html"), statusCode);
+        return getAsDOM(path, true, statusCode);
     }
 
     /**
      * Executes an ows request using the GET method and returns the result as an xml document.
      *
      * @param path The portion of the request after the context, example: 'wms?request=GetMap&version=1.1.1&..."
-     * @param skipDTD if true, will avoid loading and validating against the response document schema or DTD
+     * @param useEmptyDTD if true, will avoid loading and validating against the response document schema or DTD
      * @param statusCode Expected status code
      * @return A result of the request parsed into a dom.
      */
-    protected Document getAsDOM(final String path, boolean skipDTD, int statusCode) throws Exception {
+    protected Document getAsDOM(final String path, boolean useEmptyDTD, int statusCode) throws Exception {
         try (InputStream responseContent = get(path, statusCode)) {
-            return dom(responseContent, skipDTD);
+            return dom(responseContent, useEmptyDTD);
         }
     }
 
@@ -1285,7 +1285,7 @@ public class GeoServerSystemTestSupport extends GeoServerBaseTestSupport<SystemT
      * @return A result of the request parsed into a dom.
      */
     protected Document getAsDOM(final String path, String encoding) throws Exception {
-        return getAsDOM(path, !path.endsWith("html"), encoding);
+        return getAsDOM(path, true, encoding);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


jodygarnett:
During integration of XMLUtils I tried using parser restrictions when testing GeoServer output - including html output. This resulted in overloading w3c servers with DTD requests.

Reverting the change with this PR.

src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java
```diff
@@ -1259,20 +1259,20 @@ protected Document getAsDOM(final String path) throws Exception {
...
      * Executes an ows request using the GET method and returns the result as an xml document.
      *
      * @param path The portion of the request after the context, example: 'wms?request=GetMap&version=1.1.1&..."
-     * @param skipDTD if true, will avoid loading and validating against the response document schema or DTD
+     * @param useEmptyDTD if true, will avoid loading and validating against the response document schema or DTD
      * @param statusCode Expected status code
      * @return A result of the request parsed into a dom.
      */
-    protected Document getAsDOM(final String path, boolean skipDTD, int statusCode) throws Exception {
+    protected Document getAsDOM(final String path, boolean useEmptyDTD, int statusCode) throws Exception {
```
aaime, line 1273: skipDTD was meaningful to me, I understood what it mean
useEmptyDTD as "avoiding loading and validating" does not track to me, I don't get why they would be related....

src/restconfig/src/test/java/org/geoserver/rest/catalog/AboutControllerTest.java
```diff
@@ -96,10 +96,10 @@ public void testGetStatusAsHTML() throws Exception {
         System.setProperty("badString", "\u0007\u0008\u001b[46m");
 
         // make the request, parsing the result into a Dom object
-        Document dom = getAsDOM(BASEPATH + "/about/status", false);
```
aaime, line 99: I think there are issues also at line 64 and 72

And also in ResourceControllerTest line 193

src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java
```diff
@@ -1399,29 +1399,29 @@ protected List<BufferedImage> getAsAnimation(String path, String mime) throws Ex
     /**
      * Executes an ows request using the GET method and returns the result as an xml document.
      *
-     * <p>When working with {@code html} content use {@code skipDTD=true} to false, to avoid failure due to presense of
-     * DOCTTYPE declaration in html output.
+     * <p>When working with {@code html} content use {@code useEmptyDTD=true} to false, to avoid failure due to presense
```
aaime:
```suggestion
     * <p>When working with {@code html} content use {@code useEmptyDTD=true} to false, to avoid failure due to presence
```

aaime (changes requested):
Overall looks good, see below

src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java
jodygarnett: Yes that comment no longer applies, we are not locking down the parser used for `asDOM`

src/restconfig/src/test/java/org/geoserver/rest/catalog/AboutControllerTest.java
jodygarnett: Thanks, it did not say, but the default is `html` as shown by `checkHTMLModel(dom)` below.

src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java
jodygarnett: I find useEmptyDTD to be more accurate about what the flag does.

The javadocs on `dom(stream,useEmptyDTD, encoding)` are pretty clear?

> @param useEmptyDTD {@code true} to intercept external entity references with empty xml document.

jodygarnett:
Feedback addressed, we can wait for builds

src/main/src/test/java/org/geoserver/test/GeoServerSystemTestSupport.java
aaime: I asked an AI to explain to me what useEmptyDTD meant, this is the end of the answer I got:

------

So your instinct is correct: the name is confusing.

A clearer name would have been something like:
* ignoreExternalDTD
* disableRemoteDTDLoading
* stubExternalDTD
* offlineDTDResolution

because useEmptyDTD describes the implementation trick, not the intent.


----

So ok, useEmptyDTD is how it's implemented, not what it does.

jodygarnett: I think `stubExternalDTD` is accurate, technically it does this for DTD and XSD, so `stubExternalResources`?

petersmythe:
All green

PR #9645, opened 2026-06-22
Docs: Update main branch to 3.1.x generation

jodygarnett:
The general theme is to complete the release process and have version selector provide:

* devel: 3.1
* stable: 3.0
* maintenance: 2.28

For this change I expect we will need several PRs going concurrently (rather than backport this one):

* main: To setup 3.1.x as `latest`
* 3.0.x: To setup 3.0.x as `stable` (and the default)

Things to check:

* [ ] That version chooser works
* [ ] Version warning
* [ ] Does the [release guide](https://docs.geoserver.org/main/en/developer/release/guide/) need any manual steps?

I am not having much luck testing on my on gh-pages fork, and would like a second opinion before proceeding.

Reference:

* https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/

# Checklist

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1,5 +1,5 @@
 site_name: Documentation
-site_url: https://docs.geoserver.org/3.0.x/
+site_url: https://docs.geoserver.org/3.1.x
 repo_url: https://github.com/geoserver/geoserver
 repo_name: geoserver/geoserver
 edit_uri: edit/main/doc/
@@ -58,7 +58,8 @@ theme:
 
 extra:
   version:
-    default: latest
+    default: stable
+    alias: true
     provider: mike
 
   social:
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -61,21 +61,16 @@ jobs:
           if [ "$BRANCH" = "main" ]; then
             echo "version=latest" >> $GITHUB_OUTPUT
             echo "title=main (Latest)" >> $GITHUB_OUTPUT
-            echo "aliases=main 3.0.x" >> $GITHUB_OUTPUT
-            echo "is_default=false" >> $GITHUB_OUTPUT
-          elif [ "$BRANCH" = "2.28.x" ]; then
-            echo "version=2.28.x" >> $GITHUB_OUTPUT
-            echo "title=2.28.x" >> $GITHUB_OUTPUT
-            echo "aliases=" >> $GITHUB_OUTPUT
+            echo "aliases=main 3.1.x" >> $GITHUB_OUTPUT
             echo "is_default=false" >> $GITHUB_OUTPUT
           elif [ "$BRANCH" = "3.0.x" ]; then
             echo "version=3.0.x" >> $GITHUB_OUTPUT
             echo "title=3.0.x" >> $GITHUB_OUTPUT
-            echo "aliases=" >> $GITHUB_OUTPUT
-            echo "is_default=false" >> $GITHUB_OUTPUT
-          elif [ "$BRANCH" = "3.1.x" ]; then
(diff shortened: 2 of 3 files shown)


aaime:
@jodygarnett The title says "Update main branch to 3.1.x generation", but the PR does more:
* Consolidates the REST API URL macros (api_url3/api_url4 -> api_url)
* Restores workshop example files. Some are referenced by existing markdown (ports2.sld, sprites.*), but several (airports*.sld, ports.sld, polygon_example.css) aren't referenced anywhere; are they meant to be linked, or leftovers? And if existing docs already linked missing files, why didn't the build fail?

These are independent concerns. Could you split them into separate PRs?
Having clean PRs with one concern at a time makes the review much easier, and I could have merged some parts right away.

doc/version.py
```diff
@@ -166,7 +166,8 @@ def define_env(env):
     env.variables['docs_url'] = 'https://docs.geoserver.org'
 
     # API base URL for REST API Swagger/OpenAPI specs
-    env.variables['api_url'] = '../../api'
-    env.variables['api_url3'] = '../../../api'
-    env.variables['api_url4'] = '../../../../api'
-
+    # Uses absolute url which has been configured relative_to_docs
```
petersmythe: This will need proper testing across:
* GitHub workflow:
   * https://docs.geoserver.org/main
   * https://docs-test.geoserver.org/latest/ (no longer needed?)
   * http://petersmythe.github.io/geoserver/main (no longer needed?)
* locally built documentation
* HTML documentation (will Swagger API render anyway?)
* ...

but we might be able to remove the hack, if we now only publish to a standard depth path based on https://docs.geoserver.org/

mkdocs.yml
```diff
@@ -109,11 +114,17 @@ extra_javascript:
   - javascripts/hide-empty-toc.js
   - javascripts/version-archive.js
 
+validation:
+  links:
+    absolute_links: relative_to_docs
```
petersmythe: or does this fix the API problem above?  If so, cool!

doc/version.py
petersmythe: Testing locally was always a challenge, which is why I had a branch that would deploy to http://petersmythe.github.io/geoserver/<branch> -> https://docs-test.geoserver.org/<branch>

Is a mechanism like this still needed, for pre-deployment testing?

jodygarnett:
> @jodygarnett The title says "Update main branch to 3.1.x generation", but the PR does more

Yes my initial goal was to get read of the warnings, and was surprised to find it generating the 3.0.x docs. Let me back out the version number change and we can focus on the warnings.

I initially had warnings due to the "missing" sprite files, but it ended up being junk leftover from the conversion process.

jodygarnett:
Okay this is just focused on the version number changes. I think we will need to get it published correctly and the version selector working, and then worry about getting the correct version number passed into the documentation examples in another PR.

doc/version.py
```diff
@@ -63,7 +63,7 @@ def _guess_version(branch, default='3.1'):
     return default
 
 
-def _guess_release(branch, version, default='3.1.0'):
+def _guess_release(branch, version, default='3.0.0'):
```
bradh: Is `3.0.0` meant to be the default?

jodygarnett: Perhaps not, but my understanding is release should be the most recent published release to be used in code examples and similar. As such I would expect `3.0.0` to be correct? We may need to look up a useful value from GitHub tags as we will be unable to determine from source code alone.

jodygarnett:
Okay now that warnings are addressed I am restarting this PR changing changing `main` to 3.1 generation.

We will need a companion PR for `3.0.x` branch I expect.

PR #9647, opened 2026-06-22
changed the GeoServer login flow from plain/Base64 credentials to RSA…

ziyudebug:
…-encrypted credentials: the server creates a SecurityEncryptionManager bean with an RSA key pair, injects the public key into the page as window.gs_public_key, loads jsencrypt.min.js, and intercepts login form submissions in geoserver.js to replace username and password with CRYPT:<RSA ciphertext> before posting to j_spring_security_check; on the server side, GeoServerUserNamePasswordAuthenticationFilter decrypts those CRYPT: values with the private key before passing them to Spring Security authentication.

  

# Checklist

- [ ✓] I have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/main/CONTRIBUTING.md).
- [ ✓] I have sent a [Contribution Licence Agreement](https://docs.geoserver.org/latest/en/developer/policies/committing.html) (not required for small changes, e.g., fixing typos in documentation).
- [ ✓] First PR targets the `main` branch (backports managed later; ignore for branch specific issues).

In some deployment environments with strict security scanning and network inspection policies, transmitting credentials in plaintext (even over HTTPS) can trigger security alerts or compliance violations.

This PR introduces client-side RSA encryption for login credentials to mitigate concerns about credential exposure during transmission and to improve compatibility with security-sensitive environments where additional encryption layers are required.

The goal is to ensure GeoServer can be safely deployed in enterprise and regulated network environments without triggering false-positive security warnings related to plaintext credential handling.


aaime:
I had a look; it looks ok at a quick glance, but after a bit more thinking, the idea behind it doesn't stand up.

The motivation behind this PR is that sending credentials in plaintext is a risk, but nobody deploys a GeoServer over HTTP.  HTTPS already protects the credentials in transit. Inside a TLS session, the credentials aren't on the wire in plaintext, so there's nothing here for the RSA layer to protect that TLS isn't already protecting. 

From that premise, it gets worse: the public key gets shipped unauthenticated in the page itself, so anyone in a position to read the TLS stream can also swap the key out. There's no nonce, so the encrypted blob replays cleanly. You end up with something weaker than the certificate that's already securing the connection, dressed up to look like an extra layer of security.

Even with the security argument aside, there are more problems: the key is per-JVM and in-memory, so any clustered setup without sticky sessions breaks; there's no way to turn it off; and if the JS throws an exception, the login button just stops working.

I'm trying to guess why you got into this development to start with... maybe it's a security scanner complaining about "plaintext credentials over HTTPS"?  That would be a false positive, and the answer would be to fix it in scanner config, compliance policy, or by keeping creds out of your proxy's access logs.

If you've got a concrete deployment where HTTPS/TLS genuinely isn't enough, then please start discussing it on the [developer list](https://discourse.osgeo.org/c/geoserver/developer/63), unless you're trying to address a vulnerability, in that case, please [read the project security policy](https://github.com/geoserver/geoserver/security/policy) and then use the private security list instead. 

As it stands today, I don't think there's a version of this worth merging.

PR #9648, opened 2026-06-22
[GEOS-12140]: Better Cropping on raster mask filter, rescaling to target resolution before the crop

dromagnoli:
[GEOS-12140](https://osgeo-org.atlassian.net/browse/GEOS-12140) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

  

# Checklist

### DIFF
--- a/src/main/src/main/java/org/geoserver/security/decorators/SecuredGridCoverage2DReader.java
+++ b/src/main/src/main/java/org/geoserver/security/decorators/SecuredGridCoverage2DReader.java
@@ -5,29 +5,41 @@
  */
 package org.geoserver.security.decorators;
 
+import java.awt.Rectangle;
+import java.awt.image.RenderedImage;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
+import org.eclipse.imagen.Interpolation;
 import org.geoserver.catalog.Predicates;
 import org.geoserver.data.util.CoverageUtils;
+import org.geoserver.ows.Dispatcher;
+import org.geoserver.ows.Request;
 import org.geoserver.security.CoverageAccessLimits;
 import org.geoserver.security.WrapperPolicy;
 import org.geotools.api.coverage.grid.Format;
+import org.geotools.api.coverage.processing.Operation;
 import org.geotools.api.data.ResourceInfo;
 import org.geotools.api.data.ServiceInfo;
 import org.geotools.api.filter.Filter;
 import org.geotools.api.parameter.GeneralParameterDescriptor;
 import org.geotools.api.parameter.GeneralParameterValue;
+import org.geotools.api.parameter.ParameterNotFoundException;
 import org.geotools.api.parameter.ParameterValue;
 import org.geotools.api.parameter.ParameterValueGroup;
+import org.geotools.api.referencing.crs.CoordinateReferenceSystem;
 import org.geotools.coverage.grid.GridCoverage2D;
+import org.geotools.coverage.grid.GridCoverageFactory;
 import org.geotools.coverage.grid.io.GridCoverage2DReader;
 import org.geotools.coverage.processing.CoverageProcessor;
 import org.geotools.coverage.processing.operation.Crop;
+import org.geotools.coverage.processing.operation.Scale;
 import org.geotools.geometry.jts.JTS;
 import org.geotools.geometry.jts.ReferencedEnvelope;
+import org.geotools.referencing.CRS;
 import org.geotools.util.factory.Hints;
 import org.locationtech.jts.geom.Envelope;
 import org.locationtech.jts.geom.Geometry;
@@ -45,6 +57,9 @@ public class SecuredGridCoverage2DReader extends DecoratingGridCoverage2DReader 
 
     /** Cached crop factory */
     private static final Crop coverageCropFactory = new Crop();
+
+    /** Cached scale factory */
+    private static final Scale coverageScaleFactory = new Scale();
 
     static {
         final CoverageProcessor processor = new CoverageProcessor(new Hints(Hints.LENIENT_DATUM_SHIFT, Boolean.TRUE));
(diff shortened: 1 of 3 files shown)


src/main/src/main/java/org/geoserver/security/decorators/SecuredGridCoverage2DReader.java
```diff
@@ -142,17 +155,142 @@ static GridCoverage2D read(GridCoverage2DReader delegate, WrapperPolicy policy,
...
+                                rasterFilter.getEnvelopeInternal(), grid.getCoordinateReferenceSystem2D()));
+
+                // The underlying reader may have returned a coverage with a different resolution than the one
+                // requested. The requested gridGeometry may have been limited too, due to reaching the
+                // Max Oversampling Factor.
+                //
+                // This happens for example when the data resolution is bad and the map is heavily oversampled.
+                // We want to scale it to the requested map raster extent before cropping to the geometry.
+
+                grid = scaleToRequestedSize(grid, getRequestedMapRasterArea(), interpolation);
```
aaime: grid could be null, but scaleToRequestedSize does not check it, while it's checked below at line 182. I would just exit right away if the grid happened to be null. Please add a test for this.

src/main/src/main/java/org/geoserver/security/decorators/SecuredGridCoverage2DReader.java
```diff
@@ -142,17 +155,142 @@ static GridCoverage2D read(GridCoverage2DReader delegate, WrapperPolicy policy,
...
+    }
+
+    private static <T> T getProperty(Object object, String property, Class<T> type) {
+        if (!OwsUtils.has(object, property)) {
+            return null;
+        }
+        return OwsUtils.property(object, property, type);
+    }
+
+    private static GridCoverage2D scaleToRequestedSize(
```
aaime, line 249: Hum... the grid could be covering only part of the requested area, isn't this scaling its size up to the full requested size? That would make sense if the grid was covering the whole requested size.

src/main/src/main/java/org/geoserver/security/decorators/SecuredGridCoverage2DReader.java
```diff
@@ -142,17 +155,142 @@ static GridCoverage2D read(GridCoverage2DReader delegate, WrapperPolicy policy,
...
+    private static Rectangle getRequestedMapRasterArea() {
+        Request request = Dispatcher.REQUEST.get();
+        if (request == null
+                || request.getOperation() == null
+                || request.getOperation().getParameters() == null) {
+            return null;
+        }
+
+        for (Object parameter : request.getOperation().getParameters()) {
+            Rectangle mapExtent = getRequestedMapRasterArea(parameter);
```
aaime: Hum... looking for width/height is quite WMS-specific, and at that point you could use the dispatcher more directly (getting a WMS request object). In fact, this code path is fully WMS-specific... and at that point I would spare the reflection and just grab the parsed KVP from REQUEST, width and height should be in there, already parsed to integer.

If you want to make it fully protocol-independent instead, what about using the grid geometry being requested... shouldn't it indicate the area of interest and the target resoltuion?

aaime (commented):
Overall looks good, see comments below.

src/main/src/main/java/org/geoserver/security/decorators/SecuredGridCoverage2DReader.java
dromagnoli: Ok. I will follow that approach in getting from REQUEST.

In terms of using the requested gridGeometry it won't work.
At the beginning I was using the requested gridGeometry but there are pieces of code (I should re-check where, I think somewhere in the ReadResolutionCalculator) that won't go over the configured max oversampling factor. So when there is a big oversampling (default threshold is around 30 if my memory serves me right), they get limited and the target resolution modified and you get back again pixellated results.

src/main/src/main/java/org/geoserver/security/decorators/SecuredGridCoverage2DReader.java
dromagnoli: Good catch... I think that I could apply a correcting factor that relates requested envelope to grid envelope so that we apply the same "percentage" to the raster size.

src/main/src/main/java/org/geoserver/security/decorators/SecuredGridCoverage2DReader.java
```diff
@@ -142,17 +157,224 @@ static GridCoverage2D read(GridCoverage2DReader delegate, WrapperPolicy policy,
...
+        Rectangle requestedGridRange = requestedMapArea.gridRange;
+        int requestedWidth = (int) requestedGridRange.getWidth();
+        int requestedHeight = (int) requestedGridRange.getHeight();
+        Envelope coveredRequestedEnvelope = scalingEnvelope.intersection(coverageRequestedEnvelope);
+        // WMS 1.3.0 with geographic CRS (e.g. EPSG:4326): axis 0 = lat = vertical, axis 1 = lon = horizontal.
+        // Swap span axes so WIDTH pixels map to lon extent and HEIGHT pixels map to lat extent.
+        boolean northEast =
+                CRS.getAxisOrder(coverageRequestedEnvelope.getCoordinateReferenceSystem()) == CRS.AxisOrder.NORTH_EAST;
+        if (northEast) {
+            requestedWidth = getRequestedSize(
```
aaime, line 275: This branch is tricky but has no tests covering it.

src/main/src/main/java/org/geoserver/security/decorators/SecuredGridCoverage2DReader.java
```diff
@@ -142,17 +157,224 @@ static GridCoverage2D read(GridCoverage2DReader delegate, WrapperPolicy policy,
...
+        if (coverageCRS == null) {
+            return null;
+        }
+
+        try {
+            CoordinateReferenceSystem requestedCRS = referencedRequestedEnvelope.getCoordinateReferenceSystem();
+            if (CRS.equalsIgnoreMetadata(requestedCRS, coverageCRS)) {
+                return referencedRequestedEnvelope;
+            }
+            return referencedRequestedEnvelope.transform(coverageCRS, true);
```
aaime, line 324: Are there tests covering different CRSs, exercising the transformation?

aaime (changes requested):
Had a second look, the code now looks good, but could use a bit more testing.
I would also like to see an end-to-end WMS GetMap test (as a rule of thumb, unit tests are great to cover corner cases, but some end-to-end testing should be implemented to ensure the oevrall fix persists over time).

src/main/src/main/java/org/geoserver/security/decorators/SecuredGridCoverage2DReader.java
dromagnoli: Adding it

src/main/src/main/java/org/geoserver/security/decorators/SecuredGridCoverage2DReader.java
dromagnoli: Adding it

dromagnoli:
> Had a second look, the code now looks good, but could use a bit more testing. I would also like to see an end-to-end WMS GetMap test (as a rule of thumb, unit tests are great to cover corner cases, but some end-to-end testing should be implemented to ensure the oevrall fix persists over time).

Also added a getMap test involving a triangular crop on a low resolution image.
https://github.com/dromagnoli/geoserver/blob/63eaf83828e64e64141102b39a0bdba3270aa6a2/src/wms1_1/src/test/java/org/geoserver/wms/ResourceAccessManagerWMSTest.java#L390

dromagnoli:
All feedbacks have been addressed.

PR #9652, opened 2026-06-23
Docs: Address warnings by switching to absolute REST API links

jodygarnett:
This PR is focused on reducing warnings noticed while reviewing mkdocs output.

The key think is `api` references are always going to show up as broken links, to resolve this:

* api references are now absolute links, based on `site_url` path (so it works locally)
* validation for absolute links is set it `ignore`

The result now only shows warnings for actually broken links:

```
WARNING -  Doc file 'en/user/data/app-schema/wms-support.md' contains a link '../../tutorials/freemarker.md', but the target
           'en/user/tutorials/freemarker.md' is not found among documentation files.
WARNING -  Doc file 'en/user/datadirectory/location.md' contains a link '../../installation/upgrade3.md', but the target
           'en/installation/upgrade3.md' is not found among documentation files. Did you mean '../installation/upgrade3.md'?
```

This is a companion of:

* https://github.com/geoserver/geoserver/pull/9645

Thanks for all the hard work @petersmythe I really like how the REST API docs are include in the mkdocs output.

# Checklist


doc/en/user/styling/workshop/mbstyle/mbstyle.md
```diff
@@ -153,6 +153,11 @@ We use a point symbolizer to indicate we want this content drawn as a **Point**
 
 !!! note
     Rather than refer to many different icons separately, MBStyles use a single sprite-sheet containing all the necessary icons for the style. This is defined by the `sprite` property at the top-level of the style.
+    
+    The supprt files to be added to your **`styles`** folder:
```
aaime, line 157:
```suggestion
    The support files to be added to your **`styles`** folder:
```

doc/en/user/styling/workshop/mbstyle/mbstyle.md
```diff
@@ -153,6 +153,11 @@ We use a point symbolizer to indicate we want this content drawn as a **Point**
...
+    
+    * [sprites.json](../files/sprites.json"}{:download="sprites.json"}
```
aaime, line 159: I think we have an extra " here?

```suggestion
    * [sprites.json](../files/sprites.json}{:download="sprites.json"}
```

mkdocs.yml
```diff
@@ -109,11 +110,17 @@ extra_javascript:
   - javascripts/hide-empty-toc.js
   - javascripts/version-archive.js
 
+validation:
+  links:
+    absolute_links: relative_to_docs
+    anchors: ignore
+    unrecognized_links: warn
+    not_found: ignore
```
aaime: Hum... won't this skip page-to-page validation too?
Say one has page1 pointing to page2, but page2 is gone or has a different name... will they get a broken link report?

doc/version.py
```diff
@@ -166,7 +166,8 @@ def define_env(env):
...
     # API base URL for REST API Swagger/OpenAPI specs
-    env.variables['api_url'] = '../../api'
-    env.variables['api_url3'] = '../../../api'
-    env.variables['api_url4'] = '../../../../api'
-
+    # Uses absolute url which has been configured relative_to_docs
+    if is_snapshot:
+        env.variables['api_url'] = '/' + version + '.x/en/api'
+    else:
+        env.variables['api_url'] = '/' + branch + '/en/api'
```
aaime: The previous links were working correctly locally too, these new ones will work only from the web site and introduce risks with mismatched versions/branches. I liked the previous approach better...

Claude suggests a macro approach that might solve the issue and allow using a single URL, I'm not familiar with mkdocs enough to evaluate, but you might. It looks like this:

```
  doc/version.py — replace the PR's absolute block with:
      # API base URL; recomputed per page in on_pre_page_macros as a relative link
      env.variables['api_url'] = 'en/api'


  def on_pre_page_macros(env):
      # Relative path from the current page to en/api; mount- and version-agnostic
      depth = len(env.page.url.rstrip('/').split('/'))
      env.variables['api_url'] = '../' * depth + 'en/api'

  mkdocs.yml — no relative_to_docs needed, restore real validation:
  validation:
    links:
      anchors: ignore
      unrecognized_links: warn
      not_found: warn
```

aaime (changes requested):

doc/version.py
jodygarnett: Okay I like the proposed changes to `doc/version.py` they are fine; anything to avoid having to know how deep a document is when referencing rest api.


However I still like `validation=relative_to_docs` setting, as it allows an easy way to reference other sections.

mkdocs.yml
jodygarnett: The existing page to page links (using relative paths) is checked.

I should note that the absolute_links: relative_to_docs setting generates our relative html links at the end of the day. So they work fine locally, and when hosted on docs.geoserver.org.

Let me update the description so that is clear.

doc/version.py
jodygarnett: Oh wait in the specific case of api doc references we get lots of these warnings:
```
WARNING -  Doc file 'en/release/rest.md' contains a link 'opensearch-eo/index.html', but the target 'en/release/opensearch-eo/index.html' is not
           found among documentation files.
```

And there is no way to suppress those. The proposed change to `doc/version.py` would not succeed in suppressing these warnings.

mkdocs.yml
jodygarnett: Testing this myself and it is not working.

jodygarnett: Okay thanks for the review @aaime 
- Validating with `not_found: warn`
- Using `site_url` rather than trying to guess based on version or branch

doc/version.py
jodygarnett: Okay I was able to get a clean output with good warnings, using some inspiration from your suggestions above:

* Validating with `not_found: warn` 
* Using `site_url` rather than trying to guess based on version or branch

PR #9664, opened 2026-07-01
[GEOS-12144] "application/vnd.ogc.fg+json" output format fails when requesting non-geographical datasets

pmauduit:
[GEOS-12144](https://osgeo-org.atlassian.net/browse/GEOS-12144) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

Attempt to fix GEOS-12144: avoid to raise a NullPointerException when requesting a non-geographical dataset from the ogcapi as geojson output format.

The NPE is being thrown when GS tries to calculate the URL to the CRS (which is irrelevant considering non-geo datasets).

This has been manually/runtime tested.
  

# Checklist

### DIFF
--- a/src/extension/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/JSONFGFeatureWriter.java
+++ b/src/extension/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/JSONFGFeatureWriter.java
@@ -120,6 +120,7 @@ public class JSONFGFeatureWriter<T extends FeatureType, F extends Feature> exten
     }
 
     private String getCRSURI(CoordinateReferenceSystem crs) {
+        if (crs == null) return null;
         try {
             // prefer EPSG authority if possible, more commonly understood
             String epsgIdentifier = CRS.lookupIdentifier(Citations.EPSG, crs, false);
--- a/src/extension/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/JSONFGFeatureWriter.java
+++ b/src/extension/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/JSONFGFeatureWriter.java
@@ -112,6 +112,7 @@ public class JSONFGFeatureWriter<T extends FeatureType, F extends Feature> exten
     }
 
     private void writeCoordRefSys(GeoJSONBuilder jsonWriter, CoordinateReferenceSystem crs) {
+        if (crs == null) return;
         // write the CRS block only if needed
         if (!CRS.equalsIgnoreMetadata(DefaultGeographicCRS.WGS84, crs)) {
             jsonWriter.key(JSONFGFeaturesResponse.COORD_REF_SYS);
@@ -120,7 +121,6 @@ public class JSONFGFeatureWriter<T extends FeatureType, F extends Feature> exten
     }
 
     private String getCRSURI(CoordinateReferenceSystem crs) {
-        if (crs == null) return null;
         try {
             // prefer EPSG authority if possible, more commonly understood
             String epsgIdentifier = CRS.lookupIdentifier(Citations.EPSG, crs, false);
--- a/src/extension/ogcapi/ogcapi-features/src/test/java/org/geoserver/ogcapi/v1/features/JSONFGFeatureWriterTest.java
+++ b/src/extension/ogcapi/ogcapi-features/src/test/java/org/geoserver/ogcapi/v1/features/JSONFGFeatureWriterTest.java
@@ -7,6 +7,8 @@ package org.geoserver.ogcapi.v1.features;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
 
+import com.jayway.jsonpath.JsonPath;
+import com.jayway.jsonpath.internal.JsonContext;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
--- a/src/extension/ogcapi/ogcapi-features/src/test/java/org/geoserver/ogcapi/v1/features/JSONFGFeatureWriterTest.java
+++ b/src/extension/ogcapi/ogcapi-features/src/test/java/org/geoserver/ogcapi/v1/features/JSONFGFeatureWriterTest.java
@@ -72,7 +72,8 @@ public class JSONFGFeatureWriterTest {
 
         toTest.write(geoCollection(), out, BigInteger.TEN, false);
 
-        assertThat(out.toString(), containsString("\"type\":\"FeatureCollection\""));
+        JsonContext json = (JsonContext) JsonPath.parse(out.toString());
+        assertEquals("FeatureCollection", json.read("$.type", String.class));
     }
 
     @Test
--- a/src/extension/ogcapi/ogcapi-features/src/test/java/org/geoserver/ogcapi/v1/features/JSONFGFeatureWriterTest.java
+++ b/src/extension/ogcapi/ogcapi-features/src/test/java/org/geoserver/ogcapi/v1/features/JSONFGFeatureWriterTest.java
@@ -4,7 +4,6 @@
  */
 package org.geoserver.ogcapi.v1.features;
 
-import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.MatcherAssert.assertThat;
 
 import java.io.ByteArrayOutputStream;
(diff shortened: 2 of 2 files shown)


aaime:
We don't accept manually tested PRs, a test needs to be written for the PR to be merged, this is code in supported land.
The change per se looks ok. I believe you can find some test setup code for geometryless data in the wfs modules.

pmauduit:
I added some unit tests to test both (geo & non-geo) cases.

My concern about the test class itself is that I could not find anything for inspiration after a rapid grep, so it could read weird compared to the existing codebase. Also the assertions are based on substrings, which could be not that relevant when testing generated json.

src/extension/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/JSONFGFeatureWriter.java
```diff
@@ -120,6 +120,7 @@ private void writeCoordRefSys(GeoJSONBuilder jsonWriter, CoordinateReferenceSyst
     }
 
     private String getCRSURI(CoordinateReferenceSystem crs) {
+        if (crs == null) return null;
```
ianturton: would an empty string be a better sign here? other wise we're just pushing the NPE up a level or two

src/extension/ogcapi/ogcapi-features/src/test/java/org/geoserver/ogcapi/v1/features/JSONFGFeatureWriterTest.java
```diff
@@ -0,0 +1,86 @@
...
+        assertThat(out.toString(), containsString("\"type\":\"FeatureCollection\""));
+    }
+
+    @Test
+    public void testFeatureWriterNonGeo() throws SchemaException, IOException, ParseException {
+        OutputStream out = new ByteArrayOutputStream();
+
+        toTest.write(nonGeoCollection(), out, BigInteger.TEN, false);
+
+        assertThat(out.toString(), containsString("\"geometry\":null"));
```
ianturton: Have a look at org.geotools.geojson.GeoJSONTestSupport or possibly something like https://github.com/lukas-krecan/JsonUnit to check there are actually the same features in the output as the input

ianturton (changes requested):

src/extension/ogcapi/ogcapi-features/src/test/java/org/geoserver/ogcapi/v1/features/JSONFGFeatureWriterTest.java
pmauduit: Thanks for the hint, I will have a look

pmauduit: @ianturton I just opened the GeoJSONTestSupport class and I was wondering if there wasn't a typo here:
https://github.com/geotools/geotools/blob/main/modules/unsupported/geojson/src/test/java/org/geotools/geojson/GeoJSONTestSupport.java#L48
`f2.getID()` I'd assume ? (If so, I can open a PR on geotools, while I am on it).

ianturton: looks like it to me, so please do

src/extension/ogcapi/ogcapi-features/src/main/java/org/geoserver/ogcapi/v1/features/JSONFGFeatureWriter.java
pmauduit: relying on my tests, it does not change the behaviour of the writer (in both cases I get a `geometry: null` in the generated JSON which is what I expect), so either null or empty string should do.

pmauduit: I was actually looking at the wrong place in the generated JSON: empty string or null will alter the value of the `coordRefSys` attribute at the root of the json object.
Reading the spec, it does not look like a null or empty value has been specified: https://docs.ogc.org/is/21-045r1/21-045r1.html#ref-sys 
I am wondering then if the right fix should not be to omit the coordRefSys completely in case of non-geographical featureset.

src/extension/ogcapi/ogcapi-features/src/test/java/org/geoserver/ogcapi/v1/features/JSONFGFeatureWriterTest.java
pmauduit: done: https://github.com/geotools/geotools/pull/5757

pmauduit: checking JsonUnit it looks interesting, but am I allowed to add it in the test classpath ? grepping for it in both GS and GT, I could not see it being a dependency of any modules.

src/extension/ogcapi/ogcapi-features/src/test/java/org/geoserver/ogcapi/v1/features/JSONFGFeatureWriterTest.java
```diff
@@ -0,0 +1,86 @@
+/* (c) 2026 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license, available at the root
+ * application directory.
+ */
+package org.geoserver.ogcapi.v1.features;
+
+import static org.hamcrest.CoreMatchers.containsString;
```
mprins:
```suggestion

```

src/extension/ogcapi/ogcapi-features/src/test/java/org/geoserver/ogcapi/v1/features/JSONFGFeatureWriterTest.java
```diff
@@ -0,0 +1,86 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
```
mprins:
```suggestion
import static org.junit.Assert.assertEquals;
import static org.junit.Assert.assertNull;
```

src/extension/ogcapi/ogcapi-features/src/test/java/org/geoserver/ogcapi/v1/features/JSONFGFeatureWriterTest.java
```diff
@@ -0,0 +1,86 @@
...
+        return List.of(featureCollection);
+    }
+
+    @Test
+    public void testFeatureWriterGeo() throws SchemaException, ParseException, IOException {
+        OutputStream out = new ByteArrayOutputStream();
+
+        toTest.write(geoCollection(), out, BigInteger.TEN, false);
+
+        assertThat(out.toString(), containsString("\"type\":\"FeatureCollection\""));
```
mprins: use JsonContext like the other tests in this package

```suggestion
        JsonContext json = (JsonContext) JsonPath.parse(out.toString());
        assertEquals("FeatureCollection", json.read("$.type", String.class));
```

src/extension/ogcapi/ogcapi-features/src/test/java/org/geoserver/ogcapi/v1/features/JSONFGFeatureWriterTest.java
mprins:
```suggestion
        assertEquals("FeatureCollection", JsonPath.parse(out.toString()).read("$.type", String.class));
        assertNull(JsonPath.parse(out.toString()).read("$.features[0].geometry", String.class));
```

mprins (changes requested):
Use JsonContext / JsonPath like the other tests in this package

pmauduit:
> Use JsonContext / JsonPath like the other tests in this package

sorry for having missed the library, I updated the code

mprins (approved):

jratike80:
I suggest creating a ticket at [https://github.com/opengeospatial/ogc-feat-geo-json](https://github.com/opengeospatial/ogc-feat-geo-json) and ask how such case should be handled in JSON-FG

By the way, the standard has a note

> Earlier drafts of JSON-FG used the media types application/fg+json and application/vnd.ogc.fg+json. Implementations may continue to support these media types for backward compatibility. However, the media type application/geo+json is the preferred media type for JSON-FG documents conforming to this Standard.

But I do not know how a client should make a difference between requesting baseline GeoJSON without any JSON-FG features vs JSON-FG with "place" etc. then.

PR #9666, opened 2026-07-01
[GEOS-12146] Added clip geometry reprojection to the collection CRS in SecuredFeatureSource.

axl8713:
[GEOS-12146](https://osgeo-org.atlassian.net/browse/GEOS-12146) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

While applying a clipping filter on a layer in the context of a WFS 2.0.0 request, the filter is considered with the wrong axis order, causing the clip to return wrong results.

This issue stems from the fact that the clip polygon is actually not reprojected to the features CRS, but only on the layer native SRS.

These changes add the reprojection of the clipping polygon to the features CRS.

  

# Checklist


src/security/security-tests/src/test/java/org/geoserver/security/decorators/SecuredFeatureSourceTest.java
```diff
@@ -164,4 +176,61 @@ public void close() throws SecurityException {}
...
+        fb.add("Texas");
+        delegateCollection.add(fb.buildFeature("states.1"));
+        fb.reset();
+
+        Geometry northCarolina = reader.read("POLYGON((35 -80, 35 -78, 37 -78, 37 -80, 35 -80))");
+        fb.add(northCarolina);
+        fb.add("North Carolina");
+        delegateCollection.add(fb.buildFeature("states.2"));
+
+        // clip area as stored by GeoFence, in conventional (lon, lat) order, SRID 4326
```
aaime: In this context GeoFence does not exist, I would say something a bit more generic:


```suggestion
        // clip area as stored by the ResourceAccessManager, in conventional (lon, lat) order, SRID 4326
```

src/security/security-tests/src/test/java/org/geoserver/security/decorators/SecuredFeatureSourceTest.java
```diff
@@ -164,4 +176,61 @@ public void close() throws SecurityException {}
...
+        Geometry clip = reader.read("POLYGON((-106.8 24, -93 24, -93 36.6, -106.8 36.6, -106.8 24))");
+        clip.setSRID(4326);
+
+        VectorAccessLimits limits =
+                new VectorAccessLimits(CatalogMode.HIDE, null, Filter.INCLUDE, null, Filter.INCLUDE);
+        limits.setClipVectorFilter(clip);
+
+        SimpleFeatureSource fs = createNiceMock(SimpleFeatureSource.class);
```
aaime: No need to use a mock here, GeoTools provides a fully working facility around a collection:

DataUtilities.source(delegateCollection)

src/main/src/main/java/org/geoserver/security/decorators/SecuredFeatureSource.java
```diff
@@ -136,6 +143,20 @@ private FeatureCollection<T, F> decoratesForClipping(
...
+    private static Geometry reprojectToCollectionCRS(Geometry geometry, CoordinateReferenceSystem targetCRS) {
+        if (geometry == null || targetCRS == null || geometry.getSRID() == 0) return geometry;
+        try {
+            CoordinateReferenceSystem geomCRS = CRS.decode("EPSG:" + geometry.getSRID(), true);
+            MathTransform mt = CRS.findMathTransform(geomCRS, targetCRS, true);
+            if (!mt.isIdentity()) {
+                return JTS.transform(geometry, mt);
+            }
+        } catch (Exception e) {
+            LOGGER.log(Level.WARNING, "Could not reproject clip geometry to collection CRS", e);
```
aaime: This could be problematic, if there is a reprojection failure one can then see the entire dataset... since it's security, probably better to throw an exception? You already proved that a reprojection is needed, the original geometry is unlikely to cut anything.

aaime (changes requested):
Overall looks good, see comments below

src/security/security-tests/src/test/java/org/geoserver/security/decorators/SecuredFeatureSourceTest.java
axl8713: Nice! Fixed. 
Thanks.

src/main/src/main/java/org/geoserver/security/decorators/SecuredFeatureSource.java
axl8713: Right, thanks.
Fixed.

aaime:
LGTM

aaime (approved):

aaime:
@axl8713 the automated backports failed, you'll have to do them manually. Please remember about the 3.0.x branch (the current stable branch) and open PRs with a single backport commit.

PR #9355, opened 2026-04-14
[GEOS-12119] Workspace-scoped OGC API Styles endpoint returns styles from other workspaces

jurgispods:
[GEOS-12119](https://osgeo-org.atlassian.net/browse/GEOS-12119) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

When using the OGC API Styles extension to list styles via a workspace-scoped service like

```
myworkspace/ogc/styles/v1/styles
```

the endpoint currently returns all styles, even those present in other workspaces. Additionally, the URLs returned in the `href` field for styles from other workspaces return 404 as they are prefixed by the wrong workspace. This PR makes sure that workspace-scoped list requests only return styles in the respective workspace and global styles.
  

# Checklist

- [x] I have read the [contribution guidelines](https://github.com/geoserver/geoserver/blob/main/CONTRIBUTING.md).
- [x] I have sent a [Contribution Licence Agreement](https://docs.geoserver.org/latest/en/developer/policies/committing.html) (not required for small changes, e.g., fixing typos in documentation).
- [x] First PR targets the `main` branch (backports managed later; ignore for branch specific issues).

For core and extension modules:

- [x] New unit tests have been added covering the changes.
- [ ] [Documentation](https://github.com/geoserver/geoserver/tree/main/doc/en/user/source) has been updated (if change is visible to end users).
- [ ] The [REST API docs](https://github.com/geoserver/geoserver/tree/main/doc/en/api/1.0.0) have been updated (when changing configuration objects or the REST controllers).
- [ ] There is an issue in the [GeoServer Jira](https://osgeo-org.atlassian.net/browse/GEOS/summary) (except for changes that do not affect administrators or end users in any way).
- [ ] Commit message(s) must be in the form ``[GEOS-XYZWV] Title of the Jira ticket``.
- [x] Bug fixes and small new features are presented as a single commit.
- [x] Each commit has a single objective (if there are multiple commits, each has a separate JIRA ticket describing its goal).

Note: There is no JIRA ticket since I was not able to create one even after logging in (via Google SSO):
> You are not authorized to perform this operation. Please log in.

PR #9705, opened 2026-07-17
fix: Do not load complete catalog in workspace-scoped requests

jurgispods:
# Checklist

# Description

We noticed that simple workspace-scoped requests like 

```
/my_workspace/ows?service=wfs&version=2.0.0&request=GetCapabilities
``` 

can take a long time even on small workspaces because the `LocalWorkspaceCatalog` loads all featuretypes first before filtering them down to the virtual service's workspace. 

On core Geoserver, this is not a big issue, as the catalog lives in-memory, but on geoserver-cloud with the pgconfig backend and a large catalog of ~40k layers, a single request can take ~30s, even for small workspaces. The reason is that the complete catalog needs to be loaded into memory and filtered. This PR fixes this by loading only the relevant featuretypes from the local workspace in the first place (which is an efficient filtered db query on the pgconfig backend).

### DIFF
--- a/src/main/src/main/java/org/geoserver/catalog/impl/LocalWorkspaceCatalog.java
+++ b/src/main/src/main/java/org/geoserver/catalog/impl/LocalWorkspaceCatalog.java
@@ -128,6 +128,13 @@ public class LocalWorkspaceCatalog extends AbstractCatalogDecorator implements C
 
     @Override
     public List<FeatureTypeInfo> getFeatureTypes() {
+        WorkspaceInfo localWorkspace = LocalWorkspace.get();
+        if (localWorkspace != null) {
+            NamespaceInfo ns = super.getNamespaceByPrefix(localWorkspace.getName());
+            if (ns != null) {
+                return getFeatureTypesByNamespace(ns);
+            }
+        }
         return wrapFeatureTypesListIfNeeded(super.getFeatureTypes());
     }
 
@@ -147,8 +154,14 @@ public class LocalWorkspaceCatalog extends AbstractCatalogDecorator implements C
 
     @Override
     public List<CoverageInfo> getCoverages() {
-        List<CoverageInfo> coverages = super.getCoverages();
-        return wrapCoverageListIfNeeded(coverages);
+        WorkspaceInfo localWorkspace = LocalWorkspace.get();
+        if (localWorkspace != null) {
+            NamespaceInfo ns = super.getNamespaceByPrefix(localWorkspace.getName());
+            if (ns != null) {
+                return getCoveragesByNamespace(ns);
+            }
+        }
+        return wrapCoverageListIfNeeded(super.getCoverages());
     }
 
     @Override
--- a/src/main/src/test/java/org/geoserver/catalog/impl/LocalWorkspaceCatalogTest.java
+++ b/src/main/src/test/java/org/geoserver/catalog/impl/LocalWorkspaceCatalogTest.java
@@ -19,6 +19,7 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import org.geoserver.catalog.Catalog;
+import org.geoserver.catalog.CoverageInfo;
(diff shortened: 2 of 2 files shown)


groldan:
Hi @jurgispods,

thanks for the contribution. As marked by @mprins we'd need a jira ticket (see [CONTRIBUTING.md](https://github.com/geoserver/geoserver/blob/main/CONTRIBUTING.md)) and a matching commit message (e.g. `[GEOS-XXX] load workspace-scoped featuretypes in LocalWorkspaceCatalog`) or so. 

As for the fix itself, I'm under the impression the returned short-circuited lists should still go through `wrapFeatureTypesListIfNeeded`/`wrapCoverageListIfNeeded`? 

FWIW, this seems to be a regression I didn't catch when removing a lot of custom code from geoserver cloud to be more aligned with geoserver core, but not entirely sure.

jurgispods:
Hi @groldan, thanks for getting back so fast.

About the JIRA ticket: I've tried two accounts (Google SSO / separate e-mail), 4 different browsers (Firefox / Chrome, each desktop and mobile). _Sometimes_ I am able to log in, but I am not able to create a ticket. Either the button does not work at all or it says I need to log in and don't have permission to create a ticket. This is the same behaviour as in my two previous PRs (#8693, #9355). At this point, I think the JIRA board might be broken. What dark magic do I need to invoke to be able to create a ticket?

> As for the fix itself, I'm under the impression the returned short-circuited lists should still go through wrapFeatureTypesListIfNeeded/wrapCoverageListIfNeeded?

Ok, will take a look at that.

src/main/src/main/java/org/geoserver/catalog/impl/LocalWorkspaceCatalog.java
```diff
@@ -128,6 +128,13 @@ private FeatureTypeInfo wrapFeatureTypeIfNeeded(FeatureTypeInfo ft) {
 
     @Override
     public List<FeatureTypeInfo> getFeatureTypes() {
+        WorkspaceInfo localWorkspace = LocalWorkspace.get();
+        if (localWorkspace != null) {
+            NamespaceInfo ns = super.getNamespaceByPrefix(localWorkspace.getName());
+            if (ns != null) {
+                return getFeatureTypesByNamespace(ns);
```
jodygarnett, line 135: Just to confirm does the method call `wrapFeaturetypeListIfNeeded`?

jodygarnett (changes requested):
Thanks for the contribution, I had one question about use of `wrapFeaturetypeListIfNeeded` below.

Can I ask you to fill in the checklist (especially the bug ticket for our release notes).

src/main/src/main/java/org/geoserver/catalog/impl/LocalWorkspaceCatalog.java
jurgispods: @groldan @jodygarnett I just checked again: Yes,`getFeatureTypesByNamespace` calls `wrapFeaturetypeListIfNeeded` and `getCoveragesByNamespace` calls `wrapCoverageListIfNeeded`.

jurgispods: And about the JIRA ticket: As stated above, I was not able to create one after creating an account. I realized that I might need an invite/permissions to create a ticket, so could anyone make sure I have the proper permissions to do that? The associated e-mail is j***s.p***s@gmail.com.

jurgispods:
Hi @jodygarnett @groldan,

I believe the technical questions have been resolved. Could one of you create a JIRA ticket for me so we can get this PR reviewed and merged?

aaime:
Ticket created. The commit message, and the PR title, should both be:

[GEOS-12215] Do not load complete catalog in workspace-scoped requests

Also please, rebase on top of the main branch and force push here, to get build fixes for that Windows build.

PR #9732, opened 2026-07-22
Docs: Update version selector for consistent presentation of latest and stable

jodygarnett:
Now that I have both `main` and `3.0.x` deploying to https://docs.geoserver.org/ - I notice the version selector is inconsistent. The default is `3.0.0 stable` which is great.

image

The dropdown has different title presentation of `main (Latest) main` and `3.0.x` and `2.28.x' (Maintenance):

image

Recommendation:
```
3.1.x (Latest) main
3.0.x (Stable)
2.28.x (Maintenance)
```

The URL version selector mapping is inconsistent with `3.0.x` and `latest`. The other alias are available by direct link (for example from release announcement):

* `https://docs.geoserver.org/3.0.x/` - Selector
* `https://docs.geoserver.org/stable/` 
* `https://docs.geoserver.org/3.1.x/` 
* `https://docs.geoserver.org/latest/` - Selector
* `https://docs.geoserver.org/main/`
* `https://docs-archive.geoserver.org/2.28.x/en/user/` - Selector
* `https://docs-archive.geoserver.org/maintain/en/user/`

The default has changed between `3.0.x` and `main` over the course of writing so I believe the different branches are changing the default each time they publish!

Option 1: Choose `stable` and `latest` as the version selector targets, with `stable` as default. This is consistent.

Option 2: Choose `3.0.x` and `3.1.x` as the version selector targets, with `3.0.x` as default. This is consistent.

Option 3: choose `3.0.x` and `latest` as version selector targets, with `3.0.x` as the default.

Not sure if `main` provides value as a duplicate of `latest`.

  

# Checklist

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -54,19 +54,14 @@ jobs:
 
           if [ "$BRANCH" = "main" ]; then
             echo "version=latest" >> $GITHUB_OUTPUT
-            echo "title=main (Latest)" >> $GITHUB_OUTPUT
-            echo "aliases=main 3.1.x" >> $GITHUB_OUTPUT
+            echo "title=3.1.x (Latest) main" >> $GITHUB_OUTPUT
+            echo "aliases=3.1.x main" >> $GITHUB_OUTPUT
             echo "is_default=false" >> $GITHUB_OUTPUT
           elif [ "$BRANCH" = "3.0.x" ]; then
-            echo "version=3.0.x" >> $GITHUB_OUTPUT
-            echo "title=3.0.x" >> $GITHUB_OUTPUT
-            echo "aliases=stable" >> $GITHUB_OUTPUT
+            echo "version=stable" >> $GITHUB_OUTPUT
+            echo "title=3.0.x (Stable)" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


jodygarnett:
Alternative (which also makes sense): Choose `3.1.x` and `3.0.x` as the version selector targets:
* This approach provides a stable URL as additional releases
* The logical `stable`, `latest`, and `main` URLs remain available, just not used by default

In my opinion people will find whatever the version selector links to *only* and not notice the alias.

jodygarnett:
Like perhaps it is *fine* the way it is:

* When you want to link to "latest" you intend it to be latest, even three years from now, for example when linking to developer guide policy
* But if you are linking to a specific version, like `3.0.x` you want to stay linking to that exact set of installation instructions

With that in mind we would keep `latest`, and `3.0.x`. If someone in discourse links to the installation instructions chances are they would get the `3.0.x` install instructions, and their message would reflect the link at the time of writing.

Q: The default should be `3.0.x` in this case? 

(When I initially wrote this the default was `3.0.x`, and today it is `main`, so some troubleshooting is required).

jodygarnett (commented):
Option 3

PR #9740, opened 2026-07-24
[GEOS-12159] Fix ClassCastException reprojecting per-feature geometry…

felipecarrillo100:
[GEOS-12159](https://osgeo-org.atlassian.net/browse/GEOS-12159) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

… between 3D CRSs

ReprojectingFeatureCollection.reproject() builds a per-feature coordinate transform (used when a feature's geometry carries its own CRS, distinct from the collection's cached default-source transform) and casts the result to MathTransform2D. That cast is only safe when the transform happens to be 2-dimensional; for a genuine 3D-to-3D operation between two Geographic 3D CRSs (e.g. EPSG:4979 and EPSG:4937, where latitude, longitude and ellipsoidal height are all axes of the CRS itself, not a Z ordinate riding along on the geometry), the transform is not a MathTransform2D and the cast throws.

The constructor's own default-source transform (a few lines earlier in the same class) already performs the identical lookup without this cast, since GeometryCoordinateSequenceTransformer.setMathTransform() accepts a plain MathTransform. The per-feature path narrows the type for no reason.

Added a regression test reproducing the ClassCastException against two real Geographic 3D CRSs before the fix, passing after it.

  

# Checklist

### DIFF
--- a/src/main/src/test/java/org/geoserver/feature/ReprojectingFeatureCollectionTest.java
+++ b/src/main/src/test/java/org/geoserver/feature/ReprojectingFeatureCollectionTest.java
@@ -47,20 +47,7 @@ public class ReprojectingFeatureCollectionTest {
         }
     }
 
-    /**
-     * reproject(SimpleFeature) resolves a per-feature transform (via the geometry's own user data CRS, not the
-     * collection's cached default-source transform) through a code path that casts the result to MathTransform2D. That
-     * cast is safe whenever the transform happens to be 2-dimensional, but throws ClassCastException for a genuine
-     * 3D-to-3D coordinate operation - e.g. between two different Geographic 3D CRSs (EPSG:4979 WGS84 3D and EPSG:4937
-     * ETRS89 3D here), where latitude, longitude and ellipsoidal height are all axes of the CRS itself, not a 2D CRS
-     * with a Z ordinate merely riding along on the geometry.
-     *
-     * <p>This is exactly what happens in practice when a WFS-T feature's geometry carries an explicit CRS (e.g. one a
-     * client resolved from a GetCapabilities document) that differs from the feature type's own declared CRS, for a
-     * genuinely 3-dimensional CRS pair - the per-feature transformer has to be built for the first time (a cache miss),
-     * taking this cast path, rather than reusing the constructor's own already-cached, correctly uncast transformer for
-     * the feature type's declared CRS.
-     */
+    /** Per-feature reprojection between two 3D CRSs must not throw ClassCastException (GEOS-12159). */
     @Test
     public void testReprojectPerFeatureCRSWithGenuine3DCRS() throws Exception {
         CoordinateReferenceSystem geometryCrs = CRS.decode("EPSG:4979"); // WGS84 3D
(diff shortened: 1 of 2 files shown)


src/main/src/test/java/org/geoserver/feature/ReprojectingFeatureCollectionTest.java
```diff
@@ -44,4 +46,56 @@ public void testPerserveUserData() throws Exception {
...
+     * ETRS89 3D here), where latitude, longitude and ellipsoidal height are all axes of the CRS itself, not a 2D CRS
+     * with a Z ordinate merely riding along on the geometry.
+     *
+     * <p>This is exactly what happens in practice when a WFS-T feature's geometry carries an explicit CRS (e.g. one a
+     * client resolved from a GetCapabilities document) that differs from the feature type's own declared CRS, for a
+     * genuinely 3-dimensional CRS pair - the per-feature transformer has to be built for the first time (a cache miss),
+     * taking this cast path, rather than reusing the constructor's own already-cached, correctly uncast transformer for
+     * the feature type's declared CRS.
+     */
+    @Test
```
jodygarnett, line 51: This is very verbose for a test case! If using AI please ask it to keep comments to a minimum (so we can focus on code not reading).

src/main/src/test/java/org/geoserver/feature/ReprojectingFeatureCollectionTest.java
```diff
@@ -44,4 +46,56 @@ public void testPerserveUserData() throws Exception {
...
+        // priming step would satisfy reproject()'s cache lookup before ever reaching its own,
+        // separately-cast, per-feature transformer construction - masking the bug entirely.
+        tb.setSRS("EPSG:4937");
+        tb.add("geom", Point.class);
+
+        SimpleFeatureBuilder b = new SimpleFeatureBuilder(tb.buildFeatureType());
+        Point point = (Point) new WKTReader().read("POINT(10 20 123)");
+        // The per-feature CRS on the geometry's own user data - not the feature type's CRS -
+        // is what reproject() actually consults first (see its own javadoc).
+        point.setUserData(geometryCrs);
```
jodygarnett, line 65: For in-line comments also!

jodygarnett (changes requested):
Thanks, I believe the code is fine, the comments are a bit verbose for this codebase.

felipecarrillo100:
I have reduced the comments verbose.  Thanks for taking the time to review

aaime:
Since it's related, we approved an AI policy a few days ago, here is the proposal: https://github.com/geoserver/geoserver/wiki/GSIP-242
We are going to link from the PR template checklist.

felipecarrillo100:
Following up on my earlier email:


I realized my IDE hadn't successfully pushed my changes, so, you didn't see
any change! .

Everything is properly committed and pushed now.  Long comments have been
condensed to 1 line.

The code fix itself is just a few lines,

I used AI to generate the unit tests but if you see the impact to the code
itself is minimalistic and makes sense .

The fiix removes an unnecessary MathTransform2D cast on the per-feature
transform, matching the type already used a few lines earlier in the
constructor — This is a minimalistic change (2-lines) with no other logic
touched.

The AI was used to create a new regression test reproduces the exact
ClassCastException between two real 3D CRSs (EPSG:4979 → EPSG:4937) before
the fix and passes after it, with no other tests in the class affected.

Please let me know if everything looks good on your end this time!

aaime:
Looks good to me, the two build failures are due to erratic connection to repo.osgeo.org.

PR #9752, opened 2026-07-28
GeoFence: align to GS3

etj:
Make GeoFence work in GeoServer 3.
Uses new branch in GeoFence 3 `geofence_39`
  

# Checklist


etj:
@axl8713 as of latest commit, `geofence-server` boots and runs properly.

A note about pom and dependencies:
- `ehcache`: the hibernate version set in the pom requires `org.ehcache:ehcache:3.10.9:jakarta`, while `gt-geotiff` requires  `imageio-ext-cog-reader` --> `imageio-ext-cog-streams` --> `org.ehcache:ehcache:3.4.0`. 
   Ref: https://github.com/geosolutions-it/imageio-ext/blob/master/plugin/cog/pom.xml#L27
   Version 3.4.0 is quite old and internally references `javax.xml.bind` while the `3.10.x ` uses `jakarta.xml.bind`.
   If possible we should upgrade it.    
- `antlr`: Hibernate 7.4.5.Final needs `antlr4-runtime:4.13.2` but `gs-web-app`'s classpath resolves 4.7.1 via `gs-main` -->Eclipse Imagen's Jiffle. 
Pinned `org.antlr:antlr4-runtime` to 4.13.2 in `src/pom.xml''s `dependencyManagement`, right next to the existing Hibernate pins.
  **Jiffle provides code compiled with 4.7.1, so version is not negotiable. WIP**

etj:
New testable version:

- GF: `d188820 (HEAD -> geofence_39, origin/geofence_39) Refact out model from services API`
- GS: `be9ffad4b0 (HEAD -> geofence_build_fix_gs3, origin/geofence_build_fix_gs3) ANTLR fixes; restore assembly; model/services-api refactoring`

jodygarnett:
Hey @etj I am looking ahead to 3.0.1 release this month, this is one of the key features remaining for GS3.

(Many of the failures are due to osgeo repo timing out - so I need to ask you how this is going)

etj:
Hi @jodygarnett , I'm ironing out the last errors, I hope to have the PR ready for review within a week. Thanks for the heads up!

src/extension/geofence/geofence-server/src/main/java/org/geoserver/geofence/server/rest/RulesRestController.java
```diff
@@ -247,47 +245,47 @@ public String insert(@RequestBody(required = true) JaxbRule rule) {
...
+                adminService().shift(rule.getPriority(), 1);
             }
         }
-        Rule theRule = adminService.get(id);
-        adminService.update(rule.toRule(theRule));
+        Rule theRule = adminService().get(id);
+        adminService().update(rule.toRule(theRule));
         if (rule.getLimits() != null) {
-            adminService.setLimits(id, rule.getLimits().toRuleLimits(theRule.getRuleLimits()));
+            adminService().setLimits(id, rule.getLimits().toRuleLimits(null));
```
axl8713: This clears the previous value. Is this intentional?

src/extension/geofence/geofence-server/src/main/java/org/geoserver/geofence/server/rest/xml/JaxbRule.java
```diff
@@ -258,9 +257,8 @@ public void setAttributes(Set<LayerAttribute> layerAttributes) {
             this.layerAttributes = layerAttributes;
         }
 
-        public org.geoserver.geofence.core.model.LayerDetails toLayerDetails(
-                org.geoserver.geofence.core.model.LayerDetails details) {
-            details = new org.geoserver.geofence.core.model.LayerDetails();
+        public org.geofence.core.model.LayerDetails toLayerDetails(org.geofence.core.model.LayerDetails details) {
+            details = new org.geofence.core.model.LayerDetails();
```
axl8713: Should this follow the same null-check pattern as `toRuleLimits` above? Right now it unconditionally rewrites details, ignoring the parameter entirely.

src/extension/geofence/geofence-server/src/main/java/org/geoserver/geofence/server/rest/RulesRestController.java
```diff
@@ -247,47 +245,47 @@ public String insert(@RequestBody(required = true) JaxbRule rule) {
...
         }
         if (rule.getLayerDetails() != null) {
-            adminService.setDetails(id, rule.getLayerDetails().toLayerDetails(theRule.getLayerDetails()));
+            adminService().setDetails(id, rule.getLayerDetails().toLayerDetails(null));
```
axl8713: This clears the previous value, but check this first: https://github.com/geoserver/geoserver/pull/9752/changes#r3871124793

src/extension/geofence/geofence/src/main/java/org/geoserver/geofence/services/RestRuleReaderService.java
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    @Override
+    public PermsResult getPermissionFilter(RuleFilter filter) {
+        RESTPermsResult restResult = ruleReaderClient().getPermissionFilter(toRestFilter(filter));
+        return toPermsResult(restResult);
+    }
+
+    public void setServiceUrl(String serviceUrl) {
+        this.updatedUrl = serviceUrl;
```
axl8713, line 87: `updatedUrl` is written here without synchronization; `ruleReaderClient()` reads it from concurrent request threads with no shared lock. After an admin changes the URL, in-flight requests can keep using the stale one indefinitely. 
I would suggest using volatile on `updatedUrl`/`currentUrl`, or a shared lock with `ruleReaderClient()`.

src/extension/geofence/geofence/src/main/java/org/geoserver/geofence/cache/CacheManager.java
```diff
@@ -38,7 +40,7 @@ public class CacheManager {
     private ContainerAccessCacheLoaderFactory containerAccessCacheLoaderFactory;
 
     private LoadingCache<RuleFilter, AccessInfo> ruleCache;
-    private LoadingCache<NamePw, AuthUser> userCache;
+    private LoadingCache<RuleFilter, PermsResult> permCache;
     private LoadingCache<RuleFilter, AccessInfo> authCache;
     private LoadingCache<ContainerAccessCacheLoaderFactory.ResolveParams, ContainerLimitResolver.ProcessingResult>
             contCache;
```
axl8713: Since `init()` is called at each save and recreates these caches, they can have stale values since there is no guarantee of an update prior to a read. 
I  would suggest to set these as volatile.

src/extension/geofence/geofence/src/main/java/org/geoserver/geofence/GeofenceAccessManager.java
```diff
@@ -971,7 +978,44 @@ private void checkStyleAllowed(AccessInfo accessInfo, String styleName) {
...
+        } catch (Exception e) {
+            LOGGER.log(Level.SEVERE, "Error retrieving permissions for filter " + ruleFilter + ", blocking access", e);
+            return Filter.EXCLUDE;
+        }
+
+        if (permsResult == null) {
+            LOGGER.log(Level.SEVERE, "Unexpected null PermsResult for filter {0}, blocking access", ruleFilter);
+            return Filter.EXCLUDE;
+        }
+        return PermissionCatalogFilterHelper.buildCatalogFilter(permsResult, clazz);
```
axl8713, line 999: I think it is worth adding these lines to the above `try` block as well.

src/extension/geofence/geofence/src/main/java/org/geoserver/geofence/services/RestRuleReaderService.java
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    private synchronized org.geofence.web.rest.api.interfaces.RESTRuleReaderService ruleReaderClient() {
+        if (client == null || !updatedUrl.equals(currentUrl)) {
```
axl8713: Potential NPE on `updatedUrl` if `client` is not null.

src/extension/geofence/geofence/src/main/java/org/geoserver/geofence/config/GeoFenceSpringConfig.java
```diff
@@ -0,0 +1,170 @@
...
+
+    @Override
+    public void setApplicationContext(ApplicationContext context) {
+        this.context = context;
+    }
+
+    /**
+     * Reads {@code geofence/geofence.properties} - same file {@link GeoFenceConfigurationManager} reads/saves through.
+     */
+    private Properties loadProperties(GeoServerDataDirectory dataDirectory) {
```
axl8713: Since this method is called several times, could be worth creating a property bean for it?

src/extension/geofence/geofence/src/main/java/org/geoserver/geofence/cache/CacheManager.java
axl8713: Is this really needed? There's already the constructor calling `init()`.

axl8713 (changes requested):

etj:
Fixes according to the review:

- `RestRuleReaderService.setServiceUrl()` -- synchronized, so a URL update is guaranteed visible to `ruleReaderClient()` instead of possibly stale forever.
- `RestRuleReaderService.ruleReaderClient()` -- null-safe `Objects.equals(updatedUrl, currentUrl)`, fixing an NPE when `updatedUrl` is null on a throwaway/no-`configManager` instance.
- `CacheManager.init()` -- removed redundant `@PostConstruct` (constructor already calls it; was double-building all 4 caches on every real startup).
- `CacheManager` -- made `ruleCache`/`permCache`/`authCache`/`contCache`/`cacheConfiguration` `volatile`, fixing a stale-read race between config-save `init()` calls and concurrent cache-getter reads.
- `RulesRestController.update()` -- restored `theRule.getRuleLimits()`/`theRule.getLayerDetails()` (was `null`), fixing a partial-POST-update regression that silently wiped unset `RuleLimits`/`LayerDetails` fields.
- `JaxbRule.Limits.toLayerDetails()` -- fixed to actually merge into its `details` parameter instead of discarding it and always building fresh; deeper, pre-existing version of the same partial-update bug.
- Added `RulesRestControllerTest.testLimitsPartialUpdatePreservesUnsetFields`/`testLayerDetailsPartialUpdatePreservesUnsetFields` -- regression coverage for both fixes above.
- `GeoFenceSpringConfig` -- `loadProperties()` -> `geoFenceProperties` Spring bean, so `geofence.properties` is read once and shared instead of independently re-read by 4 different `@Bean` methods.

src/extension/geofence/geofence-server/src/main/java/org/geoserver/geofence/server/config/GeoServerDatasourcePasswordDecoder.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Override
+    public Result decode(String storedPassword) {
+        if (storedPassword == null) {
+            return Result.asIs(null);
+        }
+        String plaintext = passwordHelper.decode(storedPassword);
+        if (isPlainMarked(storedPassword)) {
+            return Result.persist(plaintext, passwordHelper.encode(plaintext));
+        }
```
axl8713, line 37: Setting geofence.datasource.password=plain:mysecret and restarting strips the plain: prefix but never rewrites it as crypt1:... and it is persisted in plaintext.

Root cause: `ConfigurationPasswordEncryptionHelper.encode()` silently no-ops when getConfigPasswordEncrypterName() returns null (just logs "Encryption disabled, no password encoder set"). 
Confirmed live: a WARNING fires at startup even though `config.xml` has pbePasswordEncoder configured; GeoServerSecurityManager's config isn't loaded yet when GefencePersistenceConfig's constructor eagerly triggers this decode/encode cycle.

etj:
[e3177f3](https://github.com/geoserver/geoserver/pull/9752/commits/e3177f3ee9abe9aa87433f90af443436c186b817)
- GeoFence honours GeoServer config reloading
- Improve/fix encryption
- Improve properties files handling
- Improve visualization of badly configured GeoFence

src/extension/geofence/geofence/src/main/java/org/geoserver/geofence/GeofenceAccessManager.java
axl8713: Unresolving this as it is still relevant.

axl8713 (changes requested):

etj:
Superseded by https://github.com/geoserver/geoserver/pull/9923

PR #9756, opened 2026-07-29
[GEOS-12165] GeoServer does not start when fileLockProvider is set: lock wait times out

fernandor777:
[GEOS-12165](https://osgeo-org.atlassian.net/browse/GEOS-12165) [Powered by Pull Request Badge](https://pullrequestbadge.com/?utm_medium=github&utm_source=geoserver&utm_campaign=badge_info)  

https://osgeo-org.atlassian.net/browse/GEOS-12165

This PR refactors the FileLockProvider class to allow same thread re-entrant lock acquisition.
  

# Checklist

### DIFF
--- a/src/platform/src/main/java/org/geoserver/platform/resource/FileLockProvider.java
+++ b/src/platform/src/main/java/org/geoserver/platform/resource/FileLockProvider.java
@@ -19,6 +19,7 @@ import java.nio.channels.OverlappingFileLockException;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -48,6 +49,7 @@ public class FileLockProvider implements LockProvider, ServletContextAware {
 
     private static final class LockEntry {
         final ReentrantLock gate = new ReentrantLock();
+        final AtomicInteger counter = new AtomicInteger(0);
         FileOutputStream fos;
         FileLock lock;
     }
@@ -80,7 +82,11 @@ public class FileLockProvider implements LockProvider, ServletContextAware {
         // below won't do that on its own: FileLock is documented as unsuitable for excluding
         // threads within the same JVM) and, being a ReentrantLock, also lets the thread that
         // already holds it re-enter without blocking
-        final LockEntry entry = lockEntries.computeIfAbsent(lockKey, k -> new LockEntry());
+        final LockEntry entry = lockEntries.compute(lockKey, (k, e) -> {
+            if (e == null) e = new LockEntry();
+            e.counter.incrementAndGet();
+            return e;
+        });
         final ReentrantLock gate = entry.gate;
         acquireGate(lockKey, gate);
 
@@ -170,7 +176,7 @@ public class FileLockProvider implements LockProvider, ServletContextAware {
                 }
             }
             IOUtils.closeQuietly(currFos);
-            lockEntries.remove(lockKey, entry);
+            detach(lockKey, entry);
             gate.unlock();
             throw (e instanceof RuntimeException) ? (RuntimeException) e : new IllegalStateException(e);
         }
@@ -190,6 +196,8 @@ public class FileLockProvider implements LockProvider, ServletContextAware {
                         releaseFileLock(lockKey, entry, file);
                     }
                 } finally {
+                    // Call detach() on every release
+                    detach(lockKey, entry);
                     gate.unlock();
                 }
             }
@@ -215,12 +223,24 @@ public class FileLockProvider implements LockProvider, ServletContextAware {
                 // should not happen: by the time release() calls into here, entry.lock was always
                 // set by this same thread's earlier successful acquire. Logged rather than thrown
                 // since lock usage is only there to prevent duplication of work, not correctness.
-                LOGGER.warning("Lock key " + lockKey + " had no valid OS lock to release; this should not happen");
+                LOGGER.warning(("Lock %s has no valid OS lock to release, current thread is: %d. This means the "
+                                + "lock was never acquired, or was released twice. Are you running two instances "
+                                + "in the same JVM using NIO locks? This case is not supported and will generate "
+                                + "exactly this error message")
+                        .formatted(lockKey, Thread.currentThread().getId()));
             }
         } catch (IOException e) {
             throw new IllegalStateException("Failure releasing lock " + lockKey, e);
-        } finally {
-            lockEntries.remove(lockKey, entry);
+        }
+    }
+
+    /**
+     * Undoes one {@code acquire()} call. Removes {@code entry} from the map once no caller is still attached to it
+     * (mirrors {@code MemoryLockProvider.LockAndCounter}).
+     */
+    private void detach(String lockKey, LockEntry entry) {
+        if (entry.counter.decrementAndGet() == 0) {
+            lockEntries.compute(lockKey, (k, e) -> (e == null || e.counter.get() == 0) ? null : e);
         }
     }
 
--- a/src/platform/src/main/java/org/geoserver/platform/resource/FileLockProvider.java
+++ b/src/platform/src/main/java/org/geoserver/platform/resource/FileLockProvider.java
@@ -88,7 +88,14 @@ public class FileLockProvider implements LockProvider, ServletContextAware {
             return e;
         });
         final ReentrantLock gate = entry.gate;
-        acquireGate(lockKey, gate);
+        try {
+            acquireGate(lockKey, gate);
+        } catch (RuntimeException e) {
+            // never got the gate, so this call never touches entry.lock/entry.fos; only undo
+            // the counter increment done above, so the entry doesn't leak in the map forever
+            detach(lockKey, entry);
+            throw e;
+        }
 
         final File file = getFile(lockKey);
         if (gate.getHoldCount() > 1) {
@@ -182,7 +189,7 @@ public class FileLockProvider implements LockProvider, ServletContextAware {
         }
     }
 
-    /** No finally block deleting the file here, it's done in the returned lock's {@code release()}. */
+    /** The file removal is done in returned lock's {@code release()}. */
     private Resource.Lock newLock(String lockKey, LockEntry entry, File file, ReentrantLock gate) {
         return new Resource.Lock() {
             boolean released;
@@ -190,6 +197,9 @@ public class FileLockProvider implements LockProvider, ServletContextAware {
             @Override
             public void release() {
                 if (released) return;
+                if (!gate.isHeldByCurrentThread()) {
+                    throw new IllegalStateException("Lock " + lockKey + " must be released by the acquiring thread");
+                }
                 released = true;
                 try {
                     if (gate.getHoldCount() == 1) {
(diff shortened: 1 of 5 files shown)


src/platform/src/main/java/org/geoserver/platform/resource/FileLockProvider.java
```diff
@@ -38,7 +44,13 @@ public class FileLockProvider implements LockProvider, ServletContextAware {
     /** The wait to occur in case the lock cannot be acquired */
     int waitBeforeRetry = 20;
 
-    MemoryLockProvider memoryProvider = new MemoryLockProvider();
+    private final ConcurrentMap<String, LockEntry> lockEntries = new ConcurrentHashMap<>();
+
+    private static final class LockEntry {
```
aaime, line 50: There is no reference count here, this can cause serious problems (that's why the memory lock provider has reference counting).

Example:

  1. T1 computeIfAbsent(K) -> E1 (entry1) is generated, takes E1.gate, takes OS lock, stores in E1.
  2. T2 computeIfAbsent(K) -> same E1, blocks in gate.tryLock (:104).
  3. T1 releases: releaseFileLock drops OS lock and removes E1 from map in its finally (:223), then gate.unlock() (:193). Removal before unlock — so at the moment T2 wakes, E1 is already unreachable.
  4. T2 owns E1.gate, holdCount 1, takes OS lock, stores it in E1. E1 is not in the map.

From here it can go wrong in different ways, e.g.:

  a) T2 nests on itself (the case we're trying to fix). T2 calls acquire(K) again: computeIfAbsent finds nothing, creates E2 with a fresh gate. getHoldCount() on E2 is 1, not 2  so it thinks this is a first acquire and goes to acquireOsLock, colliding with the FileLock T2 itself holds through E1. OverlappingFileLockException, retry loop, timeout. The hang is back.

  b) T3 arrives. Gets a new E2 (E1 is not in the map anymore), uncontended gate, then the same OS-lock collision with T2. Mutual exclusion still holds (it fails rather than letting both in), but the gate no longer queues anyone: T3 spins on OverlappingFileLockException and errors out instead of waiting its turn.

Here is a test that might reproduce the problem:

```java
package org.geoserver.platform.resource;

  import static org.junit.Assert.assertNull;
  import java.io.File;
  import java.util.concurrent.*;
  import org.junit.Rule;
  import org.junit.Test;
  import org.junit.rules.TemporaryFolder;
      
  public class OrphanRaceTest {
      @Rule
      public TemporaryFolder tempFolder = new TemporaryFolder();
      
      @Test
      public void testNestedAfterWaiterHandoff() throws Exception {
          File root = tempFolder.newFolder("lockRoot");
          FileLockProvider provider = new FileLockProvider(root, 2);
          String key = "k";
          Resource.Lock a = provider.acquire(key);
          CountDownLatch bStarted = new CountDownLatch(1);
          ExecutorService es = Executors.newSingleThreadExecutor();
          Future<Throwable> f = es.submit(() -> {
              bStarted.countDown();
              Resource.Lock outer = provider.acquire(key); // waits for A
              try {
                  Resource.Lock inner = provider.acquire(key); // nested
                  inner.release();
                  return null;
              } catch (Throwable t) {
                  return t;
              } finally {
                  outer.release();                                                          
              }
          });
          bStarted.await();
          Thread.sleep(300); // make sure B is queued on the gate
          a.release();
          Throwable t = f.get(30, TimeUnit.SECONDS);
          es.shutdownNow();
          assertNull("nested acquire after handoff failed: " + t, t);
      }
  }
```

If you can reproduce with the above (I did) I would have a serious second look at the testing class in this PR, it's not really testing what it should.

src/platform/src/main/java/org/geoserver/platform/resource/FileLockProvider.java
```diff
@@ -165,10 +170,58 @@ public void release() {
...
+
+                if (LOGGER.isLoggable(Level.FINE)) {
+                    LOGGER.fine("Lock " + lockKey + " mapped onto " + file + " released by thread "
+                            + Thread.currentThread().getId());
+                }
+            } else {
+                // should not happen: by the time release() calls into here, entry.lock was always
+                // set by this same thread's earlier successful acquire. Logged rather than thrown
+                // since lock usage is only there to prevent duplication of work, not correctness.
+                LOGGER.warning("Lock key " + lockKey + " had no valid OS lock to release; this should not happen");
```
aaime: The old log was quite a bit more comprehensive in terms of possible causes, and shared the thread id:

```java
     if (LOGGER.isLoggable(Level.FINE)) {
          LOGGER.fine(("Lock key %s for releasing lock is unknown, it means this lock was never"
                          + " acquired, or was released twice. Current thread is: %d. Are you"
                          + " running two instances in the same JVM using NIO locks? This case is"
                          + " not supported and will generate exactly this error message")
                  .formatted(lockKey, Thread.currentThread().getId()));
      }
```

I'd suggest restoring it.

aaime (changes requested):
As far as I can tell the fix is not fully working, the issue can still be reproduced in some circumstances, see below.

src/platform/src/main/java/org/geoserver/platform/resource/FileLockProvider.java
fernandor777: indeed, the issue is reproduced.  Checkint the added tests, looks like they are only checking the single thread nested scenario, lacking of the multi-thread lock passing scenario.  I'm adding this test to the test class along with the fixes (adding the counter logic, as the memory implementation one).

src/platform/src/main/java/org/geoserver/platform/resource/FileLockProvider.java
fernandor777: Agreed, restoring it.

fernandor777:
Thanks @aaime for the detailed feedback.  You are right, the multithread lock passing scenario had a bug.  I replicated it with the suggested test and added the fix as a new commit.
Please let me know if something else is missing in order to make progress on this PR.

src/platform/src/main/java/org/geoserver/platform/resource/FileLockProvider.java
```diff
@@ -63,14 +77,53 @@ public FileLockProvider(File basePath) {
...
+        // below won't do that on its own: FileLock is documented as unsuitable for excluding
+        // threads within the same JVM) and, being a ReentrantLock, also lets the thread that
+        // already holds it re-enter without blocking
+        final LockEntry entry = lockEntries.compute(lockKey, (k, e) -> {
+            if (e == null) e = new LockEntry();
+            e.counter.incrementAndGet();
+            return e;
+        });
+        final ReentrantLock gate = entry.gate;
+        acquireGate(lockKey, gate);
```
aaime: In case this times out or throws an exception, the counter is not decreased, meaning the entry won't be cleared from the map. This won't cause locking issues, but it will eventually turn into a memory leak, as new files are locked.

Suggested could look like this:

```
  final ReentrantLock gate = entry.gate;
  try {
      acquireGate(lockKey, gate);
  } catch (RuntimeException e) {
      detach(lockKey, entry);
      throw e;
  }
```

To make it testable, a package-private accessor is enough — nicer than reflection from the test, since the test is in the same package:

```
/** Number of keys currently tracked, zero when every acquisition has been released. For tests. */
int getTrackedKeyCount() {
    return lockEntries.size();
}
```

and then test wise, something like this:

```java
@Test
public void testNoEntryLeakOnAcquireTimeout() throws Exception {
    String key = "timeout-leak-key";
    Resource.Lock held = provider.acquire(key);
    ExecutorService executor = Executors.newSingleThreadExecutor();
    try {
        Future<Boolean> timedOut = executor.submit(() -> {
            try {
                provider.acquire(key).release();
                return false;
            } catch (IllegalStateException e) {
                return true;
            }
        });
        assertTrue("Acquire should have timed out", timedOut.get(60, SECONDS));
    } finally {
        executor.shutdownNow();
    }
    held.release();
    assertEquals("Timed out acquire leaked its entry", 0, provider.getTrackedKeyCount());
}

@Test
public void testNoEntryLeakOnInterruptedAcquire() throws Exception {
    String key = "interrupt-leak-key";
    Resource.Lock held = patientProvider.acquire(key);
    Thread waiter = new Thread(() -> {
        try {
            patientProvider.acquire(key).release();
        } catch (IllegalStateException expected) {
            // interrupted while waiting on the gate
        }
    });
    waiter.start();
    waiter.interrupt();
    waiter.join(60_000);
    assertFalse("Waiter did not terminate", waiter.isAlive());
    held.release();                                                                     
    assertEquals("Interrupted acquire leaked its entry", 0, patientProvider.getTrackedKeyCount());
}
```

Ah for the patient provider see the comments on the test.
To be extra safe, maybe also add assertions after each test:

```java
@After
public void assertNoLeakedEntries() {
    assertEquals("Leaked lock entries", 0, provider.getTrackedKeyCount());
    assertEquals("Leaked lock entries", 0, patientProvider.getTrackedKeyCount());
}
```

src/platform/src/main/java/org/geoserver/platform/resource/FileLockProvider.java
```diff
@@ -165,10 +176,72 @@ public void release() {
...
+    private Resource.Lock newLock(String lockKey, LockEntry entry, File file, ReentrantLock gate) {
+        return new Resource.Lock() {
+            boolean released;
+
+            @Override
+            public void release() {
+                if (released) return;
+                released = true;
+                try {
+                    if (gate.getHoldCount() == 1) {
```
aaime, line 205: This makes sense, there is a potential issue... what if the lock is passed around to a different thread that is tasked to unlock? It won't work.
I would make sure to guard against it, maybe something like this:

```
if (!gate.isHeldByCurrentThread()) throw new IllegalStateException("Lock " + lockKey + " must be released by the acquiring thread");
```

just before checking the hold coun.

src/platform/src/main/java/org/geoserver/platform/resource/FileLockProvider.java
```diff
@@ -165,10 +176,72 @@ public void release() {
...
             IOUtils.closeQuietly(currFos);
-            memoryLock.release(); // Must release memory lock on failure
+            detach(lockKey, entry);
+            gate.unlock();
             throw (e instanceof RuntimeException) ? (RuntimeException) e : new IllegalStateException(e);
         }
-        // Note: No finally block deleting the file here, it's done in the returned lock
+    }
+
+    /** No finally block deleting the file here, it's done in the returned lock's {@code release()}. */
```
aaime:
```suggestion
    /** The file removal is done in returned lock's {@code release()}. */
```

src/platform/src/main/java/org/geoserver/platform/resource/FileLockProvider.java
aaime: Btw, the same issue is happening in the memory lock provider, mind fixing that one too?

src/platform/src/test/java/org/geoserver/platform/resource/FileLockProviderTest.java
```diff
@@ -123,6 +126,152 @@ public void testInterruptionDuringAcquisition() throws Exception {
...
+
+        assertEquals("hello", IOUtils.toString(resource.in(), StandardCharsets.UTF_8));
+    }
+
+    /**
+     * A thread wait for the lock, then get it right after the other thread releases it. This thread must then be able
+     * to acquire the same lock again, like any other thread that holds the lock.
+     */
+    @Test
+    public void testNestedAcquireAfterWait() throws Exception {
```
aaime, line 271: I'm a bit worried about this test on github actions... those enviroments tend to be very erratic, and the file lock provider for the test has a timeout of only 1 second. Do you need a second provider with a higher timeout of this test? Any other test that might be using it (one of those that I suggested above migth need it for example)?
I'd call it a "patientProvider", it would be a new file, hence the reference in my other comment.

aaime (changes requested):

fernandor777:
Thanks @aaime all the comments applied, includiung the fix on the memory provider.

aaime:
Looks good, squash merged into main, the bot will hopefully take care of the backport PRs, if not, please open them manually.

